PR #16, opened 2017-12-05
Implemented packed binary formats

jadams-tresys:
Implemented parsers and unparsers for Packed Decimal, BCD, and IBM4690
Packed Decimal.  Added some tests to verify the parsing of these formats
and enabled a test that we already have that now works.

For the delimited parsers of these formats, they are currently limited
to only supporting ISO-8895-1 encoding, just like hexBinary (see
DAFFODIL-1541).

There are still some issues that are preventing the IBM TLOG schemas
from parsing correctly, but they are part of a larger bug in the
Daffodil codebase regarding delimited data and handling separators.

DFDL-211, DFDL-629, DFDL-1739

### DIFF
--- a/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -43,6 +43,7 @@ import edu.illinois.ncsa.daffodil.dpath.NodeInfo
 import edu.illinois.ncsa.daffodil.dsom.ExpressionCompilers
 import edu.illinois.ncsa.daffodil.xml.GlobalQName
 import edu.illinois.ncsa.daffodil.xml.XMLUtils
+import edu.illinois.ncsa.daffodil.util.PackedSignCodes
 import edu.illinois.ncsa.daffodil.grammar.primitives.InputValueCalc
 import edu.illinois.ncsa.daffodil.grammar.primitives.ZonedTextIntPrim
 import edu.illinois.ncsa.daffodil.grammar.primitives.PaddingInfoMixin
@@ -106,6 +107,24 @@ import edu.illinois.ncsa.daffodil.grammar.primitives.BinaryFloat
 import edu.illinois.ncsa.daffodil.grammar.primitives.BinaryDouble
 import edu.illinois.ncsa.daffodil.grammar.primitives.BinaryDecimalRuntimeLength
 import edu.illinois.ncsa.daffodil.grammar.primitives.BinaryDecimalKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedIntegerRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedIntegerKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedIntegerDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedDecimalDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedDecimalRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.PackedDecimalKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedIntegerRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedIntegerKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedIntegerDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedDecimalDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedDecimalRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.IBM4690PackedDecimalKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDIntegerRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDIntegerKnownLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDIntegerDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDDecimalDelimitedEndOfData
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDDecimalRuntimeLength
+import edu.illinois.ncsa.daffodil.grammar.primitives.BCDDecimalKnownLength
 import edu.illinois.ncsa.daffodil.grammar.primitives.DynamicEscapeSchemeCombinatorElement
 import edu.illinois.ncsa.daffodil.grammar.primitives.DelimiterStackCombinatorElement
 import edu.illinois.ncsa.daffodil.grammar.primitives.LogicalNilValue
@@ -300,7 +319,10 @@ trait ElementBaseGrammarMixin
           case Float => true
           case Double => true;
           case n: Numeric.Kind =>
-            binaryNumberRep eq BinaryNumberRep.Binary
+            binaryNumberRep match {
+              case BinaryNumberRep.Binary => true
+              case _ => false
+            }
           case _ => true
         }
       } &&
@@ -475,11 +497,21 @@ trait ElementBaseGrammarMixin
     case _ => schemaDefinitionError("Size of binary data '" + primType.name + "' cannot be determined implicitly.")
   }
 
+  /**
+   * Property consistency check for called for all binary numbers
+   *
+   * Returns -1 if the binary number is not of known length,or is of a
+   * lengthKind inconsistent with knowing the length.
+   *
+   * SDE if the lengthKind is inconsistent with binary numbers, not yet implemented
+   * for binary numbers, or not supported by Daffodil.
+   */
   private lazy val binaryNumberKnownLengthInBits: Long = lengthKind match {
     case LengthKind.Implicit => implicitBinaryLengthInBits
     case LengthKind.Explicit if (lengthEv.isConstant) => explicitBinaryLengthInBits()
     case LengthKind.Explicit => -1 // means must be computed at runtime.
-    case LengthKind.Delimited => subsetError("lengthKind='delimited' not yet supported.")
+    case LengthKind.Delimited if (binaryNumberRep == BinaryNumberRep.Binary) => subsetError("lengthKind='delimited' only supported for packed binary formats.")
+    case LengthKind.Delimited => -1 // only for packed binary data, length must be computed at runtime.
     case LengthKind.Pattern => schemaDefinitionError("Binary data elements cannot have lengthKind='pattern'.")
     case LengthKind.Prefixed => subsetError("lengthKind='prefixed' not yet supported.")
     case LengthKind.EndOfParent => schemaDefinitionError("Binary data elements cannot have lengthKind='endOfParent'.")
@@ -675,20 +707,15 @@ trait ElementBaseGrammarMixin
     }
   }
 
-  // This is the right name that the DFDL property should have had!
-  private lazy val binaryIntRep = {
-    subset(binaryNumberRep == BinaryNumberRep.Binary, "binaryNumberRep='%s' is unsupported. Only 'binary' is supported.", binaryNumberRep.toString)
-    binaryNumberRep
-  }
-
   private lazy val staticBinaryFloatRep = {
     subset(binaryFloatRepEv.isConstant, "Dynamic binaryFloatRep is not supported.")
     binaryFloatRepEv.optConstant.get
   }
 
-  val bin = BinaryNumberRep.Binary // shorthands for table dispatch
   val ieee = BinaryFloatRep.Ieee
   type BO = java.nio.ByteOrder
+
+  private lazy val packedSignCodes = PackedSignCodes(binaryPackedSignCodes, binaryNumberCheckPolicy)
 
   private def binaryIntegerValue(isSigned: Boolean) = {
     //
@@ -699,10 +726,26 @@ trait ElementBaseGrammarMixin
         binaryNumberKnownLengthInBits > 8)) {
       byteOrderRaw // must be defined or SDE
     }
-    Assert.invariant(binaryIntRep == bin)
-    binaryNumberKnownLengthInBits match {
-      case -1 => new BinaryIntegerRuntimeLength(this, isSigned)
-      case _ => new BinaryIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+    (binaryNumberRep, lengthKind, binaryNumberKnownLengthInBits) match {
+      case (BinaryNumberRep.Binary, _, -1) => new BinaryIntegerRuntimeLength(this, isSigned)
+      case (BinaryNumberRep.Binary, _, _) => new BinaryIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+      case (_, LengthKind.Implicit, _) => SDE("lengthKind='implicit' is not allowed with packed binary formats")
+      case (_, _, _) if ((binaryNumberKnownLengthInBits != -1) && (binaryNumberKnownLengthInBits % 4) != 0) =>
+        SDE("The given length (%s bits) must be a multiple of 4 when using packed binary formats", binaryNumberKnownLengthInBits)
+      case (BinaryNumberRep.Packed, LengthKind.Delimited, -1) => new PackedIntegerDelimitedEndOfData(this, isSigned, packedSignCodes)
+      case (BinaryNumberRep.Packed, _, -1) => new PackedIntegerRuntimeLength(this, isSigned, packedSignCodes)
+      case (BinaryNumberRep.Packed, _, _) => new PackedIntegerKnownLength(this, isSigned, packedSignCodes, binaryNumberKnownLengthInBits)
+      case (BinaryNumberRep.Ibm4690Packed, LengthKind.Delimited, -1) => new IBM4690PackedIntegerDelimitedEndOfData(this, isSigned)
+      case (BinaryNumberRep.Ibm4690Packed, _, -1) => new IBM4690PackedIntegerRuntimeLength(this, isSigned)
+      case (BinaryNumberRep.Ibm4690Packed, _, _) => new IBM4690PackedIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+      case (BinaryNumberRep.Bcd, _, _) => primType match {
+        case PrimType.Long | PrimType.Int | PrimType.Short | PrimType.Byte => SDE("%s is not an allowed type for bcd binary values", primType.name)
+        case _ => (lengthKind, binaryNumberKnownLengthInBits) match {
+          case (LengthKind.Delimited, -1) => new BCDIntegerDelimitedEndOfData(this)
+          case (_,  -1) => new BCDIntegerRuntimeLength(this)
+          case (_, _) => new BCDIntegerKnownLength(this, binaryNumberKnownLengthInBits)
+        }
+      }
     }
   }
 
@@ -738,16 +781,28 @@ trait ElementBaseGrammarMixin
       }
 
       case PrimType.Decimal => {
-        Assert.invariant(binaryIntRep == bin)
         if (binaryDecimalVirtualPoint > tunable.maxBinaryDecimalVirtualPoint)
           SDE("Property binaryDecimalVirtualPoint %s is greater than limit %s", binaryDecimalVirtualPoint, tunable.maxBinaryDecimalVirtualPoint)
         if (binaryDecimalVirtualPoint < tunable.minBinaryDecimalVirtualPoint)
           SDE("Property binaryDecimalVirtualPoint %s is less than limit %s", binaryDecimalVirtualPoint, tunable.minBinaryDecimalVirtualPoint)
         if (binaryNumberKnownLengthInBits == -1 ||
           binaryNumberKnownLengthInBits > 8) byteOrderRaw // must have or SDE
-        binaryNumberKnownLengthInBits match {
-          case -1 => new BinaryDecimalRuntimeLength(this)
-          case _ => new BinaryDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+
+        (binaryNumberRep, lengthKind, binaryNumberKnownLengthInBits) match {
+          case (BinaryNumberRep.Binary, _, -1) => new BinaryDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Binary, _, _) => new BinaryDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+          case (_, LengthKind.Implicit, _) => SDE("lengthKind='implicit' is not allowed with packed binary formats")
+          case (_, _, _) if ((binaryNumberKnownLengthInBits != -1) && (binaryNumberKnownLengthInBits % 4) != 0) =>
+            SDE("The given length (%s bits) must be a multiple of 4 when using packed binary formats", binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Packed, LengthKind.Delimited, -1) => new PackedDecimalDelimitedEndOfData(this, packedSignCodes)
+          case (BinaryNumberRep.Packed, _, -1) => new PackedDecimalRuntimeLength(this, packedSignCodes)
+          case (BinaryNumberRep.Packed, _, _) => new PackedDecimalKnownLength(this, packedSignCodes, binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Bcd, LengthKind.Delimited, -1) => new BCDDecimalDelimitedEndOfData(this)
+          case (BinaryNumberRep.Bcd, _, -1) => new BCDDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Bcd, _, _) => new BCDDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Ibm4690Packed, LengthKind.Delimited, -1) => new IBM4690PackedDecimalDelimitedEndOfData(this)
+          case (BinaryNumberRep.Ibm4690Packed, _, -1) => new IBM4690PackedDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Ibm4690Packed, _, _) => new IBM4690PackedDecimalKnownLength(this, binaryNumberKnownLengthInBits)
         }
       }
 
--- a/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesBCD.scala
+++ b/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesBCD.scala
@@ -1,38 +1,53 @@
-/* Copyright (c) 2012-2015 Tresys Technology, LLC. All rights reserved.
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- * Developed by: Tresys Technology, LLC
- *               http://www.tresys.com
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
- * Permission is hereby granted, free of charge, to any person obtaining a copy of
- * this software and associated documentation files (the "Software"), to deal with
- * the Software without restriction, including without limitation the rights to
- * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
- * of the Software, and to permit persons to whom the Software is furnished to do
- * so, subject to the following conditions:
- *
- *  1. Redistributions of source code must retain the above copyright notice,
- *     this list of conditions and the following disclaimers.
- *
- *  2. Redistributions in binary form must reproduce the above copyright
- *     notice, this list of conditions and the following disclaimers in the
- *     documentation and/or other materials provided with the distribution.
- *
- *  3. Neither the names of Tresys Technology, nor the names of its contributors
- *     may be used to endorse or promote products derived from this Software
- *     without specific prior written permission.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
- * SOFTWARE.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 package edu.illinois.ncsa.daffodil.grammar.primitives
 
-import edu.illinois.ncsa.daffodil.dsom._
+import edu.illinois.ncsa.daffodil.grammar.Terminal
 import edu.illinois.ncsa.daffodil.dsom.ElementBase
+import edu.illinois.ncsa.daffodil.processors.parsers.BCDIntegerRuntimeLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.BCDIntegerKnownLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.BCDDecimalRuntimeLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.BCDDecimalKnownLengthParser
+import edu.illinois.ncsa.daffodil.processors.unparsers.Unparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.BCDIntegerRuntimeLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.BCDIntegerKnownLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.BCDDecimalRuntimeLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.BCDDecimalKnownLengthUnparser
 
-case class BCDIntPrim(e: ElementBase) extends UnimplementedPrimitive(e, false)
+class BCDIntegerRuntimeLength(val e: ElementBase) extends Terminal(e, true) {
+  override lazy val parser = new BCDIntegerRuntimeLengthParser(e.elementRuntimeData, e.lengthEv, e.lengthUnits)
+
+  override lazy val unparser: Unparser = new BCDIntegerRuntimeLengthUnparser(e.elementRuntimeData, e.lengthEv, e.lengthUnits)
+}
+
+class BCDIntegerKnownLength(val e: ElementBase, lengthInBits: Long) extends Terminal(e, true) {
+
+  override lazy val parser = new BCDIntegerKnownLengthParser(e.elementRuntimeData, lengthInBits.toInt)
+
+  override lazy val unparser: Unparser = new BCDIntegerKnownLengthUnparser(e.elementRuntimeData, lengthInBits.toInt)
+}
+
+class BCDDecimalRuntimeLength(val e: ElementBase) extends Terminal(e, true) {
+  override lazy val parser = new BCDDecimalRuntimeLengthParser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, e.lengthEv, e.lengthUnits)
+
+  override lazy val unparser: Unparser = new BCDDecimalRuntimeLengthUnparser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, e.lengthEv, e.lengthUnits)
+
+}
+
+class BCDDecimalKnownLength(val e: ElementBase, lengthInBits: Long) extends Terminal(e, true) {
+  override lazy val parser = new BCDDecimalKnownLengthParser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, lengthInBits.toInt)
+
+  override lazy val unparser: Unparser = new BCDDecimalKnownLengthUnparser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, lengthInBits.toInt)
+}
--- a/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesIBM4690Packed.scala
+++ b/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesIBM4690Packed.scala
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package edu.illinois.ncsa.daffodil.grammar.primitives
+
+import edu.illinois.ncsa.daffodil.grammar.Terminal
+import edu.illinois.ncsa.daffodil.dsom.ElementBase
+import edu.illinois.ncsa.daffodil.processors.parsers.IBM4690PackedIntegerRuntimeLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.IBM4690PackedIntegerKnownLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.IBM4690PackedDecimalRuntimeLengthParser
+import edu.illinois.ncsa.daffodil.processors.parsers.IBM4690PackedDecimalKnownLengthParser
+import edu.illinois.ncsa.daffodil.processors.unparsers.Unparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.IBM4690PackedIntegerRuntimeLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.IBM4690PackedIntegerKnownLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.IBM4690PackedDecimalRuntimeLengthUnparser
+import edu.illinois.ncsa.daffodil.processors.unparsers.IBM4690PackedDecimalKnownLengthUnparser
+
+class IBM4690PackedIntegerRuntimeLength(val e: ElementBase, signed: Boolean) extends Terminal(e, true) {
+  override lazy val parser = new IBM4690PackedIntegerRuntimeLengthParser(e.elementRuntimeData, signed, e.lengthEv, e.lengthUnits)
+
+  override lazy val unparser: Unparser = new IBM4690PackedIntegerRuntimeLengthUnparser(e.elementRuntimeData, e.lengthEv, e.lengthUnits)
+}
+
+class IBM4690PackedIntegerKnownLength(val e: ElementBase, signed: Boolean, lengthInBits: Long) extends Terminal(e, true) {
+
+  override lazy val parser = new IBM4690PackedIntegerKnownLengthParser(e.elementRuntimeData, signed, lengthInBits.toInt)
+
+  override lazy val unparser: Unparser = new IBM4690PackedIntegerKnownLengthUnparser(e.elementRuntimeData, lengthInBits.toInt)
+}
+
+class IBM4690PackedDecimalRuntimeLength(val e: ElementBase) extends Terminal(e, true) {
+  override lazy val parser = new IBM4690PackedDecimalRuntimeLengthParser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, e.lengthEv, e.lengthUnits)
+
+  override lazy val unparser: Unparser = new IBM4690PackedDecimalRuntimeLengthUnparser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, e.lengthEv, e.lengthUnits)
+
+}
+
+class IBM4690PackedDecimalKnownLength(val e: ElementBase, lengthInBits: Long) extends Terminal(e, true) {
+  override lazy val parser = new IBM4690PackedDecimalKnownLengthParser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, lengthInBits.toInt)
+
+  override lazy val unparser: Unparser = new IBM4690PackedDecimalKnownLengthUnparser(e.elementRuntimeData, e.binaryDecimalVirtualPoint, lengthInBits.toInt)
+}
+
--- a/daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
+++ b/daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
@@ -1,0 +1,327 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package edu.illinois.ncsa.daffodil.util
+
+import edu.illinois.ncsa.daffodil.exceptions.Assert
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.BinaryNumberCheckPolicy
+
+import java.math.{ BigInteger => JBigInteger, BigDecimal => JBigDecimal }
+
+object PackedSignCodes {
+  def apply(signCodes: String, policy: BinaryNumberCheckPolicy) = {
+    val str = signCodes.replaceAll("\\s", "")
+    val chars = str.toCharArray()
+    Assert.invariant(chars.length == 4)
+
+    //We can convert hex to an integer value simply by subtracting 55
+    val positive = policy match {
+      case BinaryNumberCheckPolicy.Strict => List(chars(0) - 55)
+      case BinaryNumberCheckPolicy.Lax => List(chars(0) - 55, 0xA, 0xC, 0xE, 0xF)
+    }
+    val negative = policy match {
+      case BinaryNumberCheckPolicy.Strict => List(chars(1) - 55)
+      case BinaryNumberCheckPolicy.Lax => List(chars(1) - 55, 0xB, 0xD)
+    }
+    val unsigned = policy match {
+      case BinaryNumberCheckPolicy.Strict => List(chars(2) - 55)
+      case BinaryNumberCheckPolicy.Lax => List(chars(2) - 55, 0xF)
+    }
+    val zero_sign = policy match {
+      case BinaryNumberCheckPolicy.Strict => List(chars(3) - 55)
+      case BinaryNumberCheckPolicy.Lax => List(chars(3) - 55, 0xA, 0xC, 0xE, 0xF, 0x0)
+    }
+
+    new PackedSignCodes(positive, negative, unsigned, zero_sign)
+  }
+}
+
+case class PackedSignCodes(
+  positive: List[Int],
+  negative: List[Int],
+  unsigned: List[Int],
+  zero_sign: List[Int]
+) {}
+
+object DecimalUtils {
+
+  def packedToBigInteger(num: Array[Byte], signCodes: PackedSignCodes): JBigInteger = {
+    val numDigits = num.size * 2  // 2 digits stored per byte
+    val outputData = new Array[Char](numDigits-1)
+    var outputPos = 0
+    var offset = 0
+
+    // Parse and validate the last (sign) bit
+    val signNibble = (num(offset +  num.size - 1) & 0x0F)
+    val negative = signCodes.negative.contains(signNibble)
+    if (!negative && !signCodes.positive.contains(signNibble) &&
+      !signCodes.unsigned.contains(signNibble) && !signCodes.zero_sign.contains(signNibble)) {
+      throw new NumberFormatException("Invalid sign nibble: " + signNibble)
+    }
+
+    while (outputPos < outputData.size - 1) {
+      // Parse high nibble
+      val highNibble = (num(offset) & 0xFF) >>> 4
+      if (highNibble > 0x09) {
+        throw new NumberFormatException("Invalid high nibble: " + highNibble)
+      }
+
+      outputData(outputPos) = (highNibble | 0x0030).toChar
+      outputPos = outputPos + 1
+
+      // Parse low nibble
+      val lowNibble = (num(offset) & 0x0F)
+      if (lowNibble > 0x09) {
+        throw new NumberFormatException("Invalid low nibble: " + lowNibble)
+      }
+
+      outputData(outputPos) = (lowNibble | 0x0030).toChar
+      outputPos = outputPos + 1
+      offset = offset + 1
+    }
+
+    // Parse last digit
+    val lastNibble = (num(offset) & 0xFF) >>> 4
+    if (lastNibble > 0x09) {
+      throw new NumberFormatException("Invalid high nibble: " + lastNibble)
+    }
+
+    outputData(outputPos) = (lastNibble | 0x0030).toChar
+
+    val jbi = new JBigInteger(new String(outputData))
+    if (negative)
+      jbi.negate()
+    else
+      jbi
+
+  }
+
+  def packedToBigDecimal(num: Array[Byte], scale: Int, signCodes: PackedSignCodes): JBigDecimal = {
+    return new JBigDecimal(packedToBigInteger(num, signCodes), scale)
+  }
+
+  def packedFromBigInteger(bigInt: JBigInteger, nBits: Int, signCodes: PackedSignCodes): Array[Byte] = {
+    val negative = (bigInt.signum != 1)
+    val inChars = bigInt.abs.toString.toCharArray
+    var offset = 0
+    var inPos = 0
+    // Don't count the '-' symbol when determing the number of digits
+    val numDigits = inChars.length
+    val outArray = new Array[Byte](nBits/8)
+    val leadingZeros = if (numDigits % 2 == 0) (nBits/4 - numDigits - 1) else (nBits/4 - numDigits)
+
+    // Add leading double zeros if necessary
+    while ((offset * 2) < (leadingZeros - 1)) {
+      outArray(offset) = 0x00.asInstanceOf[Byte]
+      offset = offset + 1
+    }
+
+    // Need odd number of digits, pad with 0x0 if necessary
+    if (numDigits % 2 == 0) {
+      outArray(offset) = (((0x0 & 0x000F) << 4) + (inChars(inPos) & 0x000F)).asInstanceOf[Byte]
+      inPos = inPos + 1
+      offset = offset + 1
+    }
+
+    while (inPos < numDigits - 1) {
+      val firstNibble = (inChars(inPos) & 0x000F) << 4
+      inPos = inPos + 1
+      val secondNibble = inChars(inPos) & 0x000F
+      inPos = inPos + 1
+      outArray(offset) = (firstNibble + secondNibble).asInstanceOf[Byte]
+      offset = offset + 1
+    }
+
+    // Write out the last digit and sign code
+
+    val lastNibble = (inChars(inPos) & 0x000F) << 4
+    val signNibble = if (negative) (signCodes.negative(0) & 0x000F) else (signCodes.positive(0) & 0x000F)
+    outArray(offset) = (lastNibble + signNibble).asInstanceOf[Byte]
+
+    outArray
+  }
+
+  def bcdToBigInteger(bcdNum: Array[Byte]): JBigInteger = {
+    val numDigits = bcdNum.size * 2  // 2 digits stored per byte
+    val outputData = new Array[Char](numDigits)
+    var outputPos = 0
+    var offset = 0
+
+    while (offset < bcdNum.size) {
+      val highNibble = (bcdNum(offset) & 0xFF) >>> 4
+      if (highNibble > 0x09)
+        throw new NumberFormatException("Invalid high nibble: " + highNibble)
+
+      outputData(outputPos) = (highNibble | 0x0030).toChar
+      outputPos = outputPos + 1
+
+      val lowNibble = (bcdNum(offset) & 0x0F)
+      if (lowNibble > 0x09) {
+        throw new NumberFormatException("Invalid low nibble: " + lowNibble)
+      }
+
+      outputData(outputPos) = (lowNibble | 0x0030).toChar
+      outputPos = outputPos + 1
+      offset = offset + 1
+    }
+
+    new JBigInteger(new String(outputData))
+  }
+
+  def bcdToBigDecimal(bcdNum: Array[Byte], scale: Int): JBigDecimal = {
+    new JBigDecimal(bcdToBigInteger(bcdNum), scale)
+  }
+
+  def bcdFromBigInteger(bigInt: JBigInteger, nBits: Int): Array[Byte] = {
+    val inChars = bigInt.toString.toCharArray
+    var offset = 0
+    var inPos = 0
+    val outArray = new Array[Byte](nBits/8)
+    val leadingZeros = nBits/4 - inChars.length
+
+    // Add leading double zeros if necessary
+    while ((offset * 2) < (leadingZeros - 1)) {
+      outArray(offset) = 0x00.asInstanceOf[Byte]
+      offset = offset + 1
+    }
+
+    // Need even number of digits, pad with a single 0 if necessary
+    if (inChars.length % 2 != 0) {
+      outArray(offset) = (((0x0 & 0x000F) << 4) + (inChars(inPos) & 0x000F)).asInstanceOf[Byte]
+      offset = offset + 1
+      inPos = inPos + 1
+    }
+
+    while (inPos < inChars.length) {
+      val firstNibble = (inChars(inPos) & 0x000F) << 4
+      inPos = inPos + 1
+      val secondNibble = inChars(inPos) & 0x000F
+      inPos = inPos + 1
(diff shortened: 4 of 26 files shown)


daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesIBM4690Packed.scala
```diff
@@ -0,0 +1,74 @@
...
+ *     may be used to endorse or promote products derived from this Software
+ *     without specific prior written permission.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
+ * SOFTWARE.
```
stevedlawrence: Replace with the Apache license. All new files should be licensed under Apache v2.

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesIBM4690Packed.scala
```diff
@@ -0,0 +1,74 @@
...
+  override lazy val parser = new IBM4690PackedIntegerRuntimeLengthParser(e.elementRuntimeData, signed, e.lengthEv, e.lengthUnits)
+
+  override lazy val unparser: Unparser = new IBM4690PackedIntegerRuntimeLengthUnparser(e.elementRuntimeData, signed, e.lengthEv, e.lengthUnits)
+}
+
+class IBM4690PackedIntegerKnownLength(val e: ElementBase, signed: Boolean, lengthInBits: Long) extends Terminal(e, true) {
+
+  override lazy val parser = {
+    new IBM4690PackedIntegerKnownLengthParser(e.elementRuntimeData, signed, lengthInBits.toInt)
+  }
```
stevedlawrence: Remove for consistency.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+}
+
+abstract class BCDIntegerBaseUnparser(e: ElementRuntimeData)
+  extends BCDNumberBaseUnparser(e) {
+
+  override def putNumber(dos: DataOutputStream, value: JNumber, nBits: Int): Boolean = {
+    dos.putByteArray(bcdFromBigInteger(new JBigInteger(value.toString), nBits), nBits)
```
stevedlawrence: This converts value to a string, then converts that string to a JBigInteger, then the first thing bcdFromBigInteger does is convert that JBigInteger to a String/Array[Char]. Would it make sense to change bcdFromBigInteger to just accept a String and avoid the creation of a JBigInteger if we're just going to convert it back to a string?

Also, it might be a bit more readable to split this online into 2 or 3 lines/vals.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
```diff
@@ -0,0 +1,135 @@
...
+  override def putNumber(dos: DataOutputStream, value: JNumber, nBits: Int): Boolean = {
+    dos.putByteArray(bcdFromBigDecimal(new JBigDecimal(value.toString), nBits), nBits)
+  }
+}
+
+final class BCDIntegerMinLengthInBytesUnparser(minLengthInBytes: Int, e: ElementRuntimeData)
+  extends BCDIntegerBaseUnparser(e) {
+
+  override def getBitLength(state: ParseOrUnparseState): Int = {
+    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
```
stevedlawrence: Is this correct? Isn't the infoset value of a BCD number an actual number? Not a byte array? Perhaps this wasy copy/paste from hexBinary?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
```diff
@@ -0,0 +1,135 @@
...
+    val min = minLengthInBytes *  8
+    scala.math.max(len, min)
+  }
+}
+
+final class BCDDecimalMinLengthInBytesUnparser(minLengthInBytes: Int, e: ElementRuntimeData, binaryDecimalVirtualPoint: Int)
+  extends BCDDecimalBaseUnparser(e, binaryDecimalVirtualPoint) {
+
+  override def getBitLength(state: ParseOrUnparseState): Int = {
+    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
```
stevedlawrence: Same question here? Is this correct or an errant copy/paste? I don't see any tests for delimited hex binary, so I suspect this isn't thoroughly tested enough?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/IBM4690PackedDecimalUnparsers.scala
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+}
+
+abstract class IBM4690PackedIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
+  extends IBM4690PackedNumberBaseUnparser(e) {
+
+  override def putNumber(dos: DataOutputStream, value: JNumber, nBits: Int): Boolean = {
+    dos.putByteArray(IBM4690FromBigInteger(new JBigInteger(value.toString), nBits), nBits)
```
stevedlawrence: Same question here. Can we avoid the JBigInteger by having IBM4690FromBigInteger accept a String?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedDecimalUnparsers.scala
```diff
@@ -0,0 +1,138 @@
...
+
+class PackedDecimalRuntimeLengthUnparser(val e: ElementRuntimeData, signed: YesNo, binaryDecimalVirtualPoint: Int, val lengthEv: Evaluatable[JLong], val lUnits: LengthUnits)
+  extends PackedDecimalBaseUnparser(e, signed, binaryDecimalVirtualPoint)
+  with HasRuntimeExplicitLength {
+
+  override val runtimeDependencies = List(lengthEv)
+}
+
+abstract class PackedDecimalBaseUnparser(e: ElementRuntimeData, signed: YesNo, binaryDecimalVirtualPoint: Int)
+  extends PackedNumberBaseUnparser(e) {
```
stevedlawrence: signed and binaryDecimalVirtualPoint do not seem to be used (same with other new parsers). Are they needed, or are these properties not handled correctly? Maybe related, it doesn't look like the dfdl:binaryPackedSignCodes is implemented here. That should probably be implemented as part of this changeset as well.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/BCDParsers.scala
```diff
@@ -0,0 +1,103 @@
...
+}
+
+abstract class BCDIntegerBaseParser(e: ElementRuntimeData)
+  extends PrimParserObject(e) {
+
+  protected def getBitLength(s: ParseOrUnparseState): Int
+
+  def parse(start: PState): Unit = {
+    val nBits = getBitLength(start)
+    if (nBits == 0) return // zero length is used for outputValueCalc often.
```
stevedlawrence: We allow zero bit length for Integer, but not Decimal.  I think it is correct here, should we also allow it for Decimal?

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/BCDParsers.scala
```diff
@@ -0,0 +1,103 @@
...
+    val dis = start.dataInputStream
+
+    if (!dis.isDefinedForLength(nBits)) {
+      PE(start, "Insufficient bits in data. Needed %d bit(s) but found only %d available.", nBits, dis.remainingBits.get)
+      return
+    }
+
+    // AS400BCDDeimal only has conversion to Double and BigDecimal, need to convert to Number
```
stevedlawrence: Is this comment correct? I don't see any use of AS400BCDecimal? Maybe this is old? Is the nBits > 64 check needed?

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/DelimitedParsers.scala
```diff
@@ -183,3 +185,257 @@ class HexBinaryDelimitedParser(
...
+        this.PE(state, "%s - %s - Parse failed.", this.toString(), erd.diagnosticDebugName)
+        return
+      } else {
+        val num = packedToBigDecimal(fieldBytes, 0)
+        state.simpleElement.setDataValue(num)
+        if (result.matchedDelimiterValue.isDefined) state.saveDelimitedParseResult(parseResult)
+      }
+      return
+    }
+  }
```
stevedlawrence: Looks like this is copy/pasted a bunch. Can this be factored out into a common trait or base class?

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/IBM4690PackedDecimalParsers.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    val int: JNumber =
+      if (signed) {
+        if (nBits > 64) { IBM4690ToBigInteger(dis.getByteArray(nBits)) }
+        else { IBM4690ToBigInteger(dis.getByteArray(nBits)).longValue() }
+      } else {
+        if (nBits >= 64) { IBM4690ToBigInteger(dis.getByteArray(nBits)) }
+        else { IBM4690ToBigInteger(dis.getByteArray(nBits)).longValue() }
+      }
```
stevedlawrence: I don't think if (signed) is being used here correctly. If if/else branches are doing basically the same thing. How does signedness apply to these zoned/packed types?

daffodil-test/src/test/scala-new/edu/illinois/ncsa/daffodil/section13/packed/TestPacked.scala
```diff
@@ -54,4 +54,18 @@ class TestPacked {
...
+
+  @Test def testBCDCharset01(): Unit = { runner.runOneTest("bcdCharset01") }
+  @Test def testBCDCharset02(): Unit = { runner.runOneTest("bcdCharset02") }
+  @Test def testBCDCharset03(): Unit = { runner.runOneTest("bcdCharset03") }
+  //@Test def testBCDCharset04(): Unit = { runner.runOneTest("bcdCharset04") } // need bcd calendar
+
+  @Test def testIBM4690Charset01(): Unit = { runner.runOneTest("IBM4690Charset01") }
+  @Test def testIBM4690Charset02(): Unit = { runner.runOneTest("IBM4690Charset02") }
+  @Test def testIBM4690Charset03(): Unit = { runner.runOneTest("IBM4690Charset03") }
+  //@Test def testIBM4690Charset04(): Unit = { runner.runOneTest("IBM4690Charset04") } // need bcd calendar
```
stevedlawrence: I'm not sure if this is enough tests to ensure everything is correct. I think we need a lot more tests to make sure the logic in DecimalUtils.scala is correct. I think we also need some tests where we have various lengths (some delimited, some runtime expressions, some fixed), various endianess, and spaning multiple bytes. This is going to need a lot of tests to ensure that everything is correct. There are some tests in the ibm-contributed directory. Some of those might be able to be enabled and be a good starting point.

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -700,10 +718,19 @@ trait ElementBaseGrammarMixin
...
-      case _ => new BinaryIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+    (binaryNumberRep, lengthKind, binaryNumberKnownLengthInBits) match {
+      case (BinaryNumberRep.Binary, _, -1) => new BinaryIntegerRuntimeLength(this, isSigned)
+      case (BinaryNumberRep.Binary, _, _) => new BinaryIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+      case (BinaryNumberRep.Packed, LengthKind.Delimited, -1) => new PackedIntegerDelimitedEndOfData(this, isSigned)
+      case (BinaryNumberRep.Packed, _, -1) => new PackedIntegerRuntimeLength(this, isSigned)
+      case (BinaryNumberRep.Packed, _, _) => new PackedIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+      case (BinaryNumberRep.Bcd, LengthKind.Delimited, -1) => new BCDIntegerDelimitedEndOfData(this)
+      case (BinaryNumberRep.Bcd, _,  -1) => new BCDIntegerRuntimeLength(this)
+      case (BinaryNumberRep.Bcd, _,  _) => new BCDIntegerKnownLength(this, binaryNumberKnownLengthInBits)
```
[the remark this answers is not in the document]
jadams-tresys: No, BCD is always unsigned

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -700,10 +718,19 @@ trait ElementBaseGrammarMixin
...
+      case (BinaryNumberRep.Ibm4690Packed, LengthKind.Delimited, -1) => new IBM4690PackedIntegerDelimitedEndOfData(this, isSigned)
+      case (BinaryNumberRep.Ibm4690Packed, _, -1) => new IBM4690PackedIntegerRuntimeLength(this, isSigned)
+      case (BinaryNumberRep.Ibm4690Packed, _, _) => new IBM4690PackedIntegerKnownLength(this, isSigned, binaryNumberKnownLengthInBits)
+      case (_, _, binaryRep) => SDE("BinaryNumberRep '%s' is not supported.", binaryNumberRep.toString)
```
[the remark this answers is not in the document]
jadams-tresys: Will do

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+
+      nibble = (bcdNum(offset) & 0x0F)
+      offset = offset + 1
+      if (nibble > 0x09)
+        throw new NumberFormatException("Invalid low nibble")
+
+      outputData(outputPos) = (nibble | 0x0030).toChar
+      outputPos = outputPos + 1
+    }
+
```
stevedlawrence: Instead of building up a String, I'm wondering if it would be more efficient to instead to just build up a number. So for each new nibbled, something like
```
value = 0
while (...) {
  nibble = // get new nibble
  value = value * 10 + nibble
}
```
This would avoid allocating a char array, convert to a String, and then parse it into a JBigInteger. Should be a bit more efficient.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/BCDParsers.scala
jadams-tresys: I just need to remove this comment, it simply got copied around from when I based this code off of packed decimal

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+    var firstNibble: Int = 0
+    var secondNibble: Int = 0
+    val outArray: Array[Byte] = new Array[Byte](inChars.length / 2 + 1)
+    val leadingZeros: Int = if(inChars.length % 2 == 0) (nBits/4 - inChars.length + 1) else (nBits/4 - inChars.length)
+
+    // Add leading zeros if necessary
+    while(offset < (leadingZeros - 1)) {
+      outArray(offset) = 0
+      offset = offset + 1
+    }
```
stevedlawrence: Java initializes all array values to zero, so there's no need to set these. We can just set the offset to start at the right place.

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -301,7 +319,13 @@ trait ElementBaseGrammarMixin
...
           case Double => true;
           case n: Numeric.Kind =>
-            binaryNumberRep eq BinaryNumberRep.Binary
+            binaryNumberRep match {
+              case BinaryNumberRep.Binary => true
+              case BinaryNumberRep.Packed => false
+              case BinaryNumberRep.Bcd => false
+              case BinaryNumberRep.Ibm4690Packed => false
+              case _ => false
+            }
```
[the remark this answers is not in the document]
jadams-tresys: Will fix, this was simply left over as I was implementing the different packed formats one at a time.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+    if (inChars.length % 2 != 0) {
+      outArray(offset) = (inChars(inPos) & 0x000F).asInstanceOf[Byte]
+      offset = offset + 1
+      inPos = inPos + 1
+    }
+
+    while (inPos < inChars.length) {
+      firstNibble = (inChars(inPos) & 0x000F) << 4
+      inPos = inPos + 1
+      secondNibble = inChars(inPos) & 0x000F
```
stevedlawrence: firstNibble and secondNibble can be vals defines inside the while loop. Don't need to be vars.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+    var outputPos: Int = 0
+    var offset: Int = 0
+    var nibble: Int = 0
+    var negative: Boolean = false
+
+    while (offset < num.size) {
+      nibble = (num(offset) & 0xFF) >>> 4
+      if (nibble > 0x09) {
+        if ((nibble == 0xF) || (nibble == 0xD)) {
+          outputData(outputPos) = (0 | 0x0030).toChar
```
stevedlawrence: What's the purpose of the 0 in ``0 | 0x0030``? I think this is just setting the outputData to the char '0'? If so, can we just do outputData(outputPos) = '0'?

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+    outArray
+  }
+
+  def bcdFromBigDecimal(num: JBigDecimal, nBits: Int): Array[Byte] = {
+    bcdFromBigInteger(num.unscaledValue, nBits)
+  }
+
+  def IBM4690ToBigInteger(num: Array[Byte]): JBigInteger = {
+    val numDigits: Int = num.size * 2  // 2 digits stored per byte
+    val outputData: Array[Char] = new Array[Char](numDigits)
```
stevedlawrence: Same as above, might be more efficient to just build up the number rather than building an char array and converting to a number later.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedDecimalUnparsers.scala
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+}
+
+abstract class PackedIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
+  extends PackedNumberBaseUnparser(e) {
+
+  override def putNumber(dos: DataOutputStream, value: JNumber, nBits: Int): Boolean = {
+    dos.putByteArray(packedFromBigDecimal(new JBigDecimal(value.toString), nBits), nBits)
```
mbeckerle: Same comment SL made above. This is converting JNumber to string to JBigDecimal, and then packedFromBigDecimal does it's thing...

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedDecimalUnparsers.scala
mbeckerle: If these are needed for test cases like the formats on DFDLSchemas, or some of the IBM-provided tests, then we need to implement them. If not, we could leave these as NYI for now and implement later.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/DelimitedParsers.scala
```diff
@@ -183,3 +185,257 @@ class HexBinaryDelimitedParser(
...
+class PackedIntegerDelimitedParser(
+  erd: ElementRuntimeData,
+  textParser: TextDelimitedParserBase,
+  fieldDFAEv: FieldDFAParseEv,
+  isDelimRequired: Boolean,
+  signed: Boolean)
+  extends StringDelimitedParser(erd, TextJustificationType.None, MaybeChar.Nope, textParser, fieldDFAEv, isDelimRequired) {
+
+  /**
+   * HexBinary is just a string in iso-8859-1 encoding.
```
mbeckerle: Comment is cut/paste from HexBinary. Update or remove, and put in a comment saying this is similar to how hexBinary works.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/DelimitedParsers.scala
```diff
@@ -183,3 +185,257 @@ class HexBinaryDelimitedParser(
...
+  erd: ElementRuntimeData,
+  textParser: TextDelimitedParserBase,
+  fieldDFAEv: FieldDFAParseEv,
+  isDelimRequired: Boolean,
+  signed: YesNo,
+  binaryDecimalVirtualPoint: Int)
+  extends StringDelimitedParser(erd, TextJustificationType.None, MaybeChar.Nope, textParser, fieldDFAEv, isDelimRequired) {
+
+  /**
+   * HexBinary is just a string in iso-8859-1 encoding.
```
mbeckerle: HexBinary oriented comment. Same issue as above.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/DelimitedParsers.scala
mbeckerle: I second the motion! Lots of seemingly repeated code here. A base trait/class is needed.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/IBM4690PackedDecimalParsers.scala
```diff
@@ -0,0 +1,108 @@
...
+ *     without specific prior written permission.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
+ * SOFTWARE.
+ */
```
mbeckerle: Apache banner

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedDecimalParsers.scala
```diff
@@ -0,0 +1,108 @@
...
+    val dis = start.dataInputStream
+
+    if (!dis.isDefinedForLength(nBits)) {
+      PE(start, "Insufficient bits in data. Needed %d bit(s) but found only %d available.", nBits, dis.remainingBits.get)
+      return
+    }
+
+    val int: JNumber =
+      if (signed) {
+        if (nBits > 64) { packedToBigDecimal(dis.getByteArray(nBits), 0) }
```
mbeckerle: Same comment as SL made above about (signed).

Repeat of this code suggests refactor into a base trait/class.

daffodil-test/src/test/resources/edu/illinois/ncsa/daffodil/section13/packed/packed.tdml
```diff
@@ -110,5 +153,175 @@
     </tdml:infoset>
 
   </tdml:parserTestCase>  
-      
+
+   <tdml:parserTestCase name="packedCharset01" root="int01" model="s2"
+    description="hex works as a charset encoding">
```
mbeckerle: These description strings are misleading. We're not treating hex as a charset here. We do have such a charset, but that's not being used here, so these strings are misleading.

daffodil-test/src/test/scala-new/edu/illinois/ncsa/daffodil/section13/packed/TestPacked.scala
```diff
@@ -54,4 +54,18 @@ class TestPacked {
   // @Test def testHexCharset03(): Unit = { runner.runOneTest("hexCharset03") } // textNumberPattern V symbol
   @Test def testHexCharset04(): Unit = { runner.runOneTest("hexCharset04") }
 
+  @Test def testPackedCharset01(): Unit = { runner.runOneTest("packedCharset01") }
+  @Test def testPackedCharset02(): Unit = { runner.runOneTest("packedCharset02") }
+  @Test def testPackedCharset03(): Unit = { runner.runOneTest("packedCharset03") }
```
mbeckerle: Does this test work? (packedCharset03) 

If so then binaryDecimalVirtualPoint works, because the test's schema has that, as "2", but there was review commentary elsewhere indicating that this feature isn't implemented?

daffodil-test/src/test/scala-new/edu/illinois/ncsa/daffodil/section13/packed/TestPacked.scala
mbeckerle: I'd like to see unit tests in daffodil-lib-unitttest for the DecimalUtils.scala library. Basically because it can be tested there. These tests can then be insuring that the prims and parsers are invoking the right decimal util methods based on the properties provided, and detects the right errors, and not so much on whether the underlying library does the basics right.

stevedlawrence:
Remember to use DAFFODIL-XXXX instead of DFDL-XXXX for bug numbers in the comment.

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/primitives/PrimitivesBCD.scala
```diff
@@ -32,7 +32,42 @@
 
 package edu.illinois.ncsa.daffodil.grammar.primitives
 
-import edu.illinois.ncsa.daffodil.dsom._
```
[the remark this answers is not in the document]
jadams-tresys: Done

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+    val numDigits: Int = bcdNum.size * 2  // 2 digits stored per byte
+    val outputData: Array[Char] = new Array[Char](numDigits)
+    var outputPos: Int = 0
+    var offset: Int = 0
+    var nibble: Int = 0
+
+    while (offset < bcdNum.size) {
+      nibble = (bcdNum(offset) & 0xFF) >>> 4
+      if (nibble > 0x09)
+        throw new NumberFormatException("Invalid high nibble")
```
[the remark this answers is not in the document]
jadams-tresys: Negative unit tests have been added

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -480,7 +504,8 @@ trait ElementBaseGrammarMixin
     case LengthKind.Implicit => implicitBinaryLengthInBits
```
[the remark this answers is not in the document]
jadams-tresys: Adding the Asserts does break things.
The first Assert breaks HexBinary, because I believe it is handled as a string
The second Assert breaks the HexBinary calendar as it is a date/time not Numeric.Kind

jadams-tresys:
Just realized I totally forgot to test things once I rebased on master before pushing my branch up.  Currently broken as the getByteArray method now needs a FormatInfo object.  Will push up fix soon.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,342 @@
...
+    var firstNibble: Int = 0
+    var secondNibble: Int = 0
+    // Don't count the '-' symbol when determing the number of digits
+    val numDigits: Int = inChars.length
+    val outArray: Array[Byte] = new Array[Byte](nBits/8)
+    val leadingZeros: Int = if(numDigits % 2 == 0) (nBits/4 - numDigits - 1) else (nBits/4 - numDigits)
+
+    // Add leading double zeros if necessary
+    while((offset * 2) < (leadingZeros - 1)) {
+      outArray(offset) = 0X00.asInstanceOf[Byte]
```
stevedlawrence: Use a small 'x' for hex syntax.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,342 @@
...
+    val hexCodes: Map[String, List[Int]] = signCodesToHex(signCodes, policy)
+    var offset: Int = 0
+    var inPos: Int = 0
+    var firstNibble: Int = 0
+    var secondNibble: Int = 0
+    // Don't count the '-' symbol when determing the number of digits
+    val numDigits: Int = inChars.length
+    val outArray: Array[Byte] = new Array[Byte](nBits/8)
+    val leadingZeros: Int = if(numDigits % 2 == 0) (nBits/4 - numDigits - 1) else (nBits/4 - numDigits)
+
```
stevedlawrence: Some comment above about types, not necesary. Also, syntax standard-wise, check your parens after keywords like if(foo) and while(foo). There should be a space after the keyword.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,342 @@
...
+      Integer.parseInt(String.valueOf('A'), 16),
+      Integer.parseInt(String.valueOf('C'), 16),
+      Integer.parseInt(String.valueOf('E'), 16),
+      Integer.parseInt(String.valueOf('F'), 16),
+      Integer.parseInt(String.valueOf('0'), 16))
+
+    val positives: List[Int] = List(Integer.parseInt(String.valueOf(chars(0)), 16))
+    val negatives: List[Int] = List(Integer.parseInt(String.valueOf(chars(1)), 16))
+    val unsigneds: List[Int] = List(Integer.parseInt(String.valueOf(chars(2)), 16))
+    val zero_signs: List[Int] = List(Integer.parseInt(String.valueOf(chars(3)), 16))
```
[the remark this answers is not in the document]
stevedlawrence: Also something to think about with this stuff, if someone had dfdl:binaryPackedSignCodes="C D F 0" but dfdl:binaryNumberCheckPolicy="lax", parsing should accept all the valid values, but unparsing should output C D F and 0. I think the code you have will just pick the first thing in the list, which would be A B F A. So either ordering matters when building these lists, or unparsing should have an instance different from parsing with just the dfdl:binaryPackedSignCodes and ignoreing the valid stuff. Make sure to add tests for this, where it could parse one thing but unparse a different thing

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
stevedlawrence: Make sure to address this comment. var's should be avoided when possible. They make sense for things that need to increment but most other things can be val's. Might require some renaming (e.g. nibble becomes val highNibble val lowNibble), but is the preferred way to do it. Double check these functions for use of vars that could be vals.

daffodil-lib/src/test/scala/edu/illinois/ncsa/daffodil/util/TestDecimalUtils.scala
```diff
@@ -0,0 +1,1293 @@
+/*
```
stevedlawrence: Did you make up the data for these tests themselves or were you able to find a good resource for tests? Nice amount of tests!

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/IBM4690PackedDecimalUnparsers.scala
```diff
@@ -0,0 +1,61 @@
...
+import edu.illinois.ncsa.daffodil.processors.parsers.HasRuntimeExplicitLength
+import edu.illinois.ncsa.daffodil.processors.Evaluatable
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.LengthUnits
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.BinaryNumberCheckPolicy
+
+abstract class IBM4690PackedNumberBaseUnparser(
+  e: ElementRuntimeData)
+  extends PackedBinaryBaseUnparser(e) {
+
+  def fromBigInteger(num: String, nBits: Int, signCodes: String, policy: BinaryNumberCheckPolicy): Array[Byte] = IBM4690FromBigInteger(num, nBits)
```
stevedlawrence: Could IBM4690FromBigInteger and related functions be renamed? The capitalization makes it look like it's a class being allocated rather than a function being called. Functions should almost always started with a lowercase letter.  Maybe just make ibm lowercase? It also might help to keep the DecimalUtils object, e.g. ``... = DecimalUtils.ibm4690FromBigInteger(...)``

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
```diff
@@ -0,0 +1,70 @@
...
+import edu.illinois.ncsa.daffodil.io.FormatInfo
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.BinaryNumberCheckPolicy
+
+trait PackedBinaryConversion {
+  def fromBigInteger(num: String, nBits: Int, signCodes: String, policy: BinaryNumberCheckPolicy): Array[Byte]
+}
+
+abstract class PackedBinaryBaseUnparser(
+  e: ElementRuntimeData,
+  binaryPackedSignCodes: String = "",
```
stevedlawrence: What does a empty string binaryPackedSignCodes mean? Shouldn't there always be sign codes? Maybe the default value can be removed? Same as below, are there cases when a binaryNumberCheckPolicy isn't provided? Can the default be removed for that too?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
```diff
@@ -0,0 +1,70 @@
...
+  protected def getBitLength(s: ParseOrUnparseState): Int
+
+  def putNumber(dos: DataOutputStream, number: JNumber, nBits: Int, finfo: FormatInfo): Boolean = {
+
+    val num: String = number.isInstanceOf[JBigDecimal] match {
+      case true => number.asInstanceOf[JBigDecimal].toPlainString
+      case false => number.toString
+    }
+
+    dos.putByteArray(fromBigInteger(num, nBits, binaryPackedSignCodes, binaryNumberCheckPolicy), nBits, finfo)
```
stevedlawrence: Can you split this up into multiple lines? It's a bit hard to read and see which parameters go to which function.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedDecimalUnparsers.scala
```diff
@@ -0,0 +1,69 @@
...
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.LengthUnits
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.BinaryNumberCheckPolicy
+
+abstract class PackedNumberBaseUnparser(
+  e: ElementRuntimeData,
+  binaryPackedSignCodes: String,
+  binaryNumberCheckPolicy: BinaryNumberCheckPolicy)
+  extends PackedBinaryBaseUnparser(e, binaryPackedSignCodes, binaryNumberCheckPolicy) {
+
+  def fromBigInteger(num: String, nBits: Int, signCodes: String, policy: BinaryNumberCheckPolicy): Array[Byte] = packedFromBigInteger(num, nBits, signCodes, policy)
```
stevedlawrence: I've found it's usually good practice to include the object for these static functions, so ``DecimalUtils.packedFromBigInteger(...)``. Just makes it more clear to see what function is being called.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/BCDParsers.scala
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
+ * SOFTWARE.
+ */
+
```
stevedlawrence: Change to Apache license.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
```diff
@@ -0,0 +1,190 @@
...
+    val nBits = getBitLength(start)
+    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    val dis = start.dataInputStream
+
+    if (!dis.isDefinedForLength(nBits)) {
+      PE(start, "Insufficient bits in data. Needed %d bit(s) but found only %d available.", nBits, dis.remainingBits.get)
+      return
+    }
+
+    val bigDec = toBigDecimal(dis.getByteArray(nBits, start), binaryDecimalVirtualPoint, binaryPackedSignCodes, binaryNumberCheckPolicy)
```
stevedlawrence: Split up.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    val int: JNumber =
+      if (signed) {
+        if (nBits > 64) { toBigInteger(dis.getByteArray(nBits, start), binaryPackedSignCodes, binaryNumberCheckPolicy) }
+        else { toBigInteger(dis.getByteArray(nBits, start), binaryPackedSignCodes, binaryNumberCheckPolicy).longValue() }
+      } else {
+        if (nBits >= 64) { toBigInteger(dis.getByteArray(nBits, start), binaryPackedSignCodes, binaryNumberCheckPolicy) }
+        else { toBigInteger(dis.getByteArray(nBits, start), binaryPackedSignCodes, binaryNumberCheckPolicy).longValue() }
+      }
```
stevedlawrence: I'm not sure if these sign checks make sense. This is just getting a byte array and converting to a BigInteger. Why the signed check and conversion to long?

Also, what happens if the toBigInteger returns a negative value but ``signed`` is false? Should that be an SDE?

Also, I'm not sure the nBits check makes sense.  Does >64 bits of data mean a packed value can't be converted to a long? Why even bother converting to a long? Does something else depend on that?

daffodil-test/src/test/scala/edu/illinois/ncsa/daffodil/section13/packed/TestPacked.scala
```diff
@@ -54,4 +54,18 @@ class TestPacked {
...
+
+  @Test def testBCDCharset01(): Unit = { runner.runOneTest("bcdCharset01") }
+  @Test def testBCDCharset02(): Unit = { runner.runOneTest("bcdCharset02") }
+  @Test def testBCDCharset03(): Unit = { runner.runOneTest("bcdCharset03") }
+  //@Test def testBCDCharset04(): Unit = { runner.runOneTest("bcdCharset04") } // need bcd calendar
+
+  @Test def testIBM4690Charset01(): Unit = { runner.runOneTest("IBM4690Charset01") }
+  @Test def testIBM4690Charset02(): Unit = { runner.runOneTest("IBM4690Charset02") }
+  @Test def testIBM4690Charset03(): Unit = { runner.runOneTest("IBM4690Charset03") }
+  //@Test def testIBM4690Charset04(): Unit = { runner.runOneTest("IBM4690Charset04") } // need bcd calendar
```
stevedlawrence: I think it would be nice to have negative test of each of these. E.g. what happens when a packed value isn't value. Should be a ParseError with a helpful message and allow for backtracking. I think the Unit tests have good coverage for negative testing, but we want to make sure that when failures occur, Daffodil handles it right and backtracks appropriately.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,342 @@
...
+import edu.illinois.ncsa.daffodil.schema.annotation.props.gen.BinaryNumberCheckPolicy
+
+import java.math.{ BigInteger => JBigInteger, BigDecimal => JBigDecimal }
+
+object DecimalUtils {
+
+  def signCodesToHex(signCodes: String, policy: BinaryNumberCheckPolicy): Map[String, List[Int]] = {
+    // remove all spaces from the string
+    val str: String = signCodes.replaceAll("\\s", "")
+    val chars: Array[Char] = str.toCharArray()
```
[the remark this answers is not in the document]
jadams-tresys: I was originally going to do more validation here, but when I was testing things it seemed that any invalid string in the binaryPackedSignCodes field would simply get picked up by the general schema validation before the test would even run:
[debug] Test edu.illinois.ncsa.daffodil.section13.packed.TestPacked.testPackedCharset01 started
[error] Test edu.illinois.ncsa.daffodil.section13.packed.TestPacked.testPackedCharset01 failed: edu.illinois.ncsa.daffodil.tdml.TDMLException: org.xml.sax.SAXParseException; systemId: file:/home/jadams/Programming/tresys/daffodil/daffodil-test/target/scala-2.11/test-classes/edu/illinois/ncsa/daffodil/section13/packed/packed.tdml; lineNumber: 59; columnNumber: 114; cvc-attribute.3: The value 'C DFC' of attribute 'binaryPackedSignCodes' on element 'dfdl:format' is not valid with respect to its type, 'BinaryPackedSignCodes'.
[error] org.xml.sax.SAXParseException; systemId: file:/home/jadams/Programming/tresys/daffodil/daffodil-test/target/scala-2.11/test-classes/edu/illinois/ncsa/daffodil/section13/packed/packed.tdml; lineNumber: 59; columnNumber: 114; cvc-pattern-valid: Value 'C DFC' is not facet-valid with respect to pattern '(A|C|E|F) (B|D) (F) (A|C|E|F|0)' for type 'BinaryPackedSignCodes'., took 0.001 sec

So, I wasn't sure if more validation in the code was necessary.  If it is, let me know and I can add it in while I make some of the other improvements you have suggested.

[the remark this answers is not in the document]
stevedlawrence: Gotcha. I'd still like to see a cooker doing the parsing just to be consistent with the rest of the code. And then maybe just have an Assert.invariant where appropriate to ensure the length is 4.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,342 @@
...
+
+  }
+
+  def packedToBigDecimal(num: Array[Byte], scale: Int, signCodes: String, policy: BinaryNumberCheckPolicy): JBigDecimal = {
+    return new JBigDecimal(packedToBigInteger(num, signCodes, policy), scale)
+  }
+
+  def packedFromBigInteger(num: String, nBits: Int, signCodes:String, policy: BinaryNumberCheckPolicy): Array[Byte] = {
+    val negative: Boolean = (num.charAt(0) == '-')
+    // Discard any sign symbols or decimal points from the string (decimal point will be maintained in binaryVirtualDecimalPoint)
```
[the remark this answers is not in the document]
jadams-tresys: It might make the most sense to just remove the part about binaryVirtualDecimalPoint from the comment.  I was trying to convey that when we are creating the packed/bcd/ibm4690 from the BigInteger string that we do not need to keep track of the decimal point as the unparser already has the information on where the decimal point is supposed to go in the binaryVirtualDecimalPoint property.

I think I'll go ahead and change the function name to _FromBigNumber as this function works on both BigInteger numbers and BigDecimal numbers that have been converted to string.

daffodil-lib/src/test/scala/edu/illinois/ncsa/daffodil/util/TestDecimalUtils.scala
jadams-tresys: I made up the data for these tests.  I did some searching online and about the only thing I found were some other people's github repos that also implemented PackedDecimal and a few of them had some tests, but none of them were very comprehensive.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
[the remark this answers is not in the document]
stevedlawrence: I'm not even sure dfdl:binaryDecimalVirtualPoint works for unparsing, or maybe I just don't understand how it works. The way it works in parsing is you parse a number as an Int and then shift the decimal point binaryDeciamlVirtualPoint number of digist. On unparsing, we start with a decimal, so does that mean we should shift the decimal the opposite of binaryDecimalVirtualPoint to give an integer, then we just unparse the integer? And it should be an SDE if after the shift we don't get an integer? The spec isn't very clear about this. And it makes me think that this function should just assume/assert that there are no decimal points and that the number is always an int?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
jadams-tresys: So this may just be my ignorance of Scala, but I was trying to provide a generic base class that could be used by Packed, BCD, and IBM4690.  The binaryPackedSignCodes and binaryNumberCheckPolicy properties are only used by PackedDecimal, so I thought the best way to make this generic interface work would be to have default values which are simply ignored by BCD and IBM4690.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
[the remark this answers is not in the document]
jadams-tresys: I may not have worded things correctly.  I think we are both trying to make the same point, that when both parsing and unparsing we are essentially only dealing with integer values in these byte arrays.  In the case of parsing, we shift the decimal point around based on whats in BVDP.  When unparsing BVDP isn't used, but I just wanted to convey the fact that even though I am stripping the decimal point from the input string for the FromBigInteger functions, the decimal point information isn't being lost as it is still in BVDP in the schema.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
stevedlawrence: Gotcha. That's probably not ideal. I think you can change fromBigInteger to not accept a signCode and policy, so the definition just looks like this:
```
def fromBigInteger(num: String, nBits: Int): Array[Byte]
```
The IBM Packed would implement it like so:
```
override def fromBigInteger(num: String, nBits: Int): Array[Byte] = IBM4690FromBigInteger(num, nBits)
```
Likewise for BCD.

For PackedNumberBaseUnparser, you already pass in the sign code and policy (should be changed to the new data structure mentioned elsewhere) to its constructor, so it would just look like this:
```
override def fromBigInteger(num: String, nBits: Int): Array[Byte] = packedFromBigInteger(num, nBits, binaryPackedSignCodes, binaryNumberCheckPolicy)
```
With this, I think PackedBinaryBaseUnparser can just call fromBigInteger(num, nBits) and the implementations of fromBigInteger have any extra information they might need, so you can remove those two variables from the PackedBinaryBaseUnparser class and don't have to worry about defaults.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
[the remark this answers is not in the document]
stevedlawrence: But I think BVDP *should* be used during unparsing somewhere. For example the BinaryDecimalUnparserBase scales the BigDecimal by BVDP and then converts it to an BigInt and unparses the BigInt. I think we need to do something similar for these packed things. They should take the Decimal from the infost, scale it by BVDP, convert it to a BigInt, and then unparse that. (Note BinaryDecimalUnparseBase does not appear to do any error checking to ensure the BigDecimal converts to a BigInt after scalling, might be a bug). Does what I describe happen anywhere for the packed stuff?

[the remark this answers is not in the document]
mbeckerle: Yes, it is symmetric. BinaryDecimalVirtualPoint is for what is called Fixed-point decimal. It provides a scale factor that converts an integer to a decimal when parsing, and a decimal back to an integer when unparsing. Data should round-trip. This is used all the time in financial data where floating point is never used, base 10 rounding rules are required, etc. 

There's an obscure case, that *is* used I believe which is when the binaryDecimalVirtualPoint is negative when it scales an integer (on parsing) into a bigger integer, and cuts it down to a smaller integer on unparsing (have to think about rounding in this latter case, as the digits being lost aren't necessarily zeros). 

This is used for currency units like Japanese Yen, where often "000" are removed, so that what is stored in the data is thousands of Yen, but logically the data is single Yen. (Yen are like pennies - so thousands of yen is like $10 units, roughly in concept.)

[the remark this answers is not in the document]
jadams-tresys: Ok, I understand what you are gettting at now.  Currently what I am doing is not converting the BigDecimal to a BigInteger, instead I am just calling bigDec.toPlainString and doing fromBigInteger on that string, and then in FromBigInteger I am removing the decimal point character from the string.

What you are suggesting would probably be clearer in the code and removes the need for me to run a regex on the input string in FromBigInteger.  I hadn't really considered a malformed infoset when unparsing.  I'll start working on these changes.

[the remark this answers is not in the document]
stevedlawrence: Yeah, what you're doing is very similar. But it's possible that extra zeros could be added based on the BDVL or rounding issues as Mike mentions. That has to be dealt with somehow.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
stevedlawrence: Another thing related to this. I think we need to check to make sure that the parsed number is without the bounds of the infoset type. For example, assume the schema type is an xs:long. We parse nBits of data and successfully convert it to a number. It's possible that that number is out of bounds for an xs:long. We either need to determine the number of bits of these packed/zoned types that can fit in each infoset type and check at schema compilation type, or if that isn't possible, after converting to an BigInt/Decimal we need to validate that it is within range of the infoset type at runtime.

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -738,16 +770,25 @@ trait ElementBaseGrammarMixin
...
+          case (BinaryNumberRep.Binary, _, _) => new BinaryDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Packed, LengthKind.Delimited, -1) => new PackedDecimalDelimitedEndOfData(this)
+          case (BinaryNumberRep.Packed, _, -1) => new PackedDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Packed, _, _) => new PackedDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Bcd, LengthKind.Delimited, -1) => new BCDDecimalDelimitedEndOfData(this)
+          case (BinaryNumberRep.Bcd, _, -1) => new BCDDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Bcd, _, _) => new BCDDecimalKnownLength(this, binaryNumberKnownLengthInBits)
+          case (BinaryNumberRep.Ibm4690Packed, LengthKind.Delimited, -1) => new IBM4690PackedDecimalDelimitedEndOfData(this)
+          case (BinaryNumberRep.Ibm4690Packed, _, -1) => new IBM4690PackedDecimalRuntimeLength(this)
+          case (BinaryNumberRep.Ibm4690Packed, _, _) => new IBM4690PackedDecimalKnownLength(this, binaryNumberKnownLengthInBits)
```
stevedlawrence: Related to the grammar, Section 12.1.1 of the spec says that implicit alignment of packed decimals is always 8. See getImplicitAlignmentInBits in ElementBase.scala. That needs to be modified to check the binary rep.

stevedlawrence: Also related to the grammar, Section 12.3.3 says that implicit lengths are not allowed for packed decimals. Probably binaryNumberKnownLengthInBits needs an SDE based on the binary rep if lengthKind is implicit.

stevedlawrence: Also related to the grammar, Section 12.1.3 says packed decimal data must have alignment that is a multiple of 4 and is a SDE otherwise.

stevedlawrence: Section 12.3.7.2.3 says it is a schema definition error if the specified length is not a multiple of 4 bits. Need to check that at schema compilation time for constant value and at runtime for runtime values.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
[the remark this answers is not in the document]
stevedlawrence: Section 13.7.1 says when unparsing a binary number (packed decimal or twos-complement) and excess precision is supplied in the Infoset no rounding occurs. It is a processing error.

This to me implies that if the infoset element is scaled based on BDVP and the result is not an integer, then it is a processing error.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
stevedlawrence: Also related, the table in Section 13.7  bcd is not valid for xs:long, xs:int, xs:short, and xs:byte, so those should be an SDE. Note that xs:decimal and xs:integer are allowed for BCD, so we still need to do checks on unparse if the infoset value is negative.

daffodil-test/src/test/scala/edu/illinois/ncsa/daffodil/section13/packed/TestPacked.scala
jadams-tresys: Negative test cases for these have been added, as well as several negative test cases for the various edge cases (ie lenght and alignment must multiples of 4, type restrictions on BCD, etc.)

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/grammar/ElementBaseGrammarMixin.scala
jadams-tresys: Checks and tests for these edge cases have been added in my most recent commit 6d3a01f8c0b68428c063726a62cf2d25a397fd55

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,328 @@
...
+
+object PackedSignCodes {
+  def apply(signCodes: String, policy: BinaryNumberCheckPolicy) = {
+    val str = signCodes.replaceAll("\\s", "")
+    val chars = str.toCharArray()
+    Assert.invariant(chars.length == 4)
+
+    val positive = policy match {
+      case BinaryNumberCheckPolicy.Strict => List(Integer.parseInt(String.valueOf(chars(0)), 16))
+      case BinaryNumberCheckPolicy.Lax => List(Integer.parseInt(String.valueOf(chars(0)), 16), 0xA, 0xC, 0xE, 0xF)
```
stevedlawrence: This might be a little easier to read if do some math to conver the char to an int. Given an char in the range of 'A' to 'F', you can subtract 55 and that will convert it to the integer hex version. So this could just be
```
 BinaryNumberCheckPolicy.Strict => List(chars(0) - 55)
```
That isn't as obvious, so it might be worth adding a comment about what the - 55 is doing.

Also, when in Lax mode, the first item will be a duplicate of something else in the list. It might be a good idea to call .distinct on then new list to remove the duplicate. Might be helpful if we ever print out the valid sign codes to not have a duplicate.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,328 @@
...
+case class PackedSignCodes(
+  positive: List[Int],
+  negative: List[Int],
+  unsigned: List[Int],
+  zero_sign: List[Int]
+) {}
+
+object DecimalUtils {
+
+  def packedToBigInteger(num: Array[Byte], signCodes: String, policy: BinaryNumberCheckPolicy): JBigInteger = {
```
stevedlawrence: This function and similar *ToBigInteger functions are called at parse/unpars time, so it will also call PackedSignCodes.apply(...) at runtime. But we have all the information necessary (signCodes and binaryNumberCheckPolicy) at schema compile time. So we should be able to move the  PackedSignCodes.apply(...) call to somewhere in schema compile time (perhaps a lazy val in the grammar and then passed into the *Parser/Unparser constructors instead of signCode and policy), we can minimize the performance cost at runtime of building this objects. Then these functions could all take a PackedSignCodes object instead of the signCodes String and binaryNumberCheckPolicy.

Also, PackedSignCodes(...) without the 'new' keyword will cal the apply method of the object class, so .apply(...) isn't necessary and is usually left off in Scala.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
```diff
@@ -0,0 +1,195 @@
...
+      PE(start, "Insufficient bits in data. Needed %d bit(s) but found only %d available.", nBits, dis.remainingBits.get)
+      return
+    }
+
+    try {
+      val bigDec = toBigDecimal(dis.getByteArray(nBits, start), binaryDecimalVirtualPoint)
+      start.simpleElement.overwriteDataValue(bigDec)
+    } catch {
+      case n: NumberFormatException => PE(start, "NumberFormatException in packed data: \n%s", n.getMessage())
+    }
```
stevedlawrence: This is really minor, but it's probably more correct to move the overwriteDataValue outside of the try/catch block. That function shouldn't throw an NFE, and if it did, it's probably a bug that we do not want to hide in a ParseError. So instead, you can do something like this:
```scala
val bigDec = try {
  toBigDecimal(...)
} catch {
  case ...
}
start.simpleElement.overwriteDataValue(bigDec)
```
Also, it's generally good to avoid mentioning interal class names like NumberFormatException where possible, just because that doesn't mean anything to a user. Can the error message be changed to something a little more user friendly? Maybe just "Error in packed data:"?

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
```diff
@@ -0,0 +1,328 @@
+/*
```
stevedlawrence: Would changing this file and object PackedUtils or PackedDecimalUtils or something else be more appropriate and help clarify that it's only intended for these packed formats?

jadams-tresys: I was planning on also adding conversion functions for Zoned Decimal in here too, so I just went with DecimalUtils.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
jadams-tresys: Will do

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
jadams-tresys: I originally had that call outside the try block, but then bigDec was out of scope.  Hadn't thought of using the ```val bigDec = try {...}``` type of declaration, which still feels strange to me coming from C/C++ and their very simple declarations.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/util/DecimalUtils.scala
jadams-tresys: Will do

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
stevedlawrence: Yeah, it definitely feels weird at first, but is pretty standard syntax in Scala. It's also useful for things like if-statements, e.g.:
```scala
val foo = if (condition) {
  "bar"
} else {
  "baz"
}
```

jadams-tresys: After trying this out and talking to Steve, because ```PE(...)``` no longer throws, I would need to provide some sort of sane default for bigDec which isn't really possible.  So we decided to just leave the call to ```start.simpleElement.overwriteDataValue(bigDec)``` inside the try block.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/processors/parsers/PackedBinaryTraits.scala
```diff
@@ -101,7 +100,7 @@ abstract class PackedBinaryIntegerBaseParser(
       val int = toBigInteger(dis.getByteArray(nBits, start))
       start.simpleElement.overwriteDataValue(int)
```
stevedlawrence: It looks like the ``signed`` variable in PackedBinaryIntegerBaseParser is never used. I think you just need to throw a PE if signed is false but the ``val int`` result of toBigInteger is negative? Also, while you're in this section, it doesn't look like the ``signed`` variable needs a default value. All uses of this base class pass in the signed variable.

+1 with those two minor issues fixed.

jadams-tresys: BCD relies on using the default value.  PackedBinaryIntegerBaseParser is the base parser for all three packed formats integer parsers.  I know the naming I used here might be a bit confusing, but the spec also uses "packed" to refer to packed decimal, bcd, and ibm4690Packed formats.  I figured PackedBinary was a good compromise for something that all 3 formats use.

stevedlawrence: Ah yep, you're right. I missed that BCD, the default is fine.

PR #19, opened 2017-12-15
Fix eclipse classpath for daffodil-io to exclude test dirs.

mbeckerle:
DFDL-1872


stevedlawrence:
+1

stevedlawrence:
Remember to use DAFFODIL-XYZ instead of DFDL-XYZ in the commit message.

PR #20, opened 2017-12-18
Update eclipse classpaths to work with latest.

mbeckerle:
A large number of classpath changes in the master branch left things with eclipse classpaths needing
updates. I ran the classpath updater.

Another observation: You must sbt compile updateClassifiers, but you also have to 
build the cli or the scallop library won't be there on the classpath that is constructed. 

The diffs here are probably quite useless. All the classpath files were modified to have dependencies on the classpath for every jar in daffodil/lib_managed.


stevedlawrence:
+1

Regarding the scallop library, the issue is likely that the daffodil-cli subproject is not aggregated into the root project. So any command you run on the root (e.g. sbt compile, sbt test, sbt updateClassifiers) will also run on all subprojects ***except*** the daffodil-cli subproject. To run a command on the daffodil-cli subproject you must explicitly specify it. This prevents running all the CLI tests (which take a long time) when you run ``sbt test``. So in addition to running ``sbt updateClassifiers`` you also need to run ``sbt daffodil-cli/updateClassifiers``. This should be added to our eclipse set up documentation.

PR #22, opened 2018-01-04
Daffodil 1773 ambiguous names rebased

mbeckerle:
This fix done for nato-stanag-5516 latest schema.
    
    This schema depends on multiple elements of the same name, but in
    different choice branches,.... being accepted in path expressions, and
    working in the runtime - which works due to elimination of slots
    (DAFFODIL-1854).
    
    This fix is a trade. It fixes DAFFODIL-1773, and contributes to
    DAFFODIL-1869, but does so by removing a check, which re-breaks 
    DAFFODIL-1617 - those tests require an error message about ambiguity
    that we no longer get.
    
    Paths that are ambiguous just resolve to compiling a path that uses the
    name, and that's going to pull up the first matching child. We formerly
    issued a message indicating the path is a "query" i.e., can produce more
    than one result.
    
    So /foo/bar if bar has multiple occurrances, but isn't an array, is
    going to be accepted for compiling, and at runtime it is going to access
    the first /foo/bar element, and never the second/subsequent.
    
    DAFFODIL-1869, DAFFODIL-1773, DAFFODIL-1617


mbeckerle:
Ooops. created incorrectly.

PR #23, opened 2018-01-05
This fix done for nato-stanag-5516 latest schema.

mbeckerle:
This schema depends on multiple elements of the same name, but in
different choice branches,.... being accepted in path expressions, and
working in the runtime - which works due to elimination of slots
(DAFFODIL-1854).

This fix is a trade. It fixes DAFFODIL-1773, and contributes to
DAFFODIL-1869, but does so by removing a check, which re-breaks 
DAFFODIL-1617 - those tests require an error message about ambiguity
that we no longer get.

Paths that are ambiguous just resolve to compiling a path that uses the
name, and that's going to pull up the first matching child. We formerly
issued a message indicating the path is a "query" i.e., can produce more
than one result.

So /foo/bar if bar has multiple occurrances, but isn't an array, is
going to be accepted for compiling, and at runtime it is going to access
the first /foo/bar element, and never the second/subsequent.

DAFFODIL-1869, DAFFODIL-1773, DAFFODIL-1617

### DIFF
--- a/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/dsom/RuntimePropertyMixins.scala
@@ -108,7 +108,7 @@ trait TermRuntimeValuedPropertiesMixin
         val exp = ConstantExpression(qn, PrimType.HexBinary, "iso-8859-1")
         exp
       }
-      case _ => ExpressionCompilers.String.compile(qn, NodeInfo.NonEmptyString, encodingRaw)
+      case _ => ExpressionCompilers.String.compile(qn, NodeInfo.NonEmptyString, encodingRaw, decl)
     }
   }.value
 
@@ -154,7 +154,7 @@ trait TermRuntimeValuedPropertiesMixin
   lazy val outputNewLineEv = {
     val outputNewLineExpr = {
       val qn = this.qNameForProperty("outputNewLine")
-      ExpressionCompilers.String.compile(qn, NodeInfo.NonEmptyString, outputNewLineRaw)
+      ExpressionCompilers.String.compile(qn, NodeInfo.NonEmptyString, outputNewLineRaw, decl)
     }
     val ev = new OutputNewLineEv(outputNewLineExpr, termRuntimeData)
     ev.compile()
@@ -203,7 +203,7 @@ trait DelimitedRuntimeValuedPropertiesMixin
     val qn = this.qNameForProperty("initiator")
     val typeIfStaticallyKnown = NodeInfo.String
     val typeIfRuntimeKnown = NodeInfo.NonEmptyString
-    ExpressionCompilers.String.compile(qn, typeIfStaticallyKnown, typeIfRuntimeKnown, initiatorRaw)
+    ExpressionCompilers.String.compile(qn, typeIfStaticallyKnown, typeIfRuntimeKnown, initiatorRaw, decl)
   }
 
   lazy val initiatorParseEv = {
@@ -224,7 +224,7 @@ trait DelimitedRuntimeValuedPropertiesMixin
     val typeIfStaticallyKnown = NodeInfo.String
     val typeIfRuntimeKnown = NodeInfo.NonEmptyString
     val raw = terminatorRaw
-    ExpressionCompilers.String.compile(qn, typeIfStaticallyKnown, typeIfRuntimeKnown, raw)
+    ExpressionCompilers.String.compile(qn, typeIfStaticallyKnown, typeIfRuntimeKnown, raw, decl)
   }.value
 
   final def terminatorLoc = (this.diagnosticDebugName, this.path)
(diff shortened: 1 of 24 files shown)


daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -377,10 +377,59 @@ class DPathElementCompileInfo(
...
+        // For now we just check if they are siblings, i.e., have the
+        // same model group as parent.
+        //
+        //        val ambiguousSiblings = ambiguousModelGroupSiblings(retryMatchesERD)
+        //        if (ambiguousSiblings.isEmpty)
+        // BUG
+        // This results in multiple siblings with same name, path that is ambiguous
+        // just gets first one.
+        //
+        retryMatchesERD(0)
```
stevedlawrence: -1, I'm not sure I like this change as-is. Multiple times I've had to spend time tracking down unexpected behavior where the issue was just that there were ambiguous paths and Daffodil just always picked the first one. This patch essential reverts back to that behavior, which was a pain.

In fact, I think if a query-style expression is used and actually references multiple elements, we'll get an invariant failure assertion at parse time in ``def findChild`` with ``fastSeq.length == 1`` or ``found.length <= 1``.

How about instead we change ``queryMatchError`` to throw an SDW instead of an SDE so that users will be made aware that a query-style expression exists, which we don't yet support. This will still allow it though, so in ``def findChild``, we change the invariant assertions to an SDE, causing a Runtime SDE when a query-style expression returns more than one element? This way 1) we get an warning when a query style expression exists 2) query style expressions can still be used, but only if they ever refer to a single element at runtime and 3) we get a Runtime SDE if a query style expression ever returns more than one element. So we warn on potential ambiguity, and throw SDE when that ambiguity is realized in practice.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/infoset/InfosetImpl.scala
```diff
@@ -1334,10 +1334,29 @@ sealed class DIComplex(override val erd: ElementRuntimeData, val tunable: Daffod
...
   final def getChild(info: DPathElementCompileInfo): InfosetElement = {
-    if (nameToChildNodeLookup.containsKey(info.namedQName))
-      nameToChildNodeLookup.get(info.namedQName)(0).asInstanceOf[InfosetElement]
-    else
+    if (nameToChildNodeLookup.containsKey(info.namedQName)) {
+      val nodes = nameToChildNodeLookup.get(info.namedQName)
+      noQuerySupportCheck(nodes, info)
+      val node = nodes(0).asInstanceOf[InfosetElement]
+      node
+    } else
```
stevedlawrence: You probably want to rebase onto the latest master. That refactors some of this code and adds a new findChild method. That method has some asserts that require there only be a single element. Those asserts probably want to be replaced with your noQuerySupportCheck function.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/infoset/InfosetImpl.scala
```diff
@@ -1451,7 +1468,7 @@ sealed class DIComplex(override val erd: ElementRuntimeData, val tunable: Daffod
           nameToChildNodeLookup.remove(childToRemove.namedQName)
         else
           // not the last one, just drop the end
-          fastSeq.dropRight(1)
+          fastSeq.remove(fastSeq.length - 1)
```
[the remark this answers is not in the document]
stevedlawrence: I think that PR was merged recently.

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/infoset/InfosetImpl.scala
mbeckerle: Please see this additional pull request. I didn't want to rebase or all the comment log would be lost, so I rebased on a different branch, then isolated just the changes needed to rebase from this commit. 

This is the changes needed to rebase this onto master:
https://github.com/mbeckerle/incubator-daffodil/pull/4

daffodil-core/src/main/scala/edu/illinois/ncsa/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -467,7 +467,7 @@ trait ElementRuntimeValuedPropertiesMixin
   private lazy val occursCountExpr = LV('occursCount) {
     val qn = this.qNameForProperty("occursCount")
     val isEvaluatedAbove = true
-    ExpressionCompilers.JLong.compile(qn, NodeInfo.Long, occursCountRaw, isEvaluatedAbove)
+    ExpressionCompilers.JLong.compile(qn, NodeInfo.Long, occursCountRaw, decl, isEvaluatedAbove)
```
jadams-tresys: Shouldn't the decl parameter be after isEvaluatedAbove?

mbeckerle: Yeah, that little inconsistency is because one of the overloads of compile() has an optional final arg. I should probably eliminate that, or get rid of all this overloading and give them names like compilePropertyExpression,.....

mbeckerle: Fixed this in the latest commit. Refactored this confusing class, renamed methods so they're distinct by purpose of the method.

mbeckerle:
See PR 26, which is changes since these review comments. Due to rebase, those changes couldn't be on this same branch or this comment history would be overwritten and lost. Hence, the new PR.

PR #24, opened 2018-01-05
Make TDML Runner not strip all attributes

mbeckerle:
XMLUtils for comparison of test results modified to ignore dafint, and
xsi:type, but preserve all other attributes.

Many test files erroneously had e.g., 
<ex:foo type="xs:int">3</ex:foo>
which is wrong because the attribute must be xsi:type, not just type.


daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/xml/XMLUtils.scala
```diff
@@ -671,13 +676,27 @@ object XMLUtils {
...
+            // support for them, we would need to not remove them.
+            // TODO: actually check xsi:type attributes are correct - but this
+            // requires schema-aware comparison.
+            case xsiTypeAttr @ PrefixedAttribute("xsi", "type", _, _) =>
+              false // drop xsi:type attributes for now. Even if prefix xsi is not defined.
+            // This just avoids having to edit many tests to add in the xmlns:xsi=....
+            // namespace declaration.
+            case attr =>
+              true // keep all other attributes
+            //              {
```
mbeckerle: Remove commented code.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/xml/XMLUtils.scala
```diff
@@ -671,13 +676,27 @@ object XMLUtils {
...
               true
             }
-            case attr => {
-              if (ns.length > 0) {
-                !ns.contains(NS(attr.getNamespace(e)))
-              } else {
-                false
-              }
+            case dafIntAttr @ PrefixedAttribute(pre, _, _, _) if (pre ne null) && (dafIntAttr.getNamespace(e) == XMLUtils.DAFFODIL_INTERNAL_NAMESPACE.toString) => {
+              Assert.invariant(pre != "")
```
stevedlawrence: Is the only situation where pre could be an empty string is if default xmlsns was the daffodil internal namespace? Seems some like something really weird and probably dumb to do, but technically it's legal, I think? So should this assertion be removed?

daffodil-tdml/src/main/scala/edu/illinois/ncsa/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1770,33 +1769,33 @@ case class DFDLInfoset(di: Node, parent: Infoset) {
...
+  //    //
+  //    //    val ptc = parent.parent
+  //    //    val schemaNode = ptc.findModel(ptc.model)
+  //    //
+  //    // This is causing trouble, with the stripped attributes, etc.
+  //    // TODO: Fix so we can validate these expected results against
+  //    // the DFDL schema used as a XSD for the expected infoset XML.
+  //    //
+  //    expectedNoAttrs
+  //  }
```
stevedlawrence: Remove comment.

eclipse-projects/tdml/.classpath
```diff
@@ -1,96 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-        <!-- This file is updated by the UpdateEclipseClasspath app. -->
-        <classpathentry kind="src" path="src/main/scala"/><classpathentry kind="src" path="src/main/scala-debug"/><classpathentry kind="con" path="org.scala-ide.sdt.launching.SCALA_CONTAINER"/><classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/><classpathentry combineaccessrules="false" kind="src" path="/daffodil-core"/><classpathentry combineaccessrules="false" kind="src" path="/daffodil-lib"/><classpathentry combineaccessrules="false" kind="src" path="/daffodil-runtime1"/><classpathentry combineaccessrules="false" kind="src" path="/daffodil-io"/><classpathentry combineaccessrules="false" kind="src" path="/daffodil-macro-lib"/><classpathentry kind="output" path="target/eclipse/classes"/><!--
```
stevedlawrence: Was this just missed in the other patchset? I feel like this keep changing all the time? Maybe we need a different solution? Maybe they shouldn't be tracked by git and we just clearly document how to create these files if you want to use eclipse?

mbeckerle: No this is just perfecting these things over time. E.g., as we get rid of these scala-new dirs and such all the eclipse projects have to be updated. Also they were setup wrong at one point and we only discovered this by finding out that searching was generating many hits that are really all the same file but from different modules, etc. In theory this will settle down again and then these won't change again for a while.

daffodil-lib/src/main/scala/edu/illinois/ncsa/daffodil/xml/XMLUtils.scala
mbeckerle: Well, dafint is ours. Not the users choice how to use it. I don't think we want to allow use of dafint as a default namespace. Certainly I'd consider it  a programming error in daffodil if dafint ever was a default namespace hence, this is an invariant.

jadams-tresys:
+1

PR #25, opened 2018-01-09
Commits to relicense Daffodil to Apache v2

stevedlawrence:
Now that all necessary SGA's are in place (Tresys, NCSA, IBM), we can relicense Daffodil to Apache v2 and change the package name to org.apache.daffodil. This work is split up into four commits to make it easier to review. Each commit contains a detailed description, but the high level changes are:

1. Move directory structure to org/apache/daffodil. Everything here is a 100% rename so shouldn't be too difficult to review
1. Change package names and xml namespaces to org.apache. Mostly changes package/imports, but does include some code for xmlns backwards compatability
1. Remove unused files and cruft--less to worry about with relicensing
1. Relicense to Apache v2. Adds NOTICE/LICENSE files. Changes license of all files (except where we aren't supposed to, e.g. Passera/scala files)
1. Run dos2unix on all files.


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1303,13 +1303,22 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
-      case (RefQName(_, "error", DAF), args) => {
+      case (RefQName(_, "trace", DAF_APACHE), args) =>
+        DAFTraceExpr(functionQNameString, functionQName, args)
+
+      case (RefQName(_, "error", DAF_NCSA), args) => {
+        SDW("Expression daf:error is deprecated. Use fn:error instead")
+        DAFErrorExpr(functionQNameString, functionQName, args)
+      }
+
+      case (RefQName(_, "error", DAF_APACHE), args) => {
```
stevedlawrence: Actual code change other than package rename.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/IIBase.scala
```diff
@@ -180,7 +180,17 @@ abstract class IIBase( final override val xml: Node, xsdArg: XMLSchemaDocument,
...
+    val prop = getAttributeOption("schemaLocation")
+    prop.map { text =>
+      if (text.contains("edu/illinois/ncsa/daffodil")) {
+        SDW("schemaLocation property uses deprecated edu/illinois/ncsa/daffodil path instead of org/apache/daffodil. Converting to new path.")
+        text.replace("edu/illinois/ncsa/daffodil", "org/apache/daffodil")
+      } else {
+        text
+      }
+    }
+  }
```
stevedlawrence: Actual code change other than package rename. Provides backwards compatibility for schemas using old imports.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -360,20 +360,24 @@ object XMLUtils {
...
+  val EXT_PREFIX_NCSA = "daf"
+  val EXT_NS_NCSA = NS(DAFFODIL_EXTENSION_NAMESPACE_NCSA.uri)
+
+  private val DAFFODIL_EXTENSIONS_NAMESPACE_ROOT_APACHE = "urn:ogf:dfdl:2013:imp:daffodil.apache.org:2017"
+  private val DAFFODIL_EXTENSION_NAMESPACE_APACHE = NS(DAFFODIL_EXTENSIONS_NAMESPACE_ROOT_APACHE + ":ext")
+  val EXT_PREFIX_APACHE = "daf"
+  val EXT_NS_APACHE = NS(DAFFODIL_EXTENSION_NAMESPACE_APACHE.uri)
+
+  private val DAFFODIL_INTERNAL_NAMESPACE = NS(DAFFODIL_EXTENSIONS_NAMESPACE_ROOT_APACHE + ":int")
+  val INT_PREFIX= "dafint"
```
stevedlawrence: Actual code change other than package rename. Provides backwards compatability with old Daffodil xml namespaces.

NOTICE
```diff
@@ -0,0 +1,20 @@
+Apache Daffodil
```
stevedlawrence: New NOTICE file for the source.

LICENSE
```diff
@@ -1,69 +1,370 @@
-Copyright (c) 2010 NCSA.  All rights reserved.
```
stevedlawrence: New LICENSE file for the source.

daffodil-cli/NOTICE
```diff
@@ -0,0 +1,88 @@
+Apache Daffodil
```
stevedlawrence: New NOTICE file for CLI releases.

daffodil-cli/LICENSE
```diff
@@ -0,0 +1,541 @@
+
```
stevedlawrence: New LICENSE file for CLI releases.

daffodil-cli/README_How_to_Debug.txt
```diff
@@ -1,15 +0,0 @@
-Because the CLI tests spawn a sub-shell and exec the daffodil program
```
stevedlawrence: This content was consolidated into daffodil-cli/README.md since Apache Rat doesn't realize this filename as a readme and complains about missing apache license. However, the content probably doesn't actually belong in the Daffodi CLI Readme, since that is more about how to run the CLI released in zip/tar/rpm. This probably really belongs in the root README cli since that talks about how to run cli tests and it more related to source that a release.

daffodil-core/src/main/resources/daffodil.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
```
mbeckerle: Review tool won't let me add a comment on the prior file (index.html) because the diff is too large.

We should publish the HTML version of the spec on the apache daffodil web site for users. Right now the fact that this is published on our NCSA deliverables is the only place it is accessible to my knowledge.

stevedlawrence: It's currently published on the apache website at http://daffodil.apache.org/docs/dfdl/ linked under the docs menu item.

NOTICE
```diff
@@ -0,0 +1,20 @@
...
+
+  This product includes material copied or derived from W3C, Copyright 2001 (c)
+  World Wide Web Consortium (Massachusetts Institute of Technology, European
+  Research Consortium for Informatics and Mathematics, Keio University). Specific
+  files include:
+    - XMLSchema.dtd (https://www.w3.org/2001/XMLSchema.dtd)
+    - XMLSchema.xsd (https://www.w3.org/2001/XMLSchema.xsd)
+    - xml.xsd (https://www.w3.org/2001/xml.xsd)
+    - datatypes.dtd (https://www.w3.org/2001/datatypes.dtd)
+    - XMLSchema_for_DFDL.xsd (https://www.w3.org/2001/XMLSchema.dtd)
```
stevedlawrence: A discussion recently popped up on the Apache incubator list about what belongs in the NOTICE file:

https://lists.apache.org/thread.html/911ba67907cbca6925851e448ac8a7e73a921cb59efb8ca028e15845@%3Clegal-discuss.apache.org%3E

This specifically mentions the W3C license. As I understand it, the conclusion is that we should be careful what goes in a NOTICE file since it is sortof viral. And since the W3C License does not specifically call out how attribute should be made (like the Apache v2 license), it is not required to be put in the NOTICE file. The same content is already in the LICENSE file which is sufficient for attribution, so this should be removed. Same with the content in daffodil-cli/NOTICE.

PR #26, opened 2018-01-16
Daffodil 1773 - ambiguous names rebase changes - extension of PR 23

mbeckerle:
Replaces PR 23 - contains additional changes and content since that review; however, due to rebase, these changes can't be added to that PR or prior comment history would be lost.

### DIFF
--- a/daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/dsom/CompiledExpression1.scala
+++ b/daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/dsom/CompiledExpression1.scala
@@ -339,6 +339,11 @@ class DPathElementCompileInfo(
    * value is determined during DPath compilation, which requires that the
    * DPathElementCompileInfo already exists. So this must be a mutable value
    * that can be flipped during schema compilation.
+   *
+   * Note that in the case of multiple child element decls with the same name,
+   * we must make sure ALL of them get this var set.
+   *
+   * This is done on the Seq returned when findNameMatches is called.
    */
   var isReferencedByExpressions = false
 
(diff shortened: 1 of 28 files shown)


daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -383,13 +392,61 @@ class DPathElementCompileInfo(
         matchesERD
       }
 
+    //
+    // We must indicate for all children having this path step as their name
```
stevedlawrence: I'm kindof iffy on this. Technically, I'm not sure findNamedMatch is the right place for this, since there could be code that's trying to find a named match but for whatever reason they aren't used in an expression. I don't think that's the case, but the name doesn't really imply there will be side effects.

Perhaps just add a TODO to remove this when query-style expressions are supported? I think when query style expression are supported this function will return a Seq of matches, and then the caller can mark isReferencedByExpression as appropriate?

daffodil-test/src/test/scala-debug/edu/illinois/ncsa/daffodil/section23/dfdl_expressions/TestDFDLExpressionsDebug.scala
```diff
@@ -177,4 +180,8 @@ class TestDFDLExpressionsDebug {
   @Test def test_nonNeg_constructor_02a() { runner2.runOneTest("nonNeg_constructor_02a") }
 
   @Test def test_element_long_form_whitespace() { runner.runOneTest("element_long_form_whitespace") }
+
+  // DFDL-1617 - should detect errors due to query-style expressions
+  @Test def test_query_style_01 { runner5.runOneTest("query_style_01") }
+  @Test def test_query_style_02 { runner5.runOneTest("query_style_02") }
```
stevedlawrence: Do these test work now and can be moved out of scala-debug?

stevedlawrence:
+1

daffodil-runtime1/src/main/scala/edu/illinois/ncsa/daffodil/dsom/CompiledExpression1.scala
mbeckerle: Just fixing this to name it findNamedMatches and put the setting of the indicator var in a separate private method called at each place findNamedMatches is called.

daffodil-test/src/test/scala-debug/edu/illinois/ncsa/daffodil/section23/dfdl_expressions/TestDFDLExpressionsDebug.scala
mbeckerle: Yes, actually these are moving back. The prior work on this branch broke them hence moved to scala-debug. But they work now, so are moving back.

jadams-tresys:
+1

PR #29, opened 2018-01-24
Fixed bug in calculating bit length for delimited unparsing

jadams-tresys:
- getBitLength now correctly calculates the length based on the number of digits
 that will be present in the unparsed data
 - Moved existing tests out of scala-debug that now work

DAFFODIL-1739

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
@@ -48,16 +48,11 @@ class BCDIntegerRuntimeLengthUnparser(
   override lazy val runtimeDependencies = List(lengthEv)
 }
 
-final class BCDIntegerMinLengthInBytesUnparser(
-  minLengthInBytes: Int,
+final class BCDIntegerDelimitedUnparser(
   e: ElementRuntimeData)
   extends BCDIntegerBaseUnparser(e) {
 
-  override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
-    val min = minLengthInBytes * 8
-    scala.math.max(len, min)
-  }
+  override def getBitLength(state: ParseOrUnparseState): Int = { 0 }
 }
 
 abstract class BCDDecimalBaseUnparser(
@@ -87,15 +82,10 @@ class BCDDecimalRuntimeLengthUnparser(
   override lazy val runtimeDependencies = List(lengthEv)
 }
 
-final class BCDDecimalMinLengthInBytesUnparser(
-  minLengthInBytes: Int,
+final class BCDDecimalDelimitedUnparser(
   e: ElementRuntimeData,
   binaryDecimalVirtualPoint: Int)
   extends BCDDecimalBaseUnparser(e, binaryDecimalVirtualPoint) {
 
-  override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
-    val min = minLengthInBytes * 8
-    scala.math.max(len, min)
-  }
+  override def getBitLength(state: ParseOrUnparseState): Int = { 0 }
 }
--- a/daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/IBM4690PackedDecimalUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/IBM4690PackedDecimalUnparsers.scala
@@ -48,16 +48,11 @@ class IBM4690PackedIntegerRuntimeLengthUnparser(
   override lazy val runtimeDependencies = List(lengthEv)
 }
 
-final class IBM4690PackedIntegerMinLengthInBytesUnparser(
-  minLengthInBytes: Int,
+final class IBM4690PackedIntegerDelimitedUnparser(
   e: ElementRuntimeData)
   extends IBM4690PackedIntegerBaseUnparser(e) {
 
-  override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
-    val min = minLengthInBytes *  8
-    scala.math.max(len, min)
-  }
+  override def getBitLength(state: ParseOrUnparseState): Int = { 0 }
 }
 
 abstract class IBM4690PackedDecimalBaseUnparser(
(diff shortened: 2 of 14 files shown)


daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
```diff
@@ -54,7 +54,14 @@ final class BCDIntegerMinLengthInBytesUnparser(
   extends BCDIntegerBaseUnparser(e) {
 
   override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
+    val bitLen = state.currentNode.get.asSimple.dataValue.asInstanceOf[JBigInteger].bitLength
```
stevedlawrence: BigInteger.bitLength returns the length in bits in two's complement. But don't we want the length in bits in BCD? Do we need to unparse the BigInteger to a byte array first, and if that isn't at least minLength then add some padding bits? Or perhaps we pass the minLength into the bcdFromBigInteger() to accept a minLength, and the padding logic could be done there if it's easier?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
```diff
@@ -94,7 +101,7 @@ final class BCDDecimalMinLengthInBytesUnparser(
   extends BCDDecimalBaseUnparser(e, binaryDecimalVirtualPoint) {
 
   override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
+    val len = (state.currentNode.get.asSimple.dataValue.asInstanceOf[JBigDecimal].precision * 8) / 2
```
stevedlawrence: Do you want the same thing mentioned above here?

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AU.dfdl.xsd
```diff
@@ -11,7 +13,8 @@
         separatorSuppressionPolicy="anyEmpty" separatorPosition="infix"
         documentFinalTerminatorCanBeMissing='yes' byteOrder="bigEndian"
         binaryNumberRep='binary' representation="text" occursCountKind='parsed'
-        lengthUnits="bytes" textTrimKind='none' escapeSchemeRef='' />
+	lengthUnits="bytes" textTrimKind='none' escapeSchemeRef=''
+	ref="tns:daffodilTest1" />
```
stevedlawrence: Indentation isn't consistent here.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AU.tdml
```diff
@@ -7,23 +7,23 @@
...
     </document>
     <infoset>
       <dfdlInfoset>
         <list xmlns="http://www.example.org/example1" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-          <bcd type="xsd:int">0017</bcd>
-          <bcd type="xsd:int">0023</bcd>
-          <bcd type="xsd:int">0027</bcd>
+          <bcd type="xsd:int">17</bcd>
+          <bcd type="xsd:int">23</bcd>
+          <bcd type="xsd:int">27</bcd>
```
stevedlawrence: Should the ``type`` here be xsd:unsignedInt to match your changes above. Daffodil currently ignores this attribute, but we might check it one day.

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/BCDUnparsers.scala
stevedlawrence: It might be worth adding some tdml:unparserTestCase cases to ensure that minLength is behaving correctly. This feels to me like maybe we're adding more pad characters than is necessary and things just work because extra pad chars is okay?

daffodil-runtime1-unparser/src/main/scala/edu/illinois/ncsa/daffodil/processors/unparsers/IBM4690PackedDecimalUnparsers.scala
```diff
@@ -54,7 +54,21 @@ final class IBM4690PackedIntegerMinLengthInBytesUnparser(
   extends IBM4690PackedIntegerBaseUnparser(e) {
 
   override def getBitLength(state: ParseOrUnparseState): Int = {
-    val len = state.currentNode.get.asSimple.dataValue.asInstanceOf[Array[Byte]].length * 8
+    val num = state.currentNode.get.asSimple.dataValue.asInstanceOf[JBigInteger]
+    // If negative account for sign nibble
```
mbeckerle: If I understand this correctly, IBM4690 packed numbers only have sign nibbles when they are negative. If positive there is no sign nibble. Is that right? If so, please add a comment to the code stating this, as it is quite unintuitive, and otherwise someone might "fix" this code to be symmetric thinking it is broken, etc.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AU.tdml
mbeckerle: I believe we should remove these type="xsd:int". We're not generating them, not checking them, they're just misleading at best currently.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AU.dfdl.xsd
jadams-tresys: Spaces vs tabs rears its ugly head again

jadams-tresys:
Is it not an issue that one of the unparser tests fails?  The delimited packed sequence unparser tests fail because it unparses "3" as "3C" instead of "003C".  Is this an issue?  If it isn't, should I just remove the unparser tests?

stevedlawrence:
If I understand BCD correctly, 3<sub>10</sub> = 0x3C<sub>BCD</sub>. Since minLength defaults to zero if not supplied, if will unparse to the minimum number of bytes needed, which is just 0x3C. If you had minLength="2" on the other hand, then it should pad 0x3C to 2 bytes, in which case I would expect 0x003C.

So I think the test just needs to be updated to either specify minLength or to not expect the zeros (or maybe add some extra tests so both cases are covered). Same thing applies for the DelimitedPackedDecSeq test which uparses 0.03.

jadams-tresys:
I'm not sure minLength is the solution.  First off, I don't think you can specify xs:minLength on an xs:int or xs:decimal.  Also, what is really weird is that BCD and IBM4690 don't seem to have an issue and somewhere along the way the 00 (or FF in IBM4690) is getting added.  I'm not sure where they are coming from as the minLengthInBits parameter that is getting passed into the *fromBigInteger functions is always 0 and the functions output a byte array that is just [3].  So I'm not sure how this works for BCD and IBM4690 and I'm even more confused why the same thing isn't happening for packed.

jadams-tresys:
I found the section I think you were talking about with not padding anything:
section 12.3.2    dfdl:lengthKind 'delimited'
 - "When unparsing a simple element with binary representation, then for hexBinary the length is the number of bytes in the infoset value padded to the XSD minLength facet value using dfdl:fillByte, and for the other types the length is the minimum number of bytes to represent the value and any sign."

So it sounds like instead of the bug is actually in the BCD and IBM4690 unparsers that were padding the data, not Packed.  Also, your understanding of BCD is different than mine.  I thought that BCD doesn't have sign nibbles, ie. the decimal number 3 would just be 0x03, not 0x3C.  Packed uses sign nibbles and would be represented 0x3C.

jadams-tresys:
Ok, so fatherhood has not been kind to my powers of observation.  The reason that the BCD and IBM4690 unparser tests were passing when I expected them to fail was due to the fact that I labeled them with tdml:parserTestCase instead of tdml:unparserTestCase.

I have fixed these tests and updated the delimited parsers for all 3 formats to no longer take a minimum bit length parameter.

PR #31, opened 2018-01-29
Prepare for 2.1.0 release

stevedlawrence:
- Remove -SNAPSHOT from version
- Add KEYS file containing Daffodil developer signing keys, initialized
  with key from Steve Lawrence
- Remove "incubating" from the generated RPM name, that is only
  necessary on the source distribution, not on convenience binaries

DAFFODIL-1881

### DIFF
--- a/KEYS
+++ b/KEYS
@@ -1,0 +1,89 @@
+This file contains the GPG keys of Apache Daffodil developers.
+
+Users:    gpg --import KEYS
+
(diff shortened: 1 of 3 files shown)


stevedlawrence:
This will be the last commit before 2.1.0-rc1 is created. Will hold off merging until all other pull requests are merged.

PR #38, opened 2018-02-09
Move built-in-formats.xsd to test and the backwards compatibility added for it

stevedlawrence:
A handful of schemas use built-in-formats.xsd as a method to get a
default set of DFDL properties. This used to be in the
edu/illinois/ncsa/daffodil namespace, but was moved to
org/apache/daffodil with the move to Apache incubation. This move would
have broken schemas, so code was added to provide backwards
compatability. This code essentially replaced instances of
edu/illinois/ncsa with org/apache/daffodil, but that could break things
if actual schema files were in edu/illinois/ncsa, which is legit and
does happen.

However, built-in-formats.xsd is intended to only be used for testing
and not actual schemas. The GeneralPurposeFormat.xsd from IBM is
provided for this purpose and should be used instead. So move
built-in-format.xsd to src/test/resources and remove the broken
backwards compatibility hack. Also update sbt subproject dependencies so
test projects depend on the core "test" so that built-in-formats.xsd is
on the test classpath. Also update CLI test infrastructure to modify
DAFFODIL_CLASSPATH to include the daffodil-lib test resource directory
for the same reason.

DAFFODIL-1897


daffodil-cli/src/test/resources/clitests.tdml
```diff
@@ -1,35 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
stevedlawrence: There's a single test that uses the TDMLRunner to execute this test. No test using the CLI actually uses this file, so this really doesn't belong in the daffodil-cli project and doesn't really add anything that isn't already tested, so just delete it.

build.sbt
```diff
@@ -75,11 +75,11 @@ lazy val test             = Project("daffodil-test", file("daffodil-test")).conf
                               .settings(commonSettings, nopublish)
 
 lazy val testIBM1         = Project("daffodil-test-ibm1", file("daffodil-test-ibm1")).configs(TestDebug)
-                              .dependsOn(tdml)
+                              .dependsOn(tdml, core % "test->test")
                               .settings(commonSettings, nopublish)
 
 lazy val tutorials        = Project("daffodil-tutorials", file("tutorials")).configs(TestDebug)
-                              .dependsOn(tdml)
+                              .dependsOn(tdml, core % "test->test")
```
stevedlawrence: These new dependencies might require modifications to the Eclipse project/classpath files? Mike, could use give this a quick test and let me know if changes are needed? Maybe even add a pull request to this branch on stevedlarwence/incubator-daffodil if so? I don't have an eclipse environment set up so that would take a while to test for me.

daffodil-cli/src/test/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
```diff
@@ -50,7 +48,8 @@ class TestCLISaveParser {
     val shell = Util.start("")
 
     try {
-      String.format("%s save-parser -s %s -r matrix %s", Util.binPath, testSchemaFile, savedParserFile.getName()) !
```
stevedlawrence: These tests used the scala ``!`` to execute the command to save a parser, then used expected to run the command to use the saved parser. We only configure Expect to have the daffodil-lib/src/test stuff on the classpath, so this wouldn't be able to find built-in-formats.xsd. So instead use Expect to execute this command so it's gets the right classpath info.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/IIBase.scala
```diff
@@ -165,17 +165,7 @@ abstract class IIBase( final override val xml: Node, xsdArg: XMLSchemaDocument,
...
-    val prop = getAttributeOption("schemaLocation")
-    prop.map { text =>
-      if (text.contains("edu/illinois/ncsa/daffodil")) {
-        SDW("schemaLocation property uses deprecated edu/illinois/ncsa/daffodil path instead of org/apache/daffodil. Converting to new path.")
-        text.replace("edu/illinois/ncsa/daffodil", "org/apache/daffodil")
-      } else {
-        text
-      }
-    }
-  }
```
stevedlawrence: This is the hack that is now removed.

build.sbt
mbeckerle: I will fix these eclipse project classpath things. I would prefer to let you break them with this PR, then do a straight PR to fix them. Just to avoid confusion, as I have too many balls in the air at the moment.

stevedlawrence: Ok, works for me.

PR #41, opened 2018-02-14
Schema cache now works whether you check all top-levels or not.

mbeckerle:
DAFFODIL-1893


mbeckerle:
This thing was driving me crazy with wasted time working on even modest size schemas. Turns out it was pretty easy to fix.

PR #42, opened 2018-02-15
Add back deprecated support for built-in-formats.dfdl.xsd

stevedlawrence:
- Claim DFDLGeneralFormatDefinition.dfdl.xsd as our own, filling it with
      the default properties from built-in-formats.xsd. Schemas should use
      this instead of built-in-formats.xsd
    - Modify all TDML tests to use DFDLGeneralFormatDefinition.dfdl.xsd
    - Move built-in-formats.dfdl.xsd to the original NCSA location, but
      deprecate it and add an SDW to alert users not to use it
    - Reverts dependency changes to build.sbt so not all projects depend on
      core/test
    - Rename all DFDL xsd's to have a dfdl.xsd extension
    - Revert change to CLI utils that added built-in-formats.xsd to the
      classpath. That is no longer needed.
    
    DAFFODIL-1987

(Note, split into two commits for easier review. The first commit has the core changes. The second commit is just changing all the tdml files to use DFDLGeneralFormatDefinition.dfdl.xsd and add dfdl.xsd extensions to schemas missing it)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd
```diff
@@ -0,0 +1,102 @@
...
+        />
+      </dfdl:defineFormat>
+
+      <!--
+        This format is deprecated and is available only for backwards
+        compatability. It will eventually be removed and should not be used.
+      -->
+      <dfdl:defineFormat name="daffodilTest1">
+        <dfdl:format ref="GeneralFormat" />
+      </dfdl:defineFormat>
```
stevedlawrence: daffodilTest1 can be removed, it should only be available when using built-in-formats.xsd

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mbeckerle: Minor, but I think this name is excessively long. The "Definition" suffix adds nothing. Can this file just be DFDLGeneralFormat.dfdl.xsd.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd
```diff
@@ -0,0 +1,102 @@
...
+-->
+
+<xs:schema xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+  <!-- Intended for inclusion into other schemas. Has no targetNamespace on purpose -->
+  <xs:annotation>
+    <xs:appinfo source="http://www.ogf.org/dfdl/">
+
+      <dfdl:defineFormat name="GeneralFormat">
```
mbeckerle: I'm assuming this is the same as daffodilTest1. I'm not going to verify this.

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/cli_schema_04.dfdl.xsd
```diff
@@ -0,0 +1,158 @@
...
+      <dfdl:format ref="ex:daffodilTest1" initiator="" terminator="" separator=""
+        separatorSuppressionPolicy="anyEmpty" separatorPosition="infix"
+        alignment="implicit" alignmentUnits="bits" trailingSkip="0"
+        lengthKind="delimited" documentFinalTerminatorCanBeMissing="yes" encoding="ASCII"
+        leadingSkip="0" textTrimKind="none" representation="text" occursCountKind="parsed"
+        textNumberRep="standard" ignoreCase="no" initiatedContent="no" />
+    </xs:appinfo>
+  </xs:annotation>
+    
+    <xs:simpleType name="simpleAssert">
```
mbeckerle: Do we need a schema this big if we're just verifying that we get the SDW when built-in-formats.xsd is used?

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd
stevedlawrence: Sure thing

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd
stevedlawrence: Yep, just sorted

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/cli_schema_04.dfdl.xsd
stevedlawrence: I just copied another schema and changed the import--didn't realize is was so large. I'll trim it down to just what is needed.

daffodil-lib/src/main/resources/IBMdefined/GeneralPurposeFormat.xsd
```diff
@@ -43,7 +43,7 @@
 * <xs:schema xmlns:tns="http://www.ogf.org/dfdl/dfdl-1.0/examples/" ... >
 *
 * <xs:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/examples/"
-*  schemaLocation="DFDLGeneralFormatDefinition.xsd"/>
+*  schemaLocation="DFDLGeneralFormatDefinition.dfdl.xsd"/>
```
mbeckerle: This change is not correct for this file. I think this file wants to be reverted back to what it was.

btw: do we get to remove IBM's copyright on this file, or do we have to keep it and add the ASF license block?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/IIBase.scala
```diff
@@ -165,7 +165,13 @@ abstract class IIBase( final override val xml: Node, xsdArg: XMLSchemaDocument,
     res
   }.value
 
-  final lazy val schemaLocationProperty = getAttributeOption("schemaLocation")
+  final lazy val schemaLocationProperty = {
+    val prop = getAttributeOption("schemaLocation")
+    if (prop.isDefined && prop.get == "edu/illinois/ncsa/daffodil/xsd/built-in-formats.xsd") {
+      SDW("schemaLocation property uses deprecated include/import of edu/illinois/ncsa/daffodil/xsd/built-in-formats.xsd. Use org/apache/daffodil/xsd/DFDLGeneralFormatDefinition.dfdl.xsd instead and change the dfdl:format ref to \"GeneralFormat\".")
```
mbeckerle: So will all our tests that use daffodilTest1 or daffodilNewProperties1 (which is also defined in built-in-formats) all produce SDWs now, or is there a way they are suppressed for our own tests? daffodilNewProperties1 is used quite a bit.

daffodil-lib/src/main/resources/IBMdefined/GeneralPurposeFormat.xsd
stevedlawrence: Will revert changes to this file. I think because IBM signed the SGA it's correct to use just the Apache license block?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/IIBase.scala
stevedlawrence: All the tests in the Daffodil repo have been converted to using DFDLGeneralFormat now. So only external schemas will get warnings, which I'd rather not have a way to suppress. We want this annoying so people will fix it and we can remove it.

There were only a handful of tests that actually needed daffodilNewProperties1 so I just got rid of the use that format and defined all the necessary properties in individual tests. When we add new properites, these tests will likely break, but it's pretty it's easy to just add the new properties to them, there's only like 5 or so, and we won't be adding new properties all that often.

PR #44, opened 2018-02-16
Handle ChoiceGroups correctly in immediatelyEnclosingModelGroup

jadams-tresys:
This one took a while to track down for what ended up being a one line
change.  Before when immediatelyEnclosingModelGroup was called on a
child element of a ChoiceGroup, it would return the enclosing ModelGroup
of the ChoiceGroup, not the ChoiceGroup itself.  This caused issues
when the CoiceGroup's parent would have initiatedContent="yes", as the
child element of the ChoiceGroup would typically not have an initiator
defined.

This commit fixes one of the issues preventing the IBM-TLOG schema
project from working correctly.

DAFFODIL-1885

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
@@ -231,7 +231,7 @@ trait Term
         None
       }
       case gr: GroupRef => gr.asModelGroup.immediatelyEnclosingModelGroup
-      case gdd: GlobalGroupDef => gdd.groupRef.asModelGroup.immediatelyEnclosingModelGroup
+      case gdd: GlobalGroupDef => Some(gdd.groupRef.asModelGroup)
       case ged: GlobalElementDecl => ged.elementRef.immediatelyEnclosingModelGroup
       case ct: ComplexTypeBase => {
         None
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -230,6 +230,7 @@ trait Term
         Assert.invariant(this.isInstanceOf[Root])
         None
       }
+      case gcgd: GlobalChoiceGroupDef => Some(gcgd.groupRef.asModelGroup)
```
stevedlawrence: I imagine we also need a GlobalSequenceGroupDef for the case where there are nested sequences? Perhaps this can just be GlobalGroupDef, which is the parent class of Sequence and Choice GroupDefs? But actually, that case already exists just a few lines down. Maybe the bug is on that line, and that line should be changed to the ``Some(...)`` that you have here instead of getting the immediately enclosing model group?

jadams-tresys: I just tested it out and this fixes the issue and doesn't seem to cause any issues in our test suite, so that works for me.

mbeckerle: What you fixed is that this was providing not the enclosing model group, but 2-enclosings out, in the case where the term was the model-group of a global group definition.

The same bug is here for sequences.

Change the line below for parent is a GlobalGroupDef: 

  case ggd: GlobalGroupDef => Some(ggd.groupRef.asModelGroup)

This subsumes your change line, so you can remove that.

mbeckerle: Ooops. Exactly same as what SL said. My job is at the redundancy dept. is where I work.

PR #45, opened 2018-02-16
Added tests that clarify what we allow in appinfo annotations.

mbeckerle:
Users requested they can combine their annotations with DFDL annotations inside the same appinfo block.

This issue was closed with "wont fix" as it would dilute our ability to do good diagnostic messages, and there is a workaround.

This PR is just for some tests added in order to characterize Daffodil's behavior vis a vis appinfo. I.e., what we allow, what we don't. E.g., it assures that we DO allow non-native attributes on DFDL annotation elements, and that XML comments are allowed throughout them, that we don't allow duplicate ones, and that we don't allow the behavior this user was requesting.

DAFFODIL-1902

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestAppinfoSyntax.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestAppinfoSyntax.scala
@@ -1,0 +1,243 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.dsom
+
+import scala.xml.Elem
+
+import org.apache.daffodil.xml.XMLUtils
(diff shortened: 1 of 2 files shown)


daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestAppinfoSyntax.scala
```diff
@@ -0,0 +1,243 @@
...
+  val dfdl = XMLUtils.dfdlAppinfoSource // XMLUtils.DFDL_NAMESPACE
+
+  /**
+   * Test that non-native attributes can be placed on DFDL appinfo annotations.
+   */
+  @Test def testAppinfoWithNonNativeAttributes() {
+    val expected = "this is a non-native attribute"
+    val nnURI = "urn:nonNativeAttributeNamespaceURN"
+    val sc = <xs:schema xmlns:xs={ xsd } xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:foo={ nnURI }>
+               <xs:include schemaLocation="org/apache/daffodil/xsd/built-in-formats.xsd"/>
```
stevedlawrence: Should change to org/apache/daffodil/xsd/DFDLGeneralFormat.xsd and daffodiLTest1 to GeneralFormat. Same for other tests.

mbeckerle: These are entirely internal tests in daffodil-core-unittest. So I can change these, but to clarify then, is built-in-formats.xsd entirely deprecated now for all uses, or just for external uses?

stevedlawrence: I'd prefer to just deprecate it entirely. DFDLGeneralFormat is virtually the same thing, but in the org/apache namespace. Probably best to avoid uses of the old edu/illinois namespace.

stevedlawrence: Also note that the above test has built-in-formats.xsd in org/apache, a recent commit moved it back to edu/illinois for backwards compatability. That's probably the cause of the TravisCI failure. I'd guess switching to DFDLGeneralFormat should fix it

PR #46, opened 2018-02-20
Added test to illustrate issue DAFFODIL-1907.

mbeckerle:
(Just adds a test to scala debug)

This bug reported by users trying to annotate an XSD with both DFDL 
annotations and an unrelated annotation language.

Problem is that like DFDL, this additional annotation language defines
annotation elements that have xsd:attribute definitions in them. 

And Daffodil issues SDE on those attribute definitions because DFDL
doesn't use attributes.

But they're not being used to describe data. They're just being used in 
annotation elements that Daffodil (or any DFDL processor) can simply
ignore. 

DAFFODIL-1907

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/otherAnnotationLanguage.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/otherAnnotationLanguage.xsd
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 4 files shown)


stevedlawrence:
+1, but should we hold off the merge until after 2.1.0 is released?

mbeckerle:
I concur that we should not merge until after 2.1.0 is released, assuming no other patches. This fix would be for NATO, so I'm not thrilled that there are known issues bothering them and that we release anyway, but this particular issue, the right fix for it isn't 100% clear to me. So I think we have to delay it.

mbeckerle:
This will be merged as a part of PR https://github.com/apache/incubator-daffodil/pull/49

PR #48, opened 2018-02-27
All SDW have WarnIDs to support suppression of warnings.

mbeckerle:
Fixing DAFFODIL-1874, DAFFODIL-1838, and also 
https://github.com/DFDLSchemas/NACHA/issues/2

DAFFODIL-1874, DAFFODIL-1838

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -17,132 +17,115 @@
 
 package org.apache.daffodil.grammar
 
-import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.processors._
-import org.apache.daffodil.schema.annotation.props.gen._
-import org.apache.daffodil.dpath.NodeInfo.PrimType
-import org.apache.daffodil.dsom.InitiatedTerminatedMixin
-import org.apache.daffodil.dsom.ElementBase
 import java.lang.{ Long => JLong }
+
+import scala.Boolean
+import scala.Long
+
+import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.dpath.NodeInfo
+import org.apache.daffodil.dpath.NodeInfo.PrimType
+import org.apache.daffodil.dsom.ElementBase
 import org.apache.daffodil.dsom.ExpressionCompilers
+import org.apache.daffodil.dsom.InitiatedTerminatedMixin
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.grammar.primitives.AlignmentFill
+import org.apache.daffodil.grammar.primitives.BCDDecimalDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.BCDDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.BCDDecimalRuntimeLength
+import org.apache.daffodil.grammar.primitives.BCDIntegerDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.BCDIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.BCDIntegerRuntimeLength
+import org.apache.daffodil.grammar.primitives.BinaryBoolean
+import org.apache.daffodil.grammar.primitives.BinaryDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.BinaryDecimalRuntimeLength
+import org.apache.daffodil.grammar.primitives.BinaryDouble
+import org.apache.daffodil.grammar.primitives.BinaryFloat
+import org.apache.daffodil.grammar.primitives.BinaryIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.BinaryIntegerRuntimeLength
+import org.apache.daffodil.grammar.primitives.CaptureContentLengthEnd
+import org.apache.daffodil.grammar.primitives.CaptureContentLengthStart
+import org.apache.daffodil.grammar.primitives.CaptureValueLengthEnd
+import org.apache.daffodil.grammar.primitives.CaptureValueLengthStart
+import org.apache.daffodil.grammar.primitives.ComplexNilOrContent
+import org.apache.daffodil.grammar.primitives.ConvertTextBooleanPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextBytePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextCombinator
+import org.apache.daffodil.grammar.primitives.ConvertTextDatePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextDateTimePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextDecimalPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextDoublePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextFloatPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextIntPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextIntegerPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextLongPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextNonNegativeIntegerPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextShortPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextTimePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextUnsignedBytePrim
+import org.apache.daffodil.grammar.primitives.ConvertTextUnsignedIntPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextUnsignedLongPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextUnsignedShortPrim
+import org.apache.daffodil.grammar.primitives.DelimiterStackCombinatorElement
+import org.apache.daffodil.grammar.primitives.DynamicEscapeSchemeCombinatorElement
+import org.apache.daffodil.grammar.primitives.ElementCombinator
+import org.apache.daffodil.grammar.primitives.ElementParseAndUnspecifiedLength
+import org.apache.daffodil.grammar.primitives.ElementUnused
+import org.apache.daffodil.grammar.primitives.HexBinaryDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.HexBinaryEndOfBitLimit
+import org.apache.daffodil.grammar.primitives.HexBinarySpecifiedLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalRuntimeLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerRuntimeLength
+import org.apache.daffodil.grammar.primitives.Initiator
+import org.apache.daffodil.grammar.primitives.InputValueCalc
+import org.apache.daffodil.grammar.primitives.LeadingSkipRegion
+import org.apache.daffodil.grammar.primitives.LeftCenteredPadding
+import org.apache.daffodil.grammar.primitives.LiteralCharacterNilOfSpecifiedLength
+import org.apache.daffodil.grammar.primitives.LiteralNilDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.LiteralValueNilOfSpecifiedLength
+import org.apache.daffodil.grammar.primitives.LogicalNilValue
+import org.apache.daffodil.grammar.primitives.NilLiteralCharacter
+import org.apache.daffodil.grammar.primitives.OVCRetry
+import org.apache.daffodil.grammar.primitives.OnlyPadding
+import org.apache.daffodil.grammar.primitives.PackedDecimalDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.PackedDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.PackedDecimalRuntimeLength
+import org.apache.daffodil.grammar.primitives.PackedIntegerDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.PackedIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.PackedIntegerRuntimeLength
+import org.apache.daffodil.grammar.primitives.PaddingInfoMixin
+import org.apache.daffodil.grammar.primitives.PrefixLength
+import org.apache.daffodil.grammar.primitives.RightCenteredPadding
+import org.apache.daffodil.grammar.primitives.RightFill
+import org.apache.daffodil.grammar.primitives.SimpleNilOrValue
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthExplicit
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthExplicitCharacters
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthImplicit
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthImplicitCharacters
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthPattern
+import org.apache.daffodil.grammar.primitives.StringDelimitedEndOfData
+import org.apache.daffodil.grammar.primitives.StringOfSpecifiedLength
+import org.apache.daffodil.grammar.primitives.Terminator
+import org.apache.daffodil.grammar.primitives.TrailingSkipRegion
+import org.apache.daffodil.grammar.primitives.ZonedTextIntPrim
+import org.apache.daffodil.processors.TextJustificationType
+import org.apache.daffodil.schema.annotation.props.Found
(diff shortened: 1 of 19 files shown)


project/Dependencies.scala
```diff
@@ -39,7 +39,7 @@ object Dependencies {
   lazy val cli = Seq( 
     "org.fusesource.jansi" % "jansi" % "1.14",
     "org.rogach" %% "scallop" % "0.9.5", // new version avail. 2.0.5 requires code changes
-    "net.sf.expectit" % "expectit-core" % "0.8.1" % "test",
+    "net.sf.expectit" % "expectit-core" % "0.8.1" % "test"
```
mbeckerle: I do not know that this trailing comma hurts anything, but it sure looks wrong.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -17,132 +17,115 @@
...
+import org.apache.daffodil.schema.annotation.props.gen.BinaryNumberRep
+import org.apache.daffodil.schema.annotation.props.gen.LengthKind
+import org.apache.daffodil.schema.annotation.props.gen.LengthUnits
+import org.apache.daffodil.schema.annotation.props.gen.NilKind
+import org.apache.daffodil.schema.annotation.props.gen.Representation
+import org.apache.daffodil.schema.annotation.props.gen.TextNumberRep
+import org.apache.daffodil.schema.annotation.props.gen.YesNo
+import org.apache.daffodil.util.PackedSignCodes
+import org.apache.daffodil.xml.GlobalQName
+import org.apache.daffodil.xml.XMLUtils
```
stevedlawrence: What happened here? Where these just autosorted by eclipse? Perhaps we should make it part of our code review process that when we add new imports we add them sorted so that these autoformat changes don't overwhelm patch review?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
```diff
@@ -43,8 +44,12 @@ abstract class Gram(contextArg: SchemaComponent)
   extends OOLAGHostImpl(contextArg) {
 
   final def SDE(str: String, args: Any*): Nothing = context.SDE(str, args: _*)
+
+  @deprecated("2018-02-27", "A WarnID is now required.")
```
stevedlawrence: Do anymore usages of this exist? Can just be removed now?

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala
```diff
@@ -91,19 +92,22 @@ trait SavesErrorsAndWarnings {
...
    */
+  @deprecated("2018-02-27", "Must supply a WarnID now.")
   def schemaDefinitionWarningUnless(testThatWillWarnIfFalse: Boolean, str: String, args: Any*): Unit = macro SDEMacros.schemaDefinitionWarningUnlessMacro
 
   /**
    * Use this form if you need to be able to suppress the warning
    */
   def schemaDefinitionWarningUnless(warnID: WarnID, testThatWillWarnIfFalse: Boolean, str: String, args: Any*): Unit = macro SDEMacros.schemaDefinitionWarningUnlessSuppressMacro
 
+  @deprecated("2018-02-27", "Must supply a WarnID now.")
```
stevedlawrence: Can these be removed too?

project/Dependencies.scala
```diff
@@ -18,7 +18,15 @@
...
-  lazy val common = core ++ infoset ++ test
+
+  // Note cli added to commmon on purpose.
+  // Due to some obscure bug, updateClassifers was not pulling the
+  // dependencies for the daffodil-cli module into lib_managed properly.
+  // (It found the source jars and javadoc jars, but not the regular code jar for
+  // Scallop aka org.rogach.)
+  // So compilation would fail for daffodil-cli from missing org.rogach library.
+  // Added to here forces these dependencies to be resolved. 
+  lazy val common = core ++ infoset ++ test ++ cli
```
stevedlawrence: So I'm pretty sure this is a bug with SBT. The steps I've found to reliably reproduce this:
```bash
git clean -xdf
sbt daffodil-cli/compile # compiles all sources, including daffodil-cli
sbt clean # cleans everything except daffodil-cli subproject (since we don't aggregate it)
sbt updateClassifiers daffodil-cli/updateClassifiers # update classifiers, note this does not download cli dependencies (sbt bug?)
sbt daffodil-cli/compile # fails due to missing cli dependencies
```
I suspect the issue here is that ``sbt clean`` empties the lib_managed directory. But daffodil-cli thinks the lib_managed directory is up to date since daffodil-cli was never cleaned. Since daffodil-cli thinks lib_managed is correct, it doesn't pull the dependencies in lib_managed, resulting in a failure of missing dependencies. I've confirmed this issue exists with sbt 1.1.1, so an sbt update won't help. This feels like a bug in sbt, but a workaround is to just run ``sbt daffodil-cli/clean`` after running ``sbt clean``. This removes the daffodil-cli state that thinks lib_managed is correct and causes it to pull dependencies. I've confirmed that the above works if I insert that command.

If the ``sbt daffodil-cli/clean`` solution seems reasonable and works for you, I'd prefer to revert this change and not have the CLI be a dependency of everything (for example, there's no reason to pull in CLI stuff when using the JAPI). Or alternatively, just never use sbt clean and instead use ``git clean -xdf`` and just be careful to make sure all untracked files that are needed are committed.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mbeckerle: I looked into this Eclipse IDE behavior. Besides organizing the imports that are automatically inserted into groups it also alphabetizes, and there appears to be no way to turn this off for java or scala. You can specify the grouping order for Java and Scala separately. E.g., the order is java, scala, org, com. But there is no setting to get Eclipse to only add them at the bottom and don't re-order. There just isn't a setting for this. You can turn off the whole mechanism,... but this ability to just add all the imports that are missing is very time saving. So I'd prefer not to turn this off. 

So I guess yeah, we should specify that the imports should be alphabetized within groups, and the group order is java, scala, org, com (for scala code), and java, javax, org, com (for java code).

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
mbeckerle: Yes, these can be removed entirely.

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala
mbeckerle: Yes. I'll remove them all.

project/Dependencies.scala
mbeckerle: Why doesn't sbt clean automatically clean daffodil-cli like it does all the other subprojects? 

The only exceptional behavior we need/want is for test and that's simply because they take too long to run. We don't want to run the cli tests automatically, but for everything else we *do* want it to act across daffodil-cli as well. 

Or can this not be achieved?

I can revert this change certainly, and the workaround is fine, but treating daffodil-cli as an entirely separate project but still having it be a subproject is awkward.  

Does any sbt task act across all the sub modules including daffodil-cli, or is daffodil-cli entirely separate now?

stevedlawrence: Unfortunately, I'm not sure SBT has a good to way to disable aggregation of a specified command for a subproject. You can either have "do not aggregate this entire subproject" or "do not aggregate a specified command for all subprojects", but you can't combine the two.

However, SBT does have a builtin "integration" test concept. Basically, you put integration tests in ``src/it/{scala,java,resources}``, and then all integration tests can be run via ``sbt it`` or something. This is sortof similar to what we had with the scala-cli stuff, except CLI tests don't share scala/java/resources directories like they did before and can now, but I don't think that really matters with our current usage. There isn't any sharing going on. Also, I'm not sure CLI tests are really "integration" tests, but I'm not sure that really matters. I'm also not sure if we can make it work with debug, which was one of the reasons for getting rid of the scala-cli stuff.

I'll create a bug to look into if this wold work as a better alternative, and allow us to reenable aggregation and avoid these types of issues.

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala
```diff
@@ -57,7 +57,8 @@ trait ThrowsSDE {
 
   final def schemaDefinitionError(str: String, args: Any*): Nothing = SDE(str, args: _*) // long form synonym
```
mbeckerle: Above at line 85, missed deprecating another form of SDW call without warnID. Need to find where these are all called also and add more warnIDs. 

Why this review tool won't let me put a comment on an unchanged line of a file.... is annoying!

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
stevedlawrence: Sounds reasonable to me. How does it handle other things. For example, we have a handful of things that start with "net", "unit", "passera", and then some things like "import Maybe._","import NodeInfo._", etc.

Also, out of curiousity, does it do anything different with things in the same package namespace? I.e. since the packages are org.apache.daffodil, does it always put org.apache.daffodil.whatever imports at the top/bottom, or are they sorted just like any other import org?

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
```diff
@@ -21,14 +21,33 @@ import org.apache.daffodil.exceptions.ThrowsSDE
...
+  case object AppinfoNoSource extends WarnID; forceConstruction(AppinfoNoSource)
+  case object AppinfoDFDLSourceWrong extends WarnID; forceConstruction(AppinfoDFDLSourceWrong)
+  case object QueryStylePathExpression extends WarnID; forceConstruction(QueryStylePathExpression)
+  case object PatternEncodingSlashW extends WarnID; forceConstruction(PatternEncodingSlashW)
+  case object NamespaceDifferencesOnly extends WarnID; forceConstruction(NamespaceDifferencesOnly)
+  case object AlignmentNotSame extends WarnID; forceConstruction(AlignmentNotSame)
+  case object InconsistentLengthKind extends WarnID; forceConstruction(InconsistentLengthKind)
+  case object FacetExplicitLengthOutOfRange extends WarnID; forceConstruction(FacetExplicitLengthOutOfRange)
+  case object DeprecatedFunctionDAFError extends WarnID; forceConstruction(DeprecatedFunctionDAFError)
+  case object DeprectatedBuiltInFormats extends WarnID; forceConstruction(DeprectatedBuiltInFormats)
```
stevedlawrence: Typo in Deprec_t_ated.

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
```diff
@@ -21,14 +21,33 @@ import org.apache.daffodil.exceptions.ThrowsSDE
...
+  case object DeprecatedEncodingNameUSASII7BitPacked extends WarnID; forceConstruction(DeprecatedEncodingNameUSASII7BitPacked)
```
stevedlawrence: Missing C in USASCII

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
```diff
@@ -21,14 +21,33 @@ import org.apache.daffodil.exceptions.ThrowsSDE
...
+  case object UnsupportedAttributeSchemaLocation extends WarnID; forceConstruction(UnsupportedAttributeSchemaLocation)
+  case object UnsupportedAttributeBlockDefault extends WarnID; forceConstruction(UnsupportedAttributeBlockDefault)
+  case object UnsupportedAttributeFinalDefault extends WarnID; forceConstruction(UnsupportedAttributeFinalDefault)
+  case object ChoiceInsideHiddenGroup extends WarnID; forceConstruction(ChoiceInsideHiddenGroup)
+  case object DeprecatedPropertySeparatorPolicy extends WarnID; forceConstruction(DeprecatedPropertySeparatorPolicy)
+  case object PathNotToArray extends WarnID; forceConstruction(PathNotToArray)
+  case object RegexPatternZeroLength extends WarnID; forceConstruction(RegexPatternZeroLength)
```
stevedlawrence: Might be worth sorting this list for readability (e.g. all the deprecated things would be together).

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala
```diff
@@ -82,32 +83,29 @@ trait ThrowsSDE {
 trait SavesErrorsAndWarnings {
 
   def SDE(id: String, args: Any*): Nothing
```
stevedlawrence: We should probably also have a similar ErrorID for SDE's. That would be the first step towards adding support for something like JALoP or various other auditing mechanisms. Not worth doing as part of this since there's no use for those ID's like there are for warn IDs, but something to think about for the future. Could be a good task for a new developer.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -121,6 +121,25 @@
...
+                 <xsd:enumeration value="deprecatedFunctionDAFError"/>
+                 <xsd:enumeration value="deprectatedBuiltInFormats"/>
+                 <xsd:enumeration value="deprecatedEncodingNameUSASII7BitPacked"/>
+                 <xsd:enumeration value="unsupportedAttributeSchemaLocation" />
+                 <xsd:enumeration value="unsupportedAttributeBlockDefault" />
+                 <xsd:enumeration value="unsupportedAttributeFinalDefault" />
+                 <xsd:enumeration value="choiceInsideHiddenGroup" />
+                 <xsd:enumeration value="deprecatedPropertySeparatorPolicy" />
+                 <xsd:enumeration value="pathNotToArray" />
+                 <xsd:enumeration value="regexPatternZeroLength" />
```
stevedlawrence: Same typos here as above. Probably worth sorting as well.

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
mbeckerle: check.

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
mbeckerle: Check.

daffodil-lib/src/main/scala/org/apache/daffodil/api/WarnID.scala
mbeckerle: And in dafext.xsd. Really this should be generated from dafext.xsd, but I'll leave that for some other developer dabbling in code generation. 

I will alphabetize this anyway for easier maintenance.

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala
mbeckerle: See also JIRA DAFFODIL-628 (JAF Logging).

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
mbeckerle: Check

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mbeckerle: File FNFunctions.scala is a good example:

Before:

import org.apache.daffodil.processors._
import org.apache.daffodil.infoset._
import org.apache.daffodil.processors.unparsers.UnparseError
import org.apache.daffodil.api.Diagnostic
import org.apache.daffodil.api.DataLocation
import Maybe._
import org.apache.daffodil.exceptions._
import com.ibm.icu.util.Calendar
import scala.math.BigDecimal.RoundingMode
import com.ibm.icu.util.TimeZone
import org.apache.daffodil.calendar.DFDLDateTime
import org.apache.daffodil.calendar.DFDLCalendar
import org.apache.daffodil.calendar.DFDLTime
import org.apache.daffodil.calendar.DFDLDate
import java.lang.{ Number => JNumber, Byte => JByte, Short => JShort, Integer => JInt, Long => JLong, Float => JFloat, Double => JDouble, Boolean => JBoolean }

After:

import java.lang.{ Number => JNumber, Byte => JByte, Short => JShort, Integer => JInt, Long => JLong, Float => JFloat, Double => JDouble, Boolean => JBoolean }

import scala.math.BigDecimal.RoundingMode

import org.apache.daffodil.api.DataLocation
import org.apache.daffodil.api.Diagnostic
import org.apache.daffodil.calendar.DFDLCalendar
import org.apache.daffodil.calendar.DFDLDate
import org.apache.daffodil.calendar.DFDLDateTime
import org.apache.daffodil.calendar.DFDLTime
import org.apache.daffodil.exceptions._
import org.apache.daffodil.infoset._

import org.apache.daffodil.processors._
import org.apache.daffodil.processors.unparsers.UnparseError
import org.apache.daffodil.util.Maybe._

import com.ibm.icu.util.Calendar
import com.ibm.icu.util.TimeZone

I do not understand the blank line in the middle of the org.apache imports after infoset._. 

Note that it also undid the import Maybe._, adding the global package name.

mbeckerle:
Review comment changes applied. 3 commits squashed into 1.

PR #49, opened 2018-02-28
Daffodil 1907 non dfdl annotation

mbeckerle:
Allows import of other XML Schemas for other annotation languages.
    
    Annotation languages (like DFDL) use XSD attribute declarations in the
    annotation language's XML syntax.
    
    DFDL processors like Daffodil need to allow other annotation languages
    to be annotated onto DFDL schemas.
    
    This creates a problem when importing the XSD for another annotation
    language, since that XSD wants to also use things like XSD attribute
    declarations if it validates the imported XSD as a DFDL schema because
    it will refuse to accept the attribute decls. 
    
    To allow import of a full-function XML Schema for an annotation
    language, Daffodil looks at schemas for use of the DFDL_NAMESPACE. If
    the root element of the schema uses the DFDL_NAMESPACE (e.g., to define
    the dfdl prefix), then the file is treated as a DFDL schema. If the
    DFDL_NAMESPACE is not used, then the file is not imported. It is
    ignored.
    
    DAFFODIL-1907

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
@@ -28,6 +28,7 @@ import org.apache.daffodil.oolag.OOLAG
 import org.xml.sax.SAXException
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.Misc
+import org.apache.daffodil.exceptions.Assert
 
 /**
  * represents one schema document file
@@ -101,18 +102,11 @@ final class DFDLSchemaFile(val sset: SchemaSet,
     validationDiagnostics_ :+= sde
   }
 
-  def fatalError(exception: SAXParseException) = {
-    val sde = new SchemaDefinitionError(this.schemaFileLocation, "Fatal error loading schema due to %s", exception)
-    validationDiagnostics_ :+= sde
-    // parser throws out of fatalErrors.
-  }
-
-  private lazy val loader = {
-    val ldr = new DaffodilXMLLoader(this)
-    // val shouldValidate = sset.validateDFDLSchemas
-    // ldr.setValidation(shouldValidate) // TODO: Validation not occurring JIRA DFDL-1473. Fix later.
-    ldr
-  }
+  /**
+   * Called on a fatal exception. The parser/validator throws the exception after
+   * this call returns.
+   */
+  def fatalError(exception: SAXParseException) = error(exception) // same as non-fatal exception.
 
   private def loadedNode = LV('loadedNode) {
     def die(e: Throwable) = {
@@ -120,7 +114,14 @@ final class DFDLSchemaFile(val sset: SchemaSet,
     }
     val node = try {
       log(LogLevel.Resolver, "Loading %s.", diagnosticDebugName)
-      val node = loader.load(schemaSource)
+      val ldr = new DaffodilXMLLoader(this)
+      //
+      // We do not want to validate here ever, because we have to examine the
+      // root xs:schema eleemnt of a schema to decide if it is a  DFDL schema
+      // at all that we're even supposed to compile.
+      //
+      ldr.setValidation(false)
+      val node = ldr.load(schemaSource)
       schemaDefinitionUnless(node != null, "No XML Node could be loaded from %s.", schemaSource)
       node
     } catch {
@@ -132,8 +133,29 @@ final class DFDLSchemaFile(val sset: SchemaSet,
 
   lazy val node = loadedNode
 
-  def iiXMLSchemaDocument = LV('iiXMLSchemaDocument) {
+  lazy val isDFDLSchemaFile = iiXMLSchemaDocument.isDFDLSchema
+
+  lazy val iiXMLSchemaDocument = LV('iiXMLSchemaDocument) {
     val res = loadXMLSchemaDocument(seenBefore, Some(this))
+    if (res.isDFDLSchema && sset.validateDFDLSchemas) {
+      //
+      // We validate DFDL schemas, only if validation is requested.
+      // Some things, tests generally, want to turn this validation off.
+      //
+
+      val ldr = new DaffodilXMLLoader(this)
+      ldr.setValidation(true)
+      try {
+        ldr.load(schemaSource) // validate as XML file with XML Schema for DFDL Schemas
+        ldr.validateSchema(schemaSource) // validate as XSD (catches UPA errors for example)
+      } catch {
+        case _: org.xml.sax.SAXParseException =>
+          // ok to absorb this. We have captured fatal exceptions in the
+          // error handler. 
+        case e: Exception =>
+          Assert.invariantFailed("Unexpected exception type " + e)
+      }
+    }
     res
   }.value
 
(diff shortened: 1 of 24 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
```diff
@@ -107,20 +107,15 @@ final class DFDLSchemaFile(val sset: SchemaSet,
...
-
   private def loadedNode = LV('loadedNode) {
     def die(e: Throwable) = {
       SDE("Error loading schema due to %s.", Misc.getSomeMessage(e).getOrElse("an unknown error."))
     }
     val node = try {
       log(LogLevel.Resolver, "Loading %s.", diagnosticDebugName)
-      val node = loader.load(schemaSource)
+      val ldr = new DaffodilXMLLoader(this)
+      ldr.setValidation(false)
```
stevedlawrence: Does this also resolve DFDL-1473 mentioned above?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
```diff
@@ -124,6 +125,46 @@ final class XMLSchemaDocument(xmlArg: Node,
...
+  private def hasDFDLNamespaceDefinition: Boolean = {
+    val scope = xml.scope
+    val pre = scope.getPrefix(XMLUtils.DFDL_NAMESPACE)
+    val hasSomePrefixForDFDLNamespace = pre ne null
+    val res = hasSomePrefixForDFDLNamespace ||
+      {
+        val default = scope.getURI(null)
+        val hasDefaultNamespaceAsDFDLNamespace = default == XMLUtils.DFDL_NAMESPACE.toString
+        hasDefaultNamespaceAsDFDLNamespace
+      }
```
stevedlawrence: Could the inside of the brackets be moved to before the res so it just becomes ``val res = hasSomePrefixForDFDLNamespace || hasDefaultNamespaceAsDFDLNamespace``? This currently avoids some code if DFDL is the not the default namespace, but that doesn't seem all that intensive and this is a bit hard to read.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
```diff
@@ -124,6 +125,46 @@ final class XMLSchemaDocument(xmlArg: Node,
...
+   * We do this so that other annotation languages can co-exist with DFDL.
+   * That is, they can use all of XML Schema including things DFDL doesn't allow
+   * like attribute decls, but only when those schemas are only needed
+   * to process non-DFDL annotations. Since Daffodil ignores non-DFDL annotations
+   * entirely, Daffodil won't run into these non-DFDL allowed things like
+   * attribute decls. But validators like Xerces will see the regular
+   * import/include and process normally, which enables validation of
+   * all annotations, DFDL and otherwise.
+   *
+   * Fixes DAFFODIL-1909
```
stevedlawrence: Note sure it's necessary to put bug numbers in the code. If someone is interested in that, they can aways do a git blame and git log which will have bug information.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestImportOtherAnnotationSchema.scala
```diff
@@ -0,0 +1,40 @@
...
+}
+
+class TestImportOtherAnnotationSchema {
+
+  import TestImportOtherAnnotationSchema._
+
+  //DFDL-1907
+  @Test def test_importOtherAnnotationSchema1() { runner.runOneTest("importOtherAnnotationSchema1") }
+
+}
```
stevedlawrence: Out of curiosity, what happens if a DFDL schema uses types tha tare imported/included from a non-DFDL schema? I assume we would get some kind of "type not found" error since the non-DFDL schema was ignored, thus Daffodil knows nothing about those types? If so, seems like a reasonable behavior since the warning should alert the user about the schema being ignored.

mbeckerle: Exactly. It will get a missing type - or whatever kind of reference it is. That's the right behavior. Plus the warning about the schema being ignored is a pretty good diagnostic.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
mbeckerle: Potentially, but this change is probably the cause of the regressions I didn't catch in sbt test. I'll see if this one can also be addressed.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
mbeckerle: Yes, another way to do this is
{code}
val = cheapThingToCompute = .....
lazy val = expensiveThingToCompute = ....
val res = cheapThingToCompute || expensiveThingToCompute
{code}
This gets the clarity and the efficiency minus the lazy val overhead. I wouldn't do this is the runtime though, but in the compiler I'm ok with it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
mbeckerle: While I agree with you generally, this issue is such a subtle one that I thought it worthwhile to direct a programmer to read the bug ticket before messing further with this code. However I will change the phrasing so this looks less like some sort of advocated practice.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
mbeckerle: Yes. This resolves DAFFODIL-1473 also.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -518,7 +518,14 @@ class DaffodilXMLLoader(val errorHandler: org.xml.sax.ErrorHandler) {
   def load(source: DaffodilSchemaSource): scala.xml.Node = {
     var xercesNode: Node = null
     if (doValidation) {
-      xercesNode = xercesAdapter.load(source.newInputSource()) // validates
+      //
+      // We need a try/catch here as error recording needs to
```
mbeckerle: Remove comment. Try/catch is now in caller of this.

daffodil-tdml/src/test/resources/tdml-with-embedded-schema-errors.tdml
```diff
@@ -32,8 +32,7 @@
       <tdml:documentPart type="text"><![CDATA[abcdef]]></tdml:documentPart>
     </tdml:document>
     <tdml:errors>
-      <tdml:error>xs:format</tdml:error>
-      <tdml:error>invalid dfdl annotation</tdml:error>
+      <tdml:error>'import' tag should be added to</tdml:error>
```
mbeckerle: This is a negative test case where the errors being reported just are different now. It complains now about fn:string, and can't get to where it complains about xs:format.

mbeckerle:
Squashed 3 commits into 1 before merging.

PR #50, opened 2018-02-28
Move CLI tests to sbt integration

stevedlawrence:
Move CLI tests to src/it/ rather than being part of src/test. This means
that we can reenable aggregation of the daffodil-cli subproject without
fear of 'sbt test' running the CLI tests and extending the test cycle.
With aggregation reenabled, this also means that we no longer need to
run commands explicitly in the daffodil-cli project (i.e. things like
'sbt compile' or 'sbt clean' will compile/clean all subprojects,
including daffodil-cli), simplifying build process and preventing some
bugs in sbt from cropping up.

To run cli tests, use 'sbt it:test'.

To debug cli tests, use 'sbt debugIt:test'.

DAFFODIL-1911


eclipse-projects/cli-test/.project
```diff
@@ -36,5 +36,10 @@
 			<type>2</type>
 			<locationURI>DAFFODIL_ROOT/daffodil-cli/src/test</locationURI>
 		</link>
+		<link>
+			<name>src/it</name>
+			<type>2</type>
+			<locationURI>DAFFODIL_ROOT/daffodil-cli/src/it</locationURI>
+		</link>
```
stevedlawrence: Mike, does this and the .classpath modification look correct to you? I don't have an eclipse setup to test. Basically just copied things that had src/test and changed to src/it

mbeckerle: Looks right. When you've committed this ping me and I'll verify this right away. but I think this is right and it's just going to work.  One does not usually run these tests from eclipse, so really this is just making sure eclipse compiles these along with everything else when changes are made.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -43,9 +42,8 @@ object Util {
   val binPath = Paths.get(dafRoot, "daffodil-cli", "target", "universal", "stage", "bin", String.format("daffodil%s", (if (isWindows) ".bat" else ""))).toString()
 
   def getExpectedString(filename: String, convertToDos: Boolean = false): String = {
-    val rsrc = Misc.getRequiredResource(outputDir + filename)
-    //val source = scala.io.Source.fromFile(outputDir + filename)
-    val source = scala.io.Source.fromFile(rsrc)
+    val is = getClass.getResourceAsStream(outputDir + filename)
```
mbeckerle: There is a reason to use the Misc library, which is that you get a diagnostic that lists out the classpath including all the directories, jars, etc. if the thing isn't found. I think it fixes up the typical "no leading slash" problem as well.  I think I wrote the Misc.getRequiredResource when trying to debug classpath issues with include/import.

project/Dependencies.scala
```diff
@@ -39,12 +39,12 @@ object Dependencies {
...
     "org.rogach" %% "scallop" % "0.9.5", // new version avail. 2.0.5 requires code changes
-    "net.sf.expectit" % "expectit-core" % "0.8.1" % "test",
+    "net.sf.expectit" % "expectit-core" % "0.8.1" % "it,test",
   )
 
   lazy val test = Seq(
-    "junit" % "junit" % "4.11" % "test",
-    "com.novocode" % "junit-interface" % "0.11" % "test",
-    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
+    "junit" % "junit" % "4.11" % "it,test",
```
mbeckerle: Question. You did not put a space after the comma i.e., it's not "it, test" it's "it,test". Is that significant?

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
stevedlawrence: getRequiredResource returned a path inside a jar (e.g. file://path/to/foo.jar!/path/inside/jar.txt) and I could not find a way to easily get the contents of that file. I suspect in the past it found the file on the classpath so didn't need to look inside a jar. I'll reinvestigate this to figure out how to get the contents of a resource inside a jar.

mbeckerle: val dataURI = Misc.getRequiredResource("iso8859.doc.dat")
    val dataInput = dataURI.toURL.openStream()

project/Dependencies.scala
stevedlawrence: I don't think it matters to sbt, but most examples I can find of this type of thing do not use spaces. This seems to be the sbt standard for this.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
stevedlawrence: This fixed it. Thanks!

PR #51, opened 2018-03-05
Updates to LICENSE/NOTICE

stevedlawrence:
- Add explicit paths to files that are not Apache licensed
- Append -bin to LICENSE/NOTICE in daffodil-cli to clarify that it is
  for binary release files
- Add acknowledgments to UoI NCSA, Tresys, and IBM for their pre-ASF
  contributions
- Remove content from NOTICE files that is not legally required to be
  there (e.g. BSD copyrights)
- Remove Utility.scala. This was copied from the Scala library and we
  only use the most basic of features that are already accessible in scala
  lib. This file isn't needed.
- Move the PrettyPrinter.scala file out of the scalaLib directory. This
  file was originally based off of a file in the scala library but with
  a bug fixed. Since then, this file was essentially rewritten with
  completely different behavior. Since it is no longer looks anything
  like the scala lib version except for some function names, move it out
  of the directory to make this clear.
- Correct the short notice for W3C documents

DAFFODIL-1906


daffodil-cli/NOTICE-bin
```diff
@@ -0,0 +1,81 @@
...
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+Based on source code originally developed by
+- The Univerisity of Illinois National Center for Supercomputing Applications (http://www.ncsa.illinois.edu/)
+- Tresys Technology (http://www.tresys.com/)
+- International Business Machines Corporation (http://www.ibm.com)
+
+******************
```
johnament: Do not list out licenses in the NOTICE file.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PrettyPrinter.scala
```diff
@@ -15,9 +15,8 @@
  * limitations under the License.
  */
 
-package org.apache.daffodil.xml.scalaLib
+package org.apache.daffodil.xml
```
johnament: Everything beyond this line seems unrelated to NOTICE/LICENSE.

stevedlawrence: Part of the below changes removes the Utility.scala file, which came from the Scala library. We don't really use it anymore so I figured it was better to remove it so we do not need to include it in the LICENSE file that mentioned the mentions the other scala lib file (UniquenessCache.scala).

The PrettyPrinter scala file (which is the remaining of the below changes) originally came from the scala library, but we have since completely rewritten it (basically only the function names still exist), so I'm not sure it is considered a derivative work anymore. The rest of the below changes were to just make it clear that it PrettyPretty is not a scala lib file by moving it out of the scalaLib directory. This required changing some imports. So the below changes are more about removing the potential for someone to think PrettyPretty is a scala lib file with an incorrect license. That's somewhat related to this patchset, so thought it reasonable to include it here.

daffodil-cli/NOTICE-bin
stevedlawrence: I took this layout from another Apache project (can't remember which one). Looking around it seems every project has their own variant. It would be really nice if there were an ASF provided template that shows the expected format of these files. It seems the LICENSE/NOTICE files are some of the biggest hangups for incubators, and I think part of that is the lack of resources.

Simple enough change to make though. Will have a fixup shortly.

daffodil-cli/LICENSE-bin
```diff
@@ -306,9 +318,23 @@ subcomponents is subject to the terms and conditions of the following licenses.
...
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc83_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc83_02.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc7133_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc81_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc121_02.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc_41_83_04_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc82_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc121_01.dfdl.xsd
+    - daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaspc83_03.dfdl.xsd
+  The content is available under the Open Grid Format license:
```
stevedlawrence: Copyright lines missing from OGF section of license. See https://daffodil.apache.org/docs/dfdl/#_Toc398030837

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PrettyPrinter.scala
johnament: Ok, sounds fair enough.

daffodil-cli/LICENSE-bin
```diff
@@ -232,8 +238,14 @@ subcomponents is subject to the terms and conditions of the following licenses.
     OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
     OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-  This product bundles source from 'Scala'. Specifically 'Utility.scala' and
-  'UniquenessCache.scala' The source is available under a BSD LICENSE.
+  This product bundles source from 'Scala', including the following files:
```
johnament: Since this is the binary license, you'll want to point to the actual binary files and their locations.

daffodil-cli/LICENSE-bin
```diff
@@ -261,14 +273,14 @@ subcomponents is subject to the terms and conditions of the following licenses.
...
 
-  This product bundles material copied or derived from W3C. Specific files
-  include:
-    - XMLSchema.dtd (https://www.w3.org/2001/XMLSchema.dtd)
-    - XMLSchema.xsd (https://www.w3.org/2001/XMLSchema.xsd)
-    - xml.xsd (https://www.w3.org/2001/xml.xsd)
-    - datatypes.dtd (https://www.w3.org/2001/datatypes.dtd)
-    - XMLSchema_for_DFDL.xsd (https://www.w3.org/2001/XMLSchema.dtd)
-  The source is available under the following license:
+  This product bundles material copied or derived from W3C, including the
```
johnament: Same here.  If this path exists in the binary great, but more likely than not it won't.

daffodil-cli/LICENSE-bin
```diff
@@ -306,9 +318,23 @@ subcomponents is subject to the terms and conditions of the following licenses.
     permission. Title to copyright in this work will at all times remain with
     copyright holders.
 
-  This product bundles source from the Open Grid Forum Data Format Description
-  Language (DFDL) v1.0 Specification. The source is available under the following
-  license:
+  This product bundles content from the Open Grid Forum Data Format Description
+  Language (DFDL) v1.0 Specification, including the following files:
+    - daffodil-test/src/test/resources/org/apache/daffodil/ibm-tests/dpaspc7132.dfdl.xsd
```
johnament: We don't typically see test files bundled in the binary release.  Do you in fact ship tests in your binary?

daffodil-cli/NOTICE-bin
```diff
@@ -0,0 +1,74 @@
...
+The Apache Software Foundation (http://www.apache.org/).
+
+Based on source code originally developed by
+- The Univerisity of Illinois National Center for Supercomputing Applications (http://www.ncsa.illinois.edu/)
+- Tresys Technology (http://www.tresys.com/)
+- International Business Machines Corporation (http://www.ibm.com)
+
+The following NOTICE information applies to binary components distributed with this project:
+
+Apache Commons IO
```
johnament: Generally speaking, we only include apache notices if they include more restrictive information.  So the Xerces one should be replicated.

daffodil-cli/NOTICE-bin
```diff
@@ -0,0 +1,74 @@
...
+
+  This product includes software developed by the
+  JDOM Project (http://www.jdom.org/).
+
+Jansi
+  Copyright (C) 2009, Progress Software Corporation and/or its
+  subsidiaries or affiliates.  All rights reserved.
+
+Jackson JSON processor
+  # Jackson JSON processor
```
johnament: Where did the text for this come from?  I'm not saying its wrong, just want to make sure its correct.

daffodil-cli/LICENSE-bin
stevedlawrence: Check, will fix.

daffodil-cli/LICENSE-bin
stevedlawrence: Tests aren't bundled in the binary helpers, will remove.

daffodil-cli/NOTICE-bin
stevedlawrence: This comes from the META_INF/NOTICE file inside the jackson-core jar file. The file is in their github here: https://github.com/FasterXML/jackson-core/blob/master/src/main/resources/META-INF/NOTICE

daffodil-cli/NOTICE-bin
stevedlawrence: Did you mean, so *only* the Xerces one should be replicated, since it has extra stuff in the NOTICE? 

Doing some more verification, I noticed "xml-apis" jar and "xml-resolver" jar weren't referenced in either the LICENSE or NOTICE file. Both are Apache v2 licensed and part of Apache, so their NOTICE should be excluded unless it has anything extra like Xerces does, correct?

daffodil-cli/NOTICE-bin
johnament: Ok, looks good then.

daffodil-cli/NOTICE-bin
johnament: Yes, only the xerces notice needs to be propagated, since the others are just boiler plate.

PR #52, opened 2018-03-05
Fix eclipse project files for daffodil-cli module.

mbeckerle:
In the recent changes to simplify sbt tasks so that specialized tasks
aren't needed for the cli except for testing it, the src/test directory
was renamed src/it.

This change adjusts the eclipse project setup to match.
DAFFODIL-1914


stevedlawrence:
+1 to new changes

PR #54, opened 2018-03-09
Adds missing properties to DFDLGeneralFormat.dfdl.xsd

mbeckerle:
Tolerates encodingErrorPolicy='error' with a warning.

DAFFODIL-1915


mbeckerle:
I have tested this version with sbt test, sbt it:test, and I have tested all DFDLSchemas and FOUO schemas. All passes on my workstation. I have verified that eclipse is also working properly and one can run Daffodil's tests there, and tests of the various DFDL schemas there.

PR #55, opened 2018-03-19
Fixes the issue of separated empty optional elements, (ie. 1:2::4:5).

jadams-tresys:
This was one of the bugs that was preventing the IBMTLOG4690 schema
project form working correctly.  Note that formats that have these
optional empty elements will not necessarily round trip.

DAFFODIL-1886

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/LocalElementGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/LocalElementGrammarMixin.scala
@@ -108,12 +108,14 @@ trait LocalElementGrammarMixin extends GrammarMixin { self: ElementBase =>
   // thinking about how occursCountKind='stopValue' works.)
 
   private lazy val separatedContentAtMostN = prod("separatedContentAtMostN") {
-    (separatedContentAtMostNWithoutTrailingEmpties // FIXME: We don't know whether we can absorb trailing separators or not here.
-    // We don't know if this repeating thing is in trailing position, or in the middle
-    // of a sequence. There is also ambiguity if the enclosing sequence and this sequence
+    (separatedContentAtMostNWithoutTrailingEmpties
+    // There may be ambiguity here if the enclosing sequence and this sequence
     // have the same separator.
-    //      ~
-    //      RepAtMostTotalN(self, maxOccurs, separatedEmpty) // absorb extra separators, if found.
+    ~
+    (if (couldBeLastElementInModelGroup)
+      RepAtMostTotalN(self, maxOccurs, separatedEmpty) // absorb extra separators, if found.
+    else
+      EmptyGram)
     )
   }
 
@@ -206,12 +208,12 @@ trait LocalElementGrammarMixin extends GrammarMixin { self: ElementBase =>
       case (Trailing___, Implicit__, UNB, ___) if (!isLastDeclaredRequiredElementOfSequence) => SDE("occursCountKind='implicit' with unbounded maxOccurs only allowed for last element of a sequence")
       case (Trailing___, Implicit__, UNB, min) => separatedContentWithMinUnbounded
       case (Trailing___, Implicit__, max, min) if min > 0 => separatedContentWithMinAndMax
-      case (Trailing___, Implicit__, max, ___) => separatedContentAtMostN // FIXME: have to have all of them - not trailing position
+      case (Trailing___, Implicit__, max, ___) => separatedContentAtMostN
       case (TrailingStr, Implicit__, UNB, ___) if (!isLastDeclaredRequiredElementOfSequence) => SDE("occursCountKind='implicit' with unbounded maxOccurs only allowed for last element of a sequence")
       case (TrailingStr, Implicit__, UNB, ___) => separatedContentWithMinUnboundedWithoutTrailingEmpties // we're depending on optionalEmptyPart failing on empty content.
       case (TrailingStr, Implicit__, max, ___) => separatedContentAtMostNWithoutTrailingEmpties
       case (Always_____, Implicit__, UNB, ___) => separatedContentWithMinUnbounded
-      case (Always_____, Implicit__, max, ___) => separatedContentAtMostN
+      case (Always_____, Implicit__, max, ___) => separatedContentAtMostNWithoutTrailingEmpties
       case (Always_____, Parsed____, ___, __2) => separatedContentZeroToUnbounded
       case (Always_____, StopValue_, ___, __2) => separatedContentZeroToUnbounded
       case (policy /**/ , ock /****/ , max, __2) => SDE("separatorSuppressionPolicy='" + policy + "' not allowed with occursCountKind='" + ock + "'.")
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesElementKinds.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesElementKinds.scala
@@ -112,23 +112,19 @@ case class ComplexTypeCombinator(ct: ComplexTypeBase, body: Gram) extends Termin
 case class SequenceCombinator(sq: SequenceTermBase, rawTerms: Seq[Gram])
   extends Terminal(sq, true) {
 
-  override lazy val isEmpty = {
-    val rt = rawTerms
-    val frt = rt.filterNot { _.isEmpty }
-    val res = frt.isEmpty
-    res
-  }
+  lazy val terms = rawTerms.filterNot{ _.isEmpty }
 
-  private val mt: Gram = EmptyGram
-  lazy val body = rawTerms.foldRight(mt) { _ ~ _ }
+  override lazy val isEmpty = terms.isEmpty
 
-  lazy val terms = rawTerms.filterNot { _.isEmpty }
+  lazy val parsers = terms.map { term =>
+    term.parser
+  }.toArray
 
   lazy val unparsers = terms.map { term =>
     term.unparser
-  }.toVector
+  }.toArray
 
-  lazy val parser: DaffodilParser = new SequenceCombinatorParser(sq.termRuntimeData, body.parser)
+  lazy val parser: DaffodilParser = new SequenceCombinatorParser(sq.termRuntimeData, parsers)
 
   override lazy val unparser: DaffodilUnparser = {
     sq.checkHiddenSequenceIsDefaultableOrOVC
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
@@ -46,7 +46,7 @@ class ComplexTypeUnparser(rd: RuntimeData, bodyUnparser: Unparser)
   }
 }
 
-class SequenceCombinatorUnparser(ctxt: ModelGroupRuntimeData, childUnparsers: Vector[Unparser])
+class SequenceCombinatorUnparser(ctxt: ModelGroupRuntimeData, childUnparsers: Array[Unparser])
   extends CombinatorUnparser(ctxt)
   with ToBriefXMLImpl {
 
@@ -57,21 +57,18 @@ class SequenceCombinatorUnparser(ctxt: ModelGroupRuntimeData, childUnparsers: Ve
   // Sequences of nothing (no initiator, no terminator, nothing at all) should
   // have been optimized away
   Assert.invariant(childUnparsers.length > 0)
-
-  // Since some of the grammar terms might have folded away to EmptyGram,
-  // the number of unparsers here may be different from the number of
-  // children of the sequence group.
-  Assert.invariant(ctxt.groupMembers.length >= childUnparsers.length)
+  Assert.invariant(ctxt.groupMembers.length == childUnparsers.length)
 
   override lazy val childProcessors: Seq[Processor] = childUnparsers
 
   def unparse(start: UState): Unit = {
 
-    start.groupIndexStack.push(1L) // one-based indexing
 
     var index = 0
     var doUnparser = false
     val limit = childUnparsers.length
+
+    start.groupIndexStack.push(1L) // one-based indexing
 
     while (index < limit) {
       doUnparser = false
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
@@ -119,18 +119,26 @@ class DynamicEscapeSchemeParser(escapeScheme: EscapeSchemeParseEv,
   }
 }
 
-class SequenceCombinatorParser(rd: TermRuntimeData, bodyParser: Parser)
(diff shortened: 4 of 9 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/LocalElementGrammarMixin.scala
```diff
@@ -211,7 +214,7 @@ trait LocalElementGrammarMixin extends GrammarMixin { self: ElementBase =>
       case (TrailingStr, Implicit__, UNB, ___) => separatedContentWithMinUnboundedWithoutTrailingEmpties // we're depending on optionalEmptyPart failing on empty content.
       case (TrailingStr, Implicit__, max, ___) => separatedContentAtMostNWithoutTrailingEmpties
       case (Always_____, Implicit__, UNB, ___) => separatedContentWithMinUnbounded
-      case (Always_____, Implicit__, max, ___) => separatedContentAtMostN
+      case (Always_____, Implicit__, max, ___) => separatedContentAtMostNWithoutTrailingEmpties
```
jadams-tresys: I am not 100% sure that this is correct behaviour, but it results in the same behaviour as before these changes.  This prevents TestArrayOptionalElem.test_backtrack1Text from failing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -119,18 +119,26 @@ class DynamicEscapeSchemeParser(escapeScheme: EscapeSchemeParseEv,
...
+  val numChildParsers = childParsers.size
 
   def parse(start: PState): Unit = {
+    var i: Int = 0
+
     start.mpstate.groupIndexStack.push(1L) // one-based indexing
 
-    bodyParser.parse1(start)
+    while (i < numChildParsers) {
+      val parser = childParsers(i)
```
stevedlawrence: childParsers is a Seq, which is implemented in Scala as a LinkedList, which does not have constant time access to random elements. So this loop is essentially O(N<superscript>2</superscript>). Instead of a Seq, we want this to be something with constant time access like an Array, or we want to use things like .head() and .tail() to iterate through the list. An Array will probably be more efficient since I think tail() does alot of work behind the scenes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/RepParsers.scala
```diff
@@ -249,7 +249,7 @@ class RepUnboundedParser(occursCountKind: OccursCountKind.Value, rParser: Parser
     var priorState: PState.Mark = null
     var markLeakCausedByException = false
 
-    Assert.invariant(initialState.processorStatus eq Success)
+    if (initialState.processorStatus ne Success) return
```
stevedlawrence: Want to avoid returns. Scala implements them as throws which are relatively heavy. Can we instead just wrap the call to parseAllRepeats with the if(initialState.processorStatus ne Success) check?

Also, can you explain why this happens? It seems if the initial state of a RepParser is not Success, we never should have enetered the RepParser in the first place. Maybe there's a bug elsewhere?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/LocalElementGrammarMixin.scala
```diff
@@ -112,8 +112,11 @@ trait LocalElementGrammarMixin extends GrammarMixin { self: ElementBase =>
...
     // of a sequence. There is also ambiguity if the enclosing sequence and this sequence
     // have the same separator.
-    //      ~
-    //      RepAtMostTotalN(self, maxOccurs, separatedEmpty) // absorb extra separators, if found.
+    ~
+    (if (couldBeLastElementInModelGroup)
+      RepAtMostTotalN(self, maxOccurs, separatedEmpty) // absorb extra separators, if found.
+    else
+      EmptyGram)
     )
```
stevedlawrence: Does this the issue mentioned in the comment? Seems like it does, and if so the comment should be removed?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesElementKinds.scala
```diff
@@ -113,22 +113,21 @@ case class SequenceCombinator(sq: SequenceTermBase, rawTerms: Seq[Gram])
   extends Terminal(sq, true) {
 
   override lazy val isEmpty = {
-    val rt = rawTerms
-    val frt = rt.filterNot { _.isEmpty }
-    val res = frt.isEmpty
+    val ft = terms.filterNot { _.isEmpty }
+    val res = ft.isEmpty
     res
```
stevedlawrence: This looks odd to me. terms is ``rawTerms.filterNot { _.isEmpty }`` and then you do ``terms.filterNot { _.isEmpty }``. Seems like that second filterNot will do nothing, so can this just be ``val res = terms.isEmpty``?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesElementKinds.scala
```diff
@@ -113,22 +113,21 @@ case class SequenceCombinator(sq: SequenceTermBase, rawTerms: Seq[Gram])
...
   }
 
-  private val mt: Gram = EmptyGram
-  lazy val body = rawTerms.foldRight(mt) { _ ~ _ }
-
-  lazy val terms = rawTerms.filterNot { _.isEmpty }
+  lazy val terms = rawTerms.filterNot{ _.isEmpty }
+  lazy val parsers = terms.map { term =>
+    term.parser
+  }.toVector
```
stevedlawrence: I have a comment below about things being a Seq[Parser]. So I guess although the type is a Seq, the underlying class will always be a Vector. I'd say we should be specific about the type and make it Vector[Parser]. Though, wondering if this should be an Array instead. If we're only using it for random access, I think Array is going to be more efficient.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
```diff
@@ -54,25 +54,22 @@ class SequenceCombinatorUnparser(ctxt: ModelGroupRuntimeData, childUnparsers: Ve
 
   override def nom = "Sequence"
 
-  // Sequences of nothing (no initiator, no terminator, nothing at all) should
-  // have been optimized away
-  Assert.invariant(childUnparsers.length > 0)
```
stevedlawrence: In what cases can there no be zero unparsers? Seems like there's a bug elsewhere if there are zero unparsers in a sequence combinator.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
```diff
@@ -54,25 +54,22 @@ class SequenceCombinatorUnparser(ctxt: ModelGroupRuntimeData, childUnparsers: Ve
...
-
   // Since some of the grammar terms might have folded away to EmptyGram,
   // the number of unparsers here may be different from the number of
   // children of the sequence group.
-  Assert.invariant(ctxt.groupMembers.length >= childUnparsers.length)
+  Assert.invariant(ctxt.groupMembers.length == childUnparsers.length)
```
stevedlawrence: This doesn't match the above comment, and also seems odd. If the SequenceCombinatorUnparser is filtering out emptry grammars, then it seems reasonable that there might be childUnparsers than groupMembers. Or does this mean things aren't being filtered out now?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
stevedlawrence: I think this is wrong. Although the type is Seq, it is actually a Vector. I'd say change type to be Vector[Parser], or maybe Array[Parser]. I think Arrays or more efficient than vectors if all we need is random access.

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
```diff
@@ -546,7 +546,7 @@
   </tdml:defineSchema>
 
   <tdml:parserTestCase name="initiatedContentSimple1"
-    model="initiatedContentSimple1" description="TBD" root="e">
+    model="initiatedContentSimple1" description="TBD" root="e" roundTrip="false">
```
stevedlawrence: Why doesn't this round trip now?

daffodil-test/src/test/scala/org/apache/daffodil/section12/delimiter_properties/TestDelimiterProperties.scala
```diff
@@ -54,7 +54,7 @@ class TestDelimiterProperties {
   @Test def test_DelimProp_05() { runner_02.runOneTest("DelimProp_05") }
   @Test def test_DelimProp_06() { runner_02.runOneTest("DelimProp_06") }
   @Test def test_DelimProp_07() { runner_02.runOneTest("DelimProp_07") }
-  @Test def test_initiatedContentSimple1() { runner_02.runOneTest("initiatedContentSimple1") }
+  @Test def test_initiatedContentSimple1() { runner_02.trace.runOneTest("initiatedContentSimple1") }
```
stevedlawrence: Remove the .trace()

daffodil-test/src/test/scala/org/apache/daffodil/section14/sequence_groups/TestSequenceGroupInitiatedContent.scala
```diff
@@ -38,6 +38,6 @@ class TestSequenceGroupInitiatedContent {
   @Test def test_baseline() { runner_01.runOneTest("initiatedContentSeqBaseline") }
   @Test def test_1() { runner_01.runOneTest("initiatedContentSeq1") }
   @Test def test_2() { runner_01.runOneTest("initiatedContentSeq2") }
-  @Test def test_3() { runner_01.runOneTest("initiatedContentSeq3") }
+  @Test def test_3() { runner_01.trace.runOneTest("initiatedContentSeq3") }
```
stevedlawrence: Remove .trace

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
jadams-tresys: That is a good question.  I think I had set it to roundTrip="false" while debugging some things and I thought it would work now, but now it fails with an SDE saying No root element found, which is not what I would have expected.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
jadams-tresys: Added back in, doesn't cause any issues.  I think this was something that Mike and I threw in when we were initially debugging this issue and forgot to remove.

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
jadams-tresys: Nevermind, I simply accidentally deleted the root="e" when I was removing the roundTrip attribute.  This test works and roundTrips as expected.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
jadams-tresys: I think that this was another case where Mike and I were taking a look at this and thought this was something that would cause an issue with this change.  I changed it back and it doesn't seem to cause any issues.

stevedlawrence: If using == works, I would stick with that. The more restrictive an assert is, the better. I'm a little surprised it works, I would think something from groupMembers would get filtered out, but it's not completely unreasonable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
jadams-tresys: This should be addressed in the most recent commit

daffodil-core/src/main/scala/org/apache/daffodil/grammar/LocalElementGrammarMixin.scala
jadams-tresys: This should be addressed in the most recent commit

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementKindUnparsers.scala
jadams-tresys: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -119,18 +119,30 @@ class DynamicEscapeSchemeParser(escapeScheme: EscapeSchemeParseEv,
...
+      parser.parse1(start)
+      if (start.processorStatus ne Success) {
+        start.mpstate.groupIndexStack.pop()
+        return
```
stevedlawrence: return's should be avoided. Maybe remove this ``if`` block and instead change the loop to ``while (i > numChildren && start.processorStatus eq Success)``? This has the benefit that it will always execute ``groupIndexStack.pop()`` without duplicating code. Though, I'm not sure if still calling ``moveOverOneGroupIndexOnly`` would break anything, but that could be wrapped in a ``processorStatus eq Success`` check if it matters.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoiceGroupInitiatedContent.scala
```diff
@@ -47,7 +47,7 @@ class TestChoiceGroupInitiatedContent {
   @Test def test_initiatedContentChoice10() { runner_01.runOneTest("initiatedContentChoice10") }
 
   @Test def test_arrayOfChoice() { runner_01.runOneTest("arrayOfChoice") }
-  @Test def test_arrayOfChoice2() { runner_01.runOneTest("arrayOfChoice2") }
+  @Test def test_arrayOfChoice2() { runner_01.trace.runOneTest("arrayOfChoice2") }
```
stevedlawrence: Missed this trace and one below.

jadams-tresys:
I just pushed up a few more changes with a couple added tests demonstrating what works and what doesn't work with our current implementation of separatorSuppressionPolicy trailing.

jadams-tresys:
Seems like a spurious error from the Travis build.  The Travis build failed right away in code that wasn't touched at all (DaffodilTunables.scala).  Not sure what is going on with it, but it builds and tests just fine on my system.

PR #56, opened 2018-03-29
Improved toString of grammar and parser/unparser objects.

mbeckerle:
This is just to assist in debugging some thorny problems.

DAFFODIL-1920

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
@@ -180,8 +180,13 @@ abstract class NamedGram(context: SchemaComponent) extends Gram(context) {
   // Note: keep the toString really simple.
   // It causes much grief if toString uses complicated things that can fail or
   // that end up needing the name of this NamedGram again.
-  override def toString = name // + "(" + context.scPath.last + ")" //+ (if (isEmpty) "(Empty)" else "")
 
+  override def name = context match {
+    case nm: NamedMixin => nm.name
+    case _ => super.name
+  }
+
+  override def toString = "<" + name + ">" + super.name + "</" + name + ">"
 }
 
 /**
@@ -191,10 +196,5 @@ abstract class Terminal(contextArg: SchemaComponent, guard: Boolean)
   extends NamedGram(contextArg) {
 
   override def isEmpty = !guard
-
-  private lazy val realSC = context.asInstanceOf[SchemaComponent]
-  final override lazy val path = realSC.path + "@@" + diagnosticDebugName
-
-  override def toString = path // dangerous. What if realSC.path fails?
 
 }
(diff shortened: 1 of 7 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
```diff
@@ -192,9 +197,9 @@ abstract class Terminal(contextArg: SchemaComponent, guard: Boolean)
...
   override def isEmpty = !guard
 
-  private lazy val realSC = context.asInstanceOf[SchemaComponent]
-  final override lazy val path = realSC.path + "@@" + diagnosticDebugName
-
-  override def toString = path // dangerous. What if realSC.path fails?
+  //  private lazy val realSC = context.asInstanceOf[SchemaComponent]
+  //  final override lazy val path = realSC.path + "@@" + diagnosticDebugName
+  //
+  //  override def toString = path // dangerous. What if realSC.path fails?
```
stevedlawrence: Remove comments.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesElementKinds.scala
```diff
@@ -81,7 +82,14 @@ case class DelimiterStackCombinatorElement(e: ElementBase, body: Gram) extends T
   lazy val uInit = if (e.initiatorParseEv.isKnownNonEmpty) One(e.initiatorUnparseEv) else Nope
   lazy val uTerm = if (e.terminatorParseEv.isKnownNonEmpty) One(e.terminatorUnparseEv) else Nope
 
-  lazy val parser: DaffodilParser = new DelimiterStackParser((pInit.toList ++ pTerm.toList).toArray, e.termRuntimeData, body.parser)
+  lazy val delims = (pInit.toList ++ pTerm.toList)
+
+  override def toString() =
+    "<" + Misc.getNameFromClass(this) + " delims='" + delims.mkString(" ") + "'>" +
```
stevedlawrence: In the past we've had issues where delimters aren't correctly escaped according to HTML standards. This makes running things through xmllint to see a pretty indented version difficult. Shoud/can we run the delims through some sort of HTML-ifier so they are more friendly to such tools?

mbeckerle: Yes, should escape these. 
scala.xml.Utility.escape is insufficient however, as that (a) doesn't escape apostrophes because HTML doesn't have an &apos; entity (b) delims might be XML-illegal characters.

So we have to remap all XML-illegal chars to XML-legal PUA chars, but leave CR and CRLF alone, remap C0 and C1 control characters appropriately to numeric entity refs. Remap whitespace chars to visible glyphs.

We have all the pieces of this algorithm, just not in this particular combination.

mbeckerle: Fixed in most recent commit.

PR #57, opened 2018-03-30
Daffodil 1444 runtime backpointers

mbeckerle:
Removes backpointers from the RuntimeData.

This is a cleanup, but also part of preparing for improvements to the schema compiler, which was being forced to maintain backpointers due to the runtime needing runtime backpointers. 

Turns out, the runtime never used these backpointers, except to check one Assert in unparser.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
@@ -180,8 +180,13 @@ abstract class NamedGram(context: SchemaComponent) extends Gram(context) {
   // Note: keep the toString really simple.
   // It causes much grief if toString uses complicated things that can fail or
   // that end up needing the name of this NamedGram again.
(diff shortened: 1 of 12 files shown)


mbeckerle:
Created this wrong. Closing.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
```diff
@@ -192,9 +197,9 @@ abstract class Terminal(contextArg: SchemaComponent, guard: Boolean)
...
   override def isEmpty = !guard
 
-  private lazy val realSC = context.asInstanceOf[SchemaComponent]
-  final override lazy val path = realSC.path + "@@" + diagnosticDebugName
-
-  override def toString = path // dangerous. What if realSC.path fails?
+  //  private lazy val realSC = context.asInstanceOf[SchemaComponent]
+  //  final override lazy val path = realSC.path + "@@" + diagnosticDebugName
+  //
+  //  override def toString = path // dangerous. What if realSC.path fails?
```
jadams-tresys: I think the changes to this file got carried over from PR 56

PR #58, opened 2018-03-30
Eliminates backpointers from RuntimeData structures.

mbeckerle:
This cleanup is a step toward eliminating the double-linked non-sharing behavior
in the schema compiler, because maintaining these double-linked pointers
for the runtime was forcing them to exist in the schema compiler also. 

Turns out the runtime didn't need them except to check an Assert in one unparser. That check is removed, and all the backpointers can go away from RuntimeData and friends.

DAFFODIL-1444

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
@@ -150,8 +150,8 @@ final class DFDLSchemaFile(val sset: SchemaSet,
         ldr.validateSchema(schemaSource) // validate as XSD (catches UPA errors for example)
       } catch {
         case _: org.xml.sax.SAXParseException =>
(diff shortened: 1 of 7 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLSchemaFile.scala
```diff
@@ -150,8 +150,8 @@ final class DFDLSchemaFile(val sset: SchemaSet,
         ldr.validateSchema(schemaSource) // validate as XSD (catches UPA errors for example)
       } catch {
         case _: org.xml.sax.SAXParseException =>
-          // ok to absorb this. We have captured fatal exceptions in the
-          // error handler. 
+        // ok to absorb this. We have captured fatal exceptions in the
```
mbeckerle: Can revert this file. Just a whitespace change.

mbeckerle:
Nevermind. Broken idea.

PR #59, opened 2018-04-03
Add new tunable (errorOnUnsupportedJavaVersion) to allow running with older versions of Java

stevedlawrence:
Currently we intentionally throw an exception if Daffodil is not run
with Java 8 due to bugs in how Java 7 handles decoding errors. However,
there are some cases where a user has no choice except to run under Java
7. This patch adds a new tunable "errorOnUnsupportedJavaVersion" that causes
Java 8 check failures to be logged as a warning rather than thrown as an
exception. This allows the above use case to be possible, with the explicit
understanding that it is not fully tested and is know to have unexpected
behavior in some circumstances.

DAFFODIL-1921

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
@@ -21,18 +21,41 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.schema.annotation.props.gen.ParseUnparsePolicy
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.Logging
+import org.apache.daffodil.util.Misc
+import org.apache.daffodil.xml.DaffodilXMLLoader
 
 object DaffodilTunables {
 
   def apply(tunables: Map[String, String]): DaffodilTunables = {
-    new DaffodilTunables().setTunables(tunables)
+    apply().setTunables(tunables)
   }
 
   def apply(tunable: String, value: String): DaffodilTunables = {
-    new DaffodilTunables().setTunable(tunable, value)
+    apply().setTunable(tunable, value)
   }
 
-  def apply(): DaffodilTunables = new DaffodilTunables()
+  def apply(): DaffodilTunables = {
+    // override tunables from the global configuration file on the class path, if it exists
+    val (configOpt, _) = Misc.getResourceOption("/daffodil-config.xml")
+    val configTunables: Map[String, String] =
+      if (configOpt.isDefined) {
+        val loader = new DaffodilXMLLoader()
+        val node = loader.load(new URISchemaSource(configOpt.get))
+        val trimmed = scala.xml.Utility.trim(node)
+        val tunablesNode = (trimmed \ "tunables").headOption
+        val tunablesMap: Map[String, String] = tunablesNode match {
+          case None => Map.empty
+          case Some(tunableNode) => {
+            tunableNode.child.map { n => (n.label, n.text) }.toMap
+          }
+        }
+        tunablesMap
+      } else {
+        Map.empty
+      }
+
+    new DaffodilTunables().setTunables(configTunables)
+  }
 }
 
 case class DaffodilTunables(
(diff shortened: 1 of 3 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
```diff
@@ -105,7 +105,19 @@ case class DaffodilTunables(
...
+  // occur in the specific ways that Daffodil decodes data. For this reason,
+  // Daffodil throws an exception when it detects that Daffodil is not running
+  // under Java 8 or has this decoder bug. However, there are some cases where
+  // a user has no choice but to run on Java 7. Setting this tunable to false
+  // will cause Daffodil to log a warning rather than throw an exception so
+  // that a user can run Daffodil on unsupported Java versions, with the
+  // understanding that it is not fully tested and behavior may not be well
+  // defined. This boolean is experimental and should only be used by those
+  // that fully understand the risks.
+  val errorOnUnsupportedJavaVersion: Boolean = true)
```
mbeckerle: We're going to need to be able to globally set this to false and run all our tests etc without having to modify anything to specify this tunable.  For example by way of setting an environment variable. Do our tunables allow for this? 

In the past the tunables I've used only needed to be set here or there, not pervasively like we will need for this.

stevedlawrence: No, there isn't really a way to do that right now.

We could perhaps modify the TDML runner to read specific environment variables and set tunables based on that. Or similarly, we could set the default value in the DaffodilTunables class based on environment variables and then it would affect everything, (e.g. CLI, API, TDMLRunner) rather than just the TDML runner. For example:
```scala
class DaffodilTunables(
 ...
 val errorOnUnsupportedJavaVersion: Boolean = {
   val bool = scala.util.Properties.envOrElse("DAFFODIL_TUNABLE_ERROR_ON_UNSUPPORTED_JAVA_VERSION", "true")
   java.lagn.Boolean.valueOf(bool)
 }
 ...
)
```
That isn't too hairy of a change and makes it pretty easy to modify tunable values using environment variables.

Alternatively, considering that we only officially support Java 8+, perhaps we can get away with just manually changing the default value of the tunable to false when we want to do a one off Java 7 test?

mbeckerle: Well, I think we are perhaps kidding ourselves that we won't have to provide some level of support for this, so I think we really do need the env-var solution of some sort or it will just be too painful.

For each Daffodil tunable having prefix "DAFFODIL_TUNABLE_" in front of the name of the tunable seems the right choice to me. Or it could be DaffodilTunable_ given the caml-case tradition we're using.

mbeckerle: I think the right fix is not the TDML runner, but in the construction of the default values for the tunables object.

stevedlawrence: The more I think about environment variables to control this, the less I like it. It just seems like incorrect use of environment variables. Another option, which I think is a bit more of a standard, is to just use a configuration file. We already support a configuration file for the CLI, so this can just share that same logic for extracting. So I propose the following instead for the creation of DaffodilTunables:

1. Create a DaffodilTunables instance with the defaults specified in the source
2. If a global configuration file exists (e.g. /etc/daffodil/config.xml), read tunables from that and override the values set above
3. Local tunable options can then be provided to override those, for example via the -c option in the CLI, setTunables API, or TDMLRunner tunable mechanism.

So setting this tunable is just like setting any other tunable via a configuration file. We don't need any logic to extract and convert environtment variables, and the default code in DaffodilTunables isn't made messy.

I'm not sure if there is a standard way for java applications to find a configuration file in /etc, but I imagine something exists.

mbeckerle: The standard java-ish way to find a config file would be to search the classpath for it.

mbeckerle: A config file is a fine way to do this - so long as there's a way to configure it once and have it apply across many activities: tdml tests, cli, etc. without having to go in and edit all over the place.

stevedlawrence: Updated to include initializing tunables from "/daffodil-config.xml" on the classpath.

PR #61, opened 2018-04-11
Base64, gzip, and line-folding layering

mbeckerle:
Layering implementation.

Supports layerLengthKind implicit, explicit, and boundaryMark, but only certain ones for certain layerings.

See wiki page here for doc: 
    https://cwiki.apache.org/confluence/display/DAFFODIL/Proposal%3A+Data+Layering+for+base64+-+Super+Simplified

AIS not yet implemented, but base64_MIME, gzip, and line folding (IMF and iCal variants) are done.

My big concern here is that it is quite difficult to debug things. 

Java's scanner class saves us the hassle of block filling/enlarging,
etc. that is otherwise necessary when you want to use the regex
libraries against a stream.

Also put catch for ClassNotFound around compiler.reload for Java 7 vs.
Java 8. Base64 needs Java 8 classes currently, though there may be workarounds if we 
have to support Java 7.

Identified a couple cleanups where java.io.HashMap can be replaced with
our NonAllocatingMap - which uses Maybe types instead of null/notNull.

Added a charset so we can use the encoder to unparse AIS data that must
be armored with AIS_PAYLOAD_ARMORING 6 bit to 8 bit.

This is all part of layering, i.e., AIS format requires layering as does
base64, so merging this all onto same branch.

DAFFODIL-1734, DAFFODIL-1805

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
@@ -35,6 +35,8 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.SequenceRuntimeData
 import org.apache.daffodil.schema.annotation.props.Found
 import org.apache.daffodil.schema.annotation.props.PropertyLookupResult
+import org.apache.daffodil.processors.LayerTransformerEv
+import org.apache.daffodil.util.Maybe
 
 abstract class SequenceTermBase(
   final override val xml: Node,
@@ -44,7 +46,8 @@ abstract class SequenceTermBase(
   with Sequence_AnnotationMixin
   with SequenceRuntimeValuedPropertiesMixin
   with SequenceGrammarMixin
-  with SeparatorSuppressionPolicyMixin {
+  with SeparatorSuppressionPolicyMixin
+  with LayeringRuntimeValuedPropertiesMixin {
 
   requiredEvaluations(checkIfValidUnorderedSequence)
   requiredEvaluations(modelGroupRuntimeData.preSerialization)
@@ -244,6 +247,31 @@ abstract class SequenceTermBase(
       maybeCheckByteAndBitOrderEv,
       maybeCheckBitOrderAndCharset)
   }
+
+  private val layeredSequenceAllowedProps = Set("ref", "layerTransform", "layerEncoding", "layerLengthKind", "layerLength", "layerLengthUnits", "layerBoundaryMark")
+
+  final lazy val maybeLayerTransformerEv: Maybe[LayerTransformerEv] = {
+    if (maybeLayerTransformEv.isEmpty) Maybe.Nope
+    else { // need to check that only layering properties are specified
+      val localProps = this.formatAnnotation.justThisOneProperties
+      val localKeys = localProps.keySet
+      val disallowedKeys = localKeys.filterNot(k => layeredSequenceAllowedProps.contains(k))
+      if (disallowedKeys.size > 0)
+        SDE("Sequence has dfdl:layerTransform specified, so cannot have non-layering properties: %s", disallowedKeys.mkString(", "))
+
+      val lt = new LayerTransformerEv(maybeLayerTransformEv.get,
+        maybeLayerCharsetEv,
+        Maybe.toMaybe(optionLayerLengthKind),
+        maybeLayerLengthInBytesEv,
+        Maybe.toMaybe(optionLayerLengthUnits),
+        maybeLayerBoundaryMarkEv,
+        termRuntimeData)
+      lt.compile()
+      Maybe.One(lt)
+    }
+  }
+
+  final def isLayered = maybeLayerTransformerEv.isDefined
 
 }
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/LayeredSequence.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/LayeredSequence.scala
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.grammar.primitives
+
+import org.apache.daffodil.grammar.Terminal
+import org.apache.daffodil.dsom._
+import org.apache.daffodil.processors.parsers.{ Parser => DaffodilParser }
+import org.apache.daffodil.processors.unparsers.{ Unparser => DaffodilUnparser }
+import org.apache.daffodil.grammar.Gram
+import org.apache.daffodil.util.Misc
+import org.apache.daffodil.processors.parsers.LayeredSequenceParser
+import org.apache.daffodil.processors.unparsers.LayeredSequenceUnparser
+
+case class LayeredSequence(sq: SequenceTermBase, bodyTerm: Gram)
+  extends Terminal(sq, true) {
+
+  override def toString() =
+    "<" + Misc.getNameFromClass(this) + ">" +
+      bodyTerm.toString() +
+      "</" + Misc.getNameFromClass(this) + ">"
+
+  override lazy val parser: DaffodilParser =
+    new LayeredSequenceParser(sq.termRuntimeData, sq.maybeLayerTransformerEv.get, bodyTerm.parser)
+
+  override lazy val unparser: DaffodilUnparser = {
+    new LayeredSequenceUnparser(sq.modelGroupRuntimeData, sq.maybeLayerTransformerEv.get, bodyTerm.unparser)
+  }
+}
--- a/daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
@@ -1,0 +1,361 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.layers
+
+import org.apache.daffodil.util._
+import org.junit.Test
+import org.junit.Assert._
+import org.apache.daffodil.util.TestUtils
+import java.io.ByteArrayOutputStream
+import org.apache.commons.io.IOUtils
+import java.nio.charset.StandardCharsets
+import java.io.ByteArrayInputStream
+import java.io.InputStreamReader
+import scala.collection.JavaConversions._
+import org.apache.daffodil.xml.XMLUtils
+
+class TestLayers {
+
+  val example = XMLUtils.EXAMPLE_NAMESPACE
+
+  val B64Layer1Schema =
+    SchemaUtils.dfdlTestSchema(
+      <dfdl:format ref="tns:GeneralFormat"/>,
+      <xs:element name="e1" dfdl:lengthKind="implicit">
+        <xs:complexType>
+          <xs:sequence dfdl:layerTransform="base64_MIME" dfdl:layerLengthKind="boundaryMark" dfdl:layerBoundaryMark="!" dfdl:layerEncoding="iso-8859-1">
+            <xs:element name="s1" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="3"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>, elementFormDefault = "unqualified")
+
+  @Test def testParseB64Layer1() {
+    val sch = B64Layer1Schema
+    val data = "cGxl!" // encoding of "ple" + "!"
+    val infoset = <ex:e1 xmlns:ex={ example }><s1>ple</s1></ex:e1>
+    val (_, actual) = TestUtils.testString(sch, data)
+    TestUtils.assertEqualsXMLElements(infoset, actual)
+
+    val areTracing = false
+    TestUtils.testUnparsing(sch, infoset, data, areTracing)
+  }
+
+  val B64Layer2Schema =
+    SchemaUtils.dfdlTestSchema(
+      <dfdl:format ref="tns:GeneralFormat" lengthKind='delimited'/>,
+      <xs:element name="e1" dfdl:lengthKind="implicit">
+        <xs:complexType>
+          <xs:sequence dfdl:layerTransform="base64_MIME" dfdl:layerLengthKind="boundaryMark" dfdl:layerBoundaryMark="!" dfdl:layerEncoding="iso-8859-1">
+            <xs:element name="s1" type="xs:string"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>, elementFormDefault = "unqualified")
+
+  @Test def testParseB64Layer2() {
+    val sch = B64Layer2Schema
+    val data = "cGxl!" // encoding of "ple" + "!"
+    val infoset = <ex:e1 xmlns:ex={ example }><s1>ple</s1></ex:e1>
+    val (_, actual) = TestUtils.testString(sch, data)
+    TestUtils.assertEqualsXMLElements(infoset, actual)
+
+    val areTracing = false
+    TestUtils.testUnparsing(sch, infoset, data, areTracing)
+  }
+
+  val B64Layer3Schema =
+    SchemaUtils.dfdlTestSchema(
+      <dfdl:format ref="tns:GeneralFormat" lengthKind='delimited'/>,
+      <xs:element name="e1" dfdl:lengthKind="implicit">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:sequence dfdl:layerTransform="base64_MIME" dfdl:layerLengthKind="boundaryMark" dfdl:layerBoundaryMark="!" dfdl:layerEncoding="iso-8859-1">
+              <xs:element name="s1" type="xs:string"/>
+            </xs:sequence>
+            <xs:element name="s2" type="xs:string"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>, elementFormDefault = "unqualified")
+
+  @Test def testParseB64Layer3() {
+    val sch = B64Layer3Schema
+    val data = "cGxl" + "!" + "moreDataAfter"
+    val infoset = <ex:e1 xmlns:ex={ example }><s1>ple</s1><s2>moreDataAfter</s2></ex:e1>
+    val (_, actual) = TestUtils.testString(sch, data)
+    TestUtils.assertEqualsXMLElements(infoset, actual)
+
+    val areTracing = false
+    TestUtils.testUnparsing(sch, infoset, data, areTracing)
+  }
+
+  def makeGZIPData(text: String) = {
+    val baos = new ByteArrayOutputStream()
+    val gzos = new java.util.zip.GZIPOutputStream(baos)
+    IOUtils.write(text, gzos, StandardCharsets.UTF_8)
+    gzos.close()
+    val data = baos.toByteArray()
+    data
+  }
+
+  val text = """This is just some made up text that is intended to be
+a few lines long. If this had been real text, it would not have been quite
+so boring to read. Use of famous quotes or song lyrics or anything like that
+introduces copyright notice issues, so it is easier to simply make up
+a few lines of pointless text like this.""".replace("\n", " ")
+
+  @Test def testGZIPRoundTrips() {
+    val bais = new ByteArrayInputStream(makeGZIPData(text))
+    val gzis = new java.util.zip.GZIPInputStream(bais)
+    val rdr = new InputStreamReader(gzis, StandardCharsets.UTF_8)
+    val lines = IOUtils.readLines(rdr)
+    val textBack = lines.head
+    assertEquals(text, textBack)
+  }
+
+  val GZIPLayer1Schema =
+    SchemaUtils.dfdlTestSchema(
+      <dfdl:format ref="tns:GeneralFormat" layerLengthUnits="bytes" representation="binary"/>,
+      <xs:element name="e1" dfdl:lengthKind="implicit">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="len" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="4" dfdl:outputValueCalc="{ dfdl:contentLength(../x1, 'bytes') }"/>
+            <xs:element name="x1" dfdl:lengthKind="implicit">
+              <xs:complexType>
+                <xs:sequence dfdl:layerTransform="gzip" dfdl:layerLengthKind="explicit" dfdl:layerLength="{ ../len }">
+                  <xs:element name="s1" type="xs:string" dfdl:lengthKind="delimited"/>
+                </xs:sequence>
+              </xs:complexType>
+            </xs:element>
+            <xs:element name="s2" type="xs:string" dfdl:lengthKind="delimited"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>,
+      elementFormDefault = "unqualified")
+
+  def makeGZIPLayer1Data() = {
+    val gzipData = makeGZIPData(text)
+    val dataLength = gzipData.length
+    val baos = new ByteArrayOutputStream()
+    val dos = new java.io.DataOutputStream(baos)
+    dos.writeInt(dataLength)
+    dos.write(gzipData)
+    dos.write("afterGzip".getBytes(StandardCharsets.UTF_8))
+    dos.close()
+    val data = baos.toByteArray()
+    (data, dataLength)
+  }
+
+  @Test def testGZIPLayer1() {
+    val sch = GZIPLayer1Schema
+    val (data, dataLength) = makeGZIPLayer1Data()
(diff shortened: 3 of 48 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/Base64Transformer.scala
```diff
@@ -0,0 +1,43 @@
...
+  extends LayerTransformer() {
+
+  override def wrapLayerDecoder(jis: java.io.InputStream) = {
+    val b64 = java.util.Base64.getMimeDecoder().wrap(jis)
+    b64
+  }
+
+}
+
+object Base64MIMETransformerFactory
```
mbeckerle: Add scaladoc: newInstance method digests the various layering related arguments to decide how to configure and which layer transformer to create, then constructs it appropriately.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIOStreams.scala
```diff
@@ -0,0 +1,281 @@
...
+ * This can be used to forcibly stop consumption of data from a stream at
+ * a length obtained from a delimiter.
+ *
+ * The terminator string is consumed from the underlying stream (if found), and
+ * the underlying stream is left positioned at the byte after the terminator
+ * string.
+ *
+ * Thread safety: This is inherently stateful - so not thread safe to use
+ * this object from more than one thread.
+ */
```
mbeckerle: Ok, so broken.... nothing is calling/constructing one of these, so the layer decoder will go on forever.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * Shared functionality of all LayerTransformers.
+ */
+abstract class LayerTransformer()
+  extends Serializable {
+
+  protected def wrapLayerDecoder(jis: java.io.InputStream): java.io.InputStream
+
+  def wrapJavaInputStream(s: DataInputStream, fInfo: FormatInfo): java.io.InputStream = {
+    new JavaIOInputStream(s, fInfo)
```
mbeckerle: We need to additionally wrap an Explicit or Delimited Limiting Stream around this JavaIOInputStream. Otherwise the decoding will not know when to stop.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -0,0 +1,272 @@
...
+  def removeLayer(s: DataInputStream): Unit = {
+    // nothing for now
+  }
+
+  def addLayer(s: DataOutputStream, fInfo: FormatInfo): DataOutputStream = ???
+
+  def removeLayer(s: DataOutputStream): Unit = ???
+
+}
+
```
mbeckerle: Scaladoc: This class creates a java.io.InputStream from a daffodil DataInputStream.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -0,0 +1,272 @@
...
+    if (n != -1)
+      readBuf(0)
+    else
+      n
+  }
+
+  override def read(b: Array[Byte]) =
+    read(b, 0, b.length)
+
+  /**
```
mbeckerle: Revise this javadoc. This was copied here for reference when writing the code, but really this is scala code and this needs to be scaladoc, or just omitted since this is a well-documented interface already.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/LayeredSequenceUnparser.scala
```diff
@@ -0,0 +1,56 @@
...
+    val layerTransformer = layerTransformerEv.evaluate(state)
+
+    val savedDOS = state.dataOutputStream
+
+    val isAligned = savedDOS.align(layerTransformer.mandatoryLayerAlignmentInBits, state)
+    if (!isAligned)
+      UE(state, "Unable to align to the mandatory layer alignment of %s(bits)",
+        layerTransformer.mandatoryLayerAlignmentInBits)
+
+    val newDOS = layerTransformer.addLayer(savedDOS, state)
```
stevedlawrence: Thoughts on reversing the action and changing it to something like this:

    val newDos = savedDOS.addLayer(layerTransformer, state)

This is kindof how we handle buffered data streams on unparse, e.g.

    val newDOS = currentDos.addBuffered

Note sure what this affects in the big picture, if anything. Maybe it just affects which class the code is in and doesn't really matter.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -0,0 +1,272 @@
...
+   * All transformers must be registered so they are available by name.
+   *
+   * It is possible to package a transformer in a separate jar also, but then
+   * someone has to register it by calling the register method.
+   *
+   * Transformers built into the primary Daffodil jars/packages should be
+   * registered here.
+   */
```
stevedlawrence: We might want to use the java Services API for this kindof thing. Would prevent users from having to call the register function--just need to put a special file in the META-INF directory in the jar file.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
stevedlawrence: By saying "this is scala code", is this copied from the scala library. If so, we need to make sure it's the LICENSE file is updated, and ideally this would be in a scalaLib directory like some other scala stuff.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIOStreams.scala
```diff
@@ -0,0 +1,281 @@
...
+class DelimitedLengthLimitingStream(inWithMarkSupported: InputStream, terminator: String,
+  charset: Charset = StandardCharsets.ISO_8859_1,
+  maxChunkSize: Int = 1024 * 32) // 32K chunk size by default. Units are chars.
+  extends FilterInputStream(inWithMarkSupported) {
```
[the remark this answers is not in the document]
mbeckerle: I think each layer DataInputStream, or Java.io.InputStream has to support mark/reset. When we wrap a JavaIOInputStream around a DataInputStream, we implement the mark/reset in terms of what DataInputStream supports. 

The need for mark/reset here is for a very knarly painful thing about the way I'm using Java Scanner class to search for a regex. Because this operats on a reader, and both it and the reader layer read ahead a bit, we have to play tricks to get the underlying input stream to be precisely positioned immediately after the last byte that was "required" to decode the layer. 

More in next commit.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIOStreams.scala
```diff
@@ -0,0 +1,281 @@
...
+   *
+   * After we match this regex with the scanner,
+   * the input stream should end up positioned immediately before the terminator.
+   * The terminator is not part of the match, because it is in a non-capturing
+   * lookahead group.
+   *
+   * However, for assertion checking, a second capture group is defined
+   * inside the lookahead, and that should capture the terminator.
+   */
+  private val quotedTerminator = Pattern.quote(terminator) // in case pattern has non-regex-safe characters in it
```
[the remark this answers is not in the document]
mbeckerle: No specific reason. I was hoping that the scan for terminator string of a base64 region would be very lightweight and fast. E.g., I'm thinking layerTerminator is just a plain string, no WSP or other class entities, etc. That generality may not be needed, and the test burden that generality creates is not insubstantial. 

We'll revisit this point.

daffodil-io/src/test/scala/org/apache/daffodil/layers/TestBase64.scala
```diff
@@ -0,0 +1,242 @@
...
+ * When decoding from Base64, the CRLF vs. LF doesn't matter. However when
+ * encoding, the RFC specify that CRLF are inserted.
+ */
+class TestBase64 {
+  Assert.usage(scala.util.Properties.isJavaAtLeast("1.8"))
+
+  val text = """Man is distinguished, not only by his reason, but by this singular passion from
+other animals, which is a lust of the mind, that by a perseverance of delight
+in the continued and indefatigable generation of knowledge, exceeds the short
+vehemence of any carnal pleasure.""".replace("\n", " ")
```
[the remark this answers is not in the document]
mbeckerle: OK, I'll change it to Ipsum lorem or some other gibberish of my own making.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
mbeckerle: I'd like to be able to add transformers on the fly. Not high prio for now, but I am pretty sure it will come up in the future.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -245,6 +248,31 @@ abstract class SequenceTermBase(
       maybeCheckBitOrderAndCharset)
   }
 
+  private val layeredSequenceAllowedProps = Set("ref", "layerTransform", "layerEncoding", "layerLengthKind", "layerLength", "layerLengthUnits", "layerTerminator")
+
+  final lazy val maybeLayerTransformerEv: Maybe[LayerTransformerEv] = {
+    if (maybeLayerTransformEv.isEmpty) Maybe.Nope
```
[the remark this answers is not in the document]
mbeckerle: You can't SDE because they could be in scope for use by layers elsewhere in the schema. 
If they are expressed directly on this specific layered sequence, then we can SDW about them.

daffodil-io/src/main/scala/org/apache/daffodil/io/ByteBufferDataInputStream.scala
```diff
@@ -91,7 +90,13 @@ object ByteBufferDataInputStream {
       case _ => {
         // copy the contents of the stream into an array of bytes
         val bos = new ByteArrayOutputStream
-        IOUtils.copy(in, bos)
```
mbeckerle: BTW I think this change was because i needed to breakpoint on the read or write when debugging.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaOutputStreams.scala
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.daffodil.io
+
+import java.nio.charset.Charset
+
```
mbeckerle: How about the term "Boundary" instead of "terminator" to avoid confusion with the existing DFDL concept?
So a layer can have a boundary mark, not a terminator.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/AIS_PAYLOAD_ARMORING.scala
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * http://catb.org/gpsd/AIVDM.html#_aivdm_aivdo_payload_armoring
+ *
+ * To convert a string of length N, You will get 6N bits.
+ *
+ * The decoder can be used for unit testing, but the point of this class
+ * is to make the encoder available for use in un-doing the AIS Payload
+ * armoring.
+ *
+ * Always use with encodingErrorPolicy="error" since supplying a character
```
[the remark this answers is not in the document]
mbeckerle: I'll improve the comment. We don't have the EEP='error' option, so wherever this AIS encoder gets called we'll need to do the checking ourselves.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/USASCII6BitPackedDecoder.scala
```diff
@@ -26,15 +26,26 @@ import org.apache.daffodil.schema.annotation.props.gen.BitOrder
...
   // def decodeString = ((0 to 63).map { charCode => (if (charCode <= 31) charCode + 64 else charCode).toChar }).mkString
 }
 
+object USASCII6BitPackedLSBFirstCharset
+  extends USASCII6BitPacked(BitOrder.LeastSignificantBitFirst,
+    "X-DFDL-US-ASCII-6-BIT-PACKED-LSB-FIRST")
+
+object USASCII6BitPackedCharset
+  extends USASCII6BitPacked(BitOrder.LeastSignificantBitFirst,
+    "X-DFDL-US-ASCII-6-BIT-PACKED") // just an alias for the the "LSB-FIRST variant.
```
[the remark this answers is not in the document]
mbeckerle: See https://issues.apache.org/jira/browse/DAFFODIL-1926

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
mbeckerle: added https://issues.apache.org/jira/browse/DAFFODIL-1927 to record this feature request.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaOutputStreams.scala
```diff
@@ -0,0 +1,38 @@
...
+class LayerTerminatorInsertingJavaOutputStream(jos: java.io.OutputStream, terminator: String,
+  charset: Charset)
+  extends java.io.FilterOutputStream(jos) {
+
+  private var closed = false
+
+  private val terminatorBytes = terminator.getBytes(charset)
```
stevedlawrence: Rename to boundaryMark for consistency?

daffodil-io/src/test/scala/org/apache/daffodil/layers/TestBase64.scala
```diff
@@ -36,16 +36,25 @@ import org.apache.daffodil.exceptions.Assert
...
-vehemence of any carnal pleasure.""".replace("\n", " ")
-
-  val b64Text = """TWFuIGlzIGRpc3Rpbmd1aXNoZWQsIG5vdCBvbmx5IGJ5IGhpcyByZWFzb24sIGJ1dCBieSB0aGlz
-IHNpbmd1bGFyIHBhc3Npb24gZnJvbSBvdGhlciBhbmltYWxzLCB3aGljaCBpcyBhIGx1c3Qgb2Yg
-dGhlIG1pbmQsIHRoYXQgYnkgYSBwZXJzZXZlcmFuY2Ugb2YgZGVsaWdodCBpbiB0aGUgY29udGlu
-dWVkIGFuZCBpbmRlZmF0aWdhYmxlIGdlbmVyYXRpb24gb2Yga25vd2xlZGdlLCBleGNlZWRzIHRo
-ZSBzaG9ydCB2ZWhlbWVuY2Ugb2YgYW55IGNhcm5hbCBwbGVhc3VyZS4=""".replace("\n", "\r\n")
+  //
+  // This is from Don Quixote which is entirely public domain
+  //
```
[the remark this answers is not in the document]
mbeckerle: Ok, well to simplify life I'll just change this text to the text of the Apache License....!

My only goal for this text is that it is readable, so that when debugging you can easily spot if it has been clobbered.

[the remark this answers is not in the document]
stevedlawrence: I like it! That's definitely ASLv2 compatible.

daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
```diff
@@ -38,7 +38,7 @@ class TestLayers {
       <dfdl:format ref="tns:GeneralFormat"/>,
       <xs:element name="e1" dfdl:lengthKind="implicit">
         <xs:complexType>
-          <xs:sequence dfdl:layerTransform="base64_MIME" dfdl:layerLengthKind="delimited" dfdl:layerTerminator="!" dfdl:layerEncoding="iso-8859-1">
+          <xs:sequence dfdl:layerTransform="base64_MIME" dfdl:layerLengthKind="delimited" dfdl:layerBoundaryMark="!" dfdl:layerEncoding="iso-8859-1">
```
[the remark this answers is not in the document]
mbeckerle: Yes, I removed the discussion of parent delimiting markup having impact. 

I like changing the layerLengthKind to boundaryMark also. Will make that change.

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIinputStreams.scala
```diff
@@ -0,0 +1,287 @@
...
+ * The regex must have a finite maximum length match string.
+ *
+ * This can be used to forcibly stop consumption of data from a stream at
+ * a length obtained from a delimiter that is described using a regex.
+ *
+ * The delimiter matching the regex is consumed from the underlying stream (if found), and
+ * the underlying stream is left positioned at the byte after the regex match
+ * string.
+ *
+ * IMPORTANT: The delimiter regex cannot contain any Capturing Groups!
```
VTGuy08: Interesting comment.  Is there any code checking for this case to prevent misuse?

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIinputStreams.scala
```diff
@@ -0,0 +1,287 @@
...
+  /**
+   * This lazy stream stuff might look like a lot of overhead, but
+   * consider that the overhead is once per chunk, so honestly the
+   * regex match is of more concern.
+   */
+  private def chunks: Stream[String] = {
+    if (noMoreChunks) Stream()
+    else {
+      in.mark(lookAheadMax)
+      //
```
VTGuy08: I'm really appreciating all of these comments 🥇

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/ScalaXMLInfosetInputter.scala
```diff
@@ -81,6 +82,21 @@ class ScalaXMLInfosetInputter(rootNode: Node)
         }
         val text = child match {
           case t: Text => t.data
+          //
+          // Note: may be a bug in scala.xml library, but if you do this:
```
VTGuy08: Submit this to the scala peeps?

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIinputStreams.scala
mbeckerle: Hi Taylor, no nothing checks this as yet. It is possible to check, as regex are checkable that way, but it's non-trivial.  I'm thinking a comment should be sufficient. It's not like this string gets passed in from a user schema.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/ScalaXMLInfosetInputter.scala
mbeckerle: I have to revise this comment. The comment code here doesn't even work. I did have to add Atom[String] case, and I'm not sure why that should ever occur, but the example for reproducing it here doesn't work. There's at least one other place in XMLUtils where comments on XML wierdness about Text vs. PCData appear, and when I re-run those examples, they don't have the problems described in the comments anymore. I''ll fix the comment here, and update that other place to point out that the problem no longer occurs. Not sure what else I can do about this other than revisit all the use of XML parsers in Daffodil?

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LineFoldedTransformer.scala
```diff
@@ -0,0 +1,474 @@
...
+            case -1 => {
+              state = Done
+              return '\r'
+            }
+            case '\n' => {
+              state = GotCRLF
+            }
+            case _ => {
+              state = Buf1
+              return c
```
stevedlawrence: I think this should be ``return '\r'``? This state signifies that we got a CR followed by not-CR. So this should return that original CR, then transition to the Buf1 state so that the non-CR that we just read is returned next. I think right-now it would return the not-CR twice.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/LayeredSequenceParser.scala
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.daffodil.processors.parsers
+
+import org.apache.daffodil.processors.TermRuntimeData
+import org.apache.daffodil.processors.LayerTransformerEv
+import org.apache.daffodil.io.ByteBufferDataInputStream
+
+class LayeredSequenceParser(rd: TermRuntimeData,
+  layerTransformerEv: LayerTransformerEv,
+  bodyParser: Parser)
+  extends SequenceCombinatorParser(rd, bodyParser) {
```
stevedlawrence: Ah, so maybe this solves the problem I mentioned elsewhere about duplicating sequence combinator logic.

Interestingly though, does this mean that the sequence combinator will actually used properties like dfdl:separatorSuppresstionPolicy from the layered sequence (assuming is was provided in the default format)?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/LayeredSequence.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.daffodil.processors.parsers.{ Parser => DaffodilParser }
+import org.apache.daffodil.processors.unparsers.{ Unparser => DaffodilUnparser }
+import org.apache.daffodil.grammar.Gram
+import org.apache.daffodil.grammar.EmptyGram
+import org.apache.daffodil.util.Misc
+import org.apache.daffodil.processors.parsers.LayeredSequenceParser
+import org.apache.daffodil.processors.unparsers.LayeredSequenceUnparser
+
+case class LayeredSequence(sq: SequenceTermBase, rawTerms: Seq[Gram])
+  extends Terminal(sq, true) {
```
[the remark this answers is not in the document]
mbeckerle: The nest you give of LayeredSequence(this, SequenceCombinator(this, terms)) actually has different meaning. For the same reason that it is undesirable to put layering properties on the same sequence that is also carrying other properties. It becomes ambiguous which happens first: 
E.g., in 
{code}
<sequence dfdl:layerTransform="base64_MIME" dfdl:initiator="foo:"... >
{code}
Do we decode base64 first, then look for the initiator, or the other way round? By requiring a separate sequence things are 100% unambiguous, so I'd rather keep things this way for now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/LayeredSequenceParser.scala
mbeckerle: Interesting point. I should at least document whether LayeredSequenceParser overrides everything about SequenceCombinatorParser that uses the other non-layering properties, but in that case, why derive from it. I'll give this some thought and at minimum add doc.

mbeckerle: Turns out there is no good reason to inherit from SequenceCombinatorParser, so I'm eliminating that here and for the LayeredSequenceUnparser similarly.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/LayeredSequence.scala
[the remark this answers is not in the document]
mbeckerle: Further thought, and I'll put this in scaladoc, which is that while LayeredSequence looks same as sequence now, once SequenceCombinator evolves (as expected), they will differ much more substantially. LayeredSequence will still take only a single body parser, not a vector thereof.

I will change grammar and prim to take single body grammar object, not vector thereof, which will clarify a few things. The grammar will also not be wrapping things that lay down separators around, only to (hopefully) have them optimize out - this is what might not have been working due to inherited non-layer properties on the sequence. It will also prevent an array from being the direct child of a layer, since an array may need separators, etc.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -245,6 +248,31 @@ abstract class SequenceTermBase(
...
+  private val layeredSequenceAllowedProps = Set("ref", "layerTransform", "layerEncoding", "layerLengthKind", "layerLength", "layerLengthUnits", "layerBoundaryMark")
+
+  final lazy val maybeLayerTransformerEv: Maybe[LayerTransformerEv] = {
+    if (maybeLayerTransformEv.isEmpty) Maybe.Nope
+    else { // need to check that only layering properties are specified
+      val localProps = this.formatAnnotation.justThisOneProperties
+      val localKeys = localProps.keySet
+      val disallowedKeys = localKeys.filterNot(k => layeredSequenceAllowedProps.contains(k))
+      if (disallowedKeys.size > 0)
+        SDE("Sequence has dfdl:layerTransform specified, so cannot have non-layering properties: %s", disallowedKeys)
```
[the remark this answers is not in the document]
mbeckerle: Fixed, and test_layersErr1 added.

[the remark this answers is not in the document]
mbeckerle: Also, yes, other namespace attributes will be ignored. Tests test_layer2 and test_layersErr1 verify this.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -96,11 +96,26 @@ final class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectO
...
    * are morphed into direct streams.
    */
-  val id: Int = if (splitFrom == null) 0 else splitFrom.id + 1
+  private val splitID: Int = if (splitFrom == null) 0 else splitFrom.splitID + 1
+
+  /**
+   * id will be a N.M type identifier where N is the layer number,
+   * and M is the splitID.
+   */
+  val id: String = layerID.toString + "." + splitID.toString
```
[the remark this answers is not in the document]
mbeckerle: done

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/LayeredSequenceParser.scala
mbeckerle: So both the parser and unparser both call super.parse(state) or super.unparse(state) so as to reuse the guts of the parser.

Once the parser and unparser for Sequence are revised though, this may no longer make sense.  There is a lot of code reuse for unparser especially here. No matter what happens when Sequence is reimplemented this code reuse will need to be preserved. 

The answer to your question is no however, the properties like dfdl:separatorSuppressionPolicy will not apply, as the body parser for these cannot have an array (now enforced), and can be only one term, and the grammar doesn't put down separator grammar terms around it.  So it will not have any "sequence-like" behavior associated with separators. (Unless of course the single body itself IS a sequence, but then the properties are its own, not those of the layered sequence.)

PR #64, opened 2018-04-20
Added tests to daffodil-test scala-debug for JIRA ticket DAFFODIL-1923.

dfthompson:
daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScheme.tdml
  daffodil-test/src/test/scala-debug/org/apache/daffodil/section07/escapeScheme/TestEscapeSchemeDebug.scala

Added test to daffodil-tdml scala-debug for JIRA ticket DAFFODIL-1928.
Consolidated TestTDMLRunner2.scala and TestTDMLRunnerNew.scala,
duplicate tests.

  daffodil-tdml/src/test/resources/test/tdml/tdmlQuoting.tdml
  daffodil-tdml/src/test/scala-debug/org/apache/daffodil/tdml/TestTDMLRunner3.scala
  daffodil-tdml/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunner2.scala
  daffodil-tdml/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunnerNew.scala

DAFFODIL-1923 DAFFODIL-1928


daffodil-tdml/src/test/resources/test/tdml/tdmlApostrophe.tdml
```diff
@@ -0,0 +1,68 @@
...
+    	  <tns:item>Column &apos;Number&apos; Two</tns:item>
+    	  <tns:item>3</tns:item>
+    	  <tns:item>4</tns:item>
+    	  <tns:item>5</tns:item>
+  	</tns:record>
+      </dfdlInfoset>
+    </infoset>
+  </parserTestCase>
+
+</testSuite>
```
stevedlawrence: This test is a little more complicated than it needs to be for what it is testing. Ideally, tests are as small as possible so that there are no unintended side effects. This test has escape schemes, delimited text, etc. Lots of things that don't really matter for &apos; to be tested. The test for this should probably be virtually the same as the test in tdmlQuoting, just with "apos" instead of "quot". It would also be reasonable to have the test in tdmlQuoting.tdml instead of creating a new file.

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScheme.tdml
```diff
@@ -520,5 +520,128 @@
...
+        <xs:sequence dfdl:separator="," >
+          <xs:element name="item" type="xs:string" maxOccurs="unbounded"
+	    dfdl:escapeSchemeRef="tns:eBlkAllQuotes" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+  </defineSchema>
+
+  <parserTestCase name="escBlkAllQuotes" model="eBlkAllQuotes"
+    description="Section 7 defineEscapeScheme - DFDL-7-079R" root="record" roundTrip="false">
```
stevedlawrence: This test should round trip when it works. Best to set the roundTrip="true".

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScheme.tdml
```diff
@@ -520,5 +520,128 @@
...
+        <xs:sequence dfdl:separator="," >
+          <xs:element name="item" type="xs:string" maxOccurs="unbounded"
+	    dfdl:escapeSchemeRef="tns:eBlkEndParen" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+  </defineSchema>
+ 
+  <parserTestCase name="escBlkEndSame" model="eBlkEndSame"
+    description="Section 7 defineEscapeScheme - DFDL-7-079R" root="record" roundTrip="false">
```
stevedlawrence: Same here, this should also round trip.

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScheme.tdml
```diff
@@ -520,5 +520,128 @@
...
+      <xs:complexType>
+        <xs:sequence dfdl:separator="," >
+          <xs:element name="item" type="xs:string" maxOccurs="unbounded"
+	    dfdl:escapeSchemeRef="tns:eBlkAposRemovd" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+  </defineSchema>
+ 
+  <parserTestCase name="escBlkAposRemoved" model="eBlkAposRemoved"
```
stevedlawrence: This test looks like it is the same as tdmlApostrophe.tdml. Make sure to remove the duplicate.

PR #65, opened 2018-05-03
Added AIS layer with tests. Added tests matching to layer wiki page.

mbeckerle:
This is a "baby" implementation of AIS. It shows that the layering transformation can be done, uses it in one example.  The AIS format layering transformation does/undoes AIS's ascii-armoring. This is provided, usable via dfdl:layerTransformation="ais".

Implementing AIS completely beyond this is a big job. Many messages, much testing needed.

What is here shows that this project could now be undertaken successfully exploiting Daffodil's layering capability. 

Actually, a 6-bit AIS-specific text charset encoding may also be needed. This is not to be confused with the AIS ascii-armoring - which uses Daffodil's text encoder/decoder for 6-bit to Ascii, but has a slightly different mapping. But new charset encodings are 3 lines of code so I'm not going to bother putting that in for now. 

DAFFODIL-1805

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.layers
+
+import org.apache.daffodil.schema.annotation.props.gen.LayerLengthKind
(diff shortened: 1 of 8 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.daffodil.processors.charset.AIS_PAYLOAD_ARMORING
+import java.nio._
+import java.nio.charset._
+import java.io._
+import org.apache.commons.io.IOUtils
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.io.BoundaryMarkLimitingStream
+import org.apache.daffodil.io.LayerBoundaryMarkInsertingJavaOutputStream
+
+object AISTransformer {
```
mbeckerle: Make naming more specific: AIS to AIS_ASCII_ARMORING or similar. The name of the layer transform should probably also be not "ais" but "aisASCIIArmoring"

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
```diff
@@ -0,0 +1,141 @@
...
+  extends InputStream {
+  import AISPayloadArmoringTransformer._
+
+  private lazy val enc = AIS_PAYLOAD_ARMORING.newEncoder()
+
+  private lazy val bais = {
+    val armoredText = IOUtils.toString(jis, iso8859)
+    val cb = CharBuffer.wrap(armoredText)
+
+    val numBytes = (6 * armoredText.length) / 8
```
jadams-tresys: Just curious, but why 6 * armoredText.length? Is it due to the way the armoredText is encoded?

mbeckerle: The armoring scheme encodes 6 bits as an 8 bit character. So when decoding, 8 bits becomes 6, so string length * 6 bits then div by 8 for bytes.

PR #67, opened 2018-05-04
Add test showing inability to read data after a layer

stevedlawrence:
DAFFODIL-1934

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
@@ -253,5 +253,50 @@ xxx]]></tdml:documentPart>
       <tdml:error>separator</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
+  
+  <tdml:defineSchema name="s4" elementFormDefault="unqualified">
+    <dfdl:defineFormat name="general">
+      <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited" outputNewLine="%CR;%LF;" />
+    </dfdl:defineFormat>
+    <dfdl:defineFormat name="folded">
+      <dfdl:format ref="ex:general" layerTransform="lineFolded_IMF" layerLengthKind="implicit" layerLengthUnits="bytes"
+        layerEncoding="iso-8859-1" />
+    </dfdl:defineFormat>
+    <dfdl:format ref="ex:general" />
+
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
```diff
@@ -253,5 +253,50 @@ xxx]]></tdml:documentPart>
       <tdml:error>separator</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
+  
+  <tdml:defineSchema name="s4" elementFormDefault="unqualified">
+    <dfdl:defineFormat name="general">
+      <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited" outputNewLine="%CR;%LF;" />
+    </dfdl:defineFormat>
+    <dfdl:defineFormat name="folded">
+      <dfdl:format ref="ex:general" layerTransform="lineFolded_IMF" layerLengthKind="implicit" layerLengthUnits="bytes"
```
mbeckerle: Here's the bug/issue.  We haven't got a crisp description of what these lineFolded transforms actually do.

lineFolded_IMF with layerlengthKind 'implicit' means - there is no terminating this. Just go as far as you can. 

It requires that either the entire data stream is being folded, or there is a surrounding "specified length" element to stop it.

I think you want layerLengthKind 'boundaryMark'. In the lineFolded case, the boundaryMark is not variable. It is always CRLF not followed by tab/space. I.e., a single "line" of text that if too long, must be folded.

Those two behaviors, 'implicit' and 'boundaryMark' are what the code does today, but I'm not sure they are enough.

daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
```diff
@@ -253,5 +253,50 @@ xxx]]></tdml:documentPart>
...
+              </xs:complexType>
+            </xs:element>
+          </xs:sequence>
+          <xs:element name="body" type="xs:string" dfdl:initiator="" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="layers4" root="root" model="s4" roundTrip="true">
```
mbeckerle: Another PR has a layers4 test already. Can you rename this layers5 to avoid the merge conflict?

PR #68, opened 2018-05-07
Fix so trace works even when layering is being used.

mbeckerle:
DAFFODIL-1935

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
@@ -304,7 +304,6 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
       // turn debugging off
       this.cst.debugging = false
       debugOutputStream = Nope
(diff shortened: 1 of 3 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -816,7 +816,7 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
   }
 
   final override def pastData(nBytesRequested: Int): ByteBuffer = {
-    Assert.usage(isReadable)
+    Assert.usage(isReadable || areDebugging)
     if (!areDebugging)
       throw new IllegalStateException("Must be debugging.")
```
stevedlawrence: It's now impossible to hit this IllegalStateException. Note that the parse past/futureData functions do not have an assert but just have the if (!artDebugging) check. Should these all behave similarly, or are there reasons for the differences?

mbeckerle:
Ah, Travis found some regressions I missed. So this will be revised.

PR #70, opened 2018-05-22
Adding property binaryCalendarRep with values "binarySeconds" and "bi…

efinnegan:
…naryMilliseconds"

DAFFODIL-99

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -48,6 +48,7 @@ import org.apache.daffodil.grammar.primitives.CaptureContentLengthStart
 import org.apache.daffodil.grammar.primitives.CaptureValueLengthEnd
 import org.apache.daffodil.grammar.primitives.CaptureValueLengthStart
 import org.apache.daffodil.grammar.primitives.ComplexNilOrContent
+import org.apache.daffodil.grammar.primitives.ConvertBinaryDateTimeSecMilliPrim
 import org.apache.daffodil.grammar.primitives.ConvertTextBooleanPrim
 import org.apache.daffodil.grammar.primitives.ConvertTextBytePrim
 import org.apache.daffodil.grammar.primitives.ConvertTextCombinator
@@ -115,6 +116,7 @@ import org.apache.daffodil.grammar.primitives.ZonedTextIntPrim
 import org.apache.daffodil.processors.TextJustificationType
 import org.apache.daffodil.schema.annotation.props.Found
 import org.apache.daffodil.schema.annotation.props.NotFound
+import org.apache.daffodil.schema.annotation.props.gen.BinaryCalendarRep
 import org.apache.daffodil.schema.annotation.props.gen.BinaryFloatRep
 import org.apache.daffodil.schema.annotation.props.gen.BinaryNumberRep
 import org.apache.daffodil.schema.annotation.props.gen.LengthKind
@@ -462,6 +464,11 @@ trait ElementBaseGrammarMixin
     case PrimType.Short | PrimType.UnsignedShort => 16
     case PrimType.Float | PrimType.Int | PrimType.UnsignedInt | PrimType.Boolean => 32
     case PrimType.Double | PrimType.Long | PrimType.UnsignedLong => 64
+    case PrimType.DateTime => binaryCalendarRep match {
+      case BinaryCalendarRep.BinarySeconds => 32
+      case BinaryCalendarRep.BinaryMilliseconds => 64
+      case _ => schemaDefinitionError("Size of binary data '" + primType.name + "' with binaryCalendarRep='" + binaryCalendarRep + "' cannot be determined implicitly.")
+    }
     case _ => schemaDefinitionError("Size of binary data '" + primType.name + "' cannot be determined implicitly.")
   }
 
@@ -775,6 +782,25 @@ trait ElementBaseGrammarMixin
       }
 
       case PrimType.Boolean => { new BinaryBoolean(this) }
+
+      case PrimType.DateTime | PrimType.Date | PrimType.Time => {
+        (primType, binaryCalendarRep) match {
+          case (PrimType.DateTime, BinaryCalendarRep.BinarySeconds) => (lengthUnits, binaryNumberKnownLengthInBits) match {
+            case (LengthUnits.Bytes, 32) => new ConvertBinaryDateTimeSecMilliPrim(this, binaryNumberKnownLengthInBits)
+            case (_, 32)  => SDE("lengthUnits must be 'bytes' when binaryCalendarRep='binarySeconds'")
+            case (_, n) => SDE("binary xs:dateTime must be 32 bits when binaryCalendarRep='binarySeconds'. Length in bits was %s.", n)
+          }
+          case (_, BinaryCalendarRep.BinarySeconds) => SDE("binaryCalendarRep='binarySeconds' is not allowed with type %s", primType.name)
+          case (PrimType.DateTime, BinaryCalendarRep.BinaryMilliseconds) => (lengthUnits, binaryNumberKnownLengthInBits) match {
+            case (LengthUnits.Bytes, 64) => new ConvertBinaryDateTimeSecMilliPrim(this, binaryNumberKnownLengthInBits)
+            case (_, 64)  => SDE("lengthUnits must be 'bytes' when binaryCalendarRep='binaryMilliseconds'")
+            case (_, n) => SDE("binary xs:dateTime must be 64 bits when binaryCalendarRep='binaryMilliseconds'. Length in bits was %s.", n)
+          }
+          case (_, BinaryCalendarRep.BinaryMilliseconds) => SDE("binaryCalendarRep='binaryMilliseconds' is not allowed with type %s", primType.name)
+          case _ => notYetImplemented("Type %s when representation='binary' and binaryCalendarRep=%s", primType.name, binaryCalendarRep.toString)
+        }
+      }
+
       case _ => notYetImplemented("Type %s when representation='binary'", primType.name)
     }
     res
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
@@ -17,29 +17,38 @@
 
 package org.apache.daffodil.grammar.primitives
 
+import java.text.ParsePosition
+import com.ibm.icu.text.SimpleDateFormat
 import com.ibm.icu.util.Calendar
 import com.ibm.icu.util.TimeZone
+import com.ibm.icu.util.ULocale
 import org.apache.daffodil.dsom.ElementBase
 import org.apache.daffodil.grammar.Terminal
 import org.apache.daffodil.schema.annotation.props.gen.CalendarCheckPolicy
 import org.apache.daffodil.schema.annotation.props.gen.CalendarFirstDayOfWeek
 import org.apache.daffodil.schema.annotation.props.gen.CalendarPatternKind
+import org.apache.daffodil.processors.unparsers.ConvertBinaryCalendarSecMilliUnparser
 import org.apache.daffodil.processors.unparsers.ConvertTextCalendarUnparser
 import com.ibm.icu.util.Calendar
 import com.ibm.icu.util.TimeZone
 import org.apache.daffodil.processors.CalendarEv
 import org.apache.daffodil.processors.CalendarLanguageEv
+import org.apache.daffodil.processors.parsers.ConvertBinaryCalendarSecMilliParser
 import org.apache.daffodil.processors.parsers.ConvertTextCalendarParser
 import org.apache.daffodil.processors.parsers.TextCalendarConstants
 import scala.Boolean
 
-abstract class ConvertTextCalendarPrimBase(e: ElementBase, guard: Boolean)
+abstract class ConvertCalendarPrimBase(e: ElementBase, guard: Boolean)
   extends Terminal(e, guard) {
 
   protected val xsdType = "dateTime"
   protected val prettyType = "DateTime"
 
   override def toString = "to(xs:" + xsdType + ")"
+}
+
+abstract class ConvertTextCalendarPrimBase(e: ElementBase, guard: Boolean)
+  extends ConvertCalendarPrimBase(e, guard) {
 
   protected def infosetPattern: String
   protected def implicitPattern: String
@@ -144,28 +153,13 @@ abstract class ConvertTextCalendarPrimBase(e: ElementBase, guard: Boolean)
     pattern,
     hasTZ,
     localeEv,
-    calendarEv,
-    infosetPattern,
-    firstDay,
-    calendarDaysInFirstWeek,
-    calendarCheckPolicy,
-    calendarTz,
-    TimeZone.GMT_ZONE)
+    calendarEv)
 
   override lazy val unparser = new ConvertTextCalendarUnparser(
     e.elementRuntimeData,
-    xsdType,
-    prettyType,
     pattern,
-    hasTZ,
     localeEv,
-    calendarEv,
-    infosetPattern,
-    firstDay,
-    calendarDaysInFirstWeek,
-    calendarCheckPolicy,
-    calendarTz,
-    TimeZone.GMT_ZONE)
+    calendarEv)
 }
 
 case class ConvertTextDatePrim(e: ElementBase) extends ConvertTextCalendarPrimBase(e, true) {
@@ -190,4 +184,62 @@ case class ConvertTextDateTimePrim(e: ElementBase) extends ConvertTextCalendarPr
   protected override val infosetPattern = "uuuu-MM-dd'T'HH:mm:ss.SSSSSSxxx"
   protected override val implicitPattern = "uuuu-MM-dd'T'HH:mm:ss"
   protected override val validFormatCharacters = "adDeEFGhHkKmMsSuwWvVyXxYzZ".toSeq
+}
+
+abstract class ConvertBinaryCalendarPrimBase(e: ElementBase, guard: Boolean, lengthInBits: Long)
+  extends ConvertCalendarPrimBase(e, guard) {
+
+}
+
+case class ConvertBinaryDateTimeSecMilliPrim(e: ElementBase, lengthInBits: Long) extends ConvertCalendarPrimBase(e, true) {
+  protected override val xsdType = "dateTime"
+  protected override val prettyType = "DateTime"
+
+  lazy val epochCalendar: Calendar = {
+    val cal = Calendar.getInstance
+    cal.clear()
+    cal.setLenient(false)
+
+    val sdfWithTZ = new SimpleDateFormat("uuuu-MM-dd'T'HH:mm:ssZZZZ", ULocale.ENGLISH)
+    var pos = new ParsePosition(0)
+    sdfWithTZ.parse(e.binaryCalendarEpoch, cal, pos)
+
+    if (pos.getIndex != e.binaryCalendarEpoch.length || pos.getErrorIndex >= 0) {
+      // binaryCalendarEpoch didn't match the first format with timezone, so try without
+      cal.clear()
+      pos = new ParsePosition(0)
+      val sdf = new SimpleDateFormat("uuuu-MM-dd'T'HH:mm:ss", ULocale.ENGLISH)
+      cal.setTimeZone(TimeZone.UNKNOWN_ZONE)
+      sdf.parse(e.binaryCalendarEpoch, cal, pos)
+
+      if (pos.getIndex != e.binaryCalendarEpoch.length || pos.getErrorIndex >= 0) {
+        SDE("Failed to parse binaryCalendarEpoch - Format must match the pattern 'uuuu-MM-dd'T'HH:mm:ss' or 'uuuu-MM-dd'T'HH:mm:ssZZZZ'")
+      }
+    }
+
+    try {
+      cal.getTime
+    } catch {
+      case e: IllegalArgumentException => {
+        SDE("Failed to parse binaryCalendarEpoch: %s.", e.getMessage())
+      }
+    }
+
+    cal
+  }
+
+  override lazy val parser = new ConvertBinaryCalendarSecMilliParser(
+    e.elementRuntimeData,
+    !epochCalendar.getTimeZone.equals(TimeZone.UNKNOWN_ZONE),
+    e.binaryCalendarRep,
+    epochCalendar,
+    lengthInBits.toInt)
+
+  override lazy val unparser =
+    new ConvertBinaryCalendarSecMilliUnparser(
+    e.elementRuntimeData,
+    e.binaryCalendarRep,
+    epochCalendar.getTimeInMillis,
+    lengthInBits.toInt,
+    !epochCalendar.getTimeZone.equals(TimeZone.UNKNOWN_ZONE))
 }
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertBinaryCalendarUnparser.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertBinaryCalendarUnparser.scala
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processors.unparsers
+
+import org.apache.daffodil.calendar.DFDLCalendar
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.io.DataOutputStream
+import org.apache.daffodil.io.FormatInfo
+import org.apache.daffodil.processors.ElementRuntimeData
+import org.apache.daffodil.schema.annotation.props.gen.BinaryCalendarRep
+import org.apache.daffodil.util.Maybe.One
+import org.apache.daffodil.util.Misc
+
+import com.ibm.icu.util.Calendar
+
+case class ConvertBinaryCalendarSecMilliUnparser(
+  override val context: ElementRuntimeData,
+  binCalRep: BinaryCalendarRep,
+  epochTimeMillis: Long,
+  lengthInBits: Int,
+  hasTZ: Boolean)
+  extends PrimUnparser {
+
+  /**
+   * Primitive unparsers must override runtimeDependencies
+   */
+  override lazy val runtimeDependencies = Nil
+
+  protected def putNumber(dos: DataOutputStream, value: Long, nBits: Int, finfo: FormatInfo): Boolean = {
+    dos.putLong(value, nBits, finfo)
+  }
+
+  def unparse(state: UState): Unit = {
+
+    val node = state.currentInfosetNode.asSimple
+
+    val calValue = node.dataValue match {
+      case dc: DFDLCalendar => dc.calendar
+      case x => Assert.invariantFailed("ConvertTextCalendar received unsupported type. %s of type %s.".format(x, Misc.getNameFromClass(x)))
+    }
+
+    // Adjust the time based on time zone - if a time zone wasn't specified, Calendar will assume the default
+    // time zone for the user instead of TimeZone.UNKNOWN_ZONE so we need to adjust to get the correct time
+    // Note that setting the correct time zone for the calendar will not adjust the time.
+    val epochTime = if (!hasTZ) {
+      val tz = calValue.getTimeZone
+      val gmtOffset = calValue.get(Calendar.ZONE_OFFSET)
+      val dstOffset = if (tz.inDaylightTime(calValue.getTime)) tz.getDSTSavings else 0
+      epochTimeMillis - (gmtOffset + dstOffset)
+    } else {
+      epochTimeMillis
+    }
+
+    val diff: Long = binCalRep match {
+      case BinaryCalendarRep.BinarySeconds => (calValue.getTimeInMillis - epochTime) / 1000
+      case BinaryCalendarRep.BinaryMilliseconds => (calValue.getTimeInMillis - epochTime)
+      case _ => Assert.impossibleCase
+    }
+
+    val dos = state.dataOutputStream
+    val res = putNumber(dos, diff, lengthInBits, state)
+
+    if (!res) {
+      Assert.invariant(dos.maybeRelBitLimit0b.isDefined)
+      UnparseError(One(state.schemaFileLocation), One(state.currentLocation), "Insufficient space to unparse element %s, required %s bits, but only %s were available.",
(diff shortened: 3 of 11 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -1951,6 +1966,87 @@
...
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+  <!-- xs:dateTime with binaryCalendarRep='binarySeconds' with implicit length -->
+  <tdml:parserTestCase name="dateTimeBin4" root="dateTimeBin4"
+    model="SimpleTypes-binary" description="Section 5 Schema types-dateTime - DFDL-5-016R"
+    roundTrip="true">
+
+    <tdml:document>
+      <tdml:documentPart type="bits">11111111 11111110 10101110 10000000</tdml:documentPart>
```
mbeckerle: You can make a 64-bit test from this one, by just multiplying this number by 1000 to get milliseconds. Since this number uses all 32 bits, that will push it into 64-bit territory, but the date/time will be the same.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -1951,6 +1966,87 @@
...
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <dateTimeBin4>1869-12-31T00:05:00.000000+00:00</dateTimeBin4>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <!-- xs:dateTime with binaryCalendarRep='binaryMillieconds' -->
+  <tdml:parserTestCase name="dateTimeBin5" root="dateTimeBin5"
```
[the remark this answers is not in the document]
mbeckerle: I'm curious now as to what date/time 64-bits of all 1's as binaryMilliseconds is.  I wonder if it can be represented with 4-digits for year. Probably not. I don't know if the Java date/time libraries will work for this. It would be good to have tests that characterize the range that Java's date/time stuff supports. I.e., what is the max 64-bit integer for binary milliseconds that they are able to support, and what kinds of errors occur when the limits are exceeded.

Another interesting dateTime would be 9999-12-31T23-59-59.999999Z  which is the last millisecond of the "4-digit year" era. We should certainly have tests that have that, and the next millisecond after it, which would be
10000-01-01T00-00-00.000000Z.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -775,6 +782,25 @@ trait ElementBaseGrammarMixin
       }
 
       case PrimType.Boolean => { new BinaryBoolean(this) }
+
+      case PrimType.DateTime | PrimType.Date | PrimType.Time => {
+        (primType, binaryCalendarRep) match {
+          case (PrimType.DateTime, BinaryCalendarRep.BinarySeconds) => (lengthUnits, binaryNumberKnownLengthInBits) match {
+            case (LengthUnits.Bytes, 32) => new ConvertBinaryDateTimePrim(this, binaryNumberKnownLengthInBits)
```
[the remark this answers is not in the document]
efinnegan: I would think it would be ok but I added this because it's specified in the spec - "12.3.7.2.5   Length of Base-2 Binary Calendar Elements
Calendars with binary representation and dfdl:binaryCalendarRep ‘binarySeconds’ or ‘binaryMilliseconds’ are represented as a bit string which contains a base-2 representation.The specified length must be either exactly 4 bytes or exactly 8 bytes respectively.

The dfdl:lengthUnits property must be 'bytes'. It is a schema definition error otherwise."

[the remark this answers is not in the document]
mbeckerle: Yes, these restrictions on lengthUnits turned out to be a bad idea ultimately. They require a special check, and add no value over just requiring the length to ultimately be exactly 32 or 64 bits regardless of units used to express that. When a data format expresses *everything* in units of bits, these restrictions just make it awkward to shift to bytes units for these few cases. 

But the DFDL spec does require the check.... unfortunately.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +189,83 @@ object TextCalendarConstants {
...
+    if (pos.getIndex != binCalEpoch.length || pos.getErrorIndex >= 0) {
+      val errIndex = if (pos.getErrorIndex >= 0) pos.getErrorIndex else pos.getIndex
+      PE(start, "Parsing the binaryCalendarEpoch from String '%s' -  Failed to parse at index: %d.", binCalEpoch, errIndex)
+      return
+    }
+
+    if (binCalRep.eq(BinaryCalendarRep.BinarySeconds)) {
+      cal.add(Calendar.SECOND, int.intValue())
+    } else if (binCalRep.eq(BinaryCalendarRep.BinaryMilliseconds)) {
+      cal.add(Calendar.MILLISECOND, int.intValue())
```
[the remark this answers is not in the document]
efinnegan: Yes cal.add only accepts an int. I will look at setTimeInMillis.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +189,83 @@ object TextCalendarConstants {
...
+        return
+      }
+    }
+
+    val newCal = xsdType.toLowerCase() match {
+      case "time" => new DFDLTime(cal, hasTZ)
+      case "date" => new DFDLDate(cal, hasTZ)
+      case "datetime" => new DFDLDateTime(cal, hasTZ)
+      case _ => Assert.impossibleCase
+    }
```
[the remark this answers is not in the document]
mbeckerle: Not sure I understand when you say "time" and "date" aren't valid. They're allowed XSD types in DFDL.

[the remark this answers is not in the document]
efinnegan: I think he means these should be removed because for binaryCalendarRep of binarySeconds or binaryMilliseconds only dateTime is valid. They will be valid for the other types of binaryCalendarRep.

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpacalbin100_01.dfdl.xsd
```diff
@@ -52,7 +53,8 @@
 	<xs:element name="testBinaryMilliseconds" type="xs:dateTime">
 		<xs:annotation>
 			<xs:appinfo source="http://www.ogf.org/dfdl/">
-				<dfdl:element binaryCalendarRep="binaryMilliseconds" binaryCalendarEpoch="2009-12-24T05:15:00+05:00"/>
+				<dfdl:element binaryCalendarRep="binaryMilliseconds" binaryCalendarEpoch="2009-12-24T05:15:00+05:00"
+				  calendarPatternKind="explicit" calendarPattern="yyyy-MM-dd'T'HH:mm:ssZZZZ" />
```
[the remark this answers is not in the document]
mbeckerle: For binary calendars, patterns really only make sense for packed decimal or BCD representations. For twos-complement integer representations the DFDL spec is clear the pattern is not used.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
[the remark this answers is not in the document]
stevedlawrence: xs:time and xs:date aren't allowed when binaryCalendarRep is binarySeconds or binaryMilliseconds. The checks happen in the grammar, so those cases should never be hit. "not valid" probably wasn't the right words, must just that it's not possible.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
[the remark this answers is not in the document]
mbeckerle: I forgot these are signed. So there are extrema both positive future date/times and past date/times. 

Back of the envelope calculation shows that 64-bits of milliseconds can represent a delta of like 50+ million years, so yeah, they can go way beyond 4-digit years.

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpacalbin100_01.dfdl.xsd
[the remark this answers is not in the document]
stevedlawrence: Gotcha. Looks like we're just passing in too many arguments into the BinaryCalendarParser then, likely because it is extending the TextCalendarParser which does need those extra arguments. I *think* we just need the calendarEv and maybe a couple others parameters. All the other stuff should go away. And it even looks like some parameters are never used, even in the TextCalendarParser. Might be good to clean some of that up while we're in this code and splitting into a CalendarBaseParser (if needed).

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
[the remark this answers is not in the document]
efinnegan: For very large binaryMilliseconds, I don't see anything in the spec about expected behavior. Right now if adding puts it over the max long it will have a negative date.

Also, a Calendar object can either be in strict or lenient mode. When in strict mode, an IllegalIcuArgumentException  will be thrown for calling setTimeInMillis with a long that is outside of the permitted range. In lenient mode it will give the min or max date permitted (-5838269-09-20 or 5828963-12-20). Default is lenient.

i.e. With binaryCalendarEpoch="2000-06-15T03:25:19" and binaryMilliseconds (as a long) = 9223372036854775807 the resulting date is "-5838269-09-20T00:00:00"

So which do we want:
1. Very large numbers will "loop", using the min and max where necessary.
2. Very large numbers should not loop and should cap at max/min.
3. Make everything strict and throw errors if the date would be higher/lower than the max/min.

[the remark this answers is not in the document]
stevedlawrence: The following code makes it look like the actual date range is huge:
```
> val format = new java.text.SimpleDateFormat("MMM dd HH:mm:ss yyyy G")
> format.setTimeZone(java.util.TimeZone.getTimeZone("UTC"))
> val cal = java.util.Calendar.getInstance()
> cal.setTimeInMillis(Long.MaxValue)
> format.format(cal.getTime())
Sun Aug 17 02:12:55 EST 292278994 AD
> cal.setTimeInMillis(Long.MinValue)
> format.format(cal.getTime())
Sun Dec 02 11:47:04 EST 292269055 BC
```
So the min/max dates since the unix epoc are nearly 300 million years in the past and future. That's a huge timespan--seems like a very reasonable restriction to say we don't support dates before or after that. I wonder if dates that large even work with our text parsing stuff, I'd guess not.

So I'd vote for option 3, if the number of milliseconds since dfdl:binaryCalendarEpoch is greater or less than Long.MaxLong/MinLong, that should just be an error.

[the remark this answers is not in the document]
efinnegan: For the com.ibm.icu.util.Calendar the range is smaller, but still pretty large. 

"The range of handled dates has been narrowed from +/- ~300,000,000 years to +/- ~5,000,000 years. In practical terms this should not affect clients. However, it does mean that client code cannot be guaranteed well-behaved results with dates such as Date(Long.MIN_VALUE) or Date(Long.MAX_VALUE). Instead, the Calendar protected constants should be used. "

So we would have to error at their max if we go with option 3.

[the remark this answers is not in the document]
stevedlawrence: 5 million is still pretty reasonable. I'd still vote for an error at whatever long value is ICU max.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -191,3 +185,46 @@ case class ConvertTextDateTimePrim(e: ElementBase) extends ConvertTextCalendarPr
   protected override val implicitPattern = "uuuu-MM-dd'T'HH:mm:ss"
   protected override val validFormatCharacters = "adDeEFGhHkKmMsSuwWvVyXxYzZ".toSeq
 }
+
+abstract class ConvertBinaryCalendarPrimBase(e: ElementBase, guard: Boolean, lengthInBits: Long)
+  extends ConvertCalendarPrimBase(e, guard) {
+
+  lazy val epochCalendar: Calendar = {
+    val cal = Calendar.getInstance
+    val sdf = new SimpleDateFormat("uuuu-MM-dd'T'HH:mm:ss", ULocale.ENGLISH)
```
stevedlawrence: For the locale parameter, should this use dfdl:calendarLanguage instead of ULocale.ENGLISH?

Unrelated to this bug, this might be the problem with DAFFODIL-1945--it looks like whenever we create a SimpleDateFormat we do not specify a locale in the constructor, so it's probably using the default locale. I *think* we want to always provide dfdl:calenarLanguage. I'll update that bug with this information.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -191,3 +185,46 @@ case class ConvertTextDateTimePrim(e: ElementBase) extends ConvertTextCalendarPr
...
+    sdf.setTimeZone(tz)
+    cal.setTimeZone(tz)
+    cal.setLenient(false)
+    cal.setTime(sdf.parse(e.binaryCalendarEpoch))
+    cal
+  }
+
+  override lazy val parser = new ConvertBinaryCalendarParser(
+    e.elementRuntimeData,
+    (calendarTz.isDefined),
```
stevedlawrence: Remove parens.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertBinaryCalendarUnparser.scala
```diff
@@ -0,0 +1,69 @@
...
+    val calValue = node.dataValue match {
+      case dc: DFDLCalendar => dc.calendar
+      case x => Assert.invariantFailed("ConvertTextCalendar received unsupported type. %s of type %s.".format(x, Misc.getNameFromClass(x)))
+    }
+
+    val diff: Long = binCalRep match {
+      case BinaryCalendarRep.BinarySeconds => (calValue.getTimeInMillis - epochTimeMillis) / 1000
+      case BinaryCalendarRep.BinaryMilliseconds => (calValue.getTimeInMillis - epochTimeMillis)
+      case _ => Assert.impossibleCase
+    }
```
stevedlawrence: Should we check for integer overflow/underflow here? Say the calValue is really large and epochTimeMillis is a really large negative. This is going to overflow and could become a time in the valid range, but I would expect that to be an error. Same idea for underflow.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +169,37 @@ object TextCalendarConstants {
...
+  override lazy val runtimeDependencies = Nil
+
+  def parse(start: PState): Unit = {
+
+    val dis = start.dataInputStream
+    val slong: Long = dis.getSignedLong(lengthInBits, start)
+
+    if (binCalRep.eq(BinaryCalendarRep.BinarySeconds)) {
+      epochCal.add(Calendar.SECOND, slong.intValue())
+    } else if (binCalRep.eq(BinaryCalendarRep.BinaryMilliseconds)) {
```
stevedlawrence: Instead of having this conditional, and using epochCal.add vs epochCal.setTimeInMillis, could you just always use epochCal.setTimeInMillis, andwhen binCalRep is binarySecond, multiply the value by 1000? That would also mean you only need to pass in epochCal.getTimeInMillis to this method instead of the full cal.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +169,37 @@ object TextCalendarConstants {
...
+      val newTime = epochCal.getTimeInMillis + slong
```
stevedlawrence: Same question about overflow/underflow here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +169,37 @@ object TextCalendarConstants {
...
+      try {
+        epochCal.setTimeInMillis(newTime)
+      } catch {
+        case e: IllegalArgumentException => {
+          PE(start, "Adding %s to the binaryCalendarEpoch: %s.", slong, e.getMessage())
```
stevedlawrence: Using the word "adding" feels like something a developer would like to see rather than a user. Maybe something instead like "%s milliseconds from the binaryCalendarEpoch is out of range of valid values".

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +169,37 @@ object TextCalendarConstants {
...
+
+    val dis = start.dataInputStream
+    val slong: Long = dis.getSignedLong(lengthInBits, start)
+
+    if (binCalRep.eq(BinaryCalendarRep.BinarySeconds)) {
+      epochCal.add(Calendar.SECOND, slong.intValue())
+    } else if (binCalRep.eq(BinaryCalendarRep.BinaryMilliseconds)) {
+      val newTime = epochCal.getTimeInMillis + slong
+      try {
+        epochCal.setTimeInMillis(newTime)
```
stevedlawrence: This isn't thread safe. This code could be executed by multiple threads, modifying the calendar at the same time. Instead, we need to create a new thread local Calendar based on calendarEv. It looks like the Parser deals with this by cloning the calendar. In other cases we've used a ThreadLocal, so we could alternatively change our CalendarEv to return a ThreadLocal[Calendar] instead of a Calendar.

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpacalbin100_01.dfdl.xsd
```diff
@@ -43,7 +44,7 @@
 	<xs:element name="testBinarySeconds" type="xs:dateTime">
 		<xs:annotation>
 			<xs:appinfo source="http://www.ogf.org/dfdl/">
-				<dfdl:element  binaryCalendarRep="binarySeconds" binaryCalendarEpoch="1970-01-01T00:00:00+00:00"/>
```
stevedlawrence: Does including the timezone not work? I think we want to support both with and without the timezone. And since this test came from IBM, we probably want to keep it as similar to the original as possible.

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
```diff
@@ -327,7 +327,7 @@
 		<document><documentPart type="byte">000000000000003C</documentPart></document>
 		<infoset>
 			<dfdlInfoset xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-			  <testBinaryMilliseconds xsi:type="xs:dateTime">2009-12-24T05:15:00.060000+05:00</testBinaryMilliseconds>
+			  <testBinaryMilliseconds xsi:type="xs:dateTime">2009-12-24T05:15:00.060000+00:00</testBinaryMilliseconds>
```
stevedlawrence: Interesting. The original test had a binaryCalendarEpoch in the +5 timezone, and so the expected infoset value was in the +5 timezone. I would expect that the output should be in the timezone of the dfdl:calendarTimeZone property. This is because with binaryCalendarSec/MilSec no timezone is in the data, and the dfdl:calendarTimeZone property says: This property provides the time zone that will be assumed if no time zone explicitly occurs in the data. Seems like the test was wrong, and even if dfdl:binaryCalendarEpoch includes a timezone, it shouldn't affect the infoset timezone. Thoughts?

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -1951,6 +1970,236 @@
...
+  <tdml:parserTestCase name="dateTimeBin11" root="dateTimeBin6"
+    model="SimpleTypes-binary" description="Section 5 Schema types-dateTime - DFDL-5-016R"
+    roundTrip="true">
+
+    <tdml:document>
+      <tdml:documentPart type="bits">01111111 11111111 11111111 11111111 11111111 11111111 11111111 11111111</tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>millis value less than lower bounds for a Calendar</tdml:error>
```
stevedlawrence: Does the actual error message include what the range is? That might be helpful to display to a user if it's not too much of a hassle.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -2035,6 +2039,155 @@
...
+  <tdml:parserTestCase name="dateTimeBin13" root="dateTimeBin6"
+    model="SimpleTypes-binary" description="Section 5 Schema types-dateTime - DFDL-5-016R"
+    roundTrip="true">
+
+    <tdml:document>
+      <tdml:documentPart type="bits">00000010 10001101 01000110 11111011 11111100 10101110 01100010 11101001</tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>millis value greater than upper bounds for a Calendar</tdml:error>
```
mbeckerle: Let's change this error message so it says "Milliseconds" not "millis"

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
stevedlawrence: Ok, that explanation makes sense to me.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
efinnegan: No it doesn't show the range, just what it tried to set so I can look at including the range.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
efinnegan: This part is the message from the exception. Would it be better to not include it? Here's a current error message after updating with the wording that Steve suggested. The part from the exception thrown is in bold. Steve also suggesting adding the range allowed to this.

"Parse Error: 9223372036854775807 milliseconds from the binaryCalendarEpoch is out of range of valid values: **millis value less than lower bounds for a Calendar : -9223371075815256809**."

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
efinnegan: I removed the time zones from the binaryCalendarEpoch because I think Steve noted before that it shouldn't be using the calendarPattern because that doesn't apply for binarySeconds and binaryMilliseconds. So when parsing the epochCalendar I am using the default pattern for dateTime which is "yyyy-MM-dd'T'HH:mm:ss". 

But if calendarTimeZone doesn't apply here either than this is wrong too since I was using it for epochCalendar. So without specifying a pattern for epochCalendar can we allow both "yyyy-MM-dd'T'HH:mm:ss" and "yyyy-MM-dd'T'HH:mm:ssZZZZ"? Then the first would use UTC.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
efinnegan: I updated this to use calendarLanguage but then reading comments below about calendarTimeZone have me wondering if this is correct since calendarLanguage is in the same table that specifies the properties are for text or binary with a packed decimal representation, not binarySeconds or binaryMilliseconds. I would think using calendarLanguage here is still better but want to check.

mbeckerle: I think you are doing this right. Calendar language won't matter if these objects end up being used for binary calendars. For text calendars, calendar language should be used.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertBinaryCalendarUnparser.scala
efinnegan: I don't think it's possible to get a valid date from overflowing because the valid range is -184303902528000000 to 183882168921600000, so even if the calValue was the max you would need epochTimeInMillis to be at least -18078558002259951615 to overflow to a valid value but that is greater than max long.

One thing you can do is put an invalid date into binaryCalendarEpoch that overflows to a negative, but that is also a problem for a calendar with a text representation. I'm not sure if that has been discussed before.

stevedlawrence: Yeah, you're right. I don't think overflow is an issue. The range of valid values is just too small to overflow to a valid date.

I *think* we do the right thing for text parsers. It looks like the SimpleDateFormat#parse method doesn't throw an exception on failure. Instead, you just need to look at the parse position to determine if there was an error. I would assume that if the text value was out of range, then parse() would set the error position appropriately. And then calendar.getTime is called. We probably want to do something similar for the binaryCalendarEpoch and throw an SDE if there's an error. Here's the logic for that in the text date time parser:

https://github.com/apache/incubator-daffodil/blob/master/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala#L136-L157

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -191,3 +186,61 @@ case class ConvertTextDateTimePrim(e: ElementBase) extends ConvertTextCalendarPr
   protected override val implicitPattern = "uuuu-MM-dd'T'HH:mm:ss"
   protected override val validFormatCharacters = "adDeEFGhHkKmMsSuwWvVyXxYzZ".toSeq
 }
+
+abstract class ConvertBinaryCalendarPrimBase(e: ElementBase, guard: Boolean, lengthInBits: Long)
+  extends ConvertCalendarPrimBase(e, guard) {
+
+  lazy val epochCalendar: Calendar = {
```
stevedlawrence: All this stuff in here is specific to binaryCalendarRep="binarySeconds/binaryMilliseconds" so probably shouldn't be in this base class. We'll probably want this base class for the other binaryCalendarReps, so maybe just move it all into ConvertBinaryDateTimePrim? This abstract class will just be empty for now.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -191,3 +186,61 @@ case class ConvertTextDateTimePrim(e: ElementBase) extends ConvertTextCalendarPr
...
+    } catch {
+      case e: IllegalArgumentException => {
+        SDE("Failed to parse binaryCalendarEpoch: %s.", e.getMessage())
+      }
+    }
+
+    cal
+  }
+
+  override lazy val parser = new ConvertBinaryCalendarParser(
```
stevedlawrence: Should this and the below prim be renamed to something that makes it clear is for seconds/milliseconds so that it will be differentiated from other binaryCalendarRep's when they are added?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -191,26 +192,40 @@ abstract class ConvertBinaryCalendarPrimBase(e: ElementBase, guard: Boolean, len
...
+    if (pos.getIndex != e.binaryCalendarEpoch.length || pos.getErrorIndex >= 0) {
+      // binaryCalendarEpoch didn't match the first format with timezone, so try without
+      cal.clear()
+      pos = new ParsePosition(0)
+      val sdf = new SimpleDateFormat("uuuu-MM-dd'T'HH:mm:ss", ULocale.ENGLISH)
+      cal.setTimeZone(TimeZone.UNKNOWN_ZONE)
+      sdf.parse(e.binaryCalendarEpoch, cal, pos)
+
+      if (pos.getIndex != e.binaryCalendarEpoch.length || pos.getErrorIndex >= 0) {
+        SDE("Failed to parse binaryCalendarEpoch - Format must match the pattern 'uuuu-MM-dd'T'HH:mm:ss' or 'uuuu-MM-dd'T'HH:mm:ssZZZZ'")
```
mbeckerle: line too long for review tool. Can't see all the way right.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertBinaryCalendarUnparser.scala
```diff
@@ -0,0 +1,82 @@
...
+      case dc: DFDLCalendar => dc.calendar
+      case x => Assert.invariantFailed("ConvertTextCalendar received unsupported type. %s of type %s.".format(x, Misc.getNameFromClass(x)))
+    }
+
+    // Adjust the time based on time zone.
+    val epochTime = if (!hasTZ) {
+      val tz = calValue.getTimeZone
+      var gmtOffset = calValue.get(Calendar.ZONE_OFFSET)
+      if (tz.inDaylightTime(calValue.getTime)) gmtOffset += tz.getDSTSavings
+      epochTimeMillis - gmtOffset
```
stevedlawrence: Can you add a comment about what's going on here? It's not immediately obvious to me why things like DST is needed. You might also make it more scala-y by not using a var. Maybe something like:
```
val gmtOffset = ...
val dstOffset = if (tz.inDaylightTime(...)) tz.getDSTSavings else 0
epochTimeMillis - (gmtOffset + dstOffset)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesDateTime1.scala
```diff
@@ -188,3 +169,40 @@ object TextCalendarConstants {
...
+
+  def parse(start: PState): Unit = {
+
+    val dis = start.dataInputStream
+    val slong: Long = dis.getSignedLong(lengthInBits, start)
+    val cal = epochCal.clone.asInstanceOf[Calendar]
+
+    val millisToAdd: Long =  {
+      if (binCalRep.eq(BinaryCalendarRep.BinarySeconds)) { slong * 1000 }
+      else { slong }
```
stevedlawrence: Use a match/case here.

stevedlawrence:
+1 with recent changes

PR #73, opened 2018-05-30
Enable the sbt-rat plugin to perform Apache Rat checks

stevedlawrence:
One can now run 'sbt ratReport' to generate an Apache Rat report in
target/rat.txt to check for any unapproved licensed files. Running 'sbt
ratCheck' will do the same, but will return an error if any unapproved
licenses are found. Specifically, this patch does the following:

- Enable the sbt-rat plugin
- Removes the .rat-excludes file and instead uses the sbt ratExcludes
  setting, which allows exluding full paths rather than just file names
- Modify Travis CI config to perform a ratCheck on every commit
- Add matchers for the passera and Scala BSD licenses, and mark them as
  approved licenses
- Rename LICENSE-bin and NOTICE-bin to bin.LICENSE and bin.NOTICE.
  Apache Rat does not see them as NOTICE files otherwise
- Fix URL in W3C files so rat can detect the license

DAFFODIL-1937


stevedlawrence:
Release Audit Tool. It's usually run by incubator mentors/voters during releases to make sure there aren't incorrectly licensed files. But with this we can now check it continually for every change.

PR #74, opened 2018-06-12
Daffodil trailing sep

mbeckerle:
This PR is ready for preliminary review, as it is a major set of changes to Daffodil. I didn't want to take a block of time away without at least letting others get eyes on this.

See UML diagrams here:

https://cwiki.apache.org/confluence/display/DAFFODIL/Proposal%3A+Sequence+Implementation

There are a number of things not working still. E.g., the Parser MarkPool leaks Mark objects all over the place, and there are 101 daffodil-test regressions still.

Nevertheless, I think the new code organization and classes/traits are worth looking over.


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -618,8 +622,9 @@ final class ElementRuntimeData(
   @TransientParam targetNamespaceArg: => NS,
   @TransientParam thisElementsNamespaceArg: => NS,
   @TransientParam optSimpleTypeRuntimeDataArg: => Option[SimpleTypeRuntimeData],
-  @TransientParam minOccursArg: => Option[Int],
-  @TransientParam maxOccursArg: => Option[Int],
+  @TransientParam minOccursArg: => Long,
+  @TransientParam maxOccursArg: => Long,
```
stevedlawrence: Can we actually support more more than IntMax array elements. I believe we still use an Array as the underlying data structure for storing arrays so I think these values are still limited to Int's.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/DelimiterParsers.scala
```diff
@@ -158,7 +158,7 @@ class DelimiterTextParser(
           return
         } else {
           // no match and no ES in delims
-          PE(start, "Delimiter not found.")
+          PE(start, "%s '%s' not found.", delimiterType.toString, start.mpstate.delimiters.last.lookingFor)
```
stevedlawrence: This small change is a long time coming!

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NilEmptyCombinatorParsers.scala
```diff
@@ -19,14 +19,8 @@ package org.apache.daffodil.processors.parsers
...
 case class SimpleNilOrValueParser(ctxt: TermRuntimeData, nilParser: Parser, valueParser: Parser)
-  extends AltCompParser(ctxt, Seq(nilParser, valueParser))
-
-case class SimpleEmptyOrValueParser(ctxt: TermRuntimeData, emptyParser: Parser, valueParser: Parser)
-  extends AltCompParser(ctxt, Seq(emptyParser, valueParser))
+  extends ChoiceParser(ctxt, Seq(nilParser, valueParser))
 
 case class ComplexNilOrContentParser(ctxt: TermRuntimeData, emptyParser: Parser, contentParser: Parser)
-  extends AltCompParser(ctxt, Seq(emptyParser, contentParser))
+  extends ChoiceParser(ctxt, Seq(emptyParser, contentParser))
```
stevedlawrence: I like this change to replace AltCompParser with ChoiceParser, they're virtually he same. One thing to consider though is that when the ChoiceParser fails it creates parser errors that reference "Choices". There's potential confusion there. The user might think the error is related to xs:choice in the schema but it's actually just us internally trying to parse two different things.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
```diff
@@ -0,0 +1,549 @@
...
+      val child = children(scpIndex)
+      child match {
+        case parser: RepSeparatedChildParser => {
+
+          val loopState = parser.loopState(pstate)
+
+          // push new array context for array/optional
+          parser.startArray(pstate)
+
+          val maybeStartState =
```
stevedlawrence: Some thoughts on the mark leak since this is notoriously hard to track down.

I think this might be leaked. If a parse1 call throws an exception in parseOne, the processFailedChildParseResults method won't get called, which is supposed to discard/reset marks. The tryParseDetectMarkLeaks method will clean up the priorState mark leaked, but does not handle the maybeStartState. So I *think* you just need to discard the maybeStartState when priorState is discarded.

Anothe potential leak, in processSuccessfulChildParseResults, it looks like the state isn't reset/discarded if isFixedOccurs. I think some other code paths related to processSuccessful might be missing discards as well.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -0,0 +1,335 @@
...
+    //
+    // failedChild(pstate, priorState) // discards prior state
+    if (isFixedOccurs) {
+      //
+      // in fixed occurs, there are no points of uncertainty for the
+      // individual elements, so a failure is a failure of the whole loop.
+      // And any discriminator that has been set is the discriminator
+      // of some surrounding scope.
+      //
+      ParseAttemptStatus.FailedEntireArray
```
stevedlawrence: Potential mark leak here?

mbeckerle:
Closing - will incorporate changes, but this PR was really a very preliminary code review, and the code has evolved in structure substantially since this PR. A new PR with all the changes will be created when ready.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -359,6 +360,25 @@ trait Term
...
+        case Some(s) => {
+          val allRequired = s.groupMembers.filter(_.isRequired)
+          val lastDeclaredRequired = allRequired.last
+          if (s.groupMembers.indexOf(lastDeclaredRequired) < s.groupMembers.indexOf(this)) true
+          else false
+        }
+      }
+      res
+      // Since we can't determine at compile time, return false so that we can continue processing.
+      // Runtime checks will make final determination.
```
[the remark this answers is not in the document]
mbeckerle: Agree, I think the comment is misworded, or just wrong or spurious. We can't determine everything about whether something is actually trailing, but we can determine this isLastDeclaredRequiredElementOfSequence.

Also I think all these complex attribute LVs of the DSOM objects need scaladoc, as you point out in other comments, they are potentially duplicated logic, or the subtle distinction between them is not apparent from their names.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -359,6 +360,25 @@ trait Term
   def isKnownRequiredElement = false
   def hasKnownRequiredSyntax = false
 
+  def hasPotentiallyTrailingInstances: Boolean = false
+  final def isPotentiallyTrailing = LV('isPotentiallyTrailing) {
+    if (!isRequired) {
```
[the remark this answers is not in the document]
mbeckerle: For trailing separator stuff, we don't care about the required sequence child because its separator (whether pre/in/postfix) is required along with it. We care about the separator corresponding to an optional element, so that we can decide whether to absorb it without creating the element or not. 

But the language in the spec doesn't match this, so I think isPotentiallyTrailing member should be renamed to hasPotentiallySuppressedTrailingEmptySeparator, with some scaladoc explaining that this is closely related to "potentially trailing" in the spec. but is a little different.

There is starting to be a terminology problem, both in the DFDL spec, and in Daffodil. The spec uses "required" and "optional" in different senses.  We need to better distinguish those in Daffodil code. 

isRequired needs scaladoc, but I think it means "has at least 1 required instance". I'm not sure isRequired is meaningful on computed (inputValuecalc) elements. 

The term Required in the runtime however, means "at this array index, the index is less than or equal to minOccurs, and the OCK is implicit, so minOccurs matters, or if OCK is fixed or  expression, then all indexes are Required. ".  

Probably that should change to RequiredOccurrence in the runtime.

The term Optional in the runtime means "at this array index, the index is greater than minOccurs, and OCK is implicit, so minOccurs and maxOccurs matter, or OCK='parsed' for any index." 

Probably that should change to SpeculatedOccurence in the runtime.

There's also a problem with isArray and isOptional. The term Array appears in much of the runtime meaning "not scalar and fixed or variable number of occurrences". That's not consistent with isArray which means known to have the possibility of having 2 or more occurrences.  isOptional means minOccurs 0, maxOccurs 1 and isn't isArray. I think. Have to verify. But anyway Step one to sort this out is scaladoc on these to point out the differences. 

Unfortunately, all 3 of isRequired, isOptional, and isArray are passed through to the TermRuntimeData structure, forcing the runtime code to figure out how to use what they mean. There's quite a lot of code in the runtime that has to determine scalar, vs. has fixed/specified number of occurrences, vs. has number of occurrences determined by speculation (possibly after minOccurs required occurrences). It does this not by looking at isRequired, isArray, isOptional, but by testing properties that are passed through to parsers so are part of their runtime state. 

Some renaming could help. E.g, in the runtime, use better names like Scalar, Specified, or Speculated perhaps. 

I will add a TODO/FIXME comment about the unordered parent thing. There may be other things that look upward that may also have this future bug once unordered sequences are introduced.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/SequenceGrammarMixin.scala
```diff
@@ -17,36 +17,113 @@
...
+   * a time, orchestrated by the surrounding sequence's processor.
+   */
+  private def sequenceChild(child: Term, groupIndex: Int): SequenceChild = {
+    import columnConstants._
+    val (max, min, ock) = child match {
+      case e: EB => (e.maxOccurs, e.minOccurs, e.occursCountKind)
+      case _ => (1, 1, null)
+    }
+    val ssp = separatorSuppressionPolicy
+    val res = (child, sequenceKind, hasSeparator, ssp, ock, min, max) match {
```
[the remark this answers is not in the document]
mbeckerle: I took this out, but It needs to go back, as some of the error checks in this table are only relevant if hasSeparator is true. E.g., conflict of separatorSuppressionPolicy never with OCK implicit and unbounded maxOccurs is only relevant if hasSeparator is true. Otherwise the check need not, and should not occur.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -0,0 +1,124 @@
...
+ * unparser uses a Map to make the determination based on the element seen.
+ */
+case class ChoiceCombinator(ch: ChoiceTermBase, rawAlternatives: Seq[Gram])
+  extends Terminal(ch, !rawAlternatives.isEmpty) {
+
+  private lazy val alternatives = rawAlternatives.filterNot(_.isEmpty)
+
+  private lazy val parsers = alternatives.map { _.parser }.filterNot { _.isEmpty }
+
+  override def isEmpty = super.isEmpty || alternatives.isEmpty
```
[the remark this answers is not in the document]
mbeckerle: There's 2 kind of empty here. Empty as a grammar term, meaning the grammar optimizes out the entire combinator/terminal, and empty at the parser/unparser level, meaning that the parser was the NadaParser, which does show up a few places where a grammar combinator needed to make a parser/unparser distinction. In that case the generation of parser/unparser optimizes out the empty parser. 

But the two layers are distinct isEmpty optimizations. I will add comments to this effect.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ComplexTypeCombinator.scala
```diff
@@ -0,0 +1,53 @@
...
+  private lazy val u = body.unparser
+
+  override def toString() =
+    "<" + Misc.getNameFromClass(this) + ">" +
+      body.toString() +
+      "</" + Misc.getNameFromClass(this) + ">"
+
+  override lazy val parser: DaffodilParser =
+    if (p.isEmpty)
+      p
```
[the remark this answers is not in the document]
mbeckerle: This theoretically happens if the entire body of the complex type gets optimized away. 

Lots of the encapsulating wrapper combinators look at the parsers, and attempt to propagate the empty parser (NadaParser) to allow it to be fully optimized out. The runtime no longer has any actual use for NadaParsers. They always are optimized out. 

The only case I can think of is if someone has an element with a complex type that exists as a placeholder for the future, but it actually contains nothing other than doc string annotations, so fully optimizes away.  But then I'd think body.isEmpty would be true, and the whole thing would go away at the grammar level.  

I guess if you have complex types that are 100% parser-only or 100% unparser only, then these tests could be true. It *is* possible to have conditional behavior in schemas so that some parts of the schema are only for parsing or only for unparsing, but I'm not sure the schema compiler can detect that. 

ComplexTypeParser and ComplexTypeUnparser exist essentially for trace/debug purposes, and to be shared (in the future) parsers/unparsers for named complex type definitions. 

I might change these p.isEmpty to assert !p.isEmpty here instead, and see if it ever fails.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/DelimiterParsers.scala
mbeckerle: Doesn't work great tho. Sometimes you see %NL; '%NL;' not found. Other times 'xy' where x and y are those little picture Unicode chars for newline.  In the runtime code there is not currently a way to go from the delimiter or set of delimiters being found, to the schema author's expression of those delimiters. This is needed for a good diagnostic.  I'll report this inconsistency as a bug.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NilEmptyCombinatorParsers.scala
mbeckerle: Actually I think ChoiceParser is only used for choices now. The grammar no longer has the "|" operator in it at all, so there is no runtime alternation. All such conditional behavior is done by some explicit combinator. All the grammar alternatives must be resolvable at compilation time now, and use the "||" operator, which insists on turning into just one alternative at compile time.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
mbeckerle: All refactored away. I still have a few tests failing, but so far nothing appears to be leaking - in parsing anyway.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
mbeckerle: Yup. There *were* lots of these. I wasn't even trying to track them all down.

PR #75, opened 2018-06-13
Daffodil 1738 zoned decimal

jadams-tresys:
After bouncing on and off this task for a while Zoned Decimal parsing/unparsing should now be finished.

The one caveat to this is that numbers with decimal points aren't supported as the DecimalFormat library we use doesn't support the P/V symbols for indicating decimal point position in the textNumberPattern.  The code for floating point numbers is there, so in theory if support for P/V symbols was implemented it should work.


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ZonedTextParsers.scala
```diff
@@ -17,14 +17,421 @@
...
+
+    val mm = new HashMap[String, Set[String]] with MultiMap[String, String]
+
+    val dupes = mm.filter { case (k, s) => s.size > 1 }
+    val dupeStrings = dupes.map {
+      case (k, s) =>
+        "Non-distinct property '%s' found in: %s".format(k, s.mkString(", "))
+    }
+    context.schemaDefinitionUnless(dupeStrings.size == 0, dupeStrings.mkString("\n"))
+  }
```
stevedlawrence: Nothing is added to the HashMap here. The checkUnique funciton in standard text numbers ensures that various properites (e.g. groupingSep, exponentRep) don't have the same characters. Those properties don't apply here, so I think this function can be removed for zoned.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ZonedTextParsers.scala
```diff
@@ -17,14 +17,421 @@
...
+  override val xsdType = "decimal"
+  override val prettyType = "Unlimited Size Decimal"
+  override def isInvalidRange(n: java.lang.Number): Boolean = false
+
+  override def getStringFormat(n: JBigDecimal): String = {
+    n.toPlainString()
+  }
+}
+
+abstract class ZonedFormatFactoryBase[S](parserHelper: ConvertZonedNumberParserUnparserHelperBase[S]) extends Serializable {
```
stevedlawrence: This actually looks really similar to the standard text factory base. Can the zoned factory just extend that and pass in Maybe.Nope for properties that don't apply to zoned. I think the only difference is that the "+" sign needs to be stripped off of pattern?

daffodil-test/src/test/scala-debug/org/apache/daffodil/section13/zoned/TestZonedDebug.scala
```diff
@@ -0,0 +1,39 @@
...
+    runner.reset
+  }
+
+}
+
+class TestZoned {
+  import TestZoned._
+
+  @Test def testZonedTandemModified02(): Unit = { runner.runOneTest("ZonedTandemModified02") }
+  @Test def testZonedTandemModified05(): Unit = { runner.runOneTest("ZonedTandemModified05") }
```
stevedlawrence: Can you create a bug and mention it in a comment?

Also, we confirmed that this isn't a bug with the zoned implementation, but is actually a bug in how &#x80; is converted to bytes, right? It might be worth creating a really simple test case that doesn't involve any zoned stuff just to ensure there isn't something else funky going on. Maybe something that expects a single hex byte and show that &#x80; is converted to two bytes instead of one due to UTF-8.

daffodil-test/src/test/scala/org/apache/daffodil/section13/zoned/TestZoned.scala
```diff
@@ -0,0 +1,68 @@
...
+class TestZoned {
+  import TestZoned._
+
+  @Test def testZonedPatternFail01(): Unit = { runner.runOneTest("zoned_textNumberPattern_fail01") }
+  @Test def testZonedPatternFail02(): Unit = { runner.runOneTest("zoned_textNumberPattern_fail02") }
+  @Test def testZonedPatternFail03(): Unit = { runner.runOneTest("zoned_textNumberPattern_fail03") }
+  @Test def testZonedPatternFail04(): Unit = { runner.runOneTest("zoned_textNumberPattern_fail04") }
+  @Test def testZonedFloatFail01(): Unit = { runner.runOneTest("zoned_float_fail01") }
+  @Test def testZonedDoubleFail01(): Unit = { runner.runOneTest("zoned_double_fail01") }
+  //TODO: @Test def testZonedEncodingFail01(): Unit = { runner.runOneTest("zoned_encoding_fail01") }
```
stevedlawrence: I don't see this test in a TDML file anywhere, was this fixed?

jadams-tresys: So this was the test I was talking about when I wanted to know if there was any sort of list of the encodings that we supported and if there was an easy way to determine which ones were ASCII derived.  All this test would do is have an expected failure (SDE) when using a non-ASCII derived encoding.

I believe we decided that since there wasn't a simple list of supported encodings it wasn't worth the effort to set up the SDE for this case.  This test case can just be removed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ZonedTextParsers.scala
jadams-tresys: Will do.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DecimalUtils.scala
```diff
@@ -335,4 +335,142 @@ object DecimalUtils {
...
+        case TextZonedSignStyle.AsciiStandard => convertFromAsciiStandard(char)
+        case TextZonedSignStyle.AsciiTranslatedEBCDIC => convertFromAsciiTranslatedEBCDIC(char)
+        case TextZonedSignStyle.AsciiCARealiaModified => convertFromAsciiCARealiaModified(char)
+        case TextZonedSignStyle.AsciiTandemModified => convertFromAsciiTandemModified(char)
+      }
+
+      if (opneg && !negative) {
+        negative = true
+        decodedValue.insert(0, '-')
+      }
```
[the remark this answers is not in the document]
jadams-tresys: I think this was a carry over from when I didn't have a full understanding of the concept of overpunching the sign.  I assumed that in negative numbers all the digits were overpunched with the negative sign, not just one.  Will fix.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesZoned.scala
```diff
@@ -17,13 +17,162 @@
...
+  def numFormatFactory: ZonedFormatFactoryBase[S] = {
+    val h = helper
+
+    val pattern = {
+      val p = e.textNumberPattern
+
+      val noEscapedTicksRegex = """''""".r
+      val patternNoEscapedTicks = noEscapedTicksRegex.replaceAllIn(p, "")
+      val noQuotedRegex = """'[^']+'""".r
+      val patternNoQuoted = noQuotedRegex.replaceAllIn(patternNoEscapedTicks, "")
```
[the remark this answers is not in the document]
jadams-tresys: That regex should work.  I'll add it as a final check after the SDE's that I already added when checking the pattern as those give more descriptive error messages.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesZoned.scala
```diff
@@ -17,13 +17,162 @@
...
+      try {
+        new DecimalFormat(p)
+      } catch {
+        case ex: IllegalArgumentException => e.SDE("Invalid textNumberPattern: " + ex.getMessage())
+      }
+
+      p
+    }
+
+    val (roundingIncrement: MaybeDouble, roundingMode) =
```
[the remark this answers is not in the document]
jadams-tresys: This is how it was done for the standard text parser.  I'm OK with changing it as I'm not really sure what the purpose using MaybeDouble or One(e.textNumberRoundingMode) is. Was it there in case textNumberRoundingIncrement/Mode weren't defined by the schema?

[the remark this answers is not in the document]
stevedlawrence: I was just saying the type MaybeDouble shouldn't be needed due to type inference. If this comes from  the standard text parser, it's probably not worth changing. In general, I might say maybe this should be a trait like RoundingTextNumber or something, and then the code could be shared between text standard and text zoned, but it's such little code I'm not sure addng a new trait it's worth it. All that to so, probably just ignore my above comment keep this as is.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DecimalUtils.scala
```diff
@@ -335,4 +335,142 @@ object DecimalUtils {
...
   }
 
+  def convertFromAsciiStandard(digit: Int): (Int, Boolean) = {
+    if ((digit >= 48) && (digit <= 57)) // positive 0-9
+      return (digit - 48, false)
+    else if ((digit >= 112) && (digit <= 121)) // negative 0-9
+      return (digit - 112, true)
+    else
+      throw new NumberFormatException("Invalid zoned digit: " + digit)
+  }
```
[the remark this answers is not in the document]
jadams-tresys: I'll add some javadoc to this.
For AsciiStandard 0-9 represents overpunched positive 0-9 and p-y represent overpunched negative 0-9.  The boolean represents whether or not the digit was an overpunched negative. This is needed to handle the possibility of the signed digit being a 0.

For example, 123p represents -1230 in AsciiStandard, but the signed (overpunched) digit is a 0.  With the other digits I had originally just returned a negative version of them, but -0 == 0, so I needed some way to convey the fact that we are dealing with a negative number back to the where this function was called.

I will update the conversion functions to take chars instead of ints, as that will be more clear.

daffodil-lib/src/test/scala/org/apache/daffodil/util/TestDecimalUtils.scala
```diff
@@ -1249,4 +1249,478 @@ class TestDecimalUtils {
...
")
+    assertEquals(zonedToNumber("", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "-6")
+    assertEquals(zonedFromNumber("-6", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "")
+    assertEquals(zonedToNumber("", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "-7")
+    assertEquals(zonedFromNumber("-7", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "")
+    assertEquals(zonedToNumber("", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "-8")
+    assertEquals(zonedFromNumber("-8", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "")
+    assertEquals(zonedToNumber("", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "-9")
+    assertEquals(zonedFromNumber("-9", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start), "")
+  }
```
stevedlawrence: Nice test coverage! Can you add some negative test to make sure NumberFormatException is thrown at the right times?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertZonedNumberUnparser.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    val node = state.currentInfosetNode.asSimple
+    val value = node.dataValueAsString
+
+    // The type of value should have the type of S, but type erasure makes this
+    // difficult to assert. Could probably check this with TypeTags or Manifest
+    // if we find this is not the case. Want something akin to:
+    // Assert.invariant(value.isInstanceOf[S])
+
+    val opindex = pattern.indexOf('+')
```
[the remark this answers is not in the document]
stevedlawrence: I don't think you addressed this comment. I don't think pattern should be needed, just need to pass in the overpunch location.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PrimitivesTextNumber1.scala
```diff
@@ -582,17 +581,18 @@ class NumberFormatFactoryStatic[S](context: ThrowsSDE,
...
 
-  val expRep = {
-    Assert.invariant(exponentRepExpEv.isConstant)
-    getExponentRep(exponentRepExpEv.maybeConstant.get, context)
-  }
+  val expRep =
+    if (exponentRepExpEv.isEmpty) Nope else One {
+      Assert.invariant(exponentRepExpEv.get.isConstant)
+      getExponentRep(exponentRepExpEv.get.maybeConstant.get, context)
+    }
```
stevedlawrence: This is a really weird way to write this (I think I wrote the stuff this is based on). I guess it's best to be consistent with the already existing code that does something similar, but we should avoid this in the future.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ZonedTextParsers.scala
```diff
@@ -17,14 +17,110 @@
...
+
+    val opindex = pattern.indexOf('+')
+    val opl = {
+      if (opindex == 0)
+        OverpunchLocation.Start
+      else if (opindex == pattern.length - 1)
+        OverpunchLocation.End
+      else
+        OverpunchLocation.None
+    }
```
stevedlawrence: Same comment as above, this logic can be done in the primitives. The pattern shouldn't be needed, just the overpunch location.

daffodil-test/src/test/resources/org/apache/daffodil/section13/zoned/zoned.tdml
```diff
@@ -0,0 +1,427 @@
...
+      <tdml:documentPart type="text">123456789&#x80;</tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <tmod03>-1234567890</tmod03>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+
+  </tdml:parserTestCase>
+</tdml:testSuite>
```
stevedlawrence: Really nice test coverage.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DecimalUtils.scala
```diff
@@ -335,4 +336,154 @@ object DecimalUtils {
...
+    }
+    val opindex = opl match {
+      case OverpunchLocation.Start => 0
+      case OverpunchLocation.End => inStr.length - 1
+      case _ => -1
+    }
+
+    val encodedValue = {
+      if (opl == OverpunchLocation.None) {
+        inStr
```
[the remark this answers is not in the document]
stevedlawrence: Ah ok. Maybe add an assert so the code self documents that that should never happen?

daffodil-lib/src/main/scala/org/apache/daffodil/util/DecimalUtils.scala
```diff
@@ -335,4 +336,154 @@ object DecimalUtils {
...
+  def zonedToNumber(num: String, zonedStyle: TextZonedSignStyle, opl: OverpunchLocation.Value): String = {
+    val opindex = opl match {
+      case OverpunchLocation.Start => 0
+      case OverpunchLocation.End => num.length - 1
+      case _ => -1
+    }
+
+    val decodedValue = {
+      if (opl == OverpunchLocation.None) {
+        num
```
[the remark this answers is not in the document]
jadams-tresys: Yes, it should fail when DecimalFormat attempts to parse something like "A1" and I think that is an OK failure case.

PR #76, opened 2018-06-14
Allow Daffodil to build against Scala 2.12

stevedlawrence:
DAFFODIL-1652


build.sbt
```diff
@@ -92,24 +92,13 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "2.2.0-SNAPSHOT",
-  scalaVersion := "2.11.8",
+  scalaVersion := "2.12.6",
   scalacOptions ++= Seq(
```
stevedlawrence: I think we should be able to add ``crossScalaVersion := Seq("2.12.6", "2.11.12")`` to enable support for cross building with 2.11. I believe Scala 2.12 requires Java 8, and unfortunately we do have some users that are still stuck on Java 7, even though we officially only support Java 8, so ensuring that stil works would be good.

stevedlawrence: Related, we should also modify .travis.yml to include Scala versions 2.12.6 and 2.11.12 so that tests are run on both versions.

build.sbt
```diff
@@ -92,24 +92,13 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
...
-    "-Ydead-code",
-    "-Yinline" ,
-    "-Yinline-warnings",
-    "-Yopt:inline-global",
-    "-Yopt-warnings",
-    "-Ywarn-inaccessible",
-    "-Ywarn-infer-any",
-    "-Ywarn-nullary-override",
-    "-Ywarn-unused",
-    "-Ywarn-unused-import"
```
stevedlawrence: I guess these -Y options are no longer supported in Scala 2.12? Or is 2.12 just better at detecting some warnings? I don't think removing this is a blocker, but we should investigate adding back some of these warnings where it makes sense.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestGeneral.scala
```diff
@@ -97,7 +97,7 @@ class TestGeneral {
 
   @Test def test_1530_scala_version() {
     val versionStr = scala.util.Properties.versionNumberString
-    assertTrue("The scala version must begin with 2.11. The current version is: " + versionStr, versionStr.startsWith("2.11."))
+    assertTrue("The scala version must begin with 2.11 or 2.12. The current version is: " + versionStr, versionStr.startsWith("2.11.") || versionStr.startsWith("2.12."))
   }
```
stevedlawrence: I'm not even sure this test makes is necessary. Having to update this test everytime we bump the scala version seems unnecessary. As long as all the tests pass under a certain version of Scala, we should be happy. I'd say we should just remove this test.

stevedlawrence:
Turns out the new Scallop version broke some CLI tests, primarily due to ambiguities with the --validate option. Updated scallop hoping it would fix it, but it didn't. So added a commit which changes how the --validate option works, the 'mode' parameter is now required. This prevents ambiguities where the thing following --validate could be either a validate option or the input file. Now a mode *must* follow the --validate option. This breaks backwards compatibility, but is pretty minor.

stevedlawrence:
I realized that the LICENSE and eclipse classpath files needed to be updated to the new scala and scallop versions. Just added a commit that fixes those.

PR #77, opened 2018-07-05
Modifications to IO layer to support streaming input data

stevedlawrence:
- Modify the ByteBufferDataInputStream to no longer depend on
  ByteBuffers. It is now an InputSourceDataInputStream, and one can
  implement a new InputSource interface to provide Bytes. This also
  includes various changes like how bitPos and bitLimit are stored to
  simplify code (e.g. no offsets) and no longer requires a bitLimit to
  be set, since not all inputs may know a bit limit. Moves TLState
  members into the PState--the InputSourceDataInputStream class is now
  created by the user and so isn't necessarily created in thread from
  which it will be used, breaking the ThreadLocal functionality.
- Create two InputSource implementations, one using a ByteBuffer as the
  data store and one using a bucketing algorithm to support files larger
  than Int.MaxValue and can free up data that can no longer be
  backtracked to.
- Remove the DataLimits, most of these values weren't actually used.
  Instead, create new tunables for those that were and use those where
  appropriate.
- When we decode data to characters, we need to know exactly how many
  bits were used to decode each character. The Java decoders do not
  provide this information, requiring a lot of complex code to keep
  track, and even then there were bugs. This creates our own Decoders
  that provides the exact information we need and allow for further
  modifications that may be needed for things like
  dfdl:errorEncodingPolicies and dfdl:utf16Width.
- Remove the reporting and replacing decoders. Instead, we now just have
  a single decoder and it handles replacing/reporting based on the
  format info. Another benefit of our custom Decoders.
- Modifies the parse Scala/Java API to expect an
  InputSourceDataInputStream, created by the API user. Other public API
  methods are deprecated and are modified to
  create an InputSourceDataInputStream behind the scenes. API functions
  are also simplified to not take in a bit starting position or bit
  limit. This were really only used for testing, and were used fairly
  rarely. Alternative methods are used to set these values where
  necessary.
- Adds the --stream option to the CLI parse subcommand. When this is
  provided, if there is left over data at the end of a parse, the CLI
  will perform a new parse continuing where the previous left off.
- Modify the TDMLRunner to be based on java.nio.InputStreams rather than
  Channels. It was already using streams for everything and then just
  wrapping with a channel without provided any actual benefit.
- Fix issue where isAtEnd sometimes does not return a correct value. It
  now queries the underlying data stream to determine if there is more
  data or not, rather than relying on bitLimit which might not always be
  set.

DAFFODIL-934, DAFFODIL-931, DAFFODIL-1065, DAFFODIL-1565


stevedlawrence:
Note that the Travis CI build failure appears to be an out of memory error, which is fixed by bumping up the memory in PR #78.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -255,7 +255,7 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
 
   def reload(savedParser: File) = reload(new FileInputStream(savedParser).getChannel())
 
-  def reload(savedParser: DFDL.Input): DFDL.DataProcessor = {
+  def reload(savedParser: java.nio.channels.ReadableByteChannel): DFDL.DataProcessor = {
```
mbeckerle: Should this be deprecated, i.e., the new real API wants InputStream now right? I think the only reasons for wanting a Channel were to allow memory mapped I/O, which was a false requirement.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDecoder.scala
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * Decodes at most chars.remaining() characters in the chars CharBuffer. If
+   * bitPositions is provided, for each decoded character the bitPosition0b
+   * where the character decode operation finished is stored in the
+   * bitPositions LongBuffer. Upon return of the decode operation, the
+   * bitPosition0b of the InputSourceDataInputStream will be the end of the last
+   * successful character decode operation. Returns the number of successfully
+   * decode characters.
+   */
+  final def decode(dis: InputSourceDataInputStream, finfo: FormatInfo, chars: CharBuffer, bitPositions: LongBuffer = null): Int = {
```
mbeckerle: It's pretty clear from review here that the approach of implementing our own decoders, vs. all that complexity of dealing with Java decoders, is the correct trade off. These decoders are entirely straightforward, and the code using them. This is way simpler in total this way.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharset.scala
```diff
@@ -54,6 +51,7 @@ trait BitsCharset extends Serializable {
     case _ => false
   }
   def name: String
+  def aliases: Seq[String] = Nil
```
mbeckerle: See above at Line 40 - the comment needs to be updated. The motivation for BitsCharset is not the same anymore as it was. It's not about final methods, but the excessive complexity of dealing with Java charset and decoders/encoders.

I can't comment on a line that's not part of the diff, so put this comment here at the first Diff line after.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -0,0 +1,472 @@
...
+package org.apache.daffodil.io
+
+import java.nio.ByteBuffer
+
+import scala.collection.mutable.ArrayBuffer
+
+import org.apache.daffodil.exceptions.Assert
+
+
+abstract class InputSource {
```
mbeckerle: Add scaladoc for this class. It's pretty critical to the overall design, and the fact that it is NOT one of Java's or Scala's already defined abstractions is important and needs to be motivated.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -0,0 +1,472 @@
...
+
+  final def areDebugging: Boolean = _debugging
+
+  final def setDebugging(setting: Boolean): Unit = _debugging = setting
+}
+
+/**
+ * Implements the InputSource interface, reading data from a
+ * generic java.io.InputStream and storing the data in buckets of a defined
+ * size. Buckets are freed when no locks exist inside the bucket to minimize
```
mbeckerle: Mixed terminology "lock" vs. "reference count".  Add comment explaining this has nothing to do with synchronization locks. The term "lock" here just means we've incremented the ref-count so that the block will not be released until ref count goes to zero.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -0,0 +1,840 @@
...
+        // failed to skip all the necessary chars, reset the bit position back to
+        // where we started
+        setBitPos0b(startingBitPos)
+      }
+
+      skippedAllNChars
+    }
+  }
+
+  def lookingAt(matcher: java.util.regex.Matcher, finfo: FormatInfo): Boolean = {
```
mbeckerle: Did this code change much? Fundamental algorithm is still gradually enlarging the buffer until big enough to hold a match or not. 

I can''t see deltas because the diff tool isn't relating this to the prior existing code.

daffodil-io/src/main/scala/org/apache/daffodil/io/ByteBufferDataInputStream.scala
```diff
@@ -1,1270 +0,0 @@
-/*
```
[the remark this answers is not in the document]
stevedlawrence: This file is removed in this diff. It's all removals.

Regarding the two implements, there are two implementations of InputSource's in InputSource.scala. There's the BucketingInputSource which implements the bucketing algorithm, and a ByteBufferInputSource which just reads data out of a ByteBuffer. There's no bucketing there, so it should be more efficient if someone already has data in an array or byte buffer, The downside is that we cannot free up data from a ByteBuffer so it might waste memory. That's also limited to 2GB because of Java. The BucketingInputSource has a much higher limit of MaxInt * bucketSize.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDecoder.scala
stevedlawrence: Absolutely agreed. I was hesitant at first, but the decoders really weren't too difficult to write and the resulting code was sooo much simpler.

daffodil-lib/src/main/scala/org/apache/daffodil/util/BitOrder.scala
```diff
@@ -213,4 +213,60 @@ object Bits {
...
+  /**
+   * Round up the bitPosition to the nearest byte position.
+   *
+   * For example, say we want to read up to a bit position that is not a full
+   * byte. That will require reading the full byte, leaving the bytePosition
+   * past our bit position. This function can be used to determine where that
+   * byte position will be. Essentially, if bitPos0b is not a multiple of 8, we
+   * round up one byte.
+   */
+  def roundUpToByte(bitPos0b: Long): Long = {
```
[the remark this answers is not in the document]
stevedlawrence: Agreed. This method isn't really used for aligning (though I guess it could be), but the second name is much more clear about what it's doing. Fixed.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
stevedlawrence: The save/reload method weren't change, only the parse method, and that was necessary to support multiple calls to parse() using the same input. There isn't really the same motivaction here since it's just saving/reloading parsers. But it might be nice for consistency? I'll create a bug think about changing this.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
stevedlawrence: Fixed.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
stevedlawrence: Fixed.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
stevedlawrence: Not really. It's simplified some because it used to have logic about looping the fillCharBuffer since decode would stop at the first error, but now we just always replace things. There also was a lot of logic for determine lengths and that completely goes away since the decoder now keeps track of the bit position where each decoded char ends. Also I think some minor performance gains like allocation of a tuple no longer happens. But otherwise, the logic is virtually the same as before. It decodes some amount, tries a match, if that doesn't work, decode more, and repeat.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharset.scala
stevedlawrence: Fixed.

PR #79, opened 2018-07-09
TDMLRunner explicit roundTrip="twoPass" feature.

mbeckerle:
Added feature where tests are explicit about number of roundTrip tries.

I needed this feature to help with debugging the 100+ test failures I was getting in debugging the new separator/suppression implementation. But I did this change on a separate branch. 

This changes the TDML runner so that tests now must be explicit about how many passes.

This applies only to positive parse test cases. No changes for unparser tests or negative tests.

The roundTrip (and defaultRoundTrip) attributes now take these values:

"none" or "false" - no round trip

"simple" or "true" - parse - infoset must match, unparse - data must match original data exactly. This is the default behavior unless overridden with defaultRoundTrip property of test suite. 

"twoPass" - parse - infoset must match, unparse - data must NOT match original, parse - infoset must match. 

320+ tests required change from simple to twoPass.

DAFFODIL-1961

### DIFF
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
@@ -73,7 +73,7 @@
       <attribute name="suiteName" type="xs:token" use="optional"/>
       <attribute name="ID" type="xs:token" use="optional"/>
       <attribute name="description" type="xs:string" use="optional"/>
-      <attribute name="defaultRoundTrip" type="xs:boolean" use="optional"/>
+      <attribute name="defaultRoundTrip" type="tns:roundTripType" use="optional"/>
       <attribute name="defaultValidation" type="tns:validationType" use="optional"/>
       <attribute name="defaultConfig" type="xs:string" use="optional"/>
     </complexType>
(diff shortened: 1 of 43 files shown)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
```diff
@@ -151,12 +151,22 @@
...
     <attribute name="validation" type="tns:validationType" use="optional"/>
   </attributeGroup>
 
+  <simpleType name="roundTripType">
+    <restriction base="xs:token">
+      <enumeration value="false"/> <!-- means same as none -->
+      <enumeration value="none"/>
+      <enumeration value="true"/> <!-- means same as simple -->
+      <enumeration value="simple"/> <!--  parse, compare infoset, unparse - compare data -->
+      <enumeration value="twoPass"/> <!-- parse, unparse, reparse - compare infoset -->
```
stevedlawrence: Thoughts on "onePass" instead of "simple" just for consistency with "twoPass", and it's maybe a bit more clear what it's doing?

daffodil-tdml/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -77,12 +77,43 @@ import org.apache.daffodil.api.DaffodilTunables
...
+
+  /**
+   * Use to convert round trip default into enum value
+   */
+  def standardizeRoundTrip(enumStr: String): RoundTrip =
+    enumStr match {
+      case "false" | "none" => NoRoundTrip
+      case "true" | "simple" => SimpleRoundTrip
+      case "twoPass" => TwoPassRoundTrip
+    }
```
stevedlawrence: This feels like maybe it should be a companion object with an apply method that converts a string to one of those? Or perhaps something like in GeneratedCode that has builtin string to Enum values?

stevedlawrence:
Can you modify the commit message to include some of the information in the PR message? Alot of that would be useful to have, especially the descriptions of simple vs twoPass etc.
That would also be good to have on the [TDML page](https://daffodil.apache.org/tdml).

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
jadams-tresys: I agree, onePass is more clear.

mbeckerle: Agreed. This should be changed. Will make the documetnation of the feature cleaner too.

daffodil-tdml/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: If I did it that way I'd still have to override the stringToEnum methods to accomodate the backward compatibility for false and true string values. 

This usage is so simple I'm inclined to just leave it. It is true that it is not consistent in that our code is not doing ALL enums the same way, but this isn't the only example of inconsistency there, and this is one of the idiomatic ways to do enumerated objects in scala.

PR #80, opened 2018-07-13
added a layer transform that swaps bytes within an array of 32 bit words

marshalljmiller:
This is a layer transform that allows Daffodil to support parsing Link-16 J messages embedded within DIS PDUs, commonly referred to as SISO-J.  This implements https://issues.apache.org/jira/browse/DAFFODIL-1888.

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
@@ -358,4 +358,54 @@ a few lines of pointless text like this.""".replace("\n", " ")
     val areTracing = false
     TestUtils.testUnparsing(sch, infoset, data, areTracing)
   }
+
+  val le32BitData = Array[Byte](0x01,                   // BE MSBF
+                                0x43, 0x33, 0x33, 0x32, // fourbyteswap + LE LSBF (parsed right to left four bytes at a time)
+                                            0x55, 0x54, // fourbyteswap + LE LSBF
+                                0x67)                   // BE MSBF
+
+  val le32BitSchema =
+    SchemaUtils.dfdlTestSchema(
+      <dfdl:format ref="tns:GeneralFormat" bitOrder="leastSignificantBitFirst"
+                   byteOrder="littleEndian" alignmentUnits="bits" alignment="1"
+                   lengthKind="explicit" lengthUnits="bits"/>,
+      <xs:element name="e1" dfdl:lengthKind="implicit">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="s0" type="xs:hexBinary" dfdl:length="4" dfdl:byteOrder="bigEndian" dfdl:bitOrder="mostSignificantBitFirst"/>
+            <xs:element name="s1" type="xs:hexBinary" dfdl:length="4" dfdl:byteOrder="bigEndian" dfdl:bitOrder="mostSignificantBitFirst"/>
+            <xs:sequence dfdl:layerTransform="fourbyteswap" dfdl:layerLengthKind="explicit" dfdl:layerLengthUnits="bytes" dfdl:layerLength="6">
+              <xs:sequence>
+                <xs:element name="s2" type="xs:hexBinary" dfdl:length="4"/>
+                <xs:element name="s3" type="xs:hexBinary" dfdl:length="24"/>
+                <xs:element name="s4" type="xs:hexBinary" dfdl:length="8"/>
+                <xs:element name="s5" type="xs:hexBinary" dfdl:length="12"/>
+              </xs:sequence>
+            </xs:sequence>
+            <xs:element name="s6" type="xs:hexBinary" dfdl:length="4" dfdl:byteOrder="bigEndian" dfdl:bitOrder="mostSignificantBitFirst"/>
+            <xs:element name="s7" type="xs:hexBinary" dfdl:length="4" dfdl:byteOrder="bigEndian" dfdl:bitOrder="mostSignificantBitFirst"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>, elementFormDefault = "qualified")
+
+  @Test def testFourByteSwapLayer() {
+    val sch = le32BitSchema
+    val data = le32BitData
+    val infoset =
+      <e1 xmlns={ example }>
+        <s0>00</s0>
+        <s1>01</s1>
+        <s2>02</s2>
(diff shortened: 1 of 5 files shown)


daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
```diff
@@ -358,4 +358,65 @@ a few lines of pointless text like this.""".replace("\n", " ")
...
+                <xs:element name="s2" type="xs:hexBinary" dfdl:length="4"/>
+                <xs:element name="s3" type="xs:hexBinary" dfdl:length="24"/>
+                <xs:element name="s4" type="xs:hexBinary" dfdl:length="8"/>
+                <xs:element name="s5" type="xs:hexBinary" dfdl:length="4"/>
+              </xs:sequence>
+            </xs:sequence>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>, elementFormDefault = "qualified")
+  */
```
stevedlawrence: Can this commented out schema be removed or a test added to use it? Looks like it just a simpler version of the other schema? I wouldn't be against having both schemas tested.

daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
```diff
@@ -358,4 +358,65 @@ a few lines of pointless text like this.""".replace("\n", " ")
...
+
+  @Test def testFourByteSwapLayer() {
+    val sch = le32BitSchema
+    val data = le32BitData
+    val infoset = <e1 xmlns={ example }><s0>00</s0><s1>01</s1><s2>02</s2><s3>333333</s3><s4>44</s4><s5>0555</s5><s6>06</s6><s7>07</s7></e1>
```
stevedlawrence: Can you wrap and indent this XML? Just makes it easier to read tests. Scala is fine with multiline XML.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/ByteSwapTransformer.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  override def newInstance(maybeLayerCharsetEv: Maybe[LayerCharsetEv],
+    maybeLayerLengthKind: Maybe[LayerLengthKind],
+    maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
+    maybeLayerLengthUnits: Maybe[LayerLengthUnits],
+    maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
+    trd: TermRuntimeData): LayerTransformer = {
+
+    val xformer = new ByteSwapTransformer(wordsize, maybeLayerLengthInBytesEv.get)
+    xformer
```
stevedlawrence: It looks like this only supports layerLengthKind="explicit". Does it make sense to also implement implicit or boundarymark? If not, do we need to have schema definition checks to ensure the appropriate layer properties are provided? I don't see the checks for other transformers, so maybe we're missing those as well?

daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
marshalljmiller: I will remove this.

daffodil-core/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
marshalljmiller: Will do

marshalljmiller:
I will update the commit message so that it is formatted correctly and with more detail.  Should I amend the commit and force push it and then add follow up commits for the necessary code changes?

You are correct that the bug describes implementing the fourth combination of bit+byte order, but after discussing the problem with Mike on the list (https://lists.apache.org/thread.html/99bcd6fe3d75fbcaa70413ee643b7d3ce247260dc0224a78dc45bd23@%3Cdev.daffodil.apache.org%3E) we both concluded that this approach would be simpler and would achieve the desired support.

stevedlawrence:
We've found that amend and --force commits has caused old review comments to disappear from GitHub sometimes, so we tend to avoid it. The recommending way to add changes to a review is to make new commits with --fixup or --squash options. --squash with --allow-empty is a good way to create a new commit message if no actual changes are needed.

Once you get two +1's, you can rebase -i and squash/fixup everything into a single commit, do a force push, and then we'll merge the change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/ByteSwapTransformer.scala
mbeckerle: I think for many layer transforms only certain layerLengthKinds make sense. So code should reject the other layerLengthKinds as incompatible.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/ByteSwapTransformer.scala
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.processors.TermRuntimeData
+import org.apache.daffodil.processors.LayerLengthInBytesEv
+import org.apache.daffodil.processors.LayerBoundaryMarkEv
+import org.apache.daffodil.processors.LayerCharsetEv
+import org.apache.daffodil.processors.parsers.PState
+import org.apache.daffodil.io.ExplicitLengthLimitingStream
+import org.apache.daffodil.processors.unparsers.UState
+
+class ByteSwapInputStream(wordsize: Int, jis: InputStream)
```
mbeckerle: Add scaladoc. What exactly does fourByteSwap mean? I.e., if bytes are 1234 does this create 3412, or 4321 or ???

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/ByteSwapTransformer.scala
mbeckerle: Many layer transforms only 1 or a subset of the layer lengthKinds make any sense. See our limited doc on layers on the wiki here: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75979671

(which would be good to update also once this change is merged.)

marshalljmiller:
I squashed my commits and rebased on master.

stevedlawrence:
Merged. Thanks for you contribution!

PR #84, opened 2018-07-18
Daffodil 1923 - Allow use of the same character for escapeBlockEnd and escapeEscapeCharacter

jadams-tresys:
This commit allows the same character(s) to be used for the escapeBlockEnd and escapeEscapeCharacter.

Prior to this the code would first check for the escapeBlockEnd without checking to see if it was also the escapeEscapeCharacter, and would think that it hit the end of the escapeBlock when it was just the escapeEscapeCharacter.

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
@@ -214,9 +214,17 @@ class StartStateEscapeBlock(states: => ArrayBuffer[State], val blockEnd: DFADeli
   val stateName: String = "StartState"
 
   val rules = ArrayBuffer(
-    Rule { (r: Registers) => couldBeFirstChar(r.data0, blockEnd) } { (r: Registers) => r.status = StateKind.Paused },
-    Rule { (r: Registers) => { EEC.isDefined && (r.data0 == EEC.get) } } { (r: Registers) => r.nextState = EECState },
-    Rule { (r: Registers) => { r.data0 == DFA.EndOfDataChar } } { (r: Registers) => r.nextState = DFA.EndOfData },
+    Rule { (r: Registers) => couldBeFirstChar(r.data0, blockEnd)} { (r: Registers) =>
+      {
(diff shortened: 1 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
```diff
@@ -501,6 +503,11 @@ class EECStateBlock(states: => ArrayBuffer[State], blockEnd: DFADelimiter, val E
         r.nextState = StartState
       }
     },
+    Rule { (r: Registers) => couldBeFirstChar(r.data0, blockEnd) } {
+      (r: Registers) => {
+        r.status = StateKind.Paused //PTERMState
+      }
+    },
```
stevedlawrence: Can you clarify why this change is needed. I can't say if it's write or wrong, but seems odd that this changes it to check data1 first before data0. Does this instead need a change similar to the above?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
```diff
@@ -488,9 +493,6 @@ class EECStateBlock(states: => ArrayBuffer[State], blockEnd: DFADelimiter, val E
     //
     // We've already encountered EEC as data0 here
     //
-    Rule { (r: Registers) => couldBeFirstChar(r.data0, blockEnd) } {
```
mbeckerle: The comment above on lines 489-492 is just misleading now. I suggest deleting it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
mbeckerle: Can we add a test that we know drives these rules, so we can see if this data1 first thing is a bug? (I think it is.)

jadams-tresys:
Closing pull request.  I think I confused github by deleting the branch I originally did this work on and then creating a new branch with the same name. Will open new pull request.

PR #86, opened 2018-07-25
Adding value of 'bcd' to binaryCalendarRep

efinnegan:
Adds parsing and unparsing capability for binary representation of
xs:date, xs:time, and xs:dateTime with a binaryCalendarRep of 'bcd',
including known, runtime determined, and delimited lengths of data.

This uses a combination of the existing BCD parser/unparser and
text calendar parser/unparser, with some added checks for binary
calendars.

DAFFODIL-1941

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
@@ -26,6 +26,7 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.io.DataOutputStream
 import org.apache.daffodil.io.FormatInfo
 import org.apache.daffodil.processors.Evaluatable
+import org.apache.daffodil.dpath.NodeInfo
 
 trait PackedBinaryConversion {
   def fromBigInteger(bigInt: JBigInteger, nBits: Int): Array[Byte]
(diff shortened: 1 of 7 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
```diff
@@ -45,7 +46,16 @@ abstract class PackedBinaryBaseUnparser(
   override def unparse(state: UState): Unit = {
     val nBits = getBitLength(state)
     val node = state.currentInfosetNode.asSimple
-    val value = node.dataValue.asInstanceOf[JNumber]
+
+    // If this is unparsing a binary calendar use node.dataValueAsString and create a JBigDecimal
+    //   so it can be cast to a JNumber below.
+    val nodeValue =
+      node.erd.optPrimType.get match {
+      case _: NodeInfo.Date.Kind | _: NodeInfo.DateTime.Kind | _: NodeInfo.Time.Kind => new JBigDecimal(node.dataValueAsString)
```
stevedlawrence: If it's possible to switch to the BCDInteger* parsers, this can then be a JBigInteger

Also, I don't think you don't need the _: in the cases. We're probably not very consistent about it, but not having it is preferred.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/PackedBinaryUnparserTraits.scala
```diff
@@ -45,7 +46,16 @@ abstract class PackedBinaryBaseUnparser(
   override def unparse(state: UState): Unit = {
     val nBits = getBitLength(state)
     val node = state.currentInfosetNode.asSimple
-    val value = node.dataValue.asInstanceOf[JNumber]
+
+    // If this is unparsing a binary calendar use node.dataValueAsString and create a JBigDecimal
+    //   so it can be cast to a JNumber below.
```
stevedlawrence: It would be good to explain *why* we need to use dataValueAsString rather than dataValue (e.g. packed date time uses the convert combinator which sets the string value of the calendar--if we used dataValue we would get the Calendar value rather than that string. On the otherhand, with packed numbers, dataValue is already a number so we just want to use that, etc.).

Based on the code, it's clear that unparsing a packed binary calendar uses dataValueAsString, but it's not immediately clear why that's necessary.

PR #92, opened 2018-08-09
Update dependencies to their latest versions

stevedlawrence:
Updates that required changes:

- ICU4J
  - License has switched to the Unicode license, LICENSE and NOTICE
    updated.
  - No longer requires that a positive number pattern exists. That is
    required by the DFDL specification, so the check is done manually
    (i.e. the pattern cannot start with a semi-colon)
  - ICU now correctly checks roundUnnecsssary and throws an exception
    if rounding is needed on unparsing--before it would just round with
    roundHalfEven. We now catch this exception and return an
    UnparseError. To maintain backwards compatibility, the
    textNumberRoundingMode property in built-in-formats.xsd and
    DFDLGeneralFormat.dfdl.xsd is changed from roundUnnecessary to
    roundHalfEven.
  - ICU's DecimalFormat.parse() method now returns an ICU BigDecimal
    instead of a Java BigInt/BigDecimal.
  - There is a bug in ICU that if a text number pattern uses scientific
    notation but the number is +Inf, -Inf, or NaN, ICU will add the
    exponent (e.g. INFx10^0). This is wrong, so this now manually
    detects if the number is one of those values and outputs the
    appropriate inf/nanRep along with pattern prefix/suffix instead of
    using the DecimalFormat.format method.
  - Lax number parsing no longer trims whitespace, so we now manually
    perform the trim if dfdl:checkPolicy is lax.
  - Various tests are changed or moved out of scala-debug due to ICU bug
    fixes. For example ICU now handles MMM (abbreviated month) for a
    Russian calendar language. Some tests are changed just due to
    different error messages. Some just need to use twoPass round trip.
  - A few tests now fail due to new bugs in ICU. The failing features are:
      1) ICU no longer trims padding characters after a suffix,
         resulting in failures when trying to parse such numbers leading
         to ParseErrors. Padding before prefix, after prefix, and
         before suffix still work as expected.
      2) ICU Fails to parse scientific numbers where the exponent rep
         is the empty string, e.g. 12.34+2 is now invalid.
    Tests (total of 3) that require these features are disabled and
    moved to scala-debug.

- Xerces
  - Some error messages changed to now expand namespace prefixes or
    display slightly different information, requiring updates to some
    tests.

- JLine
  - Deprecated shutdown() in favor of close()

- Jackson
  - Deprecated JsonStringReader.getInstance() replaced with
    BufferRecyclers.getJsonStringEncoder()

DAFFODIL-1973, DAFFODIL-851, DAFFODIL-845

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextNumberUnparser.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextNumberUnparser.scala
@@ -17,6 +17,9 @@
 
 package org.apache.daffodil.processors.unparsers
 
+import java.lang.{ Double => JDouble }
+import java.lang.{ Float => JFloat }
+
 import org.apache.daffodil.processors._
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe._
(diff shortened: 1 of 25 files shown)


daffodil-test/src/test/scala-debug/org/apache/daffodil/section13/text_number_props/TestTextNumberPropsDebug.scala
```diff
@@ -34,34 +34,22 @@ class TestTextNumberPropsDebug {
...
   @Test def test_infnanCaseInsensitive() { runner.runOneTest("infnanCaseInsensitive") }
   @Test def test_expCaseSensitive() { runner.runOneTest("expCaseSensitive") }
+
+  // DAFFODIL-XXXX
+  @Test def test_expEmptyString() { runner.runOneTest("expEmptyString") }
+  @Test def test_expEmptyString2() { runner.runOneTest("expEmptyString2") }
+
+  // DAFFODIL-XXXX
+  @Test def test_textNumberPattern_padding09() { runner.runOneTest("textNumberPattern_padding09") }
 }
```
stevedlawrence: Need to add JIRA tickets for these new failing tests.

project/Dependencies.scala
```diff
@@ -22,30 +22,30 @@ object Dependencies {
   lazy val common = core ++ infoset ++ test
 
   lazy val core = Seq(
-    "org.scala-lang.modules" %% "scala-xml" % "1.0.6",
-    "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4",
-    "com.ibm.icu" % "icu4j" % "51.1", // new versions avail. 58.1 requires code changes
-    "xerces" % "xercesImpl" % "2.10.0",
+    "org.scala-lang.modules" %% "scala-xml" % "1.1.0",
+    "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.1",
+    "com.ibm.icu" % "icu4j" % "62.1",
```
mbeckerle: I suggest listing the JIRA tickets for the bugs, or the scala-debug test names right here in comments.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextNumberUnparser.scala
```diff
@@ -69,22 +72,36 @@ case class ConvertTextNumberUnparser[S](
...
+    val df = nff.getNumFormat(state).get
+    val dfs = df.getDecimalFormatSymbols
+
+    val strRep = value match {
+      case n: Number if n == 0 && zeroRep.isDefined => zeroRep.get
+      // We need to special case infinity and NaN because ICU4J has a bug and
+      // will add an exponent to inf/nan (e.g. INFx10^0) if defined in the
+      // pattern, which we don't want. We need to manually output the inf/nan
+      // rep plus the prefix and suffix
+      case f: JFloat if f.isInfinite =>
```
mbeckerle: In the past when we've had workarounds of bugs in 3rd party libraries I've created a unit test that will fail if the bug is fixed, so that we know we can take out the workaround. Anytime we're working around or depending on a strange behavior, a specific unit test is useful to detect if the behavior changes.

daffodil-test/src/test/scala-debug/org/apache/daffodil/section13/text_number_props/TestTextNumberPropsDebug.scala
```diff
@@ -34,34 +34,22 @@ class TestTextNumberPropsDebug {
...
+  // DFDL-861
   @Test def test_infnanCaseInsensitive() { runner.runOneTest("infnanCaseInsensitive") }
   @Test def test_expCaseSensitive() { runner.runOneTest("expCaseSensitive") }
+
+  // DAFFODIL-XXXX
+  @Test def test_expEmptyString() { runner.runOneTest("expEmptyString") }
+  @Test def test_expEmptyString2() { runner.runOneTest("expEmptyString2") }
+
+  // DAFFODIL-XXXX
+  @Test def test_textNumberPattern_padding09() { runner.runOneTest("textNumberPattern_padding09") }
```
stevedlawrence: When adding a test for the bug in ICU that caused this test to fail, I discovered the actual problem is that ICU does not update the ParsePosition when there is suffix padding. For example, if the string was "123 SUFFIX______", and the text number pattern defines an after-suffix pad character of _, the ParsePosition points to "X" rather than the end of the string. This means that we can manually check to see if all character after ParsePosition are the pad character if the pad position is after suffix to see if a string is valid, and thus maintain the original behavior. Ill still add a test to see if ICU ever fixes this bug, but this test no longer fails with the mentioned fix.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextNumberUnparser.scala
jadams-tresys: I'll have to remember that idea for other projects in the future!

PR #96, opened 2018-08-13
Fix eclipse paths.

mbeckerle:
Also moved UpdateEclipseClasspaths tool to daffodil-propgen, which has
no dependencies on libraries except for scala's XML library.

Propgen might want to be renamed someday, but it is the right module for
various sorts of code generators and tools that are used to setup the
rest of the build.

DAFFODIL-1982


mbeckerle:
I will create a ticket. I noticed recently that adding a new tunable option to the system requires redundant editing in several files. All could be easily code generated from a list that just contains the name of the tunable option, its type, and default value. That's another thing we might want to do from propgen.

PR #98, opened 2018-08-13
Eliminates backpointers from RuntimeData structures.

mbeckerle:
This is a step toward eliminating the double-linked non-sharing behavior
in the schema compiler, because maintaining these double-linked pointers
for the runtime was forcing them to exist in the schema compiler also.

DAFFODIL-1444

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
@@ -169,16 +169,14 @@ abstract class RepeatingChildUnparser(
                 // System.err.println("Stopping occurrences(1) of %s due to event %s".format(erd.namedQName, ev))
                 false // event not a start for this element
               }
-            } else if (ev.isEnd && ev.isComplex) {
-              //ok. We've peeked ahead and found the end of a complex element.
+            } else {
+              Assert.invariant(ev.isEnd)
+              // could be end of simple elemnet - we handle on the start event. Nothing to do.
+              // or could be end of complex event, i.e., we've peeked ahead and found the end of a complex element.
               // It has to be the complex element that ultimately encloses this sequence.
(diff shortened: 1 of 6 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
```diff
@@ -170,23 +170,12 @@ abstract class RepeatingChildUnparser(
                 false // event not a start for this element
               }
             } else if (ev.isEnd && ev.isComplex) {
-              val c = ev.asComplex
-              //ok. We've peeked ahead and found the end of the complex element
+              //ok. We've peeked ahead and found the end of a complex element.
+              // It has to be the complex element that ultimately encloses this sequence.
+              // Though that's not a unique element given that this sequence could be inside
+              // a global group definition that is reused in muliple places.
               //that this sequence is the model group of.
```
stevedlawrence: Should this last line of the comment be removed?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
```diff
@@ -170,23 +170,12 @@ abstract class RepeatingChildUnparser(
...
-                  false
-                }
-                case _ =>
-                  Assert.invariantFailed("Not end element for this sequence's containing element. Event %s, optParentRD %s.".format(
-                    ev, optParentRD))
-              }
+              false
             } else {
               // end of simple element. We shouldUnparse on the start event.
               // System.err.println("Stopping occurrences(3) of %s due to event %s".format(erd.namedQName, ev))
```
stevedlawrence: Both the (ev.isEnd && ev.isComplex) block and this else block now just return false. Is it worth it to combine them into a single else block, maybe with a comment that it could be either simple or complex but it doesn't matter?

PR #99, opened 2018-08-13
Fix missing checks for minOccurs after zero occurrences.

mbeckerle:
Also verify that assertion failure no longer occurs.

DAFFODIL-1964


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
```diff
@@ -254,7 +254,7 @@ abstract class RepeatingChildUnparser(
     //
     val ock = erd.maybeOccursCountKind.get
     // System.err.println("Checking for events consistent with Array index:\n ==> Array Index Stack is:" + state.arrayIndexStack)
-    val arrPos = state.arrayPos - 1 // because we advance after each unparse so this is the position of the next (non-existing) occurrence
+    // val arrPos = state.arrayPos - 1 // because we advance after each unparse so this is the position of the next (non-existing) occurrence
```
stevedlawrence: Remove comment.

daffodil-test/src/test/scala/org/apache/daffodil/section16/array_optional_elem/TestArrayOptionalElemNew.scala
```diff
@@ -18,40 +18,28 @@
 package org.apache.daffodil.section16.array_optional_elem
 
 import org.apache.daffodil.util._
-import org.apache.daffodil.tdml.DFDLTestSuite
+import org.apache.daffodil.tdml.Runner
 import org.junit.AfterClass
+import org.junit.Test
 
 object TestArrayOptionalElemNew {
```
stevedlawrence: We don't really use the scala-new thing anymore. I think it's reasonable to put this in TestArrayOptionalElem.scala. These *New.scala files should probably be deleted with their content moved to the not New files at some point. Maybe delete this file while you're at it?

PR #100, opened 2018-08-14
Moved pushDiscriminator to fix separators bug.

mbeckerle:
Also found some unused code and deleted it. Old methods not used
anywmore.

DAFFODIL-1984

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
@@ -299,6 +299,15 @@ final class OrderedSeparatedSequenceParser(
 
       val bitPosBeforeSeparator = pstate.bitPos0b
 
+      //
+      // Note: Performance. All this conditional branch logic could in principle be hoisted
+      // out and decided at schema compile time, as what is being tested here is statically known.
+      //
+      // However, that's trading a few conditional branches for a virtual function call, and that
+      // might or might not be an improvement worth making. Not worth bothering with unless
+      // some profiling makes it look like this could help.
+      //
+
       // parse prefix sep if any
       val prefixSepSuccessful =
         if (spos eq SeparatorPosition.Prefix) {
@@ -341,8 +350,6 @@ final class OrderedSeparatedSequenceParser(
           // now we parse the child
           //
           val prevBitPosBeforeChild = pstate.bitPos0b
-
-          pstate.pushDiscriminator
 
           if (pstate.dataProc.isDefined) pstate.dataProc.get.beforeRepetition(pstate, this)
 
@@ -496,7 +503,6 @@ final class OrderedSeparatedSequenceParser(
                 result
               } // end if child success/fail
             } // end val res
-            pstate.popDiscriminator
             res
           } // end if postfix
         } // end if infix
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
@@ -118,13 +118,6 @@ abstract class OrderedSequenceParserBase(
               // different, or OCK=parsed.
               //
 
-              //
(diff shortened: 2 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
```diff
@@ -299,6 +299,18 @@ final class OrderedSeparatedSequenceParser(
 
       val bitPosBeforeSeparator = pstate.bitPos0b
 
+      pstate.pushDiscriminator
```
stevedlawrence: If we pushing the discriminator here, it means that if a separator parser fails and processFailedSeparatorsWithPoU is called, the condition that checks pstate.discriminator == true will never pass and Failed_WithDiscriminatorSet will never be used (at least because of a missing separator), I think? So it seems like that condition is never going to be hit?

If so, maybe the right fix is to put the pushDiscriminator back where it was, and just remove the pstate.discriminitor == true check so that processFailedSeparatorsWithPoU always returns Failed_SpeculativeParse if a separator parser fails? Or is there some more complexity related to separators and discriminators that is missing?

mbeckerle: The pushD gives a clean false dis criminaltor at start. The only way that disc is set is when we parse The element Itself - which happens after the infix sep as you pointed out - so there can be a disc set after a postfix sep fails but not prefix or infix.
I am considering moving the push/ pop into the base class method so it will be centralized and not even an issue here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -123,7 +123,7 @@ abstract class OrderedSequenceParserBase(
               // This is the same object as the priorState before the first
               // occurrence.
               //
-              var beforeArrayState: PState.Mark = null
+              // var beforeArrayState: PState.Mark = null
```
jadams-tresys: Any reason to keep this commented out line?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -231,7 +225,6 @@ abstract class OrderedSequenceParserBase(
                           // discarded.
                           //
                           if (!isFirstIteration) pstate.discard(priorState)
```
stevedlawrence: Should this just always discard or reset to priorState?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -155,8 +148,7 @@ abstract class OrderedSequenceParserBase(
                 //
                 val priorState =
                   if (isFirstIteration) {
-                    beforeArrayState = pstate.mark("before all occurrences")
-                    beforeArrayState
+                    pstate.mark("before all occurrences")
                   } else {
                     pstate.mark("before second/subsequent occurrence")
                   }
```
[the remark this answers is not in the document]
mbeckerle: The messages stored in the pstate.mark are different, and that's useful for debugging.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
mbeckerle: oops. yes I think so.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -231,7 +225,6 @@ abstract class OrderedSequenceParserBase(
                           // discarded.
                           //
                           if (!isFirstIteration) pstate.discard(priorState)
-                          pstate.reset(beforeArrayState)
```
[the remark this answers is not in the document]
mbeckerle: removed.

PR #102, opened 2018-08-15
Added tests that exercise ambiguous separator/terminator case.

mbeckerle:
Improved stack-empty checking at end of parse, so it tests the
discriminator stack also.

DAFFODIL-1984

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
@@ -125,14 +125,20 @@ class MPState private () {
 
   val escapeSchemeEVCache = new MStackOfMaybe[EscapeSchemeParserHelper]
 
-  //var wasAnyArrayElementNonZeroLength = false
-  //var wasLastArrayElementZeroLength = true
-
   private def init {
     arrayIndexStack.push(1L)
     groupIndexStack.push(1L)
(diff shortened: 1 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -323,6 +329,26 @@ final class PState private (
...
+   * a normal parse. I.e., stacks have been popped back to their original state,
+   * pools - all items returned, etc.
+   *
+   * If for some reason parsing ends with a throw (not supposed to, but just if)
+   * then all bets are off, so this must be called ONLY on a normal return from parse call.
+   * That is, the parse can succeed or fail with diagnostics, but it must have returned normally.
+   */
+  def verifyFinalState(wasThrow: Boolean): Unit = {
+    if (!wasThrow) {
+      markPool.finalCheck
```
stevedlawrence: Should the markPool check should be done regardless of if something was thrown or not. Some throws are expected and allowed (e.g Runtime SDE) and we need to make sure the markPools are cleaned up appropriately in this case so successive parses don't fail. The other ones like stacks are less important since the PState is thrown away, but mark pool persist between parses and leaks cause real issues.

mbeckerle: Concur. Mark pool check should happen either way.

PR #105, opened 2018-08-15
Modified general format to have calendarTimeZone="" not UTC.

mbeckerle:
Updated tests to not expect "+00:00" on end of every date/time/datetime.

DAFFODIL-1930


mbeckerle:
Note that sbt test, sbt it:test all pass.
All DFDLSchemas pass, except quasiXML which required minor tweak.
All FOUO schemas pass.

PR #113, opened 2018-08-20
Clean up scaladoc

stevedlawrence:
Scaladoc does not do a good job at making links to methods pretty. So
where we have links, specify a link text that is much shorter and easier
to read than what scaladoc generates.

DAFFODIL-1992

### DIFF
--- a/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
+++ b/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
@@ -109,10 +109,10 @@ class Compiler private[sapi] () {
    * Compile DFDL schema file into a [[ProcessorFactory]]
    *
    * To allow jar-file packaging, (where schema files might be part of a jar),
-   * it is recommended to use [[Compiler#compileSource]] instead.
+   * it is recommended to use [[Compiler.compileSource]] instead.
    *
    * @param schemaFile DFDL schema file used to create a [[ProcessorFactory]].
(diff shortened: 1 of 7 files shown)


daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -138,11 +138,11 @@ class Compiler private[sapi] () {
    * Reload a saved parser from a file
    *
    * To allow jar-file packaging, (where the savedParser might be part of a jar),
-   * it is recommended to use the other version of [[Compiler#reload(savedParser:java\.nio\.channels\.ReadableByteChannel)*]] where the argument is
+   * it is recommended to use the other version of [[Compiler.reload(savedParser:java\.nio\.channels\.ReadableByteChannel)*]] where the argument is
```
mbeckerle: Honestly, I think these links that actually include the parameter types are pointless. What you want is a link to the right method, but there's no need for the method args to be displayed in the link text. Especially if the types are going to have their package names spelled out. It's just unreadable. It's much better to say "the other version of Compiler.reload" in prose than to have the method signature redundantly expressed here.

stevedlawrence: Agreed. In order to have a link, scala requires parameters if there is ambiguity due to method overloading. However, one can provide an alternative text so that the parameters aren't shown. I'll go through all the scala links to make sure we're not showing excessive information. Javadoc seems to do a better job and making readable docs without much effort.

PR #116, opened 2018-08-27
Fix comparison of calendars that do not have timezones

stevedlawrence:
The assertions was just incorrect. It is perfectly valid to call the
orderIgnoreTimeZone function where one calendar has a timezone and the
other does not, as long as they have been normalized/modified according
to the W3C spec before being called. Remove this invarient, and add a
couple extra just to make sure we are modifying calendar appropriately.
And move test out of scala debug.

DAFFODIL-1986

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/calendar/DFDLCalendar.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/calendar/DFDLCalendar.scala
@@ -24,19 +24,22 @@ import com.ibm.icu.util.TimeZone;
 import java.math.{ BigDecimal => JBigDecimal }
 
 import org.apache.daffodil.exceptions.Assert
-import scala.collection.mutable.ArraySeq
 import org.apache.daffodil.processors.parsers.TextCalendarConstants
 
 object DFDLCalendarOrder extends Enumeration {
   type DFDLCalendarOrder = Value
   val P_LESS_THAN_Q, P_GREATER_THAN_Q, P_EQUAL_Q, P_NOT_EQUAL_Q = Value
+
+  val fieldsForComparison = Array(
+    Calendar.EXTENDED_YEAR,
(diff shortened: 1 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/calendar/DFDLCalendar.scala
```diff
@@ -121,7 +121,8 @@ trait OrderedCalendar { self: DFDLCalendar =>
    * can be determined.
    */
   private def orderIgnoreTimeZone(p: DFDLDateTime, q: DFDLDateTime): DFDLCalendarOrder = {
-    Assert.invariant((p.hasTimeZone && q.hasTimeZone) || (!p.hasTimeZone && !q.hasTimeZone))
+    Assert.invariant(!p.hasTimeZone || p.calendar.getTimeZone == TimeZone.GMT_ZONE)
```
mbeckerle: The scaladoc here is not consistent with what this method does anymore. It describes a contract that P and Q must match as to whether they have, or do not have timezone. 

In what sense does this "ignore timezone" anyway? Seems to me it treats TZ like any other field, with the exception of the two assert.invariant calls.  Is this method's name actually indicative of it's behavior?

stevedlawrence: When doing the comparison of two dateTimes, the only fields that are inspected are year, month, day, hour, minute, and second. So in that sense the timezone is ignored. Technically, it is not completely ignored though, since the dateTime is normalized to UTC, so a dateTime with a non-UTC TZ will have its hours/minutes modified, which are compared.

So this function name ism't completely innacurate. That said, something like orderCompareFields might be more specific about what it's actually doing, since all it's really doing is looping over a subset of the fields of a calendar and comparing them. I'll update the name and comment to be accurate according to what the method does.

PR #117, opened 2018-08-27
Allow DFDL expressions in the message attribute of asserts/discrimina…

stevedlawrence:
…tors

- Converts the message attribute to a compiled expression that evaluates
  to a String at runtime. Creates a constant expression if no message is
  provided that is just the expression followed by "failed".
- Refactor the Assert/DiscriminatorPatternParsers into a single Parser.
  They are virtually the same except the discrimnator pattern parsers
  sets the the discriminator to true on match. This is similar to how
  AssertExpressionEvaluationParser is for both assert/discriminators
  that use tests instead of patterns.

DAFFODIL-1965

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
@@ -17,83 +17,71 @@
 
 package org.apache.daffodil.processors.parsers
 
+import java.util.regex.Matcher
+
+import org.apache.daffodil.dsom.CompiledExpression
+import org.apache.daffodil.dsom.SchemaDefinitionDiagnosticBase
 import org.apache.daffodil.processors._
 import org.apache.daffodil.util.LogLevel
-import java.util.regex.Matcher
 import org.apache.daffodil.util.OnStack
 
-abstract class AssertPatternParserBase(
-  eName: String,
-  kindString: String,
+trait AssertMessageEvaluationMixin {
+  def messageExpr: CompiledExpression[AnyRef]
+  def discrim: Boolean
+
+  def getAssertFailureMessage(state: PState): String = {
+    val message =
+      try {
+        messageExpr.evaluate(state).asInstanceOf[String]
+      } catch {
+        // Note that we intentionally catch an SDE here. This effectively
+        // converts an SDE (which is usually considered fatal) to a ParseError
+        // (which isn't fatal). But section 7.3.1 of the DFDL specification
+        // states that both types of errors should be caught and replaced with
+        // a recoverable error and an implementation defined replacement
+        // message should be provided instead.
+        case e @ (_: ParseError | _: SchemaDefinitionDiagnosticBase) => {
+          val kindString = if (discrim) "Discriminator" else "Assertion"
+          kindString + " message expression evaluation failed: " + e.getMessage
+        }
+      }
+    message
+  }
+}
+
+class AssertPatternParser(
   override val context: TermRuntimeData,
+  override val discrim: Boolean,
   testPattern: String,
-  message: String)
-  extends PrimParser {
+  override val messageExpr: CompiledExpression[AnyRef])
+  extends PrimParser
+  with AssertMessageEvaluationMixin {
   override lazy val runtimeDependencies = Vector()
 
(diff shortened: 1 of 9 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
```diff
@@ -17,83 +17,54 @@
...
+      if (!isMatch) {
+        val message =
+          try {
+            messageExpr.evaluate(start).asInstanceOf[String]
+          } catch {
+            case e @ (_: ParseError | _: SchemaDefinitionDiagnosticBase) => {
+              kindString + " message expression evaluation failed: " + e.getMessage
+            }
+          }
+        val diag = new AssertionFailed(context.schemaFileLocation, start, message)
```
mbeckerle: So, if an assert has a message expression which fails with a Schema definition error, this will be converted into an assertion failed which is a parse error. Is that right? Maybe this shouldn't catch schema definition errors at all, since they're fatal.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -140,7 +141,16 @@ class AssertExpressionEvaluationParser(
       if (testResult) {
         start.setDiscriminator(discrim)
       } else {
-        // The assertion failed. Prepare a failure message etc. in case backtracking ultimately fails from here.
+        val msg =
+          try {
+            msgExpr.evaluate(start)
+          } catch {
+            case e @ (_: ParseError | _: SchemaDefinitionDiagnosticBase) => {
```
mbeckerle: Same issue here. Should this be catching SDE?

Also, can't this code, which is common to both Pattern and Expression based Asserts, be factored into a base trait?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
stevedlawrence: The last paragraph of section 7.3.1 says:
>If a processing error or schema definition error occurs while evaluating the message expression, a recoverable error is issued to record this error (containing implementation-dependent content), then processing of the assert continues as if there was no problem and in a manner consistent with the failureType property, but using an implementation-dependent substitute message.
So I think the correct behavior is to mask the SDE and still create an assertion/parse error, just with a implementation defined message. Which in this case is message "expression failed".

mbeckerle: Wow. Ok, then that's worth specifically citing in the code in a comment, because normally I'd for sure expect masking an SDE to be a mistake.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
```diff
@@ -25,17 +25,41 @@ import org.apache.daffodil.processors._
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.OnStack
```
mbeckerle: This is great. I think the refactor below is totally worth it. It's only a few lines of code, being factored out, but there are other reasons why it's worth it. 

It makes it clear that the behavior is intended to be exactly the same between the two uses, whereas otherwise you need to have comments pointing to/from the other piece of code saying "keep these consistent", and I hate that because my IDE will jump to getAssertFailureMessage in the mixin for me, but the comments I have to navigate around myself with grep/search, and the two blocks of code can drift, and if by accident somebody fixes one and not the other, then I have to wonder if the two were intended to diverge, and the comments not fixed, or they were intended to be the same, but just not updated simultaneously.  In general, I hate doing in comments what I can do by way of code constructs. A trait is a pretty easy way to share even just a single small method. 

I also think the method name getAssertFailureMessage() adds value to the code. Makes it more readable.

stevedlawrence: Agreed, it's only a few lines but has a lot of clear benefits.

PR #120, opened 2018-09-12
Support more versions of Java

stevedlawrence:
- With the recent switch to custom decoders, we no longer are affected
  by the Java 7 decoder bug--we can now support Java 7, so remove those
  checks that required Java 8 or higher. Those checks also caused
  exceptions on older versions of Scala mixed with newer versions of
  Java, so this removes that problem as well.
- Remove the use of deprecated functions and remove some function
  ambiguities (e.g. .limit vs .limit()), enabling support for Java 9+
- Modify TravisCI configuration to build/test on all versions of Java 7
  and higher.
- Bump down the TravisCI JVM memory a bit--newer versions of Java aren't
  happy with us using so much of the available memory.
- Split the different sbt build/test steps into separate commands. Not
  only does this help with total memory usage, but since TravisCI
  executes all script commands even if previous ones fail, it ensures
  that all test/validation commands are executed, so we no longer just
  see the first failure. The only time we don't want to execute all the
  scripts is if we fail to compile, so we call travis_terminate in that
  case.
- Two tests now fail due to features added Java 9+'s regular expression
  language. These regular expressions are not valid according to the
  DFDL spec due to their use of advanced regular expression features,
  but they are now allowed since new versions of Java support them. Move
  tests to debug.
- Minor fixes to java/scaladoc found by newer versions of javadoc

DAFFODIL-1945


daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -534,16 +534,16 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
...
     val nBitsWritten = nBytesWritten * 8
     if (nBytesWritten < nWholeBytes) {
       nBytesWritten * 8
     } else {
       val isFragWritten =
         if (nFragBits > 0) {
-          bb.limit(bb.limit + 1)
-          var fragByte: Long = Bits.asUnsignedByte(bb.get(bb.limit - 1))
+          bb.limit(bb.limit() + 1)
+          var fragByte: Long = Bits.asUnsignedByte(bb.get(bb.limit() - 1))
```
mbeckerle: So, to avoid re-injecting things like this, I need to change IDE settings or environment vars or something to somehow tell scala that it is expecting a Java 10 underneath not Java 1.8?

stevedlawrence: Probably. Fedora allows multiple installations of different Java versions and uses the "alternatives" command to switch between them. I'm not sure if eclipse figures out where java is using alternatives (or probably something similar on other OSes) or if it uses some other mechanism. For sbt, alternatives is sufficient to test different java versions.

Though, TravisCI is now configured to build on all java versions, so you could just rely on the results of that to know if you missed anything or did something that wasn't compatible with different java versions. The limit() and position() methods are kindof a special case since the setters and getters have the same name, just one takes a parameter. Most java methods have separate get* and set* methods instead so this isn't an issue.

mbeckerle: Ok, no special steps needed then since TravisCI will pick this up.

stevedlawrence:
Realized the README.md file mentions a JDK 1.8 requirement, that should be changed to JDK 7+ as part of this commit.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
```diff
@@ -130,16 +130,10 @@ case class DaffodilTunables(
...
-  // Daffodil throws an exception when it detects that Daffodil is not running
-  // under Java 8 or has this decoder bug. However, there are some cases where
-  // a user has no choice but to run on Java 7. Setting this tunable to false
-  // will cause Daffodil to log a warning rather than throw an exception so
-  // that a user can run Daffodil on unsupported Java versions, with the
-  // understanding that it is not fully tested and behavior may not be well
-  // defined. This boolean is experimental and should only be used by those
-  // that fully understand the risks.
+  // Daffodil supports Java 7+. This is essentially all version of java that
+  // are in use, so we longer have any unsupported java versions. This is This
```
jadams-tresys: I believe you meant to write "so we *no* longer have any..."

jadams-tresys: Also the "This is This" is a typo

stevedlawrence:
Typo's fixed.

PR #126, opened 2018-10-15
Add default value for binaryCalendarEpoch

jadams-tresys:
Default value in DFDLGeneralFormat.dfdl.xsd is set to the Unix Epoch

DAFFODIL-2006

### DIFF
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
@@ -29,6 +29,7 @@
           alignmentUnits="bytes"
           binaryFloatRep="ieee"
           binaryNumberRep="binary"
+          binaryCalendarEpoch="1970-01-01T00:00:00"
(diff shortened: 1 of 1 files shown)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
```diff
@@ -29,6 +29,7 @@
           alignmentUnits="bytes"
           binaryFloatRep="ieee"
           binaryNumberRep="binary"
+          binaryCalendarEpoch="1970-01-01T00:00:00"
```
stevedlawrence: Should this have a UTC timezone? I guess if it has a timezone, then we'll end up with timezones in the infoset, which maybe we don't want to be the default?

+1

mbeckerle: Pretty sure it should not have a timezone. The default behavior wants to be timezone is unspecified. Otherwise everyhting starts adding +00:00 to everything.

PR #128, opened 2018-10-26
Ignore minLength when lengthKind=explicit

jadams-tresys:
According to DFDL Erratum 5.18 if dfdl:lengthKind='explicit' the facet
minLength is not used to determin the target length.

Although we are ignoring minLength, it is a schema definition wraning if
the explicit length specified is out of range of minLength and it is an
error if it is out of range of textOutputMinLength.

Also fixed a small error in the range check warning.

DFDL-1989

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
@@ -367,9 +367,9 @@ trait ElementRuntimeValuedPropertiesMixin
         case (Implicit, Binary, HexBinary) => (LengthUnits.Bytes, maxLengthLong) // fixed length
         case (Implicit, Text, String) => (lengthUnits, maxLengthLong) // fixed length
         case (Implicit, Text, AnySimpleType) => (lengthUnits, textOutputMinLength) // fixed length
-        case (Explicit, Text, String) => (lengthUnits, minLengthLong)
-        case (Explicit, Binary, HexBinary) => (LengthUnits.Bytes, minLengthLong)
-        case (Explicit, Text, AnySimpleType) => (lengthUnits, textOutputMinLength)
+        case (Explicit, Text, String) => (lengthUnits, 0L)
+        case (Explicit, Binary, HexBinary) => (LengthUnits.Bytes, 0L)
+        case (Explicit, Text, AnySimpleType) => (lengthUnits, 0L)
         case (Prefixed, _, String) => (lengthUnits, minLengthLong)
         case (Prefixed, Text, _) => (lengthUnits, textOutputMinLength)
         case (Pattern, _, String) => (lengthUnits, minLengthLong)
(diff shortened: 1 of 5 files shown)


jadams-tresys:
Based on the description in DAFFODIL-1989 I updated the tests unparseDelimitedPaddedString09 and unparsePaddedString10 to not expect the additional padding to make it reach minLength, which I am assuming is correct behaviour based on the erratum.  If someone could verify this for me I would appreciate it and I will update the test descriptions to reflect this.

mbeckerle:
I believe this is correct: lengthKind 'explicit' means minLength is used for validation only.

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberPropsUnparse.tdml
```diff
@@ -407,7 +407,11 @@
         <ex:e11 xmlns:ex="http://example.com">O</ex:e11>
       </tdml:dfdlInfoset>
     </tdml:infoset>
-    <tdml:document>O#########</tdml:document>
+    <tdml:warnings>
+      <tdml:warning>Schema Definition Warning</tdml:warning>
+      <tdml:warning>Explicit dfdl:length of 2 is out of range for facet maxLength='10'</tdml:warning>
```
mbeckerle: Suggest making these less precise like check for "explicit" "dfdl:length" "out of range" and "facet" and "maxLength='10'".

That at least gives it a chance of being a universal test that will pass for other DFDL implementations as well.

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberPropsUnparse.tdml
```diff
@@ -428,7 +432,11 @@
         <ex:e12 xmlns:ex="http://example.com">O</ex:e12>
       </tdml:dfdlInfoset>
     </tdml:infoset>
-    <tdml:document>O#########</tdml:document>
+    <tdml:warnings>
+      <tdml:warning>Schema Definition Warning</tdml:warning>
+      <tdml:warning>Explicit dfdl:length of 2 is out of range for facet maxLength='10'</tdml:warning>
```
mbeckerle: Similar to above.

jadams-tresys:
Funnily enough I had already made the warning text more general, but hadn't yet written out the file when I made the commit.  Should be included now in the most recent commit.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -367,9 +367,14 @@ trait ElementRuntimeValuedPropertiesMixin
...
-        case (Explicit, Binary, HexBinary) => (LengthUnits.Bytes, minLengthLong)
-        case (Explicit, Text, AnySimpleType) => (lengthUnits, textOutputMinLength)
+        case (Explicit, Text, String) => (lengthUnits, 0L)
+        case (Explicit, Binary, HexBinary) => (LengthUnits.Bytes, 0L)
+        case (Explicit, Text, AnySimpleType) => {
+          if (optLengthConstant.isDefined && (optLengthConstant.get < minLengthLong))
+            SDE("Explicit dfdl:length of %d is out of range for facet textOutputMinLength='%d'",
+              optLengthConstant.get, textOutputMinLength)
+          (lengthUnits, 0L)
+        }
```
stevedlawrence: Should this check also be done at runtime if dfdl:lengthKind="explicit" but dfdl:length is an expression? Or is that a validation error that might already handle?

jadams-tresys: Actually, I just double checked and it seems that there already is runtime validation and this check is unnecessary and can be set to (lengthUnits, 0L) like the other cases for lengthKind Explicit

PR #130, opened 2018-10-31
Initial commit

jadams-tresys:
Still contains debugging print statements.  The test I added should
simply SDE, but it continues past the SDE call and still hits the
Assert.

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
@@ -641,7 +641,12 @@ sealed abstract class ListOfStringLiteralBase(
   private lazy val olc = oneLiteralCooker
 
   protected def cook(raw: String, context: ThrowsSDE, forUnparse: Boolean): List[String] = {
+    if (raw.length != 0 && (raw.head.isWhitespace || raw.last.isWhitespace)) {
+      context.SDE("The property '%s' cannot start or end with the string \" \", did you mean to use '%%SP;' instead?", propNameArg)
+    }
+
     val rawList = raw.split("\\s+").toList
+
     val cooked = {
       val cookedList: ListBuffer[String] = ListBuffer.empty
       rawList.foreach(x => {
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
@@ -111,6 +111,24 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="DP_04_02" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="s1" type="xs:float"
+            dfdl:terminator=" %WSP+;" maxOccurs="unbounded" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="DP_04_03" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="s1" type="xs:float"
+            dfdl:terminator="%WSP+; " maxOccurs="unbounded" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
   </tdml:defineSchema>
 
   <tdml:defineSchema name="DelimiterProperties-Embedded2.dfdl.xsd">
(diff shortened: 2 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala
```diff
@@ -68,6 +68,13 @@ abstract class DelimiterParseEv(delimType: DelimiterTextType.Type, override val
     if (converterResult.length == 1 && converterResult(0) == "") {
       Array()
     } else {
+      System.out.println("In DelimiterParseEv: " + expr)
+      System.out.println("converterResult: " + converterResult)
+      if (converterResult.contains("")) {
+        System.out.println("Empty delimiter found!!!")
```
jadams-tresys: When running testOnly org.apache.daffodil.section12.delimiter_properties.TestDelimiterProperties this print statement (the one before the SDE call) gets displayed, but not the one after the SDE.  However, it continues processing til it hits the Assert call

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/CreateDelimiterDFA.scala
```diff
@@ -116,6 +118,7 @@ object CreateDelimiterDFA {
 
   private def buildTransitions(delim: Seq[DelimBase],
     allStates: ArrayBuffer[State], ignoreCase: Boolean): State = {
+    System.out.println("Delim: " + delim)
     assert(!delim.isEmpty)
```
jadams-tresys: This is the Assert call that fails, even after the SDE is called.

stevedlawrence: Changing from the scala assert() method to daffodils Assert.invariant() results in a stack trace that makes it a little easier to see what's going on. This is part of the stack trace:
```
[error] org.apache.daffodil.exceptions.Assert$.abort(Assert.scala:129)
[error] org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.buildTransitions(CreateDelimiterDFA.scala:123)
[error] org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.apply(CreateDelimiterDFA.scala:44)
[error] org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.apply(CreateDelimiterDFA.scala:83), took 1.291 sec
[error]     at org.apache.daffodil.exceptions.Assert$.abort(Assert.scala:129)
[error]     at org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.buildTransitions(CreateDelimiterDFA.scala:123)
[error]     at org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.apply(CreateDelimiterDFA.scala:44)
[error]     at org.apache.daffodil.processors.dfa.CreateDelimiterDFA$.apply(CreateDelimiterDFA.scala:83)
[error]     at org.apache.daffodil.processors.DelimiterUnparseEv.compute(EvDelimiters.scala:100)
[error]     at org.apache.daffodil.processors.DelimiterUnparseEv.compute(EvDelimiters.scala:83)
```
So the assert is getting called from the DelimiterUnparseEv. I suspect you just need to add logic to that Ev, similar to what you do for DelimiterParseEv.

jadams-tresys: Ah, that makes sense.  Will fix.

jadams-tresys:
Should be good to go now with the most recent commit

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala
```diff
@@ -65,10 +65,10 @@ abstract class DelimiterParseEv(delimType: DelimiterTextType.Type, override val
...
 
     val converterResult = evalAndConvert(state)
-    if (converterResult.length == 1 && converterResult(0) == "") {
-      Array()
-    } else {
-      CreateDelimiterDFA(delimType, trd, converterResult, ignoreCase)
+    (converterResult.length, converterResult(0)) match {
+      case (1, "") => Array()
+      case (_, "") => state.SDE("The string \"\" can only be used when specified as the only delimiter, not as a member of a list of delimiters")
+      case (_, _) => CreateDelimiterDFA(delimType, trd, converterResult, ignoreCase)
```
stevedlawrence: This fixes this for delimiters, but I wonder if there are other properties that also handle leading/trailing spaces incorrectly? The various Cookers in EntityReplacer.scala are the things that actually "cook" the properties values. For example, ListOfStringLiteralBase splits a property value on whitespace. I wonder if somewhere in those cookers would be a better place for the leading/trailing whitespace check. That way this issue would be resolved for all properties and not just delimiters.

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
```diff
@@ -111,6 +111,15 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="DP_04_02" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="s1" type="xs:float"
+            dfdl:terminator=" %WSP+;" maxOccurs="unbounded" />
+        </xs:sequence>
```
stevedlawrence: Might be good to have a trailing space check as well. That should also be an SDE, since the user probably meant ``"FOO%SP;"`` rather than ``"FOO "``

stevedlawrence:
+1

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -642,6 +642,9 @@ sealed abstract class ListOfStringLiteralBase(
 
   protected def cook(raw: String, context: ThrowsSDE, forUnparse: Boolean): List[String] = {
     val rawList = raw.split("\\s+").toList
+    if (((rawList(0) == "") || (rawList.last == "")) && (rawList.length != 1))
+      context.SDE("The string \"\" is not allowed at the start or end of the list if it is not the only item in the list")
+
```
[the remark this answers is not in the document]
jadams-tresys: Addressed these issues and updated the error message in the most recent commit

PR #132, opened 2018-11-05
Remove Java 10 from TravisCI build configuration

stevedlawrence:
Using Java 10 results in a deprecation error and causes the build to
fail. Since Java 10 is deprecated, let's just remove it from the build
matrix.

DAFFODIL-2016

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ jdk:
   - oraclejdk7
   - oraclejdk8
   - oraclejdk9
-  - oraclejdk10
   - oraclejdk11
 sudo: required
 dist: trusty


stevedlawrence:
They've deprecated jdk10 in favor of jdk11. Looks like no one will be able to use jdk10 on travis CI anymore.

mbeckerle:
So, JDK10 sure seemed to have a short lifespan before being deprecated. Was some egregious security bug found in it or something?

stevedlawrence:
Oracle is speeding up their release cycles so a new release happens something like every 6 months. Then on a longer cycle, like every three years, they pick a release to be the LTS release. Java 11 is that LTS, so people are told to no longer use Java 9 or 10.

PR #133, opened 2018-11-06
Enabled a few Calendar tests that now work

jadams-tresys:
There were a few Calendar tests that had been in debug for a while that
now seem to work (or only needed a few small fixes to the tests
themselves to work). They had been moved to debug as part of
DAFFODIL-1945 as part of an issue that a user was having.

DAFFODIL-1945

### DIFF
--- a/daffodil-test/src/test/scala-debug/org/apache/daffodil/section05/simple_types/TestSimpleTypesDebug.scala
+++ b/daffodil-test/src/test/scala-debug/org/apache/daffodil/section05/simple_types/TestSimpleTypesDebug.scala
@@ -98,11 +98,4 @@ class TestSimpleTypesDebug {
   @Test def test_bitOrderChangeInvalid3() { runner1.runOneTest("bitOrderChangeInvalid3") }
 
   @Test def test_bitOrderChangeInvalid() { runner3.runOneTest("bitOrderChangeInvalid") }
-
-  //////////////////////// DAFFODIL-1945 /////////////////////////////
-  @Test def test_dateCalendarDaysInFirstWeek3() { runner.runOneTest("dateCalendarDaysInFirstWeek3") }
-  @Test def test_dateCalendarDaysInFirstWeek5() { runner.runOneTest("dateCalendarDaysInFirstWeek5") }
-
-  @Test def test_dateCalendarFirstDayOfWeek03() { runner.runOneTest("dateCalendarFirstDayOfWeek03") }
-  @Test def test_dateCalendarFirstDayOfWeek04() { runner.runOneTest("dateCalendarFirstDayOfWeek04") }
 }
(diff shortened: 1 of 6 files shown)


daffodil-test/src/test/scala-debug/org/apache/daffodil/section05/simple_types/TestSimpleTypesDebug.scala
```diff
@@ -98,11 +98,4 @@ class TestSimpleTypesDebug {
   @Test def test_bitOrderChangeInvalid3() { runner1.runOneTest("bitOrderChangeInvalid3") }
 
   @Test def test_bitOrderChangeInvalid() { runner3.runOneTest("bitOrderChangeInvalid") }
-
-  //////////////////////// DAFFODIL-1945 /////////////////////////////
-  @Test def test_dateCalendarDaysInFirstWeek3() { runner.runOneTest("dateCalendarDaysInFirstWeek3") }
-  @Test def test_dateCalendarDaysInFirstWeek5() { runner.runOneTest("dateCalendarDaysInFirstWeek5") }
```
stevedlawrence: These two tests are removed from scala-debug but not added to scala.

jadams-tresys: They were already added from a previous commit, but the debug tests were never removed.

stevedlawrence: Are you sure? I only see them commented out in TestSimpleTypes.scala, lines 113 and 115.

jadams-tresys: You are right, I could have sworn I enabled them, but I guess I didn't.  I was thinking of the "FirstDayOfWeek" tests that were already enabled, not the "DaysInFirstWeek".

mbeckerle: Responding to steve's original comment..... Ok, so we need to contact the original user (per email thread in the DAFFODIL-1945 ticket) and find out what operating system, and what locale and time zone, etc. It's possible race conditions caused this transient failure, in which case keeping the tests commented out isn't helpful at finding the problem, but we need to try to reproduce.

stevedlawrence:
@jadams-tresys Commit 0b924b67de67 for DAFFODIL-2074 was able to reproduce, track down, and fix issues related to locale. You might now be able to use the steps in that bug to reproduce the locale issues for this bug and determine if the fix for DAFFODIL-2074 also fixed these issues. Would be nice to be able to close out this PR.

PR #135, opened 2018-11-12
Show different error message for OOM

jadams-tresys:
Instead of showing a generic error message, we should suggest
increasing memory when we crash due to an OutOfMemoryError.

DAFFODIL-1953

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -1355,6 +1355,20 @@ object Main extends Logging {
         log(LogLevel.Error, "%s", e.getMessage())
         1
       }
+      case e: OutOfMemoryError => {
+        System.err.println("""|
+                            |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                            |!!             Daffodil ran out of memory!              !!
+                            |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                            |
+                            | Try increasing the amount of memory by changing or
+                            | setting the DAFFODIL_JAVA_OPTS environment variable.
+                            | "DAFFODIL_JAVA_OPTS=-Xmx5G" for 5GB.
+                            |
+                            |""".stripMargin)
+        e.printStackTrace
+        1
+      }
       case e: Exception => {
         bugFound(e)
       }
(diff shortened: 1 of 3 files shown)


jadams-tresys:
Wasn't really sure of a practical way to add a test for this case.  I did test it with the CLI tool by manually decreasing the memory available it it does work.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1355,6 +1355,19 @@ object Main extends Logging {
...
         1
       }
+      case e: OutOfMemoryError => {
+        System.err.println("""|
+                            |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                            |!!             Daffodil ran out of memory!              !!
+                            |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                            |
+                            | Try increasing the amount of memory by changing or
+                            | setting the DAFFODIL_JAVA_OPTS environment variable.
```
mbeckerle: Might as well tell them how to set it i.e., for the OpenJDK JVM the parameter value is -Xmx5G for 5 gigabytes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
```diff
@@ -271,7 +271,7 @@ class ChoiceParser(ctxt: RuntimeData, val childParsers: Vector[Parser])
       case t: Throwable => {
         if (pBefore != null) {
           markLeakCausedByException = true
-          if (!t.isInstanceOf[SchemaDefinitionDiagnosticBase] && !t.isInstanceOf[UnsuppressableException]) {
+          if (!t.isInstanceOf[SchemaDefinitionDiagnosticBase] && !t.isInstanceOf[UnsuppressableException] && !t.isInstanceOf[OutOfMemoryError]) {
```
stevedlawrence: This pattern of schema checking also appears in SequenceParserBases.scala, we probably need a similar check there.

Also, rather than checking the very specific OutOfMemoryError, perhaps we instead want to check it's parent java.lang.Error,? The javadoc says java.lang.Error "indicates a serious problem that a reasonable application should not try to catch". So on any java.lang.Error we still attempt to clean up the MarkPool but don't try to warn about an unexpected exception thrown. The below message is really trying to tell when Daffodil failed to catch an exception that it should have caught elsewhere else.

This also makes me wonder if we shouldn't even be catch/using Throwable's anywhere, but instead only catch java.lang.Exceptions? I'm pretty sure we use Throwables all over the place, so it's not worth changing for this bug, but might be worth creating a new bug to reexamine our usage of Throwables vs Exceptions.

jadams-tresys: Most recent push should address the java.lang.Error check.

PR #136, opened 2018-11-13
Changing to ignore attributeFormDefault

efinnegan:
Previously an error was thrown when attributeFormDefault="qualified". This attribute
is not used by DFDL so now a warning is issued and the attribute is ignored.

Daffodil-565

### DIFF
--- a/daffodil-test-ibm1/src/test/scala-debug/org/apache/daffodil/IBMTests2.scala
+++ b/daffodil-test-ibm1/src/test/scala-debug/org/apache/daffodil/IBMTests2.scala
@@ -44,11 +44,6 @@ class IBMTestsThatThrow {
   lazy val runner1 = new DFDLTestSuite(Misc.getRequiredResource(tdml1))
   lazy val runner2 = new DFDLTestSuite(Misc.getRequiredResource(tdml2))
 
-  @Test def test_scoping_define_format_8_04() { runner1.runOneTest("scoping_define_format_8_04") } //DFDL-565 - attributeFormDefault='qualified'
-
-  @Test def test_scoping_default_format_8_01() { runner1.runOneTest("scoping_default_format_8_01") } // DFDL-565 attributeFormDefault='qualified'
-  @Test def test_scoping_define_format_8_01() { runner1.runOneTest("scoping_define_format_8_01") } // DFDL-565 attributeFormDefault='qualified'
-
   @Test def test_alignment_bytes_12_05() { runner1.runOneTest("alignment_bytes_12_05") } //DFDL-99 binary dateTime
 
   @Test def test_length_implicit_12_02() { runner1.runOneTest("length_implicit_12_02") } // implicit length string - bug in IBM test (doesn't have minLength - both are required)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaDocument.scala
@@ -118,11 +118,12 @@ final class XMLSchemaDocument(xmlArg: Node,
     val hasSchemaLocation = (xml \ "@schemaLocation").text != ""
     val hasBlockDefault = (xml \ "@blockDefault").text != ""
     val hasFinalDefault = (xml \ "@finalDefault").text != ""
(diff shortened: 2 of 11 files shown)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/dafint.xsd
```diff
@@ -17,7 +17,7 @@
 -->
 
 <xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"
-  attributeFormDefault="unqualified" elementFormDefault="qualified"
+  elementFormDefault="qualified"
   targetNamespace="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int"
```
stevedlawrence: I think for all these normal schemas (i.e. not dfdl.xsd's) we probably do want the attributeFormatDefault property. That property is valid and is used when doing normal schema validation. It's only ignore in DFDL schemas since the property isn't part of the DFDL specification.

daffodil-test-ibm1/src/test/scala-debug/org/apache/daffodil/IBMTests2.scala
```diff
@@ -44,10 +44,8 @@ class IBMTestsThatThrow {
   lazy val runner1 = new DFDLTestSuite(Misc.getRequiredResource(tdml1))
   lazy val runner2 = new DFDLTestSuite(Misc.getRequiredResource(tdml2))
 
-  @Test def test_scoping_define_format_8_04() { runner1.runOneTest("scoping_define_format_8_04") } //DFDL-565 - attributeFormDefault='qualified'
-
-  @Test def test_scoping_default_format_8_01() { runner1.runOneTest("scoping_default_format_8_01") } // DFDL-565 attributeFormDefault='qualified'
-  @Test def test_scoping_define_format_8_01() { runner1.runOneTest("scoping_define_format_8_01") } // DFDL-565 attributeFormDefault='qualified'
+  @Test def test_scoping_default_format_8_01() { runner1.runOneTest("scoping_default_format_8_01") } // Attribute "encodingErrorPolicy" was already specified - bug in IBM test
+  @Test def test_scoping_define_format_8_01() { runner1.runOneTest("scoping_define_format_8_01") } // Attribute "encodingErrorPolicy" was already specified - bug in IBM test
```
stevedlawrence: Does this test pass if the duplicate errorEncodingPolicy is removed? These tests are more about how properties are resolved, and not really about how duplicate properties are handled (which I agree, seems likea bug that IBM DFDL should catch). I'd guess the duplicate property was unintentional and can be removed and still maintain the intention of the test.

mbeckerle: This feels like our mistake adding the duplicate encodingErrorPolicy property. That is, this property encodingErrorPolicy, is relatively new. Probably hadn't even been invented when we first got a bunch of preliminary DFDL tests from IBM for incorporation into Daffodil years ago. So we subsequently added this to keep the tests working, once the new property was added to Daffodil, and required. I think adding it twice was just a mistake we made.

efinnegan: They also need textOutputMinLength added. I wasn't sure if we wanted to change these tests since they were from IBM. Is it ok to update them?

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/dafint.xsd
efinnegan: Ok I will take out the change for removing attributeFormDefault in these schemas.

daffodil-test-ibm1/src/test/scala-debug/org/apache/daffodil/IBMTests2.scala
stevedlawrence: [DAFFODIL-1616](https://issues.apache.org/jira/browse/DAFFODIL-1616) is the bug to fix the issue related to textOutputMinLength, which Mike is working on. If the tests work with that property added, it seems fine to me to modify the tests and add the property, and eventually that property will just be ignored when Mike's fix is in.

PR #138, opened 2018-11-14
Adding error when dfdl:initiator is "%"

efinnegan:
When the initiator is a percent sign the initiator was being ignored
so this will now throw an error.

DAFFODIL-1840

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
@@ -129,6 +129,56 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="DP_05" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence dfdl:initiatedContent="no"
+          dfdl:separatorPosition="infix" dfdl:separator=",">
+          <xs:element name="num1" type="xs:int"
+            dfdl:initiator="%" maxOccurs="unbounded" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="DP_06" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence dfdl:initiatedContent="no"
+          dfdl:separatorPosition="infix" dfdl:separator="%">
+          <xs:element name="num1" type="xs:int"
+            dfdl:initiator=":" maxOccurs="unbounded" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="DP_07" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence dfdl:initiatedContent="no"
+          dfdl:separatorPosition="infix" dfdl:separator=",">
+          <xs:element name="num1" type="xs:int"
+            dfdl:terminator="%" maxOccurs="unbounded" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="DP_08" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence dfdl:initiatedContent="no"
+          dfdl:separatorPosition="infix" dfdl:separator=",">
+          <xs:element name="num1" type="xs:int"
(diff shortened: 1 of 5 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -186,6 +186,9 @@ trait DelimitedRuntimeValuedPropertiesMixin
     val qn = this.qNameForProperty("initiator")
     val typeIfStaticallyKnown = NodeInfo.String
     val typeIfRuntimeKnown = NodeInfo.NonEmptyString
+    if (initiatorRaw.value == "%") {
+      SDE("Invalid dfdl:initiator. \"%\" cannot be a dfdl:initiator.")
```
mbeckerle: Suggest improve message not to just say it's wrong, but suggest the user meant "%%", and remind them that % introduces DFDL entities and so must be self-escaped if you want to literally express a single % character.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -186,6 +186,9 @@ trait DelimitedRuntimeValuedPropertiesMixin
     val qn = this.qNameForProperty("initiator")
     val typeIfStaticallyKnown = NodeInfo.String
     val typeIfRuntimeKnown = NodeInfo.NonEmptyString
+    if (initiatorRaw.value == "%") {
```
stevedlawrence: I'm not sure if this is the right fix for this, for a couple reasons:
1. This only fixes initiators, but my guess is that this same bug applies to separators and terminators, and maybe even other properties. Might be worth adding a few extra tests to make sure this issue is resolved with other properties, and not just initiator/terminator/separator.
2. If the initiator was an expression that evaluated to a percent, e.g. ``{ "%" }``, this check wouldn't find it since this only checks the raw value. We really need to check the value after it's been evaluated, whether it's a constant or an expression.

So fixing the root issue probably means a change somewhere in the Cookers, which all properties use and which are also used after expression evaluation. Take a look at the EntityReplacer, which is used by the cookers for handle char classes. The replaceEntity function already has one SDE for invalid dfdl entities using regular expressions (see the replaceEntity function), so likely a tweak just needs to be made there.

Also, when testing this, I noticed that not only does a single percent not work, but a single percent at the end also doesn't throw an error, e.g. dfdl:initiator="foo%". Sounds like whatever is checking for invalid dfdl entities doesn't handle the edge case of a percent at the end of the string, rather than just a single %.

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
```diff
@@ -129,6 +129,56 @@
...
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="DP_09" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence dfdl:initiatedContent="no"
+          dfdl:separatorPosition="infix" dfdl:separator=",">
+          <xs:element name="num1" type="xs:int"
+            dfdl:terminator="{if ('true') then '%' else 'IMPOSSIBLE'}" maxOccurs="unbounded" />
```
mbeckerle: Excellent test.

daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterProperties.tdml
```diff
@@ -810,4 +860,59 @@
...
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="percentTerminator"
+    model="DelimiterProperties-Embedded.dfdl.xsd"
+    description="A terminator of '%' causes an error"
+    root="DP_07">
+    <tdml:document><![CDATA[1#,2#,3#,4#]]></tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>Invalid DFDL Entity (%) found</tdml:error>
```
mbeckerle: This test could check for the advice about needing a self-escaped %. I would just add a <tdml:error>%%</tdml:error> to this.

PR #139, opened 2018-11-16
Updating to set encoding of CLI output to UTF-8

efinnegan:
Without an encoding specified the system default is used for outputting
the infoset in the CLI, which could result in unexpected output. This
will ensure UTF-8 is used instead of the system default.

Daffodil-2011

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -845,7 +845,7 @@ object Main extends Logging {
               case Some("-") | None => System.out
               case Some(file) => new FileOutputStream(file)
             }
-            val writer = new BufferedWriter(new OutputStreamWriter(output))
+            val writer = new BufferedWriter(new OutputStreamWriter(output, "UTF-8"))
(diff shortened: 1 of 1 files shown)


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -845,7 +845,7 @@ object Main extends Logging {
               case Some("-") | None => System.out
               case Some(file) => new FileOutputStream(file)
             }
-            val writer = new BufferedWriter(new OutputStreamWriter(output))
+            val writer = new BufferedWriter(new OutputStreamWriter(output, "UTF-8"))
```
efinnegan: I tested this by changing my settings to use DAFFODIL_JAVA_OPTS="-Dfile.encoding=ASCII". The "É" in my output showed as a "?" without this change. I'm not sure if there is a test we can add to show this though.

mbeckerle: I don't think a test is needed.

stevedlawrence:
+1

PR #140, opened 2018-11-17
TDML Runner and test changes for Cross Validation with IBM.

mbeckerle:
This review goes with one that is outside of the Daffodil repo here:

OpenDFDL/ibmDFDLCrossTester#1

This is a very large change set. 

Major work on the TDML Runner, and cleanup of tests. I am able to run tests in daffodil-test-ibm1 against IBM dfdl, using just a few setup clicks. 

I do not yet have a way to do this from sbt/command line. 

Also fixed other minor bugs that were in the way of this. I tried to separate these out into separate change sets, but was unable to do so. 
DAFFODIL-2014 - TextBidi property name was misspelled.
DAFFODIL-2015 - accept textStandardOutputCharacter (legacy prop name)
DAFFODIL-1616 - textOutputMinLength required when it shouldn't be.

TDML Runner now implements everything it does in terms of an abstract
DFDL processor defined in the tdml.processor package.

There are a number of sweeping changes in this that change many files.

First: Added explicit xs:include of DFDLGeneralFormat.dfdl.xsd to all
TDML. The Daffodil TDML runner used to implicitly include
DFDLGeneralFormat.dfdl.xsd. Now that we're cross testing, tests won't
necessarily want to do this, so it has to be done explicitly.

Second: The namespace prefixes in the root elements in many of our test
cases were not correct. Those had to be uniformly fixed so that tests
can insist on the proper namespace in result data.
These were wrong because the comparison logic in the test rig was
tolerating them being incorrect, but as part of making the TDML runner
into a viable cross-testing rig, these things are being tightened up.
The root element named by a test must match that of the infoset element.
(in fact the root element declaration is unnecessary, as the root will
be taken from the infoset's root element.

The TDML module's code has changed significantly.

The daffodil-tdml module has been split into daffodil-tdml
and daffodil-tdml-processor. The latter has dependencies on all of
daffodil. The former is independent of most of daffodil so that it can be
used with an IBM-based implementation of the tdml processor class
(Which is defined separately and outside of Daffodil's code base.)

Yes, this is yet another API for Daffodil, but it is narrower and not as
comprehensive as the SAPI/JAPI stuff. It's only and exactly what the
TDML runner uses to drive tests.

The TDML processors are always dynamically loaded from the classpath
(even Daffodil's own - just to insure we're always testing this
mechanism.) This allows one to change which processor is used by changing the
classpath, and avoids a static dependency from the daffodil-tdml module on
the daffodil-tdml-processor module.

Enable TDMLDFDLProcessors to force skip of left-over-data checking.

Moved all the IBM contributed tests to the daffodil-test-ibm1 module
out of the daffodil-test module. There were redundant tests being run in
both places. Probably this was due to an ambitious QA effort to
reorganize the IBM-supplied tests to determine requirements coverage on
the DFDL Specification, along with the Daffodil-project tests... I mean
why have more than one test for a given requirement. However, this was
not done right - moving the tests from one place to another - instead
tests were left in both places. So I have moved all the tests that use
the files supplied orignally by IBM back to the daffodil-test-ibm1
module.

So there is no longer any ibm-tests subdir of the daffodil-test module.

(Reminder: IBM supplied these a long time ago, and they are fully
approved for inclusion into Daffodil, with ASF licensing, etc. There is
no issue of source code origins here. )

Modify daffodil-test-ibm1 module to run most tests on IBM_DFDL also.

Properties were added that are required by IBM DFDL and other properties
changed: E.g.,

encodingErrorPolicy='error'
choiceLengthKind='implicit' (also added to DFDLGeneralFormat.dfdl.xsd)
binaryNumberCheckPolicy="lax" (ditto)

This obsolete property had to be removed from our
IBMDefined/GeneralPurposeFormat.xsd

textBidiTextShaped="no"

Quite a lot of the tests in daffodil-test-ibm1 pass when run on
IBM_DFDL. Some fail.

A few tests have been modified to specify daffodil only because they use
features only in Daffodil (like hidden groups and calculated values).

The failures are mixture of unexpected infoset contents (much around
calendars with timezone Z vs. "+00:00".

There are other tests that appear to be places where IBM DFDL and
daffodil have fundamentally different behavior that needs
investigation.

The runner.trace method works and will invoke the IBM_DFDL trace
facility when running tests via IBM_DFDL. This is needed to get more
than just "parse failed" from it.

A couple of other minor things.

Fixed test regression in test_alignmentPaddingOVC4.

This test shows that when textOutputMinLength and an explicit length
constant are inconsistent, we still get a warning anyway.

Removed test dateBinPacked3.

This test was only checking that binaryNumberCheckPolicy was not
defined, but it now is defined for IBM compatibility reasons, in the
DFDLGeneralFormat.dfdl.xsd.

DAFFODIL-2014, DAFFODIL-723, DAFFODIL-2015, DAFFODIL-1616


daffodil-tdml/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
```diff
@@ -0,0 +1,441 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
mbeckerle: FYI: These are not new tests. They came from TestTDMLRunner.scala which is now in the tests of daffodil-tdml-processor. 

This file is the unit tests for the daffodil-tdml, which does not include a processor. So these are true unit tests. They can't run an actual test. They are unit tests for the piece parts of TDML files, and the various things that get computed about a test case, without actually running a TDML test. 

These tests used to be in the same file with a bunch of tests that DO actually run the TDML runner. They were split out so that these can run without the need of a TDML processor.

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunner.scala
```diff
@@ -37,207 +37,6 @@ class TestTDMLRunner {
   val tns = example
   // val sub = XMLUtils.DFDL_XMLSCHEMASUBSET_NAMESPACE
 
-  @Test def testDocPart1() {
```
mbeckerle: FYI: This is where all the tests in the UnitTestTDMLRunner.scala file came from.

project/Dependencies.scala
```diff
@@ -29,6 +29,7 @@ object Dependencies {
     "xml-resolver" % "xml-resolver" % "1.2",
     "commons-io" % "commons-io" % "2.6",
     "jline" % "jline" % "2.14.6",
+    "junit" % "junit" % "4.12", // needed for TDML runner use of AssertionViolationException
```
stevedlawrence: I'd avoid setting junit as a core dependency to all projects. It's really only needed as a compile dependency for the TDML project--the other projects don't need it. Maybe the best way to do this is to create a new junit dependency val, and add just that to the tdml project, e.g.:

```scala
// Dependencies.scala
lazy val junit = "junit" % "junit" % "4.12"

lazy val test = Seq(
  junit % "it,test",
  "com.novocode" % "junit-interface" % "0.11" % "it,test",
  "org.scalacheck" %% "scalacheck" % "1.14.0" % "it,test"
)

// build.sbt
lazy val tdml = Project("daffodil-tdml", ...)
                ...
                .settings(libraryDependencies += Dependencies.junit)
```

build.sbt
```diff
@@ -63,30 +63,33 @@ lazy val sapi             = Project("daffodil-sapi", file("daffodil-sapi")).conf
                               .settings(commonSettings)
```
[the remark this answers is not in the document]
stevedlawrence: It's in here, just at the very bottom.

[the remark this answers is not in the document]
mbeckerle: Not on my screen.... See screenshot. 
screenshot at 2018-11-19 12 13 11

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1227,7 +1226,7 @@ object Main extends Logging {
             println(formatStr.format(headers: _*))
             tests.foreach { testPair =>
               testPair match {
-                case (name, Some(test)) => println(formatStr.format(name, test.model, test.root, test.description))
+                case (name, Some(test)) => println(formatStr.format(name, test.model, test.rootName, test.description))
                 case (name, None) => println(formatStr.format(name, "[Not Found]", "", ""))
               }
```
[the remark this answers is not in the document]
mbeckerle: Yes. Was hoping you'd point out exactly where this is needed. :-)

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
```diff
@@ -169,15 +169,13 @@
 
   <!-- 13.3 Properties for Bidirectional support for all Simple Types with 
     Text representation -->
-  <xsd:attribute name="textBidi" type="dfdl:YesNoEnum" />
```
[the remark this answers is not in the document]
mbeckerle: yes it is. Was artifact of the misspelling that these were not causing an error before. With spelling fixed, these now are problematic.

daffodil-tdml/src/main/scala/org/apache/daffodil/tdml/TDMLException.scala
```diff
@@ -17,16 +17,72 @@
...
+import org.apache.daffodil.api.Diagnostic
+import org.apache.daffodil.util.Maybe
+import org.junit.AssumptionViolatedException
 
-  def this(msg: String) = this(msg, Nil)
+object TDMLException {
 
-  def this(cause: Throwable) = this(cause.getMessage(), List(cause))
+  def msgWithImpl(msg: String, implementation: String) =
+    if (implementation != "none" &&
```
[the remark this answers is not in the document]
mbeckerle: Changing to an Option type makes sense. Optional arg would just get left off i think. Making it required arg is important since you really do need to consider whether the exception is implementation-specific or not.

project/Dependencies.scala
mbeckerle: Ah, that's the right way to do it.  Thanks.

daffodil-tdml/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -437,69 +432,70 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
-          parent.checkAllTopLevel,
-          root,
-          null) {
-            compileProcessor(uss, useSerializedProcessor)
-          }
-      case _ => {
-        compileProcessor(schemaSource, useSerializedProcessor)
-      }
-    }
+    val className = "org.apache.daffodil.tdml.processor.DaffodilTDMLDFDLProcessorFactory"
```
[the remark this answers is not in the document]
mbeckerle: So, I did look briefly into ServiceLoader. I think the basic API is ok.

However, this whole aspect of Java has been criticised for *not* solving the module-version-collision issue - there is no isolation of modules.  I.e., one could, legitimately, consider it entirely pointless, and highly subject to change in future SDKs.  OTOH, it does seem simpler or more approachable than the corresponding aspects of OSGI, which solves perhaps too many problems. 

My preference is just to stay with the simplest possible thing for right now. At least until an idiom for usage of this stuff appears... right now I don't have an idiom for running tests against IBM DFDL, from sbt or the command line, worked out as yet. I can run them from eclipse after tweeking classpaths of the test projects. 

I do think the class name should change here, as this one suggests this is some Daffodil-specific class, when this is the class name ANY implementation must have, in order for the class loader to find it.  The IBM DFDL implementation also implements this exact same class name.  So I'm thinking the trait/interface (or "service") definition name grows the prefix "Abstract..." on its name, and the concrete implementation is just class "TDMLDFDLProcessorFactory" which is shorter, cleaner, and admits multiple implementations.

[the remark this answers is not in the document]
stevedlawrence: Understood. Seems reasonable to hardcode the class name with the understanding that any implementations must name their class with that name. And we don't have to worry about conflicts since we only support a single implementation on the classpath. That name makes more sense to me. That Daffodil to me implied it's just for the Daffodil implementation.

As far as an idiom goes for SBT, I think we can modify the daffodil-tdml/fullClasspath setting to remove the stuff we don't want on the class path and add stuff we do. Not sure exactly the syntax, but I can help figure that out.

project/Dependencies.scala
```diff
@@ -29,6 +29,9 @@ object Dependencies {
     "xml-resolver" % "xml-resolver" % "1.2",
     "commons-io" % "commons-io" % "2.6",
     "jline" % "jline" % "2.14.6",
+  )
+
+  lazy val junit = Seq(
     "junit" % "junit" % "4.12", // needed for TDML runner use of AssertionViolationException
```
stevedlawrence: Minor, but junit is now in the file twice, once here and one in ``lazy val test``. The concern is that if we update the junit version we might not realize it needs to be updated in two places. An alternative is to not define junit as a Seq, but define it as a single dependency, then use that both here and in build.sbt, for example:
```scala
// Dependencies.scala
lazy val junit = "junit" % "junit" % "4.12"

lazy val test = Seq(
  junit % "it,test",
  "com.novocode" % "junit-interface" % "0.11" % "it,test",
  "org.scalacheck" %% "scalacheck" % "1.14.0" % "it,test"
)

// build.sbt
lazy val tdml = ...
                .settings(libraryDependencies += junit)  // note += instead of ++=
```
This way a version update is just in a single place.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/BC.dfdl.xsd
```diff
@@ -47,7 +47,7 @@
         </sequence>
         <sequence dfdl:separator="," dfdl:terminator="%NL;">
           <element name="y" type="xsd:boolean" maxOccurs="unbounded"
-            dfdl:textBooleanTrueRep="{ 1*10 }" dfdl:textBooleanFalseRep="{'a b c'}" />
+            dfdl:textBooleanTrueRep="{ xsd:string(1*10) }" dfdl:textBooleanFalseRep="{'a b c'}" />
```
stevedlawrence: Not sure it's worth investigating for this, but shouldn't DPath be able to implicitly coerce the number into a string? Or is this is a limitation with IBM's implementation that it does not do coercion?

mbeckerle: Will raise for clarification in DFDL Workgroup. Whether these coercion are implicit or explicit isn't clear in the DFDL spec for me. 

When to implicitly convert type is always a subtle thing in any language. It helps declutter things if simple implicit conversions are done. But if types are implicitly converted most of the time, then static type checking can't really find any errors for you, since a coercion can always be inserted. 

In this particular case, I think a static error is actually a better thing. Why would you write 1*10 for textBooleanTrueRep? If someone writes a numeric expression for something that's clearly supposed to be a string, probably they're just making a mistake.

project/Dependencies.scala
mbeckerle: Fixed.

daffodil-tdml/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -437,69 +432,81 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
+    // single tier of catch, with cases for all the various throws that could have happened anywhere on
+    // the three lines of actions that can throw various things.
+    //
+    // of course this will allocate an object, so not for tightest inner loops, but the code
+    // cleanup is substantial.
+    //
+    val clazz = Try(Class.forName(className))
+    val constructor = clazz.map { _.getDeclaredConstructor() }
+    val tryInstance = constructor.map { _.newInstance().asInstanceOf[AbstractTDMLDFDLProcessorFactory] }
+    val res = tryInstance.get
```
stevedlawrence: If no TDMLDFDLProcessorFactory is on the classpath this ``.get`` will fail with a ClassNotFoundException. Might be more helpful to instead throw a TDMLException with a helpful error message?

mbeckerle: fixed

daffodil-lib/src/test/scala/org/apache/daffodil/schema/annotation/props/TestGeneratedProperties.scala
```diff
@@ -138,7 +138,7 @@ class TestGeneratedProperties {
     comparePropValue(hasProps.textStringJustification, "left")
     //comparePropValue(hasProps.textStringPadCharacter, "%SP;")
     comparePropValue(hasProps.truncateSpecifiedLengthString, "no")
-    comparePropValue(hasProps.textOutputMinLength, "0")
+    // comparePropValue(hasProps.textOutputMinLength, "0")
```
jadams-tresys: Should this and the below commented out lines remain in the code?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -200,7 +207,7 @@ class DFDLTestSuite private[tdml] (
 
   val errorHandler = new org.xml.sax.ErrorHandler {
     def warning(exception: SAXParseException) = {
-      loadingExceptions == exception +: loadingExceptions
+      loadingExceptions = exception :: loadingExceptions
       // System.err.println("TDMLRunner Warning: " + exception.getMessage())
```
jadams-tresys: Probably should remove the commented out print statements

PR #143, opened 2018-12-07
Support dfdl:length="prefixed"

stevedlawrence:
- Add a concept of a "detached element". This is an element that has no
  representation in the infoset, acting only as a place to hang
  properties and as a temporary infoset element to parse unparse.
- When parsing a prefix element, this creates a new detached element,
  parses the prefix length into the element, pulls out the length from
  the element, then uses that length for parsing the actual data. The
  reverse process is performed on unparse, except in the case of binary
  number where the lengths are predetermined by the specification.
- Rename OVCRetry to SimpleTypeRetry, since this is now used to retry
  the prefix length unparser, which has the same behavior as output
  value calc. It will suspend until something figures out the prefix
  length, sets the simple value, and then it can resume and unparse that
  value.
- Add many new binary related parsers/unparsers to handle prefix
  lengths. Text representations are all handle the same using specified
  length parsers/unparsers. This involved some level of refactoring to
  keep things somewhat clean, but not what is really needed. The binary
  parsers all have subtle differences which make this challenging.
- Modified the property generator to resolve qnmaes.
- For packed numbers, split out the length determination from the
  function that converts to a byte array. With prefixed lengths, we need
  to know how much minimum space is needed without needing to create the
  entire array.
- This does not support the ability to nest prefix lengths.
  Functionally, work is complete for parsing, but on the uparsing side
  requires modifications to how UState is cloned. Nesting prefix lengths
  results in nesting suspensions, which doesn't work right now--one
  cannot create suspension from inside a suspension.
- This does not support the ability to unparse prefix lengths when
  lengthUnits is characters. This is a similar issue to trying to use
  content/value-length functions with 'characters'.

DAFFODIL-114

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
@@ -20,6 +20,7 @@ package org.apache.daffodil.grammar
 import org.apache.daffodil.dsom.Term
 import org.apache.daffodil.dsom.ElementBase
 import org.apache.daffodil.dsom.ModelGroup
+import org.apache.daffodil.dsom.DetachedElementDecl
 import org.apache.daffodil.schema.annotation.props.gen.AlignmentUnits
 import org.apache.daffodil.schema.annotation.props.gen.LengthKind
 import org.apache.daffodil.schema.annotation.props.gen.LengthUnits
@@ -125,8 +126,8 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
   }
 
   private lazy val priorAlignmentApprox: AlignmentMultipleOf = {
-    if (this.isInstanceOf[Root]) {
-      AlignmentMultipleOf(0) // root is aligned with anything
+    if (this.isInstanceOf[Root] || this.isInstanceOf[DetachedElementDecl]) {
+      AlignmentMultipleOf(0) // root and detached elements are aligned with anything
     } else {
       val (priorSibs, parent) = potentialPriorTerms
       val arraySelfAlignment =
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -25,6 +25,7 @@ import scala.Long
 import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
+import org.apache.daffodil.dsom.DetachedElementDecl
 import org.apache.daffodil.dsom.ElementBase
 import org.apache.daffodil.dsom.ExpressionCompilers
 import org.apache.daffodil.dsom.InitiatedTerminatedMixin
@@ -33,16 +34,21 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.grammar.primitives.AlignmentFill
 import org.apache.daffodil.grammar.primitives.BCDDecimalDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.BCDDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.BCDDecimalPrefixedLength
 import org.apache.daffodil.grammar.primitives.BCDDecimalRuntimeLength
 import org.apache.daffodil.grammar.primitives.BCDIntegerDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.BCDIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.BCDIntegerPrefixedLength
 import org.apache.daffodil.grammar.primitives.BCDIntegerRuntimeLength
 import org.apache.daffodil.grammar.primitives.BinaryBoolean
+import org.apache.daffodil.grammar.primitives.BinaryBooleanPrefixedLength
 import org.apache.daffodil.grammar.primitives.BinaryDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.BinaryDecimalPrefixedLength
 import org.apache.daffodil.grammar.primitives.BinaryDecimalRuntimeLength
 import org.apache.daffodil.grammar.primitives.BinaryDouble
 import org.apache.daffodil.grammar.primitives.BinaryFloat
 import org.apache.daffodil.grammar.primitives.BinaryIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.BinaryIntegerPrefixedLength
 import org.apache.daffodil.grammar.primitives.BinaryIntegerRuntimeLength
 import org.apache.daffodil.grammar.primitives.CaptureContentLengthEnd
 import org.apache.daffodil.grammar.primitives.CaptureContentLengthStart
@@ -87,12 +93,15 @@ import org.apache.daffodil.grammar.primitives.ElementParseAndUnspecifiedLength
 import org.apache.daffodil.grammar.primitives.ElementUnused
 import org.apache.daffodil.grammar.primitives.HexBinaryDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.HexBinaryEndOfBitLimit
+import org.apache.daffodil.grammar.primitives.HexBinaryLengthPrefixed
 import org.apache.daffodil.grammar.primitives.HexBinarySpecifiedLength
 import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalPrefixedLength
 import org.apache.daffodil.grammar.primitives.IBM4690PackedDecimalRuntimeLength
 import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerPrefixedLength
 import org.apache.daffodil.grammar.primitives.IBM4690PackedIntegerRuntimeLength
 import org.apache.daffodil.grammar.primitives.Initiator
 import org.apache.daffodil.grammar.primitives.InputValueCalc
@@ -102,24 +111,27 @@ import org.apache.daffodil.grammar.primitives.LiteralCharacterNilOfSpecifiedLeng
 import org.apache.daffodil.grammar.primitives.LiteralNilDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.LiteralValueNilOfSpecifiedLength
 import org.apache.daffodil.grammar.primitives.LogicalNilValue
-import org.apache.daffodil.grammar.primitives.OVCRetry
 import org.apache.daffodil.grammar.primitives.OnlyPadding
 import org.apache.daffodil.grammar.primitives.PackedDecimalDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.PackedDecimalKnownLength
+import org.apache.daffodil.grammar.primitives.PackedDecimalPrefixedLength
 import org.apache.daffodil.grammar.primitives.PackedDecimalRuntimeLength
 import org.apache.daffodil.grammar.primitives.PackedIntegerDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.PackedIntegerKnownLength
+import org.apache.daffodil.grammar.primitives.PackedIntegerPrefixedLength
 import org.apache.daffodil.grammar.primitives.PackedIntegerRuntimeLength
 import org.apache.daffodil.grammar.primitives.PaddingInfoMixin
-import org.apache.daffodil.grammar.primitives.PrefixLength
 import org.apache.daffodil.grammar.primitives.RightCenteredPadding
 import org.apache.daffodil.grammar.primitives.RightFill
 import org.apache.daffodil.grammar.primitives.SimpleNilOrValue
+import org.apache.daffodil.grammar.primitives.SimpleTypeRetry
 import org.apache.daffodil.grammar.primitives.SpecifiedLengthExplicit
 import org.apache.daffodil.grammar.primitives.SpecifiedLengthExplicitCharacters
 import org.apache.daffodil.grammar.primitives.SpecifiedLengthImplicit
 import org.apache.daffodil.grammar.primitives.SpecifiedLengthImplicitCharacters
 import org.apache.daffodil.grammar.primitives.SpecifiedLengthPattern
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthPrefixed
+import org.apache.daffodil.grammar.primitives.SpecifiedLengthPrefixedCharacters
 import org.apache.daffodil.grammar.primitives.StringDelimitedEndOfData
 import org.apache.daffodil.grammar.primitives.StringOfSpecifiedLength
 import org.apache.daffodil.grammar.primitives.Terminator
@@ -139,6 +151,7 @@ import org.apache.daffodil.schema.annotation.props.gen.TextNumberRep
 import org.apache.daffodil.schema.annotation.props.gen.YesNo
 import org.apache.daffodil.util.PackedSignCodes
 import org.apache.daffodil.xml.GlobalQName
+import org.apache.daffodil.xml.QName
 import org.apache.daffodil.xml.XMLUtils
 
 /////////////////////////////////////////////////////////////////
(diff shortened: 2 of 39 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
```diff
@@ -125,8 +126,8 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
   }
 
   private lazy val priorAlignmentApprox: AlignmentMultipleOf = {
-    if (this.isInstanceOf[Root]) {
-      AlignmentMultipleOf(0) // root is aligned with anything
+    if (this.isInstanceOf[Root] || this.isInstanceOf[DetachedElementDecl]) {
+      AlignmentMultipleOf(0) // root and detached elements are aligned with anything
```
mbeckerle: Don't understand comment here. Why are detached elements always aligned?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
```diff
@@ -295,7 +295,7 @@ case class CaptureValueLengthStart(ctxt: ElementBase)
       new NadaParser(ctxt.erd)
 
   override lazy val unparser: Unparser =
-    if (ctxt.isReferencedByValueLengthUnparserExpressions) {
+    if (ctxt.isReferencedByValueLengthUnparserExpressions || ctxt.otherwiseShouldCaptureValueRegionLength) {
```
mbeckerle: Are there other reasons for "otherwiseShouldCapture..." than prefixLength? Maybe this should just be shouldCaptureForPrefixLength?.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -187,6 +200,79 @@ trait ElementBaseGrammarMixin
...
+          name, prefixLengthType,
+          detachedElementDecl.prefixLengthType,
+          detachedElementDecl.prefixedLengthElementDecl.prefixLengthType)
+    }
+
+    subset(detachedElementDecl.lengthKind != LengthKind.Prefixed, "Nested dfdl:lengthKind=\"prefixed\" is not supported.")
+
+    detachedElementDecl
+  }
+  final lazy val prefixedLengthAdjustmentInUnits: Long = prefixIncludesPrefixLength match {
```
mbeckerle: So, is this requiring the length prefix itself to be constant length? I think this needs to be an Evaluatable ultimately, since the length prefix simple type isn't necessarily of known fixed length itself.  Seems like an obscure corner case, and it *should* be that, but I think there are formats where the prefix length is a variable-length text number delimited by something.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -50,3 +56,63 @@ trait HasRuntimeExplicitLength {
...
   }
 }
+
+
+trait PrefixedLengthParserMixin {
+
+  def prefixedLengthParser: Parser
+  def prefixedLengthERD: ElementRuntimeData
+  def lengthUnits: LengthUnits
+  def prefixedLengthAdjustmentInUnits: Long
```
mbeckerle: Don't think this can be a Long. I think it has to be an Evaluatable[Long], since the prefix length field itself can be variable length determined at runtime. 

I'm ok with putting that off to a future change. A useful comment about this limitation then would be all.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -50,3 +56,63 @@ trait HasRuntimeExplicitLength {
...
+    }
+    val adjustedLen = parsedLen - prefixedLengthAdjustmentInUnits
+    if (adjustedLen < 0) {
+      state.SDE("Prefixed length result must be non-negative after dfdl:prefixIncludesPrefixLength adjustment , but was: %d", adjustedLen)
+    }
+    adjustedLen
+  }
+
+  /**
+   * Get the prefixe length in bits. If the parse fails, this returns 0. The
```
mbeckerle: typo "prefixe"

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
```diff
@@ -0,0 +1,1820 @@
...
+
+    <xs:simpleType name="nByteTextInt" dfdl:representation="text" dfdl:lengthKind="prefixed" dfdl:prefixLengthType="ex:threeByteTextInt" dfdl:lengthUnits="bytes">
+      <xs:restriction base="xs:integer" />
+    </xs:simpleType>
+
+    <xs:element name="pl_text_string_txt_bytes" type="xs:string" dfdl:lengthKind="prefixed" dfdl:prefixLengthType="ex:threeByteTextInt" dfdl:lengthUnits="bytes" nillable="true" />
+    <xs:element name="pl_text_string_txt_bits" type="xs:string" dfdl:lengthKind="prefixed" dfdl:prefixLengthType="ex:twentyFourBitsTextInt" dfdl:lengthUnits="bits" nillable="true" />
+    <xs:element daf:parseUnparsePolicy="parseOnly" name="pl_text_string_txt_chars" type="xs:string" dfdl:lengthKind="prefixed" dfdl:prefixLengthType="ex:threeCharTextInt" dfdl:lengthUnits="characters" nillable="true" />
+
+    <xs:element name="pl_text_string_txt_bytes_includes" type="xs:string" dfdl:lengthKind="prefixed" dfdl:prefixLengthType="ex:threeByteTextInt" dfdl:lengthUnits="bytes" dfdl:prefixIncludesPrefixLength="yes" nillable="true" />
```
mbeckerle: Because of the line length here, I can't really code review this easily using the github tools. I have no horizontal scroll capability unless I scroll all the way to the bottom of the file.

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
```diff
@@ -0,0 +1,1820 @@
...
+    </xs:simpleType>
+
+    <xs:simpleType name="twentyFourBitsBinInt" dfdl:representation="binary" dfdl:lengthKind="explicit" dfdl:length="24" dfdl:lengthUnits="bits">
+      <xs:restriction base="xs:integer" />
+    </xs:simpleType>
+
+    <xs:simpleType name="threeCharTextInt" dfdl:representation="text" dfdl:lengthKind="explicit" dfdl:length="3" dfdl:lengthUnits="characters" dfdl:textNumberJustification="left" dfdl:textPadKind="padChar" dfdl:textTrimKind="padChar" dfdl:textNumberPadCharacter="X">
+      <xs:restriction base="xs:integer" /> 
+    </xs:simpleType>
+
```
mbeckerle: Ok, so I'm not seeing types for use as length prefix types, that aren't fixed length. I do see one here that looks like it is intended for use with nested prefix length stuff, but nothing with variable length as the prefix length. 

Again I'm fine with that limitation.... unless I'm just missing something... which is why I bring it up.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
stevedlawrence: As I read the spec, alignment doesn't really apply to prefix lengths. When parsing an element that has prefix length, alignment bits are skipped if necessary based on the alignment of the element. Then the prefix length is immediately parsed, then the element is immediately parsed. To ensure no alignment is done, we just pretened that the prefixed element is aligned with everything.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
stevedlawrence: otherwiseShouldCapture wasn't added for prefix length. This was just moved out of the isReferencedByVvalueLength* stuff more consistent. It can be moved back if it makes sense.

mbeckerle: Just a suggestion that all the reasons for why one might lay down the "capture" things in the grammar should either be centralized (here), or distributed in some way. We are halfway between with one case (used in an expression) explicitly mentioned here, the others abstracted away in an "otherwise" method.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberTraits.scala
stevedlawrence: The spec says:

> It is a schema definition error if the xs:simpleType [referenced by prefixLengthType] specifies any of:
> ...
> * dfdl:lengthKind 'explicit' where length is an expression
> ...

So I'm not sure it will ever need to be evaluatable. THe prefix length is always known at schema compile time. Except for the case in nested prefixes, but I think that wil result in lots more changes. I'm not sure how that would work, but I'm not sure an evaluatable will be right way to do it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberTraits.scala
stevedlawrence: fixed

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
stevedlawrence: Mentioned above, but aside for nested prefixs lengths (which isn't supported in this change), prefix types need to be implicit or explicit length and cannot be an expression, so the length must always be fixed.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
mbeckerle: So the use case that needs to work right is just a 32-bit word-aligned prefix-length integer which gives the prefix length of a "normally thought of as" byte-aligned text string. The element itself shouldn't have to specify 32-bit alignment. That should happen because the prefix is a 32-bit aligned integer word.  I believe this is actually a PL/1 data situation. Strings all have prefixes. Those are always aligned words (either 4 bytes or 2 bytes I believe). Let's just assume 4 bytes for the moment. Then two 7-character strings would look like for the first string, 4-byte length, 7-bytes of text,  then for the 2nd string 1 byte of alignment,  4-byte length, 7-bytes of text.

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
stevedlawrence: I can reformat to limit the line length.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
stevedlawrence: I'll add a test to make sure that works (I think it should). The spec does say:

> It is a schema definition error if the xs:simpleType [referenced by prefixLengthType] specifies any of:
>
> ...
> dfdl:alignment other than '1
> ...

So I think the alignment/Units properties needs to be put on the element, but the skipped bits should occur before the prefix length element is parsed.

mbeckerle: I think some of this stuff is wrong in the spec. But let's implement what it says there, and then consider correcting it.

I believe none of the properties on an element having lengthKind 'prefixed' should affect the representation of the prefix itself.  Properties like prefixIncludesPrefixLength affect the interpretation of the value from that prefix, but not the representation of the prefix. 

I believe all  properties that are supposed to apply to the prefix itself must be expressed on a simple type definition. There should be no property combining between the prefix's simple type, i.e., the pseudo-element you create for it, and the element having that prefix type. 

Let's take the case of a string with a 4-byte binary integer as prefix length. The integer has to be 32-bit aligned.

Properties on the element want to be those associated with a textual string, not those for a binary integer. 

Consider a harder case: Consider the prefix length is itself a text integer, giving the length of another text integer. Then if the property dfdl:textNumberJustification and dfdl:textTrimKind, etc. appear on the element, do those apply to the element, the prefix or both. I claim they apply ONLY to the element, NOT to the prefix. Any properties for the prefix have to be expressed on the simple type definition, which will have to be a named type definition someplace if any properties are needed.

stevedlawrence: I don't think there is any property combining going on. Some properties just aren't allowed to be on the prefix type (e.g. alignment,separator, initiators, leading/trailing skip). These all seem related to framing which kindof makes sense since the prefix type is withing the framing of the element. So any framing must comes from the element, and that surrounds both the prefix length and the element. If framing properties were allowed on the prefix element then you would have framing for the prefix inside framing of the elemtn, which seems a bit odd. I imagine this was just to simplify things--prefix lenghts are already kindof complicated.

For the example textNumberJustification/textTrimKind, that works as one would expected (I have tests for this). If those properties are on the the element then it affects the parsing of the element, if they are on the simpleType then it affects the parsing of the prefix length.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
stevedlawrence: Looking back, I think I moved things because I found the names confusing. The name isReferenedByVlueLengthUnparserExpressions should return wether or not something was refrenced by a value length expressions, and that's it. But if something wasn't referenced by a value length function, it could still return true if otherwiseSHouldCaptureValueLengthRegion was true. So the function could return something that seemed against what it said it would return. Perhaps the right fix is to just rename isReferencedByValueLengthUnparserExpression to shouldCaptureValueLengthRegionForUnparser or something? Then it does make more sense to have the otherwise stuff in there as well?

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/PropertyGenerator.scala
```diff
@@ -493,9 +493,18 @@ object Currency {
   }
 
   def generateNonEnumInstantiation(propName: String, typeName: String) = {
-    val midTemplate = """  def EUR = convertToTYPE(cacheProperty("EUR").value)
-"""
     val converterName = getConverterTypeName(typeName)
+    val midTemplate = if (converterName != "QName") {
```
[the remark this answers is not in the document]
stevedlawrence: They are sortof related. The dfdl:prefixLengthType property is a QName. Without this conversion, call element.prefixLengthProperty returned a String, and then I had to do a QName converion to be able to find the simpelType. I figured it better to do the conversion as part of the code generator so that all QName conversions can happen in one place and we don't have to remember how to convert Strings to QNames.

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
stevedlawrence: Done with a new fixup commit. Added some new defineFormat's that could be removed, reducing a lot of duplication. Wrapped lines to about 100, which seems to be about the magic number of characters that github will show on a review.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberTraits.scala
mbeckerle: Ok, looks like your code is correct. 

Hmm. I just reread the spec. It is narrower than I recall. It prohibits, for example this rep:

     123|...string of length 123 bytes...

Where the 123 is delimited by the "|". I believe I've actually seen this somewhere.
It does allow

    123...string of length 123...

where the 123 is a fixed-length text string of length 3 bytes. 

Based on what it says in the spec. the only way the prefix itself can be variable length is if it itself is this nested-prefix-length thing where the prefix length itself has  a prefix length. There's exactly one case of that in the world and it's related to an ASN.1 feature. 

I suspect this is a mistake in the spec. In general the prefix fields themselves can be variable length.
At least lengthKind 'delimited' is needed, and the special case of the prefix itself having a prefix is supposed to the "wierd thing" that can happen in addition to all the other ways the prefix itself could be variable length.

But that will have to get fixed in a rev of the DFDL spec. For now what you have seems to match the spec. so good enough.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
stevedlawrence: Fixed with latest commit.

PR #144, opened 2018-12-07
Replace OracleJDK7 with OpenJDK7

stevedlawrence:
- TravisCI no longer supports the former

DAFFODIL-2025


stevedlawrence:
Turns out things still don't work with OpenJDK. TravisCI is able to install that version, but our version of SBT depends on Scala 2.12, which doesn't work on OpenJDK 7. This isn't surprising since Scala 2.12 required Java 8, so the question is why did it work at all on Oracle JDK 7 before? Not sure it's woth the effort to figure it out--Java 7 has been EOL for few years. So I think we should just remove Java7 from TravisCI.

Note that the jars we cross compile with Scala 2.11 shoudl still work on Java 7, so Daffodil can still run on older stystems, we just can't perform automated nightlies. So perhaps we can say it should work on Java 7, but it's not officially supported anymore.

PR #145, opened 2018-12-10
Bump SBT version to 1.2.7

stevedlawrence:
- SBT 1.2.7 fixes the issue where Javadocs do not generate correctly in Java 9+

DAFFODIL-2002


mbeckerle:
So we have a few things that are annoying given the proliferation of projects. One is this number, the other is the scala compiler version number. These need to be centralized, or mostly so. Can one just set this in one place?

stevedlawrence:
Do you mean how all the different DFDL schemas also need to include an sbt and scala version number? I don't think there is a great way around that. The sbt version number definiately can't be centralized, since the build.properties file has a very simple format and is parsed very earily in the sbt startup process.

The scala version can be a little more dynamic, but I don't think there is a way to query the scala version that Daffodil uses. The best we could probably do is create an SBT plugin that the DFDL Schema projects use, and that plugin could be manually updated to match the Daffodil version. Then it just becomes a matter of updating that plugin. But that's kindof odd behavior for a plugin. It doesn't really do anything, just sets a bunch of default parameters. And it still requires people to update the plugin version to get the new settings. So still a somewhat manual process.

I'm not sure there's a great solution to that problem. When probably just need to remember to update scala and sbt versions when we update the daffodil dependency.

PR #146, opened 2018-12-10
Fix hexBinary with non-byte lengths or non-byte boundaries

stevedlawrence:
- Elements with xs:hexBinary type no longer take into account byte order
  when reading bytes. Previously, Daffodil used an algorithm where the
  hex binary was essentially the hex binary representation of the
  non-negative integer of the same bit length. This is wrong and not
  compatible with the specification (for example, the little endian hex
  binary how the bytes in the reverse order). This changes Daffodil so
  it now reads eight bits at time and converts each eight bits to hex
  binary digits. The eights bits are read according to the bitOrder, and
  if less than eight bits are available, zero bit padding is applied to
  either the left or right according to the bit order to reach a
  full byte.
- This is a non-backwards compatible change, but makes Daffodil now
  compatible with the DFDL specification. Any hexBinary element that
  either had little endian byte order or were non-byte size lengths
  will now have different output.

DAFFODIL-2019

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
@@ -22,6 +22,7 @@ import java.lang.{ Long => JLong }
 import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.processors.LengthInBitsEv
 import org.apache.daffodil.processors.Processor
+import org.apache.daffodil.schema.annotation.props.gen.BitOrder
 import org.apache.daffodil.schema.annotation.props.gen.LengthUnits
 import org.apache.daffodil.util.Numbers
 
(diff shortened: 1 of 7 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
```diff
@@ -36,7 +37,30 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
     } else if (!dis.isDefinedForLength(nBits)) {
       PENotEnoughBits(start, nBits, dis.remainingBits)
     } else {
-      val array = start.dataInputStream.getByteArray(nBits, start)
+      // We cannot use dataInputStream.getByteArray(nBits) because that
+      // function takes into account byteOrder. hexBinary should completely
+      // ignore byteOrder, only taking into account bitOrder. To do this, we
+      // want to just repeatedly read 8 bit's at a time and add them to an
+      // ByteArray, effecitively ignoring bytOrder.
```
mbeckerle: Typo "bytOrder"

stevedlawrence: Fixed.

PR #147, opened 2018-12-10
DAFFODIL-20 - All ibm-provided tests run, or are specific to other tickets.

mbeckerle:
Resolving this ticket, as tests that are remaining are covered by
other tickets.

All the ibm-provided testsn now work excepting a 4 that need missing
features. 

There are minor incompatibilities with IBM DFDL. Not documented in this
change history. 

DAFFODIL-20


daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
```diff
@@ -139,7 +143,7 @@ class TresysTests {
 
   @Test def test_nested_group_refs1() { runnerNG.runOneTest("nestedGroupRefs1") }
 
-  @Test def test_AF000() { runnerAF.runOneTest("AF000") }
+  @Test def test_AF000() { runnerAF.trace.runOneTest("AF000") }
```
stevedlawrence: Remove the trace call here. I believe this will ennable debugging for all runnerAF tests, since trace mutates the state?

PR #148, opened 2018-12-11
Daffodil 2032 eclipse

mbeckerle:
Improves eclipse project setup to have two new eclipse projects.

daffodil-test-ibm1-crossTest
daffodil-test-crossTest

These share source directories, but have separate classpaths so that one can conveniently run either Daffodil TDML tests, or cross tests against the IBM DFDL Cross Tester without having to change contexts. Just run the test from the project with the tester on classpath you want.


daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
```diff
@@ -143,7 +143,7 @@ class TresysTests {
 
   @Test def test_nested_group_refs1() { runnerNG.runOneTest("nestedGroupRefs1") }
 
-  @Test def test_AF000() { runnerAF.runOneTest("AF000") }
+  @Test def test_AF000() { runnerAF.trace.runOneTest("AF000") }
```
stevedlawrence: Remove trace.

PR #149, opened 2018-12-11
Daffodil 2034 pools

mbeckerle:
We were getting pool interactions between tests because TLRegisterPool used by the DFA/Parser stuff was a threadLocal. It wasn't cleaned up on Assertion failures and other throws like that but then the thread would get reused for a later test, and the pool would be corrupted. So you'd get this false test failure.

Removed this use of ThreadLocal, now just keeps it on the PState/UState object, which is a per-thread state block. 

The tests no longer interact. 

The interacting tests were visible if you modified the global GeneralFormat to have encodingErrorPolicy="error". Then some tests like test_hexBinary_unparse_14 will fail with Assert.nyi, but later tests (15, and 19) would fail with pool leaks on the RegisterPool. With the change this no longer happens.


mbeckerle:
Mistake setting up. Too many commits pulled together here.

PR #151, opened 2018-12-12
Add tunables for textBidi and floating properties

jadams-tresys:
Many schemas may not have these rather uncommon properties defined, but
they are technically required and according to the DFDL spec missing
properties should not be defaulted. As a stop gap solution tunables have
been added for these properties that when enabled will cause an SDE if
the properties are not defined. If they are not enabled an SDW will be
shown about the missing property.

DAFFODIL-1720

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -94,6 +94,7 @@ trait ElementBase
   requiredEvaluations(if (hasFractionDigits) fractionDigits)
   requiredEvaluations(erd.preSerialization)
   requiredEvaluations(checkForAlignmentAmbiguity)
+  requiredEvaluations(checkFloating)
 
   def name: String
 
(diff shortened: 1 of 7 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -1233,4 +1234,12 @@ trait ElementBase
     }
   }
 
+  private lazy val optionFloating = findPropertyOption("floating")
+
+  private def checkFloating = (optionFloating.isDefined, tunable.requireFloatingProperty) match {
+    case (false, false) => SDW(WarnID.FloatingError, "Property 'dfdl:floating' is required but not defined.")
+    case (false, true) => SDE("Property 'dfdl:floating' is required but not defined.")
```
stevedlawrence: Might be worth replacing this SDE with a call to "floating". That call would require the floating property and output the standard property missing SDE, which also shows some context like where it looked for the property. Same with textBidi below.

PR #152, opened 2018-12-12
Daffodil 2028 cross test

mbeckerle:
This is a partial step that improves portability of daffodil-test, but is not completing the job yet. 
But this much can be changed without any large-scale regressions, so is worth reviiew separately.
    
    Changed explicit use of encodingErrorPolicy in tests to
    encodingErrorPolicy="error" for all tests except those
    explicitly testing replace policy, and those are now restricted to
    implementation 'daffodil'.
    
    Changed explicit use of calendarTimeZone="" to calendarTimeZone="UTC" or other except
    for tests that are explicitly testing for no time zone. Those tests are
    implementation 'daffodil' only.
    
    Did NOT change GeneralFormat. A new format GeneralFormatForCrossTesting
    is available which has encodingErrorPolicy 'error' and
    calendarTimeZone='UTC' for future convenience. But this was not put in place
   across daffodil-test tests yet, because of 135 regressions that it causes. 
    
   TDML Runner doesn't cross test negative tests by default.
    
    It still runs the tests, just doesn't compare the messages.
    So the test still has to fail, just without scrutinizing the diagnostic
    messages.
    
By passing parameters to DFDLTestCase you can override this and create
tests which are designed to verify cross-test error message text or
warning text.
    
Also doesn't verify warnings on cross tests (negative or positive)
unless parameters are set.
      
Portable tests should use GeneralFormatForCrossTesting
This was NOT put in place across daffodil-test however, as it causes 135 tests to break (11 due to encodingErrorPolicy='error', and the rest due to time-zone strings needing to appear in infoset due to calendarTimeZone="UTC". Those changes are not in this PR, as they should be done in conjunction with also fixing DAFFODIL-2026 which means not just adding the TZ strings, but also xsi:type attributes so tests can do type-specific comparison. 

So this is a partial step that improves portability of daffodil-test, but is not completing the job yet.


daffodil-lib/src/main/resources/edu/illinois/ncsa/daffodil/xsd/built-in-formats.xsd
```diff
@@ -61,7 +61,7 @@
 				gets you only the new properties without all the property settings of daffodilTest1. -->
 
 			<dfdl:defineFormat name="daffodilNewProperties1">
-				<dfdl:format encodingErrorPolicy="replace" sequenceKind="ordered"
+				<dfdl:format encodingErrorPolicy="error" sequenceKind="ordered"
```
mbeckerle: Revert. We don't use this format. It was kept in for backward compat reasons. So we shouldn't be changing it.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
```diff
@@ -67,7 +67,20 @@ object Runner {
...
-  def defaultImplementationsDefaultDefault = Seq("daffodil")
+  def defaultImplementationsDefaultDefault = Seq("daffodil", "ibm")
+
+  /**
+   * By default we don't run Daffodil negative TDML tests against cross-testers.
+   * The error messages are simply too varied.
+   *
+   * Negative tests must fail, but error messages aren't compared.
+   */
+  def defaultShouldCrossTestNegativeTests = false
```
mbeckerle: rename? shouldCrossTestErrorComparisons

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
```diff
@@ -67,7 +67,20 @@ object Runner {
...
+  
+  /**
+   * By default we don't cross test warning messages because they are too varied.
+   */
+  def defaultShouldCrossTestWarnings = false
```
mbeckerle: rename? shouldCrossTestCompareWarnings

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/nested-separator-delimited.tdml
```diff
@@ -21,7 +21,7 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
   xmlns:xs="http://www.w3.org/2001/XMLSchema" 
   xmlns:ex="http://example.com" xmlns="http://example.com"
-  defaultImplementations="ibm dfdl">
+  defaultImplementations="ibm daffodil">
```
mbeckerle: Might as well add validation to the tdml XSD so that this mistake can't be made. 

We've already chosen to "own" the names of these tokens. I.e., they're not class names anymore.

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PatternTests.tdml
```diff
@@ -615,6 +597,37 @@
     </tdml:errors>
   </tdml:parserTestCase>
 
+<tdml:defineSchema name="DFDL-207-Replace">
```
mbeckerle: rename?  DFDL-207-non-portable

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
```diff
@@ -93,6 +101,17 @@
...
+      <dfdl:defineFormat name="GeneralFormat">
+        <dfdl:format ref="GeneralFormatOriginal" />
+      </dfdl:defineFormat>
+        
+      <dfdl:defineFormat name="GeneralFormatForCrossTesting">
+        <dfdl:format ref="GeneralFormatOriginal"
+          calendarTimeZone="UTC"
+          encodingErrorPolicy="error"
+        />   
+      </dfdl:defineFormat>
```
stevedlawrence: Wondering if this new format should go in src/test/resources somewhere if it really is just for testing purposes? My concen is that anything in this file is probably something we should officially support and might want to worry about backward compatability issues. If it's in src/test/resources, we can changeit whenever we want.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -164,6 +164,9 @@ private[tdml] object DFDLTestSuite {
  *
  * defaultImplementationsDefault the implementations default for the test suite will be
  * taken from this value if it is not specified on the testSuite itself.
+ *
+ * shouldCrossTestNegativeTests controls whether negative test error messages  are compared
+ * or the tests are just run to determine that they fail.
```
stevedlawrence: Add comments about the warning flag too.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
mbeckerle: Only hassle is then to use it you have to include a different file. There's 500+ include statements to change (eventually). As opposed to 500+ ref="GeneralFormat" statements. One would need to change both. 

Maybe this shouldn't be GeneralFormatForCrossTesting, but GeneralFormatPortable ?

stevedlawrence: Fair. A name change seems like a better solution.

daffodil-lib/src/main/resources/edu/illinois/ncsa/daffodil/xsd/built-in-formats.xsd
jadams-tresys: I'm a little confused about the errorEncodingPolicy stuff.  Are we changing most uses of it to "error", which we don't currently support and then relying on Daffodil switching it back to "replace" when the tests run?

Just want to make sure I am understanding what is going on.

PR #153, opened 2018-12-12
Add a distinction between how expressions and literals are compiled

stevedlawrence:
Sometimes, we compile non-expressions just to benefit from the value
conversion capabilities of DPath. For example, if a property was either
"{ 1.234 }" or just "1.234", compilation would still convert it to the
correct float value, regardless if there were curly braces or not.

This is flexible, but can lead to unexpected behavior, specifically with
the defaultValue attribute in dfdl:defineVariable. This attribute can
accept either an expression or a literal value, so we use the above
logic to compile the expression or literval value to an expression. But
this does not always work as expected. For example, if the
defineVariable type was an xs:boolean, the defaultAttribute could accept
either "{ fn:false() }" or "false". Note that "false" is not a valid
expression, so if we try to compile this literval value we get an error.
So our current behavior of compiling literals as if they were
expressions without curly braces prevents the correct use of literals in
some cases.

To resolve this, we need to make a clear distinction between how to
handle expression and literals. With this change, a property value is
now only considered an expression and compiled if it starts with an
unescaped curly brace. All other values are treated as literal values,
the appropriate type conversion is performed by the fromXMLString()
method of the primitive type, and a ConstantExpression is created.
Additionally, the DPath expression grammar is modified to be more strict
and to only accept actual expressions that start and end with a
curly--this should help to prevent accidental compilation of
non-expressions.

This change allows the use of literal values in the defaultValue
attribute in a dfdl:defineVariable element and have them converted
appropriately.

DAFFODIL-2033

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
@@ -18,6 +18,7 @@
 package org.apache.daffodil.dsom
 
 import org.apache.daffodil.dpath._
+import org.apache.daffodil.dpath.NodeInfo.PrimType
 import scala.xml.NamespaceBinding
 import org.apache.daffodil.xml.NamedQName
 import java.lang.{ Long => JLong, Boolean => JBoolean }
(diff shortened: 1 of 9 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -1061,4 +1061,67 @@
...
 
+
+  <tdml:defineSchema name="logical_default_values">
+    
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" />
+
+    <dfdl:defineVariable name="float" type="xs:float" defaultValue="-INF" />
+    <dfdl:defineVariable name="double" type="xs:double" defaultValue="1.0e-5" />
+    <dfdl:defineVariable name="bool" type="xs:boolean" defaultValue="false" />
```
mbeckerle: Are 'false' and 'true' actually values?  Does the DFDL spec really allow defaultValue="false" for booleans?
I'm not against this. I'm just not sure where the DFDL spec blesses a syntax for boolean values.

Maybe this is missing from the DFDL spec? 

I know there is no syntax for specifying nil, for example. There's no way to have an expression on say, dfdl:inputValueCalc cause a nillable element to be nilled. So there are definitely limitations on the expression language's capabilities. 

In actual expressions, you have to use fn:true() and fn:false(). The literal value syntax of "true" and "false" isn't allowed. 

I know for a boolean element the XSD attribute default="false" would be allowed, but that's not the same language of literals as DFDL string literals. E.g., in XSD attribute default="....", you can't use DFDL entities.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
```diff
@@ -91,26 +130,45 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
...
+      host)
+
+    if (compiled1.isConstant || (staticNodeInfoKind == runtimeNodeInfoKind)) {
+      compiled1
+    } else {
+      //
+      // TODO: consider passing in a flag or some other way of avoiding this
+      // duplicate compile run.
+
+      // This is, this nodeInfo.Kind is used as the target type in the DPath expression compiler, and
```
jadams-tresys: Not sure what is going on with this comment line

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
stevedlawrence: This came up in a review on changes needed to get ISO8583 to work with Daffodil:

https://github.com/DFDLSchemas/ISO8583/pull/2#pullrequestreview-183819263

You'll need to click "show resolved" to show the conversation about the defaultValue="false".

Section 7.7 of the spec says that a logical value is allowed as the defaultValue. I'm not sure the spec is clear on what is meant by a "logical value", but Steve Hanson did state in the review that "false" should be allowed.

For this implementation, this change requires that the logical type be parsable by PrimType.fromXMLString. In the boolean case that's just Scala's string.toBoolean method, which means it will accept "true" or "false" ignoring case. Note that PrimeType.fromXMLString is what we use for converting XML Infoset strings to the internal types, so in theory this allows anything that XSD would allow for a type. Seems like a reasonable interpretation of logical type. So for example, for xs:hexBinary you could do ``defaultValue="DEADBEEF"``.

Regarding inputValueCalc, this change would potentially allow something like dfd:inputValueCalc="false" to work if the type was xs:boolean, but I don't think that's new behavior. That would have worked before this change except that there are other checks  that ensure that inputValueCalc is an actual expression, and the above wouldn't be allowed.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
stevedlawrence: No idea, but it doesn't provide anything useful. WIll remove

PR #154, opened 2018-12-12
Allow lengthUnits to differ for element and its prefixed length

stevedlawrence:
Except in the case when prefixIncludesPrefixLength="yes". The spec does
not contain this excpetion so reads as if lengthUnits must always be the
same. Errata 2.76 clarifies that this should be allowed.

DAFFODIL-2035

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -207,48 +207,73 @@ trait ElementBaseGrammarMixin
     // because we need to create a detached element with the same schema
     // document/parent of the GSTD.
     val prefixLengthTypeGSTD = schemaSet.getGlobalSimpleTypeDef(prefixLengthType).getOrElse(
-      schemaDefinitionError("Failed to resolve dfdl:prefixLengthType=\"%s\" to a simpleType", prefixLengthType.toQNameString)
+      schemaDefinitionError(
+        "Failed to resolve dfdl:prefixLengthType=\"%s\" to a simpleType",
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -219,8 +219,8 @@ trait ElementBaseGrammarMixin
         schemaDefinitionError("%s is specified as a dfdl:prefixLengthType, but has a dfdl:lengthKind of %s", prefixLengthType, prefixedLengthKind)
       case LengthKind.Explicit if detachedElementDecl.optLengthConstant.isEmpty =>
         schemaDefinitionError("%s is specified as a dfdl:prefixLengthType, but has an expression for dfdl:length", prefixLengthType)
-      case LengthKind.Implicit | LengthKind.Explicit if lengthUnits != detachedElementDecl.lengthUnits =>
-        schemaDefinitionError("%s is specified as a dfdl:prefixLengthType with dfdl:lengthKind %s, but has different dfdl:lengthUnits than the element", prefixLengthType, prefixedLengthKind)
+      case LengthKind.Implicit | LengthKind.Explicit if prefixIncludesPrefixLength == YesNo.Yes && lengthUnits != detachedElementDecl.lengthUnits =>
+        schemaDefinitionError("%s is specified as a dfdl:prefixLengthType where dfdl:prefixIncludesPrefixLength=\"yes\" with dfdl:lengthKind %s, but has different dfdl:lengthUnits than the element", prefixLengthType, prefixedLengthKind)
```
mbeckerle: Lines are too long. Makes code review hard. 
Ok to re-wrap the whole section now if you want.

PR #155, opened 2018-12-12
Don't require binaryNumberRep for hexBinary

jadams-tresys:
The property binaryNumberRep should not need to be defined for hexBinary
elements.

DAFFODIL-2031

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -657,25 +657,43 @@ trait ElementBase
     (theRepresentation, thePrimType) match {
       case (Representation.Text, PrimType.HexBinary) => Assert.impossible("type xs:hexBinary with representation='text'")
       case (Representation.Text, _) => knownEncodingAlignmentInBits
+      case (Representation.Binary, PrimType.String) => Assert.impossible("type xs:string with representation='binary'")
+      // Boolean, Float, Double, and HexBinary do not require binaryNumberRep to be defined
       case (Representation.Binary, PrimType.Float | PrimType.Boolean) => 32
       case (Representation.Binary, PrimType.Double) => 64
-      case (Representation.Binary, _) => binaryNumberRep match {
-        case BinaryNumberRep.Packed | BinaryNumberRep.Bcd | BinaryNumberRep.Ibm4690Packed => 8
-        case _ => thePrimType match {
-          case PrimType.String => Assert.impossible("type xs:string with representation='binary'")
-          case PrimType.Double | PrimType.Long | PrimType.UnsignedLong => 64
-          case PrimType.Float | PrimType.Int | PrimType.UnsignedInt | PrimType.Boolean => 32
-          case PrimType.Short | PrimType.UnsignedShort => 16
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -653,13 +653,16 @@ trait ElementBase
...
+
   private def getImplicitAlignmentInBits(thePrimType: PrimType, theRepresentation: Representation): Int = {
     (theRepresentation, thePrimType) match {
       case (Representation.Text, PrimType.HexBinary) => Assert.impossible("type xs:hexBinary with representation='text'")
       case (Representation.Text, _) => knownEncodingAlignmentInBits
       case (Representation.Binary, PrimType.Float | PrimType.Boolean) => 32
       case (Representation.Binary, PrimType.Double) => 64
-      case (Representation.Binary, _) => binaryNumberRep match {
+      case (Representation.Binary, PrimType.HexBinary) => 8
+      case (Representation.Binary, _) if (optionBinaryNumberRep.isDefined) => binaryNumberRep match {
```
stevedlawrence: Is the optionBinaryNumberRep.isDefine check needed here? If the prim type is hexBInary it'll hit the new case you added and will never try to get optionBInaryNumberRep, I think? There's also a case PrimType.HexBinary => 8 at the bottom of this match case that I think is impossible to hit with the above change? Hard to tell with this complex nesting of match cases...

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -701,6 +704,7 @@ trait ElementBase
                   alignInBits, implicitAlignmentInBits, primType.name, this.knownEncodingName)
             }
             case Representation.Binary => primType match {
+              case PrimType.HexBinary => 8
```
stevedlawrence: I think the 8 here does nothing? This part of this val is just doing validation like making sure packed decimals are aligned. We don't have that restriction with hex binary since we allow hexbinary to be non-byte sized and non-byte aligned. So I think this can just be a no-op like the Float/Double/Booleans below it?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
jadams-tresys: I think this rats nest of match cases was made by me, so I took a stab at greatly simplifying the logic here and I think it is much cleaner now.  I ran the test suite with these changes and everything passes, but give it a look to make sure I didn't miss anything obvious.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -653,29 +653,29 @@ trait ElementBase
     }
   }
 
+  private lazy val optionBinaryNubmerRep = findPropertyOption("binaryNumberRep")
+  /* All binary number representations other than "binary" are of some packed format */
+  private lazy val isPackedBinaryNumber: Boolean = optionBinaryNumberRep.isDefined && (binaryNumberRep != BinaryNumberRep.Binary)
```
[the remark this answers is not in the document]
jadams-tresys: I have updated the code to not rely on whether or not binaryNumberRep is defined.
I have also tried to tidy up the areas with all the pipes, but I still prefer the use of pipes instead of having several more case statements that evaluate to the same thing.

PR #156, opened 2018-12-14
Added test to verify correct behavior for non-existant groupRef

jadams-tresys:
In version 2.2.0 this test causes an infinite loop to occur, but this
issue has been fixed in one of the commits since 2.2.0's release. This
test simply verifies that the correct error message is displayed when the
group referenced by a hiddenGroupRef does not exist.

DAFFODIL-2020


jadams-tresys:
The underlying issue was fixed by Mike in commit 92429a0fc26673d1df362916862729c03a9c9a0a and some tests were added there so this pull request may not be needed.

mbeckerle:
A more specific test for the non-exist case is always welcome.

See ticket DAFFODIL-284 .... I added a comment near the end about how this was still happening.  

This test is ok. There is still a problem with circular group refs tho.

PR #157, opened 2018-12-14
Add support for type aware checking in the TDML runner

stevedlawrence:
- One can now add xsi:type information to the expected infoset or to the
  generated infoset to have the TDML runner perform type aware
  comparisions. If both the actual and expected infosets provide type
  information, the xsi:type must be the same. If no type information is
  provided in either, string comparison is performed.
- Moved much of the Date/Time conversion code out of the runtime and
  into daffoddil-lib where the XMLUtils that perform comparision live.
  Before is was tied too closely to DPath and made it difficult to use
  outside of that. DFDLDate/Time is no longer coupled to DPath and can
  be used for generic date time tasks.
- Refactored the date/time code logic to be simplier and remove
  duplication and complexity that has grown over the years.
- Modified XMLUtils.compareAreReport make all the necessary
  modifications to compare two infosets. This used to be scattered among
  different users and wasn't consistent about what was actually done.
  This should make future comparisons, such as namespace checking,
  easier to test. Also, comparison no longer does a toString comparison
  first, it always does our custom comparison.
- This does not yet determine which tests could make use of this
  function and change them so to show cross compatbility with IBM. It
  only adds the capability and adds some tests to validate it.

DAFFODIL-2026


daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -716,24 +724,33 @@ object XMLUtils {
...
+    val noMixedWS = removeMixedWhitespace(combinedText)
+    noMixedWS
+  }
+
+  def compareAndReport(expected: Node, actual: Node, ignoreProcInstr: Boolean = true) = {
+    val expectedMinimized = prepareForDiffComparison(expected)
+    val actualMinimized = prepareForDiffComparison(actual)
+    val diffs = XMLUtils.computeDiff(expectedMinimized, actualMinimized, ignoreProcInstr)
+    if (diffs.length > 0) {
+      throw new Exception("""
```
mbeckerle: Generic Exception being thrown. Should this be an exception class that indicates by class name that this is a Diff being reported?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions.scala
```diff
@@ -114,10 +114,20 @@ case class DFDLTimeZoneFromDFDLCalendar(recipe: CompiledDPath, argType: NodeInfo
...
+      val offsetInMils = tz.getRawOffset
+      val offsetInMins = Math.abs(offsetInMils / (1000 * 60))
+      val hour = offsetInMins / 60
+      val mins = offsetInMins % 60
+
+      val signStr = if (offsetInMils >= 0) "+" else "-"
+      val hourStr = if (hour < 10) "0" + hour else hour
+      val minsStr = if (mins < 10) "0" + mins else mins
+
+      signStr + hourStr + ":" + minsStr
```
mbeckerle: Right here is where we'd change to use "Z" notation instead of more verbose "+00:00". 

We should consider this. I think people expect it, or at least find it less objectionable than "+00:00".

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
stevedlawrence: Yep, a better Exception makes sense. Will do.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions.scala
stevedlawrence: This is actually for the output of the ``dfdl:timeZoneFrom{Date,TIme,DateTIme}`` XPath functions, which according to the spec should always output +00:00 for the GMT time.

The place to change what we output to an infoset is in DFDLCalendarConversions.scala, which uses some SimpleDateFormats for outputting the timezone. So we'd just need to change thepattern from 'xxx' to 'XXX'.

That said, this conversion code should probably be in the DFDLCalendarConversions stuff. And it probably would be a good idea to stop using SimpleDateFormats for creating the output. SimpleDateFOrmat's are not thread safe and way heavier than what we need. I don't think it would be too difficult to do something like the above for dates and times too. I'll refactor that toXMLString stuff to not require SImpelDateTIme's and todo somethine more like above.

mbeckerle:
These are the tests that should work under cross testing with this new feature implemented:

test_simple_type_properties_text_calendar_13_03(org.apache.daffodil.IBMTestsThatPass)
test_simple_type_properties_text_calendar_13_04(org.apache.daffodil.IBMTestsThatPass)
test_simple_type_properties_bin_calendar_13_01(org.apache.daffodil.IBMTestsThatPass)
test_simple_type_properties_bin_calendar_13_02(org.apache.daffodil.IBMTestsThatPass)
test_simple_type_properties_text_calendar_13_01(org.apache.daffodil.IBMTestsThatPass2)
test_alignment_bytes_12_05(org.apache.daffodil.IBMTestsThatPass2)

stevedlawrence:
Added new fixup commit that removes the SimpleDateTime when creating the infoset. Avoids ThreadLocals and any overhead related to SimpleDateTIme.

Confirmed that the above tests now pass with the IBM Cross tester, plus the following tests:
* test_length_delimited_12_04
* test_length_delimited_12_01
* test_simple_type_properties_text_calendar_13_02 (not compatable)

All tests in IBMTestsThatPass pass with IBM and all but one in IBMTestsThatPass2 pass. The one failure is unrelated to type awareness.

PR #158, opened 2018-12-15
Daffodil 1080 sequences and separators - preliminary review

mbeckerle:
This is a checkpoint in work to improve Sequences and Separators.
I do not intend to merge this branch; however, review would be helpful perhaps.
There are some difficult issues here. 
    
    Tests that distinguish proper function in Daffodil, proper function in
    IBM DFDL, failures of each, and differences between them added.
    
    The tests that fail (either implementation) are in scala-debug.
    
    Parsing is working better than unparsing. Unparsing is failing in a
    number
    of important scenarios still.
    
    There is much duplication of test-case logic in order to provide
    distinct parse/unparse and daffodil/ibm variants of a test.
    
    In many cases there are 4 copies of a test. Ex: for a test named "foo":
       fooN_Np_daf
       fooN_Np_ibm
       fooN_Nu_daf
       fooN_Nu_ibm
    In these cases the test document data, and test infoset are IDENTICAL
    AND SHOULD BE KEPT THAT WAY. Eventually as we improve things we would
    hope
    to collapse these back to a single, portable and round-trip test.
    
    (alternatively, TDML Runner features allowing a single test to be
    expressed by drawing in infoset and document from shared definitions,
    would be helpful. That way one could have 4 tests, but have them
    explicitly share the same data document definition and infoset
    definition.
    You can do this today, but only by putting the infoset and document part
    into external files. I'd like to keep this self contained.)
    
    For now this split up allow us to have tests which work only some of
    the 4 possible combinations, and to split those up over scala vs.
    scala-debug.
    
    The "_ibm" tests are to understand and characterize the behavior of
    that implementation that we need to interoperate with, because in some
    cases this behavior is NOT conforming to current DFDL specification
    due to well known bugs and limitations.
    
    == Affect of changes so far on general function/regression tests ==
    
    1 regression in daffodil-test-ibm1 - AX000 fails. Complex situation at
    first glance.
    
    1 regression in daffodil-tdml-processor tests - the 'threePass' test
    doesn't work because it was depending on an empty element being created
    that is now suppressed. See below discussion.
    
    36 regressions in daffodil-test.
    A substantial number of these are due to empty elements that we were
    creating, but no longer are, and some are nilled elements we used to
    create but no longer are.
    
    There are some other failure scenarios also however.
    
    Right now the code does too much suppression of elements having zero
    length. They are not supposed to be suppressed if ZL is the nil
    representation nor if ZL is the "empty string" or "empty hexBinary"
    representation.
    
    Daffodil should implement the correct DFDL-spec-compliant behavior,
    and did prior to these changes. In order to get interop with IBM DFDL
    on some of the new EDI-style tests, I changed behavior for string and
    hexbinary elements so that they are more often suppressed as elements.
    
    However, IBM DFDL does not implement this functionality according to
    the DFDL spec, so interoperability with IBM DFDL will require a flag
    that changes this behavior. Many of these changes were added to DFDL
    spec several years ago before the 2014 revision of DFDL v1.0. But
    IBM's DFDL implementation was already in operation, so they have a
    legacy of behavior they'll have to eventually adjust. This big set of
    subtle revisions were associated with the DFDL Workgroup "Action 140"
    work/action item. Basically IBM DFDL did not implement (some ? or all
    of) the Action 140 changes.
    
    So we either need a flag, or alternatively, 36 tests have to be made
    Daffodil specific - which may be preferable. This depends on whether
    the behavior is required to make "real" schemas work portably, or just
    little test cases.
    
    DAFFODIL-1080

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
@@ -47,7 +47,7 @@ object ModelGroupFactory {
    * object. There should be only one non-Nil.
    */
   def apply(child: Node, parent: SchemaComponent, position: Int, isHidden: Boolean,
-    nodesAlreadyTrying: Set[Node] = Set()): List[ModelGroup] = {
+            nodesAlreadyTrying: Set[Node] = Set()): List[ModelGroup] = {
     if (nodesAlreadyTrying.contains(child)) {
       //
       // We are chasing our tail. Circular reference among named model groups/terms.
@@ -74,8 +74,8 @@ object ModelGroupFactory {
         case <group>{ _* }</group> => {
           val pos = parent match {
             case ct: ComplexTypeBase => 1
-            case mg: ModelGroup => position
-            case gd: GlobalGroupDef => position
+            case mg: ModelGroup      => position
+            case gd: GlobalGroupDef  => position
           }
           val isH = isHidden || parent.isHidden
           val groupRefFactory = new GroupRefFactory(child, parent, pos, isH)
@@ -83,8 +83,8 @@ object ModelGroupFactory {
           List(groupRefInstance.asModelGroup)
         }
         case <annotation>{ _* }</annotation> => Nil
-        case textNode: Text => Nil
-        case _: Comment => Nil
+        case textNode: Text                  => Nil
+        case _: Comment                      => Nil
         case _ => {
           parent.SDE("Unrecognized construct: %s", child)
         }
@@ -123,8 +123,8 @@ object TermFactory {
         }
       }
       case <annotation>{ _* }</annotation> => Nil
-      case textNode: Text => Nil
-      case _ => ModelGroupFactory(child, parent, position, false, nodesAlreadyTrying)
+      case textNode: Text                  => Nil
+      case _                               => ModelGroupFactory(child, parent, position, false, nodesAlreadyTrying)
     }
     childList
   }
@@ -144,7 +144,19 @@ abstract class ModelGroup
 
   requiredEvaluations(groupMembers)
 
+  final lazy val hasFraming =
+    hasInitiator ||
+      hasTerminator ||
+      !hasNoSkipRegions
+
+  final lazy val hasStaticallyRequiredOccurrencesInDataRepresentation = {
+    hasFraming ||
+      // or if all arms of the choice have statically required instances.
+      groupMembers.forall { _.hasStaticallyRequiredOccurrencesInDataRepresentation }
+  }
+
   def groupMembers: Seq[Term]
+
   def xmlChildren: Seq[Node]
   protected def myPeers: Option[Seq[ModelGroup]]
 
@@ -174,11 +186,11 @@ abstract class ModelGroup
   override lazy val alignmentValueInBits: JInt = {
     this.alignment match {
       case AlignmentType.Implicit => this.alignmentUnits match {
-        case AlignmentUnits.Bits => 1
+        case AlignmentUnits.Bits  => 1
         case AlignmentUnits.Bytes => 8
       }
       case align: JInt => this.alignmentUnits match {
-        case AlignmentUnits.Bits => align
+        case AlignmentUnits.Bits  => align
         case AlignmentUnits.Bytes => 8 * align
       }
     }
@@ -189,7 +201,7 @@ abstract class ModelGroup
     val echls = gms.flatMap { gm =>
       gm match {
         case eb: ElementBase => Seq(eb)
-        case gb: ModelGroup => gb.elementChildren
+        case gb: ModelGroup  => gb.elementChildren
       }
     }
     echls
@@ -204,7 +216,7 @@ abstract class ModelGroup
       case mg: ModelGroup => mg.groupMembers.map {
         _ match {
           case eb: ElementBase => eb.erd
-          case t: Term => t.termRuntimeData
+          case t: Term         => t.termRuntimeData
         }
       }
       case _ => Nil
@@ -229,14 +241,14 @@ abstract class ModelGroup
    */
   lazy val potentialLastChildren: (Seq[Term], Boolean) = {
     val (potentialLast, allOptional) = this match {
-      case ch: ChoiceTermBase => (ch.groupMembers, false)
+      case ch: ChoiceTermBase                    => (ch.groupMembers, false)
       case sq: SequenceTermBase if !sq.isOrdered => (sq.groupMembers, true) // TBD: is true correct? Are all children optional in unordered sequence?
       case sq: SequenceTermBase => {
         val maybeLast = sq.groupMembers.lastOption
         if (maybeLast.isDefined) {
           val last = maybeLast.get
           val lastIsOptional = last match {
-            case mg: ModelGroup => false // model group is mandatory
+            case mg: ModelGroup  => false // model group is mandatory
             case eb: ElementBase => !eb.isRequiredInInfoset || !eb.isRepresented
           }
           if (lastIsOptional) {
@@ -253,7 +265,7 @@ abstract class ModelGroup
     val potentialLastRepresented = potentialLast.filter { term =>
       term match {
         case eb: ElementBase => eb.isRepresented
-        case _ => true
+        case _               => true
       }
     }
     (potentialLastRepresented, allOptional)
@@ -264,8 +276,8 @@ abstract class ModelGroup
       val listOfTerms = groupMembers.map(m => {
         m match {
           case e: ElementBase if !e.isRequiredInInfoset => (Seq(e) ++ e.possibleNextTerms) // A LocalElement or ElementRef
-          case e: ElementBase => Seq(e)
-          case mg: ModelGroup => Seq(mg)
+          case e: ElementBase                           => Seq(e)
+          case mg: ModelGroup                           => Seq(mg)
         }
       }).flatten
       listOfTerms
@@ -320,7 +332,7 @@ abstract class ModelGroup
    */
   protected final def possibleFirstChildTerms: Seq[Term] = LV('possibleFirstChildTerms) {
     val firstTerms = this match {
-      case c: ChoiceTermBase => groupMembers
+      case c: ChoiceTermBase                   => groupMembers
       case s: SequenceTermBase if !s.isOrdered => groupMembers
       case s: SequenceTermBase => {
         groupMembers.headOption match {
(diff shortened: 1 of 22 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -74,17 +74,17 @@ object ModelGroupFactory {
         case <group>{ _* }</group> => {
           val pos = parent match {
             case ct: ComplexTypeBase => 1
-            case mg: ModelGroup => position
-            case gd: GlobalGroupDef => position
+            case mg: ModelGroup      => position
```
mbeckerle: Undo whitespace changes. Must have had an incorrect setting when this was last saved.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -82,8 +81,7 @@ abstract class SequenceTermBase(
  * has the sequence properties. So actual sequences, group refs to them,
  * but NOT implied sequences inside choice branches.
  */
-abstract class SequenceGroupTermBase(
-  xml: Node,
+abstract class SequenceGroupTermBase(xml: Node,
```
mbeckerle: This whitespace change I can't do anything about. There's no setting for "leave it alone unless I touch it". 
If the setting leaves the first one alone and indents the 2nd and subsequent args, then the auto formatter brings the first arg up onto the line with the "(" if it fits. ... If I use the setting that puts it on the 2nd line, then they ALL, throughout the file, get converted to that.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -434,43 +425,84 @@ trait Term
...
-          }
+  final lazy val isPotentiallyTrailing: Boolean = LV('isPotentiallyTrailing) {
+    Assert.usage(isSequenceChild || parent.isInstanceOf[ComplexTypeBase])
+    val res =
+      couldBePotentiallyTrailing &&
+        this.laterSiblings.forall { _.isPotentiallyTrailing }
+    res
+  }.value
+
+  //      (isRequiredInInfoset && isLastDeclaredRepresentedInSequence && isVariableOccurrences)) {
```
mbeckerle: Remove commented code.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -112,8 +113,7 @@ trait Separated { self: SequenceChildUnparser =>
   val childProcessors = Vector(childUnparser, sep)
 }
 
-class ScalarOrderedSeparatedSequenceChildUnparser(
-  childUnparser: Unparser,
+sealed abstract class ScalarOrderedSeparatedSequenceChildUnparserBase(childUnparser: Unparser,
```
mbeckerle: If this class is unsealed.... then... see next comment

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -126,8 +126,26 @@ class ScalarOrderedSeparatedSequenceChildUnparser(
...
+class ScalarOrderedSeparatedSequenceChildUnparser(childUnparser: Unparser,
+  srd: SequenceRuntimeData,
+  trd: TermRuntimeData,
+  sep: Unparser,
+  spos: SeparatorPosition,
+  ssp: SeparatorSuppressionPolicy,
+  ssAlgorithm: SeparatorSuppressionMode)
+  extends ScalarOrderedSeparatedSequenceChildUnparserBase(childUnparser, srd, trd, sep, spos, ssp, ssAlgorithm)
+
+class PotentiallyTrailingGroupSeparatedSequenceChildUnparser(childUnparser: Unparser,
```
mbeckerle: This class could be eliminated since it is identical to the preceeding. Where it is constructed could just do the 

new OtherClass with MixinTrait

trick. Not sure how much better I like that, but I hate these classes that just serve to forward all the constructor params.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
```diff
@@ -35,9 +35,8 @@ trait Separated { self: SequenceChildParser =>
 sealed trait RepeatingSeparatedPoU extends Separated { self: RepeatingChildParser =>
 
   private lazy val shouldRemoveZLStringHexBinaryValue_ =
-    false && // disable this feature as many test in daffodil-core depend on NOT removing these empty strings.
+    true ||
```
mbeckerle: This is where the more subtle code has to go, so that we implement the DFDL spec - and have a IBM compatible mode if needed.  IBM agressively removes zero-length strings/hexBinaries. The DFDL spec was clarified to make creation of elements have priority over ZL removal if - ZL is the nil representation or ZL is the empty string or empty hexBinary representation, i.e., if there's no syntax or emptyValueDelimiterPolicy to tell it that empty string should have distinguishing syntax, so ZL should NOT become a value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceParsers.scala
```diff
@@ -146,8 +143,24 @@ final class ScalarOrderedSeparatedSequenceChildParser(
...
-  childParser: Parser,
+final class ScalarOrderedSeparatedSequenceChildParser(childParser: Parser,
+  srd: SequenceRuntimeData,
+  trd: TermRuntimeData,
+  sep: Parser,
+  spos: SeparatorPosition,
+  ssp: SeparatorSuppressionPolicy)
+  extends ScalarOrderedSeparatedSequenceChildParserBase(childParser, srd, trd, sep, spos, ssp)
+
+final class PotentiallyTrailingGroupSeparatedSequenceChildParser(childParser: Parser,
```
mbeckerle: As above. This class could be eliminated. See corresponding comments about Unparser side.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/sepSuppression.tdml
```diff
@@ -0,0 +1,970 @@
...
+        unsupportedAttributeFormDefault 
+        encodingErrorPolicyError
+      </daf:suppressSchemaDefinitionWarnings>
+    </daf:tunables>
+  </tdml:defineConfig>
+
+  <tdml:defineSchema name="s1">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+
+    <dfdl:defineEscapeScheme name="CSVEscapeScheme">
```
mbeckerle: Escape scheme unused by these tests. Remove it.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/sepSuppression.tdml
```diff
@@ -0,0 +1,970 @@
...
+      <dfdl:escapeScheme 
+        escapeKind="escapeBlock" 
+        escapeBlockStart='"'
+        escapeBlockEnd='"' 
+        escapeCharacter='"' 
+        escapeEscapeCharacter='"' 
+        generateEscapeBlock="whenNeeded"
+        extraEscapedCharacters=", %#x0D; %#x0A;">
```
mbeckerle: Extra escaped not part of these tests. Remove

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
stevedlawrence: Seems like the mixin doesn't even really do much? Do you expect more stuff in here as this matures? Maybe it should just beome a boolean parameter that is checked in the right places?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -392,7 +431,74 @@ class OrderedSeparatedSequenceUnparser(
...
+          // however, we don't know whether to output the separator.
+          //
+          // The problem is that whether we put down this separator depends on whether
+          // the children of this group are in fact all absent and so there's nothing to output here, AND
+          // there's no subsequent siblings (i.e, this model group is truly trailing empty.)
+          //
+          // This really is a case where we can't stream.
+          //
+          // Simpler DFDL implementations could implement this behavior by some sort of backtracking on the output stream.
+          // They could use similar techniques for dealing with the storing of prefix-lengths. They don't implement
```
stevedlawrence: should the word "implement' be removed here?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
```diff
@@ -173,21 +173,33 @@ abstract class SequenceChild(
...
-    case true => new ScalarOrderedSeparatedSequenceChildUnparser(
-      childUnparser, srd, trd, sepUnparser, sq.separatorPosition, sq.separatorSuppressionPolicy,
-      this.separatorSuppressionMode)
-    case false => new ScalarOrderedUnseparatedSequenceChildUnparser(childUnparser, srd, trd)
+  lazy val sequenceChildUnparser: SequenceChildUnparser = {
+    val res =
+      (sq.hasSeparator, term.isPotentiallyTrailing) match {
+        case (true, false) => new ScalarOrderedSeparatedSequenceChildUnparser(
+        childUnparser, srd, trd, sepUnparser, sq.separatorPosition, sq.separatorSuppressionPolicy,
+          this.separatorSuppressionMode)
```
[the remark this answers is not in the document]
mbeckerle: I agree that we should embed a tool like that. 

We need a tool for XML indenting as well. Those files are full of tabs now, and CRLFs are starting to creep in here and there also.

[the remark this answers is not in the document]
jadams-tresys: For XML indenting we can use xmllint:
xmllint --c14n     -> reorder all properties/attributes alphabetically
xmllint --format   -> add new lines and indentation back in

For some reason running xmllint with both options seems to ignore the --format option, so it seems like it has to be a 2 step process.  I used this process to generate an appropriate diff for a file that by virtue of being touched by XMLSpy had every line changed.

mbeckerle:
Closing. This was never intended to be merged. It's just to gather comments.

PR #159, opened 2018-12-18
Restrict coercion of a whole expression

stevedlawrence:
While we allow coercion within an expression according to XPath, we do
not want to allow coercion to the expected expression type. For example,
if the expected type of an expression is an xs:string, we do not want to
coerce { 5 } to the string "5"--the expression should instead be { "5" }
or { xs:string(5) }. This limits coercion to only similar types.
Otherwise, the expected type of an expression must much the actual
expression result.

DAFFODIL-2021


stevedlawrence:
Need to add negative tests showing coercion errors.

stevedlawrence:
Came back to this to add some negative tests, and I dont' think this is really the complete solution. This change still allows coercion in the case where the target type and the expression result type are both numeric. This is clearly wrong in the case of decimal numbers, for example:

```xml
<xs:element name="a" type="xs:int" dfdl:inputValueCalc="{ 2.5 }" />
```
In this case, we really do not want to coerce 2.5 to an int, but this change would coerce since both are numerics.  An alternative might be to require that types match exactly, but then usability suffers. For example:

```xml
<xs:element name="a" type="xs:int" dfdl:inputValueCalc="{ 2 }" />
```

In this case, the literal 2 is technically an xs:integer so the result type and target type do not match. So if we were strict about it, this simple expression requires the use of ``xs:int(2)``. That's not terrible, but there's a good chance we would probably break a lot of schemas.

Perhaps a happy medium is to allow coercion between integer-like things? So we would automatically coerece an xs:integer to an xs:int (as long as the value was in range), but would not coerece 2.5 to 2, for example. An odd side effect of this is that one could know do this:

```xml
<xs:element name="a" type="xs:int" dfdl:inputValueCalc="{ xs:integer(2) }" />
```

And we would happily coerce that, even though it seems like maybe the user intended something else since types are explicitly provided and differ. Perhaps a better solution exists? I think it would be difficult track track when types are explicit and only check in those cases, so maybe this edge case isn't worth solving.

mbeckerle:
In what way is the expression 2 not an xs:int ? Our DPath compiler might choose to make all integers xs:integer and represent them as bignums at runtime, but that's perhaps just arbitrary. 

2 fits within the range of every signed and unsigned integer decimal and floating point type.  From the type-system perspective it is of the required type. It should be accepted for any integer type at least. 

The usual approach is that conversions that preserve information content perfectly are done automatically within numeric types, but those that involve rounding/truncation are not.

stevedlawrence:
According to w3c XPath, integer literals are alwyas xs:integer and then coerecion happens as needed. But this change is to disable coerecion of expression result. So technical 2 isn't an xs:int. Note that Daffodil does do some stuff to coerece literals down to int's' or long's (though I think technically this isn't spec compliant?), so 2 is probably considered an xs:int. But for example, we never coerce literals to xs:short's, so if you had something like this, it would fail without coerecion:

```xml
<xs:element name="a" type="xs:short" dfdl:inputValueCalc="{ 2 }" />
```

The suggestion of allowing coerence between types only when no rounding/truncation should occur seems reasonable. I'll try it out.

PR #161, opened 2019-01-04
Make the JAPI/SAPI DataProcessors serializable

stevedlawrence:
Some tools, like Apache Spark, need the ability to serialize the
DataProcessor to be sent to cluster nodes where it is then deserialized
and used to process data. We have the save/reload methods, but Spark
does not now how to use these. So this modifies the JAPI/SAPI
DataProcessors so that they are serializable, and modifies the JAPI/SAPI
tests to serialize/deserialize each DataProcessor before being used.

DAFFODIL-1440

### DIFF
--- a/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
+++ b/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
@@ -24,6 +24,8 @@ import static org.junit.Assert.fail;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.nio.channels.Channels;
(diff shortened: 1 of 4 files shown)


daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -59,8 +61,34 @@
...
 
+    public DataProcessor reserializeDataProcessor(DataProcessor dp) throws IOException, ClassNotFoundException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        ObjectOutputStream oos = new ObjectOutputStream(baos);
+        oos.writeObject(dp);
+        oos.close();
+
+        ByteArrayInputStream bais = new ByteArrayInputStream(baos.toByteArray());
+        ObjectInputStream ois = new ObjectInputStream(bais) {
+            /**
```
mbeckerle: This comment perhaps belongs on the method signature above.

Can you please explain more why we're having to do more here than just inherit from Serializable?

Is there a maintenance burden here, like all our serializable base classes have to manually call this reserialize thing?

stevedlawrence: Really, only the ``with Serialize`` stuff is all that is needed to get things to work.

But in addition to those changes, I made it so that all the JAPI and SAPI tests create a DataProcessor as normal but then also serialize and deserialize the DataProcessor, just to give some confidence that I made everything Serializable that needs to be. That's what this reserializeDataPrcoessor function does. It takes the DataProcessor that the test creates, seralizes it, deserializes it, and returns it. Then the tests use that one for the rest of the test. So this is purely used for testing within this file--no extra maintenence burden for us or users of Daffodil. Something like Spark wouldn't need this logic. They can just create a new ObjectInputStream like one would normally do.

Unfortunately, SBT closed that bug without fixing it. I think they just claim this is how SBT is expected to work so this is required. So I don't think it's ever going to be fixed.

mbeckerle: Can this explanation then go into the code. These reviews aren't a great place to capture things like this.

stevedlawrence:
Added comments explaining the purpose of the function. Also made the function prviate to help self-document that it should only be used for testing.

PR #162, opened 2019-01-08
Properly handle malformed binding files

jadams-tresys:
There were a few unhandled exceptions that were originating from being
unable to resolve QNames in external variable config files. With these
changes these exceptions should be handled more cleanly.

DAFFODIL-2044

I wanted to push this up for review before developing test cases to trigger the various places these QName exceptions can occur.  I have one test case currently that is broken due to the fact that the exception is picked up in TDMLRunner, so that test may just be removed.

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -49,7 +49,7 @@ import java.io.File
 import org.apache.daffodil.tdml.DFDLTestSuite
 import org.apache.daffodil.api.ValidationMode
 import scala.xml.Node
-import org.apache.daffodil.externalvars.Binding
+import org.apache.daffodil.externalvars.{ Binding, BindingException }
 import org.apache.daffodil.externalvars.ExternalVariablesLoader
 import org.apache.daffodil.configuration.ConfigurationLoader
 import org.apache.daffodil.api.ValidationMode
@@ -627,7 +627,9 @@ object Main extends Logging {
         val extVarBindingNodeOpt = (configNode \ "externalVariableBindings").headOption
         extVarBindingNodeOpt match {
           case None => Seq.empty
-          case Some(extVarBindingsNode) => ExternalVariablesLoader.getVariables(extVarBindingsNode, tunables)
+          case Some(extVarBindingsNode) => {
+            ExternalVariablesLoader.getVariables(extVarBindingsNode, tunables)
+          }
         }
       }
     }
@@ -1340,6 +1342,20 @@ object Main extends Logging {
     1
   }
 
+  def oomError(e: OutOfMemoryError): Int = {
+    System.err.println("""|
+                          |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                          |!!             Daffodil ran out of memory!              !!
+                          |!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                          |
+                          | Try increasing the amount of memory by changing or
+                          | setting the DAFFODIL_JAVA_OPTS environment variable.
+                          | "DAFFODIL_JAVA_OPTS=-Xmx5G" for 5GB.
+                          |
+                          |""".format(e.getMessage()).stripMargin)
+    1
+  }
+
   def main(arguments: Array[String]): Unit = {
     val ret = try {
       run(arguments)
@@ -1353,6 +1369,10 @@ object Main extends Logging {
         log(LogLevel.Error, "%s", e.getMessage())
         1
       }
+      case e: BindingException => {
+        log(LogLevel.Error, "%s", e.getMessage())
+        1
+      }
       case e: NotYetImplementedException => {
         nyiFound(e)
       }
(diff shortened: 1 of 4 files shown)


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -238,8 +238,11 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
   })
 
   def qnameConvert(s: String): RefQName = {
-    val eQN = QName.refQNameFromExtendedSyntax(s)
-    eQN.get
+    val eQN = QName.refQNameFromExtendedSyntax(s).recover {
+      case e: Throwable => throw e
```
mbeckerle: This seems like a no-op. If you don't have the recover, it should propagate this throwable exactly as this does. 

So if you want this, as a place to put a breakpoint, add a comment to that effect. Or take this out once done debugging.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -627,7 +630,22 @@ object Main extends Logging {
         val extVarBindingNodeOpt = (configNode \ "externalVariableBindings").headOption
         extVarBindingNodeOpt match {
           case None => Seq.empty
-          case Some(extVarBindingsNode) => ExternalVariablesLoader.getVariables(extVarBindingsNode, tunables)
+          case Some(extVarBindingsNode) => {
+            try {
+              ExternalVariablesLoader.getVariables(extVarBindingsNode, tunables)
+            } catch {
+              case qupe: QNameUndefinedPrefixException => {
```
mbeckerle: Shoudl be a base class or base trait here. This code doesn't really need to know what kinds of exceptions can come out of the variable bindings, only that one did. 

Your message in the second warning should be generic "Exception when processing external variable binding file".

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -472,7 +472,12 @@ final class SchemaSet(
       }
 
       val newNS = matchingDVs.head.namespace
-      val newBinding = Binding(v.varQName.local, Some(newNS), v.varValue)
+      val newBinding = try {
+        Binding(v.varQName.local, Some(newNS), v.varValue)
+      } catch {
+        case qupe: QNameUndefinedPrefixException => this.SDE("Undefined QName prefix in externally defined variable %s: %s", v.varQName, qupe.getMessage)
```
mbeckerle: Same here. You shouolnd't have to have two cases here. Just one for a base-class/base trait.

daffodil-runtime1/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesLoader.scala
```diff
@@ -83,7 +84,13 @@ object ExternalVariablesLoader {
   // return the mutated map.
 
   def loadVariables(vars: Map[String, String], referringContext: ThrowsSDE, vmap: VariableMap): VariableMap = {
-    val bindings = getVariables(vars)
+    val bindings = try {
+      getVariables(vars)
+    } catch {
+      case qupe: QNameUndefinedPrefixException => referringContext.SDE(qupe.getMessage)
```
mbeckerle: And here, and couple other places below too. Should just be one case.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -627,7 +630,22 @@ object Main extends Logging {
...
+                log(LogLevel.Warning, qupe.getMessage)
+                log(LogLevel.Warning, "Undefined QName prefix in external variable binding file. Continuing with default values")
+                Seq.empty
+              }
+              case qse: QNameSyntaxException => {
+                log(LogLevel.Warning, qse.getMessage)
+                log(LogLevel.Warning, "QName syntax error in external variable binding file. Continuing with default values")
+                Seq.empty
+              }
```
stevedlawrence: Thoughts on making these fatal errors? If a user provdies a config file with the intention of changing the behavior of how things will parse and something is wrong with the file, I'm not sure it makes sense to continue processing. If making these fatal makes sense, I think you can remove this change and let the exception bubble up, catch it in ``def main`` with other exceptions, and log an error.

daffodil-lib/src/main/scala/org/apache/daffodil/externalvars/Binding.scala
```diff
@@ -45,6 +45,9 @@ class Binding(val varQName: RefQName, val varValue: String, scope: scala.xml.Nam
       GlobalQName(varQName.prefix, varQName.local, varQName.namespace)
 }
 
+case class BindingException(message: String)
+  extends Exception("Exception when processing external variable binidng file: %s".format(message))
```
mbeckerle: Typo "binidng"

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -472,7 +472,11 @@ final class SchemaSet(
       }
 
       val newNS = matchingDVs.head.namespace
-      val newBinding = Binding(v.varQName.local, Some(newNS), v.varValue)
+      val newBinding = try {
+        Binding(v.varQName.local, Some(newNS), v.varValue)
+      } catch {
+        case e: BindingException => this.SDE("Exception when processing external variable %s: %s", v.varQName, e.getMessage)
```
stevedlawrence: I think the error message ends up a little verbose. The BindingException has a string like:
```Exception when processing external variable binidng file: %s```
So combine that with the SDE and the message would look like:
```Exception when processing external variable %s: Exception when processing external variable binidng file: %s"```
So the message would end up duplicated, I think.

daffodil-lib/src/main/scala/org/apache/daffodil/externalvars/Binding.scala
stevedlawrence: Also, a Binding doesn't necessarily need to come from a "file". Like it could from an API or the CLI. I imagine an error in one of those would still cause this Exception? Maybe just remove the work "file" if that's the case.

PR #164, opened 2019-01-11
Avoid CircularDefinition error for appinfo

jadams-tresys:
When we were detecting that an appinfo annotation did not have a source
attribute we would try to show a schema definition warning. However, the
lack of source attribute was somehow causing a CircularDefinition error
within the OOLAG/LV code when trying to print out the position of the
sequence where the error was occuring.

Instead of relying on the OOLAG/LV code to sort of reverse lookup the
sequences position, the position is now passed in as an argument to
ModelGroup.

DAFFODIL-2046

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
@@ -136,7 +136,7 @@ object TermFactory {
  * There are ultimately 4 concrete classes that implement this:
  * Sequence, Choice, SequenceGroupRef, and ChoiceGroupRef
  */
-abstract class ModelGroup
+abstract class ModelGroup(index: Int)
   extends Term
   with ModelGroupGrammarMixin
   with OverlapCheckMixin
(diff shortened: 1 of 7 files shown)


jadams-tresys:
Mike thought that there might be a similar case of avoiding OOLAG/LV calls for position in ElementBase, but I did not find anything obviously similar to what I changed in ModelGroup. If there is something in there, let me know and I'll add it to this pull request (or a new one).

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -153,14 +153,7 @@ abstract class ModelGroup
...
 
-  private def prettyIndex = LV('prettyIndex) {
-    myPeers.map { peers =>
-      {
-        if (peers.length == 1) "" // no index expression if we are the only one
-        else "[" + (peers.indexOf(this) + 1) + "]" // 1-based indexing in XML/XSD
-      }
-    }.getOrElse("")
-  }.value
+  private def prettyIndex = "[" + index + "]" // 1-based indexing in XML/XSD
```
stevedlawrence: Looks like this change caused some CLI tests to fail. I'd guess because we now always output "[X]" even if there are no peers. I think this change is fine since this is more a debug thing, so always outputting the index doesn't matter, but we should update the CLI tests so they all pass. To runt he CLI tests, you can run  ``sbt it:test``.

jadams-tresys: Fixed this issue by having it look for seq[1] instead of just seq.

PR #167, opened 2019-01-16
Daffodil 2051 tests broken in windows

ltapilot:
Remaining tests now fixed. Should be platform and OS independent.


daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AG.tdml
```diff
@@ -24,9 +24,9 @@
   <parserTestCase name="AG000" root="allZones" model="AG.dfdl.xsd"
     description="Discriminated choices"
     roundTrip="true">
-    <document><![CDATA[1
-zone 1
-]]></document>
+    <document>
+	  <documentPart type="text" replaceDFDLEntities="true"><![CDATA[1%LF;zone 1%LF;]]></documentPart>
```
mbeckerle: I wonder how many tests would break if we changed the default for replaceDFDLEntitites so it defaults to "true". That would make these less cluttered, by allowing just <document><![CDATA[...]]></document> again.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
```diff
@@ -167,8 +167,9 @@
   <tdml:parserTestCase name="ibm_format_compat_02"
     root="ABC" model="ABC_IBM.dfdl.xsd" description=""
     roundTrip="twoPass">
-    <tdml:document><![CDATA[a
-]]></tdml:document>
+    <tdml:document>
+	  <tdml:documentPart type="text" replaceDFDLEntities="true"><![CDATA[a%LF;]]></tdml:documentPart>
```
mbeckerle: I'm not going to suggest changing these, but in these short-line cases, This would also have worked since the character being inserted is a legal XML character:
<document>a&#x0A;</document>
This has little chance of being line-wrapped. The drawback, is if the character being inserted is not a legal char like LF, but is something like NUL, then this won't work, and the longer syntax as used in the code, with the DFDL entities being used, that would be the only way to express it.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/AG.tdml
stevedlawrence: I imagine anything with a percent sign would break since that would look like an invalid DFDL entitity? A quick grep shows about 65 lines that contain "document" and a percent and don't already have replaceDFDLEntities set. So I'd guess at least that many would need replaceDFDLEntities="false", which is probably about the same amount that needed to be changed in this commit so I'm not sure it would be all the different.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
stevedlawrence: Did something in your comment get stripped out by github? I'm guessing you mean to say  ``a&#xA;`` and github replaced the xA it with an actual linefeed? Using backticks seems to escape it and make it show up.

mbeckerle: I edited to fix the above so it is readable. Thanks for the catch.

mbeckerle:
The travis CI build failed due to rat check failure - a file was introduced that needs an apache license banner I believe.

stevedlawrence:
That rat failure was introduced in commit 912f7d2b73 and we missed it. It was then fixed as part of the next commit 69bb0a185e, but this PR branched off of the earlier commit. Once the comits are squashed rebased on top of current master, we can confirm that this did not introduce any new rat failures.

stevedlawrence:
Also, apparently TravisCI now supports windows:

https://blog.travis-ci.com/2018-10-11-windows-early-release

It might be a good idea to modify our .travis.yml file to enable that. There might some issues like the rpm build would probably fail on windows, but I think we can disable certain commands based on the build environment:

https://docs.travis-ci.com/user/multi-os/#example-multi-os-build-matrix

stevedlawrence:
Actually, it looking more closing, TravisCI does not support java/scala yet, so we'll have to hold off on that.

stevedlawrence:
Looks like the ratCheck is still failing. Can you please look into that? I think you maybe just need to fetch the latest and rebase ontop of master.

PR #168, opened 2019-01-18
Keep line/column info for Daffodil tests

jadams-tresys:
We had been adding line and column attributes to the schemas we were
testing against in order to provide more helpful error messages. These
attributes were causing issues with the IBM DFDL processor and were
removed.

This commit allows the TDMLRunner to detect if it is running the IBM
DFDL implementation and remove these attributes as needed, while keeping
them for the Daffodil implementation.

DAFFODIL-2050

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -24,9 +24,11 @@ import java.io.StringReader
 import java.net.URI
 
 import scala.xml.Node
+import scala.xml.Elem
 import scala.xml.NodeSeq
 import scala.xml.NodeSeq.seqToNodeSeq
 import scala.xml.SAXParseException
+import scala.xml.transform._
 import org.apache.daffodil.api.DataLocation
 import org.apache.daffodil.api.ValidationMode
 import org.apache.daffodil.api.DaffodilSchemaSource
@@ -456,6 +458,20 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
 
   final def isCrossTest(implString: String) = implString != "daffodil"
 
+  val removeLineColInfo = new RewriteRule {
+    override def transform(n: Node) = n match {
+      case e: Elem => {
+        e.copy(attributes = n.attributes.filter(a =>
+        (a.prefixedKey != "dafint:" + XMLUtils.COLUMN_ATTRIBUTE_NAME) &&
+        (a.prefixedKey != "dafint:" + XMLUtils.LINE_ATTRIBUTE_NAME) &&
+        (a.prefixedKey != "dafint:" + XMLUtils.FILE_ATTRIBUTE_NAME)))
+      }
+      case _ => n
+    }
+  }
+
+  val stripLineColInfo = new RuleTransformer(removeLineColInfo)
+
   final def isNegativeTest = optExpectedErrors.isDefined
 
   lazy val tdmlDFDLProcessorFactory: AbstractTDMLDFDLProcessorFactory = {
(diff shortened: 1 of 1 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
```diff
@@ -72,7 +72,7 @@ object Position {
  * gets positions is different. It is given just an offset into the document file/stream,
  * and it therefore must synthesize line number/col number info itself.
  */
-class DaffodilConstructingLoader(uri: URI, errorHandler: org.xml.sax.ErrorHandler, addLineColInfo: Boolean = false)
+class DaffodilConstructingLoader(uri: URI, errorHandler: org.xml.sax.ErrorHandler, addLineColInfo: Boolean = true)
```
stevedlawrence: Is it worth removeing the addLineColInfo paramter? We never use it anywhere.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -456,6 +458,20 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
   final def isCrossTest(implString: String) = implString != "daffodil"
 
+  val removeLineColInfo = new RewriteRule {
+    override def transform(n: Node) = n match {
+      case e: Elem => {
+        e.copy(attributes = n.attributes.filter(a =>
+        (a.key != XMLUtils.COLUMN_ATTRIBUTE_NAME) &&
+        (a.key != XMLUtils.LINE_ATTRIBUTE_NAME) &&
+        (a.key != XMLUtils.FILE_ATTRIBUTE_NAME)))
+      }
```
stevedlawrence: Should we also make sure the the namespace is the DAFINT namespace?  "column", "line", and "file" are pretty generic, I could see this easily removing attributes in some other namespace with the same name that shouldn't be removed.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -603,7 +619,19 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
+         * attributes that it doesn't understand, so we must remove them */
+        val processedSchema = {
+          if (isCrossTest(this.tdmlDFDLProcessorFactory.implementationName)) {
+            stripLineColInfo(defSchema.xsdSchema)
+          } else {
+            defSchema.xsdSchema
+          }
+        }
+
+        EmbeddedSchemaSource(processedSchema, defSchema.name)
```
stevedlawrence: Can you add a test to make sure that when there is an SDE in a embeded and non-embedded schema that we get a line/col number when the implementation is "daffodil"?

stevedlawrence: Also, can you double check with the [OpenDFDL/ibmDFDLCrossTester](https://github.com/OpenDFDL/ibmDFDLCrossTester)? One tests fails with ``sbt test`` with this PR with an error about unsupported attribute, so I think something still isn't being removed corrrectly.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
jadams-tresys: Will remove

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
jadams-tresys: Will fix, they have a prefixedKey method that I didn't see the first time around

PR #169, opened 2019-01-22
Fix incorrect line numbers and ensure XML errors include line numbers

stevedlawrence:
- A bug in scala-xml caused line numbers to be reset to zero after
  parsing CDATA and processing instructions. This can be worked around
  by calling lookahead() when a DaffodilConstructingLoader is
  initialized
- When using the DaffodilConstructingLoader to parse XML, if any
  exceptions are thrown, convert them to SAXParseExceptions and return
  null. This makes our errors consistent and also causes line numbers to
  be included in errors that previously did not (e.g. missing closing
  tag)
- Change the DaffodilConstructingLoader to override the elem(...)
  function to add file/line/column attributes there instead of
  overriding xTag and mkAttributes. Those functions are probably less
  safe to override since they are more internal to the scala-xml
  ConstructingLoader, and they also required position state to be
  carried around outside. Using the elem function is a bit cleaner and
  probably safer
- Remove the addColLineInfo flag. We need to handle removing line/col
  information for IBM compatibility differently as part of another
  commit. It's too difficult to use this flag in the way that's
  intended.

DAFFODIL-1476

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
@@ -72,7 +72,7 @@ object Position {
  * gets positions is different. It is given just an offset into the document file/stream,
  * and it therefore must synthesize line number/col number info itself.
  */
-class DaffodilConstructingLoader(uri: URI, errorHandler: org.xml.sax.ErrorHandler, addLineColInfo: Boolean = false)
+class DaffodilConstructingLoader(uri: URI, errorHandler: org.xml.sax.ErrorHandler)
   extends ConstructingParser({
     //
     // Note: we must open the XML carefully since it might be in some non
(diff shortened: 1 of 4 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
```diff
@@ -86,46 +86,60 @@ class DaffodilConstructingLoader(uri: URI, errorHandler: org.xml.sax.ErrorHandle
...
-   * Note that this is potentially fragile and could break if scala-xml ever
-   * makes significant changes. However, these functions, and the classes
-   * containing these functions have not been modified for many years, so I
-   * suspect they are fairly stable and it's safe to assume things won't change
-   * significantly.
+   *  @param pos      the position in the source file
+   *  @param pre      the prefix
+   *  @param local    the local name
+   *  @param attrs    the attributes (metadata)
+   *  @param scaope   the namespace binding scope
```
jadams-tresys: Spelling error on scope

PR #170, opened 2019-01-24
Fixed CLI integration tests so that they run properly on Windows.

ltapilot:
DAFFODIL-2058

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -139,6 +139,7 @@ lazy val commonSettings = Seq(
   sourceManaged := baseDirectory.value / "src_managed",
   resourceManaged := baseDirectory.value / "resource_managed",
   libraryDependencies ++= Dependencies.common,
+  parallelExecution in IntegrationTest := false
 ) ++ Defaults.itSettings ++ debugTestSettings ++ debugIntegrationTestSettings
 
 lazy val debugTestSettings = inConfig(TestDebug)(Defaults.testSettings ++ Seq(
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
@@ -34,13 +34,13 @@ class TestCLITunables2 {
 
     try {
       // note: 2>&1 is shell-speak for "connect stderr into stdout"
-      val cmd = String.format("""echo "a,b" | %s parse -s %s -TsuppressSchemaDefinitionWarnings=all 2>&1""", Util.binPath, testSchemaFile)
+      val cmd = String.format("""echo a,b| %s parse -s %s -TsuppressSchemaDefinitionWarnings=all 2>&1""", Util.binPath, testSchemaFile)
       shell.sendLine(cmd)
       shell.expect(contains("""<ex:e1 xmlns:ex="http://example.com">
   <s1>a</s1>
   <s2>b
 </s2>
-</ex:e1>"""))
+</ex:e1>""".replace("\r\n", "\n")))
       shell.sendLine("exit")
       shell.expect(eof)
     } finally {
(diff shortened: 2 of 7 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -34,13 +34,13 @@ class TestCLITunables2 {
 
     try {
       // note: 2>&1 is shell-speak for "connect stderr into stdout"
-      val cmd = String.format("""echo "a,b" | %s parse -s %s -TsuppressSchemaDefinitionWarnings=all 2>&1""", Util.binPath, testSchemaFile)
+      val cmd = String.format("""echo a,b| %s parse -s %s -TsuppressSchemaDefinitionWarnings=all 2>&1""", Util.binPath, testSchemaFile)
```
mbeckerle: Shouldn't this also be Util.echo(...)

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -58,7 +58,7 @@ class TestCLITunables2 {
 
     try {
       // note: 2>&1 is shell-speak for "connect stderr into stdout"
-      val cmd = String.format("""echo "a,b" | %s parse -s %s 2>&1""", Util.binPath, testSchemaFile)
+      val cmd = String.format("""echo a,b| %s parse -s %s 2>&1""", Util.binPath, testSchemaFile)
```
mbeckerle: and here

build.sbt
```diff
@@ -139,6 +139,7 @@ lazy val commonSettings = Seq(
   sourceManaged := baseDirectory.value / "src_managed",
   resourceManaged := baseDirectory.value / "resource_managed",
   libraryDependencies ++= Dependencies.common,
+  parallelExecution in IntegrationTest := false
```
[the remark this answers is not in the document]
stevedlawrence: Agreed, I'm not even sure this change is needed at all. I know Mark was running into memory issues when runing on windows and this was one change he made early on. But I'm not sure it really helped. I think the actual fix was changes to DAFFODIL_JAVA_OPTS in TestCLIDebugger.

Mark, can you remove this change and see if the memory issues come back? If not, I think we should remove this change completely.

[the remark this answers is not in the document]
ltapilot: I spent a couple of hours messing with this today. It turns out that the platform doesn't do a good enough job of predicting memory usage. Subsequently it starts too many tests in parallel and when the tests all ask for large blocks of RAM serious memory pressure builds at the system level. In my case the system even ran out of swap space. My system has 16GB of RAM and it is clear that this should still be able to run on systems with less than that. For these reasons I'm going to leave parallel execution OFF in this pull request. I wish it were not so, but I think that's the only way we're going to get this to run consistently and without failure. Sorry.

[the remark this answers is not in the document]
stevedlawrence: Seems reasonble in that case. The change only disables parallel executiong for integration tests (i.e. CLI tests), so it shouldn't affect the time it takes to run all our normal tests. As Josh mentions, we could disable/enable parallel execution based on the operating system, but I think I'd rather evertyhing be as similar as possible on different OS. So if it needs to be disabled on Windows, I'd vote for also disabling on Linux too.

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
ltapilot: For the record, I concur with Steve. The 'Util.echoN' function is only intended to cover the difference in 'echo' behavior 
 between Windows and Linux when the stated conditions are met.

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
ltapilot: As stated, above.

PR #173, opened 2019-01-28
Allow implicit conversion of string to HexBinary

jadams-tresys:
DAFFODIL-2056

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
@@ -78,3 +78,16 @@ case object StringToTime extends Converter {
     result
   }
 }
+
+case object StringToHexBinary extends Converter with HexBinaryKind{
+  val name = "StringToHexBinary"
+
+  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
+    val result = a match {
+      case s: String => hexStringToByteArray(s)
+      case hb: Array[Byte] => hb
+      case x => throw new NumberFormatException("%s cannot be cast to dfdl:hexBinary\ndfdl:hexBinary received an unrecognized type! Must be String or HexBinary.".format(x.toString))
+    }
+    result
+  }
+}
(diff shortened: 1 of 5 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
```diff
@@ -78,3 +78,15 @@ case object StringToTime extends Converter {
...
+  val name = "StringToHexBinary"
+
+  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
+    val result = a match {
+      case s: String => hexStringToByteArray(s)
+      case x => throw new NumberFormatException("%s cannot be cast to dfdl:hexBinary\ndfdl:hexBinary received an unrecognized type! Must be String or HexBinary.".format(x.toString))
+    }
+    result
+  }
+}
```
stevedlawrence: Does it make sense to modify XSHexBinary to call this function now so as to minimize duplicate code? So XSHexBInary basically just calls StringToHexBinary.computeValue(...), and all the actual conversion logic is only in this object?  This looks like how we do other conversions like XSTime vs StringToTime

jadams-tresys: I wasn't sure that it would, since the XSHexBinary computeValue also accepts reading in a Byte array, so I thought it might be confusing if the StringToHexBinary converter also allows reading in both a String or a Byte array.

stevedlawrence: That is a little confusing, but is at least consistent with the StringTo{Date,Time,DateTime} objects. I'd vote for consistency and less duplication. Maybe create a bug to try to figure out why the StringTo* objects will also convert things that aren't Strings? Maybe that's correct behavior and they just need to be renamed to ConverToDateTime, for example.

jadams-tresys: I've made the change and it should be pushed up now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSHexBinary.scala
```diff
@@ -90,14 +90,6 @@ case class XSHexBinary(recipe: CompiledDPath, argType: NodeInfo.Kind)
   val name = "XSHexBinary"
```
stevedlawrence: XSHexBinary mixes in HexBinaryKind, which I think isn't needed anymore? Can that be removed?

+1

PR #176, opened 2019-01-29
Add better error message for '/' vs 'div'

jadams-tresys:
Suggests using 'div' instead of '/' when this error is detected by the
expression parser.

DAFFODIL-2042

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
@@ -157,7 +157,14 @@ class DFDLPathExpressionParser[T <: AnyRef](qn: NamedQName,
           nextRdr = nextRdr.rest
           i += 1
         }
-        context.SDE("Unable to parse expression. Message: %s\nNext: %s.", msg, nextString.toString())
+        val err_msg = {
+          if (msg.contains('/')) {
+            msg + ", did you mean to use 'div'?"
+          } else {
+            msg
+          }
+        }
+        context.SDE("Unable to parse expression. Message: %s\nNext: %s.", err_msg, nextString.toString())
       }
     }
   }
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
@@ -242,6 +242,7 @@ class DFDLPathExpressionParser[T <: AnyRef](qn: NamedQName,
     })("add")
 
   def MultiplicativeExpr: Parser[Expression] = log(
+    UnaryExpr ~ (("/") ~ UnaryExpr) ^^ { _ => context.SDE("'/' is not a valid operator in DFDL Expression Syntax. Did you mean to use 'div'?") } |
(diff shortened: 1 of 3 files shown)


jadams-tresys:
While it is nice to have a more helpful error message, I imagine that this is just one of several edge cases, not to mention that the error message is coming from the Scala RegexParsers (if I understand the code correctly), so I'm not sure it is worth making this change.  I figured I could push this up for review and see what the group decides.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
```diff
@@ -157,7 +157,14 @@ class DFDLPathExpressionParser[T <: AnyRef](qn: NamedQName,
           nextRdr = nextRdr.rest
           i += 1
         }
-        context.SDE("Unable to parse expression. Message: %s\nNext: %s.", msg, nextString.toString())
+        val err_msg = {
+          if (msg.contains('/')) {
+            msg + ", did you mean to use 'div'?"
+          } else {
```
stevedlawrence: This feels pretty fragile to me, and could potentially give a lot of false negatives if there's some error message that uses a /, like maybe a path expression. Perhaps an alternative approach would be to modify the MultiplicativeExpr grammar to accept a /, but then call context.SDE() if it matches? We do similar things for `//`` in a path expression. Not sure if there might be some ambiguities and adding that might make normal paths (i.e. ``/foo/bar/baz``) look like errors in some cases, though.

jadams-tresys: While I agree that that is a better solution, it seems to break all the path tests where as my original method doesn't. I'm not entirely sure I understand the code for the UnaryExpr stuff correctly, so it's possible I did something wrong. I'll push up my changes as an example.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
```diff
@@ -235,6 +242,7 @@ class DFDLPathExpressionParser[T <: AnyRef](qn: NamedQName,
     })("add")
 
   def MultiplicativeExpr: Parser[Expression] = log(
+    UnaryExpr ~ (("/") ~ UnaryExpr) ^^ { _ => context.SDE("'/' is not a valid operator in DFDL Expression Syntax. Did you mean to use 'div'?") } |
```
jadams-tresys: Like Steve suggested, this seems like the more correct way to handle this, but it causes all the tests involving paths to break.

stevedlawrence: Yeah, I was afraid that would happen. I'm not sure this is really worth fixing. The ambiguity with path expressions makes this difficult to resolve. I believe we have a general bug to fix diagnostics coming out of DFDL expression compilation. I suspect whatever is done to make that better will be the right solution to also make this specific case better.

jadams-tresys: That's kind of what I was leaning towards too, but figured I would see if there was a reasonably simple way to add this kind of helpful error message.  I'll see if I can find the bug for better diagnostics and I'll mark the ticket for / vs div as a duplicate or something.

PR #177, opened 2019-02-06
Added note about why 2 tests fail on IBM

jadams-tresys:
DAFFODIL-2017


jadams-tresys:
Took me a bit to get the IBMCrossTester up and running and once I did I realized that the tests didn't need to be updated in order to work.  However, 2 of the tests that were listed in the DAFFODIL-2017, test_length_delimited_12_01 and 02, still fail due to IBM's implementation not allowing the separator and terminator to be the same character(s).

Added this comment to explain why these fail and mentioned bug 2018 which discusses this incompatibility.

PR #179, opened 2019-02-07
Return xs:decimal when dividing xs:long

jadams-tresys:
Originally dividing xs:long was returning a double, but it should have
been returning a decimal instead.

DAFFODIL-2063

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
@@ -86,9 +86,9 @@ object NodeInfoUtils {
       case Decimal => Decimal
       case Integer => Decimal
       case Double => Double
-      case Long => Double
+      case Long => Decimal
       case Float => Float
-      case Int => Float
+      case Int => Decimal
       case ArrayIndex => ArrayIndex
       case _ => Assert.usageError("Unsupported return type: %s".format(resultType))
     }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NumericOps.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NumericOps.scala
@@ -99,13 +99,7 @@ case object TimesInteger extends NumericOp {
   def operate(v1: JNumber, v2: JNumber): JBigInt = { asBigInt(v1).multiply(asBigInt(v2)) }
 }
 case object DivInteger extends NumericOp {
-  def operate(v1: JNumber, v2: JNumber): JBigDecimal = {
-    val v2bd = asBigDecimal(v2)
-    if (v2bd.compareTo(JBigDecimal.ZERO) == 0) {
-      throw new ArithmeticException("/ by zero")
-    }
-    asBigDecimal(v1).divide(v2bd)
-  }
+  def operate(v1: JNumber, v2: JNumber): JBigDecimal = { DivDecimal.operate(v1, v2) }
 }
 case object IDivInteger extends NumericOp {
   def operate(v1: JNumber, v2: JNumber): JBigInt = { asBigInt(DivInteger.operate(v1, v2)) }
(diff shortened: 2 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
```diff
@@ -86,7 +86,7 @@ object NodeInfoUtils {
       case Decimal => Decimal
       case Integer => Decimal
       case Double => Double
-      case Long => Double
+      case Long => Decimal
       case Float => Float
       case Int => Float
       case ArrayIndex => ArrayIndex
```
stevedlawrence: Should Int also convert to Decimal isntead of Float? So only float and double divisions return float/double, eveything else will return Decimal?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NumericOps.scala
```diff
@@ -174,13 +174,7 @@ case object TimesLong extends NumericOp {
...
 }
 case object DivLong extends NumericOp {
-  def operate(v1: JNumber, v2: JNumber): JNumber = {
-    val v2l = asLong(v2)
-    if (v2l == 0) {
-      throw new ArithmeticException("/ by zero")
-    }
-    asDouble(v1) / v2l
-  }
+  def operate(v1: JNumber, v2: JNumber): JBigDecimal = { DivDecimal.operate(v1, v2) }
```
stevedlawrence: Is this code path even possible now? I would guess that with the change above the convergedArgType in NumericExpression must be one of Decimal, Float, or Double so the case like ``("div", Long)`` and other related int types wouldn't even be possible. So DivLong and IDivLong would never be used. I would htink only DivDecimal, DivDouble, and DivFloat need to exist?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -7184,6 +7185,11 @@
     <tdml:infoset><tdml:dfdlInfoset><div21>7.290005321703885E-5</div21></tdml:dfdlInfoset></tdml:infoset>
   </tdml:parserTestCase>
 
+  <tdml:parserTestCase name="div22" root="div22" model="XPathDiv" description="Section 23 - DFDL Expressions - div"
+    config="cfg_errorOnResultCoercion">
+    <tdml:document></tdml:document>
+    <tdml:infoset><tdml:dfdlInfoset><div22>0.028125</div22></tdml:dfdlInfoset></tdml:infoset>
```
stevedlawrence: I'm not sure if this test fully tests that long div long returns a decimal. If long div long return a double, we will allow automatic upcasting to the expected xs:decimal type, so this doesn't really show that we're doing the right thing, I think. This test will probably work without the above change.

Maybe the best way to handle this is to add a unit test somewhere (maybe TestDFDLExpressionEvaluation.scala) that evaluates an expression and checks the return type is the correct thing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NumericOps.scala
jadams-tresys: I threw in a quick print statement the code path does get hit still, so I think this function and the others are still necessary.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
jadams-tresys: I believe you are correct, according to the xpath spec, 'div' on integers should also return a decimal

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NumericOps.scala
stevedlawrence: Yeah, after taking a look I see now that the convergedArgType is based on the arguemtns and not the expected result. So Div{Long,Int,etc} are divisions based on the arguments.

However, I think that means Div{Integer-like things} needs to be updated, similar to the change you've made for DivLong. So DivLong, DIvShort, DIvUnsignedBYte, etc. all have the same issue.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
jadams-tresys: I'll make the change to the unsigned variants as well.

jadams-tresys: Looking at this divResult function, it is lacking many of the smaller types (byte, short, etc) as well as all the unsigned types.  Are these types actually not supported or is this an oversight?

stevedlawrence: The parameter passed into the divResult method comes from the second value of tuple in the match case below this. That tuple value is always one of Decimal, Double, Float, Integer, ArrayIndex, Long, or Int. So I think this match/case is exhastive. Whehter or not that's correct, I'm not 100% sure.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
stevedlawrence: New updates look good, but this could use some tests to make sure we're doing the right thing. This is a corner case that really needs tests to ensure we're doing the right thing.

jadams-tresys: Just pushed up the tests. Meant to do that last week, but forgot about adding tests then. I was struggling with getting the unit tests working to verify the types, so I did what you originally suggested and had tests expecting incorrect types and then made sure the appropriate error message was displayed.

PR #182, opened 2019-02-13
Fixed instance of NadaParser not being optimized out

jadams-tresys:
SpecifiedLengthExplicitParser was not checking for isEmpty, which was
resulting in prodcuing a NadaParser.

DAFFODIL-2062


jadams-tresys:
Took a little while to track this down, but ended up being a simple fix once found.  I wouldn't be surprised if there are other issues similar to this where we don't have checks in place for Parsers that will generate NadaParsers.

PR #183, opened 2019-02-14
Fix type coerceion between signed/unsigned types

jadams-tresys:
Originally if a larger unsigned type was added to a smaller signed type,
the result would be casted to the unsigned type instead of a larger
signed type. Now we cast to a larger signed type if necessary.

DAFFODIL-2065


daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
```diff
@@ -113,24 +113,34 @@ object NodeInfoUtils {
       case (Float, _) => (Float, Float)
       case (_, Integer) => (Integer, Integer)
       case (Integer, _) => (Integer, Integer)
+      case (lat, NonNegativeInteger) if (lat.isInstanceOf[SignedInteger.Kind]) => (Integer, Integer)
```
stevedlawrence: Is there a way to remove the asInstanceOf stuff? I think you might need to use parenthesis to match the instance instead of the companion object?
```scala
case (SignedInteger(), NonNegativeInteger) => ...
```
Or maybe you needs to be bound to a variable?
```scala
case (_: SignedInteger, NonNegativeInteger) => ...
```
I need to reference myself of scala pattern matching.

jadams-tresys: The second case works and cleans up the code a lot.  I think I was overthinking things with the if statements.
I've pushed up the changes.

PR #184, opened 2019-02-15
Added sbt-scoverage code coverage reporting

jadams-tresys:
DAFFODIL-2003

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,17 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Prevents codecov from creating a test coverage comment in a pull request
+comment: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,14 @@ before_script:
   - unset _JAVA_OPTIONS
 script:
   - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION compile test:compile it:compile debug:compile || travis_terminate 1
-  - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION test
-  - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION it:test
+  - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION coverage test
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -74,6 +74,16 @@ This following command generates an [Apache RAT](https://creadur.apache.org/rat/
 $ sbt ratCheck
 ```
 
+### sbt-scoverage
```
stevedlawrence: Does it make sense to change the heading to something like "Test Coverage Report" or something? If someone doesn't know what "sbt-socverage" is, this heading isn't very useful. The content of the section goes on to explain and link to sbt-scoverage, so that's good for people that want to know more details.

README.md
```diff
@@ -74,6 +74,16 @@ This following command generates an [Apache RAT](https://creadur.apache.org/rat/
...
+
+This follow commands generate an [sbt-scoverage](https://github.com/scoverage/sbt-scoverage) test coverage report that will be located in ``target/scala-ver/scoverage-report/``.
+
+```bash
+$ sbt clean coverage test it:test
+$ sbt coverageReport
+$ sbt coverageAggregate
+```
+
```
stevedlawrence: This will be really nice! Super easy to generate reports.

Can you also modify the .travis.yml file to add the coverage commands and upload the reports to codecov.io? According to this example, it only takes a couple extra lines:

https://github.com/codecov/example-scala/blob/master/.travis.yml

jadams-tresys: That was the plan, I was just going to do it under a separate pull request, because I wasn't sure how straightforward it would be.

README.md
jadams-tresys: Will change, just kept it that way since I saw the previous section was Apache RAT.

stevedlawrence: Fair, Apache RAT probably isn't a very heading name, I'll take the blame for that.

README.md
stevedlawrence: That works for me.

stevedlawrence:
+1

README.md
jadams-tresys: I ended up just adding it to the pull request once I saw it was just a couple lines

.travis.yml
```diff
@@ -39,3 +39,7 @@ script:
   - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION daffodil-japi/genjavadoc:doc
   - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION daffodil-sapi/doc
   - sbt -J-Xms1g -J-Xmx5g ++$TRAVIS_SCALA_VERSION ratCheck
+  - sbt clean coverage test it:test
```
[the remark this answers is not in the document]
jadams-tresys: I'll add the -J options and make it so we only run the tests once.

As for needing coverageAggregate, I'm not sure since I haven't used codecov before.  I know the coverageAggregate is what made the nice index.html file with links to all the subproject coverage results.  It doesn't take long to run (just couple of seconds on my system) so my vote would be to leave it in for now and see how things look on codecov.

stevedlawrence:
Cool, guess it's working!

That said, I'm not sure we want codecov-io to add a comment to every pull request. Looks like that can be disabled with a [codecov.yml file](https://docs.codecov.io/docs/pull-request-comments)?

jadams-tresys:
Just pushed up another small change, adding a .codecov.yml file, which should disable the comment creation

.codecov.yml
```diff
@@ -0,0 +1 @@
+comment: off
```
stevedlawrence: I think this will fail the rat check. Needs a header like in .travis.yml. Otherwise looks good.
+1

PR #186, opened 2019-02-20
Added tests that illustrate DAFFODIL-2069 bug.

mbeckerle:
DAFFODIL-2069 is a complex interaction of outputValueCalc with hexBinary with bitOrder LSBF.

This PR just adds tests that illustrate the bug, commented out so all tests still pass, but illustrates the issue for whomever investigates, and lets us consider the impact of the bug. 

This bug was discovered in unit tests of the mil-std-2045 schema. One of those tests no longer passes on 2.3.0-rc1. 

See https://issues.apache.org/jira/browse/DAFFODIL-2069 for description.

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc3.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc3.tdml
@@ -1,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc3.tdml
```diff
@@ -0,0 +1,127 @@
...
+  without specific prior written permission.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH THE
+SOFTWARE.
+-->
```
stevedlawrence: Not apache license, causing rat check to fail.

mbeckerle:
Re: 2.3.0 release. I think we need to check that the binary format schemas like stanag 5516, actually still work. VMF works because none of the messages we're testing use the 2045 header fields that are hexBinary with prefix length computed by a value calc. 

These might be able to be changed to use new prefixed lengthKind support, and that may be a good workaround. I'll give that a try.

mbeckerle:
I confirmed all tests with nato-stanag-5516 work (or can be made to) on 2.3.0 and updated that schema's tests to work on 2.3.0. 

I also confirmed that prefixed lengthKind can be used in the mil-std-2045 header as a workaround for the problematic hexBinary fields, so will make that change.

I believe this bug does not need to stop the release of 2.3.0 daffodil.

PR #191, opened 2019-03-04
Ensure locale does not effect text numbers and date/times

stevedlawrence:
There were two main issues where locale could have an effect:

- We previously only set grouping separator or decimal separator if the
  text number pattern contained the grouping or decimal character.
  However, even if the pattern does not contain a grouping/decimal
  character, the value from the locale can still have an effect if the
  other does exist. So, if either exists in the pattern, we need to set
  them both to ensure that the defaults from the locale do not affect
  the result.
- When we parse an infoset string, that creates a new Calendar with some
  internal values specific to the locale. When we unparse that Calendar
  those locale specific values can have an effect on the result. To
  avoid this, copy the date/time values from the new Calendar into the
  Calendar that was created with DFDL properties, ensuring the locale
  does not effect the result.

The remaining issues were related to unit tests that didn't properly set
all the decimal format symbols, causing locale had an affect.

DAFFODIL-2074


daffodil-runtime1/src/test/scala/org/apache/daffodil/processors/input/TestICU.scala
```diff
@@ -149,4 +151,32 @@ class TestICU {
     //assertEquals(1234L, num)
     //assertEquals(7, pp.getIndex)
   }
+
+  // Shows that even if a decimal format pattern doesn't contain a decimal, the
```
mbeckerle: Should say "doesn't contain a decimal point, ..." (the point is, it's is missing 'point')

daffodil-runtime1/src/test/scala/org/apache/daffodil/processors/input/TestICU.scala
```diff
@@ -149,4 +151,32 @@ class TestICU {
...
+    val dfs = new DecimalFormatSymbols(java.util.Locale.GERMANY)
+
+    // Set the grouping separator to be the same as the decimal separator from
+    // the locale. ICU never complains
+    dfs.setGroupingSeparator(',')
+
+    // Define a pattern that only has a grouping separator--no decimal separator
+    val df = new DecimalFormat("###,###", dfs)
+
+    // We don't have a decimal character in the pattern, so one might expect
```
mbeckerle: "decimal point character" ?  In the pattern this is always a ".", so calling it the decimal point seems ok.

stevedlawrence:
Above comments resolved in fixup commit.

PR #192, opened 2019-03-04
Fix issue where suspensions with mixed byteOrder caused incorrect bitOrder to be used

stevedlawrence:
DAFFODIL-2069


daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -757,32 +757,46 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
...
+    // most of the time.
     //
+    // However, a buffered DOS maintains a fragment byte, and at some point we
+    // collapse that fragment byte into a direct DOS using this function. But
+    // we just said that there not really a byteOrder associated with a DOS,
+    // only with elements. So if we do not know the byteOrder of the fragment
+    // byte of a DOS, how do we know which function below to call when
+    // collapsing a fragment byte?
+    //
+    // Well, we do keep track of the bitOrder, since we need to know if we
```
mbeckerle: Excellent comment.

PR #193, opened 2019-03-05
Ensure content/valueLength expressions are not considered constants

stevedlawrence:
Although it should result in an error, calling contentLength() or
valueLength() on an immediate parent does not always trigger the normal
code paths that determine that an expression is not a constant. This can
lead to an unknown state and ultimately a NullPointerException. By
overriding the contentLength and valueLength methods in the FakeDINode,
we ensure that if this special case occurs we correctly determine this
expression is not a constant.

Also removes debugging information from the error message that results
when this expression is eventually evaluated.

DAFFODIL-2078

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
@@ -446,6 +446,21 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="valueLength13">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="x" dfdl:lengthKind="explicit" dfdl:length="{ 1 }">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="y" type="xs:int" maxOccurs="2" dfdl:occursCountKind="expression"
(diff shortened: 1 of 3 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -466,6 +480,20 @@
...
       </xs:complexType>
     </xs:element>
+
+    <xs:element name="contentLength3">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="x" dfdl:lengthKind="explicit" dfdl:length="{ 1 }">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="y" type="xs:int" dfdl:occursCountKind="expression" dfdl:occursCount='{ if (dfdl:contentLength(.., "bits") eq 0) then 1 else 2 }' maxOccurs="2" />
```
mbeckerle: Hard to code review because lines are too long in this file. When changes to a file are more than a screenful, you have to scroll down to find the horizontal scroll bar, which makes this very challenging.

stevedlawrence: Always forget about that, I've pushed a fixup commit to fix these.

PR #194, opened 2019-03-08
Fix poor diagnostic.

mbeckerle:
DAFFODIL-2092

This bug was found in a new-user training class in an example using dynamic separators that are the first character of a line.

If you just so happen to have a blank line (just a LF), then that LF gets lifted as the first character of the line, and made into the separator. But LF is whitespace, and separator is defined to be a list of DFDL string literals, separated by whitespace, so cannot contain whitespace characters.

The diagnostic was ineffective enough to make figuring that this is what was happening quite hard. 

Also fixes DAFFODIL-2084 (attributeFormDefault spurious warning issue)

### DIFF
--- a/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
+++ b/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
@@ -98,7 +98,7 @@ public class TestJavaAPI {
              * For more information, see https://github.com/sbt/sbt/issues/163
              */
             @Override
-            protected Class resolveClass(java.io.ObjectStreamClass desc) throws IOException, ClassNotFoundException {
+            protected Class<?> resolveClass(java.io.ObjectStreamClass desc) throws IOException, ClassNotFoundException {
                 try {
                     return Class.forName(desc.getName(), false, getClass().getClassLoader());
                 } catch (ClassNotFoundException e) {
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
@@ -47,8 +47,7 @@ final class EntityReplacer {
   val dfdlCharClassEntityName = "NL|WSP|WSP\\*|WSP\\+|ES"
 
   val entityCharacterUnicode: List[(String, String, Matcher)] =
-    List(
-      ("NUL", "\u0000", Pattern.compile("%" + "NUL" + ";", Pattern.MULTILINE).matcher("")),
+    List(("NUL", "\u0000", Pattern.compile("%" + "NUL" + ";", Pattern.MULTILINE).matcher("")),
       ("SOH", "\u0001", Pattern.compile("%" + "SOH" + ";", Pattern.MULTILINE).matcher("")),
       ("STX", "\u0002", Pattern.compile("%" + "STX" + ";", Pattern.MULTILINE).matcher("")),
       ("ETX", "\u0003", Pattern.compile("%" + "ETX" + ";", Pattern.MULTILINE).matcher("")),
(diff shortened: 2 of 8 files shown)


daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -98,7 +98,7 @@ private DataProcessor reserializeDataProcessor(DataProcessor dp) throws IOExcept
              * For more information, see https://github.com/sbt/sbt/issues/163
              */
             @Override
-            protected Class resolveClass(java.io.ObjectStreamClass desc) throws IOException, ClassNotFoundException {
+            protected Class<?> resolveClass(java.io.ObjectStreamClass desc) throws IOException, ClassNotFoundException {
```
mbeckerle: This was creating a warning in Eclipse for me.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -47,8 +47,7 @@ final class EntityReplacer {
   val dfdlCharClassEntityName = "NL|WSP|WSP\\*|WSP\\+|ES"
 
   val entityCharacterUnicode: List[(String, String, Matcher)] =
-    List(
-      ("NUL", "\u0000", Pattern.compile("%" + "NUL" + ";", Pattern.MULTILINE).matcher("")),
+    List(("NUL", "\u0000", Pattern.compile("%" + "NUL" + ";", Pattern.MULTILINE).matcher("")),
```
mbeckerle: Just eclipse changing the whitespace.... I will see if this can be avoided.

daffodil-lib/src/main/scala/org/apache/daffodil/util/SchemaUtils.scala
```diff
@@ -40,15 +40,13 @@ object SchemaUtils {
    * Constructs a DFDL schema more conveniently than having to specify all those xmlns attributes.
    */
 
-  def dfdlTestSchemaUnqualified(
-    includeImports: Seq[Node],
+  def dfdlTestSchemaUnqualified(includeImports: Seq[Node],
```
mbeckerle: It does not appear that I have a choice to tell Eclipse NOT to change whitespace like this. I would have to completely shut off auto-formatting.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
mbeckerle: Does not seem to be any way to avoid, other than turning off all auto-formatting.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -652,7 +648,11 @@ sealed abstract class ListOfStringLiteralBase(
 
   protected def cook(raw: String, context: ThrowsSDE, forUnparse: Boolean): List[String] = {
     if (raw.length != 0 && (raw.head.isWhitespace || raw.last.isWhitespace)) {
-      context.SDE("The property '%s' cannot start or end with the string \" \", did you mean to use '%%SP;' instead?", propNameArg)
+      val ws = if (raw.head.isWhitespace) raw.head else raw.last
+      val wsVisible = Misc.remapCodepointToVisibleGlyph(ws.toChar).toChar
+      val hexCodePoint = "%04x".format(ws.toInt)
+      context.SDE("The property '%s' cannot start or end with the string \"%s\"(unicode hex code point U+%s), or consist entirely of whitespace."
+        + "\nDid you mean to use character entities like '%%SP;' or '%%NL;' to indicate whitespace in the data format instead?", propNameArg, wsVisible, hexCodePoint)
     }
```
mbeckerle: Unicode is proper noun. Should be capital "U".

daffodil-lib/src/main/scala/org/apache/daffodil/util/SchemaUtils.scala
```diff
@@ -97,7 +94,7 @@ object SchemaUtils {
     scope = XMLUtils.combineScopes("ex", targetNamespace, scope)
 
     val schemaNode =
-      <xs:schema elementFormDefault={ elementFormDefault } attributeFormDefault="unqualified">
+      <xs:schema elementFormDefault={ elementFormDefault }>
```
mbeckerle: This is the DAFFODIL-2084 fix.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
stevedlawrence: That's too bad. The other syntax is definitely the right thing to do here. You should probably look into using ``git add -p`` when adding changes to a commit. It's fine if Eclipse wants to make random changes, but that doesn't mean all those changes should be comitted, like this case. Using the ``-p`` flag will let you control exactly which hunks are added to a commit and you can ignore these wrong changes. Or you chould use ``git checkout -p`` to undo specific changes.

PR #195, opened 2019-03-12
Changes to support building in IntelliJ

stevedlawrence:
- IntelliJ gets confused by our custom scala-debug configuration. Since
  scala-debug isn't all that useful and causes confusions, just remove
  it altogether and move all scala-debug tests to the scala tests
  directory but commented out.
- Fix bug where trying to get the classpath. This doesn't work on
  IntelliJ, so just return an empy classpath.
- Rename dpaext1.tdml since the name was the same as in another jar,
  which cause IntelliJ to pick up the wrong tdml file.

DAFFODIL-2093


daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
```diff
@@ -158,9 +175,35 @@ class TresysTests {
...
+  //@Test def test_BE000() { runnerBE.runOneTest("BE000") } // unordered sequences
+  //@Test def test_BE001() { runnerBE.runOneTest("BE001") }
+
+  //DFDL-1010
+  //@Test def test_BF000() { runnerBF1.runOneTest("BF000") } // unordered sequences
+  //@Test def test_BF001() { runnerBF1.runOneTest("BF001") }
+
+  //@Test def test_BG000() { runnerBG.runOneTest("BG000") } // needs text numbers: advanced properties (DFDL-452)
+
+  //@Test def test_BC000() { runnerBC.runOneTest("BC000") } // text boolean type
```
mbeckerle: Some of these tests "should" work e.g., this one says in comment it needs text boolean type, which we have now.
There's no ticket associated with it (DFDL-XXX or DAFFODIL-XXX) so this will never get revisited..... Suggest we add a structured comment we can search for e.g., "// NO JIRA TICKET" so that we can easily take a pass and find these.

daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes2.scala
```diff
@@ -79,6 +81,11 @@ class TestSimpleTypes2 {
   @Test def test_bitOrderChangeInvalid2() { runner1.runOneTest("bitOrderChangeInvalid2") }
   @Test def test_bitOrderChangeInvalid2Unparser() { runner1.runOneTest("bitOrderChangeInvalid2Unparser") }
   @Test def test_noByteOrder() { runner1.runOneTest("noByteOrder") }
+  //@Test def test_bitOrderChange() { runner1.runOneTest("bitOrderChange") }
```
mbeckerle: Same issue this file. No JIRA ticket means these tests won't get revisited even if a related bug is fixed. Add "// NO JIRA TICKET" so we can revisit and label these (unless you want to label them now, but someone else can do this.)

daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
stevedlawrence: I'm not sure how much that will really help that much since there's already alot of tests that are commented out but don't have associated bugs in the comment. Grepping for ``//.*@Test`` finds just over 300 instances of commented out bugs, but grepping for ``//.*(DFDL|DAFFODIL)-`` finds just under 200 tickets mentioned. Could be some are grouped together under a single issue, but I skimmed through commented out tests and quickly came across many that didn't have an associated bug.

Maybe we should instead create a new issue to go through commented out tests and either enable those that pass or create bugs for those that don't pass but don't have an associated bug?

mbeckerle: That's fine. I just wanted to highlight the issue since we no longer have an easy bulk way to just run all the debug tests. 

Also.... there might already be such a ticket.... vaguely recall.

stevedlawrence: I couldn't find a ticket, so I've created [DAFFODIL-2095](https://issues.apache.org/jira/browse/DAFFODIL-2095) to test our commented out bugs and make sure those that fail have associated bug numbers.

PR #196, opened 2019-03-15
Use setScale() not round() when rounding BigDecimals

jadams-tresys:
In order to perform a round/floor/ceiling on a BigDecimal and not retain
any decimal points we need to use BD.setScale with a precision of 0. When
the BD.round function is used with a precision or 0, no rounding is
performed.

DAFFODIL-2076

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
@@ -685,7 +685,7 @@ case class FNCeiling(recipe: CompiledDPath, argType: NodeInfo.Kind) extends FNOn
 
     case NodeInfo.Decimal => {
       val bd = asBigDecimal(value)
-      bd.round(new MathContext(0, RoundingMode.CEILING))
+      bd.setScale(0, RoundingMode.CEILING)
     }
     case NodeInfo.Float => asAnyRef(asFloat(value).floatValue().ceil)
     case NodeInfo.Double => asAnyRef(asDouble(value).floatValue().ceil)
@@ -699,7 +699,7 @@ case class FNFloor(recipe: CompiledDPath, argType: NodeInfo.Kind) extends FNOneA
 
     case NodeInfo.Decimal => {
       val bd = asBigDecimal(value)
-      bd.round(new MathContext(0, RoundingMode.FLOOR))
+      bd.setScale(0, RoundingMode.FLOOR)
     }
     case NodeInfo.Float => asAnyRef(asFloat(value).floatValue().floor)
     case NodeInfo.Double => asAnyRef(asDouble(value).doubleValue().floor)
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -685,7 +685,7 @@ case class FNCeiling(recipe: CompiledDPath, argType: NodeInfo.Kind) extends FNOn
 
     case NodeInfo.Decimal => {
       val bd = asBigDecimal(value)
-      bd.round(new MathContext(0, RoundingMode.CEILING))
+      bd.setScale(0, RoundingMode.CEILING)
```
mbeckerle: If this is a side-effect on this big decimal object, that object could be referenced from say, an element in the infoset. A formula elsewhere in the schema could be accessing its value and calling a round on it. 

So then the original element would have changed also.

I think you need to make a copy before you setScale.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -4324,6 +4324,8 @@
       </xs:complexType>
     </xs:element>
     
+    <xs:element name="ceil08" type="xs:decimal" dfdl:inputValueCalc="{ fn:ceiling(xs:decimal(0.125)) }"/>
```
mbeckerle: You can test my theory about the side-effect of setScale by creating a test with two elements. The first is a bigDecimal type with a fractional part. The second is dfdl:inputValueCalc=".... access first element and round it....". If setScale modifies the object, then BOTH elements will have had the scale set.

jadams-tresys: I'll add the test, but I think that it should be OK since setScale actually returns a copy with the scale set to the new value.

jadams-tresys:
Just added a test to confirm that BigDecimal.setScale does in fact return a copy of the original BigDecimal with its scale changed, so it is side effect free.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
mbeckerle: Grrrr. I guess I'm not surprised.  Numbers are supposed to be immutable, so I'm very glad this is the case.

(I think it was unwise to prefix the name with "set..." if it returns a new value. "rescale" would have been a perfectly good name. I guess whoever created the setScale() method got it through code-review without someone raising this objection. Oh well. )

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -6048,6 +6056,28 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+<!--
+    Test Name: round_19
+       Schema: XPathFunctions
+         Root: round10
+      Purpose: This test demonstrates the round function with xs:decimal and produces no side effects
```
mbeckerle: Typo "and" unnecessary.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
jadams-tresys: I agree about using "set...", but then again common sense would also lead people to believe you should use the "round" function for rounding, so I'm not sure they were really going for common sense.

stevedlawrence:
+1

PR #197, opened 2019-03-15
Fix non-thread safe code related to text date times

stevedlawrence:
We create an ICU Calendar object to handle parsing/unparsing text date
times. Unfortunately, this Calendar object isn't thread safe, but we
share it among all threads when all the DFDL calendar properties are
constant. This can lead to random errors.

To make it thread safe, change the text calendar code to never modify
the original Calendar object, but instead create a clone of it before
use and use the clone for all dateTime calculations. The effectively
treats the original calendar as immutable and avoids thread safety
issues. Also reorganize the code a bit and rename some variables to make
the purpose of the different Calendar objects more clear.

DAFFODIL-2097

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextCalendarUnparser.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextCalendarUnparser.scala
@@ -42,51 +42,72 @@ case class ConvertTextCalendarUnparser(
   override lazy val runtimeDependencies = Vector(localeEv, calendarEv)
 
   def unparse(state: UState): Unit = {
-    val locale: ULocale = localeEv.evaluate(state)
-    val calendar: Calendar = calendarEv.evaluate(state)
 
     val node = state.currentInfosetNode.asSimple
 
     val dc = node.dataValue
 
-    val calValue = node.dataValue match {
+    val infosetCalendar = node.dataValue match {
       case dc: DFDLCalendar => dc.calendar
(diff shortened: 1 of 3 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextCalendarUnparser.scala
```diff
@@ -42,51 +42,72 @@ case class ConvertTextCalendarUnparser(
...
+    // 1) The below code will modify the calendar object based on the dateTime
+    //    in the infoset. Any changes to the object would persist and affect
+    //    future unparses. By cloning, we ensure we do not modify the original
+    //    calendar object that other unparses will use.
+    //
+    // 2) Multiple threads would have access to the same Calendar object, and
+    //    so the below code could modify the same object at the same time
+    //    across threads. By cloning, we ensure that they modify different
+    //    objects.
+    val calendar = calendarOrig.clone().asInstanceOf[Calendar]
```
mbeckerle: I guess I agree, we should just allocate a new calendar here. If these things are going to be so stateful no point in bending over backwards to cache/avoid allocating them.

stevedlawrence: Yeah, it's pretty unfortunate. We could alternatively do thread local stuff, but it just becomes a pain and is more complex. And I'm not sure if there's a big gain, I would imagine the synchronizing that a thread local requires is more overhead that allocating a new Calendar object--it's not a super complex object.

Ultimately, we should switch to library that isn't so stateful.

PR #200, opened 2019-03-18
Avoid supressing original parse errors

bsloane1650:
After parsing an element, regardless of if an error was thrown,
additional state validation is conducted. If said validation fails,
following an error being thrown in the parse function, we now include
the original error as a supressed error in the final one.

DAFFODIL-2099

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
@@ -60,6 +60,8 @@ import org.apache.daffodil.util.Pool
 import org.apache.daffodil.util.Poolable
 import org.apache.daffodil.infoset.DIComplexState
 import org.apache.daffodil.infoset.DISimpleState
+import org.apache.daffodil.exceptions.UnsuppressableException
+import org.apache.daffodil.exceptions.Abort
 
 object MPState {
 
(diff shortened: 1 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -336,18 +336,26 @@ final class PState private (
...
+        Assert.invariant(this.discriminatorStack.length == 1)
+        mpstate.verifyFinalState()
+      }
+      // These we check regardless of throw or not.
+      markPool.finalCheck
+      dataInputStream.inputSource.compact // discard any storage that can be freed.
+      dataInputStream.validateFinalStreamState
+    }catch{
+      case e:Throwable => {
+        println(s"Suppressed error: ${optThrown}")
```
stevedlawrence: We probably don't want a println here, since not all uses of daffodil will be from a console, so this would get lost. Instead, we probably want to use our ``log()`` method so that the user can capture this log message in a way that they can display it. The default logger goes to stderr, but is configurable..

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -336,18 +336,26 @@ final class PState private (
...
+        if(optThrown.isDefined) e.addSuppressed(optThrown.get)
+        throw e
+      }
```
stevedlawrence: Double check the code style in your changes. Missing some spaces. For example there should be a space in ``try {``, ``} catch {``, and a space between ``if (``. Also, ``} finally {`` should all be on one line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
bsloane1650: Looking at this again, I do not see the need for any special print/log here. Since it is now included as a suppressed error of what is thrown, it should get displayed along with whatever displays that.

What I do see is that the error being thrown here will potentially be interperated as a control-flow exception in Runtime.scala, even if we are supressing a non control-flow exception. Currently this won't happen, and I can't imagine why someone would even attempt throwing, say, an SDE from here, but it is non-obvious that such would be a problem. To avoid this issue, I am going to be wrapping all the errors in a Abort (which subclasses unsupressableException).

bsloane1650:
New version uploaded

stevedlawrence:
+1 to the new changes

PR #201, opened 2019-03-18
Fix bug in handling parse errors

bsloane1650:
ProcessingError.toParseError would erroneasuly pass itself as the cause
to the ParseError instead of passings it's maybeCause

This is particuarly problematic because it would pass its maybeFormatString,
however only one of the maybeFormatString and maybeCause may be defined,
so if maybeFormatString is defined, then something will be passed for
both and no useful error message will be created.

DAFFODIL-2100

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessingError.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessingError.scala
@@ -47,13 +47,13 @@ abstract class ProcessingError protected (
    * looks like the same as other parse errors to tests that search for the
    * "Parse Error" string.
    */
-  def toParseError = new ParseError(schemaContext, dataContext, Maybe(this), maybeFormatString, args: _*)
+  def toParseError = new ParseError(schemaContext, dataContext, Maybe(this), Maybe.Nope)
 
   /**
    * Used to convert a processing error into a unparse error so that it
    * looks like the same as other unparse errors to tests that search for the
    * "Unparse Error" string.
    */
-  def toUnparseError = new UnparseError(schemaContext, dataContext, maybeCause, maybeFormatString, args: _*)
+  def toUnparseError = new UnparseError(schemaContext, dataContext, Maybe(this), Maybe.Nope)
 
 }


stevedlawrence:
It looks like this change now causes a single test to fail:
```org.apache.daffodil.section23.dfdl_expressions.TestDFDLExpressions.test_int_constructor_04```

The test expects the string "expression evaluation error". It used to get this error message:

> Parse Error: Cannot convert 'not an int!' from String type to Long (Cannot convert to type long: For input string: "not an int!"). (within Expression Evaluation Error: Cannot convert 'not an int!' from String type to Long (Cannot convert to type long: For input string: "not an int!").

But it now gets just this:

> Parse Error: Cannot convert 'not an int!' from String type to Long (Cannot convert to type long: For input string: "not an int!").

So we're losing the information that the parse error actually occurred in an expression evaluation error. There's definiately a lot of redundancy in that error message, but that fact that it occurred in an expression is useful information that I don't think we want to lose. Perhaps there's a different underlying issue here, like something using ProcessingError incorrectly? The bug mentions it's that maybeFormatString and maybeCause shouldn't both be passed in? Ifsomethign is using this incorrectly, maybe combining the paramters into a single paramter that is an ``Either[String,Throwable]`` would the type system figure out where we're doing something wrong.

It would also be helpful if you could add a tdml test that reproduces the bad error so that we can ensure the test is fixed and does not regress in the future.

**Comment by bsloane1650** [comment]
Unfortunately, the original error (which caused the poorly behaving Parse error) was in simply broken code so I don't know of any way to trigger it from a TDML. I can probably create a unit test that exersises that particular error path.

Updating the type to be an Either type seems like the correct solution. However, as we are in the runtime system, I am wondering if the use of Maybe here was a deliberate choice for performance reasons. (I believe these objects get created routinely as part of backtracking).

I am also a bit suspicious of the way the original asymetry between the parse and unparse case:

```
def toParseError = new ParseError(schemaContext, dataContext, Maybe(this), maybeFormatString, args: _*)

def toUnparseError = new UnparseError(schemaContext, dataContext, maybeCause, maybeFormatString, args: _*)
``` 

Since you are not allowed to have both maybeCause and maybeFormatString, I didn't see how the parse case could ever be correct, so I just copied the unparse case. (In particular, a caller may never pass maybeFormatString to ProcessingError if it would ever be converted to a ParseError, and I suspect that calling _.toParseError.toParseError will behave silly in that the error will be nested an additional time for each call.)

Looking at the code, it seems the info being lost by not including "this" as the cause is the modeName arguement passed to the outer ProcessingError ("Expression Evaluation" in this case). Unfortunately, ParseError makes a point of being a "Parse" error which, by definition, overwrites the value of modeName with "Parse".

Maybe it makes sense for modeName to be some sort of list? so the generated error looks more like:

```
Parse Error: Expression Evaluation Error: Cannot convert 'not an int!' from String type to Long (Cannot convert to type long: For input string: "not an int!").
```

(The repetition of "Error" is independent, but probably a good idea from the perspective of string matching).

File: daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessingError.scala [hunk 1]
```diff
@@ -47,7 +47,7 @@ abstract class ProcessingError protected (
    * looks like the same as other parse errors to tests that search for the
    * "Parse Error" string.
    */
-  def toParseError = new ParseError(schemaContext, dataContext, Maybe(this), maybeFormatString, args: _*)
+  def toParseError = new ParseError(schemaContext, dataContext, maybeCause, maybeFormatString, args: _*)
```
mbeckerle: Changing this to maybeCause isn't correct. 
'this' object itself becomes the cause of the new ParseError. You are jumping ahead one in the chain of nested exceptions by using maybeCause here.

bsloane1650:
I think it is clear that the code prior to this PR is broken for the reasons I discussed above. Would it make sense to do?:

```
def toParseError = new ParseError(schemaContext, dataContext, Maybe(this), Maybe.Nope)

def toUnparseError = new UnparseError(schemaContext, dataContext, Maybe(this), Maybe.Nope)
```

This maintains the invariant that exactly 1 of maybeCause and maybeFormatString is defined. I believe the content of formatString will still end up getting displayed when this is displayed as the cause of the resulting message.

mbeckerle:
Yes I think both toParseError and toUnparseError should be using Maybe(this) for the cause. Both should be chaining the exceptions the same way.

The concept here. Innermost exception specifies a format string. Enclosing exceptions have causes only, and *never* have format strings. At actual message creation time each exception contributes to the ultimate message string based on its class name. 

So you should get "Parse Error (caused by) Expression Evaluation Error: actual message string from innermost class goes here". Though I think "(caused by)" is a bit much. I'd just use a ":".

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessingError.scala
bsloane1650: Updated PR.

bsloane1650:
Commit message updated

PR #202, opened 2019-03-28
Daffodil 2095 commented out tests

jadams-tresys:
This pull request includes updates from a review of all our commented out tests.  Some tests work now and were enabled, others needed to have associated bug tickets created or noted, and some tests needed to be fixed before they could be used.


daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
```diff
@@ -123,40 +123,33 @@ class TresysTests {
...
   // @Test def test_nested_separator_delimited_nest2() { runnerNSD.runOneTest("nest2")}
-  // Fails infinite loop
-  // @Test def test_nested_separator_delimited_nest3() { runnerNSD.runOneTest("nest3")}
+  @Test def test_nested_separator_delimited_nest3() { runnerNSD.runOneTest("nest3")}
 
   /* Very big test data files, so each is in its own TDML file */
 
-  //  @Test def test_AB007() { runnerAB7.runOneTest("AB007") }
-  //  @Test def test_AB008() { runnerAB8.runOneTest("AB008") }
-  //  @Test def test_AB009() { runnerAB9.runOneTest("AB009") }
```
stevedlawrence: Looks like tests AB008 and AB009 were removed?

daffodil-test/src/test/resources/org/apache/daffodil/section12/aligned_data/Aligned_Data.tdml
```diff
@@ -2090,11 +2090,10 @@
   </tdml:document>
   <tdml:infoset>
     <tdml:dfdlInfoset><e10b><varArr><e>11</e><e>4</e><e>14</e><e>9</e><e>11</e></varArr><one>11</one></e10b>
-    </tdml:dfdlInfoset>
+  </tdml:dfdlInfoset>
   </tdml:infoset>
   <tdml:warnings>
-    <tdml:warning>Schema Definition Warning</tdml:warning>
-    <tdml:warning>e is an optional element or a variable-occurrence array and its alignment (2) is not the same as one's alignment (1)</tdml:warning>
+    <tdml:warning>Schema Definition Warning Palceholder</tdml:warning>
```
stevedlawrence: Was this "placeholder" intentional?

daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/AQ.tdml
```diff
@@ -76,46 +76,4 @@
...
-    <tdml:document>
-      <tdml:documentPart type="text"><![CDATA[10.5,12.01
- ]]></tdml:documentPart>
-    </tdml:document>
-
-    <tdml:errors>
-      <tdml:error>Schema definition error</tdml:error>
-    </tdml:errors>
-
-  </tdml:parserTestCase>
```
stevedlawrence: Schema and test completely removed. Was that intentional?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -3252,14 +3252,9 @@
...
-    <tdml:infoset>
-      <tdml:dfdlInfoset>
-        <s1>
-          <e1>45</e1>
-          <extra>someletters</extra>
-        </s1>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
+    <tdml:errors>
+      <tdml:error>This expression is not supported in the current option setting</tdml:error>
```
stevedlawrence: This is kindof a weird error message, I"m not even sure what it means but it looks like the issue is the test is using a 4-byte unicode character in the regular expression? Is there extra information in the error message that the test just isn't looking for that makes the problem more clear? If not, it might be worth opening a bug to make this error message better, to at least make it clear something is wrong with the regular expression as opposed to  a DFDL expression.

daffodil-test/src/test/scala/org/apache/daffodil/section07/assertions/TestAssertions.scala
```diff
@@ -98,6 +98,4 @@ class TestAssertions {
   @Test def test_assertWithMessageExpression_01() { runner.runOneTest("test_assertWithMessageExpression_01") }
   @Test def test_assertWithMessageExpression_02() { runner.runOneTest("test_assertWithMessageExpression_02") }
 
-  //DFDL-1210
-  //@Test def test_assertFailShowsValue() { runner.runOneTest("assertFailShowsValue") }
```
stevedlawrence: Was this removed by accident? Looks like the test is defined in assert.tdml.

daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
jadams-tresys: It seems that they were removed before.  There doesn't appear to be a TDML file for either of these tests anymore.

daffodil-test/src/test/resources/org/apache/daffodil/section12/aligned_data/Aligned_Data.tdml
jadams-tresys: I originally thought the SDW would have a different message for some reason, but it looks like that is the correct message to receive. I'll revert this.

daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/AQ.tdml
jadams-tresys: Yes, the schema and input were both identical to AQ000.  I'm not sure why AQ001 was expected to fail, but my guess is that someone copied AQ000 to set up a negative test case, but never changed it.

daffodil-test/src/test/scala/org/apache/daffodil/section07/assertions/TestAssertions.scala
jadams-tresys: I think it was.  I actually  needed to reopen DAFFODIL-752 for this issue, as the extra detail in assertion failed messages seems to have been lost.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
jadams-tresys: Not sure why I picked that as the error message to check for.  The error message does include more useful information and I have updated the test to check for that.

daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
mbeckerle: Long way back, but I think there were just too many variations in data volume for what was the same test really, so we removed some of them.

PR #203, opened 2019-04-01
Daffodil 752 assertion details

jadams-tresys:
Pass along more detail on assertion failures
    
We had originally provided more detail in the assertion failed messages,
    including the value that was parsed, but this detail was lost during a
    redesign of the error handling. This commit adds back in the extra
    detail for AssertionFailure's.

This is set up to merge cleanly with the changes in PR #202 so ideally it would be merged after #202 is accepted.


jadams-tresys:
This didn't work as cleanly as I expected, closing for now and will re-open once #202 is accepted.

PR #205, opened 2019-04-08
Update Line counter for new modules and schemas.

mbeckerle:
DAFFODIL-2106


daffodil-lib/src/test/scala/LineCounter.scala
```diff
@@ -30,33 +30,48 @@ object LineCounter extends App {
...
+    "tutorials",
+    "test-stdLayout")
   /**
-   * What modules within the sandbox to count
+   * What modules within the daffodil subtree sandbox to count
    */
-  val srcModulesToInclude = List("daffodil-core", "daffodil-runtime1", "daffodil-runtime1-unparser", "daffodil-lib", "daffodil-propgen", "daffodil-tdml", "daffodil-io", "daffodil-cli", "daffodil-japi", "daffodil-sapi")
-  val unitTestModulesToInclude = srcModulesToInclude
+  val srcModulesToInclude = {
+    val daf = new File(root + "daffodil")
```
stevedlawrence: This assumes the daffodil source directory is called "daffodil". Which isn't always true. Some other common names are going to be "incubator-daffodil" or "daffodil.git", "incubator-daffodil.git", etc. This shouldn't be hardcodede in here.

daffodil-lib/src/test/scala/LineCounter.scala
```diff
@@ -30,33 +30,48 @@ object LineCounter extends App {
...
+
+  /**
+   * What to exclude (subversion artifacts, anything generated, version control stuff, caches, settings, etc.)
+   */
+  val directoriesToExclude = List(".svn", ".git", ".settings", "lib_managed", "src_managed", "resource_managed", "bin", ".cache", "target", "lib", ".idea")
 
   /**
    * Where is this freakin' sandbox anyway....edit to point to yours.
    */
   val root = "/home/mbeckerle/dataiti/git/"
```
stevedlawrence: This line is specific to your layout, this code will only work on your machine. This should be generalized.

daffodil-lib/src/test/scala/LineCounter.scala
```diff
@@ -71,35 +86,24 @@ object LineCounter extends App {
...
-          new LineCounter("calabash fouo", root + "daffodil-fouo", List("calabash-test"), List("src/main", "src/test"),
-            fileSuffixesToInclude = List(".scala", ".java", ".tdml", ".xpl", ".sbt")),
-          new LineCounter("NiFi", root + "daffodil-nifi", List("nifi-daffodil-nar", "nifi-daffodil-processors"), List("src/main", "src/test")),
-          new LineCounter("Spark", root, List("daffodil-spark"), List("src/main", "src/test")))),
-      new LineCounter("all sbt files", root, List("daffodil", "fouo-schemas", "DFDLSchemas", "daffodil-calabash-extension",
-        "daffodil-nifi", "jpeg2000", "daffodil-nifi", "daffodil-spark", "ipfix", "asterix"), List(""), fileSuffixesToInclude = List(".sbt")))
+        List(new LineCounter("calabash proper", root + "daffodil-calabash-extension", List("calabash-server")),
+          new LineCounter("calabash fouo", root + "daffodil-fouo", List("calabash-test")),
+          new LineCounter("NiFi", root + "daffodil-nifi", List("nifi-daffodil-nar", "nifi-daffodil-processors")),
+          new LineCounter("Spark", root, List("daffodil-spark")))))
```
stevedlawrence: This is all specific to your directory structure, and really aren't Daffodil related, aside from the fact they they're DFDL schemas. To me. this feels like this line counter should live outside this Daffodil repo and should have a config file to specicy the directories and file extensions to include/exclude. Maybe it should live in OpenDFDL?

In fact, I'm not sure if it's evn worth maintaining this. Seems effort would be better spent using the [sbt-status](https://github.com/orrsella/sbt-stats) plugin.  The looks pretty extendable to add new metrics like TDML files, DFDL schema files, etc. Then it's just a matter a running a script that runs "sbt stats" on each project and adding up the numbers.

daffodil-lib/src/test/scala/LineCounter.scala
```diff
@@ -116,20 +120,20 @@ trait LineCounterBase {
 
   def pairs: List[(String, Int)]
 
-  final def total = pairs.map { _._2 }.sum
+  final lazy val total = pairs.map { _._2 }.sum
 
   protected final def linesWithoutCopyright(f: File): List[String] = {
```
stevedlawrence: This function no longer removes the copyright. Maybe it didn't work all that well? Seems like this is going to increase line numbers alot. Doesn't seem unreasonable, but just have to note that we shouldn't compare old numbers with new.

daffodil-lib/src/test/scala/LineCounter.scala
mbeckerle: Created an improved sbt-stats which is at OpenDFDL/sbt-stats.
I have a pull request open to the original sbt-stats project, but for now this is available via the fork.

project/plugins.sbt
```diff
@@ -32,3 +32,5 @@ dependencyOverrides ++= Seq(
   "org.codehaus.plexus" % "plexus-utils" % "3.0.17",
   "com.google.guava" % "guava" % "20.0"
 )
+
+// addSbtPlugin("com.orrsella" % "sbt-stats" % "1.0.9-SNAPSHOT")
```
stevedlawrence: Instead of doing a pull rquest to the sbt-stats plugin, what if instead of modifying the sbt-status plugin we create our own Analyzer that gathers the stats we care about (e.g. test vs src, tdml vs scala vs xml, etc.). Then we can configure the plugin in our build.sbt to use our custom analyzer. A bunch of the default stats (file size, chars, comments) seem kindof uninteresting to me. Setting own Analyzer would let us remove those and just output what we care about. Seems like an eaiser path and more configurable towards what we want than trying to upstream to sbt-stats.

mbeckerle: Can't be done. sbt-stats isn't flexible enough in the way it feeds files for processing to the analyzers. Basically, if sbt sources doesn't have the files you want stats on, you can't analyze them. So I modified it to pass separately to the analysis 4 things - sources, test sources, resources, test resources. That way analyzers can do what they want.

stevedlawrence: Ah, that's a bummer. So it only provides the files in ``src/main/{java/scala}``? Maybe a more backwards compatible change and something more sbt-y would be to do something like:

1. Add a new ``analyzeResources()`` method, which would be called in additional to analyze, and would pass in resources for when people need to analyze stuff in ``src/main/resources/``. A default method could be provided to in the base class to return nothing so existing stuff doesn't break.
2. Make it so stats can run in the "test" configuration. Then you could run ``sbt test:stats`` to get stats about things in ``src/test/{java,scala}`` and ``src/test/resources``.'

These changes would make it so output doesn't change for anyone already using the plugin, but could be used for people like us that need status about resources and tests.

mbeckerle: I'm open to these ideas. 

They are unrelated to this PR which just deletes the old line-counter and suggests, via a comment in the plugins,  that sbt-stats be used.

The PR of changes to sbt-stats is:

https://github.com/orrsella/sbt-stats/pull/13

What is there and on openDFDL/sbt-stats computes the stats we've been incorporating into presentations about Daffodil, so meets our needs. But as you point out, there might be a more extensible way to do this change though that would maintain compatibility for existing sbt-stats users.

stevedlawrence: Yeah, I would just prefer to not have to maintain a fork of sbt stats. If we could get something upstreamed that would be ideal. If they accept the current PR, that work for me.

mbeckerle: Not sure anything is being upstreamed. There's no activity for 2 years on sbt-stats.

stevedlawrence: Ugh, I guess it doesn't matter. Seems fine to me.

Only thing I'd say is that the commented out ``addSbtPlugin`` line might be confusing since the "real" sbt-stats plugin doesn't have a version 1.0.9-SNAPSHOT. Thoughts on removing that line and we can provide documentation somewhere (maybe on the wiki?) on how to install and enable the custom line counter tailored towards counting lines in Daffodil and DFDL projects?

Something like:
1. CLone our fork from XYZ
2. Install it with ``sbt localInstall``
3. Modify ``~/.sbt/1.0/plugins/plugins.sbt`` to have the ``addSbtPlugin`` line
4. Run ``sbt stats`` on any project (Daffodil or DFDL schema project) to get the stats for that project.

+1

mbeckerle: Created https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Stats+For+Reporting

stevedlawrence: 2 should actually be 'sbt publishLocal'. I've updated the wiki.

PR #206, opened 2019-04-10
Decrease time needed for continue integration builds

stevedlawrence:
- In the TravisCI configuration, we only enabled coverage when running
  tests. However, coverage actually needs to be enabled during
  compilation, since that enables a compiler plugin. By only enabling it
  when running tests, sbt will recompile when the tests are run, so we
  end up compiling Daffodil twice. Instead, always enable coverage on
  every sbt command so that we only compile once.
- In the TravisCI configuration, enable caching of ivy and sbt
  directories and provide a few extra memory options that seem to help.
- Also modify TravisCI so that when RAT check fails it cats the rat report.
- Add support for Azure Pipelines. This isn't supported (yet) by ASF
  Infra, but it has alot of advantages over TravisCI. It supports more
  parallel builds, has a larger maximum build time, organizes build
  stages better, handles our large text output, supports Windows.
  Additionally, it supports JUNit test integration so we can easily tell
  exactly which tests fail and how long each test took to run. The
  configuration file is more complex than TravisCI, but is much more
  powerful and should be our primary CI tool once ASF supports it.
- Remove the large AB tests. Not only do these fail on Windows due to
  line ending issues, but they do not provide any additional test
  coverage.
- Some CLI performance command tests expected no output, which required
  a timeout, which just takes to long. So just remove the expectation of
  nothing. The tests do continue and essentially check the same thing,
  just in a different and faster manner.

These changes get TravisCI build consistently between 30-35 minutes,
rather the 40-50+, which caused a build timeout. Note that Azure
Pipelines consistently takes between 20-25 minutes, so we should switch
as soon as ASF supports it.

DAFFODIL-2105


daffodil-cli/src/it/scala/org/apache/daffodil/executing/TestCLIexecuting.scala
```diff
@@ -123,27 +123,27 @@ class TestCLIexecuting {
...
-      val numTests = Integer.parseInt((String.format("%s test -l %s", Util.binPath, testTdmlFile) #| countTests !!).trim())
-      val numFile = Integer.parseInt((String.format(if (Util.isWindows) "cmd.exe /c more %s | find /c \"parserTestCase>\"" else "grep -c parserTestCase> %s", testTdmlFile)!!).trim())
-
-      assertTrue("Number of tests run should match the number of tests in the file.", numTests == numFile)
-      shell.sendLine("exit")
-    } finally {
-      shell.close()
-    }
-  }
+//  @Test def test_993_CLI_Executing_Listing_listAll() {
```
mbeckerle: Why keep this? Probably should just delete.

stevedlawrence: This is actually broken on Azure Pipelines on Windows, but works fine on linux. I couldn't figure out the problem so I temporarily commented it out and just forgot about it. I'll see if I can fix it, and if not I'll reenable it and then we can deal with it when ASF allows Azure Pipelines support.

stevedlawrence:
For reference, here is the Azure build I set up:

https://dev.azure.com/stevedlawrence/incubator-daffodil.git/_build?definitionId=2

daffodil-cli/src/it/scala/org/apache/daffodil/executing/TestCLIexecuting.scala
stevedlawrence: Actually, this test was just trying to be too smart. I've simplfied it a bit while still sufficiently testing listing tdml files. I've uncommented it and it should now run on windows.

PR #207, opened 2019-04-10
Added support for enumerations and TypeValueCalc

bsloane1650:
Not quite done yet. I still want to improve the test coverage. But, given the size, I wanted to start the review process as well now that I am passing my initial test set.

This is a fairly big change.

In addition to the actual implementation, and some minor hooks throughout,
this also involved moving a fair amount of the DSOM functionality into
the factory classes.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -1543,6 +1543,49 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
         DFDLSetBitsExpr(functionQNameString, functionQName, args)
       }
 
+      //Begin TypeValueCalc related functions
+      case (RefQName(_, "inputTypeCalcInt", DFDL), args) =>
+        FNTwoArgsExprInferedArgType(functionQNameString, functionQName, args,
+          NodeInfo.Int, NodeInfo.String, args(1).inherentType, DFDLXInputTypeCalcInt(_))
+
+      case (RefQName(_, "inputTypeCalcString", DFDL), args) =>
+        FNTwoArgsExprInferedArgType(functionQNameString, functionQName, args,
+          NodeInfo.String, NodeInfo.String, args(1).inherentType, DFDLXInputTypeCalcString(_))
+
+      case (RefQName(_, "outputTypeCalcInt", DFDL), args) =>
+        FNTwoArgsExprInferedArgType(functionQNameString, functionQName, args,
+          NodeInfo.Int, NodeInfo.String, args(1).inherentType, DFDLXOutputTypeCalcInt(_))
+
+      case (RefQName(_, "outputTypeCalcString", DFDL), args) =>
+        FNTwoArgsExprInferedArgType(functionQNameString, functionQName, args,
+          NodeInfo.String, NodeInfo.String, args(1).inherentType, DFDLXOutputTypeCalcString(_))
+
+      case (RefQName(_, "outputTypeCalcNextSiblingInt", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.Int, NodeInfo.AnyAtomic, DFDLXOutputTypeCalcNextSiblingInt(_, _))
+
+      case (RefQName(_, "outputTypeCalcNextSiblingString", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.String, NodeInfo.AnyAtomic, DFDLXOutputTypeCalcNextSiblingString(_, _))
+
+      case (RefQName(_, "repTypeValueInt", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.Integer, NodeInfo.AnyAtomic, DFDLXRepTypeValueInt(_, _))
+
+      case (RefQName(_, "repTypeValueString", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.String, NodeInfo.AnyAtomic, DFDLXRepTypeValueString(_, _))
+
+      case (RefQName(_, "logicalTypeValueInt", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.Integer, NodeInfo.AnyAtomic, DFDLXLogicalTypeValueInt(_, _))
+          
+      case (RefQName(_, "logicalTypeValueString", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.String, NodeInfo.AnyAtomic, DFDLXLogicalTypeValueString(_, _))
+
+      //End typeValueCalc related functions
+
       case (RefQName(_, "round-half-to-even", FUNC), args) if args.length == 1 => {
         FNOneArgMathExpr(functionQNameString, functionQName, args,
           FNRoundHalfToEven1(_, _))
@@ -1913,7 +1956,8 @@ abstract class FNTwoArgsExprBase(nameAsParsed: String, fnQName: RefQName,
   override lazy val compiledDPath = {
     val arg1Recipe = arg1.compiledDPath
     val arg2Recipe = arg2.compiledDPath
-    val res = new CompiledDPath(constructor(List(arg1Recipe, arg2Recipe)) +: conversions)
+    val recipe = constructor(List(arg1Recipe, arg2Recipe))
+    val res = new CompiledDPath(recipe +: conversions)
     res
   }
 }
@@ -1922,6 +1966,28 @@ case class FNTwoArgsExpr(nameAsParsed: String, fnQName: RefQName,
   args: List[Expression], resultType: NodeInfo.Kind, arg1Type: NodeInfo.Kind, arg2Type: NodeInfo.Kind,
   constructor: List[CompiledDPath] => RecipeOp)
   extends FNTwoArgsExprBase(nameAsParsed, fnQName, args, resultType, arg1Type, arg2Type, constructor)
+
+/*
+ * Used when the underlying constructor does not inherantly know its arguement types will be.
+ * Note that we do not actually "infer" the types here, as we can get them from the args parameter
+ * as easily as the caller, and the caller might be aware of more constraints than we are.
+ * (Eg. in the DFDLXTypeInputCalc functions, only the second parameter is infered, the first
+ * must be a String).
+ * 
+ * Note that, for arguements that should be inferred, the caller should use arg.inherentType,
+ * not arg.targetType, as the latter will look to this class to make its determination.
+ */
+case class FNTwoArgsExprInferedArgType(nameAsParsed: String, fnQName: RefQName,
+  args: List[Expression], resultType: NodeInfo.Kind, arg1Type: NodeInfo.Kind, arg2Type: NodeInfo.Kind,
+  constructor: List[(CompiledDPath, NodeInfo.Kind)] => RecipeOp)
+  extends {
+    private val constructor_ : List[CompiledDPath] => RecipeOp = (subExprs : List[CompiledDPath]) => {
+      Assert.invariant(subExprs.length == args.length)
+      val types = args.map(_.targetType)
+      val typedSubExprs = subExprs.zip(types)
+      constructor(typedSubExprs)
+    }
+  } with FNTwoArgsExprBase(nameAsParsed, fnQName, args, resultType, arg1Type, arg2Type, constructor_)
 
 sealed abstract class LengthExprBase(nameAsParsed: String, fnQName: RefQName,
   args: List[Expression], resultType: NodeInfo.Kind, arg1Type: NodeInfo.Kind, arg2Type: NodeInfo.Kind,
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
@@ -38,11 +38,18 @@ import org.apache.daffodil.api.WarnID
  * property resolution.
  *
  * The only objects that should resolve properties are
- * ElementRef, Root, LocalElementDecl, Sequence, Choice, SequenceRef, ChoiceRef
+ * ElementRef, Root, LocalElementDecl, Sequence, Choice, SequenceRef, ChoiceRef,
+ * EnumerationFactory, SimpleTypeDefFactory,
  *
- * These are all the "real" terms. Everything else is just contributing
+ * All "real" terms are able to resolve properties. Most other objects just contribute
  * properties to the mix, but they are not points where properties are
  * used to generate processors.
+ * 
+ * EnumerationFactory and SimpleTypeDefFactory are the oddballs out. In addition to 
+ * being used to generate processors, these classes our also used to generate abstract 
+ * TypeCalculators, which are not nessasarily attached to any particular element, nor 
+ * used to generate any processor (for instance, there may be a globalSimpleType whose 
+ * only purpose is to define a TypeCalculator for use in DPath expressions)
  */
 trait ResolvesProperties
   extends FindPropertyMixin { self: AnnotatedSchemaComponent =>
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ChoiceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ChoiceGroup.scala
@@ -32,6 +32,10 @@ import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.processors.ChoiceDispatchKeyEv
 import org.apache.daffodil.schema.annotation.props.gen.YesNo
 import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.schema.annotation.props.gen.ChoiceKeyKindType
+import org.apache.daffodil.schema.annotation.props.Found
+import org.apache.daffodil.processors.SimpleTypeRuntimeData
+import org.apache.daffodil.schema.annotation.props.NotFound
 
 /**
  *
@@ -101,7 +105,8 @@ abstract class ChoiceTermBase(
   extends ModelGroup(position)
   with Choice_AnnotationMixin
   with RawDelimitedRuntimeValuedPropertiesMixin // initiator and terminator (not separator)
-  with ChoiceGrammarMixin {
+  with ChoiceGrammarMixin 
+  with HasOptRepTypeMixinImpl {
 
   requiredEvaluations(branchesAreNonOptional)
   requiredEvaluations(branchesAreNotIVCElements)
@@ -118,9 +123,32 @@ abstract class ChoiceTermBase(
 
   final protected lazy val optionChoiceDispatchKeyRaw = findPropertyOption("choiceDispatchKey")
   final protected lazy val choiceDispatchKeyRaw = requireProperty(optionChoiceDispatchKeyRaw)
+  
+  lazy val optionChoiceDispatchKeyKindRaw = findPropertyOption("choiceDispatchKeyKind")
+  lazy val defaultableChoiceDispatchKeyKind = 
+    if(tunable.requireChoiceDispatchKeyKindProperty){
+      choiceDispatchKeyKind
+    } else {
+      val asString = optionChoiceDispatchKeyKindRaw.toOption.getOrElse("implicit")
+      ChoiceKeyKindType(asString, this)
+    }
+  
+  lazy val optionChoiceBranchKeyKindRaw = findPropertyOption("choiceBranchKeyKind")
+  lazy val defaultableChoiceBranchKeyKind = 
+    if(tunable.requireChoiceBranchKeyKindProperty){
+      choiceDispatchKeyKind
+    } else {
+      val asString = optionChoiceBranchKeyKindRaw.toOption.getOrElse("implicit")
+      ChoiceKeyKindType(asString, this)
+    }
 
   final lazy val isDirectDispatch = {
-    val isDD = optionChoiceDispatchKeyRaw.isDefined
+    val isDD :Boolean = defaultableChoiceDispatchKeyKind match{
+      case ChoiceKeyKindType.ByType => true
+      case ChoiceKeyKindType.Explicit => true
+      case ChoiceKeyKindType.Implicit => optionChoiceDispatchKeyRaw.isDefined
+      case ChoiceKeyKindType.Speculative => false
+    }
     if (isDD && initiatedContent == YesNo.Yes) {
       SDE("dfdl:initiatedContent must not equal 'yes' when dfdl:choiceDispatchKey is defined")
     }
@@ -129,7 +157,7 @@ abstract class ChoiceTermBase(
 
   final protected lazy val choiceDispatchKeyExpr = {
     val qn = this.qNameForProperty("choiceDispatchKey")
-    ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, choiceDispatchKeyRaw, this)
+    ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, choiceDispatchKeyRaw, this, dpathCompileInfo)
   }
 
   final lazy val choiceDispatchKeyEv = {
@@ -138,6 +166,52 @@ abstract class ChoiceTermBase(
     ev.compile()
     ev
   }
+
+  // If choiceDispatchKeyKind is byType, verify that all our children share a repType,
+  // and use that. Otherwise, there is no need to associate a repType with this choice
+  override final lazy val optRepTypeFactory: Option[SimpleTypeDefFactory with NamedMixin] = defaultableChoiceDispatchKeyKind match {
+    case ChoiceKeyKindType.ByType => {
+      val branchReptypes: Seq[SimpleTypeDefFactory with NamedMixin] = groupMembers.map(term => {
+          term match{
+            case e: ElementDeclMixin => e.typeDef match{
+              case t: SimpleTypeDefBase => t.optRepTypeDefFactory match{
+                case None => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must have a type which defines a repType")
+                case Some(x) => x
+              }
+              case _ : SimpleTypeBase => SDE("When <xs:choice> has choiceBranchKey=\"byType\", no branch can have a primitive xsd type")
+              case _ => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must be a simple type")
+            }
+            case _ => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must be a simple element")
+          }
+      })
+      val ans = branchReptypes.reduce((a, b) => {
+        /*
+         * We tolerate type objects being copies or the same object because we compare the QNames.
+         */
+        if (a.namedQName != b.namedQName) {
+          SDE("All children of an <xs:choice> with choiceDispatchKeyKind=byType must have the same reptype")
+        }
+        a
+      })
+      Some(ans)
+    }
+    case ChoiceKeyKindType.Speculative => None
+    case ChoiceKeyKindType.Explicit    => None
+    case ChoiceKeyKindType.Implicit    => None
+  }
+
+  /*
+   * When choiceBranchKeyKind=byType or choiceDispatchKeyKind=byType, we are associated with one or more repTypes (corresponding to
+   * the reptypes of our choices).
+   * Since the repValueSet represents the set of repValues that we expect to be able to be associated with an element, it does, in theory,
+   * make sense to assign a repValueSet to an entire choiceGroup. For instance, in the common case of  choiceBranchKeyKind=byType and choiceDispatchKeyKind=byType,
+   * this would just be the union of the repValueSets of all of the individual choices.
+   * However, figuring this out becomes more complicated when we start having to deal with choiceBranchKeyKind=explicit
+   * There is no indication yet that this problem is particuarly intractable, however it is non-trivial. Since we never
+   * actually need to know the optRepValueSet of an entire choiceGroup, we can simple avoid thinking about it until such a time that we do.
+   */
+  
+  override final lazy val optRepValueSet = Assert.invariantFailed("We shouldn't need to compute the optRepValueSet of a choiceGroup")
 
   final override def hasKnownRequiredSyntax = LV('hasKnownRequiredSyntax) {
     if (hasInitiator || hasTerminator) true
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
@@ -55,7 +55,8 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
     namespaces: NamespaceBinding,
     compileInfoWherePropertyWasLocated: DPathCompileInfo,
     isEvaluatedAbove: Boolean,
-    host: OOLAGHost): CompiledExpression[T] = {
+    host: OOLAGHost,
+    compileInfo: DPathCompileInfo): CompiledExpression[T] = {
 
     val res =
       if (DPathUtil.isExpression(exprOrLiteral)) {
@@ -66,7 +67,8 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
           namespaces,
           compileInfoWherePropertyWasLocated,
           isEvaluatedAbove,
-          host)
+          host,
+          compileInfo)
       } else {
         convertLiteralToConstant(
           qn,
@@ -99,6 +101,7 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
     nodeInfoKind: NodeInfo.Kind,
     property: Found,
     host: OOLAGHost,
+    compileInfo: DPathCompileInfo,
     isEvaluatedAbove: Boolean = false): CompiledExpression[T] = {
 
     compileExpression(
@@ -106,9 +109,10 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
       nodeInfoKind,
       property.value,
       property.location.namespaces,
-      propertyCompileInfo(property),
+      compileInfo,
       isEvaluatedAbove,
-      host)
+      host,
+      compileInfo)
   }
 
   /**
@@ -135,7 +139,8 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
     staticNodeInfoKind: NodeInfo.Kind,
     runtimeNodeInfoKind: NodeInfo.Kind,
     property: Found,
-    host: OOLAGHost): CompiledExpression[T] = {
+    host: OOLAGHost,
+    compileInfo: DPathCompileInfo): CompiledExpression[T] = {
 
     val isEvaluatedAbove = false
     val exprOrLiteral = property.value
@@ -146,9 +151,10 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
       staticNodeInfoKind,
       exprOrLiteral,
       namespacesForNamespaceResolution,
-      compileInfoWherePropertyWasLocated,
+      compileInfo,
       isEvaluatedAbove,
-      host)
+      host,
+      compileInfo)
 
     if (compiled1.isConstant || (staticNodeInfoKind == runtimeNodeInfoKind)) {
       compiled1
@@ -158,9 +164,10 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
         runtimeNodeInfoKind,
         exprOrLiteral,
         namespacesForNamespaceResolution,
-        compileInfoWherePropertyWasLocated,
+        compileInfo,
         isEvaluatedAbove,
-        host)
+        host,
+        compileInfo)
       compiled2
     }
   }
@@ -190,7 +197,8 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
     namespaces: NamespaceBinding,
     compileInfoWherePropertyWasLocated: DPathCompileInfo,
     isEvaluatedAbove: Boolean,
-    host: OOLAGHost): CompiledExpression[T] = {
+    host: OOLAGHost,
+    compileInfo: DPathCompileInfo): CompiledExpression[T] = {
 
     // Treat this as an expression--validate and compile it
 
@@ -204,7 +212,7 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
       qn,
       nodeInfoKind,
       namespaces,
-      compileInfoWherePropertyWasLocated,
+      compileInfo,
       isEvaluatedAbove,
       host)
     val compiledDPath = compiler.compile(expr)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/LocalElementDecl.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/LocalElementDecl.scala
@@ -32,7 +32,7 @@ sealed class LocalElementDecl(
 }
 
 /**
- * A DetachedElement is similar to a LocalElement except it will have no
+ * A QuasiElement is similar to a LocalElement except it will have no
  * representation in the infoset, acting only as a temporary element that can
  * be parsed/unparsed. As an example, this is used as an element foar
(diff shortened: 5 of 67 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -259,26 +591,91 @@ final class GlobalSimpleTypeDefFactory(xmlArg: Node, schemaDocumentArg: SchemaDo
...
+
+  override protected val optReferredToComponent = None
+
+  protected def annotationFactory(node: Node): Option[DFDLAnnotation] = Assert.invariantFailed("Should not be called")
+  protected def emptyFormatFactory: DFDLFormatAnnotation = new DFDLEnumerationFactory(newDFDLAnnotationXML("enumeration"), this)
+  protected def isMyFormatAnnotation(a: DFDLAnnotation): Boolean = Assert.invariantFailed("Should not be called")
+
+}
+
+final class EnumerationDef(
```
mbeckerle: Looks like you likely dont need EnumerationDef. All the functionality is on the Factory (where it belongs !)

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -259,26 +591,91 @@ final class GlobalSimpleTypeDefFactory(xmlArg: Node, schemaDocumentArg: SchemaDo
...
+  parent:  SimpleTypeDefBase,
+  factory: EnumerationDefFactory)
+  extends AnnotatedSchemaComponentImpl(factory.xml, parent)
+  with Enumeration_AnnotationMixin
+  with ResolvesProperties {
+
+  // Members declared in org.apache.daffodil.dsom.AnnotatedMixin
+  protected def annotationFactory(node: Node): Option[DFDLAnnotation] = Assert.invariantFailed("Should not be called")
```
mbeckerle: What are the implications of not having annotationFactory and isMyFormatAnnotation? This means, I think, that you can't have actual DFDL properties here that are found by way of the normal scoping rules. That may be fine for these enumerations, as the properties may not make sense to be scoped anyway.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -70,12 +119,51 @@ case class ChoiceCombinator(ch: ChoiceTermBase, rawAlternatives: Seq[Gram])
...
+              val conflictingRanges = dispatchBranchKeyRangeTuples.filter({ case (min, max, _, _) => min.testAsLower(asBigInt) && max.testAsUpper(asBigInt) })
+              if (conflictingRanges.length > 0) {
+                SDE(
+                  "dfdl:choiceBranchKey (%s) conflicts with dfdl:choiceBranchKeyRanges. Offending branches are:\n%s\n%s",
+                  k,
+                  kvs.map(_._2.context.runtimeData).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""),
+                  conflictingRanges.map(_._3.context).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""))
+              }
+            }
+            case Failure(_) => ()
```
mbeckerle: So if k.toLong fails, you don't throw the error/capture it as an SDE? Why is it ok to suppress this here?

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
```diff
@@ -72,6 +72,20 @@ case class DaffodilTunables(
   //
   val requireBitOrderProperty: Boolean = false,
   //
+  // If true, require that the choiceDispatchKeyKind property is specified. If false, use a
```
mbeckerle: Use scaladoc syntax to document these members.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
```diff
@@ -93,9 +93,23 @@ trait Cursor[AccessorType <: Accessor[AccessorType]] {
...
+   * must be inserted into the infoset. Therefore, in a chain with one or more calls to inspect,
+   * followed by a call to advance, the side effect of constructing the DINode will be realized on the
+   * first call to inspect.
+   *
+   * For a side-effect free variant, consider inspectPure
    */
   def inspect: Boolean
 
+  /**
+   * Like inspect, but only partially populates the accessor with data that can be computed without
```
mbeckerle: What does "partially populates" mean? I surised that you are trying to get a node that is unattached from the infoset. Is that right? Can you make this a bit clearer in intent. 

Consider whether inspectPure is a good name. Maybe it should be more explicit beause what "pure" means in this context isn't 100% clear?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
```diff
@@ -160,18 +196,24 @@ trait CursorImplMixin[AccessorType <: Accessor[AccessorType]] { self: Cursor[Acc
...
     // successful
     Assert.invariant(isFilled == true || !res)
     res
   }
 
-  private def doAdvance(isFilledValue: Boolean) = {
-    val res = fill()
+  //inpsectPure implemented in InfosetInputter.scala
+
+  private def doAdvance(isFilledValue: Boolean, advanceInput: Boolean) = {
```
mbeckerle: Having a method doAdvance with a flag argument called advanceInput is pretty confusing.  Can you provide a comment to clarify what side effects doAdvance does that aren't suppressed by advanceInput false?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -99,3 +108,42 @@ class NewVariableInstanceEndUnparser(override val context: RuntimeData)
     context.notYetImplemented("newVariableInstance")
   }
 }
+
+// typeCalculator, repTypeUnparser, e.elementRuntimeData, repTypeRuntimeData
```
mbeckerle: remove comment line?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DState.scala
```diff
@@ -95,13 +95,24 @@ case object UnparserNonBlocking extends EvalMode
...
+case class DState(val maybeSsrd:Maybe[SchemaSetRuntimeData]) {
   import org.apache.daffodil.util.Numbers._
 
   var isCompile = false
 
   var opIndex: Int = 0
 
+  /*
+   * Used by TypeValueCalc to pass the logical or representation value to the DPath runtime.
+   * In principle, TypeValueCalc can be used recursivly, so these data structures are realy a stack.
```
mbeckerle: spelling: recursively, really

daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
```diff
@@ -38,7 +38,7 @@ import org.apache.daffodil.api.WarnID
  * property resolution.
  *
  * The only objects that should resolve properties are
- * ElementRef, Root, LocalElementDecl, Sequence, Choice, SequenceRef, ChoiceRef
+ * ElementRef, Root, LocalElementDecl, Sequence, Choice, SequenceRef, ChoiceRef, Enumeration
```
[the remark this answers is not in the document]
bsloane1650: Will fix in the next commit. The above comment is already out of date anyway. The oddball objects are now EnumerationFactory and SimpleTypeDefFactory (for essentially the same reason that Enumeration was originally)

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
bsloane1650: You appear to be correct, EnumerationDef is never actually used, so I will get rid of it. I believe this means my refactoring of EnumerationDef amounts to adding "Factory" to its class name.

This is a bit weird, as it means that we have a "Factory" class that is not actually a factory, but in the context of DSOM, I think the name EnumerationDefFactory still makes sense, as "Factory" here really means "schema component considered in isolation of its context". Indeed even the SimpleTypeDed/Restriction/Union "factories" are more than just factories, since they do a fair amount of stuff unrelated to generating the "concrete" object (and, in some cases, will never be even be called for their factory method).

I'll add a comment explaining this above EnumerationDefFactory, as I suspect that is the place that will actually make people scratch their heads.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
bsloane1650: I'm not actually sure. I assume we thought this through back when this was added. Regardless, most of the annotation logic is not relevant to Enumerations as the only meaningful place to put annotations is on the object itself.

In fact, because these annotations need to be accessed from the factory object, we cannot use most of the annotation logic, which is probably what led us to dissallow  these methods in the first place.

The only potential problem that I can imagine is that these methods are needed for long-form annotations. To get language lawery, I do not believe Enumerations are defined as "annotation points", so a longform annotation might not be technically allowed there anyway (although now that we can annotate them, they ought to be annotation points). Regardless, I can't imagine someone wanting to do a longform annotation on them.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
bsloane1650: I'll add a comment explaining this.

Essentially, the only point of this try block is to detect when there is an overlap between a single choiceBranchKey and a choiceBranchKeyRange. If the single choiceBranchKey is not numeric, it cannot be part of a choiceBranchKeyRange; and, at the moment, the only numeric type we support for keys is integers up to 64 bits.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
bsloane1650: I appear to be consistent with the existing convention for these files. Should I update all the comments?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
bsloane1650: To be honest, I am not particularly happy with anything about Cursor. Most of my changes feel like a very hacky way to work through a leaky abstraction without any significant refactoring (which I am not comfortable enough with this area of the codebase to do confidently).

The core issue here is that the Cursor trait in daffodil-lib/.../util/ is really an abstract type that is completely independent from any notion of "infoset". Indeed, the daffodil-lib project is not even aware of what an Infoset is, as it does not depend on the relevent project (and can't without introducing a circular dependency). As it turns out, the only user of Curosr is InfosetCursor.

The motiviation for "inspectPure" is that the implementation of "inspect" by InfosetCursor is broken with regards to the standard meaning of "inspect" (which is normally thought of as a side-effect free method), and it is impossible to write a non-broken implementation in a backwards compatible way.

As it turns out, inspectPure isn't actually side effect free either (as it consumes input from the underlying Inputter), but it should be free of externally visible side effects.

Ultimately, the issue is that we are defining a really wonky API here. Understanding the implications of inspect vs inspectPure really does require looking at the actual implementing class to what what what side-effects inspect has.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
bsloane1650: Again, we have a really wonky API here. I'll add a comment in this part of the code explaining, but for the sake of discussion I'll include an explanation here to.

Logically speaking, there are two "streams" that we need to keep in our mind when thinking about an InfosetCursor.

There is the input stream, which contains a series of infoset events, and the output stream, which is a sequence of infoset nodes.

The first time you observe an infoset node, you consume it from the input stream. However, it is not consumed from the output stream until the "last" time you observe it (eg. you observe it through calling advance()). The advanceInput flag is used to indicate if the call to advance() is also the first time the node is being observed, in which case it also needs to consume the relevent portion of the input stream.

Arguably, this logic really belongs in the priorOpKind state machine we have sprinkled throughout the advance/inspect/inspectPure methods.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
mbeckerle: They were done wrong by prior author (which could be me.... but it got past code review at that time). 

Might as well fix them all.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
mbeckerle: The point of cursor is that unlike hasNext() and next() style APIs, one can advance cursors without any little crud objects being allocated. You advance the cursor and an accessor is modified to provide access to what is now "under" the cursor. This is ordinary APIs for big-data, e.g., database cursors. Inspect gives a 1 object look-ahead capability.  What does inspectPure add to this? Does it NOT provide access to the data being inspected? Is it just a test for whether there is or isn't a next?

bsloane1650: The issue is that in the InfosetCursor implementation, constructing the object has side effects. So, in order to use the 1 object lookahead, the side effects associated with creating an object occur when you use the lookahaed, instead of when you actually consume the object. This means that there is no way to inspect any aspect of the next object without triggering the side effects associated with creating it.

InspectPure provides a limited 1 object look ahead that can be used without triggering the side effects associated with constructing the full object like inspect would.

In more practical terms, inspectPure provides only the ElementRuntimeData, while inspect provides a DINode.

This is actually orthogonal to the difference between more traditional iterators. We would have the same problem with a next()/hasNext()/peek() style API, as our implementation of peek() would have the same side effects as inspect() does now.

The API of Cursor itself is actuall pretty sane. The problem comes when we make InfosetCursor use Accessor[DINode], as that requires making inspect()/next() have unexpected side-effects.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
bsloane1650: What is the appropriate usage of scaladoc here?

Since these are all class parameters, it would seem that the correct style would be to move all of the documentation to above the class with a bunch of @param annotations, and not have any comment where they are declared. Is this desired?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/LocalElementDecl.scala
```diff
@@ -40,9 +40,26 @@ sealed class LocalElementDecl(
...
  */
-final class DetachedElementDecl(
+sealed abstract class QuasiElementDeclBase(
   val detachedReference: ElementBase,
   xml: Node,
   parent: SchemaComponent)
   extends LocalElementDecl(xml, parent, -1){
 }
+
+//TODO rename to PrefixLengthQuasiElementDecl
```
jadams-tresys: Address this TODO before merging

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ReptypeMixins.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.xml.Null
+import scala.xml.Elem
+import org.apache.daffodil.processors.RepValueSet
+
+/*
+ * To make use of a reptype, it is common for the parser/unparser to need an element associated with said type
+ * Since such an element does not nessasarily exist in the schema itself, those elements which have a reptype also
+ * need to get a quasi element with said type.
+ */
+trait hasOptRepTypeMixinImpl extends SchemaComponent with hasOptRepTypeMixin{
```
jadams-tresys: Missing space before {

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ReptypeMixins.scala
```diff
@@ -0,0 +1,50 @@
...
+    val xmlElem = Elem(null, "QuasiElementForTypeCalc", new UnprefixedAttribute("type", repType.namedQName.toAttributeNameString, Null), namespaces, true)
+    new RepTypeQuasiElementDecl(this.enclosingElement.get, xmlElem, parent)
+  })
+
+}
+
+trait hasOptRepTypeMixin {
+  def optRepTypeFactory: Option[SimpleTypeFactory with NamedMixin]
+  
+  lazy val optRepTypeDefFactory: Option[SimpleTypeDefFactory with NamedMixin] = optRepTypeFactory match{
```
jadams-tresys: Here as well

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -166,16 +175,126 @@ final class Restriction(xmlArg: Node, val simpleType: SimpleTypeDefBase)
...
+    } else {
+      //We are some form of identity mapping, so our facetValues are also our repValues
+      facetValueSet
+    }
+  }
+  lazy val optLogicalValueSet:Option[RepValueSet[AnyRef]] = if(logicalValueSet.isEmpty) None else Some(logicalValueSet)
+
+  lazy val optRepType:Option[SimpleTypeBase] = {
+    enumerations.map(_.optRepTypeFactory)
+    ???
```
jadams-tresys: This should be addressed before being merged.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -40,14 +69,71 @@ trait NonPrimTypeMixin {
...
+        .map(_.split("\\s+").toSeq).getOrElse(Seq())
+      val repValueRangesRaw = findPropertyOptionThisComponentOnly("repValueRanges").toOption
+        .map(_.split("\\s+").toSeq).getOrElse(Seq())
+      repType.primType match {
+        case PrimType.String => {
+          if (repValueRangesRaw.size > 0) context.SDE("repValueRanges set when using a string repType")
+          val repValueSetCooked = repValueRangesRaw.map(RepValueCooker.convertConstant(_, context, false))
+          (repValueSetCooked, Seq())
+        }
+        //I need to double check the type hierarchy, but I think all integer types are subtypes of SignedInteger
```
jadams-tresys: That appears to be correct, so I would go ahead and remove this comment.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -221,64 +311,399 @@ final class LocalSimpleTypeDef(
...
+        repTypes.head
+      }
+    })
+    fromSelf.orElse(fromUnion)
+  }
+
+  override lazy val optRepValueSet: Option[RepValueSet[AnyRef]] = optRepTypeDefFactory.flatMap(repType => {
+    val primType: PrimType = repType.primType
+
+    //TODO check for SDEs
```
jadams-tresys: Deal with these TODO's

jadams-tresys:
Sorry, I missed your statement about not being done yet, so ignore my comments about dealing with TODO's, etc.

I haven't gotten through everything yet, but I've only noticed some small formatting things, so things are looking good so far.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
bsloane1650: As far as I can tell, these TODOs have already been addressed, and I just forgot to remove the comments. Removing in next commit.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
bsloane1650: There's a conversation regarding this in the dev mailing list. The TLDR is that SignedInteger isn't really being used, shouldn't exist, and has an inconsistent place in the XSD type hierarchy. 
But, I agree that this comment is correct from the perspective of the scala type system, so I'll remove it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
bsloane1650: Not sure what the original intent of this field was, so I'm just going to remove it.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
bsloane1650: Per Mike's email to me, these are class parameters, not normal fields. Since these are not part of an external API, it is not worth the added maintance effort to maintain proper scala docs, so we are leaving them as is.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1543,6 +1543,49 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
         DFDLSetBitsExpr(functionQNameString, functionQName, args)
       }
 
+      //Begin TypeValueCalc related functions
+      case (RefQName(_, "inputTypeCalcInt", DFDL), args) =>
```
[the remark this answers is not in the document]
bsloane1650: These should be DFDLX, but I believe we decided to put off creating the DFDLX namespace to later.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXTypeCalcFunctions.scala
```diff
@@ -0,0 +1,239 @@
...
+import org.apache.daffodil.dpath.NodeInfo.PrimType.String
+import org.apache.daffodil.dpath.NodeInfo.PrimType.Integer
+
+trait typeCalculatorNamedDispatch { self: FNTwoArgs =>
+  var srcType: NodeInfo.Kind = null
+  var dstType: NodeInfo.Kind = null
+
+  override def validateType(srcTypes: Seq[NodeInfo.Kind], dstType: NodeInfo.Kind): Unit = {
+    //We can't actually verify anything yet, because we need to wait until run() is
+    //called to know what calculator we are using
```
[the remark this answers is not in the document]
bsloane1650: I agree with the naming issue. The original idea was just that I was deferring type-checking to expression nodes themselves, which is a layer of abstraction lower then we normally do, so this function was providing a back door to sneak the type information into the nodes without tearing up too much existing code. If there is a pre-existing way of getting this type information at the layer I need it, I can remove this function entirely

The reason I ended up deferring to runtime is that we could, in principle, have expressions like:

{ dfdl:inputTypeCalcInt(../functionName, ../functionArg) }

where the type calculator to be used is determined by the infoset itself. Mike had suggested just not allowing this. If we insist the function name be a constant, we should be able to move the validation back to compile time (along with a handleful of other missing/invalid typeCalculator errors).

Even if someone comes along with a reason to need the dynamisism, it might still be worth it to specialize the constant case to provide SDEs at compile time.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1543,6 +1543,49 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+
+      case (RefQName(_, "logicalTypeValueInt", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.Integer, NodeInfo.AnyAtomic, DFDLXLogicalTypeValueInt(_, _))
+          
+      case (RefQName(_, "logicalTypeValueString", DFDL), args) =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.String, NodeInfo.AnyAtomic, DFDLXLogicalTypeValueString(_, _))
+
+      //End typeValueCalc related functions
```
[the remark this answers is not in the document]
mbeckerle: Note: I decided to do this the other way round: implement, see whether we like the way it came out, and only once we've revised it and are satisfied with it, do we propose it for inclusion in DFDL.

This is, in fact, the right way for a standard to be driven by an implementation. Standards should be based on experience with an implementation. So in the future I expect to do things this way.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXTypeCalcFunctions.scala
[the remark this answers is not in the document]
stevedlawrence: In DFDL XPath, everything is strongly typed, so at schema compile time you should know what the type of functionName and functionArg are, regardless if they are variables or constants. My gut feeling is that in Expression.scala where you have things like this:
```scala
      case (RefQName(_, "inputTypeCalcInt", DFDL), args) =>
        FNTwoArgsExpr(functionQNameString, functionQName, args,
          NodeInfo.Int, NodeInfo.String, NodeInfo.AnyAtomic, DFDLXInputTypeCalcInt(_))
```
You could replace FNTwoArgsExpr with a custom Expression, perhaps something that ineherits FNTwoArgsExpr and extends it to provide extra type information when it calls the DFDLXInputTypeCalcInt recipe constrcutor.

In fact, it seems like the whole getCalculator stuff could get determined at compile time. We should know the type of the arguments (because strongly typed) so the calculator should be known at compile time? I think...

daffodil-core/src/main/scala/org/apache/daffodil/dsom/QuasiElementDecl.scala
```diff
@@ -40,9 +40,25 @@ sealed class LocalElementDecl(
  * DetachedElementDecl is used as a place where properties related to the
```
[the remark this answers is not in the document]
bsloane1650: LocalElementDecl,scala is probably the better filename. The QuasiElementDecl stuff is really just a slightly specialized case of LocalElementDecl that is used far less frequently.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/QuasiElementDecl.scala
```diff
@@ -40,9 +40,25 @@ sealed class LocalElementDecl(
...
+  extends QuasiElementDeclBase(detachedReference, xml, parent){
+}
+
+final class RepTypeQuasiElementDecl(
+  detachedReference: ElementBase,
+  xml: Node,
+  parent: SchemaComponent)
+  extends QuasiElementDeclBase(detachedReference, xml, parent){
+  
+  override lazy val isQuasiElement = true
```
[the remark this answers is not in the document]
bsloane1650: Probably just an artifact of how this code was written. There was a period where (what became) PrefixLengthQuasiElementDecl and RepTypeQuasiElementDecl didn't ave a common QuasiElementDeclBase supertype. Moving this override to Base doesn't seem to break anything, so I assume that is what was meant.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -0,0 +1,453 @@
...
+    dstate.setCurrentValue(ans)
+
+  }
+
+  def supportsParse: Boolean = true
+  def supportsUnparse: Boolean = true
+}
+
+class KeysetValueTypeCalculatorOrdered[A <: AnyRef, B <: AnyRef](valueMap: HashMap[A, B], rangeTable: Seq[(RangeBound[A], RangeBound[A], B)], unparseMap: HashMap[B, A],
+                                                                 srcType: NodeInfo.Kind, dstType: NodeInfo.Kind)
```
stevedlawrence: Check the indendation for things like this. These long lines should be something like
```
class Foo(
  var1: Type,
  var2: Type,
  ...
)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -0,0 +1,453 @@
...
+  extends TypeCalculator[A, B](srcType, dstType) {
+
+  override def inputTypeCalc(x: A, xType: NodeInfo.Kind): (Maybe[B], Maybe[Error]) = {
```
stevedlawrence: It this is going to allocate a Tuple, where _1 is success and _2 is an error, we should just use an Either. Allocation is going to happen either way, so we don't gain anything by using two maybes.

Though, I'm not even sure the error message return is necessary. The error message only contains the value of ``x` and whenther it was "key" or "value"`, which the all should be able to know. I would argue this should just return a a Maybe[B], and the caller can figure out the appropriate error message if Maybe.Nope is returned.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -0,0 +1,453 @@
...
+class ExpressionTypeCalculator[A <: AnyRef, B <: AnyRef](maybeInputTypeCalc: Maybe[CompiledExpression[B]], maybeOutputTypeCalc: Maybe[CompiledExpression[A]],
+                                                         srcType: NodeInfo.Kind, dstType: NodeInfo.Kind)
+  extends TypeCalculator[A, B](srcType, dstType) {
+  override def supportsParse = maybeInputTypeCalc.isDefined
+  override def supportsUnparse = maybeOutputTypeCalc.isDefined
+
+  //The class TypeValueCalc will verify that supports(Un)Parse is true when nessasary
+  //Therefore, if we ever call the below functions, we know that the relevent Maybe object is defined.
+
+  //TODO, pass x into DPath state
```
stevedlawrence: Can this TODO be removed?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -114,42 +116,138 @@ class DynamicEscapeSchemeParser(
...
+        val key = maybeKey.get
+
+        val parserOpt1 = dispatchBranchKeyMap.get(key)
+        val parserOpt2 = 
+          if(parserOpt1.isDefined) {
+            parserOpt1
+          } else{
+            if(!dispatchKeyRangeMap.isEmpty){
+              val keyAsBigInt = BigInt(key)
+              val optAns1= dispatchKeyRangeMap.filter({case(min,max,_,_) => min.testAsLower(keyAsBigInt) && max.testAsUpper(keyAsBigInt)}).headOption
```
stevedlawrence: If we're only going to get headOption, then it probably makes sense to use ``find`` instead of ``filter``. Otherwise we have to loop through the entire vector instead of just stopping at the first one that matches.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -67,6 +83,82 @@ class IVCParser(expr: CompiledExpression[AnyRef], e: ElementRuntimeData)
...
 }
 
+/*
+ * Run a parser for an element that does not occur in the infoset
+ * Prior to running, a temporary element (of the type expected by the parser) will be created in the infoset,
+ * After running, the infoset will be reverted to its original state, but any other side effect of parsing will remain
+ *
+ * Additionally, the dataValue of the element the parser parsed will be returned
+ */
+trait withDetachedParser {
```
stevedlawrence: Captial trait name. Also, should this be WithQuasiParser? Detached was changed to Quasi in the grammar, should probably be changed here for consistency.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -67,6 +83,82 @@ class IVCParser(expr: CompiledExpression[AnyRef], e: ElementRuntimeData)
...
+     *  5) pstate.infoset.restore
+     *  6) pstate.infoset = priorElement
+     *
+     *  Note that we are only restoring the infoset. Any other side effects the repTypeParser has on pstate
+     *  (such as advancing the bit posistion) will remain.
+     *
+     *  If repTypeParser has an error (either thrown or status), we percolate it up to our caller.
+     */
+
+    Processor.initialize(distachedParser)
```
stevedlawrence: I believe this initialize call should happen at schema compile time. Any Parsers that implement this trait need to add distachedParser to their list of child parsers, which should initialize them.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -67,6 +83,82 @@ class IVCParser(expr: CompiledExpression[AnyRef], e: ElementRuntimeData)
...
 
+/*
+ * Run a parser for an element that does not occur in the infoset
+ * Prior to running, a temporary element (of the type expected by the parser) will be created in the infoset,
+ * After running, the infoset will be reverted to its original state, but any other side effect of parsing will remain
+ *
+ * Additionally, the dataValue of the element the parser parsed will be returned
+ */
+trait withDetachedParser {
+  def runDetachedParser(pstate: PState, distachedParser: Parser, erd: ElementRuntimeData): Maybe[AnyRef] = {
```
stevedlawrence: detachedParser instad of distachedParser?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -67,6 +83,82 @@ class IVCParser(expr: CompiledExpression[AnyRef], e: ElementRuntimeData)
...
+     *
+     *  If repTypeParser has an error (either thrown or status), we percolate it up to our caller.
+     */
+
+    Processor.initialize(distachedParser)
+
+    val priorElement = pstate.infoset
+    pstate.setParent(pstate.infoset.diParent)
+
+    val priorState = pstate.mark("TypeValueCalcParser.parse: About to parse repTypeQuasiElement")
```
stevedlawrence: This mark name is too specific. I assume this was created before being split into a more generic trait?

eclipse-projects/lib-test/.classpath
```diff
@@ -1,92 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-        <!-- This file is updated by the UpdateEclipseClasspath app. -->
```
stevedlawrence: This changes are in PR #213, which questions if we even want these .classpath files in version control since they are autogenerated. We may want to remove these.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -52,6 +52,11 @@ sealed abstract class PropertyLookupResult(val pname: String) extends Serializab
    * for the schema file.
    */
   def isFromDefaultFormat: Boolean
+  
+  lazy val toOption = this match{
```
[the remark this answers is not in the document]
mbeckerle: If a def, then every time toOption is called this will box a new Some(x) object. If this is immutable, then lazy val seems the right choice.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
```diff
@@ -55,7 +55,8 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
     namespaces: NamespaceBinding,
     compileInfoWherePropertyWasLocated: DPathCompileInfo,
     isEvaluatedAbove: Boolean,
-    host: OOLAGHost): CompiledExpression[T] = {
+    host: OOLAGHost,
+    compileInfo: DPathCompileInfo): CompiledExpression[T] = {
```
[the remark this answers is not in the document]
mbeckerle: As I recall,.... compileInfo is for the component on which we are doing the compilation. When compiling a property-expression however, that property may not have been expressed directly on the component, but elsewhere. E.g., dfdl:byteOrder="{ $dfdl:byteOrder }" appearing at top level. We need the compile info to tell us where the property was expressed for diagnostic purposes.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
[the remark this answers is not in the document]
stevedlawrence: But it looks like we only ever call toOption once for each property (addedin this commit). So I think we'd be storing something we never acess again, which will just waste memory. In this case, it's such a small amount I'm not sure it really matters, so I don't feel strongly about this.

I kindof viewed this similar to something like Seq.headOption, which is going to allocatea new one everytime it's called. That might be bad, but what's the overhead of boxing vs using memory that can never be garbage collected?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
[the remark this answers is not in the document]
stevedlawrence: We should probably change the name then, or at least add some documentation clarifiing the difference. compileInfoWherePropertyWasLocated and compiledInfoWherePropertyWasExpressed?

So in your example, compileInfoWherePropertyWasLocated would contain an actual element, but compileInfoWherePropertyWasExpressed might be on the dfdl:format tag or something?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -166,16 +175,121 @@ final class Restriction(xmlArg: Node, val simpleType: SimpleTypeDefBase)
...
+    } else {
+      //We are some form of identity mapping, so our facetValues are also our repValues
+      facetValueSet
+    }
+  }
+  lazy val optRepValueSet = if (repValueSet.isEmpty) None else Some(repValueSet)
+
+  lazy val logicalValueSet: RepValueSet[AnyRef] = {
+    val subsets = enumerations.map(_.logicalValueSet)
+    if (subsets.length != 0 && subsets.length != enumerations.length) {
```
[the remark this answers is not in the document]
bsloane1650: I'm not entirely sure what I was getting at here. The logicalValueSet shouldn't care about weather or not the enumerations define a repValue. Regardless, If I was trying to check for something here, I clearly am not succeeding, so might as well remove the spurious check.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
bsloane1650: We changed DetachedElement -> QuasiElement in the grammar. The idea being that the element in question is not a "real" element, but just exists to satisfy the requirements of certain existing code.

In contrast, a detachedParser strikes me as a "real" parser, so the quasi adjective does not make sense to me here.

stevedlawrence: Fair, makes sense to me.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ChoiceGroup.scala
```diff
@@ -139,6 +167,52 @@ abstract class ChoiceTermBase(
...
+          term match{
+            case e: ElementDeclMixin => e.typeDef match{
+              case t: SimpleTypeDefBase => t.optRepTypeDefFactory match{
+                case None => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must have a type which defines a repType")
+                case Some(x) => x
+              }
+              case _ : SimpleTypeBase => SDE("When <xs:choice> has choiceBranchKey=\"byType\", no branch can have a primitive xsd type")
+              case _ => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must be a simple type")
+            }
+            case _ => SDE("When <xs:choice> has choiceBranchKey=\"byType\", all branches must be a simple element")
```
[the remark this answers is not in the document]
bsloane1650: You appear to be correct. I added a test with:

```
              <xs:choice dfdl:choiceBranchKeyKind="byType" dfdl:choiceDispatchKeyKind="byType" >
                <xs:element ref="tns:one" />
                <xs:element ref="tns:two_through_100" />
              </xs:choice>
```

Which catches this

stevedlawrence:
Didn't realize I never gave a +1. I think all my comments have been addressed. This is a much needed addition.

+1 :+1:

PR #208, opened 2019-04-16
Output empty elements with self closing tags

jadams-tresys:
Before empty comlex elements would have to be represented with
<elem></elem>, which could cause some problems with non-schema aware XML
parsers. Now empty complex elements will show up as <elem /> in the
infoset.

DAFFODIL-2064

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
@@ -41,7 +41,7 @@ class XMLTextInfosetOutputter(writer: java.io.Writer, pretty: Boolean = true)
     if (prefix == null || prefix == "") elem.erd.name else prefix + ":" + elem.erd.name
   }
 
-  private def outputStartTag(elem: DIElement, name: String): Unit = {
+  private def outputStartTag(elem: DIElement, name: String, isEmpty: Boolean=false): Unit = {
     writer.write("<")
 
     writer.write(name)
@@ -58,7 +58,11 @@ class XMLTextInfosetOutputter(writer: java.io.Writer, pretty: Boolean = true)
       writer.write(" xsi:nil=\"true\"")
     }
 
-    writer.write(">")
+    if (isEmpty) {
+      writer.write("/>")
+    } else {
+      writer.write(">")
+    }
   }
 
   private def outputEndTag(elem: DIElement, name: String): Unit = {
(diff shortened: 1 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
```diff
@@ -97,19 +101,24 @@ class XMLTextInfosetOutputter(writer: java.io.Writer, pretty: Boolean = true)
   override def startComplex(complex: DIComplex): Boolean = {
     val name = getTagName(complex)
     if (pretty) outputIndentation(writer)
-    outputStartTag(complex, name)
+    outputStartTag(complex, name, complex.children.isEmpty)
```
stevedlawrence: I not sure if looking at ``complex.children.isEmpty`` is completly sufficient here. For exampe, if all the children of a complex element are from a hidden group, then ``complex.children`` won't be empty, but the resulting infoset won't have any childen.

So instead of this, we might need something like a new function in DIComplex like``def hasVisibleChildren`` (I think there's a DFDL term for "visible" in this case) that iterates through the childNodes and determines if 1) there are no children or 2) all children are hidden.

Or alterantively, since this information is needed twice (once in startComplex and again in endComplex), it might be more performatn to create a new boolean var in the DIComplex and toggle it once a non-hidden child is added?

Might be other options, but I think this does need to take into acount hidden elements.

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequenceGroup.tdml
```diff
@@ -68,7 +68,7 @@
     </tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
-        <complexEmptyContent></complexEmptyContent>
+        <complexEmptyContent/>
```
stevedlawrence: Does the TDML runner view these as not equivlent? If so we should open a bug.It shouldn't matter if the TDML infoset contains ``<foo></foo>`` or ``<foo/>``.

jadams-tresys: It does see them as equivalent, but I just wanted to use one test to make sure that it was outputting stuff correctly and had modified this test.  I can revert it back as it doesn't really make a difference.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
```diff
@@ -41,7 +41,7 @@ class XMLTextInfosetOutputter(writer: java.io.Writer, pretty: Boolean = true)
     if (prefix == null || prefix == "") elem.erd.name else prefix + ":" + elem.erd.name
   }
 
-  private def outputStartTag(elem: DIElement, name: String): Unit = {
+  private def outputStartTag(elem: DIElement, name: String, empty: Boolean=false): Unit = {
```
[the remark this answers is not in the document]
jadams-tresys: Will do

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
jadams-tresys: I've updated this with the suggested hasVisibleChildren as a boolean that gets set to true as soon as a visible child is added to the complex element

PR #210, opened 2019-04-22
Add midBitsToEnd layer transform

bsloane1650:
Add new layer transform that can move a single contiguous block
of bits from a layer's content area to the end of the layer's
content area.

DAFFODIL-2109

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
@@ -658,7 +658,7 @@ trait LayeringRuntimeValuedPropertiesMixin
   }
   
   final lazy val maybeLayerTransformArgsEv = {
-    if(optionLayerTransformArgsRaw.isDefined){
+    if (optionLayerTransformArgsRaw.isDefined) {
       val qn = this.qNameForProperty("layerTransformArgs")
       val expr = ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, layerTransformArgsRaw, decl)
       val ev = new LayerTransformArgsEv(expr, termRuntimeData)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
@@ -72,6 +72,7 @@ import org.apache.daffodil.processors.LayerCharsetEv
 import org.apache.daffodil.schema.annotation.props.TextStandardExponentRepMixin
 import org.apache.daffodil.schema.annotation.props.PropertyMixin
 import org.apache.daffodil.schema.annotation.props.Found
+import org.apache.daffodil.processors.LayerTransformArgsEv
 
 /*
  * These are the DFDL properties which can have their values come
@@ -649,6 +650,18 @@ trait LayeringRuntimeValuedPropertiesMixin
   final lazy val maybeLayerTransformEv = {
     if (optionLayerTransformRaw.isDefined) {
       val ev = new LayerTransformEv(layerTransformExpr, termRuntimeData)
+      ev.compile()
+      One(ev)
+    } else {
+      Nope
+    }
+  }
+  
+  final lazy val maybeLayerTransformArgsEv = {
+    if(optionLayerTransformArgsRaw.isDefined){
+      val qn = this.qNameForProperty("layerTransformArgs")
+      val expr = ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, layerTransformArgsRaw, decl)
+      val ev = new LayerTransformArgsEv(expr, termRuntimeData)
       ev.compile()
       One(ev)
     } else {
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
@@ -794,6 +794,17 @@
            </xsd:documentation>
          </xsd:annotation>
        </xsd:enumeration>
+       <xsd:enumeration value="midBitsToEnd">
+         <xsd:annotation>
+           <xsd:documentation><![CDATA[
+             Move the bits with indecis [x, x+y] to
+             the end of the layer block.
+
+             x,y are provided as a space delimated list in layerTransformArgs
+             ]]>
+           </xsd:documentation>
+         </xsd:annotation>
+       </xsd:enumeration>
     </xsd:restriction>
   </xsd:simpleType>
   
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
@@ -797,7 +797,7 @@
        <xsd:enumeration value="midBitsToEnd">
          <xsd:annotation>
            <xsd:documentation><![CDATA[
-             Move the bits with indecis [x, x+y] to
+             Move the bits with indices [x, x+y] to
              the end of the layer block.
 
              x,y are provided as a space delimated list in layerTransformArgs
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
@@ -27,11 +27,12 @@ import org.apache.daffodil.processors.LayerCharsetEv
 import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.io.ExplicitLengthLimitingStream
 import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.processors.LayerTransformArgsEv
 
 class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
   extends LayerTransformer() {
 
-  override def wrapLayerDecoder(jis: java.io.InputStream) = {
+  override def wrapLayerDecoder(jis: java.io.InputStream, sate: PState) = {
     val s = new java.util.zip.GZIPInputStream(jis)
     s
   }
@@ -42,7 +43,7 @@ class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
     s
   }
 
-  override protected def wrapLayerEncoder(jos: java.io.OutputStream): java.io.OutputStream = {
+  override protected def wrapLayerEncoder(jos: java.io.OutputStream, state: UState): java.io.OutputStream = {
     val s = new java.util.zip.GZIPOutputStream(jos)
     s
   }
@@ -62,6 +63,7 @@ object GZIPTransformerFactory
     maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
     maybeLayerLengthUnits: Maybe[LayerLengthUnits],
     maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
+    maybeLayerTransformArgsEv: Maybe[LayerTransformArgsEv],
     trd: TermRuntimeData): LayerTransformer = {
 
     val xformer = new GZIPTransformer(maybeLayerLengthInBytesEv.get)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
@@ -32,7 +32,7 @@ import org.apache.daffodil.processors.LayerTransformArgsEv
 class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
   extends LayerTransformer() {
 
-  override def wrapLayerDecoder(jis: java.io.InputStream, sate: PState) = {
+  override def wrapLayerDecoder(jis: java.io.InputStream, state: PState) = {
     val s = new java.util.zip.GZIPInputStream(jis)
     s
   }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
@@ -28,6 +28,7 @@ import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.io.ExplicitLengthLimitingStream
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.processors.LayerTransformArgsEv
+import org.apache.daffodil.processors.ParseOrUnparseState
 
 class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
   extends LayerTransformer() {
@@ -64,6 +65,7 @@ object GZIPTransformerFactory
     maybeLayerLengthUnits: Maybe[LayerLengthUnits],
     maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
     maybeLayerTransformArgsEv: Maybe[LayerTransformArgsEv],
+    state: ParseOrUnparseState,
     trd: TermRuntimeData): LayerTransformer = {
 
     val xformer = new GZIPTransformer(maybeLayerLengthInBytesEv.get)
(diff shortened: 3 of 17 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
```diff
@@ -27,11 +27,12 @@ import org.apache.daffodil.processors.LayerCharsetEv
 import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.io.ExplicitLengthLimitingStream
 import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.processors.LayerTransformArgsEv
 
 class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
   extends LayerTransformer() {
 
-  override def wrapLayerDecoder(jis: java.io.InputStream) = {
+  override def wrapLayerDecoder(jis: java.io.InputStream, sate: PState) = {
```
mbeckerle: Typo "sate" not "state".

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,339 @@
...
+    buff.map(b => String.format("%02x", Byte.box(b))).mkString(" ")
+  }
+
+  private def initialize(state: ParseOrUnparseState) {
+    length = layerLengthInBytesEv.evaluate(state).toInt
+    val argsString = layerTransformArgsEv.evaluate(state)
+      .trim()
+      .split("\\ +")
+    if (argsString.length != 2) {
+      state.SDE(s"""midBitsToEnd layer transform requires exactly 2 arguements, recieved ${argsString.length}: "${argsString.mkString(",")}" """)
```
mbeckerle: Typo: arguements => arguments

mbeckerle: SDE should not be given a string that has things substituted into it like this.

SDE takes a format string and then varargs, for a reason, which is Internationalization. The format string is used as a hash key into internationalized translations of the string, and then the args are selected from the arglist and embedded into various languages in different orders generally.

So don't feed SDE a string in engilish that has the args substituted in. To do so subverts the design for I18n.

daffodil-test/src/test/resources/org/apache/daffodil/layers/midBitsToEnd.tdml
```diff
@@ -0,0 +1,80 @@
...
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:ex="http://example.com" xmlns:tns="http://example.com" defaultRoundTrip="true">
+
+  <tdml:defineSchema name="s1" elementFormDefault="unqualified">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:defineFormat name="general">
+      <dfdl:format ref="ex:GeneralFormat" />
+    </dfdl:defineFormat>
+    <dfdl:defineFormat name="bits8_9_toEnd">
```
mbeckerle: Why is this specifying "7 2" if the format is "bits8_9_toEnd" ? 
The values in "7 2" the first value, the index of the first bit, should be 1 based in DFDL, not zero based.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,339 @@
...
+    }
+    try {
+      firstBitIndex = argsString(0).toInt
+      numBits = argsString(1).toInt
+    } catch {
+      case _: NumberFormatException => state.SDE(s"""midBitsToEnd layer transform requires Integer arguements, recieved: "${argsString.mkString(",")}"""")
```
mbeckerle: See comment elsewhere about I18N and not passing strings to SDE.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,339 @@
...
+     * We will end up moving a multiple of 8 bits. This is okay, since, after we shift,
+     * any extra bytes will be left passed the length byte, and so will be ignored
+     */
+    var dstByteIndex = length
+    var srcByteIndex = firstBitByteIndex
+    while (dstByteIndex < buffSize) {
+      val srcByte = buff(srcByteIndex) & bitMask
+      val srcByteNext = buff(srcByteIndex + 1) & bitMask
+
+      val byteUpper = (srcByte << firstBitBitIndex) & bitMask
```
mbeckerle: Clearly this doesn't work for different bit orders.  The direction of shift has to be different. 
Was that intentional, did we not provide code for MSBF as well as LSBF? I realize the motivating use case for this is LSBF, littleEndian, so really we only need code for that case. But the other cases ought to fail with SDE that indicate the feature is not yet implemented.

daffodil-test/src/test/resources/org/apache/daffodil/layers/midBitsToEnd.tdml
```diff
@@ -0,0 +1,80 @@
...
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="midBitsToEnd1" root="root" model="s1">
+    <tdml:document>
+      <tdml:documentPart type="bits"><![CDATA[
+      0000 0011
+      1100 0000
```
mbeckerle: This is taking 2 bits in Most-significant-bit-first positions 1 and 2, and moving them to the end, shifting everything toward the most-significant bit. The 2 extracted bits end up as the least-significant bits of the final byte. 

That's the MSBF-version of this feature. You need a test that does LSBF don't you?

daffodil-test/src/test/resources/org/apache/daffodil/layers/midBitsToEnd.tdml
bsloane1650: Fixed. I also took this opportunity to append "0b" to the various index variables to clarify that that the code is internally working 0-based (the input arguement is now converted from 1-based to 0-based during initialization)

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
bsloane1650: I believe the case implemented is MSBF (as you point out, elsewhere, that seems to be the case tested).

Is the endianness actually relevent to this transform? My understanding is that endianness refers to how bits are interpereted as atomic fields, but broadly speaking the the lower indexed bytes always come first. Eg. if we have 2 16-byte fields in the following datastream:

```
0000 0001  0000 0000 
0000 0000  0000 0000
```

Endianness will determine if the first field is 1 or 256, but the first field is always represented by the first 2 bytes. Given that the transform has no knowledge of where field boundaries are, it doesn't make sense to respond to endianess.

As an aside, I believe our motivating case is actually MSBF, but it is probably best we discuss that in a more private forum.

Regardless, it is probably best I add the LSBF support now, while this shifting logic is still fresh in my mind.

mbeckerle: You are right that byte order doesn't matter. BitOrder does since the bits are potentially spanning a byte boundary and/or smaller than a byte.  ByteOrder is constrained by BitOrder - only certain combos are allowed. Otherwise yeah, it doesn't matter.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,345 @@
...
+  protected def wrapLimitingStream(jos: OutputStream, state: UState): OutputStream = {
+    initialize(state)
+    jos
+  }
+  protected def wrapLimitingStream(jis: InputStream, state: PState): InputStream = {
+    initialize(state)
+    new ExplicitLengthLimitingStream(jis, length)
+  }
+
+}
```
stevedlawrence: Make sure to add a newline before each function defiton above.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,345 @@
...
+
+object midBitsToEndTransformerFactory
```
stevedlawrence: Capitalize object names.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,345 @@
...
+      val srcByte = buff(firstBitByteIndex0b) & bitMask
+      val byteLower = ((srcByte << firstBitBitIndex0b) & bitMask) >>> firstBitBitIndex0b
+      val byteUpper = savedBits
+      val byte = (byteLower | byteUpper).toByte
+      buff(firstBitByteIndex0b) = byte
+    }
+
+  }
+
+  private def doEncode(buff: Array[Byte]): Unit = {
```
stevedlawrence: At first look, both the doEncode and doDecode function seems overly complex to me, and potentially not optimal. Both do two shifts, one to the right and one to the left. I'm still going through these, but it seems this could be simplified some and elimiate a shift.

daffodil-test/src/test/resources/org/apache/daffodil/layers/midBitsToEnd.tdml
bsloane1650: Added LSBF test

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -655,6 +656,18 @@ trait LayeringRuntimeValuedPropertiesMixin
       Nope
     }
   }
+  
+  final lazy val maybeLayerTransformArgsEv = {
+    if(optionLayerTransformArgsRaw.isDefined){
```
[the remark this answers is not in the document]
mbeckerle: It is ok if newly added properties have defaults. 
The whole layering proposal is still a proposal. It's unclear if it will evolve to support a <dfdl:defineLayer name="..."> type annotation to group things. 
So I'm not worried about this sort of thing currently.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
```diff
@@ -794,6 +794,17 @@
            </xsd:documentation>
          </xsd:annotation>
        </xsd:enumeration>
+       <xsd:enumeration value="midBitsToEnd">
+         <xsd:annotation>
+           <xsd:documentation><![CDATA[
+             Move the bits with indecis [x, x+y] to
+             the end of the layer block.
+
+             x,y are provided as a space delimated list in layerTransformArgs
```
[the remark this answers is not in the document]
bsloane1650: Positional arguments seem simpler (particularly if these arguments are going to be computed instead of hard coded). And we don't really have named arguements anywhere else in DFDL.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
```diff
@@ -0,0 +1,345 @@
...
+  def lastBitByteIndex0b = ((firstBitIndex0b + numBits) / 8)
+  def firstBitBitIndex0b = firstBitIndex0b % 8
+  def lastBitBitIndex0b = (firstBitIndex0b + numBits) % 8
+
+  private def buffToHex(buff: Array[Byte]): String = {
+    buff.map(b => String.format("%02x", Byte.box(b))).mkString(" ")
+  }
+
+  private def initialize(state: ParseOrUnparseState) {
+    length = layerLengthInBytesEv.evaluate(state).toInt
```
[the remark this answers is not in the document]
bsloane1650: The original issue with making length a val was the state was not available at the top level. It appears the only barrier to including state at the top level is that we are serializing these objects for some reason, and cannot serialize state. Making state transient would probably work, but I instead am moving the DPath evaluations to the factory.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/midBitsToEndTransformer.scala
bsloane1650: A lot of the shifting is really just masking. It looks like our Util.Misc has a Bits.maskR and Bits.maskL that would provide a more readable solution.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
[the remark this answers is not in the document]
stevedlawrence: DFDL doesn't really have positional arguments either. Every property has it's own DFDL type. Maybe the more DFDL-y way is to have a unique property for for each argument, e.g. dfdl:mbitsToEndFirstBitIndex, dfdl:mbitsToEndNumBits, but that's probably going to get messy, and would make pluggable transforms more difficult. I agree that positional arguments are definitely simpler (though, from an implementation perspective not significantly so), but they aren't as extensible. Adding new arguments later or deprecating old ones becomes diffiult. Optional args are impossible. And it's not clear from just looking at the layerTransformRags value what the args actually mean. So there's just some limitations.

But, maybe we don't need that flexablity. I guess this is fine for now, and if we realize it's too limiting or confusing in the real world, we can always change how things work. Perhaps we should consider extensions as a sort of "beta" featue that are subject to change, and so we don't have to worry quite so much about backwards compatability.

daffodil-test/src/test/scala/org/apache/daffodil/layers/TestLayers.scala
```diff
@@ -44,4 +45,9 @@ class TestLayers {
   @Test def test_layersErr1() { runner.runOneTest("layersErr1") }
   @Test def test_layers4() { runner.runOneTest("layers4") }
 
+  @Test def test_midBitsToEndMSBF1() { midBitsToEndRunner.runOneTest("midBitsToEndMSBF1") }
+  @Test def test_midBitsToEndLSBF1() { midBitsToEndRunner.runOneTest("midBitsToEndLSBF1") }
+
+  @Test def test_midBitsToEnd_unsufficentBits_1() { midBitsToEndRunner.runOneTest("midBitsToEnd_unsufficentBits_1") }
+
```
stevedlawrence: Would it be possible to add more tests? There's a lot of complex bit logic/shifting going on up above. Would be nice to see some variation in tests to try to make sure edge cases are covered, e.g. firstBitIndex and/or numBits more than 8 bits, invalid argument values, etc.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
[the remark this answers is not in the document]
bsloane1650: DFDL does have DPath, which is where the only other function-ish feature I can think of is present in DFDL. In fact, the transformArgs itself is in DPath, and I imagine we will be adding more ways for DPath to influence layer transforms as time goes on.

mbeckerle:
I believe we determined this interesting layering transform isn't actually required for the data format we thought needed it.

So do we want to incorporate this into Daffodil, or abandon it. 

I suggest keeping the branch even if we don't incorporate it. We might build a tutorial around it "how to write a layering addition to Daffodil", and use this sort of thing as a non-trivial motivating example.

stevedlawrence:
I'd vote to just close this pull request rather than incorporating it. If we incorporate, it's just one more thing we need to support and maintain.

Note that we don't really need to keep the branch around if we do close it. Even in that case, the PR and all its comments will always exist. And the PR branch can always be fetched with
```
git fetch origin refs/pull/210/head:pr-210
```
So it is effectively part of the repository, just not one that is as visible.

I'd also say that this commit isn't that good of a tutorial for how to add a layer. It's fairly complex, and adds new features (e.g. layer arguments) that most layering things won't need. A tutorial would be good, but this has too much overhead, IMO.

mbeckerle:
I can go with that.

bsloane1650:
Closing sounds good to me as well.

PR #212, opened 2019-04-25
Catch error when delimiter expression fails

jadams-tresys:
We weren't handling a potential failure when an expression representing
a delimiter failed and it was resulting in a very unhelpful Assert
failure regarding delimiter stack lengths. Now such expression failures
will generate an SDE instead.

DAFFODIL-2083, DAFFODIL-2112

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
@@ -57,26 +57,28 @@ class DelimiterStackParser(
     val newLocalIndex = start.mpstate.delimiters.length
     start.mpstate.delimitersLocalIndexStack.push(newLocalIndex)
 
-    // evaluate and add delimiters to the stack
-    var i: Int = 0
-    while (i < delimiters.length) {
-      start.mpstate.delimiters ++= delimiters(i).evaluate(start)
-      i += 1
+    try {
+      // evaluate and add delimiters to the stack
+      var i: Int = 0
+      while (i < delimiters.length) {
+        start.mpstate.delimiters ++= delimiters(i).evaluate(start)
+        i += 1
+      }
+
+      // set the index of the newly added delimiters
+      val newDelimLen = start.mpstate.delimiters.length
+      i = newLocalIndex
+      while (i < newDelimLen) {
+        start.mpstate.delimiters(i).indexInDelimiterStack = i
+        i += 1
+      }
+
+      // parse
+        bodyParser.parse1(start)
+    } finally {
+      // pop delimiters
+      start.mpstate.delimiters.reduceToSize(start.mpstate.delimitersLocalIndexStack.pop)
     }
-
-    // set the index of the newly added delimiters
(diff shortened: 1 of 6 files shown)


jadams-tresys:
Just pushed up a small fix to the test and rebased off of the most recent approved commits, which fixed the issue preventing the passing test from round tripping.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -57,26 +57,28 @@ class DelimiterStackParser(
...
+      }
 
-    // pop delimiters
-    start.mpstate.delimiters.reduceToSize(start.mpstate.delimitersLocalIndexStack.pop)
+      // parse
+        bodyParser.parse1(start)
+    } finally {
+      // pop delimiters
+      start.mpstate.delimiters.reduceToSize(start.mpstate.delimitersLocalIndexStack.pop)
+    }
```
stevedlawrence: This looks like the right fix to clean up the delimiter stack when evaluate throws an exception.

However, the DFDL spec says this for dfdl:terminator:
> If dfdl:terminator is "" (the empty string), then the terminator region is of length zero, and no terminator is expected. It is not permitted for an expression to return an empty string, that is a schema definition error.

So we've fixed the unhelpful abort when .evaluate() fails and throws an exception, but haven't resolved the issue where this error should be an SDE instead of a PE. That issue is unrelated to this bug so might be worth just creating a new ticket to track it.

+1 :+1:

jadams-tresys: So I added a small fix to change it to an SDE.  As far as I can tell form the code, the only thing that will generate an IllegalArgumentException in DPath is the StringToNonEmptyString object, which is exactly where we want to create the SDE.

jadams-tresys: Nevermind, IllegalArgumentException does get used in other places, will remove for now.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/nameDOB_test.tdml
```diff
@@ -0,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stevedlawrence: Missing apache header, causing the rat check to fail.

jadams-tresys: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
jadams-tresys: Created DAFFODIL-2114 for the fix for generating SDE's instead of parse errors for these empty string delimter expressions

stevedlawrence:
+1 :+1:

PR #213, opened 2019-04-27
fix eclipse classpaths for no more scala-debug directories.

mbeckerle:


eclipse-projects/core-test/.classpath
```diff
@@ -1,92 +1,117 @@
...
-***********
-*********** Entries below this comment are maintained using the UpdateEclipseClasspaths
-*********** Utility and should not be modified by hand or using the Eclipse
-*********** BuildPath... GUI dialog.
-***********
```
stevedlawrence: The above notice no longer exits, making me think this was generated from eclipse instead of the UpdateEclipseClasspaths tool?

stevedlawrence:
Also, commit message is missing a bug number.

eclipse-projects/core-test/.classpath
mbeckerle: True. The update eclipse classpaths tool  needs fixing given the cross-testing rig stuff that was added last fall. I haven't revisited that yet, which shows how little programming I've been getting done of late :-(

mbeckerle:
Closed. Will address the tool for maintaining these classpaths, and probably once that is fixed we need to just remove the eclipse-isms from the check-in. 

Might make sense to even just get rid of the entire eclipse-projects substructure (or generate all of it).

PR #215, opened 2019-05-10
Ensure fragment bytes are delivered correctly when dfdl:outputValueCalc is followed by a different bitOrder

stevedlawrence:
When an OVC element is followed by an element with a different bitOrder,
the OVC causes a suspension and the buffered data output stream will
have a different bitOrder. Eventually the suspension will be resolved
and we may need to deliver a fragment byte to the direct data output
stream. We previously did that with putLong, but that uses the
FormatInfo from the suspension with the incorrect bitOrder. And
unfortunately we do not have a FormatInfo associated with the buffered
data output stream with the correct bitOrder.

To resolve this issue, this patch only uses putLong + FormatInfo when we
know the bitOrders of the direct and buffered data output streams must
be the same, i.e. the direct data output stream *does not* end on a byte
boundary. In this case, the FormatInfo from the suspension has the right
information, and we can use putLong to handle the complexities for
combining fragment bytes.

On the other hand, if the direct data output stream *does* end on a byte
boundary, then the bitOrder could be different, so we can't use putLong
since the FormatInfo is wrong. But we don't need to--we can just copy
the fragment byte from buffer to direct since it already ended on a byte
boundary.

DAFFODIL-2125

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
@@ -790,13 +790,34 @@ object DirectOrBufferedDataOutputStream {
             val wholeBytesWritten = directDOS.putBytes(ba, 0, byteCount, finfo)
             Assert.invariant(byteCount == wholeBytesWritten)
             if (nFragBits > 0) {
-              val origfrag = bufDOS.fragmentLastByte
-              val fragNum =
-                if (finfoBitOrder eq BitOrder.MostSignificantBitFirst)
-                  origfrag >> (8 - nFragBits)
-                else
-                  origfrag
-              Assert.invariant(directDOS.putLongUnchecked(fragNum, nFragBits, finfo))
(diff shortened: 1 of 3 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -790,13 +790,34 @@ object DirectOrBufferedDataOutputStream {
...
+                // its new fragment byte should be exactly the same as the
+                // buffered DOS fragment byte. So in this case, just copy the
+                // frag byte information from buffered to direct.
+                directDOS.setFragmentLastByte(bufDOS.fragmentLastByte, bufDOS.fragmentLastByteLimit)
+              } else {
+                // If the direct DOS wasn't byte aligned, then we need logic
+                // logic to write the buffered DOS fragment after the direct
+                // DOS fragment. Fortunately putLong has all of this logic.
+                // Like above, the call the putLong potentially uses the wrong
+                // finfo since it may have come from a suspensio. However, all
```
jadams-tresys: Missed an 'n' on suspension.

PR #216, opened 2019-05-10
Do not abort when an import/includes fails

stevedlawrence:
When using XML Catalogs, it's very easy for import/includes to resolve
to URI's that do not exist. This results in an IOException when trying
to open the stream. There are also other cases related to IO that could
causing opening of the stream to fail. When these failures happen, we
currently abort. This is wrong, since these exception are expected to be
able to happen.

So instead of aborting with an assert when there's an exception, we
should instead just let any exceptions from opening the stream bubble
up. The XML Catalog Resolver handles exceptions correctly, and Daffodil
handles those failures and creates a helpful SDE about what stream it
tried to open and why it failed.

DAFFODIL-2126

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
@@ -989,4 +989,35 @@ class TestCLIparsing {
     }
   }
 
+  @Test def test_CLI_Parsing_XCatalog_Resolution_Failure() {
+    val schemaFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_import_failure.dfdl.xsd")
+    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+
+    val xcatalogFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_invalid.xml")
+    val testXcatalogFile = if (Util.isWindows) Util.cmdConvert(xcatalogFile) else xcatalogFile
+
+    val DAFFODIL_JAVA_OPTS = Map("DAFFODIL_JAVA_OPTS" -> ("-Dxml.catalog.files=" + testXcatalogFile + " -Xms256m -Xmx2048m -Djline.terminal=jline.UnsupportedTerminal -Dfile.encoding=UTF-8"))
+
+    val shell = Util.startIncludeErrors("", envp = DAFFODIL_JAVA_OPTS)
+
+    try {
+      val cmd = String.format(Util.echoN("X") + "| %s parse -s %s", Util.binPath, testSchemaFile)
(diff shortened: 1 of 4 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -237,12 +237,7 @@ class DFDLCatalogResolver private ()
     optURI match {
       case None => null
       case Some(uri) => {
-        val resourceAsStream =
-          try {
-            uri.toURL.openStream() // This will work.
-          } catch {
-            case _: java.io.IOException => Assert.invariantFailed("found resource but couldn't open")
-          }
+        val resourceAsStream = uri.toURL.openStream()
```
mbeckerle: You are in daffodil-lib code, which defines the ThrowsSDE stuff. You should be able to actually report an SDE from this if resourceAsStream comes back null, or catch the IOException, and nest whatever the I/O exception was as the SDE cause.  Perhaps that happens at a high level of catch structure?
The key is that this cause here gets captured.

stevedlawrence: The DFDLCatalogResolver, where this code lives, doesn't mixin ThrowsSDE, so that would be a change to allow it to do that. In fact, this function actually gets called from deep within Xerces, and looking at the Xerces code, I think this function actually shouldn't ever throw an exception, including an SDE or IOException. I believe it should should return either an Input or null. And if null, Xerces then does some stuff to figure out why it failed and throw an actual exception, which Daffodil already handles and converts to an SDE appropriately.

So I think we should actually just catch the IOException and return null here instead.

Interestingly, the error ends up the same whether we catch and return null or let it bubble up, which is something like this:
> [error] Schema Definition Error: Error loading schema due to /this/path/does/not/exist (No such file or directory)

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -989,4 +989,34 @@ class TestCLIparsing {
...
+    val DAFFODIL_JAVA_OPTS = Map("DAFFODIL_JAVA_OPTS" -> ("-Dxml.catalog.files=" + testXcatalogFile + " -Xms256m -Xmx2048m -Djline.terminal=jline.UnsupportedTerminal -Dfile.encoding=UTF-8"))
+
+    val shell = Util.startIncludeErrors("", envp = DAFFODIL_JAVA_OPTS)
+
+    try {
+      val cmd = String.format(Util.echoN("X") + "| %s parse -s %s", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+
+      shell.expectIn(1, contains("Schema Definition Error"))
+      shell.expectIn(1, contains("/this/path/does/not/exist"))
```
stevedlawrence: This line is going to fail in windows due back vs forward slashes, need to move this into the isWIndows check below.

PR #218, opened 2019-05-16
Implement dfdl:choiceLength='explicit'

jadams-tresys:
DAFFODIL-640

May have made this more complicated than it needed to be.  For some reason I thought that choiceLength accepts an expression, but it does not.  May need to simplify some things.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
@@ -46,6 +46,7 @@ import org.apache.daffodil.schema.annotation.props.gen.DFDLBaseTypeMixin
 import org.apache.daffodil.schema.annotation.props.gen.DFDLSimpleTypeMixin
 import org.apache.daffodil.schema.annotation.props.gen.LengthAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.LengthKind
+import org.apache.daffodil.schema.annotation.props.gen.ChoiceAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.OccursAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.Representation
 import org.apache.daffodil.schema.annotation.props.gen.Sequence_AnnotationMixin
@@ -84,6 +85,7 @@ import org.apache.daffodil.schema.annotation.props.Found
 trait TermRuntimeValuedPropertiesMixin
   extends DFDLBaseTypeMixin
   with PropertyReferencedElementInfosMixin
+  with ChoiceAGMixin
   with RawCommonRuntimeValuedPropertiesMixin { decl: Term =>
 
   private lazy val encodingExpr = LV('encoding) {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
@@ -26,6 +26,7 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.cookers.ChoiceBranchKeyCooker
 import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.equality._
+import org.apache.daffodil.schema.annotation.props.gen.ChoiceLengthKind
 import org.apache.daffodil.schema.annotation.props.gen.ChoiceKeyKindType
 import org.apache.daffodil.dsom.ChoiceTermBase
 import scala.util.Try
@@ -35,6 +36,7 @@ import scala.math.BigInt
 import org.apache.daffodil.cookers.RepValueCooker
 import org.apache.daffodil.processors.RangeBound
 import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.MaybeInt
 
 /*
  * The purpose of the ChoiceCombinator (and the parsers it creates) is to
@@ -67,9 +69,19 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
 
   lazy val optChoiceDispatchKeyKind = Some(ch.choiceDispatchKeyKind)
 
+  // dfdl:choiceLength is always specified in bytes
+  private lazy val choiceLengthInBits: MaybeInt = ch.choiceLengthKind match {
+    case ChoiceLengthKind.Explicit => MaybeInt(ch.choiceLength * 8)
+    case ChoiceLengthKind.Implicit => MaybeInt.Nope
+  }
+
   lazy val parser: Parser = {
     if (!ch.isDirectDispatch) {
-      new ChoiceParser(ch.termRuntimeData, parsers.toVector)
+      val cp = new ChoiceParser(ch.termRuntimeData, parsers.toVector)
+      ch.choiceLengthKind match {
+        case ChoiceLengthKind.Implicit => cp
+        case ChoiceLengthKind.Explicit => new SpecifiedLengthChoiceParser(cp, ch.choiceRuntimeData, choiceLengthInBits.get)
+      }
     } else {
       val dispatchBranchKeyValueTuples: Seq[(String, Gram)] = alternatives.flatMap { alt =>
         val keyTerm = alt.context.asInstanceOf[Term]
@@ -176,7 +188,6 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
           new ChoiceDispatchCombinatorParser(ch.termRuntimeData, ch.choiceDispatchKeyEv, serializableMap, serializableKeyRangeMap)
         case ChoiceKeyKindType.Speculative => Assert.invariantFailed("ChoiceKeyKindType==speculative while isDirectDispatch==true")
       }
-
     }
   }
 
@@ -199,8 +210,9 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
       val mapValues = eventUnparserMap.map { case (k, v) => v }.toSeq.filterNot(_.isEmpty)
       if (mapValues.isEmpty)
         new NadaUnparser(null)
-      else
-        new ChoiceCombinatorUnparser(ch.modelGroupRuntimeData, eventUnparserMap)
+      else {
+        new ChoiceCombinatorUnparser(ch.modelGroupRuntimeData, eventUnparserMap, choiceLengthInBits)
+      }
     } else {
       // Choices inside a hidden group ref are slightly different because we
       // will never see events for any of the branches. Instead, we will just
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ChoiceAndOtherVariousUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ChoiceAndOtherVariousUnparsers.scala
@@ -21,8 +21,10 @@ import org.apache.daffodil.processors._
 import org.apache.daffodil.infoset._
 import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.dfa.DFADelimiter
+import org.apache.daffodil.schema.annotation.props.gen.ChoiceLengthKind
 import org.apache.daffodil.util.Maybe._
 import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.MaybeInt
 import org.apache.daffodil.util.Maybe._
 
 class ComplexTypeUnparser(rd: RuntimeData, bodyUnparser: Unparser)
@@ -43,7 +45,10 @@ class ComplexTypeUnparser(rd: RuntimeData, bodyUnparser: Unparser)
   }
 }
 
-class ChoiceCombinatorUnparser(mgrd: ModelGroupRuntimeData, eventUnparserMap: Map[ChoiceBranchEvent, Unparser])
+class ChoiceCombinatorUnparser(
+  mgrd: ModelGroupRuntimeData,
+  eventUnparserMap: Map[ChoiceBranchEvent, Unparser],
+  choiceLengthInBits: MaybeInt)
   extends CombinatorUnparser(mgrd)
   with ToBriefXMLImpl {
   override def nom = "Choice"
@@ -71,7 +76,18 @@ class ChoiceCombinatorUnparser(mgrd: ModelGroupRuntimeData, eventUnparserMap: Ma
       UnparseError(One(mgrd.schemaFileLocation), One(state.currentLocation), "Encountered event %s. Expected one of %s.",
         key, eventUnparserMap.keys.mkString(", "))
     }
-    childUnparser.unparse1(state)
+
+    if (choiceLengthInBits.isDefined) {
+      val suspendableOp = new ChoiceUnusedUnparserSuspendableOperation(mgrd, choiceLengthInBits.get)
+      val choiceUnusedUnparser = new ChoiceUnusedUnparser(mgrd, choiceLengthInBits.get, suspendableOp)
+
+      suspendableOp.captureDOSStartForChoiceUnused(state)
+      childUnparser.unparse1(state)
+      suspendableOp.captureDOSEndForChoiceUnused(state)
+      choiceUnusedUnparser.unparse(state)
+    } else {
+      childUnparser.unparse1(state)
+    }
   }
 }
 
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
@@ -22,6 +22,8 @@ import org.apache.daffodil.infoset.DIComplex
 import org.apache.daffodil.infoset.DIElement
 import org.apache.daffodil.infoset.DISimple
 import org.apache.daffodil.processors.ElementRuntimeData
+import org.apache.daffodil.processors.ModelGroupRuntimeData
+import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.SuspendableOperation
 import org.apache.daffodil.processors.UnparseTargetLengthInBitsEv
 import org.apache.daffodil.processors.charset.BitsCharset
@@ -37,6 +39,8 @@ import org.apache.daffodil.processors.LengthEv
 import org.apache.daffodil.processors.Evaluatable
 import org.apache.daffodil.util.MaybeJULong
 import org.apache.daffodil.io.DirectOrBufferedDataOutputStream
+import org.apache.daffodil.io.ZeroLengthStatus
+import org.apache.daffodil.io.DataOutputStream
 import passera.unsigned.ULong
 
 import java.nio.charset.MalformedInputException
@@ -405,26 +409,9 @@ sealed trait NeedValueAndTargetLengthMixin {
   }
 }
 
-class ElementUnusedUnparserSuspendableOperation(
-  override val rd: ElementRuntimeData,
-  override val targetLengthEv: UnparseTargetLengthInBitsEv,
-  override val maybeLengthEv: Maybe[LengthEv],
-  override val maybeCharsetEv: Maybe[CharsetEv],
-  override val maybeLiteralNilEv: Maybe[NilStringLiteralForUnparserEv])
-  extends SuspendableOperation
-  with NeedValueAndTargetLengthMixin {
+trait SkipTheBits { self: SuspendableOperation =>
 
-  /**
-   * determine delta between value length and target length
-   *
-   * and skip that many bits.
-   */
-  override def continuation(ustate: UState) {
-    val skipInBits = getSkipBits(ustate)
-    if (skipInBits < 0)
-      UE(ustate, "Data too long by %s bits. Unable to truncate.", -skipInBits)
-    skipTheBits(ustate, skipInBits)
-  }
+  protected val rd: RuntimeData
 
   protected final def skipTheBits(ustate: UState, skipInBits: Long) {
     if (skipInBits > 0) {
@@ -439,6 +426,29 @@ class ElementUnusedUnparserSuspendableOperation(
       log(LogLevel.Debug, "%s filled %s bits for %s DOS %s.",
         Misc.getNameFromClass(this), skipInBits, rd.diagnosticDebugName, ustate.dataOutputStream)
     }
+  }
+}
+
+class ElementUnusedUnparserSuspendableOperation(
+  override val rd: ElementRuntimeData,
+  override val targetLengthEv: UnparseTargetLengthInBitsEv,
+  override val maybeLengthEv: Maybe[LengthEv],
+  override val maybeCharsetEv: Maybe[CharsetEv],
+  override val maybeLiteralNilEv: Maybe[NilStringLiteralForUnparserEv])
+  extends SuspendableOperation
+  with SkipTheBits
+  with NeedValueAndTargetLengthMixin {
+
+  /**
+   * determine delta between value length and target length
+   *
+   * and skip that many bits.
+   */
+  override def continuation(ustate: UState) {
+    val skipInBits = getSkipBits(ustate)
+    if (skipInBits < 0)
+      UE(ustate, "Data too long by %s bits. Unable to truncate.", -skipInBits)
+    skipTheBits(ustate, skipInBits)
   }
 
 }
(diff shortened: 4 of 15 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
```diff
@@ -132,6 +133,20 @@ class SpecifiedLengthExplicitParser(
...
 
+class SpecifiedLengthChoiceParser(
+  eParser: Parser,
+  erd: RuntimeData,
+  lengthEv: Evaluatable[JLong],
+  toBits: Int)
+  extends SpecifiedLengthParserBase(eParser, erd) {
+
+  final override def getBitLength(s: PState): MaybeULong = {
+    val nBytesAsAny = lengthEv.evaluate(s)
```
mbeckerle: suggest name choiceLengthEv. There is already a lengthEv, so this is confusing.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -158,6 +162,44 @@ trait TermRuntimeValuedPropertiesMixin
       Nope
   }
 
+  protected final lazy val choiceLengthExpr = {
```
mbeckerle: As you pointed out, alas, all this is unnecessary because it's not allowed to be an expression.

I'd suggest keeping this and making it a feature, but the amount of testing involved to really  support that, and lack of a compelling use case suggest you should just take this out.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvElement.scala
```diff
@@ -187,14 +188,32 @@ class LengthInBitsEv(lengthUnits: LengthUnits,
...
+ * Hence, we have to compute this similarly at runtime.
+ */
+class ChoiceMinLengthInBitsEv(lengthUnits: LengthUnits,
+  lengthKind: LengthKind,
+  lengthEv: LengthEv,
+  rd: TermRuntimeData)
+  extends MinLengthInBitsEv(lengthUnits, lengthKind, Nope, 0, rd) {
+
+  override protected def lengthInLengthUnits(state: ParseOrUnparseState) =
+    lengthEv.evaluate(state).longValue()
```
mbeckerle: This also changes since there is no expression to evaluate.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
mbeckerle: Oh, yeah and you don't need this beause it's not an expression. But call it choiceLength anyway.

jadams-tresys: I've removed all Ev stuff and just reimpemented a simplified version of ElementUnusedUnparser without evaluatables.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +463,64 @@ class ElementUnusedUnparser(
...
+  targetLengthInBits: Long)
+  extends SuspendableOperation {
+
+  /**
+   * determine delta between value length and target length
+   *
+   * and skip that many bits.
+   */
+  override def continuation(ustate: UState) {
+    val e = ustate.currentInfosetNode.asInstanceOf[DIElement]
```
stevedlawrence: I'm wondering if we might run into issues by usign the value length of the current infoset node. For example, say we have a single complex element that is a sequence of two choiceLengthKind="explicit" choices:
```xsd
<xs:element name="foo">
  <xs:complexType>
    <xs:sequence>
      <xs:choice dfdl:choiceLengthKind="explicit" dfdl:choiceLength="X">
        ....
      </xs:choice>
      <xs:choice dfdl:choiceLengthKind="explicit" dfdl:choiceLength="X">
        ....
      </xs:choice>
    </xs:sequence>
  </xs:complexType>
</xs:element>
```
When we unparse the choice unsed stuff, both choices are going to use the length of foo to determine how much space is unused. But we really want the length of the choice content, not the length of foo. Not sure how to handle this, since there is nothing keeping track of the length of the choice content. Do we need some representation in the infoset for a choice where the length of the choice can be hung when choice length is explicit? I think we had this at one point, but maybe got rid of it?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +463,64 @@ class ElementUnusedUnparser(
...
+    val vl = e.valueLength.lengthInBits.longValue
+    val skipInBits = targetLengthInBits - vl
+    if (skipInBits < 0)
+      UE(ustate, "Data too long by %s bits. Unable to truncate.", -skipInBits)
```
stevedlawrence: I might be worth mention an explicit length choice here in this error message to make it clear that the choice content was too long for the specified length.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +463,64 @@ class ElementUnusedUnparser(
...
+    skipTheBits(ustate, skipInBits)
+  }
+
+  protected final def skipTheBits(ustate: UState, skipInBits: Long) {
```
stevedlawrence: I think this function is a copy paste? Maybe it can be pulled out into a trait?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +463,64 @@ class ElementUnusedUnparser(
...
+  }
+
+  override def test(ustate: UState): Boolean = {
+    val e = ustate.currentInfosetNode.asInstanceOf[DIElement]
+    e.valueLength.maybeLengthInBits().isDefined
+  }
+}
+
+class ChoiceUnusedUnparser(
+  override val context: TermRuntimeData,
```
stevedlawrence: Can this be a ChoiceRuntimeData?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
```diff
@@ -428,3 +429,20 @@ class SpecifiedLengthPrefixedUnparser(
...
+
+  override lazy val runtimeDependencies = Vector()
+  override lazy val childProcessors = Vector(eUnparser)
+
+  override def unparse(state: UState): Unit = {
+    val elem = state.currentInfosetNode.asInstanceOf[DIElement]
+    captureValueLengthStart(state, elem)
+    eUnparser.unparse1(state)
+    captureValueLengthEnd(state, elem)
+  }
```
stevedlawrence: I think this is where things would need to change to create a new thing in the infoset where we can hang length information. Note that I think in the example I gave above, this would capture different lengths on the "foo" elem when each choice unparse was run.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
```diff
@@ -29,10 +30,12 @@ import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.processors.Processor
 import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.Success
+import org.apache.daffodil.processors.Evaluatable
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.Maybe.One
 import org.apache.daffodil.util.MaybeULong
 import org.apache.daffodil.util.Misc
+import org.apache.daffodil.util.Numbers
```
stevedlawrence: Remove unused imports.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
```diff
@@ -132,6 +133,18 @@ class SpecifiedLengthExplicitParser(
   }
 }
 
+class SpecifiedLengthChoiceParser(
+  eParser: Parser,
+  erd: RuntimeData,
+  choiceLength: JLong,
+  toBits: Int)
+  extends SpecifiedLengthParserBase(eParser, erd) {
```
stevedlawrence: Remove toBits and change choiceLength to chocieLengthInBits, or is toBits needed by SpecifiedLengthParserBase?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -48,9 +50,22 @@ case class ChoiceCombinator(ch: ChoiceTermBase, rawAlternatives: Seq[Gram])
 
   override def isEmpty = super.isEmpty || alternatives.isEmpty
 
+  private lazy val bitsMultiplier = ch.lengthUnits match {
+    case LengthUnits.Bits => 1
+    case LengthUnits.Bytes => 8
+    case LengthUnits.Characters if ch.knownEncodingIsFixedWidth => ch.knownEncodingWidthInBits
+    case _ => 0 // zero means can't multiply to get width in bits.
+  }
```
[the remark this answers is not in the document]
mbeckerle: A choiceLengthUnits property may appear in the future. Today's byte-centric view of choices is not necessarily going to stand.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
mbeckerle: I guess I don't see the need for something on the infoset. Choices have no corresponding structure in the infoset. Can you motivate this idea further?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
mbeckerle: Good spot. This is definitely wrong to use the element's length.
The choice parser/unparser have the explicit length in their state or in the ChoiceRuntimeData object. That's the yardstick, not anything about an enclosing element.

mbeckerle: Also, let's add a test along the lines of Steve's example above to prove this works in that case.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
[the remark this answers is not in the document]
stevedlawrence: Woudl that use lengthUnits, or what that use something like choiceLengthUnits to match choiceLength? Our of curiosity, is there a reason why we have both dfdl:choiceLength and dfdl:length?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
stevedlawrence: I think we have to be careful around where the state is stored. ChoiceParser/Unparser and ChoiceRUntimeData are both shared between different parses/unparses. So putting state there needs to think about thread safety, the state might need to be in a thread local. I suggested the infoset since it is unique per parse and it has all the logic for dealing with suppressions and calculating lengths when there's relative stuff going on.

Maybe instead of adding a ChoiceNode to the infoset, each ComplexNode in the infoset has both a length state for itself **and** an array of length state for any child choices with explicit lengths? Each choice would need to statically know which state in the array is its own, so that might have some challenges. So the below might look something like:
```scala
val v1 = e.choiceValueLength[staticIndex].lengthInBits.longValue
```

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
mbeckerle: Actually, this is harder than that.
Due to the fact that the contents of the choice may have interior alignment, or variable-length things in it computed by dfdl:outputValueCalc, we can't just unparse, and know the length at the point the call returns.

I have almost the identical problem in the new separators code. I need to measure how big something is when unparsing, and decide to lay down a separator or not depending on whether it came out to be zero-length or not. 

Here the situation is almost the same I think. We need to unparse, measure relative to the choiceLength, and skip the unused bytes. But we can't know that length until, well, it's known. 

So the technique is to unparse a suspendable unparser that does nothing, just causes a new DataOutputStream to be split off of the prior, so as to establish a base position. Then unparse the choice, then unparse a suspendable unparser that does a skip N, where N is computed based on the starting position and ending positions of the data output streams being known, and the choiceLength. This would also be able to signal an error if it is too long. 

If that makes sense to everyone, then I suggest we finish review on the separators/sequences branch, merge it, and use the suspendable unparsers created there (one of them - the does-nothing one - can be reused).

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -444,7 +447,7 @@ class ElementUnusedUnparserSuspendableOperation(
 }
 
 class ElementUnusedUnparser(
-  override val context: ElementRuntimeData,
+  override val context: TermRuntimeData,
```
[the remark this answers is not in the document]
jadams-tresys: Yes, those can be reverted, will do.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +463,64 @@ class ElementUnusedUnparser(
 
 }
 
+/**
+ * This is basically a simplified version of the ElementUnusedUnparserSuspendableOperation.
+ * For explicit length choices, we do not need to handle length expressions, choiceLength is
+ * simply a non-negative integer.
+ */
+class ChoiceUnusedUnparserSuspendableOperation(
+  override val rd: TermRuntimeData,
```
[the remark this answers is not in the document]
jadams-tresys: Will do

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
jadams-tresys: I've pushed up a test and it does indeed demonstrate the issues with using valueLength instead of the choiceLength.
Also, for some reason this conversation was hidden by default and I missed it the first time around. Wonder if there is a setting to not hid conversations.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
jadams-tresys: So in the last commit I pushed up I resolved all the other issues and rebased onto the separator suppression code.  I think I know what needs to be done, but want to make sure I am understanding everything correctly.  Should I be changing ChoiceLengthUnparser to extend it with SuspendableUnparser, and then setting the suspendableOperation to SuppressionRegionSplitSuspendableOperation?  If so, should I move SuppressionRegionSplitSuspendableOperation out into its some sort of common area, as my use here has nothing to do with suppressing separators.

mbeckerle: You probably want ChoiceLengthUnparser to recursively unparse 

1) a splitter unparser - like in the separator thing, you just need to split the infoset. (Yes this could be hoisted out to a more general class name, since it really has nothing to do with separators per se.)

2) the choice branch itself, and then 

3) a suspendable unparser for the ChoiceUnusedRegion that comes after the choice branch. This will be a new suspendable unparser. The constructor for this unparser will take the DOS from after the split(1) above as an argument. 

The test for the suspendable operation of this ChoiceUnusedRegion unparser is going to use the dataOutputStream from the split (1) above. The test will pass when that data stream, and all subsequent ones up until but not including the current one (the ChoiceUnusedRegion is suspendable, so will also create a split of its own) are final. The current DOS is waiting for us to put the skip(n) into it for the unused region, so this one will NOT be final. 

The continuation method of the suspendable operations can then compute the exact length in bits of the data from the split (1) to just before the ChoiceUnusedRegion unparser (3) by following the chain of DOS from the first one created after the split (1) to current. That's the key value you need to then decide how many bits to skip based on the choiceLength value. 
You skip(n) those, then finalize the DOS. 

Hope that makes sense. Happy to discuss by phone.

jadams-tresys: I believe I have now implemented this correctly in the most recent commit

mbeckerle: One thing to keep in mind. The ChoiceLengthUnparser has slots in it that capture state. That means these must be created on the fly as needed and are not something that can be serialized as part of a saved parser/unparser, because those are shared state that can be used by multiple threads simultaneously. These ChoiceLengthUnparser instances must be per-thread.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -199,8 +206,13 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
       val mapValues = eventUnparserMap.map { case (k, v) => v }.toSeq.filterNot(_.isEmpty)
       if (mapValues.isEmpty)
         new NadaUnparser(null)
-      else
-        new ChoiceCombinatorUnparser(ch.modelGroupRuntimeData, eventUnparserMap)
+      else {
+        val cup = new ChoiceCombinatorUnparser(ch.modelGroupRuntimeData, eventUnparserMap)
+        ch.choiceLengthKind match {
+          case ChoiceLengthKind.Implicit => cup
+          case ChoiceLengthKind.Explicit => new ChoiceLengthUnparser(cup, ch.choiceRuntimeData, choiceLengthInBits)
```
mbeckerle: I think this won't work. ChoiceLengthUnparser has slots in it that are side-effected at runtime to capture length. That won't work if the unparser is being shared across threads.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
```diff
@@ -428,3 +431,39 @@ class SpecifiedLengthPrefixedUnparser(
...
+  rd: ChoiceRuntimeData,
+  targetLengthInBits: Long)
+  extends CombinatorUnparser(rd)
+  with StreamSplitter {
+
+  override lazy val runtimeDependencies = Vector()
+  override lazy val childProcessors = Vector(eUnparser)
+
+
+  private var maybeDOSStart : Maybe[DataOutputStream] = Maybe.Nope
```
mbeckerle: These slots are state that will not be thread safe.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
mbeckerle: You need the choice combinator unparser to dynamically allocate this ChoiceLengthUnparser at runtime. Not create it here in this grammar-tree.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
stevedlawrence: Instead of doing something like thread locals, can the capture/splitter logic just be moved out of those functions and into def unparse? And then the maybeDOStart/End vars just become local vals in def unparse that get passed into the creation of the ChoiceUnusedUnparser?

mbeckerle: They have to live on the heap, they outlive the call to the unparse method. But they can live in a heap allocated object that is part of the suspension state of the unparse invocation.

jadams-tresys:
All thread safety issues should hopefully be addressed now.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -67,9 +68,19 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
 
   lazy val optChoiceDispatchKeyKind = Some(ch.choiceDispatchKeyKind)
 
+  // dfdl:choiceLength is always specified in bytes
+  private lazy val choiceLengthInBits: Long = ch.choiceLengthKind match {
+    case ChoiceLengthKind.Explicit => ch.choiceLength * 8
+    case ChoiceLengthKind.Implicit => 0 // When using choiceLengthKind='implicit', this value will simply be ignored
```
stevedlawrence: What if we change the type of  ``choiceLengthInBits`` to a  ``Maybe[Long]``? Then we don't have to ignore the value. Either we have a length and we use it, or there's no value to use. Same effect, but the type is a hint that we may or may not have a value and we can't accidentally use a 0.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
```diff
@@ -24,6 +24,7 @@ import org.apache.daffodil.util.Logging
 import org.apache.daffodil.util.Maybe
 
 sealed abstract class ZeroLengthStatus
+extends Serializable
```
stevedlawrence: Any idea why this needs to be serializable now? Not a big deal, but this feels like a runtime thing that shouldn't need to be serialized?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ChoiceAndOtherVariousUnparsers.scala
```diff
@@ -43,7 +44,11 @@ class ComplexTypeUnparser(rd: RuntimeData, bodyUnparser: Unparser)
   }
 }
 
-class ChoiceCombinatorUnparser(mgrd: ModelGroupRuntimeData, eventUnparserMap: Map[ChoiceBranchEvent, Unparser])
+class ChoiceCombinatorUnparser(
+  mgrd: ModelGroupRuntimeData,
+  eventUnparserMap: Map[ChoiceBranchEvent, Unparser],
+  choiceLengthKind: ChoiceLengthKind,
+  choiceLengthInBits: Long)
```
stevedlawrence: I think if this becomes a ``Maybe[Long]`` then you don't need to pass in choiceLengthKind? So changing the type does simplify things a bit.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +470,102 @@ class ElementUnusedUnparser(
 
 }
 
+/**
+ * This is basically a simplified version of the ElementUnusedUnparserSuspendableOperation.
```
stevedlawrence: Do you mean SuppressableSeparatorUnparserSuspendableOperation? This looks much more liket tat that ElementUnused...Operation.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -46,6 +46,8 @@ import org.apache.daffodil.schema.annotation.props.gen.DFDLBaseTypeMixin
 import org.apache.daffodil.schema.annotation.props.gen.DFDLSimpleTypeMixin
 import org.apache.daffodil.schema.annotation.props.gen.LengthAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.LengthKind
+import org.apache.daffodil.schema.annotation.props.gen.ChoiceLengthKind
```
stevedlawrence: ChoiceLenghtKind no longer needs to be imported. There are also some other files that the only changes are now imports that can be backed out.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SuppressableSeparatorUnparser.scala
```diff
@@ -136,28 +137,6 @@ final class SuppressableSeparatorUnparserSuspendableOperation(
     }
   }
 
-  /**
-   * Given two DataOutputStream, determine the set of data output streams starting from the first,
-   * and in chain until (and including) we reach the second.
-   *
-   * If the two are the same DataOutputStream, we get back a sequence of just the one DOS.
-   */
-  private def getDOSFromAtoB(beforeDOS: DataOutputStream, afterDOS: DataOutputStream): Seq[DataOutputStream] = {
```
stevedlawrence: Did you move this to another file and forget to add it? It looks like it's stilll used above but I don't see it added back anywhere.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -460,6 +470,102 @@ class ElementUnusedUnparser(
...
+  with SkipTheBits {
+
+  private var zlStatus_ : ZeroLengthStatus = ZeroLengthStatus.Unknown
+
+  private var maybeDOSStart : Maybe[DataOutputStream] = Maybe.Nope
+  private var maybeDOSEnd : Maybe[DataOutputStream] = Maybe.Nope
+
+
+  def captureDOSStartForChoiceUnused(state: UState): Unit = {
+    val splitter = new RegionSplitUnparser(rd)
```
stevedlawrence: Missing file for this? I don't see RegionSplitUnparser anywhere.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
jadams-tresys: Not sure, I ran into this issue yesterday and thought I had done something completely wrong and redid my changes in a cleaner way and still hit this problem.  I guess its because we are serializing a ZeroLengthStatus as part of the SuspendableOperation now.  Not sure why it wasn't doing it before though

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
jadams-tresys: Stale comment.  It was originally based on ElementUnused but is now based off Suppressable.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
jadams-tresys: Will do.  I feel like I remember getting errors before about unused imports.  Did we change something?

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
stevedlawrence: But isn't the SuspendableOperation only allocated at Runtime? So its shouldn't need to be serialized. Again, not a big deal, but I feel like maybe I'm missing something.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
jadams-tresys: Added the missing file.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
stevedlawrence: Now that you mention, I think you're right. I think maybe when we added support for scala 2.12 we removed some build options that were scala 2.11 only. It looks like in 2.11 the build flag is ``-Ywarn-unused-import`` and scala 2.12 is ``-Ywarn-unused:imports`` (note the colon and the plural). So if we want this back (seems reasonable), we'd need to conditionally add the right option to scalacOptions based on scalaVersion.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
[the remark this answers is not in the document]
mbeckerle: Yes, there are a few places in DFDL where we have a single property, but having distinct ones would be better. A problematic area is dfdl:lengthKind. It would really be better to have separate lengthKinds for complex type elements and for simple type elements, because they are often different, but uniform. Typically complex types want lengthKind 'implicit' but simple types 'explicit' or something else. but 'implicit' is much less used for simple types. 

An example that is done properly is textNumberJustification, textStringJustification, textBooleanJustification, etc. those often want to be set differently. If there was just textJustification, you'd have to keep setting and resetting the property for different types of elements even though the normal pattern is strings justify to the left, and numbers to the right, booleans... depends on their representation. If it's 1 and 0, then right. If it is "true" and "false" then left. 

So if we introduce a lengthUnits for choices I would suggest it is a separate property choiceLengthUnits.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLength2.scala
```diff
@@ -476,34 +476,48 @@ class ElementUnusedUnparser(
...
-  targetLengthInBits: Long,
-  maybeDOSStartForChoiceUnused: Maybe[DataOutputStream],
-  maybeDOSEndForChoiceUnused: Maybe[DataOutputStream])
+  override val rd: ModelGroupRuntimeData,
+  targetLengthInBits: Long)
   extends SuspendableOperation
   with StreamSplitter
   with SkipTheBits {
 
   private var zlStatus_ : ZeroLengthStatus = ZeroLengthStatus.Unknown
```
mbeckerle: Here's the zero-length status use. This object is only created at runtime now, so yeah ZeroLengthStatus probably doesn't have to be serializable. That said. Doesn't hurt.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
mbeckerle: I think this was made serializable before Josh changed the objects to be runtime allocated. So possibly this doesn't *have to* be serializable now.

jadams-tresys: Mike is correct, it no longer needs it.

stevedlawrence:
+1 :+1:

PR #219, opened 2019-05-16
Ensure InputStreams created by includes/imports are closed

stevedlawrence:
- When we open a stream from a DaffodilInputSource, make sure we close
  the streams when we are done with them
- In our custom LSInput implementation (renamed to InputStreamLSInput),
  we provided an InputStream *and* an implementation of getStringData().
  However, when getStringData() returns non-null, Xerces ignores the
  fact that an InputStream is provided and so Xerces never closes it. If
  we instead make getStringData() return null, Xerces will then use the
  InputStream and will close it once the stream is parsed. The added
  benefit here is that Xerces *should* be able to look at the preamble
  and determine XML encoding, which we previously did not do with our
  getStringData implementation.

These changes should allow Daffodil to support a much larger number of
includes/imports without hitting "Too many open files" errors.

DAFFODIL-2130

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
@@ -23,8 +23,15 @@ package org.apache.daffodil.xml
  * See http://stackoverflow.com/questions/4446137/how-to-track-the-source-line-location-of-an-xml-element
  */
 
+import java.io.BufferedInputStream
 import java.io.File
+import java.io.InputStream
+import java.io.Reader
 import java.net.URI
+
+import javax.xml.XMLConstants
+import javax.xml.transform.sax.SAXSource
+
 import scala.collection.JavaConverters.asScalaBufferConverter
 import scala.xml.Elem
 import scala.xml.InputSource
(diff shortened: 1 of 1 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -463,9 +462,11 @@ trait SchemaAwareLoaderMixin {
    * it a plain old file or resource, and not try to play games to get it to
    * pick up the file/line/col information from attributes of the elements.
    */
-  def validateSchema(source: DaffodilSchemaSource) = {
-    val saxSource = new SAXSource(source.newInputSource())
+  def validateSchema(source: DaffodilSchemaSource): Unit = {
+    val inputSource = source.newInputSource()
+    val saxSource = new SAXSource(inputSource)
     sf.newSchema(saxSource)
+    inputSource.getByteStream().close()
```
mbeckerle: At the end of validation here can we assert that the stream is closed by xerces?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -302,35 +306,30 @@ object DFDLCatalogResolver {
...
-  def getStringData = {
-    this.synchronized {
-      val input: Array[Byte] = new Array[Byte](inputStream.available())
-      inputStream.read(input)
-      val contents = new String(input)
-      contents
-    }
-  }
-  def setBaseURI(baseURI: String) = myBaseURI = baseURI
+  def getStringData = null
```
bsloane1650: Can you add a comment explaining why this is important? Especially considering we do not have tests to catch the problems this solves.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
stevedlawrence: When Xerces comes across an include/import it calls into our resolver to get an LSInput to the resource. We don't currently keep track of the InputStreams we open to give back to Xerces, and there isn't an easy way to get access to Xerces handles of those InputStreams.

So if we wanted to assert they were all closed, I think we'd need to change our DFDLCatalogResolver to keep track of all the LSInputs that it gives Xerces in some state, and then at the end of validation we could confirm that all the LSInputs we created were properly closed. But that introduces state, and there's a handful of places where catalog resolvers are used so I'm not sure this is a trival change. And I don't think there's even a good way to tell if an InputStream is closed. Looks like the best you can do is try to read from the InputSource and catch an exception. Seems like a bit of a headache just to make sure Xerces is doing the right thing.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
stevedlawrence: Yep, will do.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
mbeckerle: Agreed. I thought it might be a simple assert here, but clearly beyond that, and unwise change.

PR #222, opened 2019-05-24
Daffodil 2135 array index

mbeckerle:
Please review only the second commit.

(The first commit is the whitespace changes which are a separate issue/pull request. Once that's fixed I'll rebase these changes on however we solve that. )

This fixes agnosc_remedy, army_drrs, and uscg_ucop schemas (all are FOUO, so not publicly available ) but these worked before the recent separator changes, but were broken by them. 

Those schemas themselves also needed minor changes, as they were depending on non-DFDL-conforming behavior from Daffodil. I will push those changes once this has been merged to master.

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
@@ -84,15 +84,6 @@ abstract class RepeatingChildUnparser(
    * event.
    */
   final def startArrayOrOptional(state: UState): Unit = {
-    //
-    // The Infoset events don't create an array start/end event for a
-    // truly optional (0..1) element.
-    //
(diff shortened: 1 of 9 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
```diff
@@ -91,8 +91,6 @@ abstract class RepeatingChildUnparser(
     // But an expression might still refer to dfdl:occursIndex() and we want that
     // to be 1, not wherever an enclosing parent array is.
     //
```
stevedlawrence: Should this comment be removed now? Since that's now handled elsewhere?

mbeckerle: Yes. Comment is misleading now. Removing.

PR #223, opened 2019-05-31
Daffodil 1444 schema comp simple types

mbeckerle:
This is incremental progress toward getting rid of the backpointers that prevent structure sharing in the Daffodil schema compiler.

There's some commits which are small cleanups. The last commit eliminates the backpointers from all simple type objects to the objects that reference them. This eliminates the Factory vs. Instance distinction for simple types (so far). 

I would like to try to carry out this task of fixing the schema compiler's algorithmic/data-structure flaws incrementally, by a number of smaller change sets like this one.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -107,9 +107,7 @@ final class ProcessorFactory(val sset: SchemaSet)
     unp
   }.value
 
-  lazy val rootElem = LV('rootElem) {
-    sset.rootElement(rootSpec)
-  }.value
+  lazy val rootElem = sset.root
 
   //
   // breaking this into these lines causes the order things are
@@ -151,7 +149,8 @@ final class ProcessorFactory(val sset: SchemaSet)
       Assert.usage(!isError)
       if (xpath != "/") rootElem.notYetImplemented("""Path must be "/". Other path support is not yet implemented.""")
       val rootERD = rootElem.elementRuntimeData
-      rootElem.schemaDefinitionUnless(rootERD.outputValueCalcExpr.isEmpty,
+      rootElem.schemaDefinitionUnless(
+        rootERD.outputValueCalcExpr.isEmpty,
         "The root element cannot have the dfdl:outputValueCalc property.")
       val validationMode = ValidationMode.Off
       val variables: VariableMap = rootElem.schemaDocument.schemaSet.variableMap
@@ -312,8 +311,7 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
           "or it is missing from the classpath which is%s".format(
             cnf.getMessage(),
             scala.util.Properties.javaVersion,
-            cpString
-          )
+            cpString)
         throw new InvalidParserException(msg, cnf)
       }
     }
@@ -335,9 +333,20 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
    */
   def compileSource(schemaSource: DaffodilSchemaSource): ProcessorFactory =
     Compiler.synchronized {
-      val noParent = null // null indicates this is the root, and has no parent
-      val sset = new SchemaSet(rootSpec, externalDFDLVariables, Seq(schemaSource), validateDFDLSchemas, checkAllTopLevel, noParent, tunablesObj)
-      val pf = new ProcessorFactory(sset)
+      //
+      // The SchemaSet object and the ProcessorFactory mutually refer to each other because
+      // the API allows the root to be specified by API calls on the processor factory, the compiler,
+      // or it can be inferred from the contents of the schema documents.
+      //
+      // The logic for assembling all those sources together is in the SchemaSet.
+      //
+      // To make these two objects refer to each other, we we create them lazily, and
+      // pass by name to SchemaSet constructor. This is the typical scala idiom
+      // for creating things that point mutually without using a side-effect which may or may not
+      // have happened a the time the information is demanded.
+      //
+      lazy val sset: SchemaSet = new SchemaSet(Some(pf), rootSpec, externalDFDLVariables, Seq(schemaSource), validateDFDLSchemas, checkAllTopLevel, tunablesObj)
+      lazy val pf: ProcessorFactory = new ProcessorFactory(sset)
       val err = pf.isError
       val diags = pf.getDiagnostics // might be warnings even if not isError
       if (err) {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
@@ -22,7 +22,7 @@ import scala.xml.Node
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.dpath.NodeInfo.PrimType
 
-trait Facets { self: RestrictionFactory =>
+trait Facets { self: Restriction =>
   import org.apache.daffodil.dsom.FacetTypes._
 
   requiredEvaluations(if (hasPattern) patternValues)
@@ -82,14 +82,14 @@ trait Facets { self: RestrictionFactory =>
     // Must be unique
     val enumerations = enumeration(xml)
     val distinctEnums = enumerations.distinct
-    if (enumerations.size != distinctEnums.size) context.SDE("Enumerations must be unique!")
+    if (enumerations.size != distinctEnums.size) SDE("Enumerations must be unique!")
     // Not a regular expression, but we plan to use it as one
     // so we must escape characters that can be interpreted as RegEx
     enumerations.map(s => escapeForRegex(s)).mkString("|")
   }
   final lazy val localWhitespaceValue: String = {
     whitespace(xml)
-    context.SDE("whitespaceValue is not implemented for DFDL v1.0 schemas but reserved for future use.")
+    SDE("whitespaceValue is not implemented for DFDL v1.0 schemas but reserved for future use.")
   }
 
   private def escapeForRegex(s: String): String = {
@@ -139,7 +139,7 @@ trait Facets { self: RestrictionFactory =>
     if (values.size > 0) {
       val (_, value) = values(0)
       Some(value)
-    } else None // context.SDE("Enumeration was not found in this context.")
+    } else None // SDE("Enumeration was not found in this context.")
   }
   // TODO: Tidy up.  Can likely replace getFacetValue with a similar call to combinedBaseFacets
   // as combinedBaseFacets should contain the 'narrowed' values.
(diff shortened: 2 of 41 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
```diff
@@ -22,7 +22,7 @@ import scala.xml.Node
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.dpath.NodeInfo.PrimType
 
-trait Facets { self: RestrictionFactory =>
+trait Facets { self: Restriction =>
```
mbeckerle: The above is the only content change in this file. The rest is formatting of code per style guide.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -40,17 +40,17 @@ import org.apache.daffodil.processors.RepValueSet
...
   with TypeChecks {
 
   Assert.invariant(xmlArg.asInstanceOf[scala.xml.Elem].label == "restriction")
 
-  def restriction(simpleType: SimpleTypeDefBase) = {
-    new Restriction(simpleType, this)
-  }
+  //  def restriction(simpleType: SimpleTypeDefBase) = {
+  //    new Restriction(simpleTypeDef, this)
+  //  }
```
mbeckerle: delete commented out code.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -69,7 +70,7 @@ trait SchemaComponent
       schemaFileLocation,
       tunable)
 
-  val context: SchemaComponent = parent
+  val context: SchemaComponent = lexicalParent
```
mbeckerle: Rename to lexicalContext. Equally weak name as "parent".

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -402,10 +408,11 @@ final class SchemaSet(
...
   }
-  def getPrimitiveTypeFactory(refQName: RefQName) = {
-    val primType = getPrimitiveType(refQName)
-    primType.map(new PrimitiveSimpleTypeFactory(_, schemaDocument))
-  }
+
+  //  def getPrimitiveTypeFactory(refQName: RefQName) = {
+  //    val primType = getPrimitiveType(refQName)
+  //    primType.map(new PrimitiveSimpleTypeFactory(_, schemaDocument))
+  //  }
```
mbeckerle: delete

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -61,12 +61,12 @@ trait TypeBase {
...
 
-  def elementBase: ElementBase = elementDecl match {
-    case eb: ElementBase => eb
-    case ged: GlobalElementDecl => ged.elementRef
-  }
+  //  def elementBase: ElementBase = elementDecl match {
+  //    case eb: ElementBase => eb
+  //    case ged: GlobalElementDecl => ged.elementRef
+  //  }
 }
```
mbeckerle: delete commented out code in this file.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -296,62 +286,40 @@ abstract class SimpleTypeDefBase(xml: Node, parent: SchemaComponent, val factory
...
-  def name: String
 }
 
-abstract class SimpleTypeDefFactory(xmlArg: Node, schemaDocumentArg: SchemaDocument)
-  extends SchemaComponentFactory(xmlArg, schemaDocumentArg)
-  with HasRepValueAttributes
-  with SimpleTypeFactory
-  with HasOptRepTypeMixinImpl
-  with ProvidesDFDLStatementMixin {
+trait SimpleTypeDefMixin extends TypeBase { self: SimpleTypeDefBase =>
```
mbeckerle: This mixin is used exactly one place, so probably should just be added to SimpleTypeDefBase and get rid of the trait. I left it this way because it moved fewer lines of code around, allowing the diff to be more of a diff, and less of a wholesale green lines for things that haven't changed. 

I will integrate this into the class that uses it post review.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementCombinator1.scala
```diff
@@ -223,8 +223,7 @@ class ElementParser(
   eBeforeParser: Maybe[Parser],
```
mbeckerle: All whitespace change. Revert.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ReptypeMixins.scala
```diff
@@ -28,23 +28,24 @@ import org.apache.daffodil.processors.RepValueSet
...
 }
 
 trait HasOptRepTypeMixin {
-  def optRepTypeFactory: Option[SimpleTypeFactory with NamedMixin]
-  
-  lazy val optRepTypeDefFactory: Option[SimpleTypeDefFactory with NamedMixin] = optRepTypeFactory match {
-    case Some(x:SimpleTypeDefFactory) => Some(x)
+  def optRepType: Option[SimpleTypeBase]
+
+  final lazy val optRepTypeDef: Option[SimpleTypeDefBase] = optRepType match {
```
bsloane1650: Not related to this change, but it might be worth adding a comment explaining why we need to distinguish between optRepType and optRepTypeDef. I had to scratch my head at this for a few minutes, and I'm the one who wrote this in the first place.

I believe the distinction is that it is possible when the repType is a primitive type. The resulting type would be difficult (maybe impossible?) to use as the type for an element, as the repType is missing any annotations that tell Daffodil what the physical characteristics are; however the type can be used purely to define a function for use by DPath expressions

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -238,57 +239,21 @@ final class RestrictionFactory(val xmlArg: Node, val simpleTypeFactory: SimpleTy
...
+  with NestingLexicalMixin {
   Assert.invariant(xmlArg.asInstanceOf[scala.xml.Elem].label == "union")
-  
-  def union(simpleType: SimpleTypeDefBase) = {
-    new Union(simpleType, this)
-  }
+
+  //  def union(simpleType: SimpleTypeDefBase) = {
+  //    new Union(simpleType, this)
+  //  }
```
bsloane1650: Delete commented code

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -54,11 +54,12 @@ trait SchemaComponent
   with PropTypes {
 
   def xml: Node
-  def parent: SchemaComponent
+  def lexicalParent: SchemaComponent
+  final lazy val optLexicalParent = Option(lexicalParent)
 
-  final override def oolagContextViaArgs = Some(parent)
+  override def oolagContextViaArgs = Option(lexicalParent)
```
bsloane1650: Is there a reason for using Option() instead of Some() here?

I believe the only functional difference is that Option(null) == None. If that is important here, it should be commented.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -169,36 +156,19 @@ trait SchemaComponent
...
+  final lazy val slashPath: String =
+    scPath.filter { _.isInstanceOf[ElementBase] }.map { _.diagnosticDebugName }.mkString("/")
 
   override def toString = diagnosticDebugName
 
   /**
-   * Includes instances. Ie., a global element will appear inside an element ref.
-   * a global group inside a group ref, a global type inside an element or for
-   * derived simple types inside another simple type, etc.
+   * Does not include instances. Ie., walks up lexical nest only.
```
bsloane1650: Does this mean that diagnostic messages will loose the ability to tell the non-lexical location of errors?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -586,37 +539,29 @@ abstract class SimpleTypeDefFactory(xmlArg: Node, schemaDocumentArg: SchemaDocum
...
+//
+//  override lazy val primType = primitiveType.primType
+//  override lazy val schemaDocument = schemaDocumentArg
+//
+//  /*
+//   * These methods don't really make sense here, but are needed by NamedMixin
+//   */
+//  override def SDE(id: String, args: Any*) = Assert.invariantFailed("Primitive types shouldn't ever have an SDE")
+//  override def schemaFileLocation = Assert.invariantFailed("Primitive types don't have a schemaFileLocation")
+//}
```
bsloane1650: Delete

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: Better yet. Just remove this, and change uses.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -250,8 +250,8 @@ abstract class ModelGroup(index: Int)
             case eb: ElementBase => !eb.isRequiredInInfoset || !eb.isRepresented
           }
           if (lastIsOptional) {
-            val (priorSibs, parent) = last.potentialPriorTerms
-            (last +: priorSibs, parent.isDefined)
+            val (priorSibs, optPriorParent) = last.potentialPriorTerms
```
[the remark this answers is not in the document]
mbeckerle: More like optPriorElementsIncludesThisParent - it's an indicator that one of the things that can be immediately prior is the parent element, i..e., this model group *could* be first child term of the parent.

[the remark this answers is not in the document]
mbeckerle: Unfortunately, this is not the lexical parent. This is still code depending on backrefs to dynamic enclosing parents.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: Yup. Will add comment. That's exactly the behavior we seek.

mbeckerle: I will be pushing a new PR later that gets rid of the null/non-null for parents entirely. I'm sick of getting NPEs.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: Yes. We will have to use the processor's node-stack at runtime to provide the dynamic nest, as we will only have the lexical nest at schema compile time.  We may be able to do better once I have the back reference table in that can provide all the context backpointers where something is being reused. There may be exactly one usage, in which case it is unambiguous.

mbeckerle: Comment added to code to this effect.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -62,22 +63,25 @@ import scala.collection.immutable.HashMap
...
   checkAllTopLevelArg: Boolean,
-  parent: SchemaComponent,
   tunableArg: DaffodilTunables)
-  extends SchemaComponentImpl(<schemaSet/>, parent) // a fake schema component
+  extends SchemaComponentImpl(<schemaSet/>, null)
   with SchemaSetIncludesAndImportsMixin {
 
+  private lazy val processorFactory = pfArg
+  lazy val root = rootElement(processorFactory.flatMap { _.rootSpec })
+
```
stevedlawrence: What's the intention for the pfArg being passed in and the creation of the root variable? From what I can tell, root isn't used.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -577,17 +577,23 @@ trait ElementBaseGrammarMixin
...
-      this.asInstanceOf[PrefixLengthQuasiElementDecl].detachedReference.impliedRepresentation == Representation.Text)
-      // If an element has text representation and has a prefixed length, it
+    else if (this.isInstanceOf[PrefixLengthQuasiElementDecl]) {
+      //
+      // This predicate used to check the detachedReference for representation text.
+      // Seems like a mistake. It's just about variable length here.
+      // You can have prefixed length on hexBinary.
+      //
+      // && this.asInstanceOf[PrefixLengthQuasiElementDecl].detachedReference.impliedRepresentation == Representation.Text)
+      //
```
stevedlawrence: Agreed, I think this is wrong, I'm not sure keeping this comment around adds anything. I think it's just confusing.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -315,9 +313,8 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
    */
   def compileSource(schemaSource: DaffodilSchemaSource): ProcessorFactory =
     Compiler.synchronized {
-      val noParent = null // null indicates this is the root, and has no parent
-      val sset = new SchemaSet(rootSpec, externalDFDLVariables, Seq(schemaSource), validateDFDLSchemas, checkAllTopLevel, noParent, tunablesObj)
-      val pf = new ProcessorFactory(sset)
+      lazy val sset: SchemaSet = new SchemaSet(Some(pf), rootSpec, externalDFDLVariables, Seq(schemaSource), validateDFDLSchemas, checkAllTopLevel, tunablesObj)
+      lazy val pf: ProcessorFactory = new ProcessorFactory(sset)
       val err = pf.isError
```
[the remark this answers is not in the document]
mbeckerle: The two objects, schemaSet, and PF, probably should just be merged at some point. But right now the SchemaSet gets responsibility for combining the various ways that the root element can be specified - which is explicitly on the compiler, explicitly on the PF, or implicitly by way of the schema's first schema document's first element.  The PF creates the SchemaSet, but the logic for root determination is in the SchemaSet.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
mbeckerle: root is used many places where optimizers ask about global properties of the schema's root element., e..g, schemaSet.root.isScannable tells us if the whole schema (for this root anyway) is all scannable text.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mbeckerle: I'll take out this comment.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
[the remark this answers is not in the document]
mbeckerle: I will add comments to the effect that these two objects are mutually referencing each other, and this is the idiom in scala for creating a circular reference chain without having to have a side-effecting assignment that may or may not have happened by the time you reference it.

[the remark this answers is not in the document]
stevedlawrence: Ah, I don't think i've seen the mutual reference pattern before. Good to know about. Makes sense to me. A comment would definitely be useful.

stevedlawrence:
+1

PR #224, opened 2019-06-03
Generate DaffodilTunables and WarnID from dafext.xsd

stevedlawrence:
- Add Tuanble and WarnID generators to generate code based on
  dafext.xsd. This makes adding new warnings or tunables much easier,
  and the schema acts as documentation.
- Modifies the parseUnparsePolicy tunable to add a "fromRoot" value,
  which is the new default. The old default of empty string is no longer
  supported.
- Modifies the ParseUnparsePolicyTunable and UnqualifiedPathStepPolicy
  enums to be props.Enums, which has improved enum functionality and
  allows for consistent logic in the generator.
- Cleaned up the dafext.xsd schema, added tunable documentation and
  default values

DAFFODIL-2117


daffodil-lib/src/main/scala/org/apache/daffodil/api/ParseUnparsePolicyTunable.scala
```diff
@@ -0,0 +1,49 @@
...
+ *   Compile for only unparsing
+ *
+ * Both:
+ *   Compile for both parsing and unparsing
+ *
+ * FromRoot:
+ *   Use the dfdl:parseUnparsePolicy from the root element to determine how to
+ *   compile the schema
+ */
+sealed trait ParseUnparsePolicyTunable extends ParseUnparsePolicyTunable.Value
```
mbeckerle: We could generate this trait also yes?  Or is there some reason this and the next one UnqualifiedPathStepPolicy need special treatment? 

If there is a reason, please add comments to code to that effect.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -0,0 +1,223 @@
...
+
+  def writeGeneratedCode(w: java.io.FileWriter) {
+    val tunables =
+      tunableNodes.map { tunableNode =>
+        val schemaName = tunableNode \@ "name"
+        val schemaType = tunableNode \@ "type"
+        val schemaDefault = tunableNode \@ "default"
+
+        schemaName match {
+          case "defaultEmptyElementParsePolicy" =>
```
mbeckerle: I am uncertain as to why this case statement? This will require editing for every new property we add.  If this is unavoidable (or unavoidable without additional work you didn't plan to do currently), then a comment to the effect that this has to be edited as well as the dafext.xsd is in order.

daffodil-lib/src/main/scala/org/apache/daffodil/api/ParseUnparsePolicyTunable.scala
stevedlawrence: Yeah, pretty sure this can be autogenerated too. Will add.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
stevedlawrence: It's only needed right now for tunables that aren't primitives (i.e. enums or lists of enums). But I guess anything that isn't a primitive can be assumed to be an enum. And then we just need the special caase for suppressSchemaDefinitionWarnings. I"ll make this change.

PR #225, opened 2019-06-03
Implement daf:lookAhead

bsloane1650:
DAFFODIL-2140


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1579,13 +1585,19 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+
       case (RefQName(_, "logicalTypeValueString", DFDL), args) =>
         FNZeroArgExpr(functionQNameString, functionQName,
           NodeInfo.String, NodeInfo.AnyAtomic, DFDLXLogicalTypeValueString(_, _))
 
       //End typeValueCalc related functions
 
+      case (RefQName(_, "lookAhead", DAF_APACHE), args) =>
+        FNTwoArgsExpr(functionQNameString, functionQName, args,
+          NodeInfo.NonNegativeInteger, NodeInfo.UnsignedInt, NodeInfo.UnsignedInt, DFDLXLookAhead(_))
```
stevedlawrence: I don't see DFDLXLookAhead defined anywhere. Did you forget to add a file?

bsloane1650: Fixed. I also took the opportunity to rename in to DAFLookAhead.

@mbeckerle  @jadams-tresys  You might want to re-review since I was missing the guts of the PR.

bsloane1650:
@mbeckerle I don't think we can merge DState with P/UState. We need to have a (degenerate) DState without a P/UState for constant folding. I also think it makes sense to keep them seperate just as an organizational matter, as our DPath interperatetor is relatively isolated from our main DFDL interperator. A back pointer is probably the way to go.

Having said that, it might make sense to merge a bunch of the setters. Now that we are passing in a pointer to the entire state, we might want to stop passing in pointers to indiviual fields of the state, and replace them with functions in DState that refer to the main state object.

mbeckerle:
w.r.t the issue of merging DState and P/UState I see your point that it's better to keep them separate, but we need a trait that defines the dependencies of DState on P/UState if the separation is going to be worth much. Would probably make DState more testable in isolation also.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
```diff
@@ -72,3 +78,63 @@ case object DAFError extends RecipeOp {
...
+    /*
+     * Since daf:lookAhead is defined to take unsigned arguements, the DPath interperater
+     * will error out on the cast if a negative arguement is supplied, so we do not need to SDE here.
+     */
+
+    Assert.invariant(offset >= 0)
+    Assert.invariant(lBitSize >= 0)
+
+    val totalLookahead = offset + lBitSize
+    if (totalLookahead > 512) {
```
stevedlawrence: This should be a constant variable, or maybe even a tunable? I don't see anything code that actually has a limit of 512, so it seems this is just to prevent people from accidentally doing the wrong thing or abusing this function. A tunable probably make sense in that case, so if someone reeeeeally wants to abuse this they can.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
```diff
@@ -72,3 +78,63 @@ case object DAFError extends RecipeOp {
...
+    }
+    //Safe since we guard on totalLookahead
+    val bitSize = lBitSize.toInt
+
+    if (!dstate.parseOrUnparseState.isDefined) {
+      Assert.invariant(dstate.isCompile)
+      /*
+        * This is an expected code path.
+        * Throwing an exception is how we indicated that this expression does
+        * cannot be reduced to a constant at compile time.
```
stevedlawrence: typo in comment "does cannot".

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
```diff
@@ -72,3 +78,63 @@ case object DAFError extends RecipeOp {
...
+      }
+      val mark = dis.markPos
+      dis.skip(offset, pstate)
+      val ans: AnyRef = if (bitSize > 63) {
+        dis.getUnsignedBigInt(bitSize, pstate)
+      } else if (bitSize == 0) {
+        Long.box(0)
+      } else {
+        Long.box(dis.getUnsignedLong(bitSize, pstate).longValue)
+      }
```
stevedlawrence: Any particular reason for the explicit Long.box() calls? I don't think we do that anywhere else. Also, should it be a JLong instead? I *think* we tend to use java types in DPath.

daffodil-test/src/test/scala/org/apache/daffodil/extensions/TestLookAhead.scala
```diff
@@ -0,0 +1,33 @@
+package org.apache.daffodil.extensions
```
stevedlawrence: Missing apache license is causing the Rat check to fail.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
```diff
@@ -72,3 +78,63 @@ case object DAFError extends RecipeOp {
...
+      dis.resetPos(mark)
+      ans
+    } else {
+      Assert.invariant(dstate.parseOrUnparseState.get.isInstanceOf[UState])
+      dstate.SDE("Cannot call daf:lookAhead() during unparse")
+    }
```
stevedlawrence: Does anyone know if there's information in the compilation process about what property an expression is being compiled for? It would be nice to throw this SDE at schema compilation time if it' ever used in a expression that could be evaluated at unparse, instead of only evaluating at runtime. I would think we would have that information, but I can't find it, so maybe not.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
bsloane1650: The explicit boxing is due to limitations in Scala's type inference (well justified in this case).

The core issue is that all DPath values are stored by Daffodil as AnyRef. When we attempt to cast an unboxed long (or integer literal) to AnyRef, Scala refused to do an automatic conversion.

I believe you are correct on the Long/JLong distinction, I am swiching these over to JLong

For reference, the following two snippets are equivalent and both should work:

    new JLong(0)
    0 : JLong

I tend to prefer the former. As long as we need to make the boxing explicit, we might as well go all the way to explicitly boxing, instead of coercing the type system to convince it to box for us.

I think the what we would normally do in this type of situation is let the ans variable take on the JLong type and let Scala handle the boxing for us. This doesn't work here since ans can also store a BigInt.

On a related note, it would be nice if we could formalize the fact that DPath values are one of a limited number of types. Ideally make it an untagged union type if Scala has a way of doing so. Passing around AnyRefs like this seems to be asking for bugs.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
bsloane1650: My main motivation for this limit was to be nice to future implementations where infinite lookahead might be less trivial. An added benifit is that it prevents people from shooting themselves in the foot with a long lookahead on streamed input.

Adding the tunable maxDAFLookaheadBits

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DAFFunctions.scala
mbeckerle: I don't know of a way in Scala 2.x to create a untagged union type. You can create type that works kind of like Maybe types that will, at runtime, restrict the type of the argument to one of several kiinds, and you can have overloaded constructors only for those types, so as to get a compile time error.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilTunables.scala
```diff
@@ -193,7 +193,13 @@ case class DaffodilTunables(
...
    * type.
    */
-  val allowExpressionResultCoercion: Boolean = true)
+  val allowExpressionResultCoercion: Boolean = true,
+  /*
+   * Max distance the daf:lookahead() function is permitted to look.
+   * Distance is defined by the distance to the last bit accessed,
+   * and so is offset + bitsize
+   */
+  val maxDAFLookaheadBits: Long = 512)
```
stevedlawrence: You might want to rebase this on master. A change was just merged that autogenerates tunables from dafext.xsd.

I'd also reccomend changing it to not mention DAF? The namespace of this will change in a future commit to DFDLX, and if it becomes part of the standard it will be moved to the DFDL namespace, so this tunable name might make sense, but we wouldn't want to change it. Maybe rename to maxLookaheadFunctionBits or maxLookaheadExpressionBits or somethign a little more generic?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -364,6 +364,15 @@
...
         </xs:element>
+        <xs:element name="maxLookaheadFunctionBits" type="xs:long" default="512" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              Max distance that the DPath lookahead function is permitted to look.
+              Distance is defined by the distance to the last bit accessed, and
+              so it is offset+bitsize.
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
```
stevedlawrence: All of the tuanbles in dafext.xsd is sorted alphabetically. Can you move this one to the right spot? Then please squash, force commit, and then we can merge this. Thanks! :+1:

bsloane1650: Fixed and rebased

PR #226, opened 2019-06-04
Sick of NPE. lexicalParent is now optLexicalParent and is Option type

mbeckerle:
This small change is another incremental improvement. I was getting NPE quite a bit due to things running up the DSOM object taxonomy looking for the root which has lexicalParent pointer of null.

I've converted this to an Option type. So you can still get bad behavior of course, it's just that you get a bit more type-safety in the code.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponentFactory.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponentFactory.scala
@@ -31,15 +31,17 @@ import scala.xml.Node
  */
 abstract class SchemaComponentFactory(
   final override val xml: Node,
-  final override val lexicalParent: SchemaComponent)
+  final override val optLexicalParent: Option[SchemaComponent])
   extends SchemaComponent
   with NestingLexicalMixin {
 
+  def this(xml: Node, lexicalParent: SchemaComponent) =
+    this(xml, Option(lexicalParent))
 }
 
 abstract class AnnotatedSchemaComponentFactory(
   final override val xml: Node,
-  final override val lexicalParent: SchemaComponent)
+  final override val optLexicalParent: Option[SchemaComponent])
   extends AnnotatedSchemaComponent
   with NestingLexicalMixin {
 
@@ -50,7 +52,7 @@ trait SchemaFileLocatableImpl
 
   def xml: scala.xml.Node
   def schemaFile: Option[DFDLSchemaFile]
-  def lexicalParent: SchemaComponent
+  def optLexicalParent: Option[SchemaComponent]
 
   /**
    * Annotations can contain expressions, so we need to be able to compile them.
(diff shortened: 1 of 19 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponentFactory.scala
```diff
@@ -79,14 +81,14 @@ trait SchemaFileLocatableImpl
...
-  def lexicalParent: SchemaComponent
+  def optLexicalParent: Option[SchemaComponent]
 
-  lazy val schemaFile: Option[DFDLSchemaFile] = lexicalParent.schemaFile
-  lazy val schemaSet: SchemaSet = lexicalParent.schemaSet
-  def schemaDocument: SchemaDocument = lexicalParent.schemaDocument
-  lazy val xmlSchemaDocument: XMLSchemaDocument = lexicalParent.xmlSchemaDocument
-  lazy val schema: Schema = lexicalParent.schema
-  def uriString: String = lexicalParent.uriString
+  lazy val schemaFile: Option[DFDLSchemaFile] = optLexicalParent.flatMap { _.schemaFile }
```
stevedlawrence: Should this be ``.get.schemaFile`` like the onces below it? Seems there's an assumption in this class that lexicalParent isDefined.

mbeckerle: Well this one really should be named optSchemaFile. The code is not very consistent about this naming convention though. 

It uses flatMap because it wants an option type back. The ones below are non-option, so they fault if used on an object without lexical parent (which is only the root).

mbeckerle: We do have a mixture here of non-opt-handling, and option code. There is rhyme and reason to this. For the most part, oolag framework lets you just right code that assumes certain things are present, and if they aren't captures it and you get an SDE out of it, ascribed to the nearest calling OOLAG LV object.  So the code doesn't have to always deal with option values. On the other hand, quite often you need to know if something is present or not, as the logic simply requires it. For this optLexicalParent, for the most part code just assumes everything has a lexical parent. The point of adding it, is just to catch algorithm problems where you run up the hierarchy unconditionally, and aren't checking for the root. This happened enough times, and NPEs are problematic enough in the debugger that I had to change this.

stevedlawrence: I was just saying this more for consistency. schemaFile is already an option, so ``optLexicalParent.get.schemaFile`` should return the that value option value. The difference between that and what you have is that if optLexicalParent isn't defined, schemaFile will be set to None rather than throwing a NoSuchElementException. But the code below already assumes optLexicalParent is defined, so the result should be exactly the same.

I guess the thing that confused me was was this line change was different than the ones below and I had to think about if there was a reason for that, and it wasn't clear to me if there was a reason or not. But returning an Option is reasonable. This is fine with me.

PR #228, opened 2019-06-07
Incremental progress on Daffodil 1444 schema compiler space/speed issue.

mbeckerle:
Has features to allow navigation of multiple backpointers from referenced components to the places referencing them.

This enables sharing in the future once the code that is depending on non-shared schema components is moved.

Also deprecates the unitary backpointers that prevent sharing, but does not remove them.

Tests demonstrate exponential growth of number of schema components being compiled. So this sharing improvement is critical.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementDeclMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementDeclMixin.scala
@@ -27,15 +27,10 @@ trait ElementLikeMixin
   extends AnnotatedSchemaComponent
   with ProvidesDFDLStatementMixin {
 
-  private lazy val elementBase = this match {
-    case ged: GlobalElementDecl => ged.elementRef
-    case eb: ElementBase => eb
-  }
-
   protected final def annotationFactory(node: Node): Option[DFDLAnnotation] = {
     node match {
       case <dfdl:element>{ contents @ _* }</dfdl:element> => Some(new DFDLElement(node, this))
-      case _ => elementBase.annotationFactoryForDFDLStatement(node, this)
+      case _ => annotationFactoryForDFDLStatement(node, this)
     }
   }
 
@@ -46,25 +41,81 @@ trait ElementLikeMixin
 /**
  * Shared by all element declarations local or global
  */
-trait ElementDeclMixin
+sealed trait ElementDeclMixin
   extends ElementLikeMixin {
 
   def namedQName: NamedQName
 
-  final override protected def optReferredToComponent = typeDef match {
+  def optReferredToComponent: Option[AnnotatedSchemaComponent]
+
+  def optImmediateSimpleType: Option[SimpleTypeBase]
+
+  def optImmediateComplexType: Option[ComplexTypeBase]
+
+  def immediateType: Option[TypeBase]
+
+  def typeName: Option[String]
+
+  def optNamedSimpleType: Option[SimpleTypeBase]
+
+  def optNamedComplexType: Option[GlobalComplexTypeDef]
+
+  def optSimpleType: Option[SimpleTypeBase]
+
+  def optComplexType: Option[ComplexTypeBase]
+
+  def namedTypeQName: Option[RefQName]
+
+  def namedType: Option[TypeBase]
+
+  def typeDef: TypeBase
+
+  final def isSimpleType: Boolean = optSimpleType.isDefined
+
+  final def isComplexType = !isSimpleType
+
+  def defaultAttr: Option[Seq[Node]]
+
+  def defaultValueAsString: String
+
+  final def primType = optSimpleType.get.primType
+
+  final def hasDefaultValue: Boolean = defaultAttr.isDefined
+
+  def isNillable: Boolean
+
+  final def simpleType: SimpleTypeBase = optSimpleType.get
+
+  final def complexType: ComplexTypeBase = optComplexType.get
+
+  /**
+   * Convenience methods for unit testing purposes.
+   */
+  final def sequence = complexType.sequence
+  final def choice = complexType.choice
+}
+
+sealed trait ElementDeclInstanceImplMixin
+  extends ElementDeclMixin {
+
+  final override lazy val optReferredToComponent = typeDef match {
     case std: SimpleTypeDefBase => Some(std)
     case ctd: ComplexTypeBase => None // in DFDL v1.0 complex types are not annotated, so can't carry properties nor statements.
     case _ => None
   }
 
-  final def immediateType: Option[TypeBase] = LV('immediateType) {
-    val st = xml \ "simpleType"
+  final override lazy val optNamedComplexType: Option[GlobalComplexTypeDef] = {
+    namedTypeQName.flatMap { qn =>
+      val gctdFactory = schemaSet.getGlobalComplexTypeDef(qn)
+      val res = gctdFactory.map { gctdf => gctdf.forElement(this) }
+      res
+    }
+  }
+
+  final override lazy val optImmediateComplexType: Option[ComplexTypeBase] = LV('optImmediateComplexType) {
     val ct = xml \ "complexType"
     val nt = typeName
-    if (st.length == 1) {
-      val lstd = new LocalSimpleTypeDef(st(0), this)
-      Some(lstd)
-    } else if (ct.length == 1)
+    if (ct.length == 1)
       Some(new LocalComplexTypeDef(ct(0), this))
     else {
       Assert.invariant(nt != "")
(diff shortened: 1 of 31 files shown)


mbeckerle:
A thought. This change introduces 41 deprecation errors. I had to switch off the checking that fails on warnings when compiling scala code. 

 Do we want a dev branch for this until those are resolved? (It will not be quick to resolve them all.)

build.sbt
```diff
@@ -110,7 +110,7 @@ lazy val commonSettings = Seq(
     "-deprecation",
     "-language:experimental.macros",
     "-unchecked",
-    "-Xfatal-warnings",
+    // "-Xfatal-warnings", // temporarily comment out because of deprecation of backpointers
```
mbeckerle: There are 41 deprecation warnings currently

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -143,16 +144,65 @@ trait SchemaComponent
...
+      case _: SchemaDocument => false
+      case _: XMLSchemaDocument => false
+      case _: DFDLSchemaFile => false
+      case _: Schema => false
+      case _: SchemaSet => false
+      case _ => true
+    }
+  }
+
+  lazy val sscd: String = {
```
mbeckerle: rename, or scaladoc. SSCD is ShortSchemaComponentDesignator

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponentFactory.scala
```diff
@@ -39,13 +51,13 @@ abstract class SchemaComponentFactory(
...
 }
 
-abstract class AnnotatedSchemaComponentFactory(
-  final override val xml: Node,
-  final override val optLexicalParent: Option[SchemaComponent])
-  extends AnnotatedSchemaComponent
-  with NestingLexicalMixin {
-
-}
+//abstract class AnnotatedSchemaComponentFactory(
```
mbeckerle: delete

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Root.scala
```diff
@@ -45,4 +47,111 @@ final class Root(defXML: Node, parentArg: SchemaDocument,
...
+    m.toMap
+  }
+
+  lazy val refPairsMap: Map[SchemaComponentFactory, Seq[String]] = {
+    refMap.toSeq.map {
+      case (to, seq: Seq[(String, _)]) => (to, seq.map { case (sscd, _) => sscd }.toSeq)
+    }.toMap
+  }
+
+  //  lazy val refMapLexicalPairs = refMap.toSeq.map{
```
mbeckerle: delete

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Root.scala
```diff
@@ -45,4 +47,111 @@ final class Root(defXML: Node, parentArg: SchemaDocument,
...
+      case e: ElementBase if (e.optComplexType.isDefined && e.complexType.isInstanceOf[GlobalComplexTypeDef]) => e.complexType
+    }
+    val ctsIDs = cts.map { _.sscd }.distinct
+    ctsIDs
+  }
+}
+
+case class RefSpec(from: SchemaComponent, to: SchemaComponentFactory, index: Int) {
+
+  override def toString = "RefSpec(" +
```
mbeckerle: really undesirable toString method.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
```diff
@@ -0,0 +1,651 @@
...
+    val root = sset.root
+    val comps = root.allComponents
+    assertEquals(509, comps.length)
+    val refMap = root.refMap
+    val numEntries = refMap.size
+    assertEquals(7, numEntries)
+    val refPairsMap = root.refPairsMap.toSeq
+    val numRefPairs = refPairsMap.map { _._2.length }.sum
+    assertEquals(13, numRefPairs)
+    assertEquals(509, root.numComponents)
```
mbeckerle: If you look at the last 2 lines of most of these tests. They show the growth of components vs. unique components.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
```diff
@@ -46,8 +46,6 @@ trait InitiatedTerminatedMixin
    */
   lazy val hasInitiator = {
     val hasOne = initiatorExpr.isKnownNonEmpty
-    if (parentSaysInitiatedContent)
-      schemaDefinitionUnless(hasOne, "Enclosing group has initiatedContent='yes', but initiator is not defined.")
     hasOne
```
stevedlawrence: I was hoping we could completely remove parentSaysInitiatedContent, but there's one place where it's still used:
```scala
lazy val initiatorDiscriminator = prod("initiatorDiscriminator", parentSaysInitiatedContent) { InitiatedContent(this) }
```
Any idea how we'll eventually remove thsi, since whether or not ``InitiatedContent`` is laid down really does depend on information about the parent. For example:
```xml
<xs:element name="foo" dfdl:initiator="foo" type="xs:string" />

<xs:element name="root">
  <xs:complexType>
    <xs:sequence dfdl:initiatedContent="yes">
      <xs:element ref="foo" />
    </xs:sequence>
    <xs:sequence dfdl:initiatedContent="no">
      <xs:element ref="foo" />
    </xs:sequence>
  </xs:complexType>
</xs:element>
```
In this case, a different grammar really does need to be laid down based on the context of where the foo element is reference. Any thoughts into how we'll handle something like this? Will the sequence pass down information to it's children so they have that content? Or the foo element have a reference to both enclosing sequences, and it will be passed an index to know which one is the current context or something?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Nesting.scala
```diff
@@ -54,24 +61,61 @@ trait NestingMixin {
  * schema components.
  */
 trait NestingLexicalMixin
-  extends NestingMixin {
+  extends NestingMixin { self: SchemaComponent =>
 
+  @deprecated("2019-06-03", "Use enclosingComponentDefs method, and deal with sharing.")
   override protected def enclosingComponentDef = optLexicalParent
 
+  final override protected lazy val enclosingComponentDefs: Seq[(SchemaComponent, Int)] =
```
stevedlawrence: Thoughts on creating a new type instead of using a tuple here? Might make things a bit more readable, something like
```scala
case class EnclosingComponentDef(lexicalParent: SchemaComponent, lexicalPosition: Int)
```
Not sure if there's a benefit or not, but might help readablility, especially once everything gets rewritten to using enclosingComponentDefs.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Nesting.scala
```diff
@@ -54,24 +61,61 @@ trait NestingMixin {
...
+    this match {
+      case hf: HasFactory => {
+        schemaSet.root.refMap.get(hf.factory) match {
+          case None => Assert.invariantFailed("There are no references to this component: " + this)
+          case Some(seq) => seq
+        }
+      }
+      case _ => Assert.invariantFailed("Should only be part of Global objects with factories.")
+    }
+  }
```
stevedlawrence: Does it make sense/is it possible for this function live on HasFactory if it's an invariantFailed if it's called on anything else?

stevedlawrence:
> A thought. This change introduces 41 deprecation errors. I had to switch off the checking that fails on warnings when compiling scala code.
> 
> Do we want a dev branch for this until those are resolved? (It will not be quick to resolve them all.)

I don't think a separate dev branch is necessary at the momemt. As long as things aren't broken, I'd prefer to get changes in master as soon as possible to minimize issues dealing with merge conflicts in the future and get people used to the new stuff.

I'd suggest either
1. Turn off ``-Xfatal-warnings`` as you did
2. Turn off ``-deprecation`` until the deprecated functions are removed. People trying to remove these deprecated functions can turn it back on during development to find the uses of deprecated functions.
3. Just don't deprecate those functions at all and just work to eventually remove them. We can make sure during reviews that new code added doesn't use them.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Nesting.scala
mbeckerle: I will incorporate this idea.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
mbeckerle: I'm pretty sure to fix this it has to happen in sequence child (or choice child or a base they both share). In generating the grammar or parser/unparser for that sequence child, that's where all the context-specific logic belongs. At that point the context is clear, and can be passed directly to anything that needs to make a decision based on it. The element ref or local element decl or child-term itself is then free of context-specific computations, and just provides information to the surrounding context.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementDeclMixin.scala
```diff
@@ -119,50 +149,83 @@ trait ElementDeclMixin
...
-      case _ => Assert.invariantFailed("Must be either SimpleType or ComplexType")
-    }
+}
+
+trait ElementDeclFactoryImplMixin
+  extends ElementDeclMixin {
+
+  final override lazy val optImmediateSimpleType: Option[SimpleTypeBase] = LV('optImmediateSimpleType) {
+    val st = xml \ "simpleType"
+    val nt = typeName
```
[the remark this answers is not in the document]
mbeckerle: Just an error as I was messing with this code. I'll remove this.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
mbeckerle: I also will set out to remove this particular context parent reference, by way of showing the new way things have to be done as code evolves to not use context pointers.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Nesting.scala
mbeckerle: I will incorporate this idea. I introduced HasFactory pretty late in this, but looking back at it, really there should be a harder separation between regular schema components and factory schema components. I'm disinclined to work too hard separating them though, since the whole objective is to eliminate the context pointers that force us to distinguish, entirely, thereby recombining the factory and instance. What really is happening is that the DSOM objects become pure factory, and the Grammar objects beccome the "instance" objects.

mbeckerle: I removed HasFactory, as it was subsumed by NestingTraversesToReferenceMixin.

mbeckerle:
One must turn off both deprecation warnings and Xfatal-warnings to successfully compile. Because even if you turn off deprecation warnings, you still get summary warnings that there were 41 deprecation warnings total, and with Xfatal-warnings, those stop the compilation from succeeding.

mbeckerle:
I've concluded the right thing is to remove as many deprecated calls as possible (some calls were from unused/dead code), and then comment out the deprecate annotations for now so we're back to zero warnings. These can be commented back in to pursue this particular thread of improvements.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
mbeckerle: I have concluded that this change is bigger and should be done in a separate change set and review. So not including this change as yet.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -252,14 +252,14 @@ sealed trait DITerm {
...
-    pre
-  }
+  //  protected final def dafPrefix = {
+  //    val ee = trd.dpathCompileInfo.immediateEnclosingCompileInfo.getOrElse {
+  //      trd.dpathCompileInfo
+  //    }
+  //    val pre = ee.namespaces.getPrefix(XMLUtils.dafintURI.uri.toString())
+  //    Assert.invariant(pre ne null)
+  //    pre
+  //  }
```
stevedlawrence: Remove commented code?

build.sbt
```diff
@@ -110,7 +110,7 @@ lazy val commonSettings = Seq(
     "-deprecation",
     "-language:experimental.macros",
     "-unchecked",
-    "-Xfatal-warnings",
+    "-Xfatal-warnings",
```
jadams-tresys: Remove extra whitespace

PR #230, opened 2019-06-07
Enable schema validation of attribute- and short-form properties

stevedlawrence:
- Use xs:anyAttribute and xs:any defined in the DFDL namespace to enable
  validation for attribute-form and short-form properties. Other
  namespaces use lax validation and will mostly likely just be ignored,
  but could still be validated if the other namespace schema defines the
  dfdl namespace and is a valid DFDL schema.
- Fix XML Schema for DFDL to allow for repValues in xs:enumerations,
  layering properties in xs:sequences, and short form dfdl:bitOrder.
- Fix tests that used incorrect properties or had typos that are now
  checked.

DAFFODIL-598, DAFFODIL-1024

### DIFF
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
@@ -556,51 +556,25 @@ this to validate when we load a DFDL Schema.
 
   <xsd:attributeGroup name="anyOther">
     <!--
-      
-      IBM's DFDL Schemas have ibm-specific attributes on them that we 
-      need to tolerate. Users may want to decorate their DFDL schemas with
-      other non-native and non-DFDL attributes and annotation sub-elements.
-      
-      So the behavior here has been switched back to the XSD Default behavior
-      which is lax validation allowing any other attribute to appear.
-      
-      The problem: this disables ability to catch certain typographical errors.
-      In fact a bunch of our negative tests will have to be fixed, as they are 
-      counting on errors from validation of these typos and mistakes and 
-      that sort of thing.
-      
-      A common error is to forget the dfdl: prefix on 
-      a short form dfdl property binding, or to include the dfdl: prefix
-      one on a long-form format annotation. This will no longer be caught by 
-      the validator. 
-      
-      We will have to build explicit checking for this into Daffodil's traversal
-      of the schema. 
-      
-      -->
-    <xsd:anyAttribute namespace="##other" processContents="lax"/> 
-    <!-- <xsd:anyAttribute namespace="##other" processContents="strict"/> -->
+    IBM's DFDL Schemas have ibm-specific attributes on them that we 
+    need to tolerate. Users may want to decorate their DFDL schemas with
+    other non-native and non-DFDL attributes and annotation sub-elements.
+    
+    The default XSD behavior (commented out below) is to allow any additional
+    attributes on a schema by performing lax validation on anything not in the
+    XSD namespace.
+    
+    The problem: since DFDL short form properties are not in the XSD namespace,
+    they are not validated. This means common typographical errors or forgetting
+    the dfdl: prefix are often not caught.
 
-
-    <!-- need some wildcard to allow xml:lang, xml:space etc. which do appear in 
-      schemas sometimes -->
-      
-    <!-- Daffodil augments schemas with dafint:file dafint:line and dafint:col -->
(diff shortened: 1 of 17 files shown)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
```diff
@@ -1757,11 +1738,19 @@ this to validate when we load a DFDL Schema.
...
+  <xsd:element name="enumeration" id="enumeration">
     <xsd:annotation>
       <xsd:documentation
         source="http://www.w3.org/TR/xmlschema-2/#element-enumeration" />
     </xsd:annotation>
+    <xsd:complexType>
+      <xsd:complexContent>
+        <xsd:extension base="sub:facet">
+          <xsd:attributeGroup ref="dfdl:RepValuesAGQualified" />
+          <xsd:attributeGroup ref="sub:anyOther" />
```
mbeckerle: This sub namespace.... do we still use it? At one time we had this hack going about not using xs or xsd for the subset-of-XML-Schema schema, but I gave up on that long ago. Is sub still essential?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
```diff
@@ -33,6 +33,24 @@
...
+    Attributes in the dfdl namespae will be strictly validated.
+  -->
+  <xsd:attributeGroup name="anyOther">
+    <xsd:anyAttribute namespace="##other" processContents="lax" />
+  </xsd:attributeGroup>
+
+  <!--
+    This group allows for any element not in the dfdl namesapce.
+    Elments in the dfdl namespae will be strictly validated.
+  -->
```
mbeckerle: What about this enables strict validation for DFDL namespace attributes?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -160,15 +160,15 @@
     <xs:element name="inputTypeCalc_unionOfKeysetValueCalcs_01">
       <xs:complexType>
         <xs:sequence>
-          <xs:element name="byte" type="tns:_1through100_union_to_string" maxOccurs="unbounded" dfdl:occursCoundKind="parsed"/>
+          <xs:element name="byte" type="tns:_1through100_union_to_string" maxOccurs="unbounded" dfdl:occursCountKind="parsed"/>
```
mbeckerle: It will be very interesting to try this on the production schemas that are big. This sort of typo has to be common.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
stevedlawrence: I'm not really sure. It's only used in XMLSchema_for_DFDL.xsd, and the sub prefix has the same namespace as the xsd and xs prefixes, so  I don't think it really provides any benefit. That said, it shows up 261 times in that file, so it'd be a pretty big diff and would hide the changes in this file. I'm fine with removing it, but we should do that as a separate bug if we want.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
stevedlawrence: It's a combination of a few things. First, this anyAttribute is in a file that has the targetNamespace set to the ``dfdl`` namespace. The ``namespace="##other"" attribte says that any attribute will be allowed if it is not in that targetNamespace. And the ``processContext="lax"`` says that for those attributes that will be allowed, they should be validated if a schema can be found ffor them, but otherwise they'll just be allowed.

So the net effect of this is that when something references dfdl:anyOther, it will allow and lax-ly validate attributes that are NOT in the dfdl namespace.

But your right that this in itself doesn't enable strict validation of DFDL attributes, so I'll update the comment.

The thing that enables strict validation is where this is referenced. The places where dfdl:anyOther is referenced additionally define specific attributes in the dfdl namespace. So the schema allows either a bunch of stuff explicitly allowed and strictly validated (normal schema validation) OR it allows anytthing not in the dfdl namespace. 

I'll update this comment to make it more clear how this works and should be used.

PR #232, opened 2019-06-12
Simplify how we show text differences when a TDML test fails

stevedlawrence:
We previously found every character that differed and showed its path,
index, and expected/actual characters. But when an XML element contains
a lot of text (e.g. hexBinary blob) with a lot of differences, this
character by character diff is not very helpful and consumed a lot
memory to build the output string. This easily leads to an OutOfMemory
exception.

Instead, when showing the diff of two XML text elements, only show the
index of the first diff plus some following characters for context. This
should be useful enough in most cases to figure out where things when
off the rails while keep memory usage low, even on large diffs.

Also change the diff output to show the actual infoset before the
differences summary. Displaying the actual infoset after the diff made
it very difficult to find the summary of differences when big infosets
were involved.

DAFFODIL-2118

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
@@ -748,16 +748,13 @@ object XMLUtils {
 Comparison failed.
 Expected (attributes stripped)
           %s
-Actual (attributes stripped)
+Actual (attributes ignored for diff)
           %s
 Differences were (path, expected, actual):
-%s
-Actual with Attributes (needed for unparse)
 %s""".format(
         removeAttributes(expected).toString,
-        removeAttributes(actual).toString,
-        diffs.map { _.toString }.mkString("- ", "\n- ", "\n"),
-        actual))
+        actual,
+        diffs.map { _.toString }.mkString("- ", "\n- ", "\n")))
     }
   }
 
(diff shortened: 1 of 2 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -748,16 +748,13 @@ object XMLUtils {
 Comparison failed.
 Expected (attributes stripped)
           %s
-Actual (attributes stripped)
+Actual (attributes ignored for diff)
```
mbeckerle: Do we have a test that verifies this ignore logic doesn't strip the xsi:nil attribute?  Just asking since Guichard had an issue related to that with EDIFACT recently, and I wanted to make sure the issue is not in the TDML comparison logic. I don't think it is, but wanted to check.

stevedlawrence: I'm not sure if we have an explicit test for the XMLUtils xml comparison logic to make sure we can find xsi:nil differences. However, I have modified a nillable test to remove ``xsi:nil="true"`` from the expected infoset and got an error saying the expected infoset did not have the attribute, so I think our xsi:nil checking logic is correct. It's very possible that there's a edgecase or a bug somewhere, but the expected cases look correct to me. Attributes are ignored, but we special case and xsi:nil as expected.

PR #233, opened 2019-06-13
Warn when DFDL properties are ignored

stevedlawrence:
During schema compilation, a cache of properties exists to speed up
subsequent property lookups. This patch takes advantage of that cache to
determine which properties have been used or not. At the end of schema
compilation, we now inspect all Terms and look for any properties that
are defined on each Term but are not in the cache, which implies that a
property was not used and we can create an SDW. These warnings should
help schema developers realize when properties are accidentally ignored
(e.g. dfdl:lengthPattern is dfdl:lengthKind is not "pattern").

Additionally, code was refactored to property lookup functions to ensure
properties are always cached. Some parts of the code would skip caching
if it used findPropertyOption or other functions instead of
cachePropertyOption. To avoid accidentally not adding a lookup to the
cache, the cache* functions are removed, the findPropertyOption always
caches, and all property lookups are changed to use this function or a
wrapper.

Also modify all tests to remove all unused properties.

DAFFODIL-1619


daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -107,10 +108,43 @@ trait Term
     statements.foreach { _.checkTerm(this) }
   }
 
+  // Used to recursively go through Terms and look for DFDL properties that
```
mbeckerle: Should be scaladoc, not just a // comment.

daffodil-test/src/test/resources/org/apache/daffodil/section12/aligned_data/Aligned_Data.tdml
```diff
@@ -267,9 +267,9 @@
     <xs:element name="e16">
       <xs:complexType>
         <xs:sequence>
-          <xs:element name="e14" type="xs:string" dfdl:terminator=":" dfdl:representation="text" dfdl:lengthKind="explicit" dfdl:length="5" dfdl:encoding="ascii"
```
mbeckerle: Ok, so looking in general at these test changes, you are not distinguishing redundant properties from irrelevant properties. E.g., having dfdl:representation="text" on a string is redundant. Having "binaryBooleanTrueRep" on a string is irrelevant, and indicative of almost certainly an error. (In fact a switch to escalate such from warning to error might be of value.)

One could also warn about properties that are unnecessary because they match what is found in the default format surrounding the schema component, and so the properties do not need to be expressed on the schema component. From what I see in this code you are not going that far. I'm fine with that. 

So the question here I guess is "Am I understanding this correctly?"

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
```diff
@@ -3462,10 +3467,10 @@
     <xs:element name="nanInvalid" dfdl:lengthKind="delimited" dfdl:terminator="]" type="xsd:double"
     dfdl:textStandardNaNRep="%#r" />
 
-    <xs:element name="infInvalidType" dfdl:lengthKind="delimited" type="xsd:long" dfdl:textStandardInfinityRep="INFINITY"
+    <xs:element name="infInvalidType" dfdl:lengthKind="delimited" type="xsd:long"
     dfdl:terminator="]" />
 
-    <xs:element name="nanInvalidType" dfdl:lengthKind="delimited" type="xsd:long" dfdl:textStandardNaNRep="NOTANUMBER"
+    <xs:element name="nanInvalidType" dfdl:lengthKind="delimited" type="xsd:long"
```
mbeckerle: This looks like a test bug. I.e., the test should have been written to use a xs:float or xs:double so that the property would be relevant. The test just looks for a parse error complaining about the NOTANUMBER syntax. But if you remove this property, it's not really anything about nan anymore, and so the element name isn't right, nor is the name of the associated test. It's really just a test that non-numeric data won't parse as a long. 

I feel like this test should have been changed into one that is looking for the new warning about dfdl:textStandardNaNRep being ignored on an xsd:long. That's probably covered elsewhere, so arguably, this test here is just plain not relevant anymore, and given the misnomers, should probably just be deleted.

To what degree have tests now evolved into "poorly named tests" that don't really test the things they were originally testing? Is the quality of the tests declining/being-diluted now? Arguably, introduction of this new uniform detection capability for unused properties obviates quite a few tests that were introducting such properties in some sense to test that they were in fact being ignored. 

Do we have a backlog issue now of re-scrutinizing tests to be sure they're still doing meaningful testing?
This is really just a discussion topic that could be held on dev@ mailing list and isn't really core to this review.

daffodil-test/src/test/resources/org/apache/daffodil/section12/aligned_data/Aligned_Data.tdml
stevedlawrence: I don't do any checks for redundant properties, though that's an interesting idea. However, there are probably cases were someone wants to be explicit about a certain property on an element even though it's redundant to the default properties. We could add it, but i'm not sure it gives as much benefit as this check.

dfdl:representation="text" was removed anytime it was on a type with xs:string or xs:hexBinary. Elements with these types have an implied representation (text or binary respectively) and so we never access the representation property. So this is correct as far as warning about an unused property. This particular one probably never means a user made an error though.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
stevedlawrence: WIll fix.

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
stevedlawrence: This is a good point. Compared to how big our test base is, I was actually surprised with how few unused properties there were. And the vast majority of the these were testing that those properties were ignored. They were just unnecessary properties (e.g. boolean props on a string type). So I dont' think this changes things significantly.

In this particular case, this test was really just testing that textStandardNanRep was ignored when the type was xs:long. I think that a reasonable test, but I agree that it probably makes sense to add the property back and expect a warning. I'll take a second pass and make sure there were any tests that I fundamentally changed that should be checking for a warning.

stevedlawrence: I've gone through the modified tests, and the i think there were only three that were actually testing that a property was ignored. One I had already noticed that tested that textNumberPattern was ignored for inputValueCalc. I had already added a dfdl:warnign for that. THe other two were infInvalidType and nanInvalidType, which tests that textStandardInfinityRep and textStandardNaNRep are ignored for a non-float type.

The vast majority of the property removals were related to dfdl:representation on xs:string and xs:hexBInary, or things like text properties defined for a binary representation, which look like they were just copy paste errors. So I think our tests are are still fine and are testing what they say they are testing.

mbeckerle: Sounds good.

PR #235, opened 2019-06-18
Daffodil 2143  2154 initiated content and nillable EDIFACT Bugs

mbeckerle:
Fixes for two bugs.

With these fixes, EDIFACT now runs portably. 

There are two commits, which are the fixes of the two bugs each separately.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
@@ -23,6 +23,9 @@ import org.apache.daffodil.grammar.GrammarMixin
 import org.apache.daffodil.grammar.primitives.InitiatedContent
 import org.apache.daffodil.grammar.primitives.Terminator
 import org.apache.daffodil.grammar.primitives.Initiator
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.compiler.ForParser
+import org.apache.daffodil.api.WarnID
 
 trait InitiatedTerminatedMixin
   extends GrammarMixin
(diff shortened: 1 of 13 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
```diff
@@ -61,7 +64,23 @@ trait InitiatedTerminatedMixin
...
 
-  private lazy val initiatorDiscriminator = prod("initiatorDiscriminator", parentSaysInitiatedContent) { InitiatedContent(this) }
+  private lazy val initiatorDiscriminator = prod(
+    "initiatorDiscriminator",
+    parentSaysInitiatedContent &&
+      immediatelyEnclosingModelGroup.map {
+        case c: ChoiceTermBase => true
+        case s: SequenceTermBase => (isArray || isOptional) &&
+          isVariableOccurrences
+      }.getOrElse(false)) {
```
stevedlawrence: Would it be possible to put this guard in an variable? It wasn't immediately clear to me that this is where the guard ended and the grammar stared.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
```diff
@@ -61,7 +64,23 @@ trait InitiatedTerminatedMixin
...
+      this match {
+        case eb: ElementBase => {
+          if (eb.minOccurs < 1)
+            SDW(WarnID.InitiatedContentMinOccursZero, "The minOccurs attribute should not be zero when dfdl:initiatedContent is 'yes'.")
```
stevedlawrence: Should this be an SDE?

mbeckerle: Yes. I'll add comments. I made this an SDW because making it an SDE breaks EDIFACT, possibly other schemas.
I will add a JIRA ticket to add tunable for this to control SDE/SDW.

mbeckerle: Actually. This is a bug. The restriction that the child of the initiatedContent group cannot be optional applies only to choices, not to sequences. The EDIFACT situation is sequences. So I can fix this.

PR #236, opened 2019-06-18
Add support for Link16 Charset 311:002

bsloane1650:
DAFFODIL-2155

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
@@ -15,6 +15,16 @@
  * limitations under the License.
  */
 
+/*
+ * This file contains special character encodings used by various US military data formats
+ * The DFI/DUI numbers may be used to lookup the definition of the encoding by those with access
+ * to the relevent specification. In cases where multiple DFI/DUI pairs make use of the same encoding,
+ * the decision of which pair to name the encoding after is arbitrary.
+ * 
+ * Unless otherwise stated, characters defined as "NO STATEMENT" are assumed to be padding characters.
+ * These characters are translated into a no-break space (U+00A0)
+ */
+
 package org.apache.daffodil.processors.charset
 
 import org.apache.daffodil.schema.annotation.props.gen.BitOrder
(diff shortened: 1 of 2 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
```diff
@@ -15,6 +15,16 @@
  * limitations under the License.
  */
 
+/*
+ * This file contains special character encodings used by various US military data formats
+ * The DFI/DUI numbers may be used to lookup the definition of the encoding by those with access
+ * to the relevent specification. In cases where multiple DFI/DUI pairs make use of the same encoding,
+ * the decision of which pair to name the encoding after is arbitrary.
```
mbeckerle: I think you should choose the lowest numbered DFI/DUI pair that uses the encoding as the identifying one.

bsloane1650: I do not believe that is possible.

Many specifications (including all those I have access to) provide only a subset of all DFI/DUI pairs, so it is not possible to know what the lowest number DFI/DUI pair is. Even with access to the global dictionary, it is possible that a lower numbered one will be added in the future.

mbeckerle: Unlikely to be added with lower numbers tho, but I take your point. You don't want to assure anyone about the numbering. I was surprised that Link11 spec, which i assumed was older than Link16 (11 being less than 16 after all), has a data dictionary that starts at DFI 2500 ! So much for sane ordering.

bsloane1650: I suspect that DFI numbers were allocated in large blocks. I wouldn't put much stock in their numbering other then as a reflection of the order the blocks were allocated in.

As far as I am aware, there is no commitment to not add new DUIs to existing DFIs either.

stevedlawrence: We should add some tests to cover this code to make sure it does the right thing.

PR #237, opened 2019-06-18
Fixes DAFFODIL-2142

olabusayoT:
-- adds warning for annotation in complexType instead of silently dropping it
-- moves checks for non-Model groups to complexTypes

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ComplexTypes.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ComplexTypes.scala
@@ -20,6 +20,9 @@ package org.apache.daffodil.dsom
 import scala.xml.Node
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.xml.QName
+import org.apache.daffodil.api.WarnID
+import scala.xml.Text
+import scala.xml.Comment
 
 abstract class ComplexTypeBase(xmlArg: Node, parentArg: SchemaComponent)
   extends SchemaComponentImpl(xmlArg, parentArg)
@@ -52,8 +55,26 @@ abstract class ComplexTypeBase(xmlArg: Node, parentArg: SchemaComponent)
     xmlChildren.flatMap {
       xmlChild =>
         {
-          val g = ModelGroupFactory(xmlChild, this, 1, false) // discards unwanted text nodes also.
-          g
+          xmlChild match {
+            case <annotation>{ annotationChildren @ _* }</annotation> => {
+              val dais = annotationChildren.find { ai =>
+                ai.attribute("source") match {
+                  case Some(n) => n.text.contains("ogf") && n.text.contains("dfdl")
+                  case _ => false
+                }
+              }
+              if (dais != None) {
+                this.SDW(WarnID.InvalidAnnotationPoint, "complexType is not a valid annotation point. Annotation ignored.")
+              }
+              None
+            }
+            case textNode: Text => None
+            case _: Comment => None
+            case _ => {
+              val g = ModelGroupFactory(xmlChild, this, 1, false)
+              Some(g)
+            }
+          }
         }
     }
   }.value
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
@@ -43,12 +43,13 @@ import org.apache.daffodil.schema.annotation.props.gen.YesNo
 object ModelGroupFactory {
 
   /**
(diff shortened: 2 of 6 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ComplexTypes.scala
```diff
@@ -52,8 +55,26 @@ abstract class ComplexTypeBase(xmlArg: Node, parentArg: SchemaComponent)
...
+          xmlChild match {
+            case <annotation>{ annotationChildren @ _* }</annotation> => {
+              val dais = annotationChildren.find { ai =>
+                ai.attribute("source") match {
+                  case Some(n) => n.text.contains("ogf") && n.text.contains("dfdl")
+                  case _ => false
+                }
+              }
+              if (dais != None) {
+                this.SDW(WarnID.InvalidAnnotationPoint, "complexType is not a valid annotation point. Annotation dropped.")
```
mbeckerle: I prefer the term "ignored" to "dropped"

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -83,9 +83,6 @@ object ModelGroupFactory {
           val groupRefInstance = groupRefFactory.groupRef
           List(groupRefInstance.asModelGroup)
         }
-        case <annotation>{ _* }</annotation> => Nil
-        case textNode: Text => Nil
-        case _: Comment => Nil
         case _ => {
           lexicalParent.SDE("Unrecognized construct: %s", child)
```
mbeckerle: I think this should not be an SDE. It should be Assert.invariantFailed("Unrecognized construct %s should be handled by caller.".format(child)), because if the code gets here with an annotation or comment node of XML that's not a user's problem with their schema, but a flaw in our code (likely a mistake during maintenance).

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ComplexTypes.scala
olabusayoT: Updates made

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
olabusayoT: Updates made!

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -44,7 +44,7 @@ object ModelGroupFactory {
 
   /**
    * Because of the contexts where this is used, we return a list. That lets users
-   * flatmap it to get a collection of model groups. Nil for non-model groups, non-Nil for the model group
+   * flatmap it to get a collection of model groups. It will return non-Nil for the model group
    * object. There should be only one non-Nil.
```
stevedlawrence: Is seems like this isn't the case anymore. It looks like it can never return non-Nil.  I think it previously returned Nil when the child passed in was not a model group (eg.. annotation, comment, textNode). But with this change, it's always either a seq, a choice, or a group. So should this now just return a ModelGroup? And callers can Map insead of flatMap?

olabusayoT: I made updates to both TermFactory and ModelGroup because they should only return a ModelGroup or Term, rather than a list because the callers handle non Model Group/non Term elements. However depending on how the caller handles aforementioned undesirable elements, a flatMap might still be necessary, as is the case with ModelGroup's caller in ComplexType.scala. Undesirable elements in the case are set to Nil, so they can be removed by the flatMap in the caller. So ModelGroupFactory's return still needs to be wrapped in a list. For TermFactory however, this isn't the case, as we remove undesirable elements, rather than doing any processing on them.

stevedlawrence: Ah, your right, the flatMap is still needed. Only suggestion is the switch to an Option for the flatmap since it only flaMaps a Nil or List of one. So replace Nil with None and List(...) with Some(...) Some(...) insead of List. Functionally it's exactly the same thing, but I think more accurately shows that you either have one thing or no things, and the no things get dropped.

Also, in the future, avoid squashing and force-pushing until all comments are address and you get enough +1's for the PR to be accepted. Github doesn't show differences between squashes at all, so it's difficult to figure out what updates have been made to the PR.

+1 :+1:

olabusayoT: Makes sense! Updates made.

PR #239, opened 2019-06-19
Fixes OVC on elements of complex type

olabusayoT:
Rather than a huge trace, we get a schema definition error if an outputValueCalc is put on element of complex type

DAFFODIL-1701

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc.tdml
@@ -315,6 +315,40 @@
 			</xs:complexType>
 		</xs:element>
 
+        <xs:element name="refSimpleElementWithOvc">
+          <xs:complexType>
+            <xs:sequence>
+              <xs:element ref="ex:st1" dfdl:outputValueCalc="{1}"/>
+            </xs:sequence>
+          </xs:complexType>
+        </xs:element>
+
+        <xs:element name="st1" type="xs:int"/>
+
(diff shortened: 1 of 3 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc.tdml
```diff
@@ -708,4 +742,63 @@
 	    </tdml:errors>
 
 	</tdml:unparserTestCase>
+
+    <tdml:unparserTestCase name="refSimpleTypeElemWithOvc"
```
mbeckerle: Why are these unparserTestCases ? Was this bug particular to unparsing? 
You can convert these to parserTestCase I think.

olabusayoT: I went with unparserTestCase since we're dealing with testing outputValueCalcs and I only cared about the results of unparsing the provided infoset (i.e roundTrip='false') and triggering the OVC code path, and not the results of parsing. Also for consistency,  since the only tests in the file are unparserTestCase.

olabusayoT: I thought OVCs were ignored during parsing.

mbeckerle: My mistake. I was reading these as InputValueCalc.... maybe I need new glasses ?

+1.

PR #240, opened 2019-06-19
Add dfdlx:choiceBranchKeyRanges

bsloane1650:
DAFFODIL-2157

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
@@ -34,9 +34,10 @@ import scala.util.Failure
 import scala.util.Success
 import scala.math.BigInt
 import org.apache.daffodil.cookers.RepValueCooker
-import org.apache.daffodil.processors.RangeBound
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.MaybeInt
+import org.apache.daffodil.util.RangeBound
+import org.apache.daffodil.cookers.IntRangeCooker
 
 /*
  * The purpose of the ChoiceCombinator (and the parsers it creates) is to
@@ -83,6 +84,15 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
         case ChoiceLengthKind.Explicit => new SpecifiedLengthChoiceParser(cp, ch.choiceRuntimeData, choiceLengthInBits.get)
       }
     } else {
+      //Verify that every alternative has some form of branch key
+      alternatives.map{ alt =>
+        val keyTerm = alt.context.asInstanceOf[Term]
+        val hasBranchKey = keyTerm.findPropertyOption("choiceBranchKey").isDefined
+        val hasBranchKeyRanges = keyTerm.findPropertyOption("choiceBranchKeyRanges").isDefined
+        if(!hasBranchKey && !hasBranchKeyRanges && ch.defaultableChoiceBranchKeyKind != ChoiceKeyKindType.ByType){
+          keyTerm.SDE("Neither dfdl:choiceBranchKey nor dfdlx:choiceBranchKeyRanges is defined.")
+        }
+      }
       val dispatchBranchKeyValueTuples: Seq[(String, Gram)] = alternatives.flatMap { alt =>
         val keyTerm = alt.context.asInstanceOf[Term]
         val uncookedBranchKeys =
@@ -95,9 +105,8 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
               val ans = repValueSet.valueSet.toSeq.map(_.toString).mkString(" ")
               ans
             }
-            case ChoiceKeyKindType.Explicit    => keyTerm.choiceBranchKey
-            case ChoiceKeyKindType.Implicit    => keyTerm.choiceBranchKey
-            case ChoiceKeyKindType.Speculative => keyTerm.choiceBranchKey
+            case ChoiceKeyKindType.Explicit | ChoiceKeyKindType.Implicit   => keyTerm.findPropertyOption("choiceBranchKey").toOption.getOrElse("")
+            case ChoiceKeyKindType.Speculative => Assert.invariantFailed("Cannot have choiceKeyKind==speculative with direct dispatch")
           }
         val cbks = {
           if(uncookedBranchKeys.isEmpty){
@@ -113,31 +122,39 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
       //Since there is not a choiceBranchKeyRanges attribute, this can only currently be populated by repType
       val dispatchBranchKeyRangeTuples: Seq[(RangeBound[BigInt], RangeBound[BigInt], Parser, Boolean)] = alternatives.flatMap { alt =>
         val keyTerm = alt.context.asInstanceOf[Term]
-        val branchKeyRanges:Set[(RangeBound[BigInt],RangeBound[BigInt])] = ch.defaultableChoiceBranchKeyKind match {
+        val branchKeyRanges:Seq[(RangeBound[BigInt],RangeBound[BigInt])] = ch.defaultableChoiceBranchKeyKind match {
           case ChoiceKeyKindType.ByType => {
             val keyTerm_ = keyTerm.asInstanceOf[ElementBase]
-            keyTerm_.simpleType.optRepValueSet.get.valueRanges.map(x=>{
+            keyTerm_.simpleType.optRepValueSet.get.valueRanges.toSeq.map(x=>{
               val x_ = x.asInstanceOf[(RangeBound[BigInt],RangeBound[BigInt])]
               (x_._1,x_._2)
             })
           }
-          case ChoiceKeyKindType.Explicit    => Set()
-          case ChoiceKeyKindType.Implicit    => Set()
-          case ChoiceKeyKindType.Speculative => Set()
+          case ChoiceKeyKindType.Explicit | ChoiceKeyKindType.Implicit    => 
+            IntRangeCooker.convertConstant(keyTerm.findPropertyOption("choiceBranchKeyRanges").toOption.getOrElse(""), context, false)
+          case ChoiceKeyKindType.Speculative => Assert.invariantFailed("Cannot have choiceKeyKind==speculative with direct dispatch")
         }
         branchKeyRanges.toSeq.map(x=>(x._1, x._2, alt.parser, alt.context.enclosingTerm.get.isRepresented))
       }
 
-      // check for duplicate dfdl:choiceBranchKeys
+      // check for duplicate branch keys
       // Our handling of ranges here is definantly suboptimal, but hopefully
       // we don't see enough distinct ranges on a single element for this to be an issue
+      // Additionally, at this point, the keys could be comming from either the choiceBranchKey family of attributes
+      // or the repValue family of attributes
+      val (branchKeyAttribute, branchKeyRangeAttribute) = ch.defaultableChoiceBranchKeyKind match{
+        case ChoiceKeyKindType.ByType => ("dfdlx:repValues","dfdlx:repValueRanges")
+        case ChoiceKeyKindType.Explicit | ChoiceKeyKindType.Implicit => ("dfdl:choiceBranchKey","dfdlx:choiceBranchKeyRanges")
+        case ChoiceKeyKindType.Speculative => Assert.invariantFailed("Cannot have choiceKeyKind==speculative with direct dispatch")
+      }
       val groupedByKey = dispatchBranchKeyValueTuples.groupBy(_._1)
       groupedByKey.foreach {
         case (k, kvs) =>
           if (kvs.length > 1) {
             SDE(
-              "dfdl:choiceBranchKey value (%s) is not unique across all branches of a direct dispatch choice. Offending branches are:\n%s",
-              k, kvs.map(_._2.context.runtimeData).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""))
+              "%s value (%s) is not unique across all branches of a direct dispatch choice. Offending branches are:\n%s",
+              branchKeyAttribute, k, 
+              kvs.map(_._2.context.runtimeData).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""))
           }
           Try(k.toLong) match {
             case Success(v) => {
@@ -145,8 +162,8 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
               val conflictingRanges = dispatchBranchKeyRangeTuples.filter({ case (min, max, _, _) => min.testAsLower(asBigInt) && max.testAsUpper(asBigInt) })
               if (conflictingRanges.length > 0) {
                 SDE(
-                  "dfdl:choiceBranchKey (%s) conflicts with dfdl:choiceBranchKeyRanges. Offending branches are:\n%s\n%s",
-                  k,
+                  "%s (%s) conflicts with %s. Offending branches are:\n%s\n%s",
+                  branchKeyAttribute, k, branchKeyRangeAttribute,
                   kvs.map(_._2.context.runtimeData).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""),
                   conflictingRanges.map(_._3.context).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""))
               }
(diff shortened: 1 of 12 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -83,6 +84,15 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
         case ChoiceLengthKind.Explicit => new SpecifiedLengthChoiceParser(cp, ch.choiceRuntimeData, choiceLengthInBits.get)
       }
     } else {
+      //Verify that every alternative has some form of branch key
+      alternatives.map{ alt =>
+        val keyTerm = alt.context.asInstanceOf[Term]
+        val hasBranchKey = keyTerm.findPropertyOption("choiceBranchKey").isDefined
+        val hasBranchKeyRanges = keyTerm.findPropertyOption("choiceBranchKeyRanges").isDefined
+        if(!hasBranchKey && !hasBranchKeyRanges && ch.defaultableChoiceBranchKeyKind != ChoiceKeyKindType.ByType){
+          keyTerm.SDE("dfdl:choiceBranchKey and dfdlx:choiceBranchKeyRanges are not defined.")
```
mbeckerle: Should say "Neither dfdl:choiceBranchKey nor dfdlx:choiceBranchKeyRanges is defined." since these are alternatives and neither-nor is singular. Of course I realize they aren't exclusive alternatives. One can have both, but neither-nor is still less ambiguous than "and".

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
```diff
@@ -84,3 +89,28 @@ object ChoiceBranchKeyCooker extends ListOfStringLiteralNonEmptyNoCharClassEntit
...
+object IntRangeCooker extends Converter[String, Seq[Range[BigInt]]] {
+  protected def convert(input: String, context: ThrowsSDE, forUnparse: Boolean): Seq[Range[BigInt]] = {
+    def run(xs: Seq[String]): Seq[(RangeBound[BigInt], RangeBound[BigInt])] = {
+      xs match {
+        case Seq() => Seq()
+        case a +: b +: rest => {
+          val a2 = BigInt(a)
+          val b2 = BigInt(b)
+          if (a2.compare(b2) > 0) {
+            context.SDE("min value must not be greater than max value")
```
mbeckerle: I think you should display the specific min and max values in this message. If someone typos and deletes a number from middle of a whitespace separated list of ranges, then this could be a little hard to isolate if there's lots of pairs.

daffodil-lib/src/main/scala/org/apache/daffodil/util/RangeBound.scala
```diff
@@ -0,0 +1,111 @@
+/*
```
mbeckerle: So I see you pulled this code out of the file below in order to share it, which is great. 

I'm going to assume this didn't change in some subtle way as well. If there's something to look at for delta review in this that's in some way subtle, please highlight for me. Otherwise I'm assuming this code is ok.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -21,6 +21,7 @@ import org.apache.daffodil.processors._
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.util.RangeBound
 
 class ComplexTypeParser(rd: RuntimeData, bodyParser: Parser)
   extends CombinatorParser(rd) {
```
mbeckerle: Did the next file ".choiceBranchKeyRanges.tdml.swp" get added by mistake?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/choiceBranchRanges/choiceBranchKeyRanges.tdml
```diff
@@ -0,0 +1,204 @@
...
+
+<tdml:testSuite xmlns:ex="http://example.com" xmlns="http://example.com"
+  xmlns:tns="http://example.com"
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  >
+
+  <tdml:defineSchema name="choiceBranchKeyRanges-Embedded.dfdl.xsd">
```
mbeckerle: You don't have to follow this convention of naming the embedded schemas with these long names and "-Embedded.dfdl.xsd". That is something that got started long ago, but I think a shorter identifier without ".dfdl.xsd" and without repeating the name of the suite, or a class name on it is probably better given that people often copy/paste TDML and adapt to new purposes. One less name to have to change if you repurpose some TDML tests for a new purpose.

daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice.tdml
```diff
@@ -1838,6 +1838,7 @@ it sure is/
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>choiceBranchKey</tdml:error>
+      <tdml:error>choiceBranchKeyRanges</tdml:error>
```
mbeckerle: Nice. Even if you change the message above (the neither-nor change), this test will still be right.

daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice.tdml
```diff
@@ -1838,6 +1838,7 @@ it sure is/
...
       <tdml:error>not defined</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
```
mbeckerle: I didn't see a test for the range list being odd length, though your code does detect that and issue an SDE.

daffodil-lib/src/main/scala/org/apache/daffodil/util/RangeBound.scala
```diff
@@ -0,0 +1,111 @@
...
+
+package org.apache.daffodil.util
+
+import java.lang.{ Long => JLong }
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.cookers.Converter
+import org.apache.daffodil.exceptions.ThrowsSDE
+import org.apache.daffodil.util.RangeBound.Range
+
+object RangeBound {
```
bsloane1650: This object is the only real change in this file. Otherwise, it is just a direct move from the typeCalculator file.

daffodil-lib/src/main/scala/org/apache/daffodil/util/RangeBound.scala
bsloane1650: Sorry, I should have highlighted it when I put up the PR. It should have a comment now. Bassically just a Range type-alias.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -113,17 +122,17 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
...
               val x_ = x.asInstanceOf[(RangeBound[BigInt],RangeBound[BigInt])]
               (x_._1,x_._2)
             })
           }
-          case ChoiceKeyKindType.Explicit    => Set()
-          case ChoiceKeyKindType.Implicit    => Set()
-          case ChoiceKeyKindType.Speculative => Set()
+          case ChoiceKeyKindType.Explicit | ChoiceKeyKindType.Implicit    => 
+            IntRangeCooker.convertConstant(keyTerm.findPropertyOption("choiceBranchKeyRanges").toOption.getOrElse(""), context, false)
+          case ChoiceKeyKindType.Speculative => Assert.invariantFailed("Cannot have choiceKeyKind==speculative with direct dispatch")
```
stevedlawrence: There's some error messages not adding in this commit that reference "dfdl:choiceBranchKeyRanges". But that's added in this commit? Are those error messages referencing the wrong thing? Looks likes some primilary choiceBranchKeyRange stuff made it into the enum change? Can you fix the error messages to have the right dfdlx namespace?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -34,9 +34,10 @@ import scala.util.Failure
 import scala.util.Success
 import scala.math.BigInt
 import org.apache.daffodil.cookers.RepValueCooker
-import org.apache.daffodil.processors.RangeBound
```
stevedlawrence: Unrelated to this change, but I was looking at the choice parsers and noticed this line to handle ths keys:
```
val optAns1= dispatchKeyRangeMap.filter({case(min,max,_,_) => min.testAsLower(keyAsBigInt) && max.testAsUpper(keyAsBigInt)}).headOption
```
Minor performance thing, but this can just be a find instead of a filter + headOption. Avoids creating a new list and and stop the search once it finds the first match instead of filtering the rest of the list. 

Another thing I noticed is that right above that we have:
```
val keyAsBigInt = BigInt(key)
```
So we are blindly converting the key to a BigInt. What happens if the dfdl:choiceDispatchKey evaluates to a string that can't be converted to an int? Should that be an SDE? If dfdl:choiceBranchKeyRanges exist, does the key need to be convertable to an int? Or do we just ignore the range mappings knowing that they can't match? If we allow both choieBranchKey and choiceBranchKeyRanges, I assume it just means we ignore mappings? Can you add some negative tests to make sure we do the right thing, whatever is decided?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/choiceBranchRanges/choiceBranchKeyRanges.tdml
mbeckerle: Ideally, the file and line numbers pointing at schema should simply be file and line numbers into the defineSchema inside the TDML file. Not some temp file.

stevedlawrence: I don't think that's the case. I'll open a bug.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
bsloane1650: I added this to my global gitignore.

I will also send a separate PR adding this to Daffodil's gitignore.

stevedlawrence: I don't feel strongly about it, but my opinion is that Daffodil's .gitignore shouldn't contain things specific to a particular development environment. It should only contain intermediate files that are part of a build/test. The idea is that we shouldn't have to try to maintain a list of all the different tmp files that different editors might want to throw in our repo. It's up to users to handle that with a global .gitignore based on their dev environment.

That said, for common things like vim and others, it could prevent this kind of issue, so I'm not totally against it. Our gitignore is already kindof a mess with different dev files scattered throughout.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
```diff
@@ -84,3 +89,28 @@ object ChoiceBranchKeyCooker extends ListOfStringLiteralNonEmptyNoCharClassEntit
 object RepValueCooker extends ListOfStringLiteralNonEmptyNoCharClassEntitiesNoByteEntities()
 
 object UpperCaseTokenCooker extends UpperCaseToken
+
+object IntRangeCooker extends Converter[String, Seq[Range[BigInt]]] {
+  protected def convert(input: String, context: ThrowsSDE, forUnparse: Boolean): Seq[Range[BigInt]] = {
```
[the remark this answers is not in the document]
bsloane1650: Maybe. All the other cookers are defined in EntityReplacer.scala. That file has already seen a bit of scope creep, as some of the cookers go beyond simply replacing entities, but this new cooker seems to be marginally farther away from anything that can be thought of as an "entityReplacer"

Additionally, I do not have a good name to give the guts of this if I move them to EntityReplacer.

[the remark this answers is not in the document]
stevedlawrence: Maybe we rename EntityRepalcer.scala to CookerImplementations or something more generic? You're right that it's definiely more than entity replacement now. Following other cookers, the implementation could be something like "IntRange" or "IntRangeCookerBase", IntRangeCookerImpl, etc. The names are a big generic, but they are in a separate "cooker" subpackage, so conflicts probably wouldn't be a problem.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
```diff
@@ -108,6 +108,7 @@
     <xsd:attribute name="encodingErrorPolicy" type="dfdl:EncodingErrorPolicyEnum"/>
     <xsd:attribute name="ignoreCase" type="dfdl:YesNoEnum" />
     <xsd:attribute name="choiceBranchKey" type="dfdl:ListOfDFDLStringLiteral" />
+    <xsd:attribute ref="dfdlx:choiceBranchKeyRanges" />
```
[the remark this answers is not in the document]
bsloane1650: I don't actually think the scala code explicitly validates this. I believe it assumes that it is an int then checks the more subtle requirements (even length and min<max). Without this change, I suspect giving it a non-integer would results in an unhandled exeption.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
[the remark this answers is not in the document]
mbeckerle: Cookers dealing with DFDL Entities should go in the central place to insure sharing. We could seal the base classes/traits to enforce that I suppose. 

But we cant require all string "conditioning" which might be termed "cooking" to live in one place. 

That's how I'd suggest we split it up.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
bsloane1650: Looking closer at this, I'm not actually sure how to write this error message.

What is happening is we have 2 potential sources for branch keys. Either we get the branch keys from the dfdl:choiceBranchKey and dfdlx:choiceBranchKeyRanges attributes, or we get the branch keys from the dfdlx:repValues and dfdlx:repValueRanges attributes of the underlying types.

To avoid duplication, we first get the keys and keyRanges from either source, then have a single codepath do the validation. This means that at the point where we emit these errors, the problem could be in either the choiceBranchKey(Ranges) attributes, or in the repValue(Ranges) attributes.

We should still be able to distinguish these two cases by using the ch.defaultableChoiceBranchKeyKind value.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
bsloane1650: I think we just ignore the ranges. In theory, you could still have non-integer keys; and if you don't you will fail fail with a dispatch error.

I'll file a seperate bug report for this. I would prefer not trying to sneak unrelated changes in.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
bsloane1650: We already IntelliJ, Eclipse, and Emacs entries, so I'm in favor of adding Vim on purely religious grounds ;)

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
bsloane1650: On second thought, the original enum proposal defined that repValues must be ints if repValueRanges was used, so this really is only an now that we have choiceBranchKeyRanges as well, so I will fix this here.

bsloane1650:
@mbeckerle 
I believe I addressed your comments. And remaining issues before I rebase and merge?

stevedlawrence:
LGTM, still a +1

stevedlawrence:
Just remember to squash before the rebase + merge.

PR #243, opened 2019-06-20
Fixed choice with dispatch key on array element.

mbeckerle:
DAFFODIL-2162

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
@@ -302,7 +302,7 @@ trait SequenceDefMixin
 
   final lazy val <sequence>{ apparentXMLChildren @ _* }</sequence> = (xml \\ "sequence")(0)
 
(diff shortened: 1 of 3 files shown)


mbeckerle:
The idea is to insure invariants the software depends on are being upheld. This particular invariant is that any sequence, even just the one implied when a choice has an array element child, can be asked for its format annotation object, and one is positively returned, even if it's an empty one containing no annotations. 

The separators of an enclosing sequence do not appear around array element occurrences for a choice branch that is an array element. To achieve that you need an explicit sequence surrounding the array element so you can express the separator. 

The rationale here is that choices don't get to express separator properties, but that's where they would have to appear if the choice branch (that is an array element) was to have separators. 

I suppose the alternative would be to say choices are transparent to separation by an enclosing sequence. But that's not true if the choice expresses initiator/terminator, which choices are allowed to have.  I lost this battle long ago in the DFDL Workgroup. My position was only sequences should have delimiters of any kind, period, but this did not win the day, as other existing format description systems including some based on XSD, allowed these sorts of properties to be expressed on a variety of constructs.

PR #245, opened 2019-06-24
Added X-DFDL-5-BIT-DFI-1661-DUI-001 char encoding

bsloane1650:
Also snuck in a typo fix for X-DFDL-6-BIT-DFI-311-DUI-002
DAFFODIL-2164

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
@@ -43,7 +43,15 @@ object BitsCharset6BitDFI264DUI001 extends {
 object BitsCharset6BitDFI311DUI002 extends {
   override val name = "X-DFDL-6-BIT-DFI-311-DUI-002"
   override val bitWidthOfACodeUnit = 6
-  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTuVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD \uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD0123456789\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
+  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTUVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD \uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD0123456789\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
+  override val replacementCharCode = 0x0
+  override val requiredBitOrder = BitOrder.LeastSignificantBitFirst
+} with BitsCharsetNonByteSize
+
+object BitsCharset5BitDFI1661DUI001 extends {
+  override val name = "X-DFDL-5-BIT-DFI-1661-DUI-001"
+  override val bitWidthOfACodeUnit = 5
+  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTUVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
   override val replacementCharCode = 0x0
   override val requiredBitOrder = BitOrder.LeastSignificantBitFirst
 } with BitsCharsetNonByteSize
--- a/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
@@ -25,6 +25,7 @@ object DaffodilCharsetProvider {
     BitsCharset5BitPackedLSBF,
     BitsCharset6BitDFI264DUI001,
     BitsCharset6BitDFI311DUI002,
+    BitsCharset5BitDFI1661DUI001,
     BitsCharsetBinaryLSBF,
     BitsCharsetBinaryMSBF,
     BitsCharsetHexLSBF,
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section10/representation_properties/encodings.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section10/representation_properties/encodings.tdml
@@ -43,6 +43,9 @@
     
   <xs:element name="sixBit1" type="xs:string" dfdl:encoding="X-DFDL-6-BIT-DFI-264-DUI-001"
     dfdl:bitOrder="leastSignificantBitFirst" dfdl:byteOrder="littleEndian"/>
+
+  <xs:element name="fiveBitDFI1661DUI001" type="xs:string" dfdl:encoding="X-DFDL-5-BIT-DFI-1661-DUI-001"
+    dfdl:bitOrder="leastSignificantBitFirst" dfdl:byteOrder="littleEndian"/>
     
 </tdml:defineSchema>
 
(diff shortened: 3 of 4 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
```diff
@@ -43,7 +43,15 @@ object BitsCharset6BitDFI264DUI001 extends {
...
-  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTuVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD \uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD0123456789\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
+  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTUVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD \uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD0123456789\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
+  override val replacementCharCode = 0x0
+  override val requiredBitOrder = BitOrder.LeastSignificantBitFirst
+} with BitsCharsetNonByteSize
+
+object BitsCharset5BitDFI1661DUI001 extends {
+  override val name = "X-DFDL-5-BIT-DFI-1661-DUI-001"
+  override val bitWidthOfACodeUnit = 5
+  override val decodeString = """\u00A0ABCDEFGHIJKLMNOPQRSTUVWXYZ\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"""
```
stevedlawrence: Can you ad some tests for this encoding?

bsloane1650: I am trying to write a test-case for this. What I came up with is:

```
  <xs:element name="fiveBitDFI1661DUI001" type="xs:string" dfdl:encoding="X-DFDL-5-BIT-DFI-1661-DUI-001"
    dfdl:bitOrder="leastSignificantBitFirst" dfdl:byteOrder="littleEndian"/>

  <tdml:parserTestCase name="fiveBitDFI1661DUI001" root="fiveBitDFI1661DUI001" model="enc1" description="X-DFDL-5-BIT-DFI-1661-DUI-001">
    <tdml:document>
      <tdml:documentPart type="bits" bitOrder="LSBFirst" byteOrder="RTL">
      00000
      00001
      00010
      00011
      00100
      00101
      00110
      00111
      01000
      01001
      01010
      01011
      01100
      01101
      01110
      01111
      10000
      10001
      10010
      10011
      10100
      10101
      10110
      10111
      11000
      11001
      11010
      11011
      11100
      11101
      11110
      11111
      </tdml:documentPart>
    </tdml:document>
    <tdml:infoset>
      <tdml:dfdlInfoset>
        <!--
        Note, the space below is actually \u00A0 no-break space
        -->
        <tns:fiveBitDFI1661DUI001><![CDATA[ ABCDEFGHIJKLMNOPQRSTUVWXYZ&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;]]></tns:fiveBitDFI1661DUI001>
      </tdml:dfdlInfoset>
    </tdml:infoset>
  </tdml:parserTestCase>
```

This fails, with the actual result of the parse being:
```
<ex:fiveBitDFI1661DUI001 xmlns:ex="http://example.com">&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;ZYXWVUTSRQPONMLKJIHGFEDCBA&#xA0;</ex:fiveBitDFI1661DUI001>
```

Note that the string has the characters in reverse order.
Given my experience in the area, I assume the problem is with my understanding of LSBF bit ordering.

As a secondary concern, Daffodil is outputing non-breaking space as &#xA0; . This is technically correct, but it is not clear if this is desireable or not.

Also, I would expect this test to fail to round-trip because of the number of "undefined" characters being mapped to u+FFFD. Once I figure out parse only for this one, I will add a second that round-trips without those characters.

mbeckerle: Yes, this is working as designed.

When you use byteOrder="RTL" the bits are interpreted as if the whole thing was on one line, and you are starting from the "right" side. But of course if you spill over to multiple lines, this just means from the end to the beginning.

bsloane1650: How should I be doing this then?

To avoid issues of text encoding, I switch my schema to be a sequence of 5 bit unsigned integers:

```
  <xs:element name="fiveBitDFI1661DUI001"
    dfdl:bitOrder="leastSignificantBitFirst" dfdl:byteOrder="littleEndian">
     <xs:complexType>
       <xs:sequence>
         <xs:element name="character" type="xs:unsignedInt" dfdl:lengthKind="explicit" dfdl:lengthUnits="bits" dfdl:length="5" dfdl:representation="binary"
           dfdl:bitOrder="leastSignificantBitFirst" dfdl:byteOrder="littleEndian" maxOccurs="unbounded" dfdl:occursCountKind="parsed"
           dfdl:alignmentUnits="bits"/>
       </xs:sequence>
     </xs:complexType>
  </xs:element>
```

When the document is specified as LSBFirst, RTL, the behaviour is consisntent with what I was seeing with text encoding:

```
<?xml version="1.0"?>
<ex:fiveBitDFI1661DUI001 xmlns:ex="http://example.com">
  <ex:character>31</ex:character>
  <ex:character>30</ex:character>
  <ex:character>29</ex:character>
  <ex:character>28</ex:character>
  <ex:character>27</ex:character>
  <ex:character>26</ex:character>
  <ex:character>25</ex:character>
  <ex:character>24</ex:character>
  <ex:character>23</ex:character>
  <ex:character>22</ex:character>
  <ex:character>21</ex:character>
  <ex:character>20</ex:character>
  <ex:character>19</ex:character>
  <ex:character>18</ex:character>
  <ex:character>17</ex:character>
  <ex:character>16</ex:character>
  <ex:character>15</ex:character>
  <ex:character>14</ex:character>
  <ex:character>13</ex:character>
  <ex:character>12</ex:character>
  <ex:character>11</ex:character>
  <ex:character>10</ex:character>
  <ex:character>9</ex:character>
  <ex:character>8</ex:character>
  <ex:character>7</ex:character>
  <ex:character>6</ex:character>
  <ex:character>5</ex:character>
  <ex:character>4</ex:character>
  <ex:character>3</ex:character>
  <ex:character>2</ex:character>
  <ex:character>1</ex:character>
  <ex:character>0</ex:character>
</ex:fiveBitDFI1661DUI001>
```

However, when I switch to LTR, the behaviour becomes hard to explain:

```
<?xml version="1.0"?>
<ex:fiveBitDFI1661DUI001 xmlns:ex="http://example.com">
  <ex:character>0</ex:character>
  <ex:character>0</ex:character>
  <ex:character>17</ex:character>
  <ex:character>4</ex:character>
  <ex:character>3</ex:character>
  <ex:character>10</ex:character>
  <ex:character>28</ex:character>
  <ex:character>24</ex:character>
  <ex:character>2</ex:character>
  <ex:character>2</ex:character>
  <ex:character>21</ex:character>
  <ex:character>12</ex:character>
  <ex:character>27</ex:character>
  <ex:character>26</ex:character>
  <ex:character>28</ex:character>
  <ex:character>25</ex:character>
  <ex:character>4</ex:character>
  <ex:character>12</ex:character>
  <ex:character>25</ex:character>
  <ex:character>20</ex:character>
  <ex:character>3</ex:character>
  <ex:character>11</ex:character>
  <ex:character>29</ex:character>
  <ex:character>26</ex:character>
  <ex:character>6</ex:character>
  <ex:character>14</ex:character>
  <ex:character>29</ex:character>
  <ex:character>28</ex:character>
  <ex:character>27</ex:character>
  <ex:character>27</ex:character>
  <ex:character>29</ex:character>
  <ex:character>27</ex:character>
</ex:fiveBitDFI1661DUI001>
```

bsloane1650: Setting aside the question of "correct way to do this", I shoved all the bits onto one line and swapped their order manually and used RTL mode.

I can see how the line breaks do not actually effect anything, but (as a human), It seems weird that RTL also means bottom-to-top.

mbeckerle: Yup, a byteOrder "RTL-per-line" feature that splits the data at line boundaries, then interprets each line as a documentPart with byteOrder "RTL". 

That might be more natural for people to use. The "RTL" feature was added specifically to make it possible to directly put a table from mil-std-2045 directly into a test case. E.g., see tests

- testMIL2045_47001D_Page70_TableB_I
- testMIL2045_47001D_Page70_TableB_I_with_string

in the daffodil-tdml-unittest module.  You may find the style they use helpful given you are dealing with formats which are mixtures of bit fields and these 5, 6, and 7 bit character set strings.

They make good use of the fact that only 0 and 1 are interpreted as contributing bits to the result, but you'll notice they have one documentPart per line for exactly the reasons expected - we want RTL, not bottom to top.

stevedlawrence:
+1

PR #246, opened 2019-06-24
WIP: Attempt to compile with scala 2.13.0

gdesrosi:
This branch contains changes to enable cross compilation with scala 2.13.0, but currently has quiet a few compilation errors if attempted to cross compile with scala 2.13.0.
Two major changes noted with the new version of scala is that:
1) the returned type of every method must explicitly be declared else it produces a syntax error that’s including void methods. Currently there are 92 compilations errors and almost if not half of them has to do with this new syntax requirements of void methods.
2)  scala has a predefined Node class which has a child method which previously used to return a mutable seq of all the child of that particular node, but in scala 2.13.0 that returned type is now immutable which causes type mismatch in at least one instance so far so to get around “toSeq” function on the child call to turn that seq back to a mutable seq as it was in the previous version.
Other changes noted with scala 2.13.0 include:
a)	Deprecation of the any2Stringadd object predefinition as well as the implicit injection of the + sign. This now causes problem when adding a file name which is a string to a Path without calling the toString method on the path because it’s no longer implied now.
b)	Stricter type checking when adding a character to a string you need to either use the Character.toString() method on the character or use string interpolation to add the two. In some instances, the string interpolation causes some issues when that value is passed to a regex and cause some illegalArgumentException.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,7 +104,7 @@ lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "2.4.0-SNAPSHOT",
   scalaVersion := "2.12.6",
-  crossScalaVersions := Seq("2.12.6", "2.11.12"),
+  crossScalaVersions := Seq("2.13.0", "2.12.6", "2.11.12"),
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation",
--- a/daffodil-japi/build.sbt
+++ b/daffodil-japi/build.sbt
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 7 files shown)


bsloane1650:
Re 2) .toSeq returns an immutable Seq.
Looking at the scala changelog https://github.com/scala/scala/releases/tag/v2.13.0 I believe what actually changed is the build it "Seq" type is now an alias for collection.immutable.Seq. Previously, it had meant collections.Seq, which includes mutable implementations.

mbeckerle:
Closed as Inactive. DAFFODIL-2152 has a link here for future reference.

PR #247, opened 2019-06-25
Fix unused properties related to implied choice sequences

stevedlawrence:
When a ChoiceBranchImpliedSequence was created, because groupMembers was
not overridden, it would actually create a complete duplicate of all the
children terms because that's what the ModelGroup groupMembers that it
inherits does. This meant that properties would be cached on different
terms than we checked for unused properties, which lead to warnings.

This overrides groupMembers for ChoiceBranchImpliedSequence so that it
does not create new terms, but is instead just a sequence of the Term it
wraps. This allows all the same behavior provided by the implied
sequence, but 1) removes the duplicate Terms and 2) allows correct
caching/checking of unused properties.

DAFFODIL-2163

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
@@ -358,6 +358,8 @@ final class ChoiceBranchImpliedSequence(rawGM: Term)
   extends SequenceTermBase(rawGM.xml, rawGM.optLexicalParent, rawGM.position)
   with SequenceDefMixin {
 
+  override final lazy val groupMembers: Seq[Term] = Seq(rawGM)
+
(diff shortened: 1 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -358,6 +358,8 @@ final class ChoiceBranchImpliedSequence(rawGM: Term)
   extends SequenceTermBase(rawGM.xml, rawGM.optLexicalParent, rawGM.position)
   with SequenceDefMixin {
 
+  override final lazy val groupMembers: Seq[Term] = LV('groupMembers) { Seq(rawGM) }.value
```
mbeckerle: does not need to be an LV. It's just taking a constructor arg. No error SDE can happen here that we want blamed/positioned  to this object, nor can anything "circular" happen. Those are the two reasons to make something an LV.

stevedlawrence: Fixed.

PR #248, opened 2019-06-25
Add debugger infosetParents config option and update trace mode

stevedlawrence:
This new config option determines how many parent elements to include
when running the "info infoset" debugger command. A value of -1 will
show the entire inforset, which is the current and default behavior. A
value of zero or more will show 0 or more parent elements, in addition
to the current element.

Also modify the trace mode commands to set this property to "1" to show
a single parent and use "info infoset" so show the infoset. This
replaces the "eval .." command which gave unfriendly results going past
the root element.

DAFFODIL-1931


stevedlawrence:
Yep, I'll make sure to update the doc page.

PR #249, opened 2019-06-25
Emit SDE when a type calculator attempts to use an undefined type.

bsloane1650:
This lays down the infrastructure which allows the DPath
interperator to lookup type calculators at compile time,
which will allow us to make their return type dependent on
the calculator they are using, instead of their functin name.

DAFFODIL-2148

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
@@ -17,14 +17,20 @@
 
 package org.apache.daffodil.dpath
 
-import org.apache.daffodil.exceptions._
-import org.apache.daffodil.dsom._
-import scala.xml.NamespaceBinding
-import org.apache.daffodil.xml._
-import scala.util.parsing.input.CharSequenceReader
+import java.math.{ BigDecimal => JBigDecimal }
+import java.math.{ BigInteger => JBigInt }
+
 import scala.util.parsing.combinator.RegexParsers
-import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
-import org.apache.daffodil.oolag.OOLAG._
+import scala.util.parsing.input.CharSequenceReader
+import scala.xml.NamespaceBinding
+
+import org.apache.daffodil.dsom.CompiledExpression
+import org.apache.daffodil.dsom.ConstantExpression
+import org.apache.daffodil.dsom.DPathCompileInfo
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.oolag.OOLAG.OOLAGHost
+import org.apache.daffodil.xml.NamedQName
+import org.apache.daffodil.xml.QNameRegex
 
 /**
  * Parses DPath expressions. Most real analysis is done later. This is
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
@@ -17,25 +17,29 @@
 
 package org.apache.daffodil.dsom
 
-import org.apache.daffodil.exceptions._
-import org.apache.daffodil.processors.VariableMap
-import org.apache.daffodil.processors.Suspension
-import org.apache.daffodil.util._
-import org.apache.daffodil.xml._
-import org.apache.daffodil.dpath._
-import org.apache.daffodil.xml.NamedQName
-import org.apache.daffodil.util.PreSerialization
-import org.apache.daffodil.dpath.NodeInfo.PrimType
-import org.apache.daffodil.processors.ParseOrUnparseState
-import org.apache.daffodil.util.TransientParam
-import org.apache.daffodil.util.Maybe
-import scala.runtime.ScalaRunTime.stringOf // for printing arrays properly.
-import org.apache.daffodil.exceptions.SchemaFileLocation
-import org.apache.daffodil.exceptions.HasSchemaFileLocation
-import org.apache.daffodil.processors.ParseOrUnparseState
+import scala.runtime.ScalaRunTime.stringOf
+
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.api.UnqualifiedPathStepPolicy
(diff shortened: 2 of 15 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -200,14 +205,23 @@ class DPathCompileInfo(
   val namespaces: scala.xml.NamespaceBinding,
   val path: String,
   override val schemaFileLocation: SchemaFileLocation,
-  val tunable: DaffodilTunables)
+  val tunable: DaffodilTunables,
+  @TransientParam val typeCalcMapArg: TypeCalcMap)
```
stevedlawrence: Does this variable need to be passed by-name? I forget why we had to do that, but all other TransientParams are by-name args.

bsloane1650: No. All passed-by-name args need to be transient; and that is normally the only reason we need to use TransientParam.

In this case, we are using TransientParam because (for an unknown reason) it is simply to possible to unserialize the value, so we avoid doing so. Like with the by-name parameters, we pull a similar trick of having a corresponding field which is serializable.

This is the same reason why the corresponding typeCalcMap field is not marked as lazy.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -36,6 +36,11 @@ import org.apache.daffodil.processors.ParseOrUnparseState
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.api.UnqualifiedPathStepPolicy
 import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.processors.SchemaSetRuntimeData
```
mbeckerle: Clean up imports. Lots of repeats here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
mbeckerle: I would have made the same observation as Steve. So this needs to go into the code as commentary so we understand that this transient param is different from others.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -429,10 +454,14 @@ object TypeCalculatorCompiler {
     }
   }
 
-  def compileExpression[A <: AnyRef, B <: AnyRef](optInputTypeCalc: Option[CompiledExpression[B]], optOutputTypeCalc: Option[CompiledExpression[A]], srcType: NodeInfo.Kind, dstType: NodeInfo.Kind): ExpressionTypeCalculator[A, B] = {
-    val maybeInputType: Maybe[CompiledExpression[B]] = optInputTypeCalc.map(Maybe(_)).getOrElse(Maybe.Nope)
-    val maybeOutputType: Maybe[CompiledExpression[A]] = optOutputTypeCalc.map(Maybe(_)).getOrElse(Maybe.Nope)
-    new ExpressionTypeCalculator(maybeInputType, maybeOutputType, srcType, dstType)
+  def compileExpression[A <: AnyRef, B <: AnyRef](
```
mbeckerle: I think we have too many overloads of compileExpression at this point. No reason to further proliferate that. Can this be called compileTypeCalcExpression ?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
bsloane1650: On second thought, there is no reason to not make this passed by-name. Even though the TransientParam is there for a different reason, spamming lazyness is still our strategy to avoid circular dependencies, and it can't hurt (much) to do so preemptivly.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -349,10 +366,19 @@ object TypeCalculatorCompiler {
...
-  def compileExpression[A <: AnyRef, B <: AnyRef](optInputTypeCalc: Option[CompiledExpression[B]], optOutputTypeCalc: Option[CompiledExpression[A]], srcType: NodeInfo.Kind, dstType: NodeInfo.Kind): ExpressionTypeCalculator[A, B] = {
-    val maybeInputType: Maybe[CompiledExpression[B]] = optInputTypeCalc.map(Maybe(_)).getOrElse(Maybe.Nope)
-    val maybeOutputType: Maybe[CompiledExpression[A]] = optOutputTypeCalc.map(Maybe(_)).getOrElse(Maybe.Nope)
-    new ExpressionTypeCalculator(maybeInputType, maybeOutputType, srcType, dstType)
+  /*
+   * compileExpression is already a highly overloaded name from the DPath expression compiler.
+   * While this technically overload that function, to avoid confusion, we are giving it a different
+   * name entirely.
+   */
+  def compileFromExpression[A <: AnyRef, B <: AnyRef](
```
stevedlawrence: Maybe rename it compileTypeCalculatorExpression or something to make it more clear how it differs from compileExpression?

daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
```diff
@@ -25,6 +25,8 @@ import scala.util.parsing.input.CharSequenceReader
 import scala.util.parsing.combinator.RegexParsers
 import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
 import org.apache.daffodil.oolag.OOLAG._
+import org.apache.daffodil.processors.TypeCalculator
+import org.apache.daffodil.processors.TypeCalculatorCompiler.TypeCalcMap
```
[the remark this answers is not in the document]
bsloane1650: Fixed.

I took this oppurtunity to just `refactor > organize imports`, so some of the other lines moved around as well.

bsloane1650:
rebased

PR #251, opened 2019-06-26
Fixes decoding unicode with surrogate pairs

olabusayoT:
Fixes decoding function for unicode characters that decode to a surrogate
pair

DAFFODIL-2123


stevedlawrence:
+1

PR #253, opened 2019-06-27
Support terminators and initiators containing %ES;

stevedlawrence:
- Modify DFA building code so that we actually compile %ES; into a DFA
  that matches anything. This allows us to use the same delimiter
  scanning code for finding ES. This also fixes the issue where a literal
  "%ES;" would be unparsed instead of an empty string.
- Modify the various delimiter scanning iterators so that they skip past
  %ES; delimiters at the appropriate time (e.g. delimiter scanning)
- Remove the logic in DelimiterParsers to no longer include special
  checks for ES since that is now handled by standard delimiter scanning
  logic and new delimiter iterator logic
- Update "delimiter not found" error to include all local delimiters
  rather than just the first
- Update DelimiterNoES cooker to mention that the no ES restriction only
  applies to lengthKind="delimited". This cooker is only used in the
  lengthKind="delimited" case.

DAFFODIL-1474, DAFFODIL-1477, DAFFODIL-1493


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DFDLDelimiter.scala
```diff
@@ -602,3 +619,15 @@ class WSPStarDelim extends WSPBase with WSP {
 
   def unparseValue(outputNewLine: String): String = ""
 }
+
+// This delimiter matches an empty space which effectively is always a
+// successful match without moving the character position. This effectively
+// acts as a maker that empty match is allowed
```
mbeckerle: Should be scaladoc.

Please add a little more description of why this isn't a stupid-nonsensical thing to do. That ES as a terminator when lengthKind is other than 'delimited' is a way of indicating an optional terminator, but that this DFA/scanner is used to consume delimiters in that case as well. This class establishes the invariant that there is always a DelimBase object used when "scanning", even when we're "scanning for nothing".

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DelimiterIterator.scala
```diff
@@ -17,85 +17,48 @@
...
-  // practicies. For exmaple, if you called di.next() twice in a row without
-  // calling hasNext, the second next() call would give you the same delimiter.
-  // So for this to work, you really do need to call hasNext before ever
-  // calling next. Although one should follow that practice with iterators, it
-  // isn't usually required.
+  // Note that this does not follow standard iterator practicies. hasNext is
+  // what moves the iterator to the next position, not next(). next() simply
+  // returns the current index in the iterator. So calling next() twice in a
+  // row will return the same thing. So it's important to call hasNext, but
+  // that's always how this is used.
```
mbeckerle: Should be scaladoc I think.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
```diff
@@ -735,3 +735,19 @@ class NLState(states: => ArrayBuffer[State], val nextState: Int, val stateNum: I
 
   def checkMatch(charIn: Char): Boolean = isNLNotCR(charIn) || isCR(charIn)
 }
+
+
+class ESState(states: => ArrayBuffer[State], val nextState: Int, val stateNum: Int)
```
mbeckerle: Add scaladoc making the point that this is a degenerate kind of rule that just succeeds immediately to handle the empty-string corner case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/DelimiterParsers.scala
```diff
@@ -74,57 +74,20 @@ class DelimiterTextParser(
...
+        Nope
+      }
 
-  override def parse(start: PState): Unit = {
     val foundDelimiter =
-      if (delimiterType == DelimiterTextType.Initiator || !start.delimitedParseResult.isDefined) {
-        //
-        // We are going to scan for delimiter text.
-        //
-        // FIXME: This is incorrect. It grabs all local delimiters even if we're last in a group so the
```
mbeckerle: I don't think you should remove this comment.
```
<sequence separator=";">
    <element foo ...>
    <sequence separator =",">
        <element bar ...>
        <element baz ...>
    </sequence>
   <element quux ...>
</sequence>
```

The above is effectively a 4-tuple foo;bar,baz;quux where the middle comma is the separator of the inner sequence. 
When scanning for the end of the baz element, we shoudlnt' be scanning for either "," or ";". Rather we should scan only for  ";". That's the bug this fixme is about. 

We should create a test for this I guess, and assuming my analysis is correct and this bug exists, create a JIRA ticket and comment out the test, marked by that ticket.

daffodil-test/src/test/scala/org/apache/daffodil/section06/entities/TestEntities.scala
```diff
@@ -92,16 +92,23 @@ class TestEntities {
...
 
   @Test def test_entity_fail_01() { runnerEntity.runOneTest("entity_fail_01") }
   @Test def test_entity_fail_02() { runnerEntity.runOneTest("entity_fail_02") }
 
-  // Regression - we used to just reject %ES; in terminators. Now we accept it, but it doesn't work
-  // right. JIRA DFDL-1477
-  // @Test def test_entity_fail_03() { runnerEntity.runOneTest("entity_fail_03") }
+  // DAFFODIL-1477
+  @Test def test_entity_fail_03a() { runnerEntity.runOneTest("entity_fail_03a") }
+  @Test def test_entity_fail_03b() { runnerEntity.runOneTest("entity_fail_03b") }
```
mbeckerle: With this fix, we need to modify schemas for mil-std-2045, which I think uses this, but especially the FOUO schemas which make heavy use of this technique. At which point those schemas become dependent on this fix, but ultimately, without this fix they're incorrect and fragile.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/DelimiterParsers.scala
stevedlawrence: Although I've confirmed that this bug does still exist, I think this comment is wrong, or at least it's in the wrong place. The purpose of DelimiterParser to to just consume delimiters, and it appears to do scana/consume the correct  delimiters correctly. For example, if I take the above schema and make all the lengths explicit, I've confirmed that it scans for what you would expect--"foo" scans only for a semicolon, "bar" scans for both the semicolon and comma, and "baz" scans for just the comma (no scanning is required for "quux").

However, I have confirmed that things fail when lengths are delimited. In fact, no actual scanning even happens in DelmiterParser in that case. All that scanning is done in StringDelimitedParser, which caches the result and then DelimiterParser just determines if i found an out of scope delimiter. So there's definitely a bug in StringDelimitedParser scanning for the wrong delimters. There might still be a bug somewhere in this parser, but that's where things start to go off the rails. And it's possible something is broken with how we build the delimiter stack.

So I think the issues are fundamental and spread out, and not just contained in this parser. So I'm not sure these comment belongs here any more, and I rather just have this information in a bug where we'll see it and keep track of it.

I'll add these two test (explicit lengths and delimited length scanning) and open new bug that has all the relevant information.

stevedlawrence:
Added new fixup commit to address scaladoc comments and add tests for existing incorrect delimiter scanning. DAFFODIL-2171 is the issue to fix that.

PR #254, opened 2019-06-27
Adds hex/utf-8 data dump on left over data

olabusayoT:
- Fixes dumpTextLine to decode whole characters rather than per byte
in keeping with the updates to converToCharRepr
- Limited to 8 bytes displayed and decoded
- Uses Java utf-8 decoder
- 1 based indexing

DAFFODIL 1387

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -95,6 +95,8 @@ import org.apache.commons.io.IOUtils
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.io.InputSourceDataInputStream
 import org.apache.daffodil.tdml.TDMLTestNotCompatibleException
+import org.apache.daffodil.io.DataDumper
+import java.nio.ByteBuffer
 
 class NullOutputStream extends OutputStream {
   override def close() {}
@@ -905,13 +907,23 @@ object Main extends Logging {
                     }
                   } else {
                     // not streaming, show left over data warning
+                    val Dump = new DataDumper
+                    val curPositionInByte0b = inStream.inputSource.position
+                    val bytesAvailable = inStream.inputSource.bytesAvailable
+                    val bytesLimit = math.min(8, bytesAvailable).toInt
+                    val destArray = new Array[Byte](bytesLimit)
+                    val destArrayFilled = inStream.inputSource.get(destArray, 0, bytesLimit)
+                    val dumpString = if (destArrayFilled) Dump.dump(Dump.TextOnly(Some("utf-8")), 0, destArray.length * 8, ByteBuffer.wrap(destArray), includeHeadingLine = false).mkString("\n") else ""
+                    val dataText = if (destArrayFilled) s"\nData (UTF-8) starting at byte ${curPositionInByte0b + 1} is: (${dumpString}...)" else ""
+                    val dataHex = if (destArrayFilled) s"\nData (Hex) starting at byte ${curPositionInByte0b + 1} is: (0x${destArray.map { a => f"$a%02x" }.mkString}...)" else ""
                     val remainingBits =
                       if (loc.bitLimit0b.isDefined) {
                         (loc.bitLimit0b.get - loc.bitPos0b).toString
                       } else {
-                        "at least " + (inStream.inputSource.bytesAvailable * 8)
+                        "at least " + (bytesAvailable * 8)
                       }
-                    log(LogLevel.Warning, "Left over data. Consumed %s bit(s) with %s bit(s) remaining.", loc.bitPos0b, remainingBits)
+                    val leftOverDataWarning = s"Left over data. Consumed ${loc.bitPos0b} bit(s) with ${remainingBits} bit(s) remaining." + dataText + dataHex
+                    log(LogLevel.Warning, leftOverDataWarning)
                     keepParsing = false
                     error = true
                   }
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
@@ -215,7 +215,7 @@ class DataDumper {
     Assert.usage(lengthInBytes >= 0)
 
     val (textDataHeader, textByteWidth, optEncName) = getTextParameters(optEncodingName)
-    val decoder = getReplacingDecoder(optEncName)
+    val decoder = getReportingDecoder(optEncName)
 
     val endByteAddress0b = math.max(startByteAddress0b + lengthInBytes - 1, 0)
     val addressHeader = """87654321  """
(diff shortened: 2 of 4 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
```diff
@@ -598,13 +599,13 @@ class DataDumper {
...
     // val cs = optEncodingName.map { Charset.forName(_) }
-    val decoder = getReplacingDecoder(optEncodingName)
+    val decoder = getReportingDecoder(optEncodingName)
     var i = startByteAddress0b
     val sb = new StringBuilder
     while (i <= endByteAddress0b) {
-      val (cR, _, _) = convertToCharRepr(i - startByteAddress0b, endByteAddress0b, byteSource, decoder)
-      sb += cR(0)
-      i += 1
+      val (cR, nBytesConsumed, _) = convertToCharRepr(i - startByteAddress0b, endByteAddress0b, byteSource, decoder)
```
mbeckerle: I wonder what will happen in this code if we have a text large enough to do debug dumps of, but it is in one of our non-8-bit charset encodings? 

Do we have tests of this, or alternatively, does this code exclude that possibility somewhere? (I imagine the latter is more likely. This kind of dump makes little sense to have text and hex if the characters are not using up units of whole bytes.

olabusayoT: It looks like we do have code that checks for non byte aligned encodings, and we default to per byte decoding based on windows-1252 in those cases. Depending on where this is called from, we'll either get text and hex (e.g trace output) or text only (e.g left over data dump).

It doesn't look like we have tests for decoding non-8 bit charset encodings.

The update to this code was intended to add multibyte sequence support to the textOnly dump as the original code didn't have support for that.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -905,13 +907,26 @@ object Main extends Logging {
                     }
                   } else {
                     // not streaming, show left over data warning
+                    val Dump = new DataDumper
+                    var (dataText, dataHex) = ("", "")
+                    var curPositionInByte0b = inStream.inputSource.position
+                    var bytesAvailable = inStream.inputSource.bytesAvailable
+                    var bytesLimit = math.min(8, bytesAvailable).toInt
+                    val destArray = new Array[Byte](bytesLimit)
+                    val ret = inStream.inputSource.get(destArray, 0, bytesLimit)
```
stevedlawrence: There are cases where this output might be misleading, primarly related when the left over bits do not start on a byte boundary. inputSource.get will get the raw underlying bytes, without regard to the bit offset. So if data ends mid-bit this is going to include those already read bits. I'm not sure of the best way to handle this. We doul use inStream.fillByteArray instead, which does take into account bits, but that stiill might be confusing showing full bytes that were offset by some amount of bits. But maybe this isn't a big deal?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -905,13 +907,26 @@ object Main extends Logging {
                     }
                   } else {
                     // not streaming, show left over data warning
+                    val Dump = new DataDumper
+                    var (dataText, dataHex) = ("", "")
+                    var curPositionInByte0b = inStream.inputSource.position
+                    var bytesAvailable = inStream.inputSource.bytesAvailable
+                    var bytesLimit = math.min(8, bytesAvailable).toInt
```
stevedlawrence: I think all these var's can be made vals. You'll have to slightly change how dataText and dataHex are set.

daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
stevedlawrence: getReplacingDecoder returns a JavaCharsetDecoder, so it can only decode full byte characters. I believe there's a bug to update the text dumper to support non-byte size charsets.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
olabusayoT: I think since the first line of the output informs the user about the consumed bits, the fact that the dump might include those bits might be ok. This dump is intended as a very rudimentary pointer at a possible issue in the data. The user could (and should) further analyze the issue if it isn't readily apparent from the dump, i think.

daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
olabusayoT: Right, we won't actually use the decoder though, due to the code in DataLoc.dump. It looks like it won't return a decoder if it's an unknown encoding or if the known encoding isn't byte aligned.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
olabusayoT: Changed made

daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
mbeckerle: I don't know if we should be trying to enhance this to produce a useful textual dump along side the byte dump for non-8-bit characters. Making that meaningful in any way is hard. I think one would have to do a binary (individual bits) dump, instead of hex to do that. Let's not do that. I think it is overkill.  Or rather, let's not do that as part of fixing THIS jira ticket, which is just about the left-over-data dump.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: It is an issue, and we debug a lot of data like this.

If we were going to try to address this, we could add to the dump a first-byte-of-dump key, which is only presented if the data ends mid-byte somwhere. It would be an individual-bits picture of the first byte of the left-over data. 
If say, the data ended at bit 5, and bit order is mostSignificantBitFirst, then this picture would look like xxxxx000 (where the 0's are the actual left over data bits in that first byte of the left over data.) If bitOrder is leastSignificantBitFirst the picture would look like 000xxxxx. 
I suggest this sort of enhancement would be best as a separate JIRA ticket however, as this is not thought through.

stevedlawrence:
Looks like something is failing to compile on scala 2.11, somehow related to scoverage: https://travis-ci.org/apache/incubator-daffodil/jobs/551770526

I've never seen that before...

olabusayoT:
> Looks like something is failing to compile on scala 2.11, somehow related to scoverage: https://travis-ci.org/apache/incubator-daffodil/jobs/551770526
> 
> I've never seen that before...

Turns out it was due to the lazy evaluation I had attempted on the DataDumper object. This breaks in 2.11, but not 2.12. Fixed now.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
olabusayoT: Created [DAFFODIL- 2173](https://issues.apache.org/jira/browse/DAFFODIL-2173 ) to account for this issue

daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
olabusayoT: As mentioned above, [DAFFODIL-378](https://issues.apache.org/jira/browse/DAFFODIL-378) was created to address this

PR #255, opened 2019-06-28
Replaces arrayIndex with occursIndex

olabusayoT:
Now the debugger uses "info occursIndex" instead of "info arrayIndex"

DAFFODIL 1478


stevedlawrence:
+1

PR #257, opened 2019-07-02
Daffodil 2165 type calc double functions

bsloane1650:
The scope expanded a bit beyond the original idea. This de-specializes all of the type-calc functions and instead infers the return type at compile time.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -30,6 +30,9 @@ import org.apache.daffodil.equality._
 import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
 import org.apache.daffodil.util.Numbers
 import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.infoset.NoNextElement
+import org.apache.daffodil.infoset.OnlyOnePossibilityForNextElement
+import org.apache.daffodil.infoset.SeveralPossibilitiesForNextElement
 
 /**
  * Root class of the type hierarchy for the AST nodes used when we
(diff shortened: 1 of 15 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
```diff
@@ -95,7 +99,7 @@ class CompiledDPath(val ops: RecipeOp*) extends Serializable {
         // in a way that is not consistent with a constant expression. Such as trying to do
         // anything with the infoset other than saving and restoring current position in the infoset.
         // Ditto trying to read a variable.
-        case _: InfosetException | _: VariableException | _: java.lang.IllegalStateException =>
+        case _: InfosetException | _: VariableException | _: NoSuchElementException | _: java.lang.IllegalStateException =>
```
stevedlawrence: What causes this NoSuchElementException? I thought we only want to catch Daffodil specific exceptions here, like the infoset stuff or variable stuff? NoSuchELementException implies to me that we have a bug somewhere and are accessing somethign we shouldn't.

bsloane1650: dfdlx:repTypeValue() and dfdlx:logicalTypeValue() attempt to read the current rep/logical value from a Maybe variable. At compile time, these are always Nope, and so throw a NoSuchElementException.

There is a good arguement to be made that we shouldn't catch NoSuchElementHere, as this will also catch bugs that shouldn't be caught. We should be able to avoid this by adding an if(isCompile) throw new IllegalStateException().

Having said that, we are currently using IllegalStateException as a sentinal exception, and that is not Daffodil specific. Those may all be used specifically as a sentinal when they are thrown, in which case it should be straightforward to replace them as well. Of course, it is much less likely for a random bug to throw IllegalStateException, so this is less imporant.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1553,46 +1556,91 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+            /*
+             * Verify that all next siblings have the same dstType
+             * In theory, we could loosen the requirement to having the same primitive type,
+             * but in the interest of being conservative, we will stick to the stricter interperatation for now
+             */
+            xs.map(other => {
+              val (otherAns, otherQName) = other
+              if(otherAns != ans){
+                SDE("dfdlx:outputTypeCalcNextSibling() requires that all the possible next siblings have the same "+
+                     "repType. However, the potential next siblings %s and %s have repTypes %s and %s respectivly",
```
mbeckerle: typo "respectively"

I found the corresponding test that exercises this error, and that is good to go as it doesn't test for this word.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1553,46 +1556,91 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+      case (RefQName(_, "repTypeValue", DFDLX), args) => {
+        val strd = compileInfo.lexicalContextRuntimeData match {
+          case strd: SimpleTypeRuntimeData => strd
+          case _ => {
+            SDE("dfdlx:repTypeValue() can only be defined on a simple type")
+          }
+        }
+        val repPrimType = strd.optRepPrimType match{
+          case Some(x) => x
+          case None => SDE("dfdlx:repTypeValue() used on type that does not have a repType")
```
mbeckerle: Missing period at end of sentence.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1706,6 +1754,43 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
     funcObj
   }
+
+  def lookupTypeCalculator(typeCalcName:Expression, requiresParse: Boolean, requiresUnparse: Boolean):TypeCalculator[AnyRef, AnyRef]={
+    /*
+     * Every functions that currently uses this defines the type calculator
+     * as their first arguement.
+     * We still take the calculators name as an arguement to avoid relying on this
+     * fact. 
+     */
```
mbeckerle: Confused by this comment. Why is the arg an Expression called typeCalcName? Why not call it typeCalcExpr ?

mbeckerle: Also "Every functions" grammar error

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
stevedlawrence: Yeah, I'd vote for changing it to an IllegalStateException like you mention, and maybe create a new bug to find those usages and change to somethign Daffodil specific so we don't hide actual bugs.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1706,6 +1754,43 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+    Assert.invariant(typeCalcName == expressions(0))
+    /*
+     * This lookup needs to occur before the main type-checker runs, because it is
+     * used to determine the type of this expression.
+     * As a result, we must do some type checking manually
+     * Also, we insist that typeCalcName is a constant, which is not even
```
mbeckerle: yeah.... you really do need to rename this argument. You can't even read these comments without doing doubletakes. "How could the name be non-constant...?" Oh it's the expression, not a name.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
mbeckerle: NoSuchElementException is a Daffodil exception isn't it? (Maybe we should rename it if that's ambiguous.) It is heavily used by Unparsing as the indicator that an expression which is forward referencing, is not ready to be evaluated yet and should be retried in the future.

At parse time the same exception is used to indicate an expression with a path step for which the corresponding element isn't in the infoset (yet at least - it has to be possible or the expression wouldn't compile, but for whatever reason the parse path didn't put it there when the expression was evaluated at runtime.) 

When doing constant folding, if you touch the infoset at all, you get NoSuchElementException, or so I thought, and that indicates, well, that you were touching the infoset so the expression is non-constant. (Or close enough anyway.)

mbeckerle: Ah. I see NoSuchElementException is a scala wrapper on a java exception. We should have our own for sure.

mbeckerle: Nevermind. I am mis-remembering again. Deleting prior comments.

stevedlawrence:
Looks like there's a compilation error in SimpelTypes.scala, incorrect number of arugments for compileTypeCalculatorFromExpression.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
bsloane1650: Bug filed: https://issues.apache.org/jira/projects/DAFFODIL/issues/DAFFODIL-2176

PR #258, opened 2019-07-02
Adds bitawareness to left over data dump

olabusayoT:
If a parse partially consumes a byte, the left over data dump prints
the rest of the consumed bytes as binary output(0bxx101011/0b101010xx)
depending on bitOrder with the x's representing already consumed bytes

Sample Output:
[warning] Left over data. Consumed 10 bit(s) with at least 16 bit(s) remaining.
Left over data (Hex) at byte 2 is: (0b011101xx)
Left over data (Hex) starting at byte 3 is: (0x7269...)
Left over data (UTF-8) starting at byte 3 is: (ri...)

DAFFODIL-2173

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
@@ -816,8 +816,8 @@ class TestCLIparsing {
       val cmd = String.format("echo 0,1,2,3,,,,| %s -t parse -s %s", Util.binPath, testSchemaFile)
       shell.sendLine(cmd)
       shell.expectIn(1, contains("Left over data. Consumed 56 bit(s) with at least"))
-      shell.expectIn(1, contains("Data (UTF-8) starting at byte 8 is: ("))
-      shell.expectIn(1, contains("Data (Hex) starting at byte 8 is: ("))
+      shell.expectIn(1, contains("Left over data (Hex) starting at byte 8 is: ("))
+      shell.expectIn(1, contains("Left over data (UTF-8) starting at byte 8 is: ("))
       shell.sendLine("exit")
       shell.expect(eof)
     } finally {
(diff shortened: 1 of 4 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -835,8 +835,64 @@ class TestCLIparsing {
...
+  @Test def test_CLI_Parsing_BitParse_LSBPartialByte_leftOverData() {
+    val schemaFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/bits_parsing.dfdl.xsd")
+    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+    val shell = Util.start("", true)
+
+    try {
+      val cmd = String.format("echo -n stri| %s parse -s %s -r leastFirst", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("Left over data. Consumed 10 bit(s) with at least 16 bit(s) remaining."
+        + "\nLeft over data (Hex) at byte 2 is: (0b011101xx)"
```
mbeckerle: So I think this is misleading a bit. This isn't hex.  True you have "0b" after this, but the two juxtaposed is confusing. Technically it is ambiguous since 0B011101 is valid hex. but you meant 0b as binary indicator followed by 8 bits with some as "x". 

Probably there should be a line the whole purpose of which is to describe the partial first byte when there is one. 
It should say "Left over data starts with partial byte.  Left over data in the first byte (binary) is (0b011101xx)." 

The subsequent lines can present the rest of the left-over data in Hex, or text or both.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -908,21 +911,36 @@ object Main extends Logging {
...
+                      val bitsToDisplay = 8 - bitsAlreadyConsumed
+                      val pbp = inStream.inputSource.position + 1
+                      inStream.fillByteArray(firstByteBitArray, bitsToDisplay, finfo)
+                      val fbs = firstByteBitArray(0).toBinaryString.takeRight(8).reverse.padTo(8, '0').reverse
+                      val bits = if (finfo.bitOrder == BitOrder.MostSignificantBitFirst) {
+                        "x" * bitsAlreadyConsumed + fbs.dropRight(bitsAlreadyConsumed)
+                      } else {
+                        fbs.takeRight(bitsToDisplay) + "x" * bitsAlreadyConsumed
+                      }
+                      val dumpString = f"\nLeft over data (Hex) at byte $pbp is: (0b$bits)"
```
mbeckerle: This is I think where you change the message to indicate binary data and emphasize the special first byte.

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
olabusayoT: Changes made

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -908,21 +911,36 @@ object Main extends Logging {
                   } else {
                     // not streaming, show left over data warning
                     val Dump = new DataDumper
-                    val curPositionInByte0b = inStream.inputSource.position
+                    val firstByteBitArray = new Array[Byte](1)
+                    val bitsAlreadyConsumed = loc.bitPos0b % 8
+                    val firstByteString = if (bitsAlreadyConsumed != 0) {
+                      val bitsToDisplay = 8 - bitsAlreadyConsumed
+                      val pbp = inStream.inputSource.position + 1
+                      inStream.fillByteArray(firstByteBitArray, bitsToDisplay, finfo)
```
stevedlawrence: Is it possible to use ``inStream.getByteArray()`` instead of making fillByteArray public? It's probably not a big deal if that's made public, but ideally it should only be used internally since it doesn't do any sanity checking like other functions do.

olabusayoT: Yes, I missed that function during the first go round. I've made the updates!

stevedlawrence:
+1, please squash and it's good to merge :+1:

PR #259, opened 2019-07-08
Incremental progress on schema compilation space/speed issue.

mbeckerle:
This change set reorganizes expression compilation so that
expressions are compiled taking all possible contexts where the expression
is reused into account, rather than compiling a separate copy for each
expression location.

This does NOT yet eliminate the excessive copying of objects in the
schema compiler. It does eliminate expressions from the set of things
preventing copy-elimination by removing the problem caused by
up-and-out relative paths.

DAFFODIL-1444

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Conversions.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Conversions.scala
@@ -18,6 +18,7 @@
 package org.apache.daffodil.dpath
 
 import org.apache.daffodil.exceptions.ThrowsSDE
+import org.apache.daffodil.exceptions.Assert
 
 /*
  * Casting chart taken from http://www.w3.org/TR/xpath-functions/#casting, with
@@ -43,22 +44,18 @@ object Conversion {
    * to convert the type. ST is the start type, the type the recipe has as its
    * output type. TT is the target type desired.
    */
-  def conversionOps(st: NodeInfo.Kind, tt: NodeInfo.Kind, context: ThrowsSDE): List[RecipeOp] = {
+  def conversionOps(st: NodeInfo.Kind, tt: NodeInfo.Kind, context: Expression): List[RecipeOp] = {
     import NodeInfo._
     val ops: List[RecipeOp] = (st, tt) match {
       case (_, Array) => Nil
       case (x, y) if (x == y) => Nil
-      case (x, Nillable) => {
-        // we don't have enough information here to check whether the
-        // item under scrutiny is nillable or not.
-        Nil
-      }
+
       case (Nothing, x) => Nil
       case (x, AnyType) => Nil
-      case (x: AnySimpleType.Kind, AnySimpleExists) => Nil
       case (x: AnyAtomic.Kind, AnyAtomic) => Nil
       case (x: Numeric.Kind, String) => List(NumericToString)
 
+      case (_: Numeric.Kind, Double) => List(NumericToDouble)
       case (Numeric, _) => NumericToString +: conversionOps(String, tt, context)
       //
       // Because of round-half-to-even, we need to know what the type of the original
@@ -75,6 +72,7 @@ object Conversion {
       case (HexBinary, String) => List(HexBinaryToString)
       case (String, NonEmptyString) => Nil
       case (x, NonEmptyString) => conversionOps(st, String, context)
+      case (x: AnyAtomic.Kind, String) => List(AnyAtomicToString)
       case (x, ArrayIndex) => conversionOps(st, Long, context) ++ List(LongToArrayIndex)
 
       case (Integer, Boolean) => IntegerToDecimal +: conversionOps(Decimal, tt, context)
@@ -96,7 +94,7 @@ object Conversion {
       case (Boolean, UnsignedShort) => BooleanToLong +: conversionOps(Long, tt, context)
       case (Boolean, Byte) => BooleanToLong +: conversionOps(Long, tt, context)
       case (Boolean, UnsignedByte) => BooleanToLong +: conversionOps(Long, tt, context)
-      case (Boolean, String) => List(BooleanToString)
+
       case (Date, DateTime) => List(DateToDateTime)
       case (Double, Decimal) => List(DoubleToDecimal)
       case (Double, Float) => List(DoubleToFloat)
@@ -111,7 +109,6 @@ object Conversion {
 
       case (Decimal, Integer) => List(DecimalToInteger)
       case (Decimal, NonNegativeInteger) => List(DecimalToNonNegativeInteger)
-      case (Decimal, Double) => List(DecimalToDouble)
       case (Decimal, Float) => DecimalToDouble +: conversionOps(Double, tt, context)
       case (Decimal, Long) => List(DecimalToLong)
       case (Decimal, UnsignedLong) => List(DecimalToUnsignedLong)
@@ -161,10 +158,8 @@ object Conversion {
       case (Long, UnsignedShort) => List(LongToUnsignedShort)
       case (Long, Byte) => List(LongToByte)
       case (Long, UnsignedByte) => List(LongToUnsignedByte)
-      case (Long, Double) => List(LongToDouble)
       case (Long, Float) => LongToDouble +: conversionOps(Double, tt, context)
 
-      case (Int, Double) => IntToLong +: conversionOps(Long, tt, context)
       case (Int, Decimal) => IntToLong +: conversionOps(Long, tt, context)
       case (Int, Float) => IntToLong +: conversionOps(Long, tt, context)
       case (Int, Integer) => IntToLong +: conversionOps(Long, tt, context)
@@ -177,7 +172,6 @@ object Conversion {
       case (Int, Byte) => IntToLong +: conversionOps(Long, tt, context)
       case (Int, UnsignedByte) => IntToLong +: conversionOps(Long, tt, context)
 
-      case (UnsignedInt, Double) => UnsignedIntToLong +: conversionOps(Long, tt, context)
       case (UnsignedInt, Decimal) => UnsignedIntToLong +: conversionOps(Long, tt, context)
       case (UnsignedInt, Float) => UnsignedIntToLong +: conversionOps(Long, tt, context)
       case (UnsignedInt, Integer) => UnsignedIntToLong +: conversionOps(Long, tt, context)
@@ -190,7 +184,6 @@ object Conversion {
       case (UnsignedInt, Byte) => UnsignedIntToLong +: conversionOps(Long, tt, context)
       case (UnsignedInt, UnsignedByte) => UnsignedIntToLong +: conversionOps(Long, tt, context)
 
-      case (UnsignedLong, Double) => UnsignedLongToLong +: conversionOps(Long, tt, context)
       case (UnsignedLong, Decimal) => UnsignedLongToLong +: conversionOps(Long, tt, context)
       case (UnsignedLong, Float) => UnsignedLongToLong +: conversionOps(Long, tt, context)
       case (UnsignedLong, Integer) => UnsignedLongToLong +: conversionOps(Long, tt, context)
@@ -202,7 +195,6 @@ object Conversion {
       case (UnsignedLong, Byte) => UnsignedLongToLong +: conversionOps(Long, tt, context)
       case (UnsignedLong, UnsignedByte) => UnsignedLongToLong +: conversionOps(Long, tt, context)
 
-      case (ArrayIndex, Double) => ArrayIndexToLong +: conversionOps(Long, tt, context)
       case (ArrayIndex, Decimal) => ArrayIndexToLong +: conversionOps(Long, tt, context)
       case (ArrayIndex, Float) => ArrayIndexToLong +: conversionOps(Long, tt, context)
       case (ArrayIndex, Integer) => ArrayIndexToLong +: conversionOps(Long, tt, context)
@@ -216,7 +208,6 @@ object Conversion {
       case (ArrayIndex, Byte) => ArrayIndexToLong +: conversionOps(Long, tt, context)
       case (ArrayIndex, UnsignedByte) => ArrayIndexToLong +: conversionOps(Long, tt, context)
 
-      case (Short, Double) => ShortToLong +: conversionOps(Long, tt, context)
       case (Short, Decimal) => ShortToLong +: conversionOps(Long, tt, context)
       case (Short, Float) => ShortToLong +: conversionOps(Long, tt, context)
       case (Short, Integer) => ShortToLong +: conversionOps(Long, tt, context)
@@ -229,7 +220,6 @@ object Conversion {
       case (Short, Byte) => ShortToLong +: conversionOps(Long, tt, context)
       case (Short, UnsignedByte) => ShortToLong +: conversionOps(Long, tt, context)
 
-      case (UnsignedShort, Double) => UnsignedShortToLong +: conversionOps(Long, tt, context)
       case (UnsignedShort, Decimal) => UnsignedShortToLong +: conversionOps(Long, tt, context)
       case (UnsignedShort, Float) => UnsignedShortToLong +: conversionOps(Long, tt, context)
       case (UnsignedShort, Integer) => UnsignedShortToLong +: conversionOps(Long, tt, context)
@@ -242,7 +232,6 @@ object Conversion {
       case (UnsignedShort, Byte) => UnsignedShortToLong +: conversionOps(Long, tt, context)
       case (UnsignedShort, UnsignedByte) => UnsignedShortToLong +: conversionOps(Long, tt, context)
 
-      case (Byte, Double) => ByteToLong +: conversionOps(Long, tt, context)
       case (Byte, Decimal) => ByteToLong +: conversionOps(Long, tt, context)
       case (Byte, Float) => ByteToLong +: conversionOps(Long, tt, context)
       case (Byte, Integer) => ByteToLong +: conversionOps(Long, tt, context)
@@ -255,7 +244,6 @@ object Conversion {
       case (Byte, Byte) => ByteToLong +: conversionOps(Long, tt, context)
       case (Byte, UnsignedByte) => ByteToLong +: conversionOps(Long, tt, context)
 
-      case (UnsignedByte, Double) => UnsignedByteToLong +: conversionOps(Long, tt, context)
       case (UnsignedByte, Decimal) => UnsignedByteToLong +: conversionOps(Long, tt, context)
       case (UnsignedByte, Float) => UnsignedByteToLong +: conversionOps(Long, tt, context)
       case (UnsignedByte, Integer) => UnsignedByteToLong +: conversionOps(Long, tt, context)
@@ -281,10 +269,75 @@ object Conversion {
       case (_, AnyAtomic) => Nil // is this correct?
 
       case (_, Exists) => Nil
-      case (_, other) =>
-        context.SDE("The type %s cannot be converted to %s.", st.name, tt.name)
+      case (_, other) => {
+        val (relevantExpr: Expression, detailMsg: String) =
+          polymorphicExpressionDiagnostics(st, tt, context)
+        context.SDE(
+          "In expression %s, the type %s cannot be converted to %s.%s",
+          relevantExpr.text,
+          st.globalQName.toQNameString,
+          tt.globalQName.toQNameString,
+          detailMsg)
+      }
     }
     ops
+  }
+
+  /**
+   * Determines a better expression for use in diagnostic messages, and computes
+   * a detailed diagnostic message with one line per use of the polymorphic expression
+   * up to 4 lines of detail. (Beyond that are suppressed/ignored.)
+   */
+  def polymorphicExpressionDiagnostics(
+    st: NodeInfo.Kind,
+    tt: NodeInfo.Kind,
+    context: Expression): (Expression, String) = {
+    val (relevantExpr: Expression, detailMsg: String) =
+      context match {
+        case stepE: StepExpression => {
+          val typeNodeGroups = stepE.stepElements.groupBy { _.typeNode }
+          Assert.invariant(typeNodeGroups.size > 0)
+          val rpp = stepE.relPathParent
+          val wholePath =
+            if (rpp.isAbsolutePath)
+              rpp.parent.asInstanceOf[RootPathExpression]
+            else
+              rpp
+          if (typeNodeGroups.size > 1) {
+            //
+            // types are not consistent for all step elements
+            //
+
+            val detailStrings: Seq[String] = {
+              typeNodeGroups.flatMap {
+                case (tn, cis) =>
+                  val tname = tn.globalQName.toQNameString
+                  val perUsePointStrings = cis.map { ci =>
+                    val sfl = ci.schemaFileLocation.locationDescription
+                    val qn = ci.namedQName.toQNameString
+                    val msg = "element %s in expression %s with %s type at %s".format(
+                      qn, wholePath.text, tname, sfl)
+                    msg
+                  }
+                  perUsePointStrings
+              }
+            }.toSeq
+            val detailPart: Seq[String] =
+              if (detailStrings.length > 4)
+                detailStrings.take(4) :+ "..."
+              else
+                detailStrings
+            val detailMessage: String =
+              "\nThe type is %s due to:\n%s".format(
+                st.name, detailPart.mkString("\n"))
+            (wholePath, detailMessage)
+          } else {
+            (wholePath, "")
+          }
+        }
+        case _ => (context, "")
+      }
+    (relevantExpr, detailMsg)
   }
 
 }
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -33,6 +33,7 @@ import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.infoset.NoNextElement
 import org.apache.daffodil.infoset.OnlyOnePossibilityForNextElement
 import org.apache.daffodil.infoset.SeveralPossibilitiesForNextElement
+import org.apache.daffodil.util.LogLevel
 
 /**
  * Root class of the type hierarchy for the AST nodes used when we
@@ -100,7 +101,10 @@ abstract class Expression extends OOLAGHostImpl()
   //
   // override for other checking beyond what is needed to do conversions
   //
-  protected def checkTypeCorrectness = true
+  protected lazy val checkTypeCorrectness =
+    children.forall { child =>
+      child.isTypeCorrect
+    }
 
   def text: String
 
@@ -151,17 +155,15 @@ abstract class Expression extends OOLAGHostImpl()
   lazy val compileInfo: DPathCompileInfo = // override in WholeExpression
     parent.compileInfo
 
-  final lazy val enclosingElementCompileInfo: Option[DPathElementCompileInfo] =
-    compileInfo.enclosingElementCompileInfo
+  final lazy val enclosingElementCompileInfos: Seq[DPathElementCompileInfo] =
+    compileInfo.enclosingElementCompileInfos
 
   final lazy val rootElement: DPathElementCompileInfo = {
-    enclosingElementCompileInfo.map {
-      _.rootElement
-    }.getOrElse {
-      compileInfo.elementCompileInfo.getOrElse {
-        Assert.invariantFailed("root doesn't have compile info")
-      }
+    val ecis = compileInfo.elementCompileInfos
+    if (ecis.length == 0) {
+      Assert.invariantFailed("Element doesn't have compile info")
     }
+    ecis.head.rootElement
   }
 
   lazy val namespaces: NamespaceBinding = parent.namespaces
@@ -250,7 +252,7 @@ trait BinaryExpMixin { self: ExpressionLists =>
   def left = children(0)
   def right = children(1)
   def op: String
-  def text = left.text + " " + op + " " + right
+  def text = left.text + " " + op + " " + right.text
 }
 
 trait BooleanExpression extends BinaryExpMixin {
@@ -616,18 +618,24 @@ case class WholeExpression(
     }
 
     if (!allowCoercion) {
+      val (relevantExpr: Expression, detailMsg: String) =
+        Conversion.polymorphicExpressionDiagnostics(inherentType, targetType, subExpr)
       if (tunable.allowExpressionResultCoercion) {
         SDW(
           WarnID.DeprecatedExpressionResultCoercion,
-          "Expression result type (%s) should be manually cast to the expected type (%s) with the appropriate constructor. " +
-            "Performing deprecated automatic conversion.",
+          "In expression %s, result type (%s) should be manually cast to the expected type (%s) with the appropriate constructor." +
+            "Performing deprecated automatic conversion.%s",
+          relevantExpr.text,
           inherentType,
-          targetType)
+          targetType,
+          detailMsg)
       } else {
         SDE(
-          "Expression result type (%s) must be manually cast to the expected type (%s) with the approrpriate constructor.",
+          "In expression %s, result type (%s) must be manually cast to the expected type (%s) with the approrpriate constructor.%s",
+          relevantExpr.text,
           inherentType,
-          targetType)
+          targetType,
+          detailMsg)
       }
     }
 
@@ -854,9 +862,28 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
   }
   requiredEvaluations(priorStep)
   requiredEvaluations(compileInfo)
-  requiredEvaluations(stepElement)
+  requiredEvaluations(stepElements)
+
+  def verifyQNames(cis: Seq[DPathElementCompileInfo]): Unit = {
+    cis.foreach { ci =>
+      if (!ci.namedQName.matches(stepQName))
+        ci.noMatchError(stepQName)
+    }
+  }
 
   // override def toString = text
+
+  // Combination of lazy val and a protected def is an idiom
+  // that enables a lazy calculation to call super.
+  //
+  final lazy val stepElements = {
+    val res = stepElementDefs
+    //    log(LogLevel.Info, "DPath Compiler: path step elements for %s are %s.",
+    //      step, res.map { _.sscd }.mkString(", "))
+    res
+  }
+
+  protected def stepElementDefs: Seq[DPathElementCompileInfo]
 
   // Note: all instances are distinct regardless of contents.
   override def equals(x: Any) = x match {
@@ -866,8 +893,8 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
 
   override def hashCode() = super.hashCode()
 
-  override def hasReferenceTo(elem: DPathElementCompileInfo): Boolean = {
-    stepElement =:= elem
+  final override def hasReferenceTo(elem: DPathElementCompileInfo): Boolean = {
+    stepElements.contains(elem)
   }
 
   lazy val stepQName = {
@@ -879,8 +906,6 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
   }
 
   override lazy val children = pred.toList
-
-  def stepElement: DPathElementCompileInfo
 
   lazy val priorStep: Option[StepExpression] = {
     val res =
@@ -919,7 +944,16 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
     res
   }
 
-  lazy val isArray: Boolean = stepElement.isArray
+  final lazy val isArray: Boolean = {
+    val (arrays, scalars) = stepElements.partition { _.isArray }
+    (arrays.length, scalars.length) match {
+      case (a, s) if (a > 0 && s > 0) =>
+        arrays.head.SDE("Path step is ambiguous. It can be to an array or a non-array element.\n" +
+          "One of the non-arrays is %s", scalars.head.schemaFileLocation.toString)
+      case (a, s) if (a == 0) => false
+      case _ => true
+    }
+  }
 
   /**
    * Used when there is no match as part of error diagnostic message.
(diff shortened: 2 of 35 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1103,64 +1104,123 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
...
 
   override def text = step
 
-  private def die = {
-    Assert.invariantFailed("should have thrown")
-  }
   /*
-   * The ERD of the element that corresponds to this path step
-   * (or SDE trying to find it.)
+   * The ERDs of the elements that corresponds to this path step
```
mbeckerle: Grammar error.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1103,64 +1104,123 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
...
       } else {
-        // since we're first we start from the element, or nearest enclosing
-        val nc = compileInfo.elementCompileInfo.getOrElse {
-          // happens for example if you have defaultValue="false" since false looks like a path step, but is really illegal. should be fn:false().
-          compileInfo.SDE("The expression path step '%s' has no defined enclosing element.", s)
-        }.findNamedChild(stepQName, this)
-        nc
+        // This is not first step so we are extension of prior step
+        //
+        //        // First we have to walk up any chain of relative upward path steps.
```
mbeckerle: Delete commented code.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1103,64 +1104,123 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
...
-        psse
-      }.map { se =>
-        val nc = se.findNamedChild(stepQName, this)
-        nc
-      }.getOrElse(die)
-      e
-    }
-    stepElem
-  }
+    // System.err.println("Named step " + this.s + " stepElements are: " + res.mkString(", "))
```
mbeckerle: Remove.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ComplexTypes.scala
```diff
@@ -70,20 +78,17 @@ abstract class ComplexTypeBase(xmlArg: Node, parentArg: SchemaComponent)
...
     }
-  }.value
+  }
 
   // provides needed polymorphism across unannotated complex types, and
   // the annotated objects.
-  lazy val localAndFormatRefProperties: Map[String, String] = {
-    Map.empty[String, String]
-  }
+  //  lazy val localAndFormatRefProperties: Map[String, String] = {
```
mbeckerle: Remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLEscapeScheme.scala
```diff
@@ -84,7 +84,8 @@ final class DFDLEscapeScheme(node: Node, decl: AnnotatedSchemaComponent, defES:
 
   final def escapeCharacterEv = LV('escapeCharacterEv) {
     val qn = this.qNameForProperty("escapeCharacter")
-    val expr = ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, escapeCharacterRaw, this, dpathCompileInfo)
+    val expr = ExpressionCompilers.String.compileProperty(qn, NodeInfo.NonEmptyString, escapeCharacterRaw, this,
+      defES.pointOfUse.dpathCompileInfo)
```
mbeckerle: Add doc. This use of the back pointer pointOfUse is *not* a problem to be remedied, or at least is not a first-order concern for improving the schema compiler. The expressions that are the escapeCharacter and escapeEscapeCharacter, like all expressions, have to have the DPathCompileInfo of the term where they apply, because they can contain relative paths, and those relative paths affect compilation of the expression. So an escape scheme object is effectively part of the state of the term which has that escape scheme. 

We could share more here if we instead just made the compileInfo  of the expressions passed in, rather than copying the whole object.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -281,12 +281,13 @@ trait ElementBase
...
   /**
-   * This is the compile info for this element. Since this might be an
-   * element ref, we optionally carry the compile info for the referenced
-   * element in that case.
+   * This is the compile info for this element term.
    */
   lazy val dpathElementCompileInfo: DPathElementCompileInfo = {
-    val ee = enclosingElement
+    val ee = enclosingElements
+    //    System.err.println("ElementBase: element " + this.shortSchemaComponentDesignator + (
```
mbeckerle: Remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -406,7 +411,7 @@ trait ElementBase
    * To be properly constructed, scala's xml Elems must share the scope (namespace bindings) of the enclosing
    * parent element, except when it adds more of its own bindings, in which case the tail is supposed to be shared.
    */
-  private lazy val minimizedScope: NamespaceBinding = {
+  private lazy val minimizedScope: NamespaceBinding = LV('minimizedScope) {
```
mbeckerle: Minimized scope isn't correct any more because a shared definition may be reused in multiple contexts that have different enclosing sets of namespace prefix definitions. 

Really we need a table here, so we can lookup the distinct minimizedScope for that context.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/GlobalElementDeclFactory.scala
```diff
@@ -47,6 +48,10 @@ class GlobalElementDeclFactory(xmlArg: Node, schemaDocumentArg: SchemaDocument)
 
   def forElementRef(eRef: AbstractElementRef) = {
     new GlobalElementDecl(xml, schemaDocument, eRef, this)
+    // memoized(eRef)
   }
 
+  private val memoized =
```
mbeckerle: unused code. Remove. Put back once we start using it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -57,41 +58,54 @@ object ModelGroupFactory {
...
-          val pos = lexicalParent match {
-            case ct: ComplexTypeBase => 1
-            case mg: ModelGroup => position
-            case gd: GlobalGroupDef => position
-          }
-          val isH = isHidden || lexicalParent.isHidden
-          val groupRefFactory = new GroupRefFactory(child, lexicalParent, pos, isH)
-          val groupRefInstance = groupRefFactory.groupRef
-          groupRefInstance.asModelGroup
+  private lazy val memoizedApply =
```
mbeckerle: Unused code. Remove and put back once we start really using it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -132,12 +135,35 @@ trait SchemaComponent
...
     ee
+  }
+
+  /**
+   * Elements that enclose this.
+   *
+   * If this is already an element, this still walks outward to find the
+   * next tier out.
+   */
+  final lazy val enclosingElements: Seq[ElementBase] = LV('enclosingElements) {
```
mbeckerle: Try moving enclosingElements, enclosingTerms, and enclosingComponents to the factory objects. Can still call them here and delegate to the factory, but there is no reason for them to reside on the instance objects.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -132,12 +135,35 @@ trait SchemaComponent
...
+      val ee = et match {
+        case eb: ElementBase => Seq(eb)
+        case sc: SchemaComponent => {
+          val scee = sc.enclosingElements
+          scee
+        }
+      }
+      ee
+    }
+    //    System.err.println("enclosingElements: Component " + this.shortSchemaComponentDesignator + (
```
mbeckerle: remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -121,9 +123,10 @@ trait SchemaComponent
   /**
    * All schema components except the root have an enclosing element.
    */
-  final lazy val enclosingElement: Option[ElementBase] = LV('enclosingElement) {
+  @deprecated("2019-06-03", "Use enclosingElements and deal with shared object with multiple referencers/enclosers.")
```
mbeckerle: Deprecations - comment out again so we don't get the slew of warnings.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -148,17 +174,30 @@ trait SchemaComponent
...
   }
 
+  /**
+   * The terms that can enclose this.
+   *
+   * Even if this is already a term, this walks outward to find
+   * those enclosing this.
+   */
   final lazy val enclosingTerms: Seq[Term] = {
+    //    if (this.shortSchemaComponentDesignator == "e=ex:a:ct:s:e2={}c:ct:s")
```
mbeckerle: remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -148,17 +174,30 @@ trait SchemaComponent
...
-        case _ => Assert.invariantFailed("Should only be Term or ComplexType")
+        case sd: SchemaDocument => Nil
+        case other => other.enclosingComponents.map { _.encloser }.flatMap {
+          case t: Term => Seq(t)
+          case x => x.enclosingTerms
+        }
       }
     }
-    et
+    //    System.err.println("enclosingTerms: Component " + this.shortSchemaComponentDesignator + (
```
mbeckerle: remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -304,18 +306,18 @@ trait Term
...
-        false
-      }
-      case Some(s) => {
-        if (s.isOrdered) {
-          val result = s.inUnorderedSequence
-          result
-        } else true
-      }
-    }
+  //  final lazy val inUnorderedSequence: Boolean =
```
mbeckerle: Delete unused code.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -396,18 +398,18 @@ trait Term
...
-        val eeGroup = ee.complexType.group
-        val res =
-          laterSiblings ++
-            (if (et eq eeGroup) Nil
-            else et.laterSiblingsWithinEnclosingElement)
-        res
-      }
-    }.getOrElse(Nil)
-  }
+  //  final lazy val laterSiblingsWithinEnclosingElement: Seq[Term] = {
```
mbeckerle: delete

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -240,57 +240,44 @@ class DPathCompileInfo(
...
-    val rel = elementCompileInfo.map {
-      parentInfo =>
-        Assert.invariant(path.startsWith(parentInfo.path))
-        path.substring(parentInfo.path.length)
-    }
-    val s = rel.getOrElse(path)
-    val res = if (s.startsWith("::")) s.substring("::".length) else s
-    res
-  }
+  //  final def relativePath: String = {
```
mbeckerle: remove

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -350,14 +336,18 @@ class DPathElementCompileInfo(
...
-      case Some(e) => e.enclosingElementPath :+ e
-    }
-  }
+    if (elementCompileInfos.isEmpty) this
+    else if (enclosingElementCompileInfos.isEmpty) this
+    else
+      // else if (elementCompileInfos.head eq this) this
+      enclosingElementCompileInfos.head.rootElement
+
+  //  final def enclosingElementPath: Seq[DPathCompileInfo] = {
```
mbeckerle: remove

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -2012,7 +2012,10 @@
       <tdml:documentPart type="text">abcd</tdml:documentPart>
     </tdml:document>
```
mbeckerle: Add a test that uses an expression like xs:string(../../foo/bar) where the expression is reused in contexts where ../../foo/bar have different contexts, and types. Since the path is just being passed to xs:string, this is polymorphic; however, if the expression were not being passed to xs:string but was just being reused to get a string, for example, and in one context ../../foo/bar is an int, the other a string, then we should get diagnostics that indicate the expression isn't correct in one or other other context, and the diagnostic should indicate both the expression location in the schema, AND the location that is problematic context.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: Changed my mind on this. Saving this cutover for next round of incremental changes to the schema compiler. This is too big a change for this change set.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
mbeckerle: This is not true yet. This will be true once we start sharing all the components.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: leaving this in for pending refactoring into factory/non-factory.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
mbeckerle: leaving in pending refactoring into factory/non-factory.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -57,41 +58,54 @@ object ModelGroupFactory {
...
+    Memoize1((child: Node) =>
+      Memoize1((lexicalParent: SchemaComponent) =>
+        Memoize1((position: JInt) =>
+          Memoize1((isHidden: JBoolean) =>
+            (nodesAlreadyTrying: Set[Node]) =>
+              nonMemoizedApply(child, lexicalParent, position, isHidden, nodesAlreadyTrying)))))
+
+  private def nonMemoizedApply(child: Node, lexicalParent: SchemaComponent, position: JInt, isHidden: Boolean,
```
bsloane1650: Can we improve the name of memoizedApply and nonMemoizedApply (and probably add a comment).

I get the sense that what we are "applying" is lexical component to a particular context. I assume the intent here is that it will allow us to still specialize terms when their lexical context requires it, but allows for us to use memoization to avoid re-computing terms that do not require it.

One issue I see with this approach (which may be unavoidable) is that it is highly attractive for developers to add new parameters without giving much thought to exponential blowup. On that note, I see that we are dependent on lexicalParent, which seems like an overly broad dependency .

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
```diff
@@ -145,13 +144,13 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
...
     val namespacesForNamespaceResolution = property.location.namespaces
-    val compileInfoWherePropertyWasLocated = propertyCompileInfo(property)
+    val compileInfoWhereExpressionWasLocated = propertyCompileInfo(property)
     val compiled1 = compileExpression(
       qn,
       staticNodeInfoKind,
       exprOrLiteral,
       namespacesForNamespaceResolution,
-      compileInfo,
+      compileInfoWhereExpressionWasLocated, // Note: Fixed. This was just compileInfo, which isn't right.
```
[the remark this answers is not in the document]
mbeckerle: Nope. I'll remove these.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
mbeckerle: I removed all the memoize stuff, as it wasn't being used, and your observations are very apropos. Memoize approaches work best when the equality comparison work involved is well understood to be either pointer-based equality or operating on simple values or simple tuples of values. 

This was half baked, and not in use as yet, so I removed it for now.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -396,18 +381,18 @@ trait Term
...
+  //    enclosingElement.flatMap { ee =>
```
stevedlawrence: You marked this conversation as resolved but didn't delete the code, keeping intentionally?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -350,14 +321,11 @@ class DPathElementCompileInfo(
...
-  final def enclosingElementPath: Seq[DPathElementCompileInfo] = {
-    enclosingElementCompileInfo match {
-      case None => Seq()
-      case Some(e) => e.enclosingElementPath :+ e
-    }
-  }
+    if (elementCompileInfos.isEmpty) this
+    else if (enclosingElementCompileInfos.isEmpty) this
+    else
+      // else if (elementCompileInfos.head eq this) this
```
stevedlawrence: Remove comment

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
+}
+
+sealed abstract class DownStepExpression(s: String, predArg: Option[PredicateExpression])
+  extends StepExpression(s, predArg) {
+
   override lazy val inherentType: NodeInfo.Kind = {
     if (!isLastStep) NodeInfo.Complex
     else {
-      if (stepElement.optPrimType.isDefined) {
+      if (stepElements.head.optPrimType.isDefined) {
```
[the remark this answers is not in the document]
bsloane1650: What if the elements have a different primType. At the moment, I believe we allow for inherentType to be pretty much anything, and just have the compiler add a cast if needed. Eg. if we have { ../foo + 1 }, where foo is a string type, I believe we would cast ../foo to some numeric type. We could have a scenerio where ../foo is a string in one context, but an int in another context. For a more reasonable example, we might have ../foo have primType decimal in one context and primType double in another (and, perhaps, primType float in yet another).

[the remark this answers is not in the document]
mbeckerle: Right, so this is an issue where the expression such as ../foo is polymorphic.  We either have to support that, or rule it out. The easiest and conservative design choice is to rule it out, so I will add a check that all uses are the same primType.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
         NodeInfo.Complex
       }
     }
   }
-
 }
 
-//TODO: Is ".[i]" ever a valid expression in DFDL?
+// TODO: Is ".[i]" ever a valid expression in DFDL?
+// Yes. For an array element, .[1] is the first occurrence.
```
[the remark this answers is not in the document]
mbeckerle: I will add test to verify if this works or not.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
+      ci
+    }
+  }
+}
 
-  override def text = "."
+sealed abstract class UpStepExpression(s: String, predArg: Option[PredicateExpression])
+  extends StepExpression(s, predArg) {
+
+  override def text = ".." // + "{" + stepElement.path + "}"
```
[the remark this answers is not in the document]
mbeckerle: It was trying to provide the name of the element where ".." lands. But I think it's not important and other diagnostics will display the element names. We don't need it in this text string. So I'll remove it.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
+      if (isFirstStep) {
+        Assert.invariant(!isAbsolutePath)
+        //
+        // This looks like 2 hops up, but it is really 1 hop up.
+        // first position ourselves on nearest enclosing element, or self if
+        // we are an element.
+        // Then take the enclosing elements to get ".." upward move.
+        //
+        val p = compileInfo.elementCompileInfos.flatMap { _.enclosingElementCompileInfos }
+        p.headOption.getOrElse { relPathErr() }
```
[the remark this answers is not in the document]
mbeckerle: I think i abused headOption quite a bit in this code. I'll revisit and simplify.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
 
+  override lazy val inherentType: NodeInfo.Kind = NodeInfo.Complex
 }
 
-case class Up(predArg: Option[PredicateExpression]) extends StepExpression(null, predArg) {
+case class Up(predArg: Option[PredicateExpression]) extends UpStepExpression(null, predArg) {
+
   override lazy val compiledDPath = {
-    if (isLastStep && stepElement.isArray && targetType == NodeInfo.Array) {
+    if (isLastStep && stepElements.forall { _.isArray } && targetType == NodeInfo.Array) {
```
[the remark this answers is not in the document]
mbeckerle: Good catch. Yes they have to be consistently either all array or all not-array, similar to having to be all the same primType which wasn't being checked either. Paths where the last step is an array are only for the argument expression to the fn:count function I believe.  So if fn:count(../../../foo) is an array in some contexts, and non-array in others, that's definitely a bug. 

This check and the one above for consistent primtType also need tests that insure the inconsistencies are caught.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-      //  no prior step, so we're the first step
-      this.compileInfo.elementCompileInfo.getOrElse {
-        relPathErr()
+  override lazy val stepElements = stepElementDefs
+
+  protected def stepElementDefs: Seq[DPathElementCompileInfo] = {
+    if (this.isFirstStep) {
+      val ecs = compileInfo.elementCompileInfos
+      if (ecs.isEmpty) {
+        SDE("There is no enclosing element")
```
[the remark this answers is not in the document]
mbeckerle: Should be assert.invariantFailed I think, because this is a self-step so some sort of ./foo where . turns out to not have any element context? This is not possible I think.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
  */
 case class Up2(s: String, predArg: Option[PredicateExpression])
-  extends StepExpression(s, predArg) {
+  extends UpStepExpression(s, predArg) {
+
   override lazy val compiledDPath = new CompiledDPath(UpMove)
 
   requiredEvaluations(stepQName)
 
   override def text = ".." // + "{" + stepElement.path + "}"
```
mbeckerle: remove commented part.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1103,64 +1103,106 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
...
+        new DownArray(dpathElementCompileInfos.head)
+      }
     } else {
       //
       // Note: DFDL spec allows a[exp] if a is not an array, but it's a processing
       // error if exp doesn't evaluate to 1.
       // TODO: Implement this.
       if (pred.isDefined) subsetError("Indexing is only allowed on arrays. Offending path step: '%s%s'.", step, pred.get.text)
-      new DownElement(dpathElementCompileInfo)
+      new DownElement(dpathElementCompileInfos.head)
```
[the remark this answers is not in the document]
mbeckerle: All the compile infos for a final downward step "foo" in ../../foo must all be to an element named "foo", so since navigation down to a named child element is always the same regardless of its type, any of them will do.

However, I need to verify that this is only taking the element name + namespace information, not using anything else about the element. Perhaps these should be changed to take only that information, and not the whole dpathElementCompileInfo, since this downward step is actually representing a polymorphic downward step to any of the possible named children (having the same name and namespace) that this path could be stepping downward into.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -200,7 +238,7 @@ trait SchemaComponent
         case sgd: GlobalSequenceGroupDef => "sgd=" + sgd.namedQName
         case cg: Choice => "c" + (if (cg.position > 1) cg.position else "")
         case sg: Sequence => "s" + (if (sg.isHidden) "h" else "") + (if (sg.position > 1) sg.position else "")
-        case sc => Assert.invariantFailed("Unexpected component type: " + sc)
+        case unknown => "unk=" + unknown.toString()
```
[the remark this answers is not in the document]
mbeckerle: SSCD strings are used in logging and printf debug-by-print statements, and if they can fail you get stack overflows, etc. I was running into those so I rigged this to never fail by putting this unknown case in as the stop-gap after I found I had missed a case. (which I also added). 

I will switch this back to Assert.invariantFailed now that this is debugged, and put a more informative message in it indicating that it failed trying to construct an SSCD for a particular object class. That will make it quick to fix if we in fact run into it again.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
mbeckerle: Odd. I will make sure this code is deleted. .

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Ah. There are lots of polymorphic possibilities. 

fn:count(pathToArray) cares only that the path is to an array, but does not care if in some cases the path is an array of simple type, another case path is an array of complex type. 

There are functions where only the node matters, not its type. E.g., fn:local-name(path) or fn:namespace-uri(path) which work for any node at all. 

Or fn:nilled(path) which works for either simple or complex type elements since both can be nillable.

Also functions dfdl:contentLength(path, units) and dfdl:valueLength(path, units) are both potentially polymorphic in that they measure size, but don't care what the thing actually is. 

dfdl:checkConstraints(path) works for any simple type element. 

fn:exists(node) only cares about the node existing, doesn't care about the type.

So really to check consistency we have to do this in a function-specific manner, or we can insist on uniformity even when the functions being used don't strictly speaking require it.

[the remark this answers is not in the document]
bsloane1650: Are we still able to make the conservative argument? Daffodil already has an established behavior, and that behavior is required by the spec (at least for implementations that choice to implement DPath). This sounds to me like we will be deliberately introducing a regression, and it is not clear to me we have a compelling reason to do so. Further, the broken behavior is just weird from a user standpoint. You can define a valid type T, and use it at location A xor location B. That type of restriction does not strike me as reasonable.

I see a few ways around this problem without introducing a regression:

A) Detect when such a situation would arise and specialize the expression/term/type/etc as many times as needed. Eg. If we have an element <bar> that references { ../foo } at 3 locations where foo is a string, and 5 where for is an int, then we compile 2 instances of <bar>. I thought this was the idea behind the memoizedApply function I saw earlier. Further, I suspect we will find this mechanism would be useful as a way of avoiding other tricky cases arising from de-duplication. Having said that, as I noted in my comment on memoizedApply, we would need to be careful to avoid overusing this mechanism.

B) Detect when such a situation would occur and add a runtime cast that can dispatch on the type (or on a flag set in an ERD object). This wouldn't add any overhead in the common case since we only need to include such a node when the types do not match.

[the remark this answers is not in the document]
mbeckerle: Good points. 

I am not sure we're really looking at a regression here. I am not sure any expressions in any DFDL schemas require any polymorphism even though they theoretically could. 

That said, I suppose the right algorithm is just to generalize the types. So if ../foo is int in some places and float in others and string in others, then the least upper bound is AnySimpleType for those. Then the DPath compiler would error if that expression appeared in a context where AnySimpleType isn't sufficient. So fn:localName(../foo) would work, but ../foo + 7 would fail to compile.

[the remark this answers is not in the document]
bsloane1650: It is still a regression in that we would be introducing a bug where one did not previously exist. The only question is if anyone will notice. You are probably correct that no one would notice; but if someone did notice, it would clearly be a bug in Daffodil (or, at least non standards compliant behavior). As Daffodil continues to grow in popularity and capabilities, it seems to me to be only a matter of time until someone tries to do this for some reason. 

At a minimum, can we make this a Not Yet Implemented SDE? That seems less confusing, both to a hypothetical user that stumbles onto this, and a developer who notices that we are emitting an SDE for legal behaviour.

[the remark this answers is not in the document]
mbeckerle: I will look at the type-generalization approach first, and go to a NYI SDE if that gets too complex for this change set. 

I actually think the type generalization won't be very complex. everything needed to do a type lower bound is already there in the nodeinfo stuff, and issuing the diagnostics is no more complex than issuing the diagnostics about type incompatibility that I was initially considering.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -131,16 +131,18 @@ trait Term
...
 
-    localProps.foreach { case (prop, (value, _)) =>
-      if (!usedProperties.contains(prop)) {
-        SDW(WarnID.IgnoreDFDLProperty, "DFDL property was ignored: %s=\"%s\"", prop, value)
-      }
+    localProps.foreach {
+      case (prop, (value, _)) =>
+        if (!usedProperties.contains(prop)) {
+          SDW(WarnID.IgnoreDFDLProperty, "DFDL property was ignored: %s=\"%s\"", prop, value)
+        }
```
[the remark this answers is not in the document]
mbeckerle: Yes, auto format. Cannot find a way to adjust it not to do this. Basically case keyword starts on a new line in all cases.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: This turns out to be a tricky issue. Since expressions are polymorphic, it's unclear exactly which downward step is failing here if this fails at runtime. When we were copying the expressions, the diagnostic info (schema file location) for a corresponding downward step was easily identifiable. I will have to investigate.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Ticket is DAFFODIL-2182. Test (which fails) added.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Type generalization implemented. Please see the new tests which actually do things that are polymorphic of int/float and int/Complex type.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Conversions.scala
```diff
@@ -18,6 +18,7 @@
 package org.apache.daffodil.dpath
```
mbeckerle: This file moved to daffodil-core from daffodil-runtime1. Not needed in the runtime.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Conversions.scala
```diff
@@ -43,22 +44,23 @@ object Conversion {
...
     import NodeInfo._
     val ops: List[RecipeOp] = (st, tt) match {
       case (_, Array) => Nil
       case (x, y) if (x == y) => Nil
-      case (x, Nillable) => {
-        // we don't have enough information here to check whether the
-        // item under scrutiny is nillable or not.
-        Nil
-      }
+      // case (x, Nillable) => {
```
mbeckerle: remove commented code throughout file.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +964,192 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-        val pt = stepElement.optPrimType.get
-        pt
-      } else {
-        NodeInfo.Complex
-      }
+    stepElements.map { _.typeNode }.reduce {
+      (left, right) =>
+        NodeInfoUtils.typeLeastUpperBound(left, right)
     }
+    //
```
mbeckerle: remove commented code - this diagnostic behavior is elsewhere now.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Still open issue.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -943,155 +950,148 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-      ps5
+  override protected def stepElementsDef: Seq[DPathElementCompileInfo] = {
+    val p = super.stepElementsDef
+    val res = p.filter { ci => ci.namedQName.matches(stepQName) }
+    if (res.isEmpty) {
+      p.headOption.foreach { _.noMatchError(stepQName) } // error about first one.
+    }
+    if (res.length > 1) {
+      // There are multiple upward paths to elements of the same name.
+      // So long as they all ultimately type-check, this is ok.
```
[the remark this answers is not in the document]
mbeckerle: Refactored.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/NodeInfoUtils.scala
```diff
@@ -100,57 +99,82 @@ object NodeInfoUtils {
...
+      // For each abstract type that could be the least upper bound of the two
+      // arg types, we must pick a concrete type to convert everything into.
+      val lubImplementationType = lub match {
+        case SignedNumeric => NodeInfo.Double
+        case _ => lub
+      }
+      (lubImplementationType, lubImplementationType)
     }
+
+    //      (leftArgType, rightArgType) match {
```
mbeckerle: Remove commented code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -253,16 +258,16 @@ object NodeInfo extends Enum {
    * hierarchy. We have it as a parent of both complex type and simple type since
    * both of them can be nillable.
    */
-  protected sealed trait NillableKind extends AnyType.Kind
-  case object Nillable extends TypeNode(AnyType, List(Complex, AnySimpleType)) with NillableKind {
-    type Kind = NillableKind
-  }
+  //  protected sealed trait NillableKind extends AnyType.Kind
```
mbeckerle: remove commented code - several of these type taxonomy nodes have been removed (2 or 3)

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
```diff
@@ -0,0 +1,400 @@
...
+          </e2>
+        </ex:r>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:warnings>
+      <tdml:warning>should be manually cast</tdml:warning>  
+    </tdml:warnings>
+  </tdml:parserTestCase>
+
+  <tdml:defineSchema name="s2a" elementFormDefault="unqualified">
```
mbeckerle: Please review these test cases carefully.

In this first one, you see that to get the choiceDispatchKey expression to work, you have to do lots of manual conversions. That's because ../foo/bar path expression has different types for the two different reuses of this group.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
```diff
@@ -0,0 +1,400 @@
...
+
+  <tdml:parserTestCase name="test_polymorphic_expr_2b" model="s2b" root="r"
+    description="Polymorphic path expression">
+    <tdml:document>1|1|a|2|1|b</tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>expression ../foo/bar</tdml:error>
+      <tdml:error>SignedNumeric</tdml:error>
+      <tdml:error>cannot be converted to xs:int</tdml:error>
+      <tdml:error>element bar with xs:int type at Location</tdml:error>
```
mbeckerle: There are actually two diagnostics about different element bar. They have different locations in the TDML schema. The TDML runner can't specify that there are two distinct matches for the string.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
```diff
@@ -0,0 +1,400 @@
...
+    <xs:group name="g">
+      <!-- 
+        The path expression below is going to have different type
+        elements for the different instances of this group in the 
+        final schema. 
+        
+        Ultimately they all are converted to string.
+       -->
+      <xs:sequence>
+      <xs:choice dfdl:choiceDispatchKey="{ ../foo/bar }">
```
bsloane1650: Should this have an explicit cast to xs:string ?
I believe the existing behavior is that an automatic cast at the top level is generated, but will emit a depreciation warning. Unless we are explicitly checking that the depreciation warning is being generated, we should probably future-proof this and insert the cast.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
```diff
@@ -0,0 +1,400 @@
...
+    </xs:element>
+ </tdml:defineSchema>
+
+  <tdml:parserTestCase name="test_polymorphic_expr_3" model="s3" root="r"
+    description="Polymorphic path expression">
+    <tdml:document>1|1|a|2|1|b</tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>cannot be converted to xs:int</tdml:error>
+      <tdml:error>Left operand for operator '+' must have numeric type</tdml:error>
```
bsloane1650: What does the context look like for this error? Ideally, we would have both: the location where the expression is defined (eg. the lexical context of g), and the location where its usage causes a problem (eg. e1/foo)

bsloane1650: We should add this to the test; as this seems like a place where it would be easy to accidentally break our diagnostics without noticing.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
```diff
@@ -0,0 +1,400 @@
...
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+
+    <xs:element name="r">
+      <xs:complexType>
+        <xs:sequence dfdl:separator="|">
+          <xs:element name="e1" type="xs:string" minOccurs="3" maxOccurs="3">
+            <xs:annotation>
+              <xs:appinfo source="http://www.ogf.org/dfdl/">
+                <dfdl:assert>{ if (dfdl:occursIndex() eq 1) then fn:true() else ( .[1] eq '1' ) }</dfdl:assert>
```
bsloane1650: { . } and { .[1] } are the same thing when the current node is an element, aren't they? I don't think "." is supposed to elevate you to the array containing the current element. I tested this here: https://www.freeformatter.com/xpath-tester.html with the xml:

```
<foo>
  <a>1</a>
  <a>2</a>
  <a>3</a>
</foo>
```

where the expression `foo/a[2]/.[1]` resolved to <a>2</a>

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
bsloane1650: Nevermind, I see we are checking for such a warning.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
bsloane1650: Can you add a test case that combines xs:float, xs:decimal, xs:double, as well as one that combines xs:anyURI and xs:string (if we even support xs:anyURI, which I am not sure we do).

According to XPath 2.0, float and decimal are both autocast to double, and anyURI can be autocast to string.

I am not clear on what the implications of this for us are. In the expression "{ ../foo/bar + 1 }", it is not clear to me what type the addition should take place in. I believe our current behaviour is that the literal "1" is cast to whichever type bar has in the given context, which is then the resulting type of the entire expression (although I have been wrong about our coercien rules before). To make this work polymorphically, we would need to decide what type we want ../foo/bar to be, and insert the auto-casting, which would slightly change the sementics at the call-sights which end up relying on the cast. In fact, we might have a simmiliar problem when ../foo/bar is a combination of different decimal types: the addition of a new context in which the expression is used may change the precision with which it is computed at all contexts.

mbeckerle: DFDL does not have anyURI type.

I do have example of a path that is shared and has type int/float, and float/complex-type. 

This is a point where the DFDL spec is unclear. It takes no position on whether expressions are shared or copied/inlined per context. What I've done is make it so that if properties are expressed on a reference term (element ref, sequence ref, or choice ref) then the properties are copied. This is a natural consequence of the fact that the properties might appear on the ref, not on the shared object being referenced, so they may in fact be distinct property values despite referencing a shared definition.

However, if an expression appears on a local element, sequence or choice, then it is inside an element decl, or group definition, so then it is not being shared, and so no copying is being done.  BUT, a relative path within it it that is upward-and-outward (e.g., ../../foo/bar) may reference outside the global decl/definition. This is the case in my test examples. In this case the path is (now, with this change set) being interpreted as a polymorphic expression ... one that must make sense in all contexts, and must have a common definition shared definition in all contexts.

So if we have int and float interpretations we will get double behavior since the LUB of int and float is SignedNumeric, and the most general representation of that (the type all such can be converted into) is double. Same with float and decimal.  But if it is int and decimal, you'll get decimal behavior since decimal is a supertype of int. 

I can add examples that show the int/decimal case, so as to show you don't always end up with double. The only way to test this in TDML is to have a test that requires an auto-coercion be added, and then detect the warning for this, or possibly set a flag that makes the lack of this auto-coercion to cause an error. The error can be scrutinized for "decimal". I will construct one of these examples.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
mbeckerle: Hmmm. There was DFDL workgroup discussion about this a long time back. I will dig in those archives. However, as this is now isolated in a bug ticket about exactly this issue, I will add your comments to that ticket, and not worry about this for this change set.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
mbeckerle: Tests added

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
mbeckerle: Added checks that two different contexts for the different types, are exhibited.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Resolved. There is a check elsewhere. This code just asserts now.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions3.tdml
bsloane1650: Thinking about this some more, it seems like we want to behave as-if we are inlining everything. The alternative is to have spooky action at a distance.

If we write schema with ```<xs:group ref="foo"/>``` in our schema today, then 5 years from now, extend our schema with a choice to support a new version of the format in question and re-use the group foo in a different context, we do not want that to change the behavior on any of the pre-existing messages. Similarly, if we see 2 identical group definitions in the same file, it should not change the behavior of the schema to combine them into a single group definition.

mbeckerle: Yes. I guess I agree. Full referential transparency should be the goal.
Whether you copy the reusable thing inline or keep it shared should not change the behavior. 

There is another alternative - parametric - require people to bind variables rather than allow these upward-and-outward paths. The parameters have types. Another equivalent possibility: require encapsulating any upward-and-outward path expression in a constructor call.. So instead of ../../foo/bar you must write xs:double( ../../foo/bar ) (or other constructor) which is another way of forcing a type-declaration - in this case the type of foo/bar must be suitable to be an arg to the double constructor, which means I believe that it just has to have a value, since those constructors even take strings. This is, I guess, just weakening the type checking substantially. 

Ultimately, keeping track of every point of usage, the types at those points of use, and generating N specializations for the N actual situations that arise, is not impossible. It's just harder. 

E.g., consider an expression with multiple paths in it, each of which is upward-and-out to a variety of different types. The number of instance you need is potentially the product of those.... 

I would probably approximate this as follows: If there is exactly one type for all usages share. Otherwise copy.  I would certainly hope the vast bulk of expressions are used monomorphically. 

Are there any other ideas for how to achieve full referential transparency here?

bsloane1650: Full inlining is probably sufficient. As you said, there is not much use for polymorphic expressions. The time I could see this arising would be with massive super schemas that try to parse everything (eg. a dispatch on a files magic number) and developed a modular library of components. Even in such a schema it seems unlikely, but we have no experience with such schema yet. The issue here is that a small change could cause a significant performance degradation for hard to debug reasons.

What we might want to do is emit a warning when this situation arises. If this turns out to be something people actually want to do we can deal with it then. Otherwise, these polymorphic expressions are probably mistakes.

mbeckerle: Ah. Good point. We can just issue a diagnostic if there is polymorphism required, and see what happens. It is certainly true that the tests I have for polymorphic expressions are 100% contrived situations. I tend to agree that allowing polymorphism is more likely to just be allowing a mistake than a true intended usage.  

This is far easier to implement so I will go with this for now.

mbeckerle: I have implemented Not Yet Implemented SDE for polymorphism. 

The diagnostics show the examples where the polymorphism is happening, and the line numbers and types involved.

PR #260, opened 2019-07-09
Adds support for non-spacing and combining chars

olabusayoT:
Pads non-spacing/combining chars with spaces so they show up properly
aligned in the debugger without combining with our ~ delimiters or
whitespace. Since they sometimes combine with preceding and following
characters (e.g \u093f or \u064d), so we pad them on either side.
This has the effect of displaying the char with an additional space but it is aligned 
without the risk of it consuming our tilde.

DAFFODIL-2168

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
@@ -486,6 +486,11 @@ class DataDumper {
       case Some(dec) => {
         val bb = ByteBuffer.allocate(6)
         var cb = CharBuffer.allocate(1)
+        var cr = CoderResult.OVERFLOW
+        var nConsumedBytes = 0
+        var remapped = ""
+        var nCols = 0
+        val INVALID_CODEPOINT = -1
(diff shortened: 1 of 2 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/io/Dump.scala
```diff
@@ -516,24 +544,41 @@ class DataDumper {
...
+                case e: IllegalArgumentException => {
+                  -1
+                }
+              }
+            } else allChars(0)
+
+          val (r: String, n: Int) =
+            if (allChars.length > 1) {
+              if (uCodePoint == -1) {
+                allChars.map(c => homogenizeChars(c)).foldLeft(("", 0)) {
```
mbeckerle: I don't understand this piece of code. Please add comment as to what this case is doing.

olabusayoT: I updated the code for clarity, adding comments and doing some additional checks so we don't need to rely on exception handling for illegal surrogate pairs!

PR #263, opened 2019-07-22
Add more obscure charsets needed for some mil formats.

mbeckerle:
Also minor changes to support debug.

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
@@ -22,10 +22,17 @@ import org.apache.daffodil.exceptions.Assert
 object DaffodilCharsetProvider {
 
   lazy val charsets = Seq(
+    BitsCharset3BitDFI336DUI001,
+    BitsCharset3BitDFI746DUI002,
+    BitsCharset3BitDFI747DUI001,
+    BitsCharset4BitDFI746DUI002,
+    BitsCharset5BitDFI769DUI002,
+    BitsCharset5BitDFI1661DUI001,
     BitsCharset5BitPackedLSBF,
     BitsCharset6BitDFI264DUI001,
     BitsCharset6BitDFI311DUI002,
-    BitsCharset5BitDFI1661DUI001,
+    BitsCharsetBase4LSBF,
+    BitsCharsetBase4MSBF,
     BitsCharsetBinaryLSBF,
     BitsCharsetBinaryMSBF,
     BitsCharsetHexLSBF,
@@ -42,8 +49,7 @@ object DaffodilCharsetProvider {
     BitsCharsetUTF16LE,
     BitsCharsetUTF32BE,
     BitsCharsetUTF32LE,
-    BitsCharsetUTF8
-  )
+    BitsCharsetUTF8)
 
   private lazy val charsetMap = {
     val nameCharsetPairs = charsets.flatMap { cs =>
--- a/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
@@ -20,7 +20,7 @@
  * The DFI/DUI numbers may be used to lookup the definition of the encoding by those with access
  * to the relevent specification. In cases where multiple DFI/DUI pairs make use of the same encoding,
  * the decision of which pair to name the encoding after is arbitrary.
- * 
+ *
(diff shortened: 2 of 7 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
```diff
@@ -48,6 +48,46 @@ object BitsCharset6BitDFI311DUI002 extends {
...
+
+object BitsCharset3BitDFI747DUI001 extends {
+  override val name = "X-DFDL-3-BIT-DFI-747-DUI-001"
+  override val bitWidthOfACodeUnit = 3
+  override val decodeString = """AEGHJKLM"""
+  override val replacementCharCode = 0x0
+  override val requiredBitOrder = BitOrder.LeastSignificantBitFirst
+} with BitsCharsetNonByteSize
+
+object BitsCharset5BitDFI769DUI002 extends {
```
mbeckerle: Add TDML tests that exercise these charset encodings in ways that prove they will work properly for the real use cases.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
```diff
@@ -25,7 +25,14 @@ object DaffodilCharsetProvider {
     BitsCharset5BitPackedLSBF,
     BitsCharset6BitDFI264DUI001,
     BitsCharset6BitDFI311DUI002,
+    BitsCharset3BitDFI336DUI001,
+    BitsCharset4BitDFI746DUI002,
+    BitsCharset3BitDFI746DUI002,
+    BitsCharset3BitDFI747DUI001,
+    BitsCharset5BitDFI769DUI002,
```
stevedlawrence: Might be worth sorting these just for extra clarity.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/X_DFDL_MIL_STD.scala
```diff
@@ -48,6 +48,46 @@ object BitsCharset6BitDFI311DUI002 extends {
...
+object BitsCharset3BitDFI746DUI002 extends {
+  override val name = "X-DFDL-3-BIT-DFI-746-DUI-002"
+  override val bitWidthOfACodeUnit = 3
+  override val decodeString = """ABCDEFGH"""
+  override val replacementCharCode = 0x0
+  override val requiredBitOrder = BitOrder.LeastSignificantBitFirst
+} with BitsCharsetNonByteSize
+
+object BitsCharset3BitDFI747DUI001 extends {
+  override val name = "X-DFDL-3-BIT-DFI-747-DUI-001"
```
stevedlawrence: Minor and probably not worth changing, but I've never been a fan of charset names that contain DFI/DUI's. It seems so specific to a single use case of a charset. Granted most of them probably are the only cases, but still seems odd. Especially since encodings *could* be shared by by different DFI/DUIs. For example, maybe DFI 123 DUI 456 has the same encoding. Using this encoding for that DFI/DUI now will look like a bug in the schema, even though it's would be correct. But maybe the solution there is to just add aliases?

I'd maybe rather see something like X-DFDL-TDL-NNNN-M, where NNNN a zero-padded number that we increment for every new TDL encoding we come across, and M is the bit length. So this might be X-DFDL-TDL-0010-3.

mbeckerle: The notion here is that we pick one of the dfi/dui as the "canonical" usage, and any other dfi/dui that share the encoding would use that same encoding, and the fact that it is shared would not look like a mistake, as much as intention to share the "canonical" one. Lots of things are shared across dfi/duis so this seems an ok course.  Or we could do aliases as you suggest. There are numerous specs that could be re-using these encodings, but all of them are these DFI/DUI-style specs, so I think there is little gained by avoiding the DFI/DUI identification. 

There is an argument to get the dfi/dui notation entirely out of Daffodil, but in doing so we just make it hard for anyone using Daffodil to know if the charset they need is in fact part of daffodil, so we'd end up documenting them as X-DFDL-TDL-NNNN-M is the charset needed for DFI X DUI Y anyway.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
mbeckerle: They are sorted by the DFI number. Not the length. But given the mixture of different encodings here, some of which are these obscure DFI/DUI things, some of which are more understandable things, I take your point. i'll sort these alphanumerically.

PR #264, opened 2019-07-24
WIP: DAFFODIL-2169

bsloane1650:
The groundwork for adding type-safty to our DPath and infoset values.

This doesn't actually effect any of the existing code yet, but should give any idea on what I am planning. Given the volume of effort needed to actually incorporate this into the codebase, I wanted to get feedback before I attempted to do so.

When I do thread it through, I suspect I will run across other issues. I'll try to keep the "meaningful" changes separate from the boilerplate updates to type signatures.

The bad:
1) We need to support having DINodes, but also need the DataValue type to be visible to Daffodil-lib (because we will want to use it for the number-conversion functions). I am accomplishing this by defining a trait it DataValue.scala that is instantiated only by DINode, along with an implicit to convert back to a DINode. I don't think this will cause additional memory allocations, but it might invoke an additional typecheck when we try to get extract a DINode.
2) We still need to thread this through all of Daffodil. 

The good:
1) DataValue is not AnyRef
2) By using phantom types, we can impose an arbitrary type hierarchy on our DataValue.I suspect the only useful useful distinction here is going to be Nullable vs NonNullable.
We might also be able to take advantage of the seperate DataValue types for each supported underlying type. In particular, we might use them to tighten the return type when subclassing. In theory, we could have done that when using AnyRef, but not that the types are still marked as DataValue it may be clearer what is going on.
I also included DataValuePrimitive to indicate types which are definined in Daffodil-lib. This is mostly just a demo that this is possible to define more complicated type hierarchies. I don't actually see a way to take advantage of it.


daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
```diff
@@ -0,0 +1,91 @@
...
+  implicit def toDataValue(v: Float) : DataValueFloat = new DataValue(v:JFloat)
+  implicit def toDataValue(v: Byte) : DataValueByte = new DataValue(v:JByte)
+  implicit def toDataValue(v: Int) : DataValueInt = new DataValue(v:JInt)
+  implicit def toDataValue(v: Short) : DataValueShort = new DataValue(v:JShort)
+
+  val NoValue = new DataValue(new NoValueObj)
+
+}
+
+protected class NoValueObj {
```
mbeckerle: If this is just a singleton, why not an object. Any why not make it  a private member of the DataValue object?

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
```diff
@@ -0,0 +1,91 @@
...
+trait Nullable extends NonNullable;
+sealed trait DataValuePrimitiveType
+
+final class DataValue[+T <: AnyRef, +X<:AnyRef] private (val v: T) extends AnyVal {
+  def isEmpty = DataValue.NoValue.v eq v
+  def isDefined = !isEmpty
+  def value = v
+  override def toString = if (isEmpty) "NoValue" else "DataValue("+ v.toString +")"
+  
+  def getAnyRef = v.asInstanceOf[AnyRef]
```
mbeckerle: No cast needed above. v is already AnyRef. I would hope the scala compiler knows that, but ...?

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
```diff
@@ -0,0 +1,91 @@
...
+ * class.
+ * The DINode class itself defines an implicit DataValueDINode -> DINode conversion
+ */
+trait DataValueDINodeTrait extends NonNullable;
+
+trait NonNullable;
+trait Nullable extends NonNullable;
+sealed trait DataValuePrimitiveType
+
+final class DataValue[+T <: AnyRef, +X<:AnyRef] private (val v: T) extends AnyVal {
```
mbeckerle: You don't use type X anywhere, though you provide it in type parameters below. 
First, the name X should be something like NullableT or something explanatory.  But I'm not clear on how you intend to use this.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import java.lang.{ Boolean => JBoolean, Number => JNumber, Long => JLong, Double => JDouble, String => JString, Float => JFloat, Byte => JByte, Integer => JInt, Short => JShort }
+import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
+import org.apache.daffodil.calendar.DFDLCalendar
+
+/*
+ * DINode is not visible to Daffodil-lib, so we cannot actually reference this here.
+ * We are trusting the rest of the codebase to only apply DataValueDINode to the DINode
+ * class.
+ * The DINode class itself defines an implicit DataValueDINode -> DINode conversion
```
mbeckerle: Why are you bothering to put this in daffodil-lib? It feels like part of the runtime to me, and this should just be part of InfosetImpl.scala or related file.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -95,6 +96,10 @@ sealed trait DINode {
 
 }
 
+object DINode{
+  implicit def toDINode(v: DataValueDINodeTrait):DINode = v.asInstanceOf[DINode]
```
mbeckerle: This cast is only needed because of the daffodil-lib location of the DataValue class. This whole thing can be mixins to DINode and then there's no casting I think.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
bsloane1650: Turned into an object. I cannot make it private, because "val NoValue = new DataValue(NoValueObj)" causes it to escape its defining scope.

bsloane1650: Disregard above. Making it an object prevents using it as a type, which means I cannot define the DataValueEmpty type.

I'm not convinced that we will ever actually use said type, but I don't see the harm of declaring it as a class that gets initialized to a singleton.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
bsloane1650: Technically, v is of type T, so I would have to explicitly add the type annotation of getAnyRef, otherwise getAnyRef would return type T. That would probably be fine, since Scala can upcast at the callpoint just as easily. This asInstanceOf is mostly just to be consistent with the rest. The only downside I can see is that it might invoke an extra type-check at runtime, but I would hope Scala knows not to bother actually checking that an object is an AnyRef (and even if it wasn't, as you point out, the type constraint explicitly guarantees that this is just an upcast, so Scala has a second way of knowing to omit it at runtime).

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
bsloane1650: X is a phantom type. It is even less meaningful that the type variable T (which could be called, say, UnderlyingType).
The purpose of X is to create a type hierarchy of the various type aliases I define below.
In particular, it will allow us to define a type that can be any subset of the support types.
For instance, I am currently using it to define:
* DataValueNonEmpty - which excluded DataValueEmpty
* DataValuePrimitive - which excludes DataValueDINode

We could also, define DataValueInteger, DataValueNumeric, etc.

This will allow us to add types in a limited context without while still leveraging the type system to insure that they do not occur everywhere DataValue is used.
For instance, we currently have that the "defaultVal" could be set to the singleton object "UseNilForDefault".

When I get around to actually supporting that, we can define another type that does not include UseNilForDefault as part of what it would allow.
We might find we can make useful distinctions between subsets of the types I already have. I am hopeful I will at least be able to make the empty/non-empty distinction useful. Not so confident that any other distinction will work out.

This whole sub-exercise might also turn into me wrestling with the Scala/Java type system, in which case I will likely drop it.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
bsloane1650: I believe I was looking at our Numbers.scala file, which currently is defined in terms of AnyRef. I believe our only use of it is for objects that are DataValue; so I wanted to try include that in the code that is DataValue aware. Particuarly the as_ methods.

I suppose we could move these methods to DataValue itself. The issue there is that we need to distinguish between the accesors that just do an .asInstanceOf, and those that dispatch on the type and potentially do a runtime conversion.

I can try doing that, then hopefully remove most of what is in Numbers.scala.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
bsloane1650: To need for a mixin. This whole exercise is just because DataValue doesn't have access to the DINode type.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
mbeckerle: No big deal. Just thought we could hide it better, but if it's problematic, don't bother.

daffodil-lib/src/main/scala/org/apache/daffodil/util/DataValue.scala
mbeckerle: Ah that's what is going on here. Well you should explain that in the code in comments.

Good luck. I agree that if this just turns into type-system craziness then it is not worth it. However, I still think having this AnyVal class helps.

bsloane1650:
Closing to submit updated PR

PR #269, opened 2019-09-23
DAFFODIL-2200. xs:import problems with xsd files provided in daffodil-lib

dineshchitlangia:
Jira had reported 2 issues:
1. schemaLocation attributes should just be the filenames
2. missing schemaLocation in XMLSchema_for_DFDL.xsd when importing dfdl namespace.

Issue 1 was not seen and probably fixed by another jira in the past.

This PR addresses issue 2.

### DIFF
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
@@ -39,11 +39,12 @@ this to validate when we load a DFDL Schema.
    "http://www.w3.org/2001/XMLSchema"
   -->
 
-  <xsd:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/" />
+  <xsd:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/"
+    schemaLocation="DFDL_part3_model.xsd" />
 
   <xsd:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
     schemaLocation="dfdlx.xsd" />
-    
+
   <xsd:import namespace="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int"
     schemaLocation="dafint.xsd"/>
 
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
@@ -27,8 +27,8 @@
   attributeFormDefault="unqualified"
   elementFormDefault="qualified">
 
-  <xs:import namespace="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int" schemaLocation="xsd/dafint.xsd" />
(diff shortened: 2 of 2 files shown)


daffodil-lib/src/main/resources/org/apache/daffodil/xsd/XMLSchema_for_DFDL.xsd
```diff
@@ -40,10 +40,11 @@ this to validate when we load a DFDL Schema.
   -->
 
   <xsd:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/" />
+    schemaLocation="DFDL_part3_model.xsd" />
```
mbeckerle: The line above has "/>" on it, so this line is broken/not well formed. Tests are not passing as a result.

dineshchitlangia:
> See other comment to fix.
> 
> You should squash what will be 3 commits together and then force-push your changes again. That way it will be one single commit when it hits the master.
> 
> Thanks

@mbeckerle Updated the PR as advised. I will recheck after CI run.

dineshchitlangia:
@mbeckerle it looks like for some reason one of the CI phases timed out at Integration test.
Could you help to re-trigger it pls?

mbeckerle:
Will try again tomorrow. My current attempts to re-try all tests then fail with a git checkout error that makes no sense:
   git checkout --progress --force 77f6143198b29368072669fae0d6fbebb14b6807
   ##[error]fatal: reference is not a tree: 77f6143198b29368072669fae0d6fbebb14b6807

stevedlawrence:
It looks like the GitHub actions CI is failing due to https://github.com/actions/checkout/issues/23. I think we'll just have to rely on TravisCI until this issue is resolved.

Though, it looks like the first part of the bug is still an issue. The ``dfdlx.xsd`` file has the following lines:
```xml
  <xs:import namespace="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int" schemaLocation="xsd/dafint.xsd" />
  <xs:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/" schemaLocation="xsd/DFDL_part3_model.xsd" />
```
The bug report says that the schemaLocation attributes should be this:
```xml
  <xs:import namespace="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int" schemaLocation="dafint.xsd" />
  <xs:import namespace="http://www.ogf.org/dfdl/dfdl-1.0/" schemaLocation="DFDL_part3_model.xsd" />
```
I.e., the "xsd" directory is removed from the schemaLocation. I'd doesn't look like that was fixed as another PR. I *think* the bug is right and the xsd should be removed.

Also, I recommend changing the commit message to remove the squashed commit messge "fixed broken tag" since the squashed commit doesn't actually fix any broken tags. Also, in the Daffodil project, the standard is to list the bug number at the end of the commit message, so your commit message should look something like this:

> xs:import problems with xsd files provided in daffodil-lib
>
> Removed extra whitespace
>
> DAFFODIL-2200

dineshchitlangia:
@stevedlawrence thank you for checking this PR. I have addressed review comments in latest commit.

dineshchitlangia:
Thank you all for review and guidance on my first contribution to this project!

PR #271, opened 2019-10-02
Fixes and tests twoPass Mac LE bug

olabusayoT:
Fixes bug where data containing CRs aren't replaced with LFs during
twoPass testing.

DAFFODIL-2209


stevedlawrence:
+1

PR #274, opened 2019-10-08
WIP: Daffodil 2202 initial codegen

mbeckerle:
Reorganized.

First commit is just adding the daffodil-codegen module with the codegen library in it.

Second commit is new code-generation capability in the daffodil-core module, which is the schema compiler. This doesn't work yet. This WIP pr is to exhibit the issue. 

When viewing this PR, view only the commits other than the first one so you can see the actual changes and not just the addition of the library. 

DAFFODIL-2202


daffodil-core/src/main/scala/org/apache/daffodil/runtime2/generators/ElementParserGenerator.scala
```diff
@@ -0,0 +1,176 @@
...
+  override def generateCode(cgState: CodeGeneratorState): CodeGeneratorState = {
+
+    val childDecls =
+      context.elementChildren.map { child =>
+        if (child.optPrimType.isDefined)
+          new FieldDeclaration(toPrimitive(child.optPrimType.get), child.name)
+        else
+          new FieldDeclaration(toComplexType(child), child.name)
+      }
+
```
mbeckerle, line 57: Below is where we get in trouble.
Line 61 specifically has no "target" because the target is 'this' class object's e1 field. 
But later when generating code we get a null pointer exception when it tries to access the target of this FieldReference object.

daffodil-core/src/test/scala/org/apache/daffodil/runtime2/TestRuntime2.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Test
+import java.io.ByteArrayInputStream
+
+import org.apache.daffodil.util.{SchemaUtils, Misc}
+import org.apache.daffodil.runtime2.generators.CodeGeneratorState
+
+class TestRuntime2 {
+
+  @Test def test1(): Unit = {
+
```
mbeckerle, line 30: This is our super-basic example. Just one root r, with one e1 element inside of type int.

daffodil-runtime2/src/main/java/org/apache/daffodil/codegen/ast/AssignementExpression.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private final Expression target;
+    private final Node value;
+
+    AssignementExpression(Expression target, Node value) {
+        super(target.getType());
+        this.target = target;
+        this.value = value;
+    }
+
```
mbeckerle, line 31: This getTarget() is going to return null for the assignment to e1.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
```diff
@@ -18,17 +18,13 @@
 package org.apache.daffodil.grammar
 
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.processors.parsers.Parser
-import org.apache.daffodil.processors.unparsers.Unparser
-import org.apache.daffodil.dsom.SchemaComponent
 import org.apache.daffodil.oolag.OOLAG.OOLAGHostImpl
-import org.apache.daffodil.compiler.ParserOrUnparser
-import org.apache.daffodil.compiler.BothParserAndUnparser
+import org.apache.daffodil.compiler.{ParserOrUnparser, BothParserAndUnparser}
```
mbeckerle, line 22: revert these import changes. Intellij IDEA is doing this automatically due to incorrect settings.

daffodil-core/src/main/scala/org/apache/daffodil/runtime2/generators/ElementParserGenerator.scala
mbeckerle, line 57: Below at line 61 is where the issue is.

mbeckerle: We get a NPE later when the code generation is trying to access the target of this field reference and gets null.

tuxji:
Let's close this pull request since we have opened a new pull request to replace it: <https://github.com/apache/incubator-daffodil/pull/422>.

PR #275, opened 2019-10-10
Added smaller test to reproduce vCard failure.

mbeckerle:
Fails currently so is commented out in the scala driver.

DAFFODIL-1034


stevedlawrence:
+1

PR #276, opened 2019-10-11
Updated Unordered Sequence code to better support vCard

jadams-tresys:
The vCard schema had triggered some edge cases that weren't found during
the development of unordered sequences. This commit deals with those
edge cases.

DAFFODIL-1034

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
@@ -200,7 +200,10 @@ final class SeqCompParser(context: RuntimeData, val childParsers: Vector[Parser]
 
 }
 
-class ChoiceParser(ctxt: RuntimeData, val childParsers: Vector[Parser])
+class ChoiceParser(
+  ctxt: RuntimeData,
+  val childParsers: Vector[Parser],
+  val unordered: Boolean = false)
   extends CombinatorParser(ctxt) {
   override lazy val runtimeDependencies = Vector()
   override lazy val childProcessors = childParsers
@@ -218,6 +221,7 @@ class ChoiceParser(ctxt: RuntimeData, val childParsers: Vector[Parser])
       var parser: Parser = null
       val limit = childParsers.length
       var returnFlag = false
+      var discriminatedFailure = false
       while (!returnFlag && i < limit) {
         parser = childParsers(i)
         i += 1
@@ -258,6 +262,14 @@ class ChoiceParser(ctxt: RuntimeData, val childParsers: Vector[Parser])
             pstate.discard(pBefore) // because disc set, we don't unwind side effects on input stream & infoset
             pBefore = null
             pstate.setFailed(allDiags)
+
+            // We are actually dealing with an unordered sequence, which uses ChoiceParser behind
+            // the scenes. Because a discriminator was set, but the discriminated content failed,
+            // we need to make sure that this discriminator information gets passed back up to the
+            // unordered sequence parser and will know to fail the entire unordered sequence.
+            if (unordered)
+              discriminatedFailure = true
+
             returnFlag = true
           } else {
             //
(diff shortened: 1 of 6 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
```diff
@@ -281,6 +293,10 @@ class ChoiceParser(ctxt: RuntimeData, val childParsers: Vector[Parser])
       }
 
       pstate.popDiscriminator
+
+      if (discriminatorFailed)
+        pstate.setDiscriminator(true)
+
```
stevedlawrence: Is this setting the correct discriminator? I was originally thinking that we would need to push a new discriminator somewhere, and then check that result someehwere else to know when to bail on the unordered sequence. But maybe that's not necessary and this is correct? This could maybe use a comment to explain where the discrimnator that this is setting gets pushed onto the stack?

jadams-tresys: The way I viewed the discriminator stack (which may be incorrect) is that the discriminator we are actually interested in is on the unordered sequence, but because we are implementing uoseq's with a ChoiceParser, that discriminator essentially gets pushed one level down to the ChoiceParser.  By setting the discriminator here, we are essentially just lifting it out of the ChoiceParser and back into the Unordered Sequence.

I added comments to the part of the code where discriminatorFailed gets set to true, but perhaps I should add something here too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -227,12 +227,18 @@ abstract class SequenceParserBase(
               pstate.setSuccess()
               isDone = true
             }
+
+            // We successfully parsed a discriminator, but failed to parse the discriminated content.
+            // Do not continue trying to parse other memebers of the unordered sequence.
+            case DiscriminatedFailure if (!isOrdered) => isDone = true
```
mbeckerle: If I am correct above and DiscriminatedFailure is only for unordered case, then I think the if(!isUnordered) should not be a conditional branch here, but an Assert.invariant(isUnordered) before the assignment of isDone = true.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -304,17 +310,21 @@ abstract class SequenceParserBase(
         }
         case _: FailedParseAttemptStatus => { // MissingSeparator with failure will match here
           Assert.invariant(pstate.isFailure)
-          if (hasPoU && !wasDescriminatorSet &&
+          if (hasPoU && !wasDiscriminatorSet &&
             (roStatus.isInstanceOf[RequiredOptionalStatus.Optional])) {
             // we back up and finish the array at the prior element if any.
+            val cause = pstate.processorStatus.asInstanceOf[Failure].cause
```
mbeckerle: Unused.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
mbeckerle: Add comment to this setDiscriminator call pointing out that the inner choice discriminator is now off the stack, so this is setting the discriminator of the outer unordered sequence loop insuring that while we have two physical discriminators (outer for the loop, inner for the choice), that this is making it behave as if there was only a single discriminator.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
jadams-tresys: That was left over from some debugging.  It is removed now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
jadams-tresys: Fixed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
```diff
@@ -154,6 +154,14 @@ object ParseAttemptStatus {
    */
   case object MissingItem extends Missing
 
+  /**
+   * Means that a failure has occurred after succesfully parsing a discriminator.
+   * For example, in a choice with initiated content if an initiator is
+   * successfully parsed, but the content of the choice branch fails, we should
+   * not continue to attempt to parse the other branches.
+   */
+  case object DiscriminatedFailure extends FailedParseAttemptStatus
```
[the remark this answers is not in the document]
jadams-tresys: Done.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
jadams-tresys: Done

PR #277, opened 2019-10-15
Remove continue-on-error from github actions

stevedlawrence:
The intention of this option was that as long as compilation succeeded
we would run all test steps, even if some one of the test steps failed.
This would allows us to potentially find multiple failures in a single
build instead of just the first failure.

Unfortunately, this option just makes it so the step is considered a
success regardless of pass/fail, so this actually hid errors. We
definitely don't want that, so remove the option where it is used.

DAFFODIL-2215


daffodil-lib/src/test/scala/org/apache/daffodil/xml/test/unit/TestXMLUtils.scala
```diff
@@ -284,7 +284,7 @@ class TestXMLUtils {
   @Test def testBlobDiff01() {
     val path1 = createBlobFile("A1B2C3")
     val path2 = createBlobFile("A1B2C3")
-    val diff = XMLUtils.computeBlobDiff("path", path1.toString, path2.toString)
+    val diff = XMLUtils.computeBlobDiff("path", path1.toUri.toString, path2.toUri.toString)
```
mbeckerle: wasn't expecting these fixes in this PR. I'm ok with it, just isn't mentioned in the description.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/sepSuppression.tdml
```diff
@@ -58,10 +58,10 @@
       for exploring when groups are 'potentially trailing' meaning their 
       separator (in enclosing sequence) might be suppressed/not-required.
      -->
-    <xs:element name="ptg1" dfdl:initiator="%NL;">
+    <xs:element name="ptg1" dfdl:initiator="|">
```
mbeckerle: Also not expecting these changes in this change set. These have nothing to do with the continue-on-error from github actions? Or are these the tests that were failing but were being masked?

daffodil-lib/src/test/scala/org/apache/daffodil/xml/test/unit/TestXMLUtils.scala
stevedlawrence: When I made the PR, I didn't realize there were actual failures (I noticed this originally in another PR that did have failures). But it turned out, we've had tests failing on windows for a while and I didn't notice until the option was removed and a build happened. Since it was kindof related, I quickly fix the tests so they'd pass on windows. I'll fix the commit message when I squash everything together once all the tests pass.

PR #278, opened 2019-10-17
Add tests for recent bugs found. Improve a trailingEmptyStrict test.

mbeckerle:
Just adds tests (commented out) for these bugs to assist in diagnosis/fixing later. 

DAFFODIL-2219, DAFFODIL-2217

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-5.tdml.xml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-5.tdml.xml
@@ -1,0 +1,169 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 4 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-5.tdml.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mbeckerle: banner

daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-6.tdml.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mbeckerle: banner

PR #273, opened 2019-10-03
WIP: Add User Defined Functions Capability

olabusayoT:
Requesting preliminary review of Daffodil's UDF capability implementation.

Tests are still in progress so code/changes in daffodil-test and daffodil-udf/src/test should be ignored.

PR #279, opened 2019-10-23
User Defined Functions feature

olabusayoT:
Review for UDF feature (with changes from PR #273 implemented)

Tests are still in progress

- Adds daffodil-udf submodule that has the files needed to write a UDF
- Adds UDFExpr/UDFExprCall classes
- Adds readme doc (which will be posted on website when complete)
- Adds checks, warnings and errors to our UDFService loader
 - invalid UDFs and UDFPs are dropped, and emit a warning
 - if an attempt is made to use a dropped UDF, an SDE will occur
- Adds error handling from UDFs during evaluate execution

DAFFODIL-2186

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
@@ -1,0 +1,589 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.udf
+
+import org.junit.AfterClass
+import org.junit.Test
+import junit.framework.Assert._
+import org.apache.daffodil.CLI.Util
+import net.sf.expectit.matcher.Matchers.contains
+import net.sf.expectit.matcher.Matchers.anyOf
+import net.sf.expectit.matcher.Matchers.allOf
+import net.sf.expectit.matcher.Matchers.eof
+import java.nio.file.FileSystems
+import java.nio.file.Files
+import scala.collection.JavaConverters._
+
+class TestCLIUdfs {
+
+  lazy val testUdfsPaths = {
+    val dir = FileSystems.getDefault.getPath(Util.daffodilPath("daffodil-udf/target"))
+    val allClassFiles = Files.walk(dir).iterator.asScala
+      .filter { f => Files.isRegularFile(f) && f.toString.endsWith("class") }
+    val commonPrefix = allClassFiles
+      .flatMap {
+        fp =>
+          val filePath = fp.toString
+          val indexOfOrg = filePath.indexOfSlice("org")
+          if (indexOfOrg >= 0) {
+            val prefix = filePath.splitAt(indexOfOrg)._1
+            Some(prefix)
+          } else None
+      }.toList.distinct
+    commonPrefix
+  }
+
+  /**
+   * Tests the case when no User Defined Functions on classpath, and a schema makes
+   * no User Defined Function calls, so they don't get loaded
+   */
+  @Test def test_noUdfsLoaded_regular_schema() {
+    val schemaFile = Util.daffodilPath("daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd")
+    val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
+
+    val shell = Util.start("")
+
+    try {
+      val cmd = String.format("echo -n strng| %s -v parse -s %s -r fn_func", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+      shell.expect(
+        allOf(
+          contains("<fn_func>"),
+          contains("data>strng</data>"),
+          contains("<value>Hello,strng</value>"),
+          contains("</fn_func>")))
+
+      shell.send("exit\n")
+      shell.expect(eof)
+      shell.close()
+    } finally {
+      shell.close()
+    }
+  }
+
+  /**
+   * Tests the case when no User Defined Function are loaded, but a schema makes a
+   * User Defined Function call
+   */
+  @Test def test_noUdfsLoaded_udf_schema() {
+    val schemaFile = Util.daffodilPath("daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd")
+    val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
+
+    val shell = Util.startIncludeErrors("")
+
+    try {
+      val cmd = String.format("echo -n strng| %s -v parse -s %s -r user_func1", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+      shell.expectIn(
+        1,
+        allOf(
+          contains("[info] No User Defined Functions loaded."),
+          contains("[error] Schema Definition Error: Unsupported function: jsudf:replace")))
+
+      shell.send("exit\n")
+      shell.expect(eof)
+      shell.close()
+    } finally {
+      shell.close()
+    }
+  }
+
+  /**
+   * Tests the case when no User Defined Function are loaded, due to
+   * ServiceConfigurationError such as an invalid META-INF file referencing a class
+   * missing from the classPath
+   *
+   * The schema makes a User Defined Function call
+   */
+  @Test def test_noUdfsLoaded_MissingClassInMetaInfFile() {
+    val schemaFile = Util.daffodilPath("daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd")
+    val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
+
+    val metaForNonExistentClass = "daffodil-udf/src/test/resources/org/badmetainf/nonexistentclass/"
+
+    val dafClassPath =
+      (testUdfsPaths :+ Util.daffodilPath(metaForNonExistentClass))
+        .mkString(":")
+
+    val shell = Util.startIncludeErrors("", envp = Map("DAFFODIL_CLASSPATH" -> dafClassPath))
+
+    try {
+      val cmd = String.format("echo -n strng| %s -v parse -s %s -r user_func1", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+      shell.expectIn(
+        1,
+        allOf(
+          contains("[error] Error while loading User Defined Function Providers:" +
+            " org.apache.daffodil.udf.UserDefinedFunctionProvider:" +
+            " Provider org.nonexistentclass.example.StringFunctions.StringFunctionsProvider not found"),
+          contains("[info] No User Defined Functions loaded."),
+          contains("[error] Schema Definition Error: Unsupported function: jsudf:replace")))
+
+      shell.send("exit\n")
+      shell.expect(eof)
+      shell.close()
+    } finally {
+      shell.close()
+    }
+  }
+
+  /**
+   * Tests the case when no User Defined Function are loaded, due to absent META-INF
+   * file, but a schema makes a User Defined Function call
+   */
+  @Test def test_noUdfsLoaded_MissingMetaInfFile() {
+    val schemaFile = Util.daffodilPath("daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd")
+    val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
+
+    val dafClassPath =
+      testUdfsPaths.mkString(":")
+
+    val shell = Util.startIncludeErrors("", envp = Map("DAFFODIL_CLASSPATH" -> dafClassPath))
+
+    try {
+      val cmd = String.format("echo -n strng| %s -v parse -s %s -r user_func1", Util.binPath, testSchemaFile)
+      shell.sendLine(cmd)
+      shell.expectIn(
+        1,
+        allOf(
+          contains("[info] No User Defined Functions loaded."),
+          contains("[error] Schema Definition Error: Unsupported function: jsudf:replace")))
+
+      shell.send("exit\n")
+      shell.expect(eof)
+      shell.close()
+    } finally {
+      shell.close()
+    }
+  }
+
+  /**
+   * Tests the case when a User Defined Function Provider returns null for its list of UDFs
+   */
+  @Test def test_UDFPClass_NoUdfClasses() {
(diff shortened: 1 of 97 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/UserDefinedFunctionBase.scala
```diff
@@ -0,0 +1,93 @@
...
+  errorCause: Maybe[Throwable],
+  errorStr: Maybe[String])
+  extends ProcessingError(errorInfo, schemaContext, dataContext, errorCause, errorStr)
+
+/**
+ * User Defined Function Exception class to wrap fatal errors from the UDF
+ * into Daffodil Aborts
+ */
+case class UserDefinedFunctionFatalErrorException(description: String = "", cause: Throwable, classOfInterest: String = "")
+  extends Abort(description ++ cause.getMessage) {
```
mbeckerle: You probably want at least one space between these two parts of the message.

daffodil-test/src/test/scala/org/apache/daffodil/udf/TestUdfInSchemas.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  @AfterClass def shutDown {
+    runner.reset
+  }*/
+
+}
+
+class TestUdfInSchemas {
+  import TestUdfInSchemas._
+  /*  @Test def test_udf_argsIncorrect() { runner.runOneTest("udf_argsIncorrect") }
```
mbeckerle: Why are these all commented out?

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -0,0 +1,271 @@
...
+      out.defaultWriteObject()
+    }
+
+    @throws(classOf[java.io.IOException])
+    @throws(classOf[java.lang.ClassNotFoundException])
+    private def readObject(in: ObjectInputStream): Unit = {
+      in.defaultReadObject()
+      this.method
+    }
+  }
```
stevedlawrence: Can you remove the writeObject since all it oes it all the default? We should probably also call "method" in the constructor to force the lookup when this is created rather than the first time the lazy "method" is used so all lookups happen at compile time or deserialization time.

daffodil-test/src/test/scala/org/sgoodudfs/example/StringFunctions/Reverse.scala
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sgoodudfs.example.StringFunctions
+
+import org.apache.daffodil.udf.UserDefinedFunction
+import org.apache.daffodil.udf.UserDefinedFunctionIdentification
+import com.sun.xml.internal.bind.v2.runtime.IllegalAnnotationsException
```
stevedlawrence: Can this exception be removed? Since these are good examples, these should be perfect. And an import of someting internal seems odd.

Also, I'd recommend combining the example scala UDF's and provider in a single file. I think in the real world that's how people should organize their scala UDFs so we should show that in our example. Java's requirement of splitting out classes into separate files is annoying, especially to people familar with Scala.

daffodil-test/src/test/scala/org/apache/daffodil/udf/TestUdfInSchemas.scala
olabusayoT: I was holding off on tests till the code was as close to complete as possible. Tests are in progress now.

daffodil-udf/src/test/scala/org/sbadudfs/udfexceptions/StringFunctions/Reverse.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * UDF for exceptions negative tests in Scala
+ *
+ * Throws exception on evaluate
+ */
+@UserDefinedFunctionIdentification(name = "reverse",
+  namespaceURI = "http://scala.udf.com")
+class Reverse extends UserDefinedFunction {
+  final case class CustomException(
```
stevedlawrence: Looks like this final case class is causing build warnings/failures. Not sure what the problem is, but the solution might be to just move it outside of the UDF, or define it in the provider and share it among these bad UDFS?

eclipse-projects/runtime1/.classpath
```diff
@@ -1,92 +1,112 @@
...
-          </classpathentry><classpathentry sourcepath="lib_managed/srcs/org.jdom/jdom2/jdom2-2.0.6-sources.jar" exported="true" kind="lib" path="lib_managed/jars/org.jdom/jdom2/jdom2-2.0.6.jar">
-            <attributes>
-                  <attribute name="javadoc_location" value="jar:file:lib_managed/docs/org.jdom/jdom2/jdom2-2.0.6-javadoc.jar!/"/>
-                </attributes>
-          </classpathentry>
-      </classpath>
\ No newline at end of file
+	<classpathentry kind="src" path="src/main/scala"/>
+	<classpathentry kind="con" path="org.scala-ide.sdt.launching.SCALA_CONTAINER"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
```
stevedlawrence: I think we need to be careful about these classpath changes? Don't we have Utility to update these? Does that utility need to be updated to generate the udf classpath? I don't mind if that's done as a separate commit.

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
olabusayoT: Hmm, if we force the lookup at compile time (i.e in the constructor), then we'll do a lookup for every UDF found on the classpath. Even if it's never used. As it stands right now, we only do a lookup at deserialization, and even then only for the UDFs in the schema.

mbeckerle: I think the right behavior should be only UDFs that are used by the schema should get looked up. I should be able to have some older, stale/broken UDF on the classpath, so long as I don't use it it shouldn't bother me. Make sense?

In general, object construction time should never do anything that can possibly fail/throw an exception. Otherwise when it fails, well the object itself isn't even fully constructed and that can make it hard to figure out what is going on.  ( This is one of the reasons scala lazy val is so useful just for ordinary object-oriented programming. )

olabusayoT: Hmm agreed. We could force the lookup to happen during compile time, when we do the udf lookup (lookupUserDefinedFunctionCallingInfo). This optimizes things for the regular parse, but duplicates the lookup if we are doing a saved parse as the UDFMethod.method will get thrown out during serialization, and repopulated at deserialization.

(As it stands right now, we actually do the method lookup for the regular parse at execution time, which is not ideal.)

daffodil-udf/src/main/java/org/apache/daffodil/udf/exceptions/UserDefinedFunctionFatalException.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.udf.exceptions;
+
+/**
+ * Exception to throw to abort parsing/unparsing.
+ */
+public class UserDefinedFunctionFatalException extends Exception {
+  public UserDefinedFunctionFatalException(String errorMessage) {
```
bsloane1650: Should this accept a cause as well?

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -0,0 +1,271 @@
...
+        goodFunctionClasses.foreach { fc =>
+          val fcClassName = fc.getName
+          val fInfo = fc.getDeclaredAnnotation(classUserDefinedFunctionIdentification)
+          val fns = fInfo.namespaceURI()
+          val fname = fInfo.name()
+          val invalidFns = Misc.isNullOrBlank(fns)
+          val invalidFname = Misc.isNullOrBlank(fname)
+          val evalMethods = fc.getMethods.filter(_.getName == evaluateMethodName)
+
+          if (invalidFns || invalidFname) {
```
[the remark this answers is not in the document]
olabusayoT: I think my idea behind this was to group the annotation checks, but have separate error message for the name and the namespace. So all errors on the annotations would be reported once (or at the same time), and an error with the first checked one doesn't mask the other's error message. Is this too granular?

daffodil-test/src/test/java/org/jgoodudfs/example/StringFunctions/Compare.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package org.jgoodudfs.example.StringFunctions;
+
+import org.apache.daffodil.udf.UserDefinedFunction;
+import org.apache.daffodil.udf.UserDefinedFunctionIdentification;
+
+@UserDefinedFunctionIdentification(
+    name = "compare",
+    namespaceURI = "http://goodudfs.StringFunctions")
```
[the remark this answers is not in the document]
olabusayoT: If you mean for the example UDFs, we can definitely update it, but code wise we don't actually place any restrictions on the namespaceURI of the UDF (as long as its a valid string), so any collisions would need to be resolved by the writer of the schema/UDF

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -0,0 +1,271 @@
...
+            val paramTypes = evaluateMethod.getParameterTypes
+            val returnType = evaluateMethod.getReturnType
+
+            val initParamTypeConv = paramTypes.map { c => NodeInfo.fromClass(c) }
+            val initRetTypeConv = NodeInfo.fromClass(returnType)
+            val badParamIndex = initParamTypeConv.indexOf(None)
+            val invalidParamType = badParamIndex >= 0
+            val invalidRetType = initRetTypeConv.isEmpty
+
+            if (invalidParamType || invalidRetType) {
```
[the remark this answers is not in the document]
olabusayoT: same idea as with these as in https://github.com/apache/incubator-daffodil/pull/279#discussion_r339737162, it's to group the evaluate method check

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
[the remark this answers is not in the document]
mbeckerle: I understand what you were going for here, but I think I'm with Brandon on style here. I think this is too granular. Think of it this way: under maintenance, I want to add another check,... I have to add the predicate in two places, not one.

daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
```diff
@@ -0,0 +1,565 @@
...
+  /**
+   * Tests the cases when a function class:
+   *   doesn't have an evaluate function
+   *   has overloaded the evaluate function
+   *   has unsupported param types
+   *   has unsupported return type
+   */
+  @Test def test_UDFClass_noEvaluate() {
+    val schemaFile = Util.daffodilPath("daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd")
+    val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
```
mbeckerle: I'm sick of seeing this redundant windows conditionalization, so FYI: I created DAFFODIL-2226 which is to go back to all the CLI-oriented testing not just of the UDFs, but everywhere, and abstract out the need for this conditionalization.

daffodil-udf/README.md
```diff
@@ -39,39 +39,82 @@ The class can provide as many User Defined Function classes as it wishes, as lon
...
+#### User Defined Function Exceptions
+
+Daffodil proves two exception classes for users that wish to throw an error from within their UDFs.
+* `UserDefinedFunctionProcessingError` can be thrown when the implementer wishes to cause backtracking during parsing
+* `UserDefinedFunctionFatalException` can be thrown when the implementer wishes processing to be aborted all together
+
+All other exceptions are treated as `UserDefinedFunctionFatalException`.
+
+### UDF Registration
+Once implemented, a UDF can be expected to have one of the following structure
```
mbeckerle: Typo: need final "s" on "the following structures"

daffodil-udf/README.md
```diff
@@ -39,39 +39,82 @@ The class can provide as many User Defined Function classes as it wishes, as lon
...
+      example/
+          StringFunctions/
+            StringFunctionsProvider.java //UDF Provider
+            Compare.java //UDF class
+            Replace.java //UDF class
+  META-INF/
+    services/
+      org.apache.daffodil.udf.UserDefinedFunctionProvider
+```
+Each UDF is registered by including the fully qualified name of its provider in the `META-INF/services/org.apache.daffodil.udf.UserDefinedFunctionProvider`. The META-INF folder must be accessible from the root of whatever paths are on the classpath, otherwise it won't be picked up by ServiceLoader.
```
mbeckerle: It's not clear from the way this is worded that org.apache...UserDefinedFunctionProvider is the name of a file that is supposed to be textual and contain the name. Say "in a text file named 'org.apache...UserDefinedFunctionProvider'"

Also I learned an interesting style idea when writing markdown which is https://rhodesmill.org/brandon/2012/one-sentence-per-line/. 

I'm not suggesting further change here in your file, but it's an idea I'm planning to embrace when writing markdown from here on.

daffodil-udf/src/test/scala/org/sgoodudfs/example/SupportedTypesFunctions/SupportedTypesFunctionsProvider.scala
```diff
@@ -0,0 +1,236 @@
...
+ * limitations under the License.
+ */
+package org.sgoodudfs.example.SupportedTypesFunctions
+
+import org.apache.daffodil.udf.UserDefinedFunction
+import org.apache.daffodil.udf.UserDefinedFunctionProvider
+import org.apache.daffodil.udf.UserDefinedFunctionIdentification
+
+/**
+ * Example User Defined Function Provider in Scala
```
mbeckerle: Can you explain what this file is about?  It's weird and immediately caught my eye. This comment is boilerplate. You are testing all the implicit boxing conversions inbound and for return types right? A user would never do anything like this. Right?

olabusayoT: Ah so, this file is supposed to be contain UDFs that exercise the types we support. So checking that we can pass in (params) and receive back the types we claim to support.

eclipse-projects/runtime1/.classpath
olabusayoT: We do, see https://github.com/apache/incubator-daffodil/pull/279/commits/7293a23a308dea72743502ddf4228e4ba4265094

mbeckerle:
Codecov report indicates you are missing tests that manually throw the fatal and non-fatal exceptions you have defined for intentional explicit use by end-users. I thought I saw tests of those though, so this may be spurious.

Other than that the non-coverage things look to be limitations of the scala compiler. E.g., it seems to always say this line is not covered:

`val Some(x) = y`

probably because it doesn't know that y is always type Some on this code path, so it thinks this assignment can fail possibly, and that failure path is not covered by tests of course, because it actually is not really possible. Rewriting as:

`val x = y.get`

Should remove the coverage red-mark, because the type-check code path is now over in the definition of the get method of the Option type, not here in this code. 

The latter may also generate better/smaller code. Or rather, we know for sure what y.get is going to do, whereas the pattern assignment may or may not (depending on optimizers) contain calls to unapply methods and other artifacts of how pattern destructuring works.

PR #281, opened 2019-10-31
Remove heap sized limit for parsing

jadams-tresys:
This involved setting a maximum number of buckets and removing old
buckets once that maximum has been reached.

This only fixes the parse side of DAFFODIL-2194.  The unparse side of things still has a hard 2GB limit.

DAFFODIL-2194


daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    try {
+      val cmd = String.format("%s parse -s %s %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      val result = shell.expect(contains("tns:blob_01")).toString
+      // Use +7 to drop the 'file://' to get the path of the file
+      val generated_blob = result.substring(result.indexOf("file://") + 7).takeWhile(_ != '<')
+
+      //val checksum = String.format("%s parse -s %s %s", Util.binPath, testSchemaFile, testInputFile)
+      val cmp = String.format("/usr/bin/cmp %s %s", generated_blob, testInputFile)
```
stevedlawrence: This is going to fail on windows. Maybe refactor the logic in XMLUtils used for TDML comparing blobs so that we can use it here as well? Not everything in a CLI tests has to be done via shell.send/expect.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -142,8 +145,9 @@ abstract class InputSource {
  *
  * @param inputStream the java.io.Inputstream to read data from
  * @param bucketSize the size of each individual bucket
+ * @param maxNumberOfBuckets the maximum number of buckets that can be created before beginning to remove old ones
  */ 
-class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1 << 13)
+class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1 << 13, maxNumberOfBuckets: Int = 32768)
```
stevedlawrence: If I'm doing the math right, a bucketsize of 1<<13 is 4096 bytes, that times 32768 comes out to a little over 100MB. So the most we could backtrack by default is 100MB. That doesn't seem totally unreasonble?

A though, maybe instead of maxNumberOfBuckets passed it, we pass in maxCacheSizeInBytes or something and then convert that to maxNumberOfBuckets internally? That way users don't have to do the math I did above.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -357,18 +372,25 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
...
+
     buckets(bucketIndex.toInt).refCount += 1
   }
 
   def releasePosition(bytePos0b: Long): Unit = {
     val (bucketIndex, _) = bytePositionToIndicies(bytePos0b)
+
+    if (buckets(bucketIndex.toInt) == null)
+      return
+
```
stevedlawrence: Instead of returning, does it make sense to ignore the refCount for that bucket, but then continue to release buckets? I guess there now might be a bunch of releaseble buckets, but I guess also who cares? Since the next read will result in the backtrack exception? If it doesn't make sense to just say who care and not try to release, this could maybe use a comment as to why we don't care.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -188,7 +189,14 @@ final class PState private (
...
     // threadCheck()
     val m = markPool.getFromPool(requestorID)
-    m.captureFrom(this, requestorID)
+    try {
+      m.captureFrom(this, requestorID)
+    } catch {
+      case b: BacktrackingException => {
+        markPool.returnToPool(m)
+      }
+      throw b
```
stevedlawrence: I assume this comes from a lock throwing the exception? This would go away too, I think?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -531,7 +531,15 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
     val m = markPool.getFromPool(requestorID)
     m.assignFrom(cst)
     markStack.push(m)
-    inputSource.lockPosition(m.bytePos0b)
+    try {
+      inputSource.lockPosition(m.bytePos0b)
+    } catch {
+      case b: BacktrackingException => {
```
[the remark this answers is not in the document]
jadams-tresys: Makes sense.  Will change.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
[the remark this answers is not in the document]
stevedlawrence: We'd have to change the API in order to support it as a tunable. Right now, the user creates the InputSourceDataInputStream explicitly, which doesn't know anything about tunables. We either need a way to get tunables from the DataProcessor and pass it in as an argument, or maybe the DataProcessor becomes a factory that creates InputSourceDataInputStreams and itself passes in the necessary tunable information?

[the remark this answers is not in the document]
stevedlawrence: Speaking of which, our current API doesn't allow changing the bucketSize or this new maxNumbeOfBuckets capabilities. The JAPI and SAPI only allow passing in an InputStream, but has not configurability. Perhaps supporting that from tuanbles becomes part of this new API. Probably needs to be part of the next release so that people can grow this limit if needed, since before it was only limited by memory.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -22,6 +22,9 @@ import java.nio.ByteBuffer
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.exceptions.ThinThrowable
+
+case class BacktrackingException() extends Exception with ThinThrowable
```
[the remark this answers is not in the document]
stevedlawrence: Definitely agreed. I think it's also important to point out that the act of backtracking (i.e releasePosition) itself does not cause an error. The error only occurs when someone tries to read from that point. I could imagine a case where you try to parse this huge optional thing, we eventually find out it's optional and throw away the hole thing, and then the parse is done. We can actually return a valid infoset in that case, even though we've backtracked too far to read anymore data.

Note that if someone tried to use this for a second parse (i.e. stream) then would hit the backtrack error. This suggests that the comment I made above about marking an input source as invalid vs throw a backtrack error the bucket is negative, the right solution is throw when negative since the input source wouldn't be invalid in this case.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -340,6 +352,9 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
       while (bytesStillToGet > 0) {
         val bytesToGetFromCurrentBucket = Math.min(bucketSize - byteIndex, bytesStillToGet).toInt
 
+        if (buckets(bucketIndex.toInt) == null)
```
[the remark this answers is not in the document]
mbeckerle: Interesting point. If one call to parse runs into this bucket overflow situation, they "ruin" the input stream for subsequent calls to parse, should the parse return in error. We can't unwind back to the starting position for that call to parse. We've lost that location. 

Clearly a Runtime SDE is fatal, and not recoverable, but it indicates more than just a failure of that call to parse. Rather,  the whole input stream is also unusable now, unless we can somehow know that it was successfully restored to where it started with the parse call.  The first top-level call to parse for a data input stream could save the current position. On unwind from that parse call, we could check if the stream has been restored to that position (or can be restored to that position - and restore it to that when possible), and if not mark it as in some "broken due to pre-advance by failed parse" state. Then an attempt to reuse that data input stream would throw indicating why the stream is unsuitable for use.

daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
```diff
@@ -0,0 +1,191 @@
...
+    val schemaFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/1GB_blob.dfdl.xsd")
+    val inputFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/1GB_blob.bin")
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile)) else (schemaFile, inputFile)
+
+    //val shell = Util.startIncludeErrors("", envp = DAFFODIL_JAVA_OPTS)
+    val shell = Util.startIncludeErrors("", envp = DAFFODIL_JAVA_OPTS)
+
+    try {
+      // Create 1GB blob file
+      val dd = String.format("/usr/bin/dd if=/dev/zero of=%s bs=1M count=1000", testInputFile)
```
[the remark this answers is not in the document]
mbeckerle: We can feed 1G+ of data to a pipe using only smallish buffers using Scala/Java code. 
Thread 1 writes into a PipedOutputStream, and main thread's daffodil's input stream is created from the corresponding PipedInputStream, and reads from it as normal.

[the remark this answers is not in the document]
stevedlawrence: That's still going to result in a blob files being created since we only support file based blobs. It it does at least avoid haing the input files having to be on disk. Definiately a better route to go. Also, it doesn't seem too unreasonable to require 1GB+ of free space when running integration tests. This are mostly only ever run on CI machines where that shouldn't be a problem as long as we make sure to delete the blobs after a test finishes.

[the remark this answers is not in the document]
mbeckerle: I am coming around to your view that tests that push the real boundaries of this stuff should be manual test steps that we just run on specific machines. E.g., I think we need to do a 16GB test on a 4GB JVM that parses data including huge blob, then unparses, I also think we need  a test where that same huge blob comes out of a gzipped layer, so the blob, while huge, can come from a tiny test file (assuming the file is for example 16GB of the same 8 bytes 0xDEADBEEF)
I want to be sure there is nothing finite and problematic in the layering stuff. I fear there may be, though it mostly just reuses the I/O layer, so there's a chance this just composes.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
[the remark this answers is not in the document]
jadams-tresys: I think my most recent commit addresses this with the added check on Pstate.isDataStreamValid, but I'm not sure I understand how this situation works where the user initiates another parse with the same data input stream, so I don't have any tests to cover this.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -145,16 +173,21 @@ abstract class InputSource {
  *
  * @param inputStream the java.io.Inputstream to read data from
  * @param bucketSize the size of each individual bucket
- * @param maxNumberOfBuckets the maximum number of buckets that can be created before beginning to remove old ones
+ * @param maxCacheSizeInBytes the maximum number of buckets that can be created before beginning to remove old ones
```
mbeckerle: Description isn't right or doesn't match name of parameter. This is the total size of the buckets, not the number of buckets.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -24,7 +24,33 @@ import scala.collection.mutable.ArrayBuffer
...
+ * not BLOB specific. A format with a choice, then in the first branch of the
+ * choice, a group and/or array of small data items, ultimately totaling in size
+ * to more than the backtrack limit, and then a branch failure, will cause this
+ * backtracking limit error with no BLOBs being used.
+ *
+ * There is no limit created by this module of code to the size of a any data
+ * item having to fit within JVM byte array maximums, nor the JVM memory
+ * footprint even. Only a limitation on backtracking distance is created.
+ *
+ * It is also worth noting that it is not the act of backtracking does not
```
mbeckerle: Phrasing is not right.  "is not".... "does not".

daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
```diff
@@ -82,40 +69,41 @@ class TestBlob {
...
+   * automatic regression tests on the build servers.  In order to run them you
+   * will need to unzip the file:
+   *
+   * daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/2_01GB.zip
+   *
+   * and make sure that the file "2_01GB.bin" ends up in:
+   *
+   * daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/
+   */
+  /*@Test def test_2GB_blob() {
```
mbeckerle: You can comment out a test by just commenting out the Test marker
`/* @Test */
def thisTestCommentedOut() {....
}`

mbeckerle:
RatChecks are failing because of new binary files that have to be made exceptions perhaps?

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/gen_blob.py
```diff
@@ -0,0 +1,18 @@
+#! /usr/bin/python
```
mbeckerle: Need apache banner. RatCheck is failing due to this.

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/gen_blob.py
```diff
@@ -0,0 +1,18 @@
...
+    parser.add_argument("-o", "--output", help="Path to the file that will be generated")
+    args = parser.parse_args()
+
+    bytes_written = 0
+    with open(args.output, 'wb') as fout:
+        while (bytes_written < (args.size * 1024000)):
+            fout.write(binascii.unhexlify('deadbeef'))
+            bytes_written += 4
+
+    print("Wrote %d bytes" % bytes_written)
```
stevedlawrence: Was this used to generate 2_01GB.zip? In general, Apache reccommends that anything that is generated not be included in a repo, but instead include instructions on how to generate it. Either way a user needs to perform an action to get the 2GB file (either run this script or unzip a file), so I'd rather them just run this script.

Another though, it might make for a more interesting test if the first 8 bytes were an unsigned long of the number of bytes. That's more realistic to how most blob files are going to look, and it allows the schema to have outputValueCalc for unparse and runtime calculated lengths parse. And then you can have a single schema that could parse anything that this scrip could generate without having to modify the schema.

daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
```diff
@@ -23,31 +23,20 @@ import java.io.File
...
-  val output9 = Util.getExpectedString("output9.txt")
-  val output10 = Util.getExpectedString("output10.txt")
-  val output12 = Util.getExpectedString("output12.txt")
-
-  // Call md5sum with two different files and return if they match
-  /*def md5sum(shell: Expect, a: String, b: String): Boolean {
-    val cmdA = String.format("/usr/bin/md5sum %s | awk '{print $1}'", a)
-    shell.sendLine(cmd)
-    val resA = shell.expect("\n").toString
+  def md5sum(shell: Expect, blob_path: String, checksum: String): Unit = {
```
stevedlawrence: MIght make more sense to put this in the Utils.scala file. Other CLI tests that aren't blobs might find this useful.

daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
```diff
@@ -64,15 +53,13 @@ class TestBlob {
...
-      shell.sendLine("echo $?")
-      shell.expect(contains("0\n"))
+      md5sum(shell, generated_blob, "a8430718a3dc7cab13d28c8838e298c4")
 
       // Clean up blobs
-      shell.sendLine("rm -rf daffodil-blobs")
+      if (Util.isWindows)
+        shell.sendLine("rmdir /Q /S daffodil-blobs")
+      else
+        shell.sendLine("rm -rf daffodil-blobs")
```
stevedlawrence: Thoughts on putting this is a finally block? If something fails we could be leaving around some giant files.

Also suggest moving this to a utility function.

daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
stevedlawrence: I actually kindof prefer commenting out the whole thing. It makes it much more obvious that a test isn't going to be run when the whole thing is syntax highlighted as a comment rather than just the one ``@Test`` annotation.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -145,16 +173,21 @@ abstract class InputSource {
...
+  maxCacheSizeInBytes: Int = 512000)
   extends InputSource {
 
   private class Bucket {
     var refCount = 0
     val bytes = new Array[Byte](bucketSize)
   }
 
+  private val maxNumberOfBuckets = maxCacheSizeInBytes / bucketSize
+
```
stevedlawrence: Whap happens if maxCacheSizeInBytes is less than bucketSize? maxNumberOfBuckets will be zero. Does that still allow one bucket to exist or will things fail? Do things even work if we allow only a single bucket? I'm wondering if we should force maxNumberOfBuckets to be at least 1, or maybe 2?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -325,8 +361,10 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
     } else {
       val (bucketIndex, byteIndex) = bytePositionToIndicies(curBytePosition0b)
 
-      if (buckets(bucketIndex.toInt) == null)
-        throw new BacktrackingException
+      if ((bucketIndex < 0) || (buckets(bucketIndex.toInt) == null)) {
+        _isValid = false
```
stevedlawrence: Should this _isValid change happen somewhere more generically instead? I'm thinking maybe in doParse(), where if the final parse result fails because of an SDE then the InputSource becomes invalid? While it's true that this specific case makes the InputSource invalid, any SDE (which this will eventually become one) also make it invalid.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -325,8 +361,10 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
...
+        throw new BacktrackingException("Attempted to backtrack to position %d, which exceeds maximum backtrack length of %d".format(bucketIndex, maxCacheSizeInBytes))
```
stevedlawrence: This says "position %d", but passes in bucketIndex as the parameter, which isn't a position, but an internal thing that really won't mean anything to a user. I think we instead want curBytePosition or something? Might also say "backtrack to byte %d" since sometimes we report bit positions.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -379,20 +419,17 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
   def lockPosition(bytePos0b: Long): Unit = {
     val (bucketIndex, _) = bytePositionToIndicies(bytePos0b)
     Assert.invariant(bucketIndex < buckets.length)
-    if (buckets(bucketIndex.toInt) == null)
-      throw new BacktrackingException
-
-    buckets(bucketIndex.toInt).refCount += 1
+    if (!(buckets(bucketIndex.toInt) == null))
```
stevedlawrence: ``!= null`` would be more clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -228,6 +228,8 @@ class DataProcessor(val ssrd: SchemaSetRuntimeData)
     var optThrown: Maybe[Throwable] = None
     try {
       try {
+        Assert.usageErrorUnless(state.dataInputStreamIsValid, "Attempted to use an invalid input source. This can happen due to our position in the input source being preadvanced by a failed parse beyond the length that we can backtrack")
```
stevedlawrence: "preadvanced" sounds odd here. Is there maybe better way to say this? To me that sounds like we're moving forward in the input source but this is really a backtrack issue.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -142,15 +173,21 @@ abstract class InputSource {
  *
  * @param inputStream the java.io.Inputstream to read data from
  * @param bucketSize the size of each individual bucket
+ * @param maxCacheSizeInBytes the maximum number of buckets that can be created before beginning to remove old ones
  */ 
-class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1 << 13)
+class BucketingInputSource(
+  inputStream: java.io.InputStream,
+  bucketSize: Int = 1 << 13,
+  maxCacheSizeInBytes: Int = 512000)
```
stevedlawrence: This is only 0.5 MB. Any concerns with this? While it should be large enough for any backtracking case, we certainly have files that are significantly larger than this. So this could break backwards compatbilitiy if they actually relied on the ability to backtrack that far. And there is currently no way for a user to change it if it is too small.

This seems like a reasonable default only if there is a way for the user to change it. Maybe that can be done as another bug for this release? I know another comment mentioned some ways to allow configuring this via tunables.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -340,6 +390,11 @@ class BucketingInputSource(inputStream: java.io.InputStream, bucketSize: Int = 1
       while (bytesStillToGet > 0) {
         val bytesToGetFromCurrentBucket = Math.min(bucketSize - byteIndex, bytesStillToGet).toInt
 
+        if ((bucketIndex < 0) || (buckets(bucketIndex.toInt) == null)) {
+          _isValid = false
+          throw new BacktrackingException("Attempted to backtrack to position %d, which exceeds maximum backtrack length of %d".format(bucketIndex, maxCacheSizeInBytes))
```
stevedlawrence: This duplicates the backtracking string. Thoughts on having the BacktrackingException just accept a psotion and max backtrack length, and then the exception can build the string in the constructor?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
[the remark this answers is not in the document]
stevedlawrence: I think you'd have to write some unit test code, and maybe a CLI test using the ``--stream`` option would be good too.

Essentially, you need data + a schema that will fail due to a backtracking exception. But you need to continue to try to parse that. I *think* the CLI --stream option will abort the parse and won't try again, so it would be good to have that CLI tests. A unit test would look something like

```
val input = // create an InputSourceDataInputStream()
val output = // create infoset outputter

val pr1 = dataProcessor.parse(input, output) // should cause an backtracking SDE result
val pr2 = dataProcessor.parse(input, output) // should cause an invalid usage error
```
So we try to parse the data once, that fails and makes our input source invalid. Then we try to parse again using the same input source, but that should fail since the input is not invalid.

Thinking about this, this doesn't even need to be a backtracking SDE. Any runtime SDE should work.

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/gen_blob.py
jadams-tresys: I'm a little confused about the schema having an outputValueCalc. I updated the schema to parse the first 8 bytes as blobLength, and then had the actual blob element reference that length. Should I be using an outputValueCalc somewhere?

mbeckerle: Yes, have that length Element have an OVC: like this:
`dfdl:outputValueCalc="{ dfdl:valueLength(../blobelement) }"`

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
```diff
@@ -945,4 +946,55 @@ class TestScalaAPI {
...
+    try {
+      res = dp.parse(input, outputter)
+    } catch {
+      case e: Abort => {
+        assertTrue(e.getMessage().contains("Usage error"))
+        assertTrue(e.getMessage().contains("invalid input source"))
+      }
+    }
+  }
+
```
stevedlawrence: Would be good to duplicate this in the Java API. Should behave exactlty the same, but in general, anything added as a ScalaAPI test is also added as a Java API test.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
jadams-tresys: Missed this comment somehow.  Should be addressed in the most recent commit.

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
jadams-tresys: Done

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -969,4 +969,52 @@ public void testJavaAPI18() throws IOException, ClassNotFoundException {
...
+   */
+    @Test
+    public void testJavaAPI19() throws IOException, ClassNotFoundException {
+      // Demonstrate that we cannot use the API to continue a parse with an invalid InputSource
+      // ie. after a runtime SDE. This test needs to be run with an input file larger than 256MB
+      org.apache.daffodil.japi.Compiler c = Daffodil.compiler();
+      c.setValidateDFDLSchemas(false);
+      java.io.File schemaFile = getResource("/test/japi/blob_backtracking.dfdl.xsd");
+      c.setDistinguishedRootNode("root", null);
+      ProcessorFactory pf = c.compileFile(schemaFile);
```
stevedlawrence: Looks like this test is failing here. I think blob_backtracking.dfdl.xsd need to be added?

jadams-tresys: Yup, forgot to add it, also I forgot to comment it out.

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -984,7 +984,7 @@ public void testJavaAPI18() throws IOException, ClassNotFoundException {
    * daffodil/daffodil-sapi/src/test/resources/test/sapi
    *
    */
-    @Test
+    /*@Test*/
```
stevedlawrence: Instead of commenting these out, can we just use a schema that will throw a Runtime SDE someway other than excessive backtracking?

jadams-tresys: Done

PR #282, opened 2019-10-31
Specify the appropriate shell for installing GitHub Actions dependencies

stevedlawrence:
GitHub Actions seems to have changed the default shell on Windows from
cmd.exe to powershell. We relied on the cmd.exe 'mklink' built-in to add
sbt to the PATH on windows, but that doesn't work in powershell. Ensure
we use cmd.exe for this step.

Also switch to REM for comments in the windows script. YAML does not
treat hashes as comments inside multi-line scalars, so they end being
fed directly to cmd.exe. Change them to cmd.exe style comments.

DAFFODIL-2225

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,16 +42,18 @@ jobs:
           sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
           sudo apt-get -y update
           sudo apt-get -y install rpm sbt
+        shell: bash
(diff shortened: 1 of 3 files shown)


stevedlawrence:
No idea why that one failed, looks like something just got stuck on github's end. The tests never even got triggers. I've triggered a re-run of a test just to double check.

stevedlawrence:
After rebasing onto master, I discovered a couple minor issues that caused failures on windows that were masked because GitHub changed the default behavior and broke our builds. Those issues are fixed as part of this.

olabusayoT:
+1

PR #283, opened 2019-11-05
Adds constructors for TDML Runner

olabusayoT:
-- useful for Java calls
-- fixes typo in UDF tests


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
```diff
@@ -98,6 +98,15 @@ class Runner private (elem: scala.xml.Elem, dir: String, file: String,
...
 
+  /*
+   * these constructors are for use by Java programs
+   */
+  def this(dir: String, file:String) =
+    this(null, dir, file)
+
+  def this(elem: scala.xml.Elem) =
+    this(elem, null, null)
+
```
stevedlawrence: Can you add some TDML unit tests that use these constructors? Also, even though this is a small change, it's probably still a good idea to create a bug and add it to the commit message so there's tracability.

PR #284, opened 2019-11-06
Setup so tutorials are deployed to daffodil-site.

mbeckerle:
The XSLT and other infrastructure are now maintained as part of
daffodil-site.

release-candidate script copies tutorial files from daffodil tree
to daffodil-site tree.

Assumes daffodil-site changes which put tutorial infrastructure and
asciidoc capability into deployed site.

DAFFODIL-2004


tutorials/src/test/scala/org/apache/daffodil/tutorials/TestTutorials.scala
```diff
@@ -26,7 +26,7 @@ object TestTutorials {
   val runner2 = Runner("/", "test1.tutorial.tdml.xml")
   val runner3 = Runner("/", "testTDMLTutorialFile2.tutorial.tdml.xml")
```
stevedlawrence: Maybe only semi related to this change, but no tests are run using the runner3. Should that be delete along with the testTDMLTutorialFile? I'm not sure if that file should has a purpose or if it's just cruft. Also, what is test1.tutorial.tdml? Is that a tutorial or testing the tutorial capabilities? Should it be in src/test/ or maybe it just needs a more descriptive name other than test1?

PR #289, opened 2019-11-12
Daffodil 2192 warnings - nextElementResolver rewrite

mbeckerle:
All tests pass.

All tests pass for VMF, USMTF ATO, and HL7-v2.7
 
Compile times are still slow, but back to where we can tolerate them.


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1695,7 +1695,12 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           case erd: ElementRuntimeData => erd
           case _ => SDE("dfdlx:outputTypeCalcNextSibling can only be defined on an element")
         }
-        val resolver = erd.nextElementResolver
+        val resolver = erd.partialNextElementResolver
+        //
+        // FIXME: This is broken because it is looking outward and upward for enclosing contexts
```
mbeckerle: remove comments. No longer relevant. The partial resolver is not looking beyond the scope of the current lexically enclosing group. 

This could be a problem if there is a group-ref/def boundary between the "siblings" in this typeCalc pair.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/TermRuntime1Mixin.scala
```diff
@@ -0,0 +1,189 @@
...
+    val schemaFileLocation = context.schemaFileLocation
+    val (sibs, isRequiredInInfoset) = possibles match {
+      case Closed(sibs) => (sibs, true)
+      case Open(sibs) => (sibs, false)
+    }
+    val eltMap = sibs.map {
+      sib => (sib.e.namedQName, sib.e.erd)
+    }.toMap.asInstanceOf[Map[QNameBase, ElementRuntimeData]]
+    val resolver = eltMap.size match {
+      case 0 => new NoNextElement(schemaFileLocation)
```
mbeckerle: Do we actually use NoNextElement resolver, or is it redundant with Closed state of possibles (which turns into the flag isRequiredInInfoset being true, and at runtime the nextElement method of NextElementResolver deals with Closed nature.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -0,0 +1,304 @@
...
+import org.apache.daffodil.processors.ElementRuntimeData
+import org.apache.daffodil.util.MStackOf
+import org.apache.daffodil.processors.TermRuntimeData
+import org.apache.daffodil.util.ResettableIterator
+import scala.language.implicitConversions
+
+/**
+ * Mixin for InfosetInputter
+ */
+trait NextElementResolver {
```
mbeckerle: Add 

`self: InfosetInputter => `

to trait decl.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -0,0 +1,298 @@
...
+
+  // TODO: PERFORMANCE: We really should be interning all QNames so that comparison of QNames can be pointer equality
+  // or nearly so. We're going to do tons of lookups in hash tables, which will compute the hash code, find it is equal,
+  // compare the entire namespace string character by character, only to say yes, and yes will be by far the vast
+  // bulk of the lookup results.  Pointer equality would be so much faster....
+  //
+  def maybeNextElement(nqn: NamedQName, hasNamespace: Boolean): Maybe[ElementRuntimeData] =
+    maybeNextElement(nqn.local, if (hasNamespace) nqn.namespace.toStringOrNullIfNoNS else null, hasNamespace)
+
+  def allPossibleNextElements: Seq[ElementRuntimeData]
```
mbeckerle: Maybe rename from "all" possible to remove "all" prefix. It's not all anymore, it's a partial set. It's just within this lexically-enclosing model group.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -380,12 +386,13 @@ class DataProcessor(val ssrd: SchemaSetRuntimeData)
...
     state.setProcessor(rootUnparser)
 
-    /* Verify that all stacks are empty */
+    // Verify that all stacks are empty
     Assert.invariant(state.arrayIndexStack.length == 1)
     Assert.invariant(state.groupIndexStack.length == 1)
     Assert.invariant(state.childIndexStack.length == 1)
     Assert.invariant(state.currentInfosetNodeMaybe.isEmpty)
     Assert.invariant(state.escapeSchemeEVCache.isEmpty)
+
```
mbeckerle: Add check here that the TRD stack is also empty.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -763,12 +769,179 @@ final class ElementRuntimeData(
...
+ * expected, the resulting infoset event will contain a DIElement having an
+ * ErrorElementRuntimeData ERD. The unparser can then inspect the infoset event, and
+ * if it is expecting an EndElement event, it can issue a UnparseError that
+ * correctly identifies that an expected EndElement event was not received.
+ *
+ * In all cases, there is no recovering from these errors in the Unparser; hence,
+ * we don't really care if these bogus DIElement nodes having these Error ERDs are
+ * spliced into the infoset or not.
+ */
+sealed abstract class ErrorERD(local: String, namespaceURI: String)
```
mbeckerle: Is there a better way to do this so that I don't have to construct this fully bogus object?
There's no union types in Scala..... I'd have to make ERD a trait, and then have the real and error subclasses.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -763,12 +769,179 @@ final class ElementRuntimeData(
...
+  /**
+   * Causes unparse error with diagnostic about unexpected element.
+   *
+   * Pass argument true if the context is one where no element was expected
+   * (e.g., because an EndElement was expected.)
+   *
+   * Pass false if the Term's ordinary nextElementResolver list of possibilities
+   * is what was expected.
+   */
+  def toUnparseError(nothingWasExpected: Boolean = false) = {
```
mbeckerle: This is never called. Tests need to be created that get these messages.

But, those will fail because this is never called. Something is missing. 

Also terminology needs to be made the same with the messages in the Element combinator unparsers, and the PartialNextElementResolvers.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -763,12 +769,179 @@ final class ElementRuntimeData(
...
+          sqnx, allPossiblesString)
+    }
+  }
+}
+
+/**
+ * Used when unparsing to indicate that a next element event was detected that is
+ * unexpected.
+ */
+final class NamespaceAmbiguousElementErrorERD(
```
mbeckerle: No test insures this situation is detected, or the string "found mulitple matches" isn't in any test I can find. 

Need test to cover this case. 

Since there are no calls to toUnparseError however, this message can't get issued. 

So something is wrong because this isn't being invoked ever.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -78,6 +79,9 @@ abstract class UState(
   extends ParseOrUnparseState(vbox, diagnosticsArg, dataProcArg, tunable)
   with Cursor[InfosetAccessor] with ThrowsSDE with SavesErrorsAndWarnings {
 
+  def pushTRD(rd: RuntimeData): Unit
+  def popTRD(rd: RuntimeData): RuntimeData
```
mbeckerle: doc  that the rd is passed to popTRD so we verify we got what we expected to pop.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -422,6 +430,15 @@ final class UStateMain private (
     this(inputter, new VariableBox(vmap), diagnosticsArg, dataProcArg,
       dataOutputStream, initialSuspendedExpressions, tunable)
 
+  /**
+   * Peek at the infoset inputter, and get the root element so we don't lose
+   * access to it.
+   */
+  val originalInfosetElement: DIElement = {
```
mbeckerle: This  depends on the fact that initialization when creating a new UState pushes the rootERD and then inspects, which caches the root element, allowing us to get it here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -604,6 +621,23 @@ final class UStateMain private (
...
+    rd match {
+      case trd: TermRuntimeData => inputter.pushTRD(trd)
+      case _ => // ok
+    }
+  }
+  final override def popTRD(rd: RuntimeData) = {
+    rd match {
+      case trd: TermRuntimeData => {
+        val poppedTRD = inputter.popTRD()
+        Assert.invariant(poppedTRD eq rd)
```
mbeckerle: PUt in better assertion that prints out the two trds so that we have a clue if this assertion in fact fails.

daffodil-test/src/test/scala/org/apache/daffodil/section16/array_optional_elem/TestUnparseArrayFixedOptionalElemNew.scala
```diff
@@ -17,36 +17,37 @@
...
-   * Avoid memory leak of adding more and more test suites to static objects as we run more and more test suites.
-   */
-  @AfterClass def tearDown() {
-    runner_fixed.reset
-    runner_imp.reset
-    runner_parsed.reset
-    runner_expr.reset
-  }
-}
+//import org.junit.Test
```
mbeckerle: delete file.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -0,0 +1,115 @@
+package org.apache.daffodil.runtime1
```
mbeckerle: Add banner to this and all other files in this new package.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ParticleMixin.scala
```diff
@@ -85,12 +85,16 @@ trait RequiredOptionalMixin { self: ElementBase =>
...
    * seemingly required elements (based on minOccurs) optional or
    * repeating.
+   *
+   * OutputValueCalc elements are treated as optional. If present they are supposed to get their
+   * values ignored and overwritten by the computation.
    */
   final lazy val isRequiredInInfoset: Boolean = {
     val res = {
-      if (isScalar) true
+      if (isScalar) !this.isOutputValueCalc
```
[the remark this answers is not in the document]
mbeckerle: How about isRequiredStreamingUnparserEvent ?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ParticleMixin.scala
```diff
@@ -85,12 +85,16 @@ trait RequiredOptionalMixin { self: ElementBase =>
...
       else if (isOptional) false
-      else if (isArray) isArraywithAtLeastOneRequiredArrayElement
+      // Treat all arrays as non-required so that we can tolerate 
+      else if (isArray) false // isArraywithAtLeastOneRequiredArrayElement
```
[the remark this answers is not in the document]
mbeckerle: tolerate invalid infosets where the number of events is fewer than the array minimum occurrences. 
I'll fix the comment. 

We have lots of places where true false mean "known true" vs. "not known, could be either". 

In this case false means "is not required" so "may or may not be in the event stream". 

I'll put scaladoc comments to explain this better.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -65,16 +65,18 @@ trait SchemaComponent
 
   lazy val tunable: DaffodilTunables = optLexicalParent.get.tunable
 
-  lazy val dpathCompileInfo: DPathCompileInfo =
+  lazy val dpathCompileInfo: DPathCompileInfo = {
+    lazy val parents = enclosingComponent.map { _.dpathCompileInfo }.toSeq
     new DPathCompileInfo(
-      enclosingComponents.map { _.encloser.dpathCompileInfo },
+      parents,
```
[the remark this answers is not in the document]
mbeckerle: Yes. Coding for debug. I needed a variable so I could look at the value of this.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -81,6 +82,51 @@ trait HasTermCheck {
   }
 }
 
+/**
+ * A set of possibilities for elements that can arrive as events
+ * in a valid infoset.
```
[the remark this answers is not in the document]
mbeckerle: Will rename "Sibling" to "PossibleNextElement", or perhaps PNE because these are case classes and tests are going to pattern match on these quite a bit. 

I can rename isRequiredInInfoset to match isRequiredStreamingUnparserEvent, but perhaps it should be overrideIsRequiredStreamingUnparserEvent.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfoset.scala
```diff
@@ -43,22 +47,58 @@ object TestInfoset {
...
+    // A prior version of this code just pulled events to force the
+    // infoset to be constructed. That doesn't work anymore.
+    // The structure of the schema must be respected and the
+    // TermRuntimeData must be pushed/popped appropriately.
+    //
+    // Really the only way to do this is to invoke the unparser since that is
+    // what maintains the schema dynamic runtime context.
+    //
+    val inputter = new ScalaXMLInfosetInputter(xmlElem)
+    val dummyOutStream = new java.io.FileOutputStream("/dev/null")
```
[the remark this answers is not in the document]
mbeckerle: apache commons io has NullOutputStream. Will use that.

daffodil-lib/src/main/scala/org/apache/daffodil/util/MStack.scala
```diff
@@ -282,9 +284,9 @@ protected abstract class MStack[@specialized T] private[util] (
    * Creates and iterator over the stack in LIFO order.
    *  @return an iterator over the elements of the stack.
    */
-  def iterator: Iterator[T] = new Iterator[T] {
+  def iterator: Iterator[T] = new ResettableIterator[T] {
     private var currentIndex = index
-    private val initialIndex = index
+    private var initialIndex = index
```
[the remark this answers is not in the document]
mbeckerle: Yes. 

index is state of the push/pop stack.

initial/current index are state of the iterator which is for working from current top of stack down the stack. A reset of the iterator resets initialIndex to current push/pop index. 

I'll add comments to this effect.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -791,6 +807,125 @@ trait Term
...
+   * Does not follow backpointers from group defs to group refs.
+   *
+   * It is ok, and has no negative impact on sharing for a Term to reference
+   * its lexically enclosing model group.
+   *
+   * Within the lexically enclosing model group, if it is a sequence, then
+   * this computation involves subsequent siblings, children of those siblings.
+   */
+
+  lazy val possibleNextLexicalSiblingElementsInInfoset: PossibleNextElements = {
```
mbeckerle: At end of review, move this whole algorithm into the runtime1 package. This is very specific to the streaming unparser of runtime1. 

Given that isRequiredInInfoset is renamed to isRequiredSteramingUnpaserEvent, this should be renamed to possibleNextLexicalStreamingUnparserEvent

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1695,9 +1695,15 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           case erd: ElementRuntimeData => erd
           case _ => SDE("dfdlx:outputTypeCalcNextSibling can only be defined on an element")
         }
-        val resolver = erd.nextElementResolver
+        val resolver = erd.partialNextElementResolver
+        val followingERDs = resolver.allPossibleNextElements.dropWhile { _ == erd }
+        //
+        // FIXME: This is broken because it is looking outward and upward for enclosing contexts
+        // all the possible elements. This isn't possible anymore statically, since a group can be
+        // shared. You can do the immediate lexical siblings of an element only. Maybe that's sufficient?
```
mbeckerle: Add tests to cover this case where the sibling isn't properly situated relative to the one invoking this nextSibling function.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -128,20 +133,19 @@ abstract class InfosetInputter
       case e: InvalidInfosetException =>
         UnparseError(One(nodeStack.top.erd.schemaFileLocation), Nope, "Infoset does not start with StartDocument event: " + e.getMessage)
     }
+
+    this.pushTRD(rootElementInfo)
```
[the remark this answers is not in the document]
mbeckerle: Nope. Needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
mbeckerle: Created  https://issues.apache.org/jira/browse/DAFFODIL-2236

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
mbeckerle: There is a test for this. test_nextSibling_04

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/TermRuntime1Mixin.scala
mbeckerle: It is not redundant. NoNextElement resolver is used for a complex element with no element children. It is also used for a sequence that is the model group of a complex type.

mbeckerle:
There will be at least one more commit. Recent changes broke VMF again. The fact that we're not sharing is causing ERD comparisons to fail, because there are multiple instances of the same element that should be considered equivalent even though they get separate objects.

stevedlawrence:
Would it be possible to merge PR #290 and then rebase this on top? It would be easier to review all these commits as one change if those weren't included.

mbeckerle:
I'll rebase this onto 290 once merged to get those deltas out.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -0,0 +1,310 @@
...
+   * no mapping.
+   */
+  final def nextElement(name: String, nameSpace: String, hasNamespace: Boolean): ElementRuntimeData = {
+    val iter = trdStack.iter
+    iter.reset()
+    var firstOne: Boolean = true
+    var maybeERD: Maybe[ElementRuntimeData] = Nope
+    import util.control.Breaks._
+
+    breakable {
```
stevedlawrence: I think I've read break should be avoided because it's really slow, like backed by heavy exceptiosn or something?

I *think* you can just put the second half of the wile loop inside the default case or your trd match and you don't need the break? Might be someother way to adjust it to get rid of the breakable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -0,0 +1,310 @@
...
+class OnlyOnePossibilityForNextElement(
+  trd: TermRuntimeData,
+  val nextERD: ElementRuntimeData,
+  isRequiredStreamingUnparserEvent: Boolean)
+  extends PartialNextElementResolver {
+
+  val nqn = nextERD.namedQName
+
+  override def maybeNextElement(local: String, namespace: String, hasNamespace: Boolean): Maybe[ElementRuntimeData] = {
+    lazy val sqn = StepQName(None, local, NS(namespace))
```
stevedlawrence: Move sqn into if block where it's used? No need to evaluate if it's never needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -763,12 +769,148 @@ final class ElementRuntimeData(
...
+}
+
+/**
+ * Used when unparsing to indicate that multiple elements could be next that
+ * differ only by namespace. This means for some event sources (like JSON) we
+ * don't know which element to create.
+ */
+final class NamespaceAmbiguousElementErrorERD(
+  optTRD: Option[TermRuntimeData],
+  local: String,
```
stevedlawrence: We do actually have a test schema that tests for the warning: combinatiosn_04 in namespaces.tdml. I think the problem though is that there is no way to use a JsonInfosetInputter in the TDML runner to create the infoset. I think we'd need a special unit test that uses JSON. Just pointing out that we at least have a schema that should reproduce this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -78,6 +79,24 @@ abstract class UState(
   extends ParseOrUnparseState(vbox, diagnosticsArg, dataProcArg, tunable)
   with Cursor[InfosetAccessor] with ThrowsSDE with SavesErrorsAndWarnings {
 
+  /**
+   * Push onto the dynamic TRD context stack, so long as the argument is a TermRuntimeData.
+   * Otherwise it is ignored.
```
stevedlawrence: Just curious, what types of things aren't TermRuntimeData's that we are pushing on the stack that hsould be ignored?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -789,6 +811,183 @@ trait Term
...
+      //
+      // In the actual runtime, this information is never on top of the dynamic context
+      // stack. It is always down at least one on the stack. It is only used when the
+      // term itself has not provided a match to the incoming event, and we need to
+      // see if things following the term provide a match.
+      //
+      case stb: SequenceTermBase =>
+        followingLexicalSiblingStreamingUnparserElements
+      case _ =>
+        possibleSelfPlusNextLexicalSiblingStreamingUnparserElements
```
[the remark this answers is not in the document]
mbeckerle: This is the big algorithmic change in the system.

The focus was on the element looking at what is next. The nextElementResolver told you what was next, including if it was OVC or defaulted. That depended on just pulling events weaving the whole infoset together. 

Now, the unparsers do the pulling, and they have to do the pushTRD/popTRD to provide the right dynamic context for the events. The focus is on the element decl unparser which is looking for that element's event or one that can follow it in its lexically containing model-group, or if that's not found, we look on this TRD stack for elements that can come later in the dynamic nest of model groups.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -789,6 +811,183 @@ trait Term
...
+          //
+          // This case only applies to when we are analyzing a sequence, but it is
+          // being considered as contributing possible elements that are after the
+          // end of a Term.
+          //
+          // In this case, we DO recursively walk into the sequence's own children
+          //
+          val subTerms = stb.groupMembers
+          val res =
+            subTerms.map { _.possibleSelfPlusNextLexicalSiblingStreamingUnparserElements }.headOption
```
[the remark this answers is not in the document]
mbeckerle: yup. I think they're going to get computed elsewhere anyway, but we can be lazier here.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -789,6 +811,183 @@ trait Term
...
+          res.getOrElse(Open(Nil))
+        }
+      }
+      val res: PossibleNextElements = (thisItself, followingLexicalSiblingStreamingUnparserElements) match {
+        case (Closed(pnes1), poss2) => Closed(pnes1) // thisItself is closed, you'll never look at the following.
+        case (Open(pnes1), Closed(Nil)) => Open(pnes1) // case of Open(...) followed by end of complex element.
+        case (poss1, Closed(pnes2)) => Closed((poss1.pnes ++ pnes2).distinct)
+        case (poss1, poss2) => Open((poss1.pnes ++ poss2.pnes).distinct)
+      }
```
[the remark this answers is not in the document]
mbeckerle: Just breaking out the first case is acceptable. Will change.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -789,6 +811,183 @@ trait Term
...
+          //
+          val sibPossibles = sibTerms.map { _.possibleSelfPlusNextLexicalSiblingStreamingUnparserElements }
+          //
+          // split the possibles at the first closed set,
+          // including the first closed one, if there are any closed ones
+          //
+          val (opens, closedAndAfter) = sibPossibles.span { _.isInstanceOf[Open] }
+          val optClosed = closedAndAfter.headOption
+          val possibleFromSibs = opens ++ optClosed
+          val hasClosed = optClosed.isDefined
```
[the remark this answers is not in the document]
mbeckerle: I think it's clearer if I just rename optClosed to optFirstClosed, and possibleFromSibs to opensPlusFirstClosed

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/TermRuntime1Mixin.scala
```diff
@@ -47,8 +56,108 @@ import org.apache.daffodil.schema.annotation.props.gen.LengthKind
  */
 trait TermRuntime1Mixin { self: Term =>
 
+  import PossibleNextElements._
```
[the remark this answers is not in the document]
mbeckerle: Agreed.
I can have my cake and eat it to like this:

`private val PNE = PossibleNextElements`

then below I use `PNE.Closed(x) `and `PNE.DoNotUse`, etc.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
```diff
@@ -190,8 +190,14 @@ sealed abstract class ElementUnparserBase(
 
     doBeforeContentUnparser(state)
 
+    if (erd.isComplexType)
+      state.pushTRD(erd.optComplexTypeModelGroupRuntimeData.get)
```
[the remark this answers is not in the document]
mbeckerle: We push model groups. A complexType has a model group. A simple type does not. It has to be a container where there are elements that could need to get resolved.  Will comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -1345,13 +1349,15 @@ sealed trait DIFinalizable {
...
  */
-sealed class DIComplex(override val erd: ElementRuntimeData, val tunable: DaffodilTunables)
+sealed class DIComplex(override val erd: ElementRuntimeData)
   extends DIElement
   with DIComplexSharedImplMixin
   with InfosetComplexElement
   with DIFinalizable // with HasModelGroupMixin
   { diComplex =>
 
+  def tunable: DaffodilTunables = toRootDoc.tunable
```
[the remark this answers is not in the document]
mbeckerle: At the cost of a slot we can make it lazy val. I'd prefer to leave it like this. 

Interestingly, there is a tunables object on the ERD of elements so we can get tunables from there too, but I wonder if the two kinds of tunables are the same? The ones on the ERDs of elements have to be those created at schema compile time. But some tunables are varied at runtime.....

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
mbeckerle: removedd

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: I will change to TRD only. There's one place that needs a downcast but that's better. 
We dont' have a trait/base that is TermUnparser, so the choiceBranchMap maps choiceBranchEvent to Unparser, and an Unparser might not be a term. E.g., unparser for statements aren't terms.
Now, it just so happens that choiceBranchMap only contains Unparsers for Terms, so downcasting there is better, or I could mess with the choiceBranchMap, but I'd rather not get into that.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
mbeckerle: I t tried to make a little JSON test rig. Where you can unparse from XML to data, parse from data to JSON, then unparse from JSON to data, and compare. 

I wrote it, but I couldn't get it to accept the schemas in the daffodil-tests namespaces tests specifically. I keep getting SDEs having to do with include/import, yet the same schema file imports and includes and works fine running under the TDML runner. 

I will push this as a separate PR.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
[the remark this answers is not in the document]
stevedlawrence: Yeah, that should be sufficient.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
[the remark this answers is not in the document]
stevedlawrence: Getting the tunable from the ERD seems like it's maybe a bug. As you point out, the ERD tunables are created at compile time, and tunables can changed for runtime. Wonder if anything is even using the tunable.

I'm okay with leaving it as is. If it shows up in profiling we can fix it.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
```diff
@@ -108,7 +108,7 @@ trait Cursor[AccessorType <: Accessor[AccessorType]] {
    * Like inspect, but only partially populates the accessor with data that can be computed without
    * side effects
    */
-  def inspectPure: Boolean
+  // def inspectPure: Boolean
```
stevedlawrence: Remove comment. Was something fixed so inspectPure is no longer needed?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -400,114 +427,283 @@ abstract class InfosetInputter
...
  *
+ * Unparsers are also responsible for looking at the incoming event, and deciding if we
+ * should first fill in any defaultable elements, or compute any dfdl:outputValueCalc elements.
+ *
+ * This changes the behavior of infoset inputters. You can no longer just pull events and have them
+ * get connected together into an infoset.
+ *
+ * Note that populating an accessor does NOT perform any side-effects on the infoset.
+ * That is left up to the Unparser combinators, largely the Element Combinators.
  */
```
stevedlawrence: This is making me wonder if the infoset event stuff is even really needed anymore? THe unparser a tied much more closely to the infoset inputter now. For example, they are now triggering the creation of new elements, whereas previously this was all handled by the infoset inputter and the unparsers could just assume the infoset nodes were created automatically.

I haven't put much thought into this, so this might be completely wrong, but is there any benefit to removing the infoset event stuff, and just moving more of this logic into  individual unparsers? I feel like the infoset accessor stuff was to support streaming and decouple unparsers from infoset creation, but that seems to not be the case anymore.

So I guess the question is, does it make sense to have Unparsers directly access the InfosetInputter without this indirection from the accessor?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -483,7 +508,7 @@ final class UStateMain private (
   override def advance: Boolean = inputter.advance
   override def advanceAccessor: InfosetAccessor = inputter.advanceAccessor
   override def inspect: Boolean = inputter.inspect
-  override def inspectPure: Boolean = inputter.inspectPure
+  // override def inspectPure: Boolean = inputter.inspectPure
```
stevedlawrence: Remove comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -112,83 +148,61 @@ abstract class InfosetInputter
...
 
   override def toString() = {
     indent + "************* STATE ************************\n" +
-      indent + "nextElementResolver = " + nextElementResolver + "\n" +
-      indent + "nodeStack = " + nodeStack + "\n" +
+      indent + "infoStack = " + infoStack + "\n" +
       indent + "********************************************"
   }
 
-  override def inspectPure: Boolean = {
```
bsloane1650: There is still an InspectPure case in ./daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala. I assume we want to remove that as well?

Also, check comments in ./daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala and ./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala for lingering references.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -400,114 +427,283 @@ abstract class InfosetInputter
...
  *
- * We maintaint the invariant that node==null || node.erd==erd
+ * The Unparsers will examine the incoming DIElement, and associate them parent with child if
+ * they decide to keep them.
  *
+ * Unparsers are also responsible for looking at the incoming event, and deciding if we
+ * should first fill in any defaultable elements, or compute any dfdl:outputValueCalc elements.
+ *
+ * This changes the behavior of infoset inputters. You can no longer just pull events and have them
+ * get connected together into an infoset.
```
bsloane1650: Phrasing. Comments shouldn't describe behavior in terms of previous versions of the code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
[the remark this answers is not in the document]
mbeckerle: I have made a pass at removing tunable from RuntimeData, really DPathCompileInfo.
This is a simple, but big enough change that I want to do it in a separate change. 

It's definitely a bug to store tunables on DPathCompileInfo. We only need one thing from it at schema compile time which is unqualifiedPathStepPolicy.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Cursor.scala
mbeckerle: Yes, inspectPure is no longer needed because pulling events no longer has side-effects of splicing things into the infoset. That's up to the caller of the inspect now (the Unparsers).

mbeckerle: Next commit removes all traces of inspectPure, including comments, etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
mbeckerle: The infoset inputter still feels like a useful abstraction to me. It still creates the infoset elements and resolves the elements. It also solves the problem of inferring array boundaries (start/end) from occurrences of element start/end, which is necessary for XML.  

One point of criticism is that it doesn't make it simpler if a representation has its own explicit arrays, as JSON does. In some sense, InfosetInputters should have startArray() and endArray() methods that are directly called for JSON, and for XML, the inputter synthesizes and calls these event callbacks when it infers the array boundaries.

Instead, we treat all infoset event sources as if they were "as dumb as" XML, and always infer array boundaries. This is ok, as our infoset is XML-like. E.g., has no notion of Array with no elements in it. JSON has empty arrays as first class things, but we can't represent that. 

The accessor is just an internal API thing to avoid allocation of little event tuple objects all the time. You advance/inspect and you get back two things: start/end, and "info" which is arrayERD, or a DIElement.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
mbeckerle: Thanks. I removed all the inspectPure references. Next commit.

PR #290, opened 2019-11-14
Daffodil 2233 runtime1

mbeckerle:
See DAFFODIL-2233

There is no functional change in this PR. Just code reorganization to better separate concerns. 

See also the section "Appendix: Schema Compiler grammar and runtime1 Packages"
in https://github.com/apache/incubator-daffodil-site/pull/12 which is a section that is actually about this change, and not about the other changes discussed in that memo.

This is only part way to moving the runtime1 specific stuff from dsom into a runtime1-specific set of mixins. Nothing is yet done with DPath for example. 

Of course eventually it would be good not to do this with mixins at all, but via a list of backend-specific objects for various backends. 

For now, just separating concerns this way helps. It allows us to be clear that dsom and grammar packages are supposed to be runtime/backend-independent.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -18,10 +18,8 @@
 package org.apache.daffodil.grammar
 
 import java.lang.{ Long => JLong }
-
 import scala.Boolean
 import scala.Long
-
 import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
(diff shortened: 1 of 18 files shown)


stevedlawrence:
Also don't forget to squash and add the bug number to the commit.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -156,6 +156,45 @@ import org.apache.daffodil.xml.QName
...
+import org.apache.daffodil.grammar.primitives.ConvertTextNonNegativeIntegerPrim
+import org.apache.daffodil.grammar.primitives.ConvertZonedUnsignedIntPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextUnsignedShortPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextIntegerPrim
+import org.apache.daffodil.grammar.primitives.ConvertTextDoublePrim
+import org.apache.daffodil.grammar.primitives.LiteralCharacterNilOfSpecifiedLength
+import org.apache.daffodil.grammar.primitives.DelimiterStackCombinatorElement
+import org.apache.daffodil.grammar.primitives.ConvertTextCombinator
+import org.apache.daffodil.runtime1.ElementBaseRuntime1Mixin
+import org.apache.daffodil.grammar.primitives.ConvertZonedNonNegativeIntegerPrim
```
jadams-tresys: Are all these new import statements needed (I'm guessing they may be for #289 ).  Also, just for my own knowledge, I'm assuming we are avoiding doing an `import org.apache.daffodil.grammar.primitives._` in order to make it clear where things are coming from, is that correct?

mbeckerle: I don't fine it helpful that these are enumerated. Our standard is to do them individually, but I am not sure why. In this case, it just makes the code harder to work on. If you import primitives._ then code-completion works and you get the symbols you want without having to remember their specific spellings or search around. 

I don't think the import statements are useful for isolating dependencies or anything like that.  

I will change this to just use a wildcard import for grammar.primitives._

mbeckerle:
I agree that runtime1 is completely arbitrary. Code names more conducive to being useful sub-project names would be better ultimately. I'm thinking what we're calling runtime2, which has this code-gen thing in it, should be runtime-cg, and the original runtime should be runtime-primary or runtime-basic.

PR #291, opened 2019-11-19
Switch to file output stream for large data

jadams-tresys:
Previously, we exclusively were using a ByteArrayOutputStream to handle
buffered output streams, which has a hard limit of 2GB. This commit
allows the output stream to change to a FileOutputStream to a temporary
file once the buffer grows to a specified size, removing the 2GB limit.

DAFFODIL-2194


daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+    _fileOS = new java.io.FileOutputStream(tempFile)
+    this.flush
+    this.writeTo(_fileOS)
+    bytesWritten = count
+    this.reset
+    isFile = true
+  }
+
+  override def write(b: Array[Byte]) = {
+    if ((_fileOS == null) && ((count + b.length) > maxBufferSizeInBytes))
```
stevedlawrence: These ``_fileOS == null`` checks are kindof ugly. Thoughts on making this so this class just extends OutputStream instead of ByteArrayOutputStream, and then make it a wrapper for an output stream, which might change from ByteArrayOutputStream to FileOutputStream? I'm thinking something like:
```scala
  var stream = new ByteArrayOutputStream()

  @inline
  private def writeBytes(...): Unit {
    if (!isFile && needsSwitching) {
      val newStream = new java.io.FileOutputStream(...)
      stream.flush
      stream.asInstanceOf[ByteArrayOutputStream].writeTo(newStream)
      stream = newStream
    }
    bytesWritten += len
  }

  override def write(...) = {
    writeBytes(...)
    stream.write(...)
  }
}
```
Makes it so all the overrirde functions are basically just stream.whatever(), except for write which just calls the switch thing. Another benefit is once the switch happens, the old ByteArrayOutputStream can be garbage collected, whereas before it couldn't.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+  private var _byteArrayOS = this
+  private var _fileOS: java.io.FileOutputStream = null
+
+  var isFile: Boolean = false
+  var tempFile: File = null
+  var bytesWritten = count
+
+  def switchOS() = {
+    Assert.invariant(_fileOS == null)
+    tempFile = File.createTempFile("daffodil-", ".tmp")
```
stevedlawrence: There probably needs to be a tunable that says what directory we create this temp file in. Some applications may only be allow to write to certain directories, which might not include /tmp, so this needs to be configurable. Might also be a good idea to make the prefix and suffix tunable as well.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -765,15 +818,30 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
       // no fragment bytes anywhere - just take the bytes
 
       val nBytes = (bufferNBits / 8).toInt
-      val nBytesPut = directDOS.putBytes(ba, 0, nBytes, finfo)
+      val nBytesPut = {
+        if (this.bufferingJOS.isFile)
+          directDOS.putFile(
```
stevedlawrence: Is a flush here needed to make sure that the data is actually written to the file before we start reading it?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -765,15 +818,30 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
...
+            this.bufferingJOS.tempFile.toPath,
+            this.bufferingJOS.bytesWritten * 8, 4096, finfo)
```
stevedlawrence: Can we just use nBytes instead of bytesWritten?

stevedlawrence: 4096 is hardcoded. Should that be chunkSizeInBytes instead?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -765,15 +818,30 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
...
       // fragment byte on directDOS, fragment byte on bufDOS, or both.
 
       val nFragBits = this.fragmentLastByteLimit
       val byteCount = this.bufferingJOS.getCount()
-      val wholeBytesWritten = directDOS.putBytes(ba, 0, byteCount, finfo)
+      val wholeBytesWritten = {
+        if (this.bufferingJOS.isFile)
+          directDOS.putFile(
+            this.bufferingJOS.tempFile.toPath,
+            this.bufferingJOS.bytesWritten * 8, 4096, finfo)
```
stevedlawrence: Same comment here as above.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -765,15 +818,30 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
...
 
       val nBytes = (bufferNBits / 8).toInt
-      val nBytesPut = directDOS.putBytes(ba, 0, nBytes, finfo)
+      val nBytesPut = {
+        if (this.bufferingJOS.isFile)
+          directDOS.putFile(
+            this.bufferingJOS.tempFile.toPath,
+            this.bufferingJOS.bytesWritten * 8, 4096, finfo)
+        else
+          directDOS.putBytes(ba, 0, nBytes, finfo)
```
stevedlawrence: Where does ba come from? Seems like we should only be gettingg that if this is not a file based output stream?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -119,6 +119,20 @@
             </xs:documentation>
           </xs:annotation>
         </xs:element>
+        <xs:element name="outputStreamChunkSizeInBytes" type="xs:int" default="4096" minOccurs="0">
```
mbeckerle: This is an incredibly small number in this modern age. I would suggest a minimum of 64K.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
```diff
@@ -174,7 +174,7 @@ trait DataOutputStream extends DataStreamCommon
    * It is a usage error if bitLengthFrom1 is not greater than or equal to 1.
    *
    */
-  def putByteArray(ba: Array[Byte], bitLengthFrom1: Int, finfo: FormatInfo): Boolean
+  def putByteArray(ba: Array[Byte], bitLengthFrom1: Int, finfo: FormatInfo, ignoreByteOrder: Boolean = false): Boolean
```
mbeckerle: ignoreByteOrder begs for scaladoc. Otherwise,.. how the heck can that be a sensible thing to do?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -807,7 +875,6 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
       }
     }
   }
-
 }
 
 final class BufferedBlobDataOutputStream private[io] (
```
stevedlawrence: I question if this even needs to exist anymore. Seems to me that BlobDataOutputStream is really just a normal DirectOrBufferedDataOutputStrema except the ByteArrayOrFileOutputStreamWithGetBuf should be initialized with a File and never even create the ByteArrayOutputStream. Maybe we just need a special constructor/function that sets that all up instead of having this special class.

I think that would then get rid of some of the hacks where this overrides bufferingJOS and getJavaOutputStream.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -119,6 +119,20 @@
...
+          <xs:annotation>
+            <xs:documentation>
+              When writing file data to the output stream during unparse, this is the maximum number of bytes to write at a time.
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element name="maxByteArrayOutputStreamBufferSizeInBytes" type="xs:long" default="2097152000" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              When unparsing, this is the maximum size of the buffer that the ByteArrayOutputStream can grow to before switching to a file based output stream.
```
stevedlawrence: Can you wrap these descriptions?

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -657,6 +668,64 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
...
+      val bytesRead = fileStream.read(array, 0, bytesToRead.toInt)
+      if (bytesRead == -1) {
+        fileHasData = false
+      } else {
+        val bitsToPut = Math.min(bytesRead * 8, bitsToRead)
+        val ret = putByteArray(array, bitsToPut.toInt, finfo, ignoreByteOrder=true)
+        if (!ret) {
+          fileStream.close()
+          throw new BlobIOException(
+            "Failed to write BLOB data: %s".format(path.toString))
```
mbeckerle: Is this really about BLOB? I thought this would apply any time the buffer grows too big. You probably want an I/O Exception certainly, but not one specific to BLOB.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+  def switchOS() = {
+    Assert.invariant(_fileOS == null)
+    tempFile = File.createTempFile("daffodil-", ".tmp")
+    _fileOS = new java.io.FileOutputStream(tempFile)
+    this.flush
+    this.writeTo(_fileOS)
+    bytesWritten = count
+    this.reset
+    isFile = true
+  }
```
stevedlawrence: Is it possible that there might be a case where a user does not want to switch to a file output, but instead wants to have an exception thrown? I can maybe imagine performance cirticial systems would want to avoid this and cause a failure so they can adjust their schema. Maybe add a tunable that will throw an exception if the maxBufferSize limit for a single buffer is reached?

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -657,6 +668,64 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
...
+    }
+
+    fileStream.close()
+
+    // calculate the skip bits
+    val nFillBits = remainingBitsToPut
+    if (nFillBits > 0) {
+      val ret = skip(nFillBits, finfo)
+      if (!ret) {
+          throw new BlobIOException(
```
mbeckerle: How can this fail? 

Can this skip call if nFillBits is enough to cause any existing fragment byte to be filled out, resulting in writing 1 more byte to the output, which could fail if the file-system is full? 

Or are we re-establishing the fragment byte here? 

If the latter, then this can't fail for any user-sensible reason, so should  just be a Assert.invariant(ret) to insist the skip call succeeds.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+  maxBufferSizeInBytes: Long = 2000 * (1 << 20))
+  extends java.io.ByteArrayOutputStream {
+  private var _byteArrayOS = this
+  private var _fileOS: java.io.FileOutputStream = null
+
+  var isFile: Boolean = false
+  var tempFile: File = null
+  var bytesWritten = count
+
+  def switchOS() = {
```
mbeckerle: Suggest better name "switchToFileOutputStream"

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+      bytesWritten += 1
+    } else
+      super.write(b)
+  }
+
+  override def close() = if (_fileOS != null) _fileOS.close() else super.close()
+  override def flush() = if (_fileOS != null) _fileOS.flush() else super.flush()
+
+  def getCount() = if (isFile) bytesWritten else count
   def getBuf() = buf
```
mbeckerle: Should getBuf fail if you have switched to a file output stream? Assert.usage(_fileOS eq null, ...)

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -836,8 +903,10 @@ final class BufferedBlobDataOutputStream private[io] (
    }
 
   protected override def deliverContent(directDOS: DirectOrBufferedDataOutputStream, finfo: FormatInfo) = {
+    directDOS.putFile(path, lengthInBits, blobChunkSizeInBytes, finfo)
+  }
 
-    val blobStream =
+    /*val blobStream =
```
mbeckerle: remove commented out code.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
mbeckerle: Magic number 4096 ?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -961,7 +1030,12 @@ object DirectOrBufferedDataOutputStream {
    * Factory for creating new ones/
    * Passing creator as null indicates no other stream created this one.
    */
-  def apply(jos: java.io.OutputStream, creator: DirectOrBufferedDataOutputStream, isLayer: Boolean = false) = {
+  def apply(
+    jos: java.io.OutputStream,
+    creator: DirectOrBufferedDataOutputStream,
+    isLayer: Boolean = false,
+    chunkSizeInBytes: Long = 4096,
```
mbeckerle: Tiny size for modern I/O. 64K minimum is what I'd suggest. 
Shouldn't this default come from the tunable now?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
mbeckerle: I've often wondered if this is tunable at the system level. I.e., can you just setup linux/windows/whatever so that it creates temp files in some location as the default or a per-user setting, like using environment variables or something?

stevedlawrence: Potentially, but I can image use cases were it would be a different directory per-file/job, so a global property might not be sufficient. To accomplish this you could use tunables and change the value before each call to parse.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -657,6 +668,64 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
...
+            "Failed to skip %s bits after BLOB data: %s".format(nFillBits, path.toString))
+      }
+    }
+    bitsWritten / 8
```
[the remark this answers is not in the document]
jadams-tresys: I was trying to mimic how putBytes works, as this is essentially what we are replacing in DirectOrBufferedDOS.deliverContent.  putBytes returns the number of bytesWritten, which is then used to verify that we have written out the correct number of bytes.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,30 +33,79 @@ import org.apache.daffodil.util.Maybe
...
+import java.io.File
 
 /**
  * This simple extension just gives us a public method for access to the underlying byte array.
  * That way we don't have to make a copy just to access the bytes.
  */
-private[io] class ByteArrayOutputStreamWithGetBuf() extends java.io.ByteArrayOutputStream {
+private[io] class ByteArrayOrFileOutputStreamWithGetBuf(
+  chunkSizeInBytes: Long = 4096,
+  maxBufferSizeInBytes: Long = 2000 * (1 << 20))
```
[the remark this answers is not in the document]
jadams-tresys: chunkSizeInBytes can definitely be removed.  The default values are useful for some of the unit test code, but that can be updated to use something reasonable.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: I originally did extend OutputStream, but we were making heavy use of the count and buffer that ByteArrayOutputStream provides, but as long as we enforce the access of these things through the appropriate get() functions we should be able to just use OutputStream.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
jadams-tresys: Meant to change this, but completely forgot before I pushed this up for review.  Will fix.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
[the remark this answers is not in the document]
stevedlawrence: I think the difference here is that putBytes can only put full bytes, whereas putFile can put fragment byes if lengthInBits is not a multiple of 8. If this doesn't return a boolean it should return the exact number of things that were supposed to be put, which in this case is bits.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
stevedlawrence: What we might want to do is make is to the default value of the new tunable is the empty string. And we could interpret emptu string as we should use the java.io.tmpdir system property. That way this directly can be controlled globally by setting the system property and setting the tunable to "", or it could be set on a per java, compiler, etc. basis.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
stevedlawrence: I think 4096 was taken from the blob code I wrote. And I think I used 4096 because thats the size of a page on modern kernels. So i have no idea if it was reasonable. Though 64K feels a little big to me. That's a decent chunk of memory that Java is going to need to allocate. For comparison, the JDK Files.copy (which I think is similar to what we're doing?) uses a array of size BUFFER_SIZE:

https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/java/nio/file/Files.java#L2902

which is initialized to 8192:

https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/java/nio/file/Files.java#L2752

So maybe we should bump it up? Might be interesting to do a quick performance test to see if there's a big differentce bettween 4k, 8k, and 64k.

Though, I guess in theory this chunkSize is only ever going to be used when there are huge files (either BLOBS or large buffers) so maybe an allocation of 64k is going to miniscule compared to reading 2G+ of data?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: I think that this may be addressed by a logging message about switching to file based output, which can suggest changing the schema to avoid this situation.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
mbeckerle: Just sayin',  these sizes, like 4K, are the same buffer sizes that were used when I was in college in 1979 using pdp11 running at 8 Megahertz. 

This is sequential I/O, not random.  No compromises for storing small fragments of data inefficiently are needed. 

With CPUs and I/O systems being nearly 3 decimal orders of magnitude faster, I think megabytes are the new Kilobytes.   

Of course perf testing should be the answer, but I would not bother to performance test yet. There is so much baggage in the unparser currently (due to optimizations that are supposed to happen, but which do not take place) that you'd never see anything pop up. 

See: https://issues.apache.org/jira/browse/DAFFODIL-2238. 

We are executing tons of excess overhead, capturing the content and value length for every element, winding into and out of fillBytes unparsers that do nothing, etc. These things need to get fixed first. Just print out the unparser for a small simple binary schema. It's scary huge.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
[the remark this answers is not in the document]
jadams-tresys: Will change

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,81 +34,82 @@ import org.apache.daffodil.util.Maybe
...
+
+private[io] class ByteArrayOrFileOutputStream(
+  maxBufferSizeInBytes: Long,
+  tempFilePath: Path,
+  useFileOnly: Boolean)
+  extends java.io.OutputStream
+  with Logging {
+  val filePath = {
+    if (tempFilePath.toString == "")
+      Paths.get(System.getProperty("java.io.tmpdir"))
```
stevedlawrence: Thoughts on moving this logic the the Tunable code? Defaulting to java.io.tmpdir sort of feels to me like a tunable behavior and not a behavior global to every byte array/file output stream. TunableGenerator has logic to be able to special case what is generated for certain tunables if that does make sense.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStream.scala
```diff
@@ -96,6 +97,15 @@ trait DataOutputStream extends DataStreamCommon
 
   def maybeAbsBitPos0b: MaybeULong
 
+  /**
+   * These values are used for output streams that could change from
+   * ByteArray's to File based output streams.
+   */
+  def chunkSizeInBytes: Int
+  def maxBufferSizeInBytes: Long
+  def tempFilePath: Path
```
stevedlawrence: Maybe ``tempDirPath`` to make it clear this is a path to a directory that temp files will be created in?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,81 +34,82 @@ import org.apache.daffodil.util.Maybe
...
+    else
+      tempFilePath
+  }
+
+  var stream: java.io.OutputStream = {
+    if (useFileOnly)
+      new java.io.FileOutputStream(File.createTempFile("daffodil-", ".tmp", filePath.toFile))
```
stevedlawrence: What happenes if createTempFile fails? Should we throw the new FileIOException, or is whatever exception it throws reasonable?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,81 +34,82 @@ import org.apache.daffodil.util.Maybe
...
   override def write(b: Array[Byte]) = {
-    if ((_fileOS == null) && ((count + b.length) > maxBufferSizeInBytes))
-      switchOS()
-
-    if (_fileOS != null) {
-      _fileOS.write(b)
-      bytesWritten += b.length
-    } else
-      super.write(b)
+    writeBytes(b.length)
```
stevedlawrence: writeBytes feels like an odd name, considering it doesn't always write bytes. Only writes bytes from the buffer to the file if switching. Not sure of a different name though.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,81 +34,82 @@ import org.apache.daffodil.util.Maybe
...
-private[io] class ByteArrayOrFileOutputStreamWithGetBuf(
-  chunkSizeInBytes: Long = 4096,
-  maxBufferSizeInBytes: Long = 2000 * (1 << 20))
-  extends java.io.ByteArrayOutputStream {
-  private var _byteArrayOS = this
-  private var _fileOS: java.io.FileOutputStream = null
-
-  var isFile: Boolean = false
+private[io] class ByteArrayOutputStreamWithGetBuf() extends java.io.ByteArrayOutputStream {
+  def getCount = count
```
stevedlawrence: I don't think this getCount is used anymore? I think ByteArrayOrFileOutputStream keeps track of the count regardless of if the stream is array based or file based?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -142,12 +144,28 @@ private[io] class ByteArrayOrFileOutputStreamWithGetBuf(
  * is not related to a layer, that means the associated Java OutputStream came
  * from the user and it is the users responsibility to close it. The isLayer
  * provides the flag to know which streams should be closed or not.
+ *
+ * chunkSizeInBytes is used when the buffered output stream is using a file as
+ * its buffer. This is the size of chunks that will be read into memory before
+ * being written out to a file.
```
stevedlawrence: Should be "before being written to the direct output stream". The direct output stream might not also be a file

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -278,42 +296,23 @@ class DirectOrBufferedDataOutputStream private[io] (
...
-      bufferedBlob.setNonZeroLength()
-    bufferedBlob.setFinished(finfo)
-
-    // now split the blob DOS to a normal buffered DOS, we'll return this and
-    // expect the caller to update the UState accordingly so that all future
-    // unparsed data is written to this DOS. This DOS will eventually be
-    // delivered to the blob DOS once it becomes direct
-    val buffered = bufferedBlob.addBuffered()
-    buffered
+    this.putFile(path, lengthInBits, blobChunkSizeInBytes, finfo)
```
stevedlawrence: I'm not sure this is correct. This will cause Daffodil to write the entire contents of the blob to the current data output stream. If the current DOS buffered this will write the full contents to that buffer. Eventually that buffer will fill up and switch to file based and write the entire content to a file, and then we'll finish writing to the blob file to that other file.

So this could be very inefficient if the DOS is buffered.

Instead, I still think we want similar logic as existed before where we split to two DOS, but instead of creating a special BufferedBlobDataOutputStream we want to create a DirectOrBufferedDataOutputstream where the ByteArrayOrFIleOutputStream never uses a ByteArrayOutputStream and goes straight to the FileOutputStream. And the FileOutputStream should be the backed by the blob file instead of a temporary file.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -811,35 +810,39 @@ class DirectOrBufferedDataOutputStream private[io] (
...
-          directDOS.putFile(
-            this.bufferingJOS.tempFile.toPath,
-            this.bufferingJOS.bytesWritten * 8, 4096, finfo)
-        else
-          directDOS.putBytes(ba, 0, nBytes, finfo)
+        if (bufOS.isFile) {
+          bufOS.flush
+          val nBitsPut = directDOS.putFile(
+            bufOS.tempFile.toPath,
+            bufOS.nBytes * 8, chunkSizeInBytes, finfo)
```
stevedlawrence: I'm not sure this is correct for blobs. Although blob files are multiples of bytes, we actually support fragment bytes from the blob file. So I think instead of nBytes it should be bufferNBits. The assert probably also wants to be different since putBytes returns the number of bytes put, but putFile returns the number of bits, which isn't always a multiple of 8.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -811,35 +810,39 @@ class DirectOrBufferedDataOutputStream private[io] (
...
-            this.bufferingJOS.tempFile.toPath,
-            this.bufferingJOS.bytesWritten * 8, 4096, finfo)
-        else
-          directDOS.putBytes(ba, 0, byteCount, finfo)
+        if (bufOS.isFile) {
+          bufOS.flush
+          val nBitsPut = directDOS.putFile(
+            bufOS.tempFile.toPath,
+            bufOS.nBytes * 8, chunkSizeInBytes, finfo)
+          nBitsPut / 8
```
stevedlawrence: Same commnt here as above.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/BlobLengthUnparser.scala
```diff
@@ -75,21 +75,7 @@ abstract class BlobUnparserBase(override val context: ElementRuntimeData)
...
-    // Daffodil to immediately deliver the blob content to the direct output
-    // stream and make the blob data output stream the new direct. Since the
-    // blob data output stream is marked as finished in addBufferedBlob, the
-    // second data output stream will then be delivered, finally making it the
-    // direct stream.
-    val dos = state.dataOutputStream
-    val newDOS = dos.addBufferedBlob(path, lengthInBits, state.tunable.blobChunkSizeInBytes, state)
-    state.dataOutputStream = newDOS
-    dos.setFinished(state)
+    state.dataOutputStream.addBlobFile(path, lengthInBits, state.tunable.blobChunkSizeInBytes, state)
```
stevedlawrence: I think this logic is still needed, for similar reasons as mentione above.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -48,33 +48,31 @@ private[io] class ByteArrayOutputStreamWithGetBuf() extends java.io.ByteArrayOut
...
+  tempDirPath: Path,
+  existingFile: Path)
+  extends java.io.OutputStream with Logging {
 
-  var stream: java.io.OutputStream = {
-    if (useFileOnly)
-      new java.io.FileOutputStream(File.createTempFile("daffodil-", ".tmp", filePath.toFile))
+  val dirPath = {
+    if (tempDirPath.toString == "")
+      Paths.get(System.getProperty("java.io.tmpdir"))
```
stevedlawrence: Thoughts on moving this to the tunable logic? It feels to me like this class should always expect to receive a valid path, and the caller should be reponsible for figuring on that path. In our case, the caller value comes from tunables, and the tunable code generator should contain the logic for "tunable value of emptry string should become java.io.tmpdir".

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -48,33 +48,31 @@ private[io] class ByteArrayOutputStreamWithGetBuf() extends java.io.ByteArrayOut
...
     else
-      new ByteArrayOutputStreamWithGetBuf()
+      tempDirPath
   }
 
-  var isFile: Boolean = useFileOnly
-  var tempFile: File = null
-  var nBytes: Long = 0
+  var isFile: Boolean = (existingFile != null)
```
stevedlawrence: All this null stuff makes me thing existingFile should be a ``Maybe`` so that the logic becomes ``existingFIle.isDefined``. That's more scala-y. And I suspect nothing should really use existingFile. I would expect existingFile would set tempFile (if its defined) and then all the other logic is based on tempFile. So if tempFile is set  then this OutputStream is file-based, otherwise it ByteArray-based.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -152,20 +150,17 @@ private[io] class ByteArrayOrFileOutputStream(
  * maxBufferSizeInByte is the size that the ByteArrayOutputStream will grow to
  * before switching over to a FileOutputStream
  *
- * tempFilePath is the path where temporary files will be created when switching
+ * tempPath is the path where temporary files will be created when switching
```
stevedlawrence: Typo? Should this be ``tempDirPath``?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -152,20 +150,17 @@ private[io] class ByteArrayOrFileOutputStream(
...
  * to a file based buffer
  *
- * useFileOnly indicates that we don't want to create a memory based output
- * buffer, and instead go straight to useing a file based buffer regardless of
- * size.
```
stevedlawrence: Needs a description for ``existingFile``.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -809,11 +829,12 @@ class DirectOrBufferedDataOutputStream private[io] (
...
   protected def deliverContent(directDOS: DirectOrBufferedDataOutputStream, finfo: FormatInfo) = {
-
-    val bufOS = this.bufferingJOS
     val bufferNBits = this.relBitPos0b // don't have to subtract a starting offset. It's always zero in buffered case.
+    val bufOS = this.bufferingJOS
 
-    if (directDOS.isEndOnByteBoundary && this.isEndOnByteBoundary) {
+    if (existingFile != null) {
+      directDOS.putFile(bufOS.tempFile.toPath, bufferNBits.toLong, chunkSizeInBytes, finfo)
+    } else if (directDOS.isEndOnByteBoundary && this.isEndOnByteBoundary) {
```
stevedlawrence: It's not clear to my why we need this ``existingFIle != null`` condition. It seems odd this that branch is only taken if existingFIle is set, but not if we switched from buffering to file based? Should those be handled exactly the same? The below code does check if the bufOS.isFile, can't we use that same logic there?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -822,7 +843,7 @@ class DirectOrBufferedDataOutputStream private[io] (
           bufOS.flush
           val nBitsPut = directDOS.putFile(
             bufOS.tempFile.toPath,
-            bufOS.nBytes * 8, chunkSizeInBytes, finfo)
+            bufferNBits.toLong, chunkSizeInBytes, finfo)
           nBitsPut / 8
         } else
           directDOS.putBytes(bufOS.getBuf, 0, nBytes, finfo)
```
stevedlawrence: Comment on the context below. We call directDOS.putFile, writing the whole contents of the file. But that doesn't take into account the length of the blob, is that correct? I'm wondering if this logic was broken before and just not tested?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
stevedlawrence: For reference, the change here would be done here in TunableGenerator:

https://github.com/apache/incubator-daffodil/blob/master/daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala#L144-L153

You could create a custom TunableBase that sets the scalaConversion and scalaDefinition to values that have the empty string equals tmp dir logic, and then add a special case in the highlighted section in that link to create that.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: My last commit had this mistake, but the original Blob code did not.  I'm making this a little more explicit by having an existingFileLengthInBits value that gets passed in, which would be used for the blob length.

stevedlawrence: I'm wondering if we need that at all. The DOS keeps tracking of the relative bit position of this DOS via ``relBitPos0b``. That is effectively the number of bits that have been written to this DOS. And this value is also set whent he blob DOS is created. Seems like that value can always be used to determine how many bits should be delivered from the underlying output stream, wheter it be file based or byte array based. Maybe that simplifies things a bit?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: This has been made into a FileIOException

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: You are correct. This will be removed.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: I have made the appropriate changes to just use the relBitPos0b instead of trying to keep track of the existing file length via function parameters.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -193,7 +200,7 @@ class PrimitiveTunable(name: String, schemaType: String, schemaDefault: String)
   }
 
   private val scalaDefault = schemaType match {
-    case "xs:string" => "\"" + schemaDefault + "\""
+    case "xs:string" if (name != "tempFilePath") => "\"" + schemaDefault + "\""
     case _ => schemaDefault
```
stevedlawrence: It's a little confusing to have the tempFilePath logic split in two places. What about just having here
```scala
case "xs:string" if (name == "tempFilePath") => "System.getProperty(\"java.io.tempdir\")"
case "xs:string" => "\"" + schemaDefault + "\""
```

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/BlobLengthUnparser.scala
```diff
@@ -75,20 +75,9 @@ abstract class BlobUnparserBase(override val context: ElementRuntimeData)
...
-    // Blob, and contains all the logic for delivering the blob content to the
-    // current data output stream once it becomes direct without loading all
-    // the blob data into memory. The second is a buffered data output stream
-    // that Unparsers following this blob one will write data to. Note that if
-    // the current data stream is direct, the call to setFinished will cause
-    // Daffodil to immediately deliver the blob content to the direct output
-    // stream and make the blob data output stream the new direct. Since the
-    // blob data output stream is marked as finished in addBufferedBlob, the
-    // second data output stream will then be delivered, finally making it the
-    // direct stream.
```
stevedlawrence: This comment is still valid and provides useful information about how blobs are unparsed. It might need to be tweaked a bit since there is no longer a special BLOB DOS, but i think this information is important.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -193,7 +200,7 @@ class PrimitiveTunable(name: String, schemaType: String, schemaDefault: String)
...
   }
```
stevedlawrence: What happens if someone does someting like this:
```scala
dp.setTunable("tempFilePath", "/custom/path")
...
dp.setTuanble("tempFilePath", "")
```
So they effectively want to set it back to the default? Since the ``scalaDefault`` change only affects the initial value of the tunable, the second setTunable call won't revert back to java.io.tmpdir. Which is potentially reasonable, but what will the behavior be? If the Path is the empty, will that fail, will it be the current directory, or something else? Not sure I feel strongly about what happens, but just want to make sure whatever hapens is reasonable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -304,7 +305,14 @@ class DataProcessor(val ssrd: SchemaSetRuntimeData)
...
 
   def unparse(inputter: InfosetInputter, outStream: java.io.OutputStream) = {
-    val out = DirectOrBufferedDataOutputStream(outStream, null) // null means no other stream created this one.
+    val out = DirectOrBufferedDataOutputStream(
+      outStream,
+      null, // null means no other stream created this one.
+      isLayer=false,
+      tunablesObj.outputStreamChunkSizeInBytes,
+      tunablesObj.maxByteArrayOutputStreamBufferSizeInBytes,
+      Paths.get(tunablesObj.tempFilePath))
```
stevedlawrence: Might be advantageous to modify the TunableGenerator so that the tempFilePath variable is of type File so we don't have to duplicate this conversion here and below (and anywhere else it might eventually be used). Also makes it easier to catch any conversion errors in the setTunable logic rather than where it's actuall used.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -225,7 +295,13 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
...
    */
   def addBuffered(): DirectOrBufferedDataOutputStream = {
-    val buffered = new DirectOrBufferedDataOutputStream(this, isLayer)
+    val buffered = new DirectOrBufferedDataOutputStream(
+      this,
+      isLayer,
+      chunkSizeInBytes,
+      maxBufferSizeInBytes,
+      tempDirPath,
+      maybeExistingFile)
```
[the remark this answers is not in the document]
jadams-tresys: You are correct. Will fix

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,32 +34,81 @@ import org.apache.daffodil.util.Maybe
...
+    checkBuffer(1)
+    stream.write(b)
   }
 
-  def hexDump = {
-    (0 to (count - 1)).map { i => "%2x".format(buf(i).toInt & 0xFF) }.mkString(".")
+  def getBuf() = {
+    Assert.usage(!isFile, "Attempted to call getBuf on FileOutputStream")
+    stream.asInstanceOf[ByteArrayOutputStreamWithGetBuf].getBuf
+  }
```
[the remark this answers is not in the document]
jadams-tresys: Agreed.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -33,32 +34,81 @@ import org.apache.daffodil.util.Maybe
...
+  /**
+   * Check to see if there is enough room in the ByteArrayOutputStream for the
+   * specified length. If there is not, switch to FileOutputStream.
+   */
+  @inline
+  private def checkBuffer(lengthInBytes: Long): Unit = {
+    if (!isFile && (nBytes + lengthInBytes > maxBufferSizeInBytes)) {
+      log(LogLevel.Info, "Switching to file based output stream. If this is performance critical, you may want to consider re-organizing your schema to avoid this if possible.")
+      tempFile = try {
+        File.createTempFile("daffodil-", ".tmp", tempDirPath.toFile)
```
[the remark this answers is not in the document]
jadams-tresys: I believe my next commit will address this. Calls to putFile will be inside a try block, with a finally section that deletes the temporary file.  The file will also be set to deleteOnExit in the case of some other unhandled exception.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
jadams-tresys: You're right, will fix

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -843,10 +860,15 @@ class DirectOrBufferedDataOutputStream private[io] (
...
-            bufOS.tempFile.toPath,
-            bufferNBits.toLong, chunkSizeInBytes, finfo)
+          bufOS.close
+          val nBitsPut = try {
+            directDOS.putFile(bufOS.getFile.toPath, bufferNBits.toLong, chunkSizeInBytes, finfo)
+          } finally {
+            // Make sure we delete the file after it was put in the directDOS or
+            // if we encountered an error
+            if (bufOS.isTempFile)
+              bufOS.getFile.delete()
```
stevedlawrence: Whatabout the case where there is an unparse error after switching to a file? In that case, this DOS will never be delivered and it won't be deleted? We need some way to delete all the temp files from all the DOS that haven't been delivered.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -196,16 +189,23 @@ class PrimitiveTunable(name: String, schemaType: String, schemaDefault: String)
...
 
   override val scalaDefinition = s"""val ${name}: ${scalaType} = ${scalaDefault}"""
-  override val scalaConversion = s"""case "${name}" => this.copy(${name} = value.to${scalaType})"""
+  override val scalaConversion = {
+		if (name == "tempFilePath")
+			s"""case "${name}" => this.copy(${name} = new java.io.File(value))"""
+		else
+			s"""case "${name}" => this.copy(${name} = value.to${scalaType})"""
+	}
```
stevedlawrence: This now has a special case for both scalaDefault and scalaConversion, which are the only two things that need to be implemented in TunableBase. Maybe instead just create a new TunableBase unique for tempFilePath that has the appropriate scalaDefault/Conversion and instantiate it in writeGeneratedCode above? Isolates all this special casing to a single class and leves PrimitiveTunable generic.

daffodil-io/src/test/scala/org/apache/daffodil/io/TestDirectOrBufferedDataOutputStream.scala
```diff
@@ -113,7 +113,7 @@ class TestDirectOrBufferedDataOutputStream {
 
   @Test def testCollapsingTwoBuffersIntoDirect {
 
-    val baos = new ByteArrayOrFileOutputStream(2000 * (1 << 20), Paths.get("."), Maybe.Nope)
+    val baos = new ByteArrayOrFileOutputStream(2000 * (1 << 20), new File("."), Maybe.Nope)
     val layered = newDirectOrBufferedDataOutputStream(baos, null)
 
     val hw = "Hello World!"
```
stevedlawrence: Do we need some new TDML or unit tests that validate that these new tunables work and switching to a file works correctly?  I dont think any of these tests have small enough limits to ever switch to a file, and since this is tunable based we can actually have tests for it, unlike the other 2GB changes. A good starting point would be to look at the code coverage and make sure all the new paths are hit.

jadams-tresys: You are correct, I have tested them by manually changing the tunable.  I forgot that with the tunable we can actually set up some TDML test cases.  Will do.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -802,89 +924,29 @@ class DirectOrBufferedDataOutputStream private[io] (var splitFrom: DirectOrBuffe
...
+   * Clean up any temporary files that were generated
+   */
+  def cleanUp(): Unit = {
+    var maybeStream = Maybe(this)
+    while (maybeStream.isDefined) {
+      val stream = maybeStream.get
+      val next = stream._following
+      if (stream.bufferingJOS.isTempFile && stream.bufferingJOS.maybeFile.isDefined)
+        stream.bufferingJOS.maybeFile.get.delete()
+      maybeStream = next
```
stevedlawrence: I'm not sure if this really matters, but I think this will do a double delete.

1. Say we have a direct DOS and we split off to a buffered DOS.
2. Assume the buffered DOS switches from byte-based to file-base.
3. No errors occur during the unparse and we deliver the buffered DOS to the direct DOS.
4. After the content is delievered, we delete the temp file and make the buffered DOS the new direct DOS.
    * Even though this DOS is direct, we still have a pointer to bufferingJOS which is still file-based (though the file has been deleted). This is potentially an existing bug--we probably do not want to hang on to the bufferingJOS once it's been delivered, but that is an already existing bug so probably not worth worrying about for this.
5. We call cleanup(), this DOS bufferingJOS is a temp file and maybeFile.defined, so we will delete the file. But we already deleted the file in step 4, so have a double delete.

Fortunately, I think FIle.delete just returns false when the file doesn't exist, so it might not really be a big deal, but it does sort of feel wrong to try to delete a file that we already deleted. I *think* the fix would be to only do a delete if stream.isBuffering, since a buffering stream that exists here was never delivering and so it was also neve deleted.

I might also recommend refactoring out the while loop and just walk the _following DOS's calling cleanUp, e.g.

```scala
def cleanUp(): Unit = {
  // logic to test if temp and delete if so

  if (_following.isDefined)
    _following.get.cleanUp()
}
```

Avoids the Maybe's and feels more scala-y.

jadams-tresys: Perhaps I should have put in a comment, but I did expect it to do a double delete and was OK with that as File.delete will simply return false if the file does not exist.  It does not throw an error.

Will make the change to check for isBuffering and refactor out the while loop.

PR #292, opened 2019-11-20
Adds serializability checks for UDFs

olabusayoT:
-- adds check to catch any issues with serializability at UDF
initialization time (i.e when the udf object is created)
-- adds associated tests/test classes

DAFFODIL-2235

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
@@ -66,7 +66,7 @@ class TestCLIUdfs {
       shell.expect(
         allOf(
           contains("<fn_func>"),
-          contains("data>strng</data>"),
+          contains("<data>strng</data>"),
           contains("<value>Hello,strng</value>"),
           contains("</fn_func>")))
 
@@ -406,7 +406,7 @@ class TestCLIUdfs {
         1,
         allOf(
           contains("[error] User Defined Function 'ssudf:reverse' Error: UDF Error!"),
-          contains("org.apache.daffodil.dpath.UserDefinedFunctionFatalErrorException: "),
+          contains("org.apache.daffodil.udf.UserDefinedFunctionFatalErrorException: "),
           contains("at org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse.evaluate")))
 
       shell.send("exit\n")
@@ -471,7 +471,9 @@ class TestCLIUdfs {
           contains("[error] Error initializing User Defined Function:"),
           contains("http://example.com/scala/udf:rev-words."),
           contains("Error thrown: org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords$CustomException: UDF Error!"),
-          contains("[error] Schema Definition Error: Unsupported function: ssudf:rev-words")))
+          contains("[error] User Defined Function 'http://example.com/scala/udf:rev-words' Error: UDF Error!"),
+          contains("org.apache.daffodil.udf.UserDefinedFunctionFatalErrorException:"),
+          contains("at org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords")))
 
       shell.send("exit\n")
       shell.expect(eof)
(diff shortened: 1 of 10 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
```diff
@@ -588,4 +591,75 @@ class TestCLIUdfs {
...
+    val metaInfForSomeUdfA = "daffodil-udf/src/test/scala/org/sbadudfs/functionclasses2/StringFunctions/"
+
+    val dafClassPath =
+      (testUdfsPaths :+ Util.daffodilPath(metaInfForSomeUdfA))
+        .mkString(java.io.File.pathSeparator)
+
+    val shell = Util.startIncludeErrors("", envp = Map("DAFFODIL_CLASSPATH" -> dafClassPath))
+
+    try {
+      val cmd = String.format(Util.echoN("strng") + "| %s -v save-parser -s %s -r user_func4", Util.binPath, testSchemaFile)
```
stevedlawrence: The echoN stuff isn't needed to save a parser. Daffodil will just ignore anything from stdin. Should be safe to remove.

daffodil-cli/src/it/scala/org/apache/daffodil/udf/TestCLIUdfs.scala
```diff
@@ -49,6 +50,8 @@ class TestCLIUdfs {
       }.toList.distinct
     commonPrefix
   }
+  lazy val savedParserFile = new File("savedParser.xsd.bin")
```
stevedlawrence: Thoughts on creating this file via Files.createTempFile in the test? That way 1) it doesn't pollute the Daffodil directory so no one accidentally commits it and 2) it ensure there's a unique name so there aren't any issue where a subsequent test uses an old saved parser if somethign weird happens (e.g. save-parser fails, but continues the test with an old saved-parser).

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -226,13 +228,35 @@ object UserDefinedFunctionService extends Logging {
...
+             * This is to protect against any errors thrown when we are trying to
+             * initialize the UDFs. It will catch any exceptions and emit them as the reason the UDF
+             * is unsupported
+             */
               case e: ReflectiveOperationException => {
                 log(LogLevel.Error, "Error initializing User Defined Function: %s. Error thrown: %s",
                   udfid, e.getCause)
                 None
               }
+              case e: Exception => {
```
stevedlawrence: In general we try to avoid catching a general exception, but it doesn't seem unreasonable to have a generic one here. Are there other exceptions you know of that were being hit, or is this just to be safe?

If we do want to keep the generic Exception case, can we remove the ReflectiveOperationException since Exception should handle that? Looks like the only different is ``e.getCause`` vs just ``e``.

olabusayoT: The addition of this generic exception is mainly to be safe as the ReflectiveOperationException catches errors thrown by our default UDFP createUDF function (and as these tend the wrap the underlying issue, we call a getCause on the e). 

We don't catch errors caused by the improper programming of a UDFP implementation such as a scala.MatchError (which returned a huge ugly "This is a bug" message in my case; though on 2nd thought, this might not be fully necessary as the error message does identify the UDF createUDF as the culprit...so I can go either way). Since the exception is not wrapped in anything, we need to pass it directly to log.

If we only want to krrp generic, I could refactor it to check if e is a ReflectiveOperationException within the log call, and call a e.getCause if it is.

mbeckerle: Let's say the serialization code for a UDF throws a divide-by-zero arithmetic exception. 

What do we want to happen? I think we want a Daffodil UDF-oriented exception object with a cause that is the arithmetic exception.  I believe we should not just log this. We need to let the exception propagate. 

The "this is a bug" banner is appropriate for this situation. Some of these "User" defined functions will be ours. Their failure really would be a bug. 

Usually, if you are catching very general things like Exception, you can repackage things into other exception objects, but you should end by re-throwing something.

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -226,12 +228,34 @@ object UserDefinedFunctionService extends Logging {
...
+                val actualCause = e match {
+                  case _: ReflectiveOperationException => e.getCause
+                  case x => x
+                }
                 log(LogLevel.Error, "Error initializing User Defined Function: %s. Error thrown: %s",
-                  udfid, e.getCause)
-                None
+                  udfid, actualCause)
+                throw new UserDefinedFunctionFatalErrorException(s"User Defined Function '$udfid' Error",
```
mbeckerle: Liking this better.

stevedlawrence: Agreed, looks good. Please squash and this can be merged.

PR #293, opened 2019-11-22
WIP: JSON test rig

mbeckerle:
Trying to get it to run this namespaces test on json, but 
current status is that it SDEs with symbol-resolution/namespace/prefix
issues.

Same tests work fine running under the TDML runner, so somehow this
test rig and TDML runner aren't assembling the parts of a schema the
same way?

DAFFODIL-2240


mbeckerle:
Closed as inactive. The JIRA ticket refers back to this PR for future reference.

PR #294, opened 2019-11-22
Removed DaffodilTunables object from the Infoset entirely.

mbeckerle:
They aren't stored on the DIDocument even.
Nor are they available on the RuntimeData/DPathCompileInfo.

They are saved in the state of the DataProcessor object when it is
serialized. They exist nowhere else in the runtime data structures.

They go from there, as modified by setTunable() calls, into
the PState/UState for runtime usage. They, or
some value they contain, must be passed everywhere they are needed
at runtime.

DAFFODIL-2242

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -176,7 +176,7 @@ final class ProcessorFactory(val sset: SchemaSet)
       if (rootElem.numComponents > rootElem.numUniqueComponents)
         log(LogLevel.Info, "Compiler: component counts: unique %s, actual %s.",
           rootElem.numUniqueComponents, rootElem.numComponents)
-      val dataProc = new DataProcessor(ssrd)
+      val dataProc = new DataProcessor(ssrd, tunable)
       if (dataProc.isError) {
         // NO longer printing anything here. Callers must do this.
         //        val diags = dataProc.getDiagnostics
@@ -248,7 +248,7 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
   def setExternalDFDLVariable(variable: Binding) = externalDFDLVariables.enqueue(variable)
   def setExternalDFDLVariables(variables: Seq[Binding]) = variables.foreach(b => setExternalDFDLVariable(b))
   def setExternalDFDLVariables(extVarsFile: File): Unit = {
-    val extVars = ExternalVariablesLoader.getVariables(extVarsFile, tunablesObj)
+    val extVars = ExternalVariablesLoader.getVariables(extVarsFile)
     setExternalDFDLVariables(extVars)
   }
 
@@ -258,10 +258,6 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
 
   def setTunables(tunables: Map[String, String]): Unit = {
     tunablesObj = tunablesObj.setTunables(tunables)
-  }
-
-  def resetTunables(): Unit = {
-    tunablesObj = DaffodilTunables()
   }
 
   /**
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
@@ -31,6 +31,7 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.oolag.OOLAG.OOLAGHost
 import org.apache.daffodil.xml.NamedQName
 import org.apache.daffodil.xml.QNameRegex
+import org.apache.daffodil.BasicComponent
 
 /**
  * Parses DPath expressions. Most real analysis is done later. This is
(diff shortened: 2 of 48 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
```diff
@@ -52,14 +53,14 @@ class DFDLPathExpressionParser[T <: AnyRef](
   namespaces: NamespaceBinding,
   context: DPathCompileInfo,
   isEvaluatedAbove: Boolean,
-  host: OOLAGHost) extends RegexParsers {
+  host: OOLAGHost with HasTunable) extends RegexParsers {
```
mbeckerle: Maybe this OOLAGHost with HasTunable should just become SchemaComponent.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -260,10 +260,6 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
     tunablesObj = tunablesObj.setTunables(tunables)
   }
 
-  def resetTunables(): Unit = {
```
stevedlawrence: Just above this is a call to ExternalVariablesLoader.getVariables() which passes ina  tunablesObj.  Does that need to change too? Seems odd that variable stuff is getting tunable information from the Compiler, and not a DataProcessor, but mabye that happens so early it needs to?

daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
stevedlawrence: Would be nice if we didn't need the ``with HasTunable`` all over the place.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -840,20 +840,9 @@ sealed trait DIElement
...
-    else {
-      parent match {
-        case null =>
-          orig.erd.toss(new InfosetNoRootException(orig, erd))
-        case elt: DIElement => elt.toRootDoc1(orig)
-        case _ =>
-          orig.erd.toss(new InfosetNoRootException(orig, erd))
-      }
-    }
+    else toParent.toRootDoc
```
stevedlawrence: Did you intended to make this a lazy val? I think the other review mentioned keeping a def unless it came up in profiling?

mbeckerle: Force of habit. I can make this a def. But first I'm going to check if it is even used anymore and remove it if it is not.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
mbeckerle: It's all the same tunables "stuff". The compiler constructs the tunablesObj but it can be modified. The DP gets it from the compiler post modifications. it gets serialized and back when the DP is serialized. Then it can be modified again. Finally that gets tucked into the PState/UState for use at runtime.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
mbeckerle: This is used exactly one place in the runtime. When an expression uses an absolute path, then this is invoked to get back up to the root. 

This can be done differently, just walking up to the deepest element on the currentNodeStack in pstate/ustate, so I'm going to suggest we should drop this method entirely.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -162,15 +167,14 @@ class DataProcessor(val ssrd: SchemaSetRuntimeData)
 
   def setTunable(tunable: String, value: String): Unit = tunablesObj = tunablesObj.setTunable(tunable, value)
   def setTunables(tunables: Map[String, String]): Unit = tunablesObj = tunablesObj.setTunables(tunables)
```
stevedlawrence: Right agove this is this:
```scala
  def setExternalVariables(extVars: File, tunable: DaffodilTunables): Unit = {
    ssrd.variables = ExternalVariablesLoader.loadVariables(extVars, ssrd, ssrd.variables, getTunables)
  }
```
That looks like another tunable related bug I think? Tunables are passed into that method, but then they are ignore and getTunables is called to get tunables. Seems odd to ignore the tunables. Not sure where it's actually called.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
stevedlawrence: Ah yeah, this makes perfect sense. If setExternalDFDLVariables is called on the Compiler it makes sense to use the tunables from the compiler. As long as there is no weird behavior if you later chnage the tunable.

mbeckerle: Ok, so the reason that the tunables are passed to the external variables getVariables call, is because the getVariables call has to resolve variable names, and the resolver requires the tunables so that it can have access to unqualifiedPathStepPolicy, which is necessary if what you are resolving is a path step.

For variable resolving, I think we can hard wire this to a particular value (because it doesn't matter), and then the getVariables call won't need a tunable argument.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/DFDLExpressionParser.scala
mbeckerle: Is now BasicComponent defined in runtime1 module.

mbeckerle:
I pushed a change that should fix the scala 2.11 issues. A recursive definition needed a type. It seems scala 2.12 doesn't *always* require types for recursive definitions. It was working fine without it.

PR #295, opened 2019-11-22
Added test to tutorials module that runs a TDML test from java code.

mbeckerle:
This is an example in response to a bug submitted by a user. 

DAFFODIL-2241

### DIFF
--- a/tutorials/src/test/java/org/apache/daffodil/tutorials/TestTDMLFromJava.java
+++ b/tutorials/src/test/java/org/apache/daffodil/tutorials/TestTDMLFromJava.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


tutorials/src/test/java/org/apache/daffodil/tutorials/TestTDMLFromJava.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.daffodil.tutorials;
+
+import org.apache.daffodil.tdml.Runner;
+
+import org.junit.Test;
+
+public class TestTDMLFromJava {
+
+static Runner runner = new Runner("/", "tdmlTutorial.tdml.xml");
```
olabusayoT: The indentation of the contents of the test class is missing

tutorials/src/test/java/org/apache/daffodil/tutorials/TestTDMLFromJava.java
```diff
@@ -0,0 +1,33 @@
...
+
+// DAFFODIL-2241 Java Junit Test TDML - invoke TDML from Java
+@Test
+void test_bugReportParse1() {
+  runner.runOneTest("dateTimeTest");
+}
```
stevedlawrence: Are there other functions we sould also add tests for while were add it? I'm thinking runAllTests and maybe some different runner constructors?

Also, recommend changing tthe name from "bugReportParse1"? "bugReport" is pretty generic and doesn't really describe what this is testing.

PR #296, opened 2019-11-27
Code improvements to restrict scoped property lookup to only terms.

mbeckerle:
A cleanup which prevents a class of errors where you lookup properties not on an object that actually has full visibility to scoping. 

This cleanup is helpful in some of the refactoring for DAFFODIL-1444 (schema compiler space/speed)

DAFFODIL-2244

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/GlobalElementDecl.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/GlobalElementDecl.scala
@@ -28,17 +28,10 @@ final class GlobalElementDecl(
   with GlobalElementComponentMixin
   with ElementDeclFactoryDelegatingMixin
   with NestingTraversesToReferenceMixin
-  with ResolvesProperties // 
-    // Technically, this is not a term, so shouldn't resolve scoped properties
-  // and doesn't take into account combining ref with def properties.
-  // But we need to get one local property: dfdl:choiceBranchKey. In order to 
-  // enforce that it is NOT specified on global defs/decls.
-  // So we mixin ResolvesProperties, but shouldn't use it except to probe dfdl:choiceBranchKey
-  {
-
-  //   global elements combined with element references referring to them can
-  //   be multiple occurring (aka arrays) hence, we have to have things
-  //   that take root and referenced situation into account.
+  // Below needed so we can insure some properties are
+  // NOT on global element decls that are allowed on local element decls.
+  // Such as dfdl:choiceBranchKey
+  with ResolvesLocalProperties {
 
   final override def delegate = factory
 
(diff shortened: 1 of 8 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/GlobalElementDecl.scala
```diff
@@ -28,18 +28,9 @@ final class GlobalElementDecl(
   with GlobalElementComponentMixin
   with ElementDeclFactoryDelegatingMixin
   with NestingTraversesToReferenceMixin
-  with ResolvesProperties // 
-    // Technically, this is not a term, so shouldn't resolve scoped properties
-  // and doesn't take into account combining ref with def properties.
-  // But we need to get one local property: dfdl:choiceBranchKey. In order to 
-  // enforce that it is NOT specified on global defs/decls.
-  // So we mixin ResolvesProperties, but shouldn't use it except to probe dfdl:choiceBranchKey
+  with ResolvesLocalProperties // TODO: what properties do we resolve locally on global elements?
```
stevedlawrence: Did you figure out this TODO?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -260,12 +261,7 @@ abstract class SequenceGroupTermBase(
 trait SequenceDefMixin
   extends AnnotatedSchemaComponent
   with GroupDefLike
-  with ResolvesProperties //
-  // Technically, this is not a term, so shouldn't resolve scoped properties
-  // and doesn't take into account combining ref with def properties.
-  // But we need to get one local property: dfdl:hiddenGroupRef.
-  // So we mixin ResolvesProperties, but shouldn't use it except to obtain dfdl:hiddenGroupRef.
-  {
+  with FindPropertyMixin {
```
stevedlawrence: Should this be ResolvesLocalProperties for dfdl:hiddenGroupRef instead of FindPropertyMixin?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/GlobalElementDecl.scala
mbeckerle: No, I didn't look into it further. I think there are several that we look for a local definition of, if only to give warnings about. 

I'll search and update this.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
mbeckerle: This gets mixed into classes which define ResolvesScopedProperties, or ResolvesLocalProperties so this use of FindPropertyMixin is just to get it to compile without errors. 

Turns out if something has ResolvesScopedProperties mixed in, that doesn't mean all its properties are scoped. It still checks the list of local-only properties like outputValueCalc for elements, and in this case for sequences the dfdl:hiddenGroupRef property, and doesn't "scope" those.  So this does the right thing whether mixed into a class that also mixes in the "scoped" or "local" flavor of property resolving.

PR #299, opened 2019-12-11
Daffodil 2250 cleanup runtime data vs. compile info

mbeckerle:
No functional change here. Just a cleanup. 

Lots of places in the code are using the RuntimeData objects that have no need of anything there beyond just the CompileInfo part of it. This change just decouples that code from anything we change about the RuntimeData objects. 

DAFFODIL-2250

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
@@ -27,15 +27,6 @@ import org.apache.daffodil.compiler.ForParser
 import org.apache.daffodil.processors.unparsers.NadaUnparser
 import org.apache.daffodil.processors.parsers.NadaParser
 
-abstract class UnaryGram(context: Term, rr: => Gram) extends NamedGram(context) {
-  private lazy val r = rr
-
-  final override lazy val gram = {
-    if (r.isEmpty) EmptyGram
-    else this
-  }
-}
-
 /**
  * BinaryGram isn't really 'binary' it's n-ary. It is called binary because it comes from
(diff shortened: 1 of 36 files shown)


mbeckerle:
I force pushed this to remove the two commits now part of master, that previously were overlapping with the DAF-2242 PR.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/Grammar.scala
```diff
@@ -27,14 +27,14 @@ import org.apache.daffodil.compiler.ForParser
...
 
-abstract class UnaryGram(context: Term, rr: => Gram) extends NamedGram(context) {
-  private lazy val r = rr
-
-  final override lazy val gram = {
-    if (r.isEmpty) EmptyGram
-    else this
-  }
-}
+//abstract class UnaryGram(context: Term, rr: => Gram) extends NamedGram(context) {
```
stevedlawrence: Delete comment.

mbeckerle: If this is your only comment I need an approval so I can fix this and push it.

stevedlawrence: I think usually Apache just consideres a +1 comment  approval and doesn't require the actual GitHub "Aprove" thing, especially since many communities dont' even use github. But as a communiy we can decide that we can require the github "approve" thing as well. Might be worth starting a conversastion on dev@ if you'd prefer that. I have no objectiosn to it, and there are some benefits. I'll also marked this as approved.

PR #300, opened 2019-12-12
Use a container for building release candidates

stevedlawrence:
There are too many environment configurations that can lead to difficult
and inconsistent builds. This switches to containers to build and
publish release candidates. Also makes the release candidate script more
mature, including the ability to use --dry-run to test a build without
publishing anything.

The Dockerfile is used to create a new container based on Fedora 30
minimal, install dependencies, set up the environment, and installs the
release candidate script as the entrypoint to the container. This script
has the logic to download all the necessary repos, build everything,
copy build artifacts to the right places, gives the user a list of
things to check, and if everything looks good, the steps to publish the
release candidate.

DAFFODIL-2227


containers/release-candidate/daffodil-release-candidate
```diff
@@ -0,0 +1,255 @@
+#!/bin/bash
```
mbeckerle: This script was release-candidate.sh before. Is there a reason for the redundant long name now? I mean adding daffodil- to everything proved to be a mistake (albeit pretty benign except having to type it alot). Should we cease with that now?

Also, I liked the ".sh" convention.  Admittedly Unix systems have always done the "no extension" thing when things are runnable. But I never really liked that except for commands you run often. This doesn't qualify.

stevedlawrence: The reason for the name change is pretty minor. The instructions for building the container say to use the container name of "daffodil-release-candidate", which is a reasonable and sufficiently unique name to differentiate from other containers on the system.

When you run this container with that name, e.g.

    podman run ... daffodil-release-candidate

any arguments you provide after that are provided to the entrypoint script. So for example you can run this to do a test build:

    podman run ... daffodil-release-candidate --dry-run

And the entrypoint script will change its behavior so it doesn't publish anything. If you give bad arguments or the --help option, then it will print the usage. But the usage displays the name of the script, not the name of the container (the script doesn't even know it's in a container). By giving the entrypoint the same name as the name as the container, the usage just looks more like what someone might expect and almost feels like you're not even running in a container, e.g.:

    podman run ... daffodil-release-candidate --help
    usage: daffodil-release-candidate [OPTION]...
    Options:
      -n, --dry-run   Run all commands but do not actually publish anything
      -h, --help      Display this help and exit

If the script name were something different, it wouldn't match the container name and the usage might be slightly confusing. Granted, you can use whatever name you want for the container, so it won't necessarily match the script name, but that's the container name suggested in the readme so that it does match.

Same reasoning for dropping the .sh. Purely just so the container name matches the script name in the usage.

I don't feel strongly about this if it seems unnecssary.

mbeckerle: I'm fine with this rationale. Leave it as you have it.

PR #12, opened 2017-11-29
Revised daffodil-io module to require passing in a FormatInfo object.

mbeckerle:
Not expect this to get approved without substantial revisions.
This is another large patch set - many files are touched.

This is just one part in addressing this major issue of daffodil
schema compiler slowness/speed.

It should also fix DAFFODIL-1843

The FormatInfo object is used to obtain the very commonly used format
properties like byteOrder, bitOrder, etc. This eliminates the need to
call setters that set these properties, eliminates the need to
save/restore them on backtracking, and when evaluating unparser
suspensions. Instead, the daffodil runtime just passes an object
(the PState/UState) which contains this information - getting it from
the ERD of an element most likely, or if runtime defined, from an
Evaluatable object which caches the property value on the infoset
element.

Once incorporated into daffodil, this will facilitate elimination of
most of the "change" processors (e.g., EncodingChangeParser,
ByteOrderChangeParser) the insertion of which uses compile-time
context-upward analysis.

There is some overhead in this change - performance impact will have to
be measured. It should reduce backtracking overheads to compensate
perhaps.

DAFFODIL-1444, DAFFODIL-1843

PR #301, opened 2019-12-13
DAFFODIL-2169 Add typesafty to our handling of infoset data values.

bsloane1650:
For performance reasons, we have been making extensive use of
AnyRef to store data values in our infoset, and throughout the runtime.
This allows us to avoid boxing, but prevents the type system from validating
that unexepected types to not infect our values.

This commit introduces a value type, DataValue, which provides a 0 overhead (eg. no boxing)
wrapper around AnyRef, while only allowing a limited number of types to populate it.

In doing so, an unrelated bug of mixing Java and Scala BigInts was uncovered and corrected.
Such a bug should be much less likely to occur in the future.

DAFFODIL-2169
DAFFODIL-2188 (unconfirmed)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -34,6 +34,9 @@ import org.apache.daffodil.dpath.NodeInfo.PrimType
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.api.WarnID
 import java.lang.{ Integer => JInt }
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveOrUseNilForDefaultOrNull
+import org.apache.daffodil.infoset.DataValue
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveOrUseNilForDefaultOrNull
 
 /**
  * Note about DSOM design versus say XSOM or Apache XSD library.
@@ -327,11 +330,11 @@ trait ElementBase
    * nor any PUA-remapped characters. This insures the default/fixed value can still be
    * used for ordinary XML-schema validation outside of Daffodil/DFDL.
    */
-  final lazy val defaultValue: Option[AnyRef] = {
+  final lazy val defaultValue: DataValuePrimitiveOrUseNilForDefaultOrNull = {
     if (isDefaultable && (isScalar || isArraywithAtLeastOneRequiredArrayElement)) {
       val dv =
         if (isNillable && useNilForDefault =:= YesNo.Yes) {
-          Some(UseNilForDefault) // singleton object indicator
+          DataValue.UseNilForDefault
         } else {
           //
           // Note: no remapping PUA chars or otherwise messing with the text of the default value
@@ -342,8 +345,8 @@ trait ElementBase
           val value = primType.fromXMLString(str)
           value
         }
-      Some(dv)
-    } else None
+      dv
+    } else DataValue.NoValue
   }
 
   lazy val unparserInfosetElementDefaultingBehavior: UnparserInfo.InfosetEventBehavior = {
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextBooleanUnparser.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextBooleanUnparser.scala
@@ -38,7 +38,7 @@ case class ConvertTextBooleanUnparser(
 
     val node = state.currentInfosetNode.asSimple
 
-    val boolValue = node.dataValue match {
+    val boolValue = node.dataValue.getAnyRef match {
       case JTrue => {
         val textBooleanTrueReps: List[String] = textBooleanTrueRepEv.evaluate(state)
         Assert.invariant(textBooleanTrueReps.length >= 1)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
@@ -17,71 +17,101 @@
 
 package org.apache.daffodil.dpath
 
-import org.apache.daffodil.util.Numbers._
-import org.apache.daffodil.calendar.DFDLDateTime
+import java.lang.{ Boolean => JBoolean }
+import java.lang.{ Byte => JByte }
+import java.lang.{ Double => JDouble }
+import java.lang.{ Float => JFloat }
+import java.lang.{ Integer => JInt }
+import java.lang.{ Long => JLong }
+import java.lang.{ Short => JShort }
+import java.math.{ BigDecimal => JBigDecimal }
+import java.math.{ BigInteger => JBigInt }
+
 import org.apache.daffodil.calendar.DFDLDate
-import java.lang.{ Byte => JByte, Short => JShort, Integer => JInt, Long => JLong, Float => JFloat, Double => JDouble, Boolean => JBoolean }
-import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
+import org.apache.daffodil.calendar.DFDLDateTime
+import org.apache.daffodil.infoset.DataValue.DataValueBigDecimal
+import org.apache.daffodil.infoset.DataValue.DataValueBigInt
+import org.apache.daffodil.infoset.DataValue.DataValueBool
+import org.apache.daffodil.infoset.DataValue.DataValueByte
+import org.apache.daffodil.infoset.DataValue.DataValueDate
+import org.apache.daffodil.infoset.DataValue.DataValueDateTime
+import org.apache.daffodil.infoset.DataValue.DataValueDouble
+import org.apache.daffodil.infoset.DataValue.DataValueFloat
+import org.apache.daffodil.infoset.DataValue.DataValueInt
+import org.apache.daffodil.infoset.DataValue.DataValueLong
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueShort
+import org.apache.daffodil.infoset.DataValue.DataValueString
+import org.apache.daffodil.infoset.DataValue.DataValueTime
+import org.apache.daffodil.util.Numbers.asBigDecimal
+import org.apache.daffodil.util.Numbers.asBigInt
+import org.apache.daffodil.util.Numbers.asBoolean
+import org.apache.daffodil.util.Numbers.asByte
+import org.apache.daffodil.util.Numbers.asDouble
+import org.apache.daffodil.util.Numbers.asFloat
+import org.apache.daffodil.util.Numbers.asInt
+import org.apache.daffodil.util.Numbers.asLong
+import org.apache.daffodil.util.Numbers.asShort
 import org.apache.daffodil.xml.XMLUtils
 
 case object BooleanToLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = JLong.valueOf(if (asBoolean(a) == true) 1L else 0L)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = JLong.valueOf(if (asBoolean(a.getAnyRef) == true) 1L else 0L)
 }
 
 case object BooleanToString extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = if (asBoolean(a) == true) "true" else "false"
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueString = if (asBoolean(a.getAnyRef) == true) "true" else "false"
 }
 
 case object DateTimeToDate extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDate = {
+    a.getAnyRef match {
       case dt: DFDLDateTime => dt.toDate
       case _ => throw new NumberFormatException("xs:dateTime expected but an invalid type was received.")
     }
   }
 }
 case object DateTimeToTime extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueTime = {
+    a.getAnyRef match {
       case dt: DFDLDateTime => dt.toTime
       case _ => throw new NumberFormatException("xs:dateTime expected but an invalid type was received.")
     }
   }
 }
 case object DateToDateTime extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDateTime = {
+    a.getAnyRef match {
       case d: DFDLDate => d.toDateTime
       case _ => throw new NumberFormatException("xs:date expected but an invalid type was received.")
     }
   }
 }
 case object DecimalToInteger extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asBigDecimal(a).toBigInteger()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = asBigDecimal(a.getAnyRef).toBigInteger()
 }
 case object DecimalToLong extends Converter {
   val MAX_VALUE = JBigDecimal.valueOf(Long.MaxValue)
   val MIN_VALUE = JBigDecimal.valueOf(Long.MinValue)
 
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigDecimal(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
+    val res = asBigDecimal(a.getAnyRef)
     if (res.compareTo(MIN_VALUE) == -1 || res.compareTo(MAX_VALUE) == 1) throw new NumberFormatException("Value %s out of range for Long type.".format(res))
     asLong(res)
   }
 }
 case object DecimalToDouble extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asDouble(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = asDouble(a.getAnyRef)
 }
 case object DecimalToNonNegativeInteger extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigDecimal(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = asBigDecimal(a.getAnyRef)
     if (res.compareTo(JBigDecimal.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to a non-negative integer.".format(res))
     res.toBigInteger()
   }
 }
 case object DecimalToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigDecimal(a).toBigInteger()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = asBigDecimal(a.getAnyRef).toBigInteger()
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to a non-negative integer.".format(res))
 
     if (res.compareTo(NodeInfo.UnsignedLong.Max) == 1) throw new NumberFormatException("Value %s out of range for UnsignedLong type.".format(res))
@@ -89,8 +119,8 @@ case object DecimalToUnsignedLong extends Converter {
   }
 }
 case object DecimalToBoolean extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val d = asBigDecimal(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
+    val d = asBigDecimal(a.getAnyRef)
     val comp = d.compareTo(JBigDecimal.ZERO)
     val b =
       if (comp == 0) false
@@ -99,34 +129,34 @@ case object DecimalToBoolean extends Converter {
   }
 }
 case object DoubleToDecimal extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = JBigDecimal.valueOf(asDouble(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal = JBigDecimal.valueOf(asDouble(a.getAnyRef))
 }
 case object DoubleToFloat extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val f = asFloat(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueFloat = {
+    val f = asFloat(a.getAnyRef)
     f
   }
 }
 case object DoubleToLong extends Converter {
   val MAX_VALUE = JBigDecimal.valueOf(Long.MaxValue)
   val MIN_VALUE = JBigDecimal.valueOf(Long.MinValue)
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigDecimal(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
+    val res = asBigDecimal(a.getAnyRef)
     if (res.compareTo(MIN_VALUE) == -1 || res.compareTo(MAX_VALUE) == 1) throw new NumberFormatException("Value %s out of range for Long type.".format(res))
-    asLong(a)
+    asLong(a.getAnyRef)
   }
 }
 case object DoubleToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigInt(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = asBigInt(a.getAnyRef)
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned long.".format(res))
     if (res.compareTo(NodeInfo.UnsignedLong.Max) == 1) throw new NumberFormatException("Value %s out of range for UnsignedLong type.".format(res))
     else res
   }
 }
 case object DoubleToBoolean extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val d = asDouble(a).doubleValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
+    val d = asDouble(a.getAnyRef).doubleValue()
     val b =
       if (d == 0.0) false
       else if (d.isNaN()) false
@@ -135,75 +165,75 @@ case object DoubleToBoolean extends Converter {
   }
 }
 case object FloatToDouble extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asDouble(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = asDouble(a.getAnyRef)
 }
 case object IntegerToDecimal extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = new JBigDecimal(asBigInt(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal = new JBigDecimal(asBigInt(a.getAnyRef))
 }
 case object IntegerToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asBigInt(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = asBigInt(a.getAnyRef)
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned long.".format(res))
     if (res.compareTo(NodeInfo.UnsignedLong.Max) == 1) throw new NumberFormatException("Value %s out of range for UnsignedLong type.".format(res))
     else res
   }
 }
 case object LongToBoolean extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asBoolean(if (asLong(a) == 0) false else true)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = asBoolean(if (asLong(a.getAnyRef) == 0) false else true)
 }
 case object LongToByte extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val l = asLong(a).longValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueByte = {
+    val l = a.getLong
     if (l > Byte.MaxValue || l < Byte.MinValue) throw new NumberFormatException("Value %s out of range for Byte type.".format(l))
-    asByte(a)
+    asByte(l)
   }
 }
 case object LongToDecimal extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = JBigDecimal.valueOf(asLong(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal = JBigDecimal.valueOf(asLong(a.getAnyRef))
 }
 case object LongToDouble extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asDouble(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = asDouble(a.getAnyRef)
 }
 case object LongToFloat extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asFloat(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueFloat = asFloat(a.getAnyRef)
 }
 case object LongToInt extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val l = asLong(a).longValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueInt = {
+    val l = asLong(a.getAnyRef).longValue()
     if (l > Int.MaxValue || l < Int.MinValue) throw new NumberFormatException("Value %s out of range for Int type.".format(l))
-    asInt(a)
+    asInt(a.getAnyRef)
   }
 }
 
 case object LongToInteger extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = JBigInt.valueOf(asLong(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = JBigInt.valueOf(asLong(a.getAnyRef))
 }
 
 case object LongToShort extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val l = asLong(a).longValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueShort = {
+    val l = asLong(a.getAnyRef).longValue()
     if (l > Short.MaxValue || l < Short.MinValue) throw new NumberFormatException("Value %s out of range for Short type.".format(l))
-    asShort(a)
+    asShort(a.getAnyRef)
   }
 }
 
 case object LongToArrayIndex extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asLong(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
+    val res = asLong(a.getAnyRef)
     res
   }
 }
 case object LongToUnsignedByte extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asLong(a).longValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueShort = {
+    val res = asLong(a.getAnyRef).longValue()
     if (res < 0) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned byte.".format(res))
     if (res > 255) throw new NumberFormatException("Value %s out of range for unsigned byte.".format(res))
-    asShort(a)
+    asShort(a.getAnyRef)
   }
 }
 case object LongToUnsignedInt extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val r = asLong(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
+    val r = asLong(a.getAnyRef)
     val res = r.longValue()
     if (res < 0) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned int.".format(res))
     if (res > 0xFFFFFFFFL) throw new NumberFormatException("Value %s out of range for unsigned int.".format(res))
@@ -211,37 +241,37 @@ case object LongToUnsignedInt extends Converter {
   }
 }
 case object LongToUnsignedShort extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = asLong(a).longValue()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueInt = {
+    val res = asLong(a.getAnyRef).longValue()
     if (res < 0) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned short.".format(res))
     if (res > 65535) throw new NumberFormatException("Value %s out of range for unsigned short.".format(res))
-    asInt(a)
+    asInt(a.getAnyRef)
   }
 }
 
 case object LongToNonNegativeInteger extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = JBigInt.valueOf(asLong(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = JBigInt.valueOf(asLong(a.getAnyRef))
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to a non-negative integer.".format(res))
     res
   }
 }
 
 case object LongToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = JBigInt.valueOf(asLong(a))
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = JBigInt.valueOf(asLong(a.getAnyRef))
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to a non-negative integer.".format(res))
     else res
   }
 }
 
 case object NumericToDouble extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asDouble(a)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = asDouble(a.getAnyRef)
 }
 
 case object StringToBoolean extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val str = a.asInstanceOf[String]
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
+    val str = a.getString
     val res =
       if (str == "true" || str == "1") true
       else if (str == "false" || str == "0") false
@@ -250,24 +280,24 @@ case object StringToBoolean extends Converter {
   }
 }
 case object StringToDecimal extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = new JBigDecimal(a.asInstanceOf[String])
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal = new JBigDecimal(a.getString)
 }
 case object StringToDouble extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val str = a.asInstanceOf[String]
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = {
+    val str = a.getString
     val d =
       if (str == XMLUtils.PositiveInfinityString) JDouble.POSITIVE_INFINITY
       else if (str == XMLUtils.NegativeInfinityString) JDouble.NEGATIVE_INFINITY
       else if (str == XMLUtils.NaNString) JDouble.NaN
       else str.toDouble
-    asAnyRef(d)
+    d
   }
 }
 case object StringToLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
     val res =
       try {
-        asAnyRef(a.asInstanceOf[String].toLong)
+        a.getString.toLong
       } catch {
         case nfe: NumberFormatException => {
           val e = new NumberFormatException("Cannot convert to type long: " + nfe.getMessage())
@@ -278,8 +308,8 @@ case object StringToLong extends Converter {
   }
 }
 case object StringToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = new JBigInt(a.asInstanceOf[String])
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res = new JBigInt(a.getString)
     if (res.compareTo(JBigInt.ZERO) == -1) throw new NumberFormatException("Negative value %s cannot be converted to an unsigned long.".format(res))
     if (res.compareTo(NodeInfo.UnsignedLong.Max) == 1) throw new NumberFormatException("Value %s out of range for UnsignedLong type.".format(res))
     else res
@@ -308,8 +338,8 @@ case object StringToUnsignedLong extends Converter {
  * In all other cases, fn:boolean raises a type error [err:FORG0006].
  */
 case object FNToBoolean extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res = a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
+    val res = a.getAnyRef match {
       case b: JBoolean => b.booleanValue()
       case s: String => if (s.length == 0) false else true
       case d: JDouble => if (d.isNaN() || d == 0) false else true
@@ -327,6 +357,6 @@ case object FNToBoolean extends Converter {
       //case s: Sequence if s(0) == Node => true
       case _ => throw new NumberFormatException("Invalid argument type.")
     }
-    asAnyRef(res)
+    res
   }
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
@@ -17,16 +17,17 @@
 
 package org.apache.daffodil.dpath
 
-import com.ibm.icu.text.SimpleDateFormat
-import com.ibm.icu.util.Calendar
-
-import org.apache.daffodil.calendar.DFDLCalendar
 import org.apache.daffodil.calendar.DFDLDate
 import org.apache.daffodil.calendar.DFDLDateConversion
 import org.apache.daffodil.calendar.DFDLDateTime
 import org.apache.daffodil.calendar.DFDLDateTimeConversion
 import org.apache.daffodil.calendar.DFDLTime
 import org.apache.daffodil.calendar.DFDLTimeConversion
+import org.apache.daffodil.infoset.DataValue.DataValueByteArray
+import org.apache.daffodil.infoset.DataValue.DataValueDate
+import org.apache.daffodil.infoset.DataValue.DataValueDateTime
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueTime
 
 case object AnyAtomicToString extends ToString {
   val name = "AnyAtomicToString"
@@ -35,8 +36,8 @@ case object AnyAtomicToString extends ToString {
 case object StringToDate extends Converter {
   val name = "StringToDate"
 
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val result = a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDate = {
+    val result = a.getAnyRef match {
       case cal: DFDLDateTime => cal.toDate
       case cal: DFDLDate => cal
       case str: String => DFDLDateConversion.fromXMLString(str)
@@ -49,8 +50,8 @@ case object StringToDate extends Converter {
 case object StringToDateTime extends Converter {
   val name = "StringToDateTime"
 
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val result = a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDateTime = {
+    val result = a.getAnyRef match {
       case cal: DFDLDateTime => cal
       case cal: DFDLDate => cal.toDateTime
       case str: String => DFDLDateTimeConversion.fromXMLString(str)
@@ -63,8 +64,8 @@ case object StringToDateTime extends Converter {
 case object StringToTime extends Converter {
   val name = "StringToTime"
 
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val result = a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueTime = {
+    val result = a.getAnyRef match {
       case cal: DFDLDateTime => cal.toTime
       case cal: DFDLTime => cal
       case str: String => DFDLTimeConversion.fromXMLString(str)
@@ -77,8 +78,8 @@ case object StringToTime extends Converter {
 case object StringToHexBinary extends Converter with HexBinaryKind {
   val name = "StringToHexBinary"
 
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val result = a match {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueByteArray = {
+    val result = a.getAnyRef match {
       case s: String => hexStringToByteArray(s)
       case hb: Array[Byte] => hb
       case x => throw new NumberFormatException("%s cannot be cast to dfdl:hexBinary\ndfdl:hexBinary received an unrecognized type! Must be String or HexBinary.".format(x.toString))
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
@@ -17,24 +17,30 @@
 
 package org.apache.daffodil.dpath
 
-import org.apache.daffodil.util.Numbers._
-import org.apache.daffodil.util.Misc
 import java.math.{ BigInteger => JBigInt }
+
+import org.apache.daffodil.infoset.DataValue.DataValueBigInt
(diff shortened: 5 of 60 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
```diff
@@ -23,74 +23,158 @@ import org.apache.daffodil.calendar.DFDLDate
 import java.lang.{ Byte => JByte, Short => JShort, Integer => JInt, Long => JLong, Float => JFloat, Double => JDouble, Boolean => JBoolean }
 import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
 import org.apache.daffodil.xml.XMLUtils
+import org.apache.daffodil.infoset.DataValue.DataValueLong
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
```diff
@@ -23,74 +23,158 @@ import org.apache.daffodil.calendar.DFDLDate
...
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueDate
+import org.apache.daffodil.infoset.DataValue.DataValueDouble
+import org.apache.daffodil.infoset.DataValue.DataValueDouble
 
 case object BooleanToLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = JLong.valueOf(if (asBoolean(a) == true) 1L else 0L)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = JLong.valueOf(if (asBoolean(a) == true) 1L else 0L)
```
bsloane1650: This shouldn't work...

We should need asBoolean(a.getAnyRef)

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
```diff
@@ -23,74 +23,158 @@ import org.apache.daffodil.calendar.DFDLDate
...
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDateTime = {
+    a.getAnyRef match {
       case d: DFDLDate => d.toDateTime
       case _ => throw new NumberFormatException("xs:date expected but an invalid type was received.")
     }
   }
 }
 case object DecimalToInteger extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = asBigDecimal(a).toBigInteger()
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = asBigDecimal(a.getAnyRef).toBigInteger()
```
bsloane1650: All of this asType(x.getAnyRef) stuff could be cleaned up as well.

Because of how our libraries are structured, the asType functions do not have access to the DataValue type.

We could probably move the functions (or wrappers for them) to the DataValue class, so we have asFoo() methods which do intellegent coercien to complement our getFoo() methods which just assume the underlying value is already the correct type.

I'm not convinced this is worth the effort at this point.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps2.scala
```diff
@@ -27,6 +27,14 @@ import org.apache.daffodil.calendar.DFDLDateTime
...
 import org.apache.daffodil.calendar.DFDLTime
 import org.apache.daffodil.calendar.DFDLTimeConversion
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueDate
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueDateTime
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueTime
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
+import org.apache.daffodil.infoset.DataValue.DataValueByteArray
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
```diff
@@ -20,37 +20,54 @@ package org.apache.daffodil.dpath
 import org.apache.daffodil.util.Numbers._
 import org.apache.daffodil.util.Misc
 import java.math.{ BigInteger => JBigInt }
+import org.apache.daffodil.infoset.DataValue.DataValueInt
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLConstructors.scala
```diff
@@ -20,6 +20,20 @@ package org.apache.daffodil.dpath
 import org.apache.daffodil.util.Numbers._
 import java.lang.{ Byte => JByte, Short => JShort, Integer => JInt, Long => JLong }
 import java.math.{ BigInteger => JBigInt, BigDecimal => JBigDecimal }
+import org.apache.daffodil.infoset.DataValue.DataValueLong
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions.scala
```diff
@@ -23,6 +23,14 @@ import org.apache.daffodil.cookers.EntityReplacer
 import org.apache.daffodil.dsom._
 import org.apache.daffodil.exceptions._
 import org.apache.daffodil.util.Numbers._
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions2.scala
```diff
@@ -22,6 +22,9 @@ import org.apache.daffodil.infoset.LengthState
 import java.lang.{ Long => JLong }
 import passera.unsigned.ULong
 import org.apache.daffodil.infoset._
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -29,6 +29,9 @@ import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.processors.parsers.ParseError
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXTypeCalcFunctions.scala
```diff
@@ -26,15 +26,17 @@ import org.apache.daffodil.dpath.NodeInfo.PrimType.IntegerKind
 import org.apache.daffodil.dpath.NodeInfo.PrimType.String
 import org.apache.daffodil.dpath.NodeInfo.PrimType.Integer
 import org.apache.daffodil.processors.TypeCalculator
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXTypeCalcFunctions.scala
```diff
@@ -26,15 +26,17 @@ import org.apache.daffodil.dpath.NodeInfo.PrimType.IntegerKind
...
   extends FNTwoArgs(typedRecipes.map(_._1)) {
 
   val srcType = typedRecipes(1)._2
 
-  override def computeValue(arg1: AnyRef, arg2: AnyRef, dstate: DState): AnyRef = {
-    calc.inputTypeCalcRun(dstate, arg2, srcType)
-    dstate.currentValue
+  override def computeValue(arg1: DataValuePrimitive, arg2: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+    calc.inputTypeCalcRun(dstate, arg2.getAnyRef, srcType)
+    dstate.currentValue.getNonNullable
```
bsloane1650: I don't think this will actually fail if currentValue is NoValue.

Not that we were checking for that case before this refactor, but maybe an assertion here would be warrented?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXTypeCalcFunctions.scala
```diff
@@ -69,7 +71,7 @@ case class DFDLXOutputTypeCalcNextSibling(a: CompiledDPath, b: NodeInfo.Kind) ex
     val primType = nextSibling.erd.optPrimType.get
 
     val x = nextSibling.dataValue
-    typeCalculator.outputTypeCalcRun(dstate, x, primType)
+    typeCalculator.outputTypeCalcRun(dstate, x.getAnyRef, primType)
```
bsloane1650: As with the expressions, it seemed too complicated to make the typeCalculator aware of DataValue, so this is one of those boundaries where we cast to AnyRef

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPath.scala
```diff
@@ -34,6 +34,14 @@ import org.apache.daffodil.api.Diagnostic
 import org.apache.daffodil.util.Numbers._
 import org.apache.daffodil.processors.parsers.DoSDEMixin
 import org.apache.daffodil.processors.parsers.PState
+import org.apache.daffodil.infoset.DataValue.DataValueNonEmpty
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPath.scala
```diff
@@ -159,12 +167,12 @@ final class RuntimeExpressionDPath[T <: AnyRef](qn: NamedQName, tt: NodeInfo.Kin
    * so as to detect deadlocks.
    */
   def evaluateForwardReferencing(state: ParseOrUnparseState, whereBlockedInfo: Suspension): Maybe[T] = {
-    var value: Maybe[AnyRef] = Nope
+    var value:Maybe[T] = Nope
```
bsloane1650: This type of change occurs in several places. Due to type erasure, it does not actually do much of anything.

I assume the previous use of Maybe[AnyRef] was just an oversight.

I was not systemetic in looking for these. Where you see me fix this, what probably happened was I accidently changed the type to DataValue, then set it back to Maybe[T] when I realized my mistake. Fixing it is not acutally relevent to what I was trying to do, but I see no reason to go out of my way to avoid fixing it in the places where I did.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPath.scala
```diff
@@ -195,26 +203,26 @@ final class RuntimeExpressionDPath[T <: AnyRef](qn: NamedQName, tt: NodeInfo.Kin
     recipe.run(dstate)
   }
 
-  private def processForwardExpressionResults(dstate: DState): AnyRef = {
-    val v: AnyRef = {
+  private def processForwardExpressionResults(dstate: DState): Maybe[T] = {
```
bsloane1650: Not sure why this changed. Looking at the code again, just AnyRef should still be fine

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPath.scala
```diff
@@ -256,22 +267,22 @@ final class RuntimeExpressionDPath[T <: AnyRef](qn: NamedQName, tt: NodeInfo.Kin
   def isConstant = false
   def constant: T = Assert.usageError("Not a constant.")
 
-  private def processExpressionResults(dstate: DState): AnyRef = {
+  private def processExpressionResults(dstate: DState): Maybe[T] = {
```
bsloane1650: Again, the Maybe here seems uneeded.

AnyRef -> T still makes sense

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
```diff
@@ -38,6 +38,13 @@ import org.apache.daffodil.dpath.NodeInfo.PrimType
 import org.apache.daffodil.dsom.DPathCompileInfo
 import org.apache.daffodil.xml.GlobalQName
 import org.apache.daffodil.processors.TypeCalculator
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
```
bsloane1650: Clean Impors

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DState.scala
```diff
@@ -29,6 +29,13 @@ import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.dsom.DPathCompileInfo
 import org.apache.daffodil.xml.GlobalQName
 import org.apache.daffodil.processors.TypeCalculatorCompiler.TypeCalcMap
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNBases.scala
```diff
@@ -24,17 +24,30 @@ import org.apache.daffodil.exceptions.Assert
 import java.lang.{ Integer => JInt, Long => JLong, Double => JDouble, Boolean => JBoolean }
 import java.math.{ BigInteger => JBigInt, BigDecimal => JBigDecimal }
 import org.apache.daffodil.util.Numbers
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean Imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNBases.scala
```diff
@@ -92,10 +105,10 @@ case class BooleanOp(op: String, left: CompiledDPath, right: CompiledDPath)
 case class NegateOp(recipe: CompiledDPath) extends RecipeOpWithSubRecipes(recipe) {
   override def run(dstate: DState) {
     recipe.run(dstate)
-    val value: AnyRef = dstate.currentValue match {
-      case i: JInt => Numbers.asAnyRef(i * -1)
-      case l: JLong => Numbers.asAnyRef(l * (-1L))
-      case d: JDouble => Numbers.asAnyRef(d * -1.0)
+    val value: DataValuePrimitive = dstate.currentValue.getAnyRef match {
+      case i: JInt => DataValue.unsafeFromAnyRef(Numbers.asAnyRef(i * -1))
```
bsloane1650: I don't think we actually need to use AnyRef here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -35,15 +35,63 @@ import org.apache.daffodil.calendar.DFDLTime
 import org.apache.daffodil.calendar.DFDLDate
 import java.lang.{ Number => JNumber, Byte => JByte, Short => JShort, Integer => JInt, Long => JLong, Float => JFloat, Double => JDouble, Boolean => JBoolean }
 import java.math.{ BigDecimal => JBigDecimal, BigInteger => JBigInt }
+import org.apache.daffodil.infoset.DataValue.DataValueBool
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -40,6 +40,20 @@ import org.apache.daffodil.xml.NoNamespace
 import org.apache.daffodil.xml.QName
 import org.apache.daffodil.xml.RefQName
 import org.apache.daffodil.xml.XMLUtils
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/SuspendableExpression.scala
```diff
@@ -24,6 +24,9 @@ import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe._
 import org.apache.daffodil.processors.Suspension
 import org.apache.daffodil.util.LogLevel
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/UserDefinedFunctionBase.scala
```diff
@@ -38,11 +40,11 @@ case class UserDefinedFunctionCall(
...
   extends FNArgsList(recipes) {
 
-  override def computeValue(values: List[Any], dstate: DState) = {
-    val jValues = values.map { _.asInstanceOf[Object] }
+  override def computeValue(values: List[DataValuePrimitive], dstate: DState) = {
+    val jValues = values.map { _.getAnyRef.asInstanceOf[Object] }
     try {
       val res = evaluateFxn.method.invoke(userDefinedFunction, jValues: _*)
-      res
+      DataValue.unsafeFromAnyRef(res)
```
bsloane1650: I assuming we are actually checking the type before the value reaches here...

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSConstructors.scala
```diff
@@ -20,22 +20,28 @@ package org.apache.daffodil.dpath
 import org.apache.daffodil.util.Numbers._
 import org.apache.daffodil.calendar.DFDLDate
 import org.apache.daffodil.calendar.DFDLTime
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean Imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -41,6 +41,9 @@ import org.apache.daffodil.xml.NS
 import org.apache.daffodil.xml.NamedQName
 import org.apache.daffodil.xml.NoNamespace
 import org.apache.daffodil.xml.StepQName
+import org.apache.daffodil.infoset.DataValue
+import org.apache.daffodil.infoset.DataValue.DataValueAny
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -72,6 +75,8 @@ abstract class CompiledExpression[+T <: AnyRef](
   valueForDebugPrinting: AnyRef)
   extends ContentValueReferencedElementInfoMixin with Serializable {
 
+  DataValue.assertValueIsNotDataValue(valueForDebugPrinting)
```
bsloane1650: This was very useful for doing the refactor. Now that it is done, I am not sure how much this assertion is still needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
```diff
@@ -22,6 +22,9 @@ import org.apache.daffodil.xml.NS
 import org.apache.daffodil.util.MaybeBoolean
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.processors.ElementRuntimeData
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -54,6 +54,14 @@ import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.api.DaffodilTunables
 import java.util.HashMap
 import org.apache.daffodil.dsom.DPathElementCompileInfo
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/Evaluatable.scala
```diff
@@ -440,6 +441,7 @@ trait ExprEvalMixin[T <: AnyRef]
           }
         }
       }
+    DataValue.assertValueIsNotDataValue(expressionResult)
```
bsloane1650: This assertions may also not be needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -60,6 +60,9 @@ import org.apache.daffodil.xml.NamedQName
 import org.apache.daffodil.processors.unparsers.UnparseError
 import org.apache.daffodil.util.Misc
 import org.apache.daffodil.api.UnqualifiedPathStepPolicy
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -33,6 +33,9 @@ import org.apache.daffodil.util.PreSerialization
 import org.apache.daffodil.util.RangeBound
 import org.apache.daffodil.util.TransientParam
 import org.apache.daffodil.xml.QNameBase
+import org.apache.daffodil.infoset.DataValue
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -28,6 +28,13 @@ import org.apache.daffodil.util.Maybe.Nope
 import org.apache.daffodil.util.Maybe.One
 import org.apache.daffodil.xml.GlobalQName
 import org.apache.daffodil.xml.NamedQName
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -45,6 +45,11 @@ import org.apache.daffodil.infoset.DIDocument
 import org.apache.daffodil.infoset.DIComplex
 import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.processors.Processor
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
```
bsloane1650: Clean imports

daffodil-test/.settings/org.scala-ide.sdt.core.prefs
```diff
@@ -1,92 +1,7 @@
-eclipse.preferences.version=1
```
bsloane1650: Revert

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
```diff
@@ -20,37 +20,54 @@ package org.apache.daffodil.dpath
...
 case object HexStringToUnsignedLong extends Converter {
-  override def computeValue(a: AnyRef, dstate: DState): AnyRef = {
-    val res =
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
+    val res: DataValueBigInt =
       try {
-        val str = a.asInstanceOf[String]
-        BigInt(str, 16)
+        val str = a.getString
+        new JBigInt(str, 16)
```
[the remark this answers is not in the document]
mbeckerle: So do we have a test that verifies that enums using the typecalc stuff all work properly now?
That's the *real* reason this was revisited.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -1252,15 +1260,15 @@ sealed class DISimple(override val erd: ElementRuntimeData)
           else if (f == Float.NegativeInfinity) XMLUtils.NegativeInfinityString
           else f.toString
         }
-        case _ => dataValue.toString
+        case x => x.toString
```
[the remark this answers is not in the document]
mbeckerle: Should we make dataValue.toString cause an Assert.usageError? Or is messing with toString a super bad idea (e.g., blows up the debugger???)

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+ * with more ability to control the inheritense relationship between various DataValue types.
+ * In theory, this would be accomplished by establishing a type hierarchy on the types
+ * used for T. Indeed, if Scala had type-classes as a first-class feature that is likely
+ * the approach we would have taken. However, since most of the types we use for T
+ * are out of our control, we are unable to control what type structure they have.
+ * By carefully selecting X in our type aliases, it is possible for to define types
+ * that can represent any subset of types which we want. It is further possible to
+ * allow for automatic upcasting between these types where appropriate.
+ *
+ * We inline all methods because attempts to call non-inlined methods trigger boxing
```
mbeckerle: I don't think that's right. Did you disassemble the byte code? I think polymorphic usage of AnyVal types triggers boxing, but just calling a method on one doesn't. It's the equivalent of calling a Java static method.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+  @inline implicit def toDataValue(v: JLong): DataValueLong = new DataValue(v)
+  @inline implicit def toDataValue(v: JDouble): DataValueDouble = new DataValue(v)
+  @inline implicit def toDataValue(v: JBigInt): DataValueBigInt = new DataValue(v)
+  @inline implicit def toDataValue(v: JString): DataValueString = new DataValue(v)
+  @inline implicit def toDataValue(v: URI): DataValueURI = new DataValue(v)
+  @inline implicit def toDataValue(v: JFloat): DataValueFloat = new DataValue(v)
+  @inline implicit def toDataValue(v: JByte): DataValueByte = new DataValue(v)
+  @inline implicit def toDataValue(v: JInt): DataValueInt = new DataValue(v)
+  @inline implicit def toDataValue(v: JShort): DataValueShort = new DataValue(v)
+  @inline implicit def toDataValue(v: DINode): DataValueDINode = new DataValue(v)
```
mbeckerle: Wait what? We shouldn't need this. v: DINode  what could this possibly mean.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+  @inline def unsafeFromMaybeAnyRef(v: Maybe[AnyRef]) = {
+    if (v.isDefined) {
+      new DataValue(v.get)
+    } else {
+      NoValue
+    }
+  }
+
+  val NoValue: DataValueEmpty = new DataValue(null)
+  val UseNilForDefault: DataValueUseNilForDefault = new DataValue(new UseNilForDefaultObj)
```
mbeckerle: scala doc on this sentinel value would be useful.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+    /*
+   *
+   * In our CompileExpressions classes, we use type variables declared as T <: AnyRef
+   *
+   * Ideally, we would have declared T <: DataValuePrimitive
+   * However, we need to be able to refer to something of the form Maybe[T].
+   * In theory, it should be possible to take a T <: DataValuePrimitive,
+   * and construct a T' <: DataValuePrimitiveNullable, such that T <: T'
+   * In practice, it does not appear to be easy to tell Scala's type system what we want to do,
+   * so we instead punt on this issue and require the caller to translate to/from AnyRef at the boundery.
```
mbeckerle: typo: boundery => boundary

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+   *
+   * In theory, if a caller forgets to do so, and instead passes in a DataValue instead of AnyRef,
+   * the compiler would issue a type error because DataValue is an AnyVal type, and so does not inherit from AnyRef.
+   *
+   * In practice, Scala will "helpfully" box the DataValue for us, which causes all sorts of problems. For instance,
+   * x.asInstanceOf[String] does not work when x is a DataValueString (even though an unboxed DataValueString is literally just a String at runtime).
```
mbeckerle: Can we create an implicit conversion from DataValue to AnyRef that just fails? At runtime anyway we should be able to. Getting it to fail to compile would be great, but I think this full-service boxing is consider a feature.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  @inline def unsafeFromAnyRef(v: AnyRef) = new DataValue(v)
+  @inline def unsafeFromMaybeAnyRef(v: Maybe[AnyRef]) = {
+    if (v.isDefined) {
+      new DataValue(v.get)
+    } else {
+      NoValue
+    }
+  }
+
```
mbeckerle: How about 

implicit def toAnyRef(dv: DataValue): AnyRef = Assert.usageError("Please don't auto box like this scala. Please please!")

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
```diff
@@ -80,13 +83,13 @@ trait InfosetComplexElement extends InfosetElement {
 
 trait InfosetSimpleElement extends InfosetElement {
 
-  def dataValue: Any
+  def dataValue: DataValuePrimitiveNullable
```
mbeckerle: So is this whole use of the phantom types just so we can deal with the stupid sentinel value that indicates useNilForDefault?

Blech. I suggest some names then like adding "Hack" as a suffix as in DataValuePrimitiveNilDefaultHack instead of the seemingly purposeful DataValuePrimitiveNullable. 

Also, terminology suggestion. XML Schema uses the terms nil and nillable. Java and Scala use the term Null as in null pointer. We should be consistent with this. When we're talking about DFDL's notion of Nil we should use Nillable in the identifiers and type names. 

So I think i'm suggesting Nullable => Nillable throughout here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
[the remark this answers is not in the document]
mbeckerle: if I call toString on a scala Int, you're saying they box the Int then, and then invoke toString on it?  Or do they treat built in types differently from user defined AnyVal types?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/Evaluatable.scala
mbeckerle: Given the potential for auto-boxing if you accidently pass a DataValue where a underlying value is wanted, I think the check is worthwhile.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+ * the approach we would have taken. However, since most of the types we use for T
+ * are out of our control, we are unable to control what type structure they have.
+ * By carefully selecting X in our type aliases, it is possible for to define types
+ * that can represent any subset of types which we want. It is further possible to
+ * allow for automatic upcasting between these types where appropriate.
+ *
+ * We inline all methods because attempts to call non-inlined methods trigger boxing
+ */
+final class DataValue[+T <: AnyRef, +X <: AnyRef] private (val v: T) extends AnyVal {
+  @inline def isEmpty = DataValue.NoValue.v eq v
```
mbeckerle: Why isn't isEmpty just (this == DataValue.NoValue)

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+  @inline implicit def toDataValue(v: DINode): DataValueDINode = new DataValue(v)
+
+  @inline implicit def toDataValue(v: Long): DataValueLong = new DataValue(v: JLong)
+  @inline implicit def toDataValue(v: Double): DataValueDouble = new DataValue(v: JDouble)
+  @inline implicit def toDataValue(v: Boolean): DataValueBool = new DataValue(v: JBoolean)
+  @inline implicit def toDataValue(v: Float): DataValueFloat = new DataValue(v: JFloat)
+  @inline implicit def toDataValue(v: Byte): DataValueByte = new DataValue(v: JByte)
+  @inline implicit def toDataValue(v: Int): DataValueInt = new DataValue(v: JInt)
+  @inline implicit def toDataValue(v: Short): DataValueShort = new DataValue(v: JShort)
+
```
mbeckerle: Am I right that you didn't provide DataValue.apply(v: Long) = new DataValue(v: Long) for each acceptable DataValue anyref type?

So I have to write DataValue.toDataValue(myLong) instead of DataValue(myLong) ? 

Or maybe with these implicits you can just have a DataValue.apply(dv: DataValue) = dv and you pass any of the accepted types and it auto coerces them to a DataValue ? 

Or maybe you just don't need this. The implicits are enough and you never need an explicit constructor?

I'm not suggesting a change here necessarily at all. I just fully expected object DataValue to have a flurry of apply methods.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+import org.apache.daffodil.util.Maybe.Nope
+import java.io.ObjectOutputStream
+import java.io.ObjectInputStream
+import java.net.URI
+
+/*
+ * These traits are used for the phantom type X. When considering the type of a
+ * DataValue. See the comment on the DataValue class below.
+ */
+trait NonNullable extends Nullable;
```
mbeckerle: Ok, I am confused. 

Nullable has to do with the DataValue having a "Null" that is the NoValue sentinel. I suggest name is misleading. Null suggests Nil which is not what this is about at all. It's about the fact that the DataValue class represents a hybrid of a Maybe[T] type with a constructor system that prevents constructing (or selecting out) anything but one of our approved values. So to me DataValuePrimitive and NoValue and DataValuePrimitiveOrNoValue are the right names. "Null" just confuses the heck out me for some reason. I believe it is going to work to have NoValue be both an object name and type name. 

tUseNilForDefault is about the sentinel hack we use to indicate that if defaulting, we should instead set the nilled flag. Suggest UseNilForDefaultSentinelType.

Please add scaladoc to this effect so that these things don't get conflated in readers minds.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+  @inline def unsafeFromAnyRef(v: AnyRef) = new DataValue(v)
+  @inline def unsafeFromMaybeAnyRef(v: Maybe[AnyRef]) = {
+    if (v.isDefined) {
+      new DataValue(v.get)
+    } else {
+      NoValue
+    }
+  }
+
+  val NoValue: DataValueEmpty = new DataValue(null)
```
mbeckerle: Since you have unsafe constructors I suggest there should *never* be a null pointer in the value slot of these things. NoValue should be represented by new DataValue(NoValueIndicator) where NoValueIndicator is some other private sentinel. That way even your unsafe constructor can reject null if passed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
mbeckerle: Nevermind. You can't say "this" in an AnyVal because there is no "this".

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+ * that can represent any subset of types which we want. It is further possible to
+ * allow for automatic upcasting between these types where appropriate.
+ *
+ * We inline all methods because attempts to call non-inlined methods trigger boxing
+ */
+final class DataValue[+T <: AnyRef, +X <: AnyRef] private (val v: T) extends AnyVal {
+  @inline def isEmpty = DataValue.NoValue.v eq v
+  @inline def isDefined = !isEmpty
+  @inline def value = v
+  @inline override def toString = if (isEmpty) "NoValue" else "DataValue(" + v.toString + ")"
```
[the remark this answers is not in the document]
mbeckerle: This is the right thing. 

I wonder, however, if the debugger in Eclipse, which is effectively the Java debugger, will call this toString or really only sees the underlying object.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+}
+
+object DataValue {
+  type DataValueAny = DataValue[AnyRef, Nullable]
+  type DataValueNonEmpty = DataValue[AnyRef, NonNullable]
+  type DataValuePrimitive = DataValue[AnyRef, NonNullable with DataValuePrimitiveType]
+  type DataValuePrimitiveNullable = DataValue[AnyRef, Nullable with DataValuePrimitiveType]
+  type DataValuePrimitiveOrUseNilForDefaultOrNull = DataValue[AnyRef, tUseNilForDefault]
```
mbeckerle: This one I think is trying to make a distinction we don't need. UseNilForDefault sentinel is just another notion of NoValue plus "oh yeah and if you're considering defaulting this thing, then here's another tidbit of info".

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -0,0 +1,209 @@
...
+  type DataValueDouble = DataValue[JDouble, NonNullable with DataValuePrimitiveType]
+  type DataValueBigInt = DataValue[JBigInt, NonNullable with DataValuePrimitiveType]
+  type DataValueString = DataValue[JString, NonNullable with DataValuePrimitiveType]
+  type DataValueURI = DataValue[URI, NonNullable with DataValuePrimitiveType]
+  type DataValueFloat = DataValue[JFloat, NonNullable with DataValuePrimitiveType]
+  type DataValueByte = DataValue[JByte, NonNullable with DataValuePrimitiveType]
+  type DataValueInt = DataValue[JInt, NonNullable with DataValuePrimitiveType]
+  type DataValueShort = DataValue[JShort, NonNullable with DataValuePrimitiveType]
+  type DataValueDINode = DataValue[DINode, NonNullable with DataValuePrimitiveType]
+  type DataValueUseNilForDefault = DataValue[UseNilForDefaultObj, NonNullable]
```
mbeckerle: This one is the type for data values that can only be UseNilForDefault ? We don't need this do we?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -327,11 +330,11 @@ trait ElementBase
    * nor any PUA-remapped characters. This insures the default/fixed value can still be
    * used for ordinary XML-schema validation outside of Daffodil/DFDL.
    */
-  final lazy val defaultValue: Option[AnyRef] = {
+  final lazy val defaultValue: DataValuePrimitiveOrUseNilForDefaultOrNull = {
```
mbeckerle: Yeah, so here I think it would be fine, and less confusing, if both DataValue.NoValue and DataValue.UseNilForDefault were of type NoValue, so that defaultValue is of type DataValuePrimitiveOrNoValue, just as DISimple's value slots are. The fact that only defaultValue can hold the sentinel indicating useNilForDefault, .... that's such a hack. Using the type system for this, rather than just letting this thing be NoValue and having a boolean flag, which would be so much more maintainable. 

I truly regret the sin of putting in the UseNilForDefault object hack. Basically because you figured out a way to make it type safe! At the cost of having to understand how that can be possible. We truly should go back to a flag.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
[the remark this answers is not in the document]
bsloane1650: It doesn't look like we ever constructed an isolated testcase to demonstrate the original problem.

I'll regenerate the original Tresys schema that spawned the bug report with enum support enabled. If that works, we know the problem is fixed.

I suppose I should also isolate a testcase to add to our regression suit. This commit is large enough as is without that, so I'll plan on doing that as a seperate commit.

Also, if this happens to not fix the original problem, I would not consider that to be blocking, so I think we can merge this in before verifying it fixes everything.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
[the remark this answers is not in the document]
bsloane1650: Just tested this:

`object Test{
  def main(args: Array[String]): Unit = { 
    val x:Int = 1;
    val y=x.toString()
  }
}
`

Compiles to:

`
public final class Test$ {
  public static final Test$ MODULE$;

  public static {};
    Code:
       0: new           #2                  // class Test$
       3: invokespecial #12                 // Method "<init>":()V
       6: return

  public void main(java.lang.String[]);
    Code:
       0: iconst_1
       1: istore_2
       2: iload_2
       3: invokestatic  #20                 // Method scala/runtime/BoxesRunTime.boxToInteger:(I)Ljava/lang/Integer;
       6: invokevirtual #24                 // Method java/lang/Object.toString:()Ljava/lang/String;
       9: astore_3
      10: return
`

Switching to
`

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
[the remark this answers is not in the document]
bsloane1650: I have observed both.

Normally, the debugger calls the underlying toString, and shows no evidence that the variable is a DataValue.

The only time I recall seeing DataValue in the debugger is when I was tracing down DataValue.toString getting called, and the variable I was looking up ended up being boxed.

This makes sense. If the variable is not boxed, the JVM should have no idea it is not the underlying type.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: You are correct. these methods do not trigger boxxing.

Also, the inline annotation appears to do nothing. When I tested, AnyVal methods get compiled and used as static methods regardless of the inline annotation.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: One of the things I avoided doing was distinguishing the infoset dataValue from DPath values.

Since DINode are legal as DPath values, they are therefore a primitive type.

I suppose we could distinguish this in the same was we distinguish UseNilForDefault, but I am not convinced it is worth the complexity. In particular, we will need to explicitly downcast whenever we want to assign a DPath value to an infoset dataValue. Since this downcast would not actually verify the type involved is not a DINode, creating this distinction might just give us complication and a false sense of security.

We can get away with it with UseNilForDefault because we only have 1 place where we need the downcast.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: Sadly, this doesn't work.

The below program does not get caught

```
object Test{
  def main(args: Array[String]): Unit = {
    val x:Foo = new Foo("x")
    print(x)
  }

}

class Foo(val v:String) extends AnyVal{

} 

object Foo{
  import scala.language.implicitConversions
  implicit def toAnyRef(f:Foo):AnyRef = {
    println("Cast to AnyRef")
    f.v
  } 
} 
```

Putting in `val y:AnyRef = x` gives the compile error:

```test.scala:4: error: the result type of an implicit conversion must be more specific than AnyRef
```

Without the implicit, this gives a compile error at runtime.

The core issue here is that boxing is not a cast. The goal of boxing is to make the system behave as-if the AnyVal type were a typical object type. Scala does not give us a hook to override boxing behaviour.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
bsloane1650: Now that I think about it, there isn't really a reason we could have NoValue=DataValue(null). I believe this is what we do with the Maybe class.

My understanding is the 'nullable type' is standard terminology for a type which has some form of a "no value" sentinal value. It is confusing here probably because almost all types in Scala are nullable (notably, AnyVal types are not nullable).

I would use Maybe,Optional, or Opt in the name, but those are already taken by our other answers to nullable types.

I have no major problem with DataValuePrimitiveOrNoValue, the type name is already pretty long as is.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: I just relied entirely on the implicits.

Given the implicit, `DataValue(x)` would be equivalent to `x:DataValueAny` (or whatever suitable specialization covers the set of types you wrote apply for).

I have not even found any use for the above approach, instead finding it much clearer to put the annotation of the variable that gets written to. Eg:

val x:DataValuePrimitive = bar match{
 case a => 1
 case b => "a"
}

bsloane1650: I just relied entirely on the implicits.

Given the implicit, `DataValue(x)` would be equivalent to `x:DataValueAny` (or whatever suitable specialization covers the set of types you wrote apply for).

I have not even found any use for the above approach, instead finding it much clearer to put the annotation of the variable that gets written to. Eg:

val x:DataValuePrimitive = bar match{
 case a => 1
 case b => "a"
}

The one exception is a cast from AnyRef, which does require DataValue.unsafeFromAnyRef. In this case, I think we do not want an apply method, to force the developer to make sure that this is what they want to do.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: UseNilForDefault is not a NoValueType.

Almost everywhere we have DataValuePrimitiveOrNoValue, UseNilForDefault is not a legal value.

The `trait NonNullable extends Nullable` also turns out to be important in order to get auto upcasting to work. Once you get your mind around this (which I have yet to do), adding a third trait to the hierarchy is simple.

This contravariance stuff is confusing enough as is. I would prefer to keep it standard inheritance hierarchy that we flip upside down; instead of trying to combine it with the singleton definitions.

bsloane1650: UseNilForDefault is not a NoValueType.

Almost everywhere we have DataValuePrimitiveOrNoValue, UseNilForDefault is not a legal value.

The `trait NonNullable extends Nullable` also turns out to be important in order to get auto upcasting to work. Once you get your mind around this (which I have yet to do), adding a third trait to the hierarchy is simple.

This contravariance stuff is confusing enough as is. I would prefer to keep it standard inheritance hierarchy that we flip upside down; instead of trying to combine it with the singleton definitions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
bsloane1650: Disregard my first paragraph. NoValue=DataValue(null) is exactly what we are currently doing.

bsloane1650: Disregard my first paragraph. NoValue=DataValue(null) is exactly what we are currently doing.

bsloane1650: Disregard my first paragraph. NoValue=DataValue(null) is exactly what we are currently doing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: I think the question here is what we want `unsafeFromAnyRef(null)` to mean.

In my mind, `null` is AnyRef's version of NoValue, so it makes sense that it gets converted into NoValue. I don't see any reason why our unsafe converter shouldn't allow this. If the caller wants to allow for NoValue, then not using null would mean they need to wrap their conversion in an if statement so they can produce NoValue explicitly.

If they do not want to allow NoValue, then they are responsible for making sure they are providing a legal value, which would including not being null in this case.

It sounds like you want a safeAnyRefToDataValuePrimitive functions. We could write such a function; it would just require a bunch of type checks as well as a null check (and we might also want a simmilar function for our other DataValue types, including safeAnyRefToDataValuePrimitiveNullable).

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: No. This is just copy pasta. Every type we use gets a type alias, an asFoo accessor, and an implicit toDataValue converter.

This seems simpler then deciding which types need which things.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
bsloane1650: I have no problem with switching to using just NoValue with a seperate boolean flag.

Short of doing that, we should keep what we are doing explicit.

Keep in mind that UseNilForDefault != NoValue, so all you really accomplish by putting them both in the NoValue type-class is re-introducing the hack of sticking UseNilForDefault where it does not belong.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps3.scala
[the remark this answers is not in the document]
mbeckerle: So long as we do cover the enum fix ultimately I'm ok with more than one commit/PR.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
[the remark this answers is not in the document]
mbeckerle: Yeah, I guess since toString is a virtual function, you must have an object to invoke it on. 

Hopefully for some set of small ints, they return the same object rather than allocating new ones.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
mbeckerle: Perhaps doesn't inline because we have debug flags in our compile options?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
mbeckerle: Well, DINodes are never the values of DINodes. We may pass them in DPath to do things like get fn:namespace or something, but they never can be returned by say, inputValueCalc, or outputValueCalc. They can never become the value of an InfosetElement.

That's what makes this confusing. 

I would put in a comment like "Reserved for future use if we use DataValues in DPath."

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
mbeckerle: I'll buy into null being AnyRef's version of NoValue.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
mbeckerle: If you want to keep Nullable as the name because "nullable type" is a standard term, then you need to explain that in comments, because it wasn't standard to me and I'm reasonably well informed on these things, though not expert.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
mbeckerle: Ah. In the above comment I hadn't realized that UseNilForDefault doesn't actually show up as the value of a DISimple. It only appears in the ERD to signify that the value, if not present, can be defaulted not to the default value, but to nil.  
So I withdraw the suggestion.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
mbeckerle: We can keep it the way you have it. I would like explanation of Nullable in scaladoc so that it's less confusing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
bsloane1650: This code path is not covered by our tests.

Regardless, I fixed the issue since I noticed it. I'm not dealing with the coverage issue now. If we care about that, we should look at our test coverage separately.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: I removed the inline portion of this comment.

I did my testing using the scalac directly from the commandline. It wouldn't suprise me if debug mode was the default though.

Regardless, I see no reason not to inline these.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/DataValue.scala
bsloane1650: This isn't a future use. This commit uses the DataValue type as part of DPath:

Note arg1 below is a DINode.

```
abstract class FNTwoArgsNodeAndValue(recipes: List[CompiledDPath])
  extends RecipeOpWithSubRecipes(recipes) {
  override def run(dstate: DState) {

    val recipe1 = recipes(0)
    val recipe2 = recipes(1)

    val savedNode = dstate.currentNode
    dstate.resetValue()
    recipe1.run(dstate)
    val arg1 = dstate.currentNode

    dstate.setCurrentNode(savedNode)
    recipe2.run(dstate)
    val arg2 = dstate.currentValue.getNonNullable

    dstate.setCurrentValue(computeValue(arg1, arg2, dstate))
  }

  def computeValue(arg1: DataValuePrimitive, arg2: DataValuePrimitive, dstate: DState): DataValuePrimitive

  override def toXML = toXML(recipes.map { _.toXML })
}
```

There is certainly merit in having a seperate type (alias) for types that are usable in the infoset and types that are usable within DPath; I'm just not convinced it will actually be worth it given how our code is currently structured.

mbeckerle: Yeah. The case you cite is for the dfdl:contentLength and dfdl:valueLength functions, and I suppose fn:count would be similar. Nodes there are arguments. Never results, but the DPath engine has to pass nodes around by way of creating suspensions that eventually compute their values. Methods there might even be returning nodes. So I guess I'm pretty well convinced.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/Infoset.scala
bsloane1650: Comment added above the type alias

mbeckerle:
There are no new tests under daffodil-test in this squashed commit. 
Is there a TDML test that exercises the enum feature the way we expect customers will use it?

bsloane1650:
> There are no new tests under daffodil-test in this squashed commit.
> Is there a TDML test that exercises the enum feature the way we expect customers will use it?

There is no functional change in this commit.

I did fix to bugs/mistakes, however:

* 1 was simply dead code. We were testing if the result of eval was ever a PState. 
* A second was a Scala BigInt vs Java BigInteger mistake. This turned out not to be the cause of the type-calc issue, and I do not know of a code path that would expose this mistake (not that I looked particuarly hard for one).

We have some tests from the original PR at daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc 

None of these were created with pedagogy in mind, but some of the early examples  in inputTypeCalc.tdml (namely the keysetValue examples) show the simple usecase; where the parser consumes a byte and outputs a string and the unparser does the inverse.

I am working on a seperate PR to fix the original typeCalc problem that started this. That fix builds off of the work here, but involves a fairly significant refactoring of the typeCalc system, so I want to keep it as a seperate PR.

That PR will add a TDML test that demonstrates the more advanced usecase where both the integer and string representations are preservered in the infoset.

For pedagogical purposes, it might makes sense to create a separate tdml file with just example test cases. As is, the useful examples are mixed in with more unit-testy ones.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -55,24 +103,31 @@ case class FNAbs(recipe: CompiledDPath, argType: NodeInfo.Kind) extends FNOneArg
...
+  }
 }
 
 // No such function in DFDL v1.0
 // But leave this here because we probably will want to add it as a
 // daffodil extension function and then eventually hope it gets into DFDL1.1
+// 12/12/2019 - We refactored how these functions are defined. The args should now be of type DataValuePrimitive
 //case class FNStringJoin(recipes: List[CompiledDPath]) extends FNTwoArgs(recipes) {
 //  override def computeValue(arg1: AnyRef, arg2:AnyRef, dstate: DState) = {
 //    val values = arg1.asInstanceOf[List[String]]
```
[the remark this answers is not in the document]
stevedlawrence: This entire comment doesn't really seem worth keeping to me, or at least not updating to use the new DataValuePrimitive. The fn:string-join function looks like it is supposed to accept a sequence of arguments and concats them all together. So it's fn:concat, but with varargs. Daffodil doesn't support sequences, and I don't think that's anywhere on the nearterm roadmap. And the commented function only accepts two args anyway, so doesn't even seem correct. Seems like this is just cruft.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ConvertTextBooleanUnparser.scala
```diff
@@ -38,7 +38,7 @@ case class ConvertTextBooleanUnparser(
 
     val node = state.currentInfosetNode.asSimple
 
-    val boolValue = node.dataValue match {
+    val boolValue = node.dataValue.getAnyRef match {
```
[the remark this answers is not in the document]
bsloane1650: I was just spamming getAnyRef in all of these match statements. In this case, getBoolean probably makes more sense.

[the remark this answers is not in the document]
bsloane1650: Having said that, I don't want to go through and update all of these.

Previously, we were just using anyRef's anyway, and were inconsistently handling cases where the type was not as expected. Given that this is supposed to be a non-functional change, I do not want to update any of the bevaiours, so will just leave it as getAnyRef.

PR #302, opened 2019-12-13
Remove GitHub Actions linux dependency step

stevedlawrence:
We currently run apt-get to install linux dependencies like RPM and SBT
as part of our Github Actions workflow. Unfortunately, GitHub has
updated their container to include a Microsoft repository that is buggy
and started breaking builds everyone's builds that ran apt. We could
remove just this buggy repo, but other repos could always break in the
future.

Fortunately, it the GitHub Actions container now includes SBT and RPM by
default, so our dependencies are already met. So remove the Install
Linux Dependencies step so we never have to worry about random
repositories breaking our build.

DAFFODIL-2253

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,15 +36,6 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v1.0.0
 
-      - name: Install Dependencies (Linux)
-        run: |
-          echo "deb https://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
-          sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
-          sudo apt-get -y update
-          sudo apt-get -y install rpm sbt
-        shell: bash
-        if: runner.os == 'Linux'
-
       - name: Install Dependencies (Windows)
         run: |
           choco install sbt


.github/workflows/main.yml
```diff
@@ -36,15 +36,6 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v1.0.0
 
-      - name: Install Dependencies (Linux)
-        run: |
-          echo "deb https://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
-          sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
-          sudo apt-get -y update
```
olabusayoT: Don't we still need this update statement, or are all the dependencies kept up to date on their end?

stevedlawrence: Good question. I could see an argument both for and against this. Argument for is we probably do always want to test with the latest version of dependencies. Argument against is that random repo updates could break things (like what led to this PR).

In practice, I'm not sure package updates really matters. Thisis because the only dependnecies Daffodil really has to the system are SBT, RPM, and Java. All other dependiecies come from maven with explicit versions.

For SBT, our build.properties specifies the version to use and SBT autodownloads it. So it doesn't really matter what's on the system as long as it's newer than some really old version.

For RPM, RPM is really good about not breaking backwards compatbility, so we should be fine even if we're using a slightly older version. Newer versions would still work if someone used them.

And for Java, our GitHub actions workflow explicitly states to use the latest version of Java. That's not effected by an apt update.

So I don't think an apt-update would ever really effect our builds, except for breaking them when microsoft breaks their repo. If we ever do start depending on anything on the system, then we probably do want to think about adding this back, but as it is right now, I don't think it's neccessary, and potentially leads to errors.

olabusayoT: Ah ok, makes sense

bsloane1650: FYI, apt-get update does not actually update the system. It updates apt's package database.

apt-get upgrade is what would update all packages installed on the system.

PR #303, opened 2019-12-17
Daffodil 2188 typecalc bigint

bsloane1650:
This builds off of the still in review DPath type safety PR.

Until that gets merged, ignore the first few commits of this PR, as those are really part of the other one.


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1909,7 +1909,7 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
     funcObj
   }
 
-  def lookupTypeCalculator(typeCalcExpr: Expression, requiresParse: Boolean, requiresUnparse: Boolean): TypeCalculator[AnyRef, AnyRef] = {
+  def lookupTypeCalculator(typeCalcExpr: Expression, requiresParse: Boolean, requiresUnparse: Boolean): TypeCalculator = {
```
bsloane1650: We no longer attempt to represent the types a particular calculator uses with the Scala type system. The src/dst types are saved as fields of the calculator. From the perspective of the Scala type system, all calculators are now DataValue -> DataValue

daffodil-lib/src/main/scala/org/apache/daffodil/util/RangeBound.scala
```diff
@@ -1,111 +0,0 @@
-/*
```
bsloane1650: Since RangeBound now depends on DataValue, we had to move it out of daffodil-lib.

For lack of a better place to put it, it is now part of TypeCalculator.scala

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -19,6 +19,7 @@ package org.apache.daffodil.processors
 
 import scala.collection.immutable.HashMap
 import scala.collection.immutable.HashSet
+import scala.math.BigInt
```
bsloane1650: This can't be right...

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -31,12 +32,36 @@ import org.apache.daffodil.processors.parsers.ParseError
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe.One
+import org.apache.daffodil.util.Numbers
 import org.apache.daffodil.util.PreSerialization
-import org.apache.daffodil.util.RangeBound
 import org.apache.daffodil.util.TransientParam
 import org.apache.daffodil.xml.QNameBase
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
```
bsloane1650: Clean imports

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -125,67 +148,79 @@ abstract class TypeCalculator[A <: AnyRef, B <: AnyRef](val srcType: NodeInfo.Ki
 
   @throws(classOf[java.io.IOException])
   final private def writeObject(out: java.io.ObjectOutputStream): Unit = serializeObject(out)
+
+  protected def normalizeArg(x: DataValuePrimitive, xType: NodeInfo.Kind): DataValuePrimitive = {
```
bsloane1650: All of this normalizeArg stuff shouldn't be nessasary. In theory we know at compile time exactly what types we expect to see, and can specialize then. In practice, we convert all integral types to JBigInt at compile time, and so must do the same at runtime.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -412,3 +451,83 @@ object TypeCalculatorCompiler {
   }
 
 }
+
```
bsloane1650: The below block is a modified version of the RangeBound file we removed above.

Instead of having a type parameter, it assumes the bound is a DataValuePrimitive.

This is a good example of why we might want a richer DataValue hierarchy. In theory we could declare this with DataValueNumeric instead. I am aware of no reason we cannot do this, but in practice, declaring new DataValue types seems to be very finicky work, so I do not think it is worth the benifit here.

Also, in practice RangeBound only every uses JBigInt at the moment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ElementKindParsers.scala
```diff
@@ -21,7 +21,10 @@ import org.apache.daffodil.processors._
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.util.RangeBound
+import java.math.{BigInteger => JBigInt}
+import org.apache.daffodil.infoset.DataValue.DataValueString
```
bsloane1650: Clean imports

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
```diff
@@ -90,20 +90,18 @@ object RepValueCooker extends ListOfStringLiteralNonEmptyNoCharClassEntitiesNoBy
...
+object IntRangeCooker extends Converter[String, Seq[(JBigInt, JBigInt)]] {
+  protected def convert(input: String, context: ThrowsSDE, forUnparse: Boolean): Seq[(JBigInt, JBigInt)] = {
+    def run(xs: Seq[String]): Seq[(JBigInt, JBigInt)] = {
       xs match {
         case Seq() => Seq()
         case a +: b +: rest => {
-          val a2 = BigInt(a)
-          val b2 = BigInt(b)
-          if (a2.compare(b2) > 0) {
+          val a2 = new JBigInt(a)
```
mbeckerle: Even though this usage is transient, we're stamping out all use of BigInt in favor of JBigInt. I think that's good, but have you done that comprehensively, or selectively in this change set?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
mbeckerle: Add a comment to the code stating just what you said. This should be recorded for someone else reading the code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
mbeckerle: Put above comment into the code please.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/typeCalcFunctions.tdml
```diff
@@ -281,5 +313,81 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+  <tdml:parserTestCase name="sparse_enum_01"
+    root="sparse_enum01" model="inputTypeCalc-Embedded.dfdl.xsd" description="Extensions - inputTypeCalc keysetValue transform">
+
+    <tdml:document>
+    <tdml:documentPart type="byte">
+    00
```
mbeckerle: If you want this line per byte preserved, which I see you do given that each byte corresponds to an infoset item and those are vertically spread, you need to put <![CDATA[ .... ]]> around them. Then XML formatting won't screw with your whitespace.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -153,12 +160,12 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
...
             SDE(
               "%s value (%s) is not unique across all branches of a direct dispatch choice. Offending branches are:\n%s",
-              branchKeyAttribute, k, 
+              branchKeyAttribute, k,
               kvs.map(_._2.context.runtimeData).map(rd => rd.diagnosticDebugName + " " + rd.schemaFileLocation.locationDescription).mkString("- ", "\n- ", ""))
           }
           Try(k.toLong) match {
             case Success(v) => {
-              val asBigInt = BigInt(v)
+              val asBigInt = JBigInt.valueOf(v)
```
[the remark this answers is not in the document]
bsloane1650: Possibly. There were several BigInt -> JBigInt changes. At least one of them was the bug fix.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
bsloane1650: Selectively. I didn't have a good way of finding all instances of scala BigInt. Switching over to DataValue uncovered a lot of them, but it is possible that we have more hiding someplace.

Based on some of the recent debugging that I have been doing, it occurs to me that I can use the dissembler to comprehensibly search for scala's BigInt. I'll give that a go and see if it uncovers anything.

build.sbt
```diff
@@ -105,7 +105,7 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "2.5.0-SNAPSHOT",
-  scalaVersion := "2.12.6",
+  scalaVersion := "2.11.12",
```
bsloane1650: Revert

daffodil-runtime1/src/main/scala-2.11/org/apache/daffodil/infoset/DataValue.scala
```diff
@@ -99,8 +99,8 @@ final class DataValue[+T <: AnyRef, +X <: AnyRef] private (val v: T) extends Any
   @inline def getString = v.asInstanceOf[JString]
   @inline def getURI = v.asInstanceOf[URI]
 
-  @inline def getNonNullable = this.asInstanceOf[DataValue[T, X with NonNullable]]
-  @inline def getNullablePrimitive = this.asInstanceOf[DataValue.DataValuePrimitiveNullable]
+  @inline def getNonNullable:DataValue[T, X with NonNullable] = new DataValue(v)
+  @inline def getNullablePrimitive:DataValue.DataValuePrimitiveNullable = new DataValue(v)
```
bsloane1650: Not actually relevent, but using asInstanceOf above introduces boxing.

mbeckerle: Really. Wow, so you can't even asInstanceOf an AnyVal type without it being boxed....

We need a section in our wiki about using AnyVal types in the coding style and coding for performance things. In code review we should be looking for this sort of thing. Right now I think it just says there are some non-boxing and value class types like MStack and suggests using them, but doesn't give a do's and don'ts list.

bsloane1650: Yeah. I did a grep through a dissasembly of our source code, and found 188 instances of Maybe."<init>" and 78 instances of DataValue."<init>"

One place these come up seem to be pass-by-name parameters, which would probably account for alot of these. I didn't look through all of them, so I do not know if there are other places we introduced boxing.

I did verify that all of the toDataValue functions do not introduce boxing, so we are doing at least something write.

I googled around and found a recent Pre-SIP to introduce true non boxing wrapper types: https://contributors.scala-lang.org/t/pre-sip-unboxed-wrapper-types/987

It is worth keeping an eye on this and switching over to it when/if it becomes available.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
bsloane1650: I can confirm that we still have usages of scala BigInt and BigDecimal in places. The few I looked into turned out to just be unit tests.

It is probably worth cleaning these up as well, but seems out of scope for this PR.

PR #308, opened 2019-12-30
Fixed issue unparsing optional unordered elements

jadams-tresys:
A user discovered an issue when the first element of an unordered
sequence is optinal and does not have a maxOccurs > 1. There was a logic
error that was resulting in the element being treated as a scalar
element when it should have been treated as an array, due to it being
optional.

DAFFODIL-2261


jadams-tresys:
The test coverage was all from the user, so I can't take credit for that.
Looks like I may need to make a line ending change though to get this test working on Windows.

PR #309, opened 2019-12-31
Fixed incorrect separator suppression with ock=expression

jadams-tresys:
A user ran into a bug involving incorrect separator suppresion when a
sequence was using occursCountKind='expression', which should behave as
if separatorSuppressionPolicy='never'.

DAFFODIL-2262

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
@@ -532,14 +532,27 @@ class RepOrderedExactlyNSequenceChild(sq: SequenceTermBase, e: ElementBase, grou
 
 }
 
-class RepOrderedExactlyTotalOccursCountSequenceChild(sq: SequenceTermBase, e: ElementBase, groupIndex: Int)
+class RepOrderedExpressionOccursCountSequenceChild(sq: SequenceTermBase, e: ElementBase, groupIndex: Int)
   extends RepElementSequenceChild(sq, e, groupIndex) {
 
   lazy val sequenceChildParser: SequenceChildParser = sq.hasSeparator match {
-    case true => new RepOrderedExactlyTotalOccursCountSeparatedSequenceChildParser(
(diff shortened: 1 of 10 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/usertests/Book2.dfdl.xsd
```diff
@@ -0,0 +1,64 @@
...
+								dfdl:occursCount="{ fn:count(../../header/title) }"
+								dfdl:occursCountKind="expression" />
+						</xs:sequence>
+					</xs:complexType>
+				</xs:element>
+			</xs:sequence>
+		</xs:complexType>
+	</xs:element>
+      
</xs:schema>
```
stevedlawrence: Looks like there's a few carriage returns in this file, mixing windows and linux line endings. Would be nice to remove them.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
```diff
@@ -532,15 +532,28 @@ class RepOrderedExactlyNSequenceChild(sq: SequenceTermBase, e: ElementBase, grou
...
       childParser, e.occursCountEv, srd, erd, sepParser, sq.separatorPosition, separatedHelper)
-    case false => new RepOrderedExactlyTotalOccursCountUnseparatedSequenceChildParser(childParser, e.occursCountEv, srd, erd,
+    case false => new RepOrderedExpressionOccursCountUnseparatedSequenceChildParser(childParser, e.occursCountEv, srd, erd,
       unseparatedHelper)
   }
+
+  // This class is only used for sequences with occursCountKind="expression",
+  // which means that the separatorSuppressionPolicy is not applicable and the
+  // implied behavior is separatorSuppresionPolicy="never"
+  override lazy val sequenceChildUnparser: SequenceChildUnparser = sq.hasSeparator match {
```
mbeckerle: Looks like this crazy factoring of the sequences stuff does work. This fix was straightforward and didn't require any refactoring.

PR #310, opened 2020-01-02
Add more automation to release candidate container

stevedlawrence:
- Install the wix shell script as part of creating the container instead
  of getting it from the cloned daffodil repo. This avoids having a
  broken symlink before the daffodil repo is cloned
- Change how we clone the svn repo to avoid unneeded directories
- svn remove any previous release candidates if they exist. Needed for
  when doing an rc2, for example
- svn add the new release files so they are ready for commit if all
  checks pass
- Put final steps in a script to automate release after all checks pass
- Add some missing suppression of popd output

DAFFODIL-2264

### DIFF
--- a/containers/release-candidate/daffodil-release-candidate
+++ b/containers/release-candidate/daffodil-release-candidate
@@ -84,7 +84,7 @@ if [ $? -ne 0 ]; then
    echo -e "\n!!! Ensure you have created a bind mount of your gitconfig file, for example with -v ~/.gitconfig:/root/.gitconfig !!!\n"
    exit
 fi
-popd
+popd &> /dev/null
 rm -rf $TEST_GIT_DIR
 
 if [ ! -d /root/.ssh/ ]; then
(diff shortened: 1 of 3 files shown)


containers/release-candidate/daffodil-release-candidate
```diff
@@ -170,6 +165,8 @@ sbt \
   "daffodil-japi/genjavadoc:doc" \
   "daffodil-sapi/doc" \
 
+echo "Removing old release candidates..."
+svn delete --force $DAFFODIL_DIST_DIR/$VERSION-* 2> /dev/null
```
mbeckerle: All this piping errors to dev null seems very problematic to me. Under maintenance if one goofs up the pushd/popd pairings,  or goof up the definitions of things like $DAFFODIL_DIST_DIR you won't get any diagnostics.  At least not from these commands that will fail. So long as that's not a problem (because the errors would be obvious elsewhere in the script) then I'm ok with it I guess. 

If the script is mostly straight line code, then there's little once can mess up under maintenance, so my concern may be unwarranted. 

Is there a less fragile way to do this? E.g., just redirect to a temp file,.....

stevedlawrence: Yeah, I was being a little lazy here. The reason for the /dev/null redirect here is that there won't be an existing release candidate if this is the first one. So svn will output an error that there's nothing to delete. I just wanted to hide that, but you're right that this could potentially hide other real errors.

It's probably a better idea to check for the existence before deleting. I'll make the update.

PR #314, opened 2020-01-13
Add support for textStandardBase

stevedlawrence:
Add new primitive, parser, and unparser to handle non base-10 text
numbers. Actual logic from converting to/from bases is handled by Java's
builtin functions.

To maintain backwards compatibility, if the textStandardBase property is
not defined it defaults to "10" unless the requireTextStandardBaseProperty
tunable is set to true, default to false.

Also noticed some unordered sequence tests were commented out and work
with minor tweaks due to incorrect tests.

DAFFODIL-840

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -605,7 +605,12 @@ trait ElementBaseGrammarMixin
   }
 
   private lazy val textStandardNumber = prod("textStandardNumber", textNumberRep == TextNumberRep.Standard) {
-    ConvertTextCombinator(this, stringValue, textConverter)
+    val converter = textStandardBaseDefaulted match {
+      case 10 => textConverter
+      case 2 | 8 | 16 =>  textStandardNonBaseTenConverter
+      case _ => Assert.impossible()
+    }
+    ConvertTextCombinator(this, stringValue, converter)
   }
 
   private lazy val textZonedNumber = prod("textZonedNumber", textNumberRep == TextNumberRep.Zoned) {
@@ -625,13 +630,17 @@ trait ElementBaseGrammarMixin
     }
   }
 
+  private lazy val textStandardNonBaseTenConverter = {
+    primType match {
+      case _: NodeInfo.Integer.Kind => ConvertNonBaseTenTextNumberPrim(this)
+      case _ => SDE("dfdl:textStandardBase=\"%s\" cannot be used with %s", textStandardBaseDefaulted, primType.globalQName)
+    }
+  }
+
   private lazy val textZonedConverter = {
     primType match {
-      case PrimType.Double | PrimType.Float =>
-        SDE("dfdl:textNumberRep=\"zoned\" is not allowed for %s", primType.globalQName)
-      case _: NodeInfo.Numeric.Kind => ConvertZonedNumberPrim(this)
-      case PrimType.HexBinary | PrimType.Boolean | PrimType.Date | PrimType.Time | PrimType.DateTime | PrimType.AnyURI | PrimType.String =>
-        Assert.invariantFailed("textZonedConverter only to be used for numeric types")
+      case _: NodeInfo.Decimal.Kind => ConvertZonedNumberPrim(this)
+      case _ => SDE("dfdl:textNumberRep=\"zoned\" cannot be used with %s", primType.globalQName)
     }
   }
 
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
@@ -95,10 +95,34 @@ object TextNumberBase {
       case "8" => 8
       case "10" => 10
       case "16" => 16
-      case _ => self.schemaDefinitionError("Illegal number base: " + str) // validation will have checked. So this shoudn't happen.
+      case _ => self.schemaDefinitionError("For property textStandardBase, value must be 2, 8, 10, or 16. Found: %s", str)
     }
   }
+}
+trait TextStandardBaseMixin extends PropertyMixin {
 
+  def tunable: DaffodilTunables
+
+  private def optionTextStandardBase = findPropertyOption("textStandardBase")
+
+  /**
+   * Daffodil 2.5.0 and older ignored the textStandardBase property, behaving
+   * as if the value was always 10. Newer versions of Daffodil support this
+   * property, but we don't want to require it and potentially break old
+   * schemas that do not define it. So we check if we should require the
+   * property based on a tunable, and if we shouldn't require it and it's not
+   * defined, then we warn and default to 10.
+   */
+  final lazy val textStandardBaseDefaulted = {
+    val numStr =
+      if (tunable.requireTextStandardBaseProperty || optionTextStandardBase.isDefined) {
+        getProperty("textStandardBase")
+      } else {
+        SDW(WarnID.TextStandardBaseUndefined, "dfdl:textStandardBase property is undefined. Defaulting to 10.")
+        "10"
+      }
+    TextNumberBase(numStr, this)
+  }
 }
 
 sealed trait SeparatorSuppressionPolicy extends SeparatorSuppressionPolicy.Value
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NonBaseTenTextNumberParser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NonBaseTenTextNumberParser.scala
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processors.parsers
+
+import java.math.{ BigInteger => JBigInt }
+import java.lang.{ Number => JNumber }
+
+import org.apache.daffodil.dpath.NodeInfo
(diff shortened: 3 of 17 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -625,6 +630,16 @@ trait ElementBaseGrammarMixin
     }
   }
 
+  private lazy val textStandardNonBaseTenConverter = {
+    primType match {
+      case PrimType.Double | PrimType.Float | PrimType.Decimal =>
+        SDE("dfdl:textStandardBase=\"%s\" cannot be used with %s", textStandardBaseDefaulted, primType.globalQName)
+      case _: NodeInfo.Numeric.Kind => ConvertNonBaseTenTextNumberPrim(this)
+      case PrimType.HexBinary | PrimType.Boolean | PrimType.Date | PrimType.Time | PrimType.DateTime | PrimType.AnyURI | PrimType.String =>
```
bsloane1650: Can this be a wildcard match instead of listing every type?

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
```diff
@@ -95,10 +95,34 @@ object TextNumberBase {
       case "8" => 8
       case "10" => 10
       case "16" => 16
-      case _ => self.schemaDefinitionError("Illegal number base: " + str) // validation will have checked. So this shoudn't happen.
+      case _ => self.schemaDefinitionError("For property textStandardBase, value must be 2, 8, 10, or 16. Found: %s", str)
```
bsloane1650: Aren't we also testing for this in ElementBaseGrammarMixin? Why do we need to check twice?

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
```diff
@@ -145,7 +145,7 @@
 
     <tdml:parserTestCase name="simple_type_properties_text_number_13_03" root="base_group"
 		model="./fvt/ext/dpa/dpanum_properties.dfdl.xsd" 
-		 description="Section 13.5 Specification of number base - 2,8, and 16">
+		 description="Section 13.5 Specification of number base - 2,8, and 16" roundTrip="twoPass">
```
bsloane1650: I would prefer to avoid having twoPass test cases, as they obscure what is actually going on. Can we instead update the document/infoset to match what we are expecting in the second pass. If we car about the details of the intermediate parses, it would be clearer to break this test into multiple tests that should the individual parse/unparse steps with roundTrip="None"

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/BG.tdml
```diff
@@ -25,25 +25,25 @@
     description="Text number properties">
```
bsloane1650: We are making many changes to existins tests in the test-ibm1 project. Do we have compatability concerns to worry about?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
stevedlawrence: Yeah, that's probably the right thing to do in this case. I usually like to list everything because scala will warn you if the match/case isn't exhaustive. So it sort of forces you to think about every case. And if we add a new type (e.g. like we did recently for blobs) this will error and we'll be forced to think about how this new type applies to this match. But in this case I'm not sure if it's worth it.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
stevedlawrence: Yep, the SDE in ElemenetBaseGrammarMixin should just be an assert.

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
stevedlawrence: I usually prefer two-pass over none since they both require that the parse matches the infoset, but you also get some unparse testing to at least make sure things unparse somewhat correctly. Which is all I'm really caring about in this particular test. I generally avoid changing IBM tests (unless they're completely broken), so this gives me some assurance that this is doing the right thing. (Note that two pass is required because there's a leading zero, which doesn't unparse).

The new tests added for textStandardBase are more strict, and do have separate tests for parse vs unparse when we want to ensure that unparse is doing the right thing. So while this test could potentially hide issues due to two pass, other tests should catch those issues.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/BG.tdml
stevedlawrence: I don't think these are actually IBM tests. They're just incorrectly in the test-ibm1 project (looks like they got moved in commit 406c470a5b in 2013). All the AA-BG tests were either added by Tresys or inherited from the original Daffodil codebase. So it's fine to change these (and they were definitely broken).

daffodil-test-ibm1/src/test/resources/test-suite/ibm-contributed/dpaext2.tdml
stevedlawrence: I'll also add that a few of other tests in this file use roundTrip="twoPass", so it is at least consistent with those.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -605,7 +605,12 @@ trait ElementBaseGrammarMixin
   }
 
   private lazy val textStandardNumber = prod("textStandardNumber", textNumberRep == TextNumberRep.Standard) {
-    ConvertTextCombinator(this, stringValue, textConverter)
+    val converter = textStandardBaseDefaulted match {
+      case 10 => textConverter
+      case 2 | 8 | 16 =>  textStandardNonBaseTenConverter
```
mbeckerle: I think we should support base 256 also, where the charcodes 0 to 255 are the digits.

Base e (2.71828...) has also been suggested somewhere.....

Just kidding.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -625,6 +630,15 @@ trait ElementBaseGrammarMixin
     }
   }
 
+  private lazy val textStandardNonBaseTenConverter = {
+    primType match {
+      case PrimType.Double | PrimType.Float | PrimType.Decimal =>
```
mbeckerle: While this is shorter, you are depending on the set of types not changing here. E.g., if a new type is introduced (quad-precision floats?) then this will break. 

It is better to positively select the integer types that support this than use negation of the the floating point types that do not. 

You can still do this without enumerating all the integer types because they're all subtypes of PrimType.Integer.Kind 

`case _: PrimType.Integer.Kind => ...`

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NonBaseTenTextNumberParser.scala
```diff
@@ -0,0 +1,75 @@
...
+    if (baseStr == "") {
+      PE(state, "Unable to parse %s from empty string", context.optPrimType.get.globalQName)
+      return
+    }
+
+    // must explicitly check for and error on text that start with + or -
+    // because DFDL does not allow a leading sign character, but parseInt and
+    // friends will accept them
+    val firstChar = baseStr(0)
+    if (firstChar == '-' || firstChar == '+') {
```
mbeckerle: Please add to comment that DFDL spec says textNumberPattern is not used when textStandardBase is not 10.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/BG.tdml
mbeckerle: 75 or so tests did originate with IBM, which donated them to the DFDL Workgroup and the Daffodil effort 7 years back. 

The AA-ZZ (aka the "two letter tests") originated with the original Daffodil NCSA code base before anybody including myself started working with this code. 

So calling this bunch of tests "ibm1" is really misleading. Nevertheless keeping them distinct from the more organized and methodically created daffodil-test suite makes sense.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
bsloane1650: The DFDL spec is clear on what valid values are for textStandardBase, and they are only 2, 8, 10, 16.  I would want a compelling reason to deviate from this.

Base 256 sounds more like binary encoding; I'm not sure we gain much by allowing it as a textStandardBase as well. Eg. when the string "0" means 48, I do not think you are working with text.

stevedlawrence: What's the use case for base-256? A base-256 encoded number would probably be full of non-display characters, so I'm not sure why that would ever be used in practice?

Implementation also becomes a bit more difficult, since right now we just rely on Java utils to converts to/from base-2,8,16 encoded numbers. Supporting anything means we have to implement our own logic since that's all java supports. Not hard, but more work and more code to maintain. Just want to make sure there's a good use case for it.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
stevedlawrence: Will do, I'll also fix up Zoned with a similar logic while I'm there.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NonBaseTenTextNumberParser.scala
stevedlawrence: WIll do

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
stevedlawrence: @mbeckerle , thoughts on base 256? I think all other comments have been addressed and this commit is ready to merge otherwise.

mbeckerle: You missed the "Just Kidding" at the end of my comment. :-)

+1

stevedlawrence: Ah, I thought you were just talking about base e. I'll merge this.

PR #318, opened 2020-02-03
Add new TimeTracker for more advanced performance analysis

stevedlawrence:
Because of our nested parsers, it can sometimes be difficult to
determine how much time each processor takes on average to complete it's
parse/unparse, even when using a profiler. This adds a new timer that
makes it much easier to determine how much total time particular parsers
take and on average how long they take. This can help to prioritize
performance optimizations.

The intended use is to modify the Parser.scala parse1() function to look
like this:

  try {
    TimeTracker.track(parserName) {
      parse(pstate)
    }
  }

The resulting output gives a break down of all the parsers and how
performant they are. Once slow processors are found, it can then be
useful to wrap subsections of the parse() method in a call to
TimeTracker.track() to determine which sections are slow. This incurs
less overhead than traditional profilers and give more control over
which sections are actually profiled.

DAFFODIL-1883

### DIFF
--- a/daffodil-macro-lib/src/main/scala/org/apache/daffodil/util/TimerMacros.scala
+++ b/daffodil-macro-lib/src/main/scala/org/apache/daffodil/util/TimerMacros.scala
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.util
+
+import scala.reflect.macros.blackbox.Context
+
+object TimeTrackerMacros {
+
+  def trackMacro(c: Context)(name: c.Tree)(body: c.Tree) = {
+    import c.universe._
+
+    val startTime = TermName(c.freshName)
+    val endTime = TermName(c.freshName)
+    val childrenTime = TermName(c.freshName)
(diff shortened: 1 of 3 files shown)


stevedlawrence:
An example of the output, adding the TImeTracker around the parse() call like mentioned in the commit message and parsing a large CSV-like data format, we get the following stats:
```
Name                                 Time     Pct  Average    Count
LiteralNilDelimitedEndOfDataParser  3.330  34.03%     4030   826140
StringDelimitedParser               2.455  25.09%     4184   586640
DelimiterTextParser                 1.038  10.61%      879  1180480
SimpleNilOrValueParser              0.985  10.07%     1192   826140
OrderedSeparatedSequenceParser      0.806   8.23%    10232    78720
ElementParser                       0.404   4.13%      342  1180520
DelimiterStackParser                0.308   3.15%      244  1259220
ChoiceParser                        0.226   2.31%     5750    39360
SeqCompParser                       0.113   1.15%      318   354300
ConvertTextNumberParser             0.060   0.61%  1489652       40
OrderedUnseparatedSequenceParser    0.058   0.60%  2922016       20
ConvertTextCombinatorParser         0.000   0.00%     8825       40
Total Time: 9.783
```
So we can see we spend a third of the time trying to parse delimted Nils, a quarter of the time trying to parse delimited strings, a thenth parsing actual delimiters, and so on. So to get any real benefits, we really just need to focus on speeding up these parsers. Note that this isn't too surpsing, we would expect delimiter scanning to take up a good chunk of the time, but it's helpful to see that in number form.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/util/TimerMacros.scala
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.util
+
+import scala.reflect.macros.blackbox.Context
+
+object TimeTrackerMacros {
+
+  def trackMacro(c: Context)(name: c.Tree)(body: c.Tree) = {
```
bsloane1650: The usage of childrenTimeStack could use a comment.

My first reading was that each block had its own childrenTimeStack, that each of the children pushed onto. This seems to be backwards to what is actually happening, where there is a single stack, where the top element is the total time that all of the children parsers took. What is actually happening is reasonable (and seems like the correct thing to do), but a comment explaining it would be nice.

stevedlawrence: Yeah, your reading of the code is correct. I'll add a coment to that affect and update our performance page to describe how one might use this new timer.

PR #319, opened 2020-02-03
Fixes some bugs, vulnerabilities and security hotspots

olabusayoT:
- fixes branches in conditional structure with same implementation
- removes double brace initialization
- replaces arr.toString() with 'new String(arr)'
- fixes unused return val
- fixes some indentation (2 spaces instead of 5/6)

DAFFODIL-2272


olabusayoT:
> +1 
> You mentioned there's a configuration to turn off code some smells. Does that config need to be included in the repo, or is it something that we do to the sonarcloud.io settings once we get it up and running?

Oh good point, I believe the config settings might be a 2 parter with a text properties file (specifically for multi-module projects), and some web ui configs. I've created https://issues.apache.org/jira/projects/DAFFODIL/issues/DAFFODIL-2275 to track setting up and configuring the UI. Once the account is set up, I'll know if we need to commit the properties file I think!

PR #323, opened 2020-02-11
Configure Sonarcloud.io

olabusayoT:
- sonarcloud workflow with compilation of code and tests, and sonarcloud
action to do the scanning and upload to site based on
sonar-scanner.properties specification
- properties file with organization, projectKey, module information etc
- we didn't configure for scoverage information as that is already provided by codecov
- test run can be found here: https://sonarcloud.io/dashboard?branch=daffodil-2275-sonarqube-config&id=olabusayoT_incubator-daffodil
- we're awaiting final configuration changes on sonarcloud.io that should reduce the amount of codesmells

DAFFODIL-2275

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,0 +1,49 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: SonarCloud Workflow
+on: [push]
+jobs:
+  sonarCloudTrigger:
+    name: SonarCloud Trigger
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        scala_version: [ '2.12.6', '2.11.12' ]
+    steps:
+
+      ############################################################
+      # Setup
+      ############################################################
+
+      - name: Checkout Repository
+        uses: actions/checkout@master
+
+      ############################################################
+      # Build & Scan
+      ############################################################
+
+      - name: Compile
+        run: $SBT compile test:compile it:compile
+        shell: bash
+        env:
+          SBT: sbt -J-Xms1024m -J-Xmx5120m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m ++${{ matrix.scala_version }} coverage
+
+      - name: SonarCloud Scan
+        uses: sonarsource/sonarcloud-github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,24 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+sonar.organization=apache
+sonar.projectKey=apache_inbubator-daffodil
+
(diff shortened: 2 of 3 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,49 @@
...
+name: SonarCloud Workflow
+on: [push, pull_request]
+jobs:
+  sonarCloudTrigger:
+    name: SonarCloud Trigger
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        scala_version: [ '2.12.6', '2.11.12' ]
```
stevedlawrence: Does the scala version actually matter? I would think if it's just doing static analysis it wouldn't be compiling anything and so wouldn't even need scala. Related, do we need the Compile step below?

sonar-project.properties
```diff
@@ -0,0 +1,32 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
stevedlawrence: Curious, can this file be made into a dot file so it can be hidden? Some systems (e.g. travis) support both hidden and non-hidden config files. If possible, I'd prefer hidden since this just clutters the root directory.

sonar-project.properties
```diff
@@ -0,0 +1,32 @@
...
+
+sonar.organization=apache
+sonar.projectKey=apache_inbubator-daffodil
+
+# --- optional properties ---
+
+# defaults to project key
+#sonar.projectName=My project
+# defaults to 'not provided'
+#sonar.projectVersion=1.0
```
stevedlawrence: Can we remove these commented out properties?

sonar-project.properties
```diff
@@ -0,0 +1,32 @@
...
+
+# Path is relative to the sonar-project.properties file. Defaults to .
+sonar.modules=daffodil-cli,daffodil-core,daffodil-io,daffodil-japi,daffodil-lib,daffodil-macro-lib,daffodil-propgen,daffodil-runtime1,daffodil-runtime1-unparser,daffodil-sapi,daffodil-tdml-lib,daffodil-tdml-processor,daffodil-test,daffodil-test-ibm1,daffodil-udf
+sonar.sources=src/main
+sonar.tests=src/it,src/test
+sonar.java.binaries=target/**/classes
+sonar.java.test.binaries=target/**/test-classes
+sonar.java.libraries=../lib_managed/**/*.jar
```
stevedlawrence: I guess maybe this is why you compile things? Does sonar cube do static analysis on jars/classes in addition to source?

stevedlawrence:
Anohther though, is it posible to have sonarcub do something if a pull request adds new bugs/vulnerabiliies? For example, codecov will sometimes add a comment to a PR. Can/is sonar cube configured to do something similar? Note that we probably only want it to complain if there are new issues, since we know there are going to be things like code smells that we aren't going to fix anytime soon.

mbeckerle:
The number of code smells: 800-ish, doesn't seem unreasonable to me. I looked at them, and while I don't see them as major flaws anywhere, they do seem like real coding-style complaints where we are not following naming conventions, or not following guidlines to document empty classes, etc. 

It's always best if you have this sort of tool and you set a policy of "no messages" from it. I.e., just like for the scala compiler where we have a no-warnings policy. 

These code-smells can all be fixed quite quickly. Yes there are 800 of them, so even at 30 seconds per that's a day of work. But since misery loves company, I'd suggest we all get in a room, race to fix them, and when they're all done, whoever fixed the most gets free lunch or something.

.github/workflows/sonarcloud.yml
olabusayoT: Yes, sonarqube (..cloud )depends on the files in target/{classes,test-classes} for analysis. Which is why we need the compile step, as well as the scala versions.

sonar-project.properties
olabusayoT: Unfortunately the github action doesn't support the hidden file option and explicitly requires it be named that. Sonar-scanner cmd provides the project.settings option, which they do not. We could fork the repo to allow us to use this property, but that opens a host of maintenance issues we mioght not want to deal with.

sonar-project.properties
olabusayoT: Yes, the binaries are required (according to their manula) But the libraries are optional, and it claims providing this gives it a fuller picture and not providing them might lead to imprecise data. I'll test locally to see how much of a different it makes towards results, but the compilation step is necessary for the scan to yield any results.

sonar-project.properties
olabusayoT: Done

olabusayoT:
@stevedlawrence  looks like there might be a way with PR decorations. I'm looking into it now

@mbeckerle I agree, I think the number now might not be reflective of our final desired state. I think our first step would be to come up with standard naming conventions and code style guide (i.e if they differ from Sonar's convention). Once that's updated on sonarcloud, we'd have an exact full picture of what it is we want. I think so far, all I've updated is to allow method names to have underscores in them e.g test_ but it's not entirely strict, as a method could also be do_Something and pass, which is not a desired outcome.

olabusayoT:
@stevedlawrence looks like they used to support inline comments, then dropped it in favor or Checks and the Sonarcloud UI, but enough people complained so they're looking into it (https://community.sonarsource.com/t/sonarcloud-github-pull-request-analysis-no-inline-comments/7371/35) Not sure how long it'd take to implement, but for not that feature is not supported.

sonar-project.properties
stevedlawrence: Definitelyl not worth forking, this is fine.

stevedlawrence:
According to that comment chain, it sounds like the sonarcloud check will fail if the PR adds new issues, so that should be sufficient. I guess we'll see what happens once it starts working.

stevedlawrence:
Unfortunately according to https://github.community/t5/GitHub-Actions/Allow-secrets-to-be-shared-with-trusted-Actions/td-p/34278 it sounds like secrets (asdie from GITHUB_TOKEN) are not made available during pull requests to prevent potentially leaking the secrets. We might only be able to trigger the scan when commits are merged, which is kindof unfortunate.

olabusayo:
You're right. That's quite unfortunate.

Sonarcloud doesn't give the option to manage the permissions of tokens, or to create local users, so exposure of SONAR_TOKEN could have undesirable security consequences. So until Github or Sonarqube figures a way around this, we'll only run on a scan on push (aka merge), and we can check the results then.

PR #324, opened 2020-02-14
DAFFODIL-1444 Schema Compiler Space/Speed

mbeckerle:
All schemas work, and compile MUCH faster now. 
    
    Instrumentation results of counting shared terms:
    
    These are compiler instrumentation output for runs of 
    the TestRefMap  testRefMapExplosion1 to testRefMapExplosion7
    
    These are compiler instrumentation output WITHOUT the new sharing of
    group members/substructure:
    
    [Schema had 0 elements.]
    [Schema memoized 0 shared terms.]
    [Schema had 0 elements.]
    [Schema memoized 0 shared terms.]
    [Schema had 31 elements.]
    [Schema memoized 4 shared terms.]
    [Schema had 63 elements.]
    [Schema memoized 5 shared terms.]
    [Schema had 127 elements.]
    [Schema memoized 6 shared terms.]
    [Schema had 255 elements.]
    [Schema memoized 7 shared terms.]
    [Schema had 511 elements.]
    [Schema memoized 8 shared terms.]
    
    This is WITH the new sharing enabled:
    
    [Schema had 0 elements.]
    [Schema memoized 0 shared terms.]
    [Schema had 0 elements.]
    [Schema memoized 0 shared terms.]
    [Schema had 9 elements.]
    [Schema memoized 4 shared terms.]
    [Schema had 11 elements.]
    [Schema memoized 5 shared terms.]
    [Schema had 13 elements.]
    [Schema memoized 6 shared terms.]
    [Schema had 15 elements.]
    [Schema memoized 7 shared terms.]
    [Schema had 17 elements.]
    [Schema memoized 8 shared terms.]
    
    
    You can see the exponential growth in number of elements is not detected
    any longer.
    
    There are still bugs to be fixed. Not all schemas work, and there are
    some unexpected test failures that have to be debugged. 
    
    Same output instrumentation for EDIFACT:
    
    WITHOUT new sharing:
    [Schema had 18820 elements.]
    [Schema memoized 690 shared terms.]
    
    WITH new sharing:
    
    [Schema had 1992 elements.]
    [Schema memoized 690 shared terms.]


daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -338,6 +339,8 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
    */
   def compileSource(schemaSource: DaffodilSchemaSource): ProcessorFactory =
     Compiler.synchronized {
+      Memoize1.reset
```
mbeckerle: remove. Not using Memoize1 anymore.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -364,7 +367,9 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
           log(LogLevel.Compile, "ProcessorFactory completed with no errors.")
         }
       }
-      log(LogLevel.Compile, "Schema had %s elements.", ElementBase.count)
+      log(LogLevel.Info, "Schema had %s elements.", ElementBase.count)
+
+      log(LogLevel.Info, "Schema memoized %s shared terms.", Memoize1.size)
```
mbeckerle: remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
```diff
@@ -200,6 +202,41 @@ abstract class AnnotatedSchemaComponentImpl(
...
+ * What we want is pure value based - depends on the properties and their values only.
+ */
+case class PropEnv(localProps: Seq[Set[(String, String)]], defaultPropSource: Seq[Set[(String, String)]], optNext: Option[scala.xml.Node]) {
+  //
+  // We depend on equality on PropEnv being set equality for localProps, pointer equality for defaultPropSource, and optNext.
+}
+
+object PropEnvFactory {
+  private val factory =
+    Memoize1((localProps: Seq[Set[(String, String)]]) =>
```
mbeckerle: can remove this. I have convinced myself that these are being compared for equality properly.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -107,11 +135,82 @@ trait Term
...
 
+  /**
+   * If two terms have the same propEnv, they have identical properties
+   * in scope.
+   */
+  lazy val propEnv: PropEnv = {
+    val localPropsSets = this.nonDefaultFormatChain.propertyPairsSets
+    val defaultPropObj = this.defaultFormatChain.propertyPairsSets
+    val next = refersToForPropertyCombining
+    PropEnvFactory(localPropsSets, defaultPropObj, next.map { _.xml })
```
mbeckerle: Can get rid of factory. Just construct the case class.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Memoize1.scala
```diff
@@ -52,6 +52,7 @@ class Memoize1[-T <: AnyRef, +R] private (f: T => R) extends (T => R) {
       case None => {
         val y = f(x)
         vals.put(x, y)
+        Memoize1.count += 1
```
mbeckerle: can revert all changes to Memoize1. Came in handy while experimenting, but code doesn't need to use it.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
```diff
@@ -61,7 +61,9 @@ trait ModelGroupGrammarMixin
         // there are no local delimiters.
         //
         ) {
-        val content = initiatorRegion ~ groupContent ~ terminatorRegion
+        val content = initiatorRegion ~ (
+          if (turnOnSharing) sharedModelGroup.groupContent
+          else groupContent) ~ terminatorRegion
         self match {
```
mbeckerle: The above is 1/2 the magic here. If turnOnSharing, then the groupContent grammar term is taken from the shared object, not the unique object. Hence it, and the parser/unparser generated by it, are shared.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -160,7 +162,15 @@ abstract class ModelGroup(index: Int)
...
 
-  def groupMembers: Seq[Term]
+  final def turnOnSharing = true
+
+  final def groupMembers: Seq[Term] =
+    if (turnOnSharing && this.isSharingATerm)
+      sharedModelGroup.groupMembers1
+    else
+      groupMembers1
+
```
mbeckerle: The above is 1/2 the magic. If turnOnSharing, and this term has a shared proto-term, then use that to get the group members. This eliminates all the combinatorial explosion of creating DSOM objects recursively.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -107,11 +135,82 @@ trait Term
...
+  }
+
+  lazy val shared: SharedTerm =
+    Term.sharedTermFactory(optionDef.xml, propEnv, makeSharedTerm(this))
+
+  /**
+   * True if this term is sharing a shared term. That is, this term
+   * is not the proto-term for this term + propenv combination.
+   */
+  lazy val isSharingATerm = this ne shared
```
bsloane1650: How could this boolean be useful? If we have two terms sharing a SharedTerm, as we are currently implemented, one of them is going to be the shared term itself; and we are not making any explicit decision on which term gets that honor.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
mbeckerle: Note to self: add commentary to Memoize1 about singleton and memory leaking.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -160,7 +162,15 @@ abstract class ModelGroup(index: Int)
       groupMembers.forall { _.hasStaticallyRequiredOccurrencesInDataRepresentation }
   }
 
-  def groupMembers: Seq[Term]
+  final def turnOnSharing = true
```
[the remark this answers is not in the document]
mbeckerle: Yes, as well as turn the Info messages back to Compile level.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
mbeckerle: It is actually essential currently to allowing us to distinguish when to use the regular method for groupMembers and traverse the child XML and create DSOM objects, vs. delegate and ask the shared thing to do that. 

However, I think I have a better way to do this than this general pair of Term and Term where one is the unique and the other is the shared ''proto term" but they can be confused. 

I think the way to go is to just memoize exactly the things we need. E.g., groupMembers, and groupContent. Those can be memoized per pair of (defXML, propEnv), which is what I'm memoizing the entire Term with right now. 

That will get us out of this flaky subtlety of two objects that both can do everything, but which to use for what, and no type-safety to help us know.

mbeckerle: Eliminated problematic isSharingATerm by eliminating term sharing. We now share exactly the things needed, and nothing else.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -130,6 +130,22 @@ object TermFactory {
...
+
+  private val f = new SharedFactory[SharedType]
+
+  /**
+   * Evaluates the argument only if passed a new key it has never seen before.
+   *
+   * If it has seen the key, it returns the
+   * initial sharedArg evaluation from the first time it encountered them.
+   */
+  final def sharedGroupMembersFactory(key: SharedFactory.ShareKey, sharedArg: => SharedType) =
```
bsloane1650: We're using a private type alias in the signature of a public function?

This wont cause any problems (as Scala considers them to be equivalent), but seems confusing.

In this case, it is not clear to me that aliasing the type to SharedType actually clarifies much to the reader; as both places it is used already have "shared" in their name.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -40,7 +40,63 @@ import org.apache.daffodil.xml.QName
...
 import org.apache.daffodil.dsom.ExpressionCompilers
 import org.apache.daffodil.runtime1.ElementBaseRuntime1Mixin
+import org.apache.daffodil.dsom.SchemaSet
+import org.apache.daffodil.dsom.SharedFactory
+import org.apache.daffodil.grammar.primitives.ConvertTextBooleanPrim
+import org.apache.daffodil.grammar.primitives.LeadingSkipRegion
+import org.apache.daffodil.grammar.primitives.SimpleNilOrValue
+import org.apache.daffodil.grammar.primitives.LiteralValueNilOfSpecifiedLength
 
+trait SharedSimpleValueFactoryMixin { sset: SchemaSet =>
```
bsloane1650: Didn't we define an almost identical trait earlier; with the only difference being SharedType differed.

Would it make more sense to have a common SharedTermFactoryMixin[A] trait that is paramiterized by the the type of term that is being shared? If we need seperate traits for each type (e.g. for pattern matching), we should be able to defined concrete sub-traits along the lines of:

trait SharedSimpleValueFactoryMixin extends SharedTermFactoryMixin[Gram]

mbeckerle: Yeah, I got rid of these. Now I just define members on SchemaSet, and use them appropriately in dsom or the grammar. No more mixins onto SchemaSet

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
mbeckerle: Yeah. I simplified this and got rid of these named members, or rather I jsut define them on SchemaSet directly rather than this mixin stuff.

mbeckerle:
Ok, everything doesn't work yet. The vmf_s2s schema passes on master, but not this branch. I get errors about paths that are ambiguous to an array an a non-array. 
Regular vmf works fine.
I also have test failures in 'sbt it:test'.

But this is getting really close.

mbeckerle:
I figured out the vmf_s2s problem. 
We check path expressions more thoroughly now, and one of the checks is that a path has to always refer to an array, or a non-array, not a mixture of them. vmf_s2s has a use of msi:FRI in a hand-crafted schema for a message. The FRI is not inside an array, but FRI uses a discriminator that refers to fn:count(..) which formula ".." has to refer to an array, and usually does. But in the one handcrafted message schema, ".." refers to a scalar. That mix is no longer allowed. By fixing the handcrafted schema to use FRI in the same idiom it was designed for, the problem goes away.

mbeckerle:
The 'sbt it:test' failures were spurious. I had JVM out of memory messages in that run also. A second run is clean, and Travis says all is working.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -160,7 +160,10 @@ abstract class ModelGroup(index: Int)
       groupMembers.forall { _.hasStaticallyRequiredOccurrencesInDataRepresentation }
   }
 
-  def groupMembers: Seq[Term]
+  final lazy val groupMembers: Seq[Term] =
+    schemaSet.sharedGroupMembersFactory.getShared(shareKey, groupMembers1)
```
stevedlawrence: It might be worth adding a comment around some of these parts, considering this is part of the "magic". For example, it seems important to point out that groupMembers1 (any chance a better name could be used) is a lazy val and that the ``getShared`` value parameter is passed by reference. If either of those changes I think this would still end up allocating some unnecessary things.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/PropProviders.scala
```diff
@@ -107,6 +115,9 @@ final class ChainPropProvider(leafProvidersArg: Seq[LeafPropProvider], forAnnota
 
   override def toString() = Misc.getNameFromClass(this) + "(" + forAnnotation + ")"
 
+  final lazy val propertyPairsSets: Seq[Set[(String, String)]] =
```
stevedlawrence: It's still not clear to me why this is a Seq[Set[...]]. Is this because properties can come from different locations (e.g. on the term, on a ref, on a dfdl:ref) and this captures the different properties from different locations? Why can't this be flattened?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
```diff
@@ -71,5 +73,7 @@ trait ModelGroupGrammarMixin
     finalContent
   }
 
+  private lazy val groupContent1 = schemaSet.sharedGroupContentsFactory.getShared(shareKey, groupContent)
```
stevedlawrence: Same question as with groupMembers1, can we find a better name?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Memoize1.scala
```diff
@@ -60,4 +61,11 @@ class Memoize1[-T <: AnyRef, +R] private (f: T => R) extends (T => R) {
 
 object Memoize1 {
   def apply[T <: AnyRef, R](f: T => R) = new Memoize1(f).apply _
+
+  private var count: Long = 0L
+
+  def size = count
+  def reset {
+    count = 0
+  }
```
stevedlawrence: Doesn't even look like we use Memoize except for in tests making sure it works. It's an interesting class, but we don't seem to need it? Thoughts on getting rid of it? It seems to have decent tests, so it's not a big deal if we keep it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
mbeckerle: I will add comments.

I will add documentation.

public groupMembers, 
protected groupMembersDef, 
and on the objects which aren't Terms, 
public groupMembersNotShared.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
mbeckerle: I am doing a few renames here, but in general this will be groupContent and protected groupContentDef following the same pattern as for ModelGroup.groupMembers.

The _content will become groupContentWithInitiatorTerminator since that's what is is.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Memoize1.scala
mbeckerle: While we are not using it, I did use it extensively while figuring this stuff out, and eventually removed it in favor of the SharedFactory idiom which is more purpose built, and so easier to understand. 

I will remove it from daffodil-lib however, and just try to keep in in some side-pony project for future use.

PR #327, opened 2020-02-27
updated dependency to 2.9.8 for jackson-core.

mbeckerle:
Doing work on the runtime 2 branch (daffodil-2202-initial-codegen) has been painful as it has a requirement for 2.9.8 of jackson-core.

Updating master branch to this makes keeping daffodil-2202-initial-codegen rebased to master much easier. 

DAFFODIL-2288

### DIFF
--- a/eclipse-projects/cli-test/.classpath
+++ b/eclipse-projects/cli-test/.classpath
@@ -43,9 +43,9 @@
       <attribute name="javadoc_location" value="jar:file:lib_managed/docs/org.codehaus.woodstox/stax2-api/stax2-api-4.1-javadoc.jar!/"/>
     </attributes>
   </classpathentry>
-  <classpathentry sourcepath="lib_managed/srcs/com.fasterxml.jackson.core/jackson-core/jackson-core-2.9.6-sources.jar" exported="true" kind="lib" path="lib_managed/bundles/com.fasterxml.jackson.core/jackson-core/jackson-core-2.9.6.jar">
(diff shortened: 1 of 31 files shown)


eclipse-projects/cli-test/.classpath
```diff
@@ -43,9 +43,9 @@
       <attribute name="javadoc_location" value="jar:file:lib_managed/docs/org.codehaus.woodstox/stax2-api/stax2-api-4.1-javadoc.jar!/"/>
     </attributes>
```
mbeckerle: The .classpath files are all changes created by the UpdateEclipseClasspath utility that maintains these files.

project/Dependencies.scala
```diff
@@ -34,7 +34,7 @@ object Dependencies {
   lazy val infoset = Seq(
     "org.jdom" % "jdom2" % "2.0.6",
     "com.fasterxml.woodstox" % "woodstox-core" % "5.1.0",
```
mbeckerle: Below is the only ACTUAL change in this PR.

stevedlawrence: Anytime we update a dependency, we also need to update the LICENSE and NOTICE files (and also validate that the dependencies license didn't change). In this case I think just ``daffodil-cli/bin.NOTICE`` needs to be updated since we only distribute jackson code with the CLI, and it's Apache licensed.

stevedlawrence:
Yeah, it's definitely not releated. The GitHub action is failing to checkout the branch:
not clear to me why GitHub actions is failing checkout the branch. Perhaps a force push will fix it.

> git checkout --progress --force 74473951f860bbb1299246028950b3ec68eab4a1
> ##[error]fatal: reference is not a tree: 74473951f860bbb1299246028950b3ec68eab4a1

That hash clearly exists (you can even click it in this comment). Hopefully a force push will fix it.

PR #328, opened 2020-03-01
DAFFODIL-2288 bump jackson version to 2.10.2

gezapeti:


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/JsonInfosetOutputter.scala
```diff
@@ -45,7 +45,7 @@ class JsonInfosetOutputter private (writer: java.io.Writer, pretty: Boolean, dum
   // is false).
   private val isFirstChildStack = MStackOfBoolean()
 
-  private val stringEncoder = BufferRecyclers.getJsonStringEncoder() // thread-safe instance of a string encoder
+  private val stringEncoder = JsonStringEncoder.getInstance()
```
gezapeti: Based on changes in Jackson, the JsonStringEncoder is thread safe after 2.10.

eclipse-projects/cli-test/.classpath
```diff
@@ -43,9 +43,9 @@
       <attribute name="javadoc_location" value="jar:file:lib_managed/docs/org.codehaus.woodstox/stax2-api/stax2-api-4.1-javadoc.jar!/"/>
     </attributes>
   </classpathentry>
-  <classpathentry sourcepath="lib_managed/srcs/com.fasterxml.jackson.core/jackson-core/jackson-core-2.9.8-sources.jar" exported="true" kind="lib" path="lib_managed/bundles/com.fasterxml.jackson.core/jackson-core/jackson-core-2.9.8.jar">
```
mbeckerle: Did you find the automated way to maintain these classpath files?
sbt updateClassifiers
followed by running the UpdateClasspathFiles in src/main/scala in the daffodil-propgen module?

project/Dependencies.scala
```diff
@@ -34,7 +34,7 @@ object Dependencies {
   lazy val infoset = Seq(
     "org.jdom" % "jdom2" % "2.0.6",
     "com.fasterxml.woodstox" % "woodstox-core" % "5.1.0",
-    "com.fasterxml.jackson.core" % "jackson-core" % "2.9.8"
+    "com.fasterxml.jackson.core" % "jackson-core" % "2.10.2"
```
mbeckerle: There is one more file to update which is bin.NOTICE in daffodil-cli. 
One must also verify that the license for jackson-core hasn't changed in the new version.

eclipse-projects/cli-test/.classpath
gezapeti: No, I've used sed and stuff to do the replacements.  :)

project/Dependencies.scala
gezapeti: I've updated bin.NOTICE: https://github.com/apache/incubator-daffodil/pull/328/files#diff-1f1ec1d6ff78037b56d3b4855b5e75e4R63 (probably not in the first patch, I've done a force push since)
The license haven't changed for 7 years: https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.10.2/src/main/resources/META-INF/LICENSE

eclipse-projects/cli-test/.classpath
mbeckerle: Great. For future reference, there is such a tool, but it's more important if new libraries are added or old ones removed. A version change, as you noticed, can be automated pretty easily. I did the last one with emacs tags query replace. :-)

PR #329, opened 2020-03-02
WIP: Add noPretty to disable pretty print for XML and JSON formats

gezapeti:
DAFFODIL-2234

I'm not sure this is the correct way to implement this option. I couldn't figure out a way to place it in a subcommand, so I went with a general command there.  No tests yet, I wanted to clarify if the path is correct.

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -307,6 +307,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
   val trace = opt[Boolean](descr = "run the debugger with verbose trace output")
   val verbose = tally(descr = "increment verbosity level, one level for each -v")
   val version = opt[Boolean](descr = "Show version of this program")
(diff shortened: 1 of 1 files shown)


stevedlawrence:
Potentially. DAFFODIL-1958 would be a really nice bug to resolve. The entire CLI is basically one giant function. Cleaning up the Scallop config is the first step in making that main function easier to update and manage. Definiately a worthy bug to work on.

stevedlawrence:
I think we at the least need to move the option out of the global options and into the parse options. Also, keep in mind that although it's a small, it could be considered part of the CLI "API", so changing things later potentially has issues with backwards compatibilty.

But as long as this is moved out of the global options I'm fine with this change.

gezapeti:
I've stopped working on this one and started with DAFFODIL-1958, but I find it hard to wrap my head around the logic of scallop - and working from home eats up all the free cycles I could spend on that. I'm fine with moving the option to parsing option and we can change things under the hood if the option name itself is good.

mbeckerle:
Closing as inactive. DAFFODIL-2234 links back here for future reference.

PR #331, opened 2020-03-09
Fix deep stack sizes when serializing some schemas

stevedlawrence:
The "parents" val in a DPathCompileInfo is a backpointer to all
DPathCompileInfo's that reference it. The problem with this is that when
elements are shared, these backpointers create a highly connected graph
that requires a large stack to serialize using the default java
serialization as it jumps around parents and children. To avoid this
large stack requirement, we make the parents backpointer transient. This
prevents jumping back up to parents during serialization and results in
only needing a stack depth relative to the schema depth. Once all that
serialization is completed and all the DPathCompileInfo's are
serialized, we then manually traverse all the DPathCompileInfo's again
and serialize the parent sequences (via the serailizeParents method).
Because all the DPathCompileInfo's are already serialized, this just
serializes the Sequence objects and the stack depth is again relative to
the schema depth.

On complex schemas, this saw an order of magnitude reduction in stack
size during serialization.

DAFFODIL-2283

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
@@ -212,7 +212,47 @@ class DPathCompileInfo(
   extends ImplementsThrowsSDE with PreSerialization
   with HasSchemaFileLocation {
 
-  lazy val parents = parentsArg
+
+  /**
+   * This "parents" val is a backpointer to all DPathCompileInfo's that
+   * reference this DPathCompileInfo. The problem with this is that when
+   * elements are shared, these backpointers create a highly connected graph
+   * that requires a large stack to serialize using the default java
+   * serialization as it jumps around parents and children. To avoid this large
+   * stack requirement, we make the parents backpointer transient. This
+   * prevents jumping back up to parents during serialization and results in
+   * only needing a stack depth relative to the schema depth. Once all that
+   * serialization is completed and all the DPathCompileInfo's are serialized,
+   * we then manually traverse all the DPathCompileInfo's again and serialize
+   * the parent sequences (via the serailizeParents method). Because all the
+   * DPathCompileInfo's are already serialized, this just serializes the
+   * Sequence objects and the stack depth is again relative to the schema
+   * depth.
+   */
+  @transient
+  val parents = parentsArg
+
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -212,7 +212,48 @@ class DPathCompileInfo(
...
+   * only needing a stack depth relative to the schema depth. Once all that
+   * serialization is completed and all the DPathCompileInfo's are serialized,
+   * we then manually traverse all the DPathCompileInfo's again and serialize
+   * the parent sequences (via the serailizeParents method). Because all the
+   * DPathCompileInfo's are already serialized, this just serializes the
+   * Sequence objects and the stack depth is again relative to the schema
+   * depth.
+   */
+  @transient
   lazy val parents = parentsArg
```
mbeckerle: lazy val no longer needed as parentsArg is not a by-name argument. Could just rename the arg to parents.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -301,6 +341,19 @@ class DPathElementCompileInfo(
     unqualifiedPathStepPolicy,
     typeCalcMap, lexicalContextRuntimeData) {
 
+  override
```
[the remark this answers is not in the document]
stevedlawrence: Yeah,not sure what I was thinking here. Fixed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
stevedlawrence: Correct. Will fix.

stevedlawrence: Note that there was some issue with ``@TransientParam`` not working as expected if I just renamed the parameter to ``parents`` and made it a ``val``. Change this  from ``lazy val`` to ``val`` does work as expected though.

stevedlawrence: Looks like val's in a constructor parameter cannot be made transient. Keeping the parentsArg and using a lazy val I think is the normal solution to this kindof thing.

PR #332, opened 2020-03-09
Refactor isHidden

olabusayoT:
- Removes isHidden from ElementRuntimeData
- Adds isHidden flag to ModelGroup/ModelGroupRuntimeData
- Adds private hiddenDepth counter (and associated functions)
increment/decrement logic (HiddenGroupCombinators) for PState/UState
- overrides increment/decrement functions in UStateForSuspension to
fail on UsageError as these have no place in that context
- Adds check to verify  we are out of hidden state at the end of
parsing/unparsing
- Adds isHidden check to ChoiceGroupRef
- Adds isHidden/setHidden logic to InfosetElement/DIElement
- Adds tests for group defs used in both hidden/non-hidden sequence &
choice contexts
-- adds "info hidden" to debugger with tests
-- moves InfoOccursIndex to maintain alphabetic ordering of objects

DAFFODIL-2281

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
@@ -899,6 +899,207 @@ class TestCLIdebugger {
     }
   }
 
+  @Test def test_CLI_Debugger_InfoHidden_1() {
+    val schemaFile = Util.daffodilPath(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequencesWithHiddenRefs.dfdl.xsd")
+    val inputFile = Util.newTempFile("testInput_", ".tmp", optFileContents = Some("2~3"))
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) {
+      (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile.getAbsolutePath))
+    } else {
+      (schemaFile, inputFile)
+    }
+
+    val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
+
+    try {
+      val cmd = String.format("%s -d parse -s %s -r e5 %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("(debug)"))
+
+      shell.sendLine("break f")
+      shell.sendLine("display info hidden")
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("<f>2</f>"))
+      shell.sendLine("quit")
+    } finally {
+      shell.close()
+    }
+  }
+
+  @Test def test_CLI_Debugger_InfoHidden_2() {
+    val schemaFile = Util.daffodilPath(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequencesWithHiddenRefs.dfdl.xsd")
+    val inputFile = Util.newTempFile("testInput_", ".tmp", optFileContents = Some("2~3"))
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) {
+      (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile.getAbsolutePath))
+    } else {
+      (schemaFile, inputFile)
+    }
+
+    val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
+
+    try {
+      val cmd = String.format("%s -d parse -s %s -r e4 %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("(debug)"))
+
+      shell.sendLine("break f")
+      shell.sendLine("display info hidden")
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("<f>3</f>"))
+      shell.sendLine("quit")
+    } finally {
+      shell.close()
+    }
+  }
+
+  @Test def test_CLI_Debugger_InfoHidden_3() {
+    val schemaFile = Util.daffodilPath(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/ChoicesInHiddenContexts.dfdl.xsd")
+    val inputFile = Util.newTempFile("testInput_", ".tmp", optFileContents = Some("2,3"))
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) {
+      (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile.getAbsolutePath))
+    } else {
+      (schemaFile, inputFile)
+    }
+
+    val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
+
+    try {
+      val cmd = String.format("%s -d parse -s %s -r e8 %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("(debug)"))
+
+      shell.sendLine("break a")
+      shell.sendLine("break h")
+      shell.sendLine("break g")
+      shell.sendLine("break e")
+      shell.sendLine("break f")
+      shell.sendLine("display info hidden")
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("<a>2</a>"))
+      shell.expect(contains("<g />"))
+    } finally {
+      shell.close()
+    }
+  }
+
+  @Test def test_CLI_Debugger_InfoHidden_4() {
+    val schemaFile = Util.daffodilPath(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/ChoicesInHiddenContexts.dfdl.xsd")
+    val inputFile = Util.newTempFile("testInput_", ".tmp", optFileContents = Some("[6~]9"))
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) {
+      (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile.getAbsolutePath))
+    } else {
+      (schemaFile, inputFile)
+    }
+
+    val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
+
+    try {
+      val cmd = String.format("%s -d parse -s %s -r e9 %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("(debug)"))
+
+      shell.sendLine("break e")
+      shell.sendLine("break f")
+      shell.sendLine("break g")
+      shell.sendLine("break h")
+      shell.sendLine("break i")
+      shell.sendLine("display info path hidden")
+
+      shell.sendLine("continue")
+      shell.expect(contains(":f"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":i"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":h"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":e"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":f"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":f"))
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":g"))
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":i"))
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":h"))
+      shell.expect(contains("hidden: false"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":e"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains(":f"))
+      shell.expect(contains("hidden: true"))
+
+      shell.sendLine("continue")
+      shell.expect(contains("<h />"))
+    } finally {
+      shell.close()
+    }
+  }
   /* See DFDL-1264
   @Test def test_3585_CLI_Debugger_simpleDebugger_unparse() {
     val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchema.dfdl.xsd")
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/GroupRef.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/GroupRef.scala
@@ -30,6 +30,11 @@ trait GroupRef { self: ModelGroup =>
 
   final def referredToComponent = groupDef
 
+  /**
+   * Override in sequenceGroupRef and choiceGroupRef for hidden groups
+   */
+  def isHidden: Boolean
+
   final override lazy val optReferredToComponent = Some(referredToComponent)
 
   final override protected lazy val groupMembersDef = LV('groupMembers) {
@@ -101,14 +106,11 @@ final class SequenceGroupRef(
   refXML: Node,
   refLexicalParent: SchemaComponent,
   positionArg: Int,
-  isHiddenArg: Boolean)
+  override val isHidden: Boolean)
   extends SequenceGroupTermBase(refXML, refLexicalParent, positionArg)
   with GroupRef {
 
   private lazy val globalGroupDef = globalGroupDefArg // once only
-
-  override def isHidden = isHiddenArg
-  override def isHiddenGroupRef = isHiddenArg
 
   private lazy val sgd = groupDef.asInstanceOf[GlobalSequenceGroupDef]
 
@@ -129,7 +131,7 @@ final class ChoiceGroupRef(
   refXML: Node,
   refLexicalParent: SchemaComponent,
   positionArg: Int,
-  isHiddenArg: Boolean)
+  override val isHidden: Boolean)
   extends ChoiceTermBase(refXML, Option(refLexicalParent), positionArg)
   with GroupRef {
 
@@ -137,10 +139,7 @@ final class ChoiceGroupRef(
 
   private lazy val globalGroupDef = globalGroupDefArg // once only
 
-  override def isHidden = isHiddenArg
-  override def isHiddenGroupRef = isHiddenArg
-
-  override lazy val groupDef = globalGroupDef
+  override lazy val groupDef = LV('cgrGroupDef){ globalGroupDef }.value
 
   private lazy val cgd = groupDef.asInstanceOf[GlobalChoiceGroupDef]
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
@@ -115,21 +115,6 @@ trait SchemaComponent
     schemaSet.variableMap
   }.value
 
-  /**
-   * Whether the component is hidden.
-   *
-   * Override this in the components that can hide - SequenceGroupRef
-   * and ChoiceGroupRef
-   */
-  def isHidden: Boolean = isHiddenLV
-
-  private lazy val isHiddenLV = {
-    val optEC = enclosingTerms.headOption // FIXME: broken if there are hidden and non-hidden uses
-    optEC match {
-      case None => false
-      case Some(ec) => ec.isHidden
-    }
-  }
 
   lazy val schemaComponent: LookupLocation = this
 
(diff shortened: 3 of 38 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/HiddenGroupCombinatorUnparser.scala
```diff
@@ -0,0 +1,44 @@
...
+  override lazy val runtimeDependencies = Vector()
+
+  def unparse(start: UState): Unit = {
+    start.incrementHiddenDef
+
+    // unparse
+    bodyUnparser.unparse1(start)
+
+    start.decrementHiddenDef
+  }
```
stevedlawrence: We probably want to wrap this in a a try/finally so that the hiddenef count always decremented even if the bodyUnparse throws some exception.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SequenceTermRuntime1Mixin.scala
```diff
@@ -76,7 +77,8 @@ trait ChoiceBranchImpliedSequenceRuntime1Mixin { self: ChoiceBranchImpliedSequen
       None,
       Maybe.Nope,
       Maybe.Nope,
-      Maybe.Nope)
+      Maybe.Nope,
+      isHidden)
```
mbeckerle: This should always be false here. I.e., a ChoiceBranchImpliedSequence cannot be created for a group ref. So you can write
```
{
Assert.invariant(!isHidden)
false
}
```

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ChoiceAndOtherVariousUnparsers.scala
```diff
@@ -29,69 +29,62 @@ import org.apache.daffodil.util.MaybeInt
...
+  optDefaultableBranchUnparser: Option[Unparser],
+  firstBranchUnparser: Unparser)
   extends CombinatorUnparser(mgrd)
   with ToBriefXMLImpl {
   override def nom = "Choice"
 
   override lazy val runtimeDependencies = Vector()
 
-  override lazy val childProcessors = eventUnparserMap.map { case (k, v) => v }.toSeq.toVector
+  override lazy val childProcessors = (eventUnparserMap.map { case (k, v) => v }.toSeq ++ optDefaultableBranchUnparser).toVector
```
mbeckerle: line too long. Please break at the ++

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HiddenGroupCombinatorParser.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  override lazy val runtimeDependencies = Vector()
+
+  def parse(start: PState): Unit = {
+    start.incrementHiddenDef
+
+    // parse
+    bodyParser.parse1(start)
+
+    start.decrementHiddenDef
```
stevedlawrence: Same here, probably need a try/finally.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1497,6 +1482,17 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
         }
       }
 
+      object InfoHidden extends DebugCommand with DebugCommandValidateZeroArgs {
+        val name = "hidden"
+        override lazy val short = "hid"
+        val desc = "display whether or not we're within a hidden state"
```
mbeckerle: "... within a hidden group." would better match DFDL spec terminology.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
```diff
@@ -68,9 +69,16 @@ trait ModelGroupGrammarMixin
...
           case s: SequenceTermBase => DelimiterStackCombinatorSequence(s, content)
         }
-      } else { groupContentDef }
+      } else {
+        groupContent
+      } //TODO: confirm this works; changed from groupContentDef
 
-    finalContent
+      val mgrd = self.modelGroupRuntimeData
+      if (mgrd.isHidden) {
```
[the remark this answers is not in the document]
mbeckerle: Good point. This is the schema compiler, so we shouldn't be asking the runtime data objects, but the DSOM object. 
The schema compiler creates the runtime data objects, but shouldn't use them. What they contain is in service of the runtime. 

Interestingly, it's possible that ModelGroupRuntimeData doesn't need an isHidden member. Whether the parser/unparser is traversing a hidden group ref or not is represented by way of there being a HiddenGroupParser/Unparser constructed. So possibly we don't even need ModelGroupRuntimeData.isHidden.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -732,9 +728,8 @@ trait Term
   final def possibleFirstChildElementsInInfoset: Seq[ElementBase] = LV('possibleFirstChildElementsInInfoset) {
     val pfct = possibleFirstChildTerms
     val firstChildren = pfct.flatMap {
-      case e: ElementBase if e.isHidden => Nil
       case e: ElementBase => Seq(e)
-      case s: SequenceTermBase if s.isHidden => Nil
+      case gr: GroupRef if gr.asModelGroup.isHidden => Nil
```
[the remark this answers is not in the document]
stevedlawrence: Just curious how difficult it would be to move isHidden (or isHiddenGroupRef, agreed that's amore clear name) to GroupRef? Now that you mention it, that does feel like that's maybe a more natural place for this property.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -746,7 +741,7 @@ trait Term
    *
    * What stops this is when the end of an enclosing element has to be next.
    */
-  final def possibleNextChildElementsInInfoset: Seq[ElementBase] = LV('possibleNextChildElementsInInfoset) {
+  def possibleNextChildElementsInInfoset: Seq[ElementBase] = LV('possibleNextChildElementsInInfoset) {
```
stevedlawrence: Any particular reason final was removed? It doesn't look like this patch changes anything to override this function.

olabusayoT: I think this was an oversight. I'd changed it to a val during debugging, and forgot the "final" during restoration

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1497,6 +1482,17 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
         }
       }
 
+      object InfoHidden extends DebugCommand with DebugCommandValidateZeroArgs {
+        val name = "hidden"
+        override lazy val short = "hid"
```
stevedlawrence: I think if you don't define ``short`` then it just defaults to the first letter of the name, which in this case is "h" and is a reasonable short form for ``hidden``. Nothing else starts with h.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1231,21 +1231,6 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
         }
       }
 
-      object InfoOccursIndex extends DebugCommand with DebugCommandValidateZeroArgs {
```
stevedlawrence: While we're reording things, it might be worth it to reorder the ``info`` subcommands Seq so they're alphabetical, and to make it so it's not on one ginormous line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessorStateBases.scala
```diff
@@ -480,6 +480,17 @@ abstract class ParseOrUnparseState protected (
...
+  private var _hiddenDepth = 0
+
+  def incrementHiddenDef = {
+    _hiddenDepth += 1
+  }
+  def decrementHiddenDef = {
+    _hiddenDepth -= 1
+  }
+
+  def withinHiddenNest: Boolean = _hiddenDepth > 0
```
stevedlawrence: Related to what Mike mentioned in the debugger, changing this variable to ``withinHiddenGroup`` is maybe more consistent with the DFDL spec.

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequencesWithHiddenRefs.dfdl.xsd
```diff
@@ -0,0 +1,94 @@
...
+  </xs:element>
+  <xs:element name="e5" dfdl:lengthKind="delimited">
+    <xs:complexType>
+      <xs:sequence dfdl:separator="~">
+        <xs:group ref="ex:s2" />
+        <xs:sequence dfdl:hiddenGroupRef="ex:s2" />
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+</xs:schema>
```
stevedlawrence: It might also be useful to have a test where a hidden group ref contains another hidden group ref just to ensure that nesting works as expected.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessorStateBases.scala
mbeckerle: The word "nest" here is important. The concept here is that the runtime is beneath/inside a hidden group, but not directly/statically, but anywhere within the dynamic context nest. 

I'm open to other names, but the goal is to denote the dynamic extent of the hidden group.

stevedlawrence: Fair enough. Should we modify the description of "info hidden" to mention "...within the nesting context of a hidden group" or something like that to be consistent, or is that distinction to verbose/excessive for the debugger?

mbeckerle: That would be good to do.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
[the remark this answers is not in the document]
olabusayoT: After looking a tthe code, I realized we didn't actually need ModelGroupRuntimeData.isHidden, nor ModelGroup.isHidden (this was moved into GroupRef)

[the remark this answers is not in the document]
mbeckerle: :+1:

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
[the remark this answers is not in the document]
olabusayoT: I decided to move isHidden to GroupRef, as isHidden inherited from MG was only ever possibly true for these cases.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SequenceTermRuntime1Mixin.scala
olabusayoT: The Runtime structures no longer have an isHidden member

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -217,27 +217,37 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
...
+        // will never see events for any of the branches. Instead, we will just
+        // always pick the branch in which every thing is defaultble or OVC.
+        val defaultableBranchUnparser = alternatives.find(_.context.runtimeData =:= defaultableBranchRD).get.unparser
+        Some(defaultableBranchUnparser)
+      } else {
+        ch match {
+          case cgr: GroupRef if cgr.isHidden => {
+            /*
+             * This is a requirement for terms inside a hidden group. If we're a hidden group ref, we know
+             * our descendants need this requirements, so it's a SDE if they don't meet it.
```
mbeckerle: "need this requirements"... ambiguous. Did you mean "meets this requirement" or "requires this" ?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
```diff
@@ -257,9 +267,8 @@ case class ChoiceCombinator(ch: ChoiceTermBase, alternatives: Seq[Gram])
     if (mapValues.isEmpty)
       new NadaUnparser(null)
     else {
-      val firstBranchUnparser = alternatives(0).unparser
       new ChoiceCombinatorUnparser(ch.modelGroupRuntimeData, eventUnparserMap,
-        choiceLengthInBits, optDefaultableBranchUnparser, firstBranchUnparser)
+        choiceLengthInBits, optDefaultableBranchUnparser)
```
mbeckerle: Before committing this, we really must test mil-std-2045 and vmf. These have been turning up many problems related to choices for me.

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -900,10 +900,14 @@ class TestCLIdebugger {
   }
 
   @Test def test_CLI_Debugger_InfoHidden_1() {
-    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequencesWithHiddenRefs.dfdl.xsd")
-    val inputFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/test_isHidden1.txt")
-    val (testSchemaFile, testInputFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile)) else (schemaFile, inputFile)
-
+    val schemaFile = Util.daffodilPath(
```
mbeckerle: Really bugs me that every one of these tests (not just your new ones) has this repeated isWindows conditionalization. 

We really should be able to abstract this away so that tests don't have this baggage over and over. 

These same 6 lines of code are repeated a lot. 

```
    val (testSchemaFile, testInputFile) = if (Util.isWindows) {
      (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile.getAbsolutePath))
    } else {
      (schemaFile, inputFile)
   } 
     val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
stevedlawrence: Doesn't look like this was resolved.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
olabusayoT: I meant "require this", it's been update to clarify

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -243,12 +227,12 @@ trait SchemaComponent
         case ct: ComplexTypeBase => "ct"
         case st: SimpleTypeDefBase => "st=" + st.namedQName.toQNameString
         case st: SimpleTypeBase => "st=" + st.primType.globalQName.toQNameString
-        case cgr: ChoiceGroupRef => "cgr" + (if (cgr.position > 1) cgr.position else "") + "=" + cgr.groupDef.namedQName.toQNameString
+        case cgr: ChoiceGroupRef => "cgr" + (if (cgr.isHidden) "h" else "") + (if (cgr.position > 1) cgr.position else "") + "=" + cgr.groupDef.namedQName.toQNameString
```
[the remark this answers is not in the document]
olabusayoT: So there is a difference b/w the toString and the toQNameString, the former considers the namespace in the case of no prefix, and will output '{ns}local' instead of the latter's 'local', but it doesn't look like there are any other differences beyond that. I'll update to 'toQNameString" so the same bhv is maintained across the cases.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ChoiceCombinator.scala
olabusayoT: Both have been tested and are passing

mbeckerle: +1 then.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -578,14 +563,13 @@ trait Term
   protected def possibleFirstChildTerms: Seq[Term]
```
mbeckerle: Unused.  Can be deleted.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/GroupRef.scala
```diff
@@ -129,18 +131,15 @@ final class ChoiceGroupRef(
...
 
   requiredEvaluationsIfActivated(groupDef)
 
   private lazy val globalGroupDef = globalGroupDefArg // once only
 
-  override def isHidden = isHiddenArg
-  override def isHiddenGroupRef = isHiddenArg
-
-  override lazy val groupDef = globalGroupDef
+  override lazy val groupDef = LV('cgrGroupDef){ globalGroupDef }.value
```
stevedlawrence: Is it important for the symbol used in the LV match the variable name, maybe for debugging or diagnostic purposes?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
```diff
@@ -94,16 +94,19 @@ abstract class Gram(contextArg: SchemaComponent)
    * Note: This should not evaluate the argument unless it has to.
    */
   def ~(qq: => Gram) = {
-    lazy val q = qq.deref
+    val q = qq.deref
```
stevedlawrence: The comment above says this shouldn't evaluate qq unless it needs to. But this looks like it will always evaluate it now. But maybe it always did that regardless?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
```diff
@@ -94,16 +94,19 @@ abstract class Gram(contextArg: SchemaComponent)
...
+        if (q.isEmpty) {
           EmptyGram
-        else q
-      } else if (q.isEmpty) self
-      else {
+        } else {
+          q
+        }
+      } else if (q.isEmpty) {
+        self
```
stevedlawrence: I think this is the only change here? I'm not sure I understandt this change. What issue does this fix?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -96,20 +91,23 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
       // get recomputed, but it can appear with a stale value (left-over from parse)
       // in the arriving infoset for unparse.
       //
-      val optDefaultBranch = {
-
+      val optDefaultBranchIfHidden = {
```
mbeckerle: Naming. The "ifHidden" suffix is misleading. This is the default branch, hidden or not.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -135,6 +135,9 @@ class TypeValueCalcUnparser(typeCalculator: TypeCalculator, repTypeUnparser: Unp
 
     val origInfosetElement = ustate.currentInfosetNode
     val tmpInfosetElement = Infoset.newElement(repTypeRuntimeData).asInstanceOf[DISimple]
+    if (ustate.withinHiddenNest)
```
mbeckerle: It's possible you don't need this. These temp elements are transient for repType. But it doesn't hurt, and it is consistent that these be marked hidden. So I'd leave it in.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -76,12 +76,8 @@ trait NextElementResolver { self: InfosetInputter =>
         // set of possible elements.
         //
         maybeERD = resolver.maybeNextElement(name, nameSpace, hasNamespace)
-        //
-        // If the ERD is marked hidden, then we never expect any events corresponding
-        // to its elements.
-        //
-        if (maybeERD.isDefined && maybeERD.get.isHidden)
-          maybeERD = Nope // clobber it back to Nope
+        // we address the cases where ERD is in a hidden context where the elements are assigned
```
mbeckerle: Rephrase. "The cases where .... are addressed where..." 
I had a hard time following the comment in the current wording.

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequenceGroup.tdml
```diff
@@ -1406,9 +1406,9 @@
       <tdml:documentPart type="text"><![CDATA[145,2]]></tdml:documentPart>
     </tdml:document>
     <tdml:errors>
-      <tdml:error>choice</tdml:error>
-      <tdml:error>hidden group</tdml:error>
-      <tdml:error>outputValueCalc</tdml:error>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>hidden choice</tdml:error>
+      <tdml:error>must be fully defaultable or define dfdl:outputValueCalc</tdml:error>
```
mbeckerle: Better to be less specific about the error messages so that minor changes to them may not break the test.
I would just have the error checks look for "defaultable" and "outputValueCalc" somewhere in the message.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/GroupRef.scala
olabusayoT: There's a comment under LV, where it says 2 symbols with the same name could cause issues that tend to be difficult to debug, so I named it something different since SGR's groupDef already has a groupDef LV. I also found some precedence where the symbol didn't match the variable name (such as in org/apache/daffodil/dsom/SchemaDocIncludesAndImportsMixin.scala:151) so I figured it'd be ok

daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala
olabusayoT: Ah that was an oversight, from when i was trying to debug the dropped HGR. Will fix

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -578,14 +563,13 @@ trait Term
...
 
   /*
-     * Returns list of Elements that could be the first child in the infoset of this model group or element.
-     */
+   * Returns list of Elements that could be the first child in the infoset of this model group or element.
+   */
   final lazy val possibleFirstChildElementsInInfoset: Seq[ElementBase] = LV('possibleFirstChildElementsInInfoset) {
```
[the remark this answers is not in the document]
olabusayoT: I've created DAFFODIL-2305 to clean up these elements and some of the tests can't be cleanly deleted, so this can be merged separately.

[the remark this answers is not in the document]
mbeckerle: Good idea.

PR #334, opened 2020-03-20
Add a special unparser for empty choice branches

stevedlawrence:
When unparsing a choice, it is possible that the default branch to
unparsing could be completely empty (e.g. a sequence with a parse only
assertion). In these cases, that empty branch optimizes to a
NadaUnparser. Normally NadaUnparser's are optimized out or else they
will throw an error at runtime. However, the choice combinator unparse
still expects to have an Unparser to unparse, even if it does nothing.
To support this, this creates an empty choice branch unparser that does
nothing, but allows for the choice combinator to have something to
unparse.

DAFFODIL-2296

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice-unparse2.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice-unparse2.tdml
@@ -216,4 +216,35 @@
     </tdml:infoset>
   </tdml:unparserTestCase>
 
+  <tdml:defineSchema name="c3" elementFormDefault="unqualified">
+
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+
+    <xs:element name="choice" dfdl:terminator=";">
+      <xs:complexType>
+        <xs:choice>
+          <xs:sequence>
+            <xs:annotation>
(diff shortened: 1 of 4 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice-unparse2.tdml
```diff
@@ -216,4 +216,35 @@
...
+
+  </tdml:defineSchema>
+
+  <tdml:unparserTestCase name="choice_default_branch_is_empty" model="c3" root="choice">
+    <tdml:document>;</tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:choice />
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
```
olabusayoT: Should this be switched to match the convention of input then output?

stevedlawrence: That's a good question. I'm not sure what our convention is, or if we have one. The other unparserTestCases in this file have the document first followed by the infoset, even though that's sort of backwards than one might suspect. I am able to find some tdml files that have the infoset first for unparserTestCases, but I think most of them follow this pattern. 

I think my inclination is to keep consistent with the rest of the file, since it doesn't seem we really have an enfoced convention, but I don't feel strongly.

olabusayoT: Ok, that makes sense!

olabusayoT:
+1

PR #339, opened 2020-03-26
Switch completely to GitHub actions for CI

stevedlawrence:
- TravisCI fairly frequently has random failures not related to
  Daffodil. And even when it doesn't fail, it often takes a long time to
  run. It also doesn't support Windows. Let's just remove it.
- Azure was more stable, and supports Windows, but Apache does not
  support Azure, and some tests still don't work on Windows. Since this
  isn't actively tested, remove it.
- GitHub actions has all the features we needs, supports both Windows
  and Linux and seems to be more stable as far as random failures go. So
  completely switch to GitHub actions. Also enable code coverage
  reporting in GitHub Actions, which was previously only done by
  TravisCI. Also remove duplication of SBT environment variable now
  that GitHub actions supports that

DAFFODIL-2307


stevedlawrence:
According to https://codecov.io/gh/apache/incubator-daffodil/pull/339 it looks like the coverage went down a little with this change. Not sure what's different though...will investigate

PR #343, opened 2020-03-28
Daffodil 2302 fixes external variables API difficulties

mbeckerle:
Turned into a large change set.

There is a functional change here. External variables are no longer saved when we save a binary compiled object. They are irrelevant to schema compilation. They are applied just before runtime. 

DAFFODIL-2302

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -113,7 +113,13 @@ lazy val commonSettings = Seq(
     "-language:experimental.macros",
     "-unchecked",
     "-Xfatal-warnings",
-    "-Xxml:-coalescing",
+    "-Xxml:-coalescing"
+  ),
+  // Workaround issue that some options are valid for javac, not javadoc.
+  // These javacOptions are for code compilation only. (Issue sbt/sbt#355)
+  javacOptions in Compile in compile ++= Seq(
+    "-Werror",
+    "-Xlint:deprecation"
   ),
   logBuffered := true,
   transitiveClassifiers := Seq("sources", "javadoc"),
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
@@ -103,7 +103,7 @@ class TestCLISaveParser {
       val saveCmd = String.format("%s save-parser -s %s -r row2 -c %s %s", Util.binPath, testSchemaFile, testConfigFile, savedParserFile.getName())
       shell.sendLine(saveCmd)
 
-      val cmd = String.format("echo 0,1,2| %s parse --parser %s", Util.binPath, savedParserFile.getName())
+      val cmd = String.format("echo 0,1,2| %s parse --parser %s -c %s", Util.binPath, savedParserFile.getName(), testConfigFile)
       shell.sendLine(cmd)
       shell.expect(contains(output12))
 
@@ -262,6 +262,10 @@ class TestCLISaveParser {
     }
   }
 
+  /**
+   * Note that in Daffodil 2.6.0 behavior of external variables changed. They are not saved as part of binary
+   * compiling. They are a runtime-thing only.
+   */
   @Test def test_3508_CLI_Saving_SaveParser_extVars() {
 
     val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section07/external_variables/external_variables.dfdl.xsd")
@@ -273,13 +277,13 @@ class TestCLISaveParser {
     val shell = Util.startIncludeErrors("")
 
     try {
-      var cmd = String.format("%s -v save-parser -s %s -r row2 -D\"{http://example.com}var1=25\" \"{http://example.com}var3=7\" %s", Util.binPath, testSchemaFile, savedParser)
+      var cmd = String.format("%s -v save-parser -s %s -r row2  %s", Util.binPath, testSchemaFile, savedParser)
       shell.sendLine(cmd)
       shell.expectIn(1, (contains("[info] Time (saving)")))
       assertTrue("save-parser failed", parserFile.exists())
 
       shell.sendLine();
-      cmd = String.format("echo 0| %s parse --parser %s\n", Util.binPath, savedParser)
+      cmd = String.format("""echo 0| %s parse --parser %s -D"{http://example.com}var1=25" "{http://example.com}var3=7" """, Util.binPath, savedParser)
       shell.sendLine(cmd)
       shell.expect(contains("<tns:row2 xmlns:tns=\"http://example.com\">"))
       shell.expect(contains("<cell>25</cell>"))
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -32,33 +32,35 @@ import java.nio.file.Paths
 import java.net.URI
 import java.util.Scanner
 
-import scala.xml.SAXParseException
+import scala.xml.{SAXParseException, Node}
 import org.rogach.scallop
-import org.apache.daffodil.debugger.{ InteractiveDebugger, TraceDebuggerRunner, CLIDebuggerRunner }
+import org.apache.daffodil.debugger.{InteractiveDebugger, CLIDebuggerRunner, TraceDebuggerRunner}
 import org.apache.daffodil.util.Misc
 import org.apache.daffodil.util.Timer
 import org.apache.daffodil.xml._
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.compiler.Compiler
-import org.apache.daffodil.api.WithDiagnostics
+import org.apache.daffodil.api.{WithDiagnostics, URISchemaSource, ValidationMode, DFDL, DaffodilTunables}
 import org.apache.daffodil.util.Logging
 import org.apache.daffodil.util.LogLevel
 import org.apache.daffodil.util.LogWriter
 import org.apache.daffodil.util.LoggingDefaults
 import org.apache.daffodil.exceptions.NotYetImplementedException
 import java.io.File
+
 import org.apache.daffodil.tdml.DFDLTestSuite
-import org.apache.daffodil.api.ValidationMode
-import scala.xml.Node
-import org.apache.daffodil.externalvars.{ Binding, BindingException }
+import org.apache.daffodil.externalvars.{Binding, BindingException}
 import org.apache.daffodil.externalvars.ExternalVariablesLoader
 import org.apache.daffodil.configuration.ConfigurationLoader
 import org.apache.daffodil.api.ValidationMode
+
 import scala.language.reflectiveCalls
 import scala.concurrent.Future
 import java.util.concurrent.Executors
+
 import scala.concurrent.ExecutionContext
 import org.rogach.scallop.ScallopOption
+
 import scala.concurrent.duration.Duration
 import scala.concurrent.Await
 import org.apache.daffodil.xml.QName
@@ -66,7 +68,7 @@ import org.apache.daffodil.compiler._
 import org.apache.daffodil.dsom.ExpressionCompilers
 import org.apache.daffodil.compiler.InvalidParserException
 import java.net.URI
-import org.apache.daffodil.api.URISchemaSource
+
 import org.apache.daffodil.tdml.TDMLException
 import org.apache.daffodil.xml.RefQName
 import org.rogach.scallop.ArgType
@@ -93,11 +95,11 @@ import javax.xml.transform.dom.DOMSource
 import javax.xml.transform.stream.StreamResult
 import javax.xml.parsers.DocumentBuilderFactory
 import org.apache.commons.io.IOUtils
-import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.io.InputSourceDataInputStream
 import org.apache.daffodil.tdml.TDMLTestNotCompatibleException
 import org.apache.daffodil.io.DataDumper
 import java.nio.ByteBuffer
+
 import org.apache.daffodil.io.FormatInfo
 import org.apache.daffodil.schema.annotation.props.gen.BitOrder
 import org.apache.daffodil.udf.UserDefinedFunctionFatalErrorException
@@ -611,7 +613,7 @@ object Main extends Logging {
    * @param vars The individual variables input via the command line using the -D command.
    * @param configFileNode The Node representing the configuration file if there is one.
    */
-  def retrieveExternalVariables(vars: Map[String, String], configFileNode: Option[Node], tunables: DaffodilTunables): Seq[Binding] = {
+  def retrieveExternalVariables(vars: Map[String, String], configFileNode: Option[Node]): Seq[Binding] = {
     val configFileVars: Seq[Binding] = configFileNode match {
       case None => Seq.empty
       case Some(configNode) => {
@@ -621,13 +623,13 @@ object Main extends Logging {
         extVarBindingNodeOpt match {
           case None => Seq.empty
           case Some(extVarBindingsNode) => {
-            ExternalVariablesLoader.getVariables(extVarBindingsNode)
+            Binding.getBindings(extVarBindingsNode)
           }
         }
       }
     }
 
-    val individualVars = ExternalVariablesLoader.getVariables(vars)
+    val individualVars = ExternalVariablesLoader.mapToBindings(vars)
 
     val bindings = overrideBindings(individualVars, configFileVars)
     bindings
@@ -638,8 +640,7 @@ object Main extends Logging {
     val processor = Timer.getResult("reloading", compiler.reload(savedParser))
     displayDiagnostics(processor)
     if (!processor.isError) {
-      processor.setValidationMode(mode)
-      Some(processor)
+      Some(processor.withValidationMode(mode))
     } else None
   }
 
@@ -686,17 +687,14 @@ object Main extends Logging {
   }
 
   def createProcessorFromSchema(schema: URI, rootNS: Option[RefQName], path: Option[String],
-    extVars: Seq[Binding],
     tunables: Map[String, String],
     mode: ValidationMode.Type) = {
-    val compiler = Compiler()
-
-    compiler.setExternalDFDLVariables(extVars)
-    compiler.setTunables(tunables)
-
-    rootNS match {
-      case None => // nothing
-      case Some(RefQName(_, root, ns)) => compiler.setDistinguishedRootNode(root, ns.toStringOrNullIfNoNS)
+    val compiler = {
+      val c = Compiler().withTunables(tunables)
+      rootNS match {
+        case None => c
+        case Some(RefQName(_, root, ns)) => c.withDistinguishedRootNode(root, ns.toStringOrNullIfNoNS)
+      }
     }
 
     // Wrap timing around the whole of compilation
@@ -709,9 +707,8 @@ object Main extends Logging {
     val pf = Timer.getResult("compiling", {
       val processorFactory = compiler.compileSource(schemaSource)
       if (!processorFactory.isError) {
-        val processor = processorFactory.onPath(path.getOrElse("/"))
+        val processor = processorFactory.onPath(path.getOrElse("/")).withValidationMode(mode)
         displayDiagnostics(processor)
-        processor.setValidationMode(mode)
         Some(processor) // note: processor could still be isError == true
         // but we do definitely get a processor.
       } else {
@@ -809,28 +806,23 @@ object Main extends Logging {
 
         val validate = parseOpts.validate.toOption.get
 
-        val cfgFileNode = parseOpts.config.toOption match {
+        val cfgFileNode: Option[Node] = parseOpts.config.toOption match {
           case None => None
           case Some(pathToConfig) => Some(this.loadConfigurationFile(pathToConfig))
         }
 
-        val processor = {
+        val processor: Option[DFDL.DataProcessor] = {
           if (parseOpts.parser.isDefined) {
-            val p = createProcessorFromParser(parseOpts.parser(), parseOpts.path.toOption, validate)
-            p.get.setExternalVariables(retrieveExternalVariables(parseOpts.vars, cfgFileNode, p.get.getTunables()))
-            p
+            createProcessorFromParser(parseOpts.parser(), parseOpts.path.toOption, validate)
           } else {
             val tunables = retrieveTunables(parseOpts.tunables, cfgFileNode)
-            val tunablesObj = DaffodilTunables(tunables)
-            val extVarsBindings = retrieveExternalVariables(parseOpts.vars, cfgFileNode, tunablesObj)
-
-            //val schema = new URI(parseOpts.schemaString())
-            createProcessorFromSchema(parseOpts.schema(), parseOpts.rootNS.toOption, parseOpts.path.toOption, extVarsBindings, tunables, validate)
+            createProcessorFromSchema(parseOpts.schema(), parseOpts.rootNS.toOption, parseOpts.path.toOption, tunables, validate)
           }
-        }
+        }.map{ _.withExternalVariables(retrieveExternalVariables(parseOpts.vars, cfgFileNode))}
 
         val rc = processor match {
-          case Some(processor) if (!processor.isError) => {
+          case Some(proc) if (!proc.isError) => {
+            var processor = proc
             val input = parseOpts.infile.toOption match {
               case Some("-") | None => System.in
               case Some(file) => {
@@ -840,7 +832,8 @@ object Main extends Logging {
             }
             val inStream = InputSourceDataInputStream(input)
 
-            processor.setValidationMode(validate)
+            processor = processor.withValidationMode(validate)
+
             setupDebugOrTrace(processor.asInstanceOf[DataProcessor], conf)
 
             val output = parseOpts.output.toOption match {
@@ -975,17 +968,13 @@ object Main extends Logging {
 
         val processor = {
           if (performanceOpts.parser.isDefined) {
-            val p = createProcessorFromParser(performanceOpts.parser(), performanceOpts.path.toOption, validate)
-            p.get.setExternalVariables(retrieveExternalVariables(performanceOpts.vars, cfgFileNode, p.get.getTunables()))
-            p
+            createProcessorFromParser(performanceOpts.parser(), performanceOpts.path.toOption, validate)
           } else {
             val tunables = retrieveTunables(performanceOpts.tunables, cfgFileNode)
-            val tunablesObj = DaffodilTunables(tunables)
-            val extVarsBindings = retrieveExternalVariables(performanceOpts.vars, cfgFileNode, tunablesObj)
-
-            createProcessorFromSchema(performanceOpts.schema(), performanceOpts.rootNS.toOption, performanceOpts.path.toOption, extVarsBindings, tunables, validate)
+            createProcessorFromSchema(performanceOpts.schema(), performanceOpts.rootNS.toOption, performanceOpts.path.toOption, tunables, validate)
           }
-        }
+        }.map{ _.withExternalVariables(retrieveExternalVariables(performanceOpts.vars, cfgFileNode)) }
+         .map{ _.withValidationMode(validate) }
 
         val rc = processor match {
           case Some(processor: DataProcessor) if (!processor.isError) => {
@@ -1018,8 +1007,6 @@ object Main extends Logging {
               dataSeq(index)
             }
             val inputsWithIndex = inputs.zipWithIndex
-
-            processor.setValidationMode(validate)
 
             implicit val executionContext = new ExecutionContext {
               val threadPool = Executors.newFixedThreadPool(performanceOpts.threads())
@@ -1103,17 +1090,12 @@ object Main extends Logging {
 
         val processor = {
           if (unparseOpts.parser.isDefined) {
-            val p = createProcessorFromParser(unparseOpts.parser(), unparseOpts.path.toOption, validate)
-            p.get.setExternalVariables(retrieveExternalVariables(unparseOpts.vars, cfgFileNode, p.get.getTunables()))
-            p
+            createProcessorFromParser(unparseOpts.parser(), unparseOpts.path.toOption, validate)
           } else {
             val tunables = retrieveTunables(unparseOpts.tunables, cfgFileNode)
-            val tunablesObj = DaffodilTunables(tunables)
-            val extVarsBindings = retrieveExternalVariables(unparseOpts.vars, cfgFileNode, tunablesObj)
-
-            createProcessorFromSchema(unparseOpts.schema(), unparseOpts.rootNS.toOption, unparseOpts.path.toOption, extVarsBindings, tunables, validate)
+            createProcessorFromSchema(unparseOpts.schema(), unparseOpts.rootNS.toOption, unparseOpts.path.toOption, tunables, validate)
           }
-        }
+        }.map{ _.withExternalVariables(retrieveExternalVariables(unparseOpts.vars, cfgFileNode)) }
 
         val output = unparseOpts.output.toOption match {
           case Some("-") | None => System.out
@@ -1190,9 +1172,8 @@ object Main extends Logging {
         }
         val tunables = retrieveTunables(saveOpts.tunables, cfgFileNode)
         val tunablesObj = DaffodilTunables(tunables)
-        val extVarsBindings = retrieveExternalVariables(saveOpts.vars, cfgFileNode, tunablesObj)
 
-        val processor = createProcessorFromSchema(saveOpts.schema(), saveOpts.rootNS.toOption, saveOpts.path.toOption, extVarsBindings, tunables, validate)
+        val processor = createProcessorFromSchema(saveOpts.schema(), saveOpts.rootNS.toOption, saveOpts.path.toOption, tunables, validate)
 
         val output = saveOpts.outfile.toOption match {
           case Some("-") | None => Channels.newChannel(System.out)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -25,7 +25,7 @@ import java.nio.channels.Channels
 import java.util.zip.GZIPInputStream
 import java.util.zip.ZipException
 
-import scala.collection.mutable.Queue
+import scala.collection.immutable.Queue
 import scala.xml.Node
 
 import org.apache.daffodil.ExecutionMode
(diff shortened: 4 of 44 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -480,66 +480,70 @@ final class SchemaSet(
...
-      finalExternalVariables.enqueue(newBinding)
-    })
-    finalExternalVariables
-  }
-
-  override def variableMap = LV('variableMap) {
+  lazy val allDefinedVariables = schemas.flatMap{ _.defineVariables }
+  lazy val allExternalVariables = allDefinedVariables.filter{ _.external }
+
+//  /**
```
mbeckerle: Remove

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala
```diff
@@ -480,66 +480,70 @@ final class SchemaSet(
...
 
     // At this point we want to try to figure out which, if any, external
     // variables did not have a namespace specified.
-    val finalExternalVariables = resolveExternalVariableNamespaces(alldvs)
-
-    val finalVMap =
-      ExternalVariablesLoader.loadVariables(finalExternalVariables, this, vmap)
-
-    finalVMap
+//    val finalExternalVariables = resolveExternalVariableNamespaces(alldvs)
```
mbeckerle: Remove

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
```diff
@@ -107,10 +108,47 @@ import org.apache.daffodil.exceptions.Assert
  */
 object QName {
 
+  /**
+   * Construct a RefQName from a QName string, and a scope.
+   * @param qnameString
```
mbeckerle: Remove @param and friends. 
I hate these redundant things. They're always obsolete relative to the code.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
```diff
@@ -414,7 +430,7 @@ final case class GlobalQName(prefix: Option[String], local: String, namespace: N
 /**
  * base trait for Qnames that reference other things
  */
-protected sealed trait RefQNameBase extends QNameBase
+sealed trait RefQNameBase extends QNameBase
```
mbeckerle: Can go back to Protected.

daffodil-runtime1/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesLoader.scala
```diff
@@ -53,17 +54,9 @@ object ExternalVariablesLoader {
     varsKVP
   }
 
-  def getVariables(vars: Map[String, String]): Seq[Binding] = {
```
mbeckerle: Loving this IDEA quick "Find Usages" feature. So easy to find methods that are not in use anywhere!

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -96,15 +106,41 @@ class SerializableDataProcessor(
...
   with HasSetDebugger with Serializable
   with MultipleEventHandler {
 
-  def tunable = getTunables
+  /**
+   * The compilerExternalVars argument supports the deprecated feature to assign external var bindings
+   * on the compiler object.
+   *
+   * These are just incorporated into the initial variable map of the data processor.
+   * @param ssrd
```
mbeckerle: remove @param and friends

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -114,70 +150,156 @@ class DataProcessor(
...
 
-  def setValidationMode(mode: ValidationMode.Type): Unit = { ssrd.validationMode = mode }
-  def getValidationMode() = ssrd.validationMode
-  def getVariables = ssrd.variables
-  def getTunables = tunablesObj
+  /**
+   * Returns a data processor with the same schema, but with no tunables, no external variable bindings,
+   * and the default validation mode.
+   */
+  def clear(): DataProcessor = new DataProcessor(ssrd, tunables)
```
mbeckerle: Scaladoc is not consistent with behavior. This preserves the tunables.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -114,70 +150,156 @@ class DataProcessor(
...
+
+  private def newVariableMap(extVars: File, tunable: DaffodilTunables): VariableMap = {
+    ExternalVariablesLoader.loadVariables(extVars, ssrd, variableMap, getTunables)
+  }
+
+  private def newVariableMap(extVars: Seq[Binding]): VariableMap = {
+    ExternalVariablesLoader.loadVariables(extVars, ssrd, variableMap)
+  }
+
+//  private def getVariables(vars: Map[String, String], scope: scala.xml.NamespaceBinding, externalVRDs: Seq[VariableRuntimeData]): Seq[Binding] = {
```
mbeckerle: Remove.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -178,6 +188,7 @@ class VariableMap private (vTable: Map[GlobalQName, List[List[Variable]]])
   def getVariableBindings(qn: GlobalQName): List[List[Variable]] = {
     vTable.get(qn).get
   }
+
   //  def getVariableBindings(vrd: VariableRuntimeData): List[List[Variable]] = {
```
mbeckerle: remove

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -17,43 +17,47 @@
...
-import org.apache.daffodil.processors.{ ParseResult => SParseResult }
-import org.apache.daffodil.processors.{ UnparseResult => SUnparseResult }
-import org.apache.daffodil.util.{ ConsoleWriter => SConsoleWriter }
-import org.apache.daffodil.util.{ FileWriter => SFileWriter }
-import org.apache.daffodil.util.{ LogWriter => SLogWriter }
-import org.apache.daffodil.util.{ LoggingDefaults => SLoggingDefaults }
-import org.apache.daffodil.util.{ NullLogWriter => SNullLogWriter }
-import org.apache.daffodil.externalvars.ExternalVariablesLoader
+import org.apache.daffodil.api.{DataLocation => SDataLocation}
+import org.apache.daffodil.api.{Diagnostic => SDiagnostic}
```
stevedlawrence: Looks like yoru auto-import formatter broke? It changed everything to not use spaces?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -257,7 +264,7 @@ class Compiler private[japi] () {
    * @param tunables a map of key/value pairs, where the key is the tunable name and the value is the value to set it to
    */
   def setTunables(tunables: java.util.AbstractMap[String, String]): Unit = {
-    sCompiler.setTunables(tunables.asScala.toMap)
+    sCompiler = sCompiler.withTunables(tunables.asScala.toMap)
```
stevedlawrence: Shoudl all these setters be marked as deprecated and we add with* methods?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -267,7 +274,7 @@ class Compiler private[japi] () {
  * Do not use the ProcessorFactry constructor to create a [[ProcessorFactory]].
  * Instead, use [[Compiler#compileFile(java.io.File)]]
  */
-class ProcessorFactory private[japi] (pf: SProcessorFactory)
+class ProcessorFactory private[japi] (var pf: SProcessorFactory)
```
stevedlawrence: Can the var be made private?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -419,7 +426,7 @@ class LocationInSchemaFile private[japi] (lsf: SLocationInSchemaFile) {
 /**
  * Compiled version of a DFDL Schema, used to parse data and get the DFDL infoset
  */
-class DataProcessor private[japi] (dp: SDataProcessor)
+class DataProcessor private[japi] (var dp: SDataProcessor)
```
stevedlawrence: Can this var be made private?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
```diff
@@ -477,6 +489,15 @@ protected trait RefQNameFactoryBase[T] {
...
 
+  /**
+   * This variant to resolve normal QNames such as in a ref="pre:local" syntax, or
+   * a path step like these steps .../foo/bar:baz/quux.
+   *
+   * @param qnameString
+   * @param scope
+   * @param unqualifiedPathStepPolicy
+   * @return
+   */
```
stevedlawrence: Empty scaladoc on the variables.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
```diff
@@ -132,22 +140,57 @@ object DFDL {
...
 
   trait DataProcessor extends WithDiagnostics {
-    def setValidationMode(mode: ValidationMode.Type): Unit
-    def getValidationMode(): ValidationMode.Type
+
+    /**
+     * Returns a data processor with the same schema, no external variable bindings,
+     * and the default validation mode.
+     */
+    def clear(): DataProcessor
```
stevedlawrence: Is this just so users don't need to keep a reference to the original DataProcessor

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -176,7 +179,7 @@ class Compiler private[sapi] () {
    *                  no namespace. Set to to NULL to figure out the namespace.
    */
   def setDistinguishedRootNode(name: String, namespace: String): Unit =
```
stevedlawrence: Same questions as the Java API. Should we deprecate these, and can the var constructor paremters be made private.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: I have not yet gone so far as to modify sapi/japi, and the TDMLDFDLProcessor classes. These are one-indirection away from the actual compiler, PF, and DataProcessor objects. It's pretty easy to make them work properly with setters. 

That said, yes I do think a side-effect free design point is better, as it eliminates broad classes of otherwise subtle bugs. I just didn't want to go that far in this patch.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -686,17 +687,14 @@ object Main extends Logging {
...
-    compiler.setTunables(tunables)
-
-    rootNS match {
-      case None => // nothing
-      case Some(RefQName(_, root, ns)) => compiler.setDistinguishedRootNode(root, ns.toStringOrNullIfNoNS)
+    val compiler = {
+      val c = Compiler().withTunables(tunables)
+      rootNS match {
+        case None => c
+        case Some(RefQName(_, root, ns)) => c.withDistinguishedRootNode(root, ns.toStringOrNullIfNoNS)
```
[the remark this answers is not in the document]
mbeckerle: The right thing is for compileSource to take an optional argument which is the root spec. If unspecified you get "infer from schema" behavior. 

So yes I suppose we should ditch withDistinguishedRootNode entirely. It's purpose is actually to make converting setter-based code more straightforward. 

val compiler = Compiler()
compiler.setDistinguishedRootNode(...)

just becomes

var compiler = Compiler()
compiler = compiler.withDistinguishedRootNode(...)

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -192,34 +216,63 @@ final class ProcessorFactory(val sset: SchemaSet)
...
+
+  /**
+   * Identical method to what we have in ProcessorFactory, but I couldn't
+   * come up with a way to share it in the HavingRootSpec mixin because it calls copy. So it is duplicated.
+   * Only 2 lines though. .
+   */
+  override def withDistinguishedRootNode(name: String, namespace: String): Compiler =
+    copy(rootSpec = getRootSpec(name, namespace))
+
+  @deprecated("2020-03-26", "Use withValidateDFDLSchemas.")
```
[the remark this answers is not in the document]
mbeckerle: Seriously, they reversed the arg order from java's syntax.
Why why why....
I will see if @deprecated(since = "2016-03-26", message="...") works also, since that's clearer. 

Also it seems that the since argument can be "Daffodil 2.6.0" instead of a date, which is better.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -232,31 +285,85 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
...
+  def withExternalDFDLVariable(variable: Binding): Compiler =
+    copy(externalDFDLVariables = externalDFDLVariables.enqueue(variable))
+
+  @deprecated("2020-03-26", "Use DataProcessor.withExternalVariables.")
+  def setExternalDFDLVariables(variables: Seq[Binding]): Unit =
+    variables.foreach(b => externalDFDLVariables = externalDFDLVariables.enqueue(b))
+
+  @deprecated("2020-03-26", "Use DataProcessor.withExternalVariables.")
+  def withExternalDFDLVariables(variables: Seq[Binding]): Compiler =
+    withExternalDFDLVariablesImpl(variables)
```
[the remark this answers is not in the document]
mbeckerle: Temporary place-holding. Yes these should go away. 

Converting from setters to with-ers (?) is step 1.

Refactoring to pass the external vars bindings down to where you ultimately create a data processor is step 2. 

Only step 1 is required to fix threading-related problems. Which I wanted to get away from entirely first.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -232,31 +285,85 @@ class Compiler(var validateDFDLSchemas: Boolean = true)
...
+
+  //
+  // Not deprecated so that we can implement the deprecated things
+  // and reuse code.
+  // When the deprecated methods go away, so should this.
+  //
+  def withExternalDFDLVariablesImpl(variables: Seq[Binding]): Compiler = {
```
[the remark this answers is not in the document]
mbeckerle: Can't . It is used in japi/sapi, and the TDML processor.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Yes. I forgot that var or val makes it public.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
mbeckerle: Yeah, I hate these scaladoc redundant tag things. They're either empty, or not maintained properly because the code changes, and they do not. 

I am using Intellij IDEA now, and haven't go tthese things properly customized yet. There may be whitespace and indent things that are wrong too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
mbeckerle: Yes. Also so that you can use the deprecated setters, and take the resulting DP and get a clean one to start from again. 

I'm not sure *we* need this method. I call it one place and I'm not sure it is needed there.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Fixed. There is a setting for this. It's not set by default. Added a comment to the IDEA page on wiki.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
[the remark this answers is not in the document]
mbeckerle: Here is the rock/hard-place I ended up.

If we need to still support the deprecated functionality, then we need the "withDistinguishedRootNode" method in order to still have a non-deprecated way to implement the deprecated behavior.

Until there is a way to silence deprecation warnings that are created from inside other deprecated things, there's no good way to do this other than introduce another "interim" API, albeit one that works more functionally.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: It is much too involved for this patch set. Took me another whole day pulling the other threads of this review. SAPI/JAPI, and the TDMLDFDLProcessor classes all still have imperative APIs. 

They're much safer now, in that if you say, spawn lots of parse calls on varous threads, and then start calling things like setExternalDFDLVariables again on the DP object, nothing bad will happen, because those setter calls displace the state with a new state.

daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
```diff
@@ -336,6 +334,17 @@ class Fakes private () {
     def setTunables(tunables: Map[String, String]): Unit = {}
     def getTunables(): DaffodilTunables = { tunablesObj }
 
+    def clear(): DFDL.DataProcessor = this
```
mbeckerle: don't need this clear method.
Other methods should all get override keyword

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -189,8 +203,9 @@ class Compiler private[japi] () {
    * @param namespace namespace of the root node. Set to empty string to specify
    *                  no namespace. Set to to NULL to figure out the namespace.
    */
+  @deprecated("Pass argument to compileSource, or compileFile.", "Since 2.6.0")
```
mbeckerle: arguments plural

mbeckerle: Probably shouldnt' deprecate this for now. Consistent with not changing JAPI/SAPI at all.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -73,190 +73,169 @@ object ForParser extends ParserOrUnparser
...
+  /*
+   * compilerExternalVarSettings supports the deprecated API
+   * where external variable settings can be supplied to the compiler
+   * instance.
+   *
+   * The non-deprecated API is to deal with external variable settings
+   * on the data processor instance. This is passed here so that
+   * the PF can propagate it to the DP.
+   */
+  var compilerExternalVarSettings: Queue[Binding],
```
mbeckerle: Why is this a Queue of bindings? There is no "unset" for external variables, so while you could depend on order that they are being set to overwrite one with a new value, it's not a general technique because if you want the variable back to its default value "whatever that is", there's no way to express that.

daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
```diff
@@ -262,6 +262,10 @@ class TestCLISaveParser {
     }
   }
 
+  /**
+   * Note that in Daffodil 2.6.0 behavior of external variables changed. They are not saved as part of binary
+   * compiling. They are a runtime-thing only.
+   */
```
stevedlawrence: Maybe we shoudl add a warning in the CLI

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -73,190 +73,169 @@ object ForParser extends ParserOrUnparser
...
-class InvalidParserException(msg: String, cause: Throwable = null) extends Exception(msg, cause)
 
-class Compiler(var validateDFDLSchemas: Boolean = true)
-  extends DFDL.Compiler
-  with Logging
-  with HavingRootSpec {
+  def withDistinguishedRootNode(name: String, namespace: String) : Compiler = {
+    Assert.usage(name ne null)
+    copy(optRootName = Option(name), optRootNamespace = Option(namespace))
+  }
```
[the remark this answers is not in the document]
mbeckerle: You have to have this in order to support the old API style (which is still in place in JAPI/SAPI, and the TDML Processor), without having to call a deprecated method, which you can't do without us turning off -Xfatal-warnings at compilation time. So even though we don't technically want this method, there's nothing wrong with it per-se. It's functional anyway. 

I'd rather not have it, just because I'd like there to ultimately be one way to do things, and the best functional interface is one function with a bunch of optional args.  

But, I think we have to have this or we have to convert JAPI/SAPI and the TDML processor all at the same time, which is, frankly, too much change all at once, and the implications of it are not clear.  Even if we wanted to change the JAPI/SAPI and TDML processor APIs, we still have to support their deprecated behavior for now, so we'd *still* need this "with-X" method for the time being.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -122,8 +125,14 @@ class Compiler private[japi] () {
...
   @throws(classOf[java.io.IOException])
-  def compileFile(schemaFile: File): ProcessorFactory = {
-    val pf = sCompiler.compileFile(schemaFile)
+  def compileFile(schemaFile: File): ProcessorFactory = compileFile(schemaFile, None, None)
+
+  @throws(classOf[java.io.IOException])
+  def compileFile(schemaFile: File,
+    optRootName: Option[String],
+    optRootNamespace: Option[String]): ProcessorFactory = {
+
```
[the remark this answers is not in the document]
mbeckerle: Good point. These can be string or null and I'll option them inside.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -114,70 +147,126 @@ class DataProcessor(
...
+  def withExternalVariables(extVars: Seq[Binding]): DataProcessor = {
+    copy(variableMap = newVariableMap(extVars))
+  }
+
+  @deprecated("Use withTunables.", "Since 2.6.0")
+  def setTunable(tunable: String, value: String): Unit = tunables = tunables.setTunable(tunable, value)
+
+  def withTunable(tunable: String, value: String): DataProcessor = copy(tunables = tunables.setTunable(tunable, value))
+
+  @deprecated("Use withTunables.", "Since 2.6.0")
```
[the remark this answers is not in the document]
mbeckerle: I should have guessed  that the arg name is "since", so since="Since ..." wouldn't be clean code.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/RootSpec.scala
```diff
@@ -33,4 +33,9 @@ case class RootSpec(ns: Option[NS], name: String) {
     val nsStr = ns.getOrElse("")
     "{" + nsStr + "}" + name
   }
+}
+
+object RootSpec {
+  def makeRootSpec(optName: Option[String], optNamespace:Option[String]) =
+    optName.map{ RootSpec(optNamespace.map{ NS(_)}, _) }
```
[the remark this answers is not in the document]
mbeckerle: This is an API, so I can just add an Assert.usage(test) that checks this additional usage mistake.

daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
mbeckerle: There are things we can check at compile time. First is that the settings are going to be legal, which requires that the variables being set are in fact declared 'external', and that the value provided can be converted to the proper type. 

A warning, that the external vars are checked, but not saved, as part of save-parser makes sense at the CLI and at the save API call. Internally, however, where we save/restore (e.g., for every TDML test) we don't want this warning.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Ok. I will take a pass at deprecating all setters everywhere on the basic API, the JAPI/SAPI and TDML processor APIs. 

The headache is updating all the tests that call these things now using what will be deprecated methods.

stevedlawrence: If updating the test is too much of pain we could change the sbt config so that warnings are not treated as errors for the test target. So we would have have a different ``scalacOptions in Test :=`` in the sbt build settings.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -73,190 +73,169 @@ object ForParser extends ParserOrUnparser
...
-    val unp = if (generateUnparser) rootElem.document.unparser else new NotUnparsableUnparser(rootElem.erd)
-    Processor.initialize(unp)
-    unp
-  }.value
-
-  lazy val rootElem = sset.root
+  def withDistinguishedRootNode(name: String, namespace: String) : ProcessorFactory = {
+    Assert.usage(name ne null)
+    copy(optRootSpec = RootSpec.makeRootSpec(Option(name), Option(namespace)))
+  }
```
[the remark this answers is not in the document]
mbeckerle: You can't do that. The distinguished root must be specified before you call compileFile. 

I guess I am convinced we should deprecate setDistinguishedRootNode, and NOT provide a replacement that is officially part of the API. 

Usage should be to pass the distinghished root at the compile call.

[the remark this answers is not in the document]
mbeckerle: You have to do
`Compiler().withDistinguishedRootNode(n,ns).compileFile(file)`
in that order. 
The call to compileFile/compileSource must be the last thing you do for compiling that file.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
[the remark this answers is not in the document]
mbeckerle: I have updated everything now to deprecate all setters in JAPI/SAPI, and TDML processor.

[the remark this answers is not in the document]
mbeckerle: btw: this method on compiler withDistinguishedRootNode is needed to implement the deprecated setter in JAPI/SAPI, and the TDML processor. It's not part of the API officially. Not in DFDL.Compiler.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -73,190 +73,169 @@ object ForParser extends ParserOrUnparser
...
-class Compiler(var validateDFDLSchemas: Boolean = true)
-  extends DFDL.Compiler
-  with Logging
-  with HavingRootSpec {
+  def withDistinguishedRootNode(name: String, namespace: String) : Compiler = {
+    Assert.usage(name ne null)
+    copy(optRootName = Option(name), optRootNamespace = Option(namespace))
+  }
 
+  @deprecated("Use constructor argument.", "Since 2.6.0")
```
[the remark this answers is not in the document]
mbeckerle: That method is needed to implement deprecated methods of JAPI/SAPI, etc. but we don't need to advertise 2 ways of doing this. 

This withValidateDFDLSchemas is not on DFDL.Compiler, only the internal Compiler class.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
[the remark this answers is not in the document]
mbeckerle: The preferred style is to pass the args. 

The setter on DFDL.Compiler is deprecated.

The internal Compiler has a withDistingushedRootNode method, as we need it to implement the deprecated setter in SAPI/JAPI. But it's not "official"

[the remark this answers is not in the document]
mbeckerle: E.g. SAPI.Compiler doesn't have a withDistinghishedRootNode, just a deprecated setter and args on compileSource and compileFile.

mbeckerle:
There is one fix left to do which is the unresolved comment about checking the external vars if specified on the compiler.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -396,57 +460,117 @@ class LocationInSchemaFile private[sapi] (lsf: SLocationInSchemaFile) {
...
+  @deprecated("Use withDebugging.", "2.6.0")
+  def setDebugging(flag: Boolean) {
+    dp = dp.withDebugging(flag)
+  }
 
   /**
-   * Set the debugger runer
+   * Obtain a new [[DataProcessor]] instance with debugging enabled or disabled.
+   *
+   * Before enabling, [[DataProcessor#withDebugger(DebuggerRunner)]] must be called to obtain a [[DataProcessor]] with a non-null debugger.
```
stevedlawrence, line 484: When generating the scaladoc, I get warnings that this can't be found. I think if you just use ``DataProcessor.withDebugger`` without the arguments it should be able to find it. In scaladoc, if you provide arguments in a scaladoc link, you have to include the parameter name. Same issue with ``setDebugger`` above.

mbeckerle: This is because I copied code from JAPI to SAPI when I thought they were identical, but they were not. The "code" was identical, not the scaladoc comments. Will fix.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -200,9 +220,10 @@ class Compiler private[sapi] () {
    *                   then no namespace is used. With not preceded by "{namespace}",
    *                   then Daffodil will figure out the namespace.
    */
+  @deprecated("Use DataProcessor.setExternalVariables", "2.6.0")
   def setExternalDFDLVariables(extVarsMap: Map[String, String]): Unit = {
```
stevedlawrence, line 224: When I build the scaladoc,, for some reason this method doesnt show up as deprecated. Same issue with this function in the DataProcessor. And it's only this variable setter. All the other variable setters are marked as deprecated. Looks correct to me though...

build.sbt
```diff
@@ -113,7 +113,11 @@ lazy val commonSettings = Seq(
     "-language:experimental.macros",
     "-unchecked",
     "-Xfatal-warnings",
-    "-Xxml:-coalescing",
+    "-Xxml:-coalescing"
+  ),
+  javacOptions ++= Seq(
+    "-Werror",
+    "-Xlint:deprecation"
   ),
```
stevedlawrence, line 121: I noticed an issue when running ``sbt universal:packageBin``. This command calls ``sbt doc`` for all subprojects. In particular, it calls ``sbt daffodil-udf/doc``. It looks like because there is Java code in this repo is calls javadoc, and these javacOptions get passed to it. But these aren't valid options for javadoc, only for javac. There's an sbt issue talking about this behavior: https://github.com/sbt/sbt/issues/355

Looks like the solution is to do ``javacOptions in Compile in compile ++= ...``

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
mbeckerle: Probably bug in scaladoc. There are multiple overloads of this method name, and they even have the same number of arguments, same return tyhpe. Only the arg type is different. Easy to imagine a bug there.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -458,13 +582,29 @@ class DataProcessor private[sapi] (dp: SDataProcessor)
    *                then no namespace is used. If not preceded by anything,
    *                then Daffodil will figure out the namespace.
    */
-  def setExternalVariables(extVars: Map[String, String]) = dp.setExternalVariables(extVars)
+  def setExternalVariables(extVars: Map[String, String]) =
```
stevedlawrence, line 585: Ah, I think this is the one I was seeing not being deprecated. I tihnk I was just confused before. This is the only one that's missing an @deprecated annotation. Everything else looks good.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
mbeckerle: Decided to leave this a queue. It's possible that bindings are ordered. Rather apply them in order than worry about de-duplicating them.

build.sbt
mbeckerle: Fixed

stevedlawrence:
Also, looks like a test failed, looks like it can't serialize a Binding, which is odd since the externalVars Binding is supposed to be transient...maybe it something else?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -96,15 +106,41 @@ class SerializableDataProcessor(
...
+   * on the compiler object.
+   *
+   * These are just incorporated into the initial variable map of the data processor.
+   */
+  def this(ssrd: SchemaSetRuntimeData, tunables:DaffodilTunables, compilerExternalVars: Queue[Binding] = Queue.empty) =
+    this(ssrd, tunables, false, None, ValidationMode.Off,
+      ExternalVariablesLoader.loadVariables(compilerExternalVars, ssrd, ssrd.originalVariables),
+      compilerExternalVars)
+
+  private def copy(
```
jadams-tresys, line 135: This will come in handy with some of the issues I was running into with making Variables mutable.  Was running into issues with the DataProcessor being shared between threads, so the parse thread was effecting the state of variables for unparse and I was going to need to implement a copying function just like this.

PR #344, opened 2020-03-31
Use reference equality for dsom shared keys

stevedlawrence:
By using the default object equality, it is possible that two chunks of
XML could look exactly the same and will be shared, even though their
contexts are different and should not be shared. By using reference
equality, XML nodes that happen to have the exact same content but in
different contexts will be seen as different and will not be shared.

DAFFODIL-2308

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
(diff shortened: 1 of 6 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
```diff
@@ -0,0 +1,55 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<!--
+  This schema contents references to two completely different elements
```
mbeckerle: Typo contents -> contains

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
```diff
@@ -0,0 +1,55 @@
...
+  (simpleElement1 and simpleElement2), which are exactly the same (including
+  line numbers) except for the root elements and a differentiator. This schema
```
mbeckerle: differentiator?

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
stevedlawrence: Fixed.

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SimilarModelGroupRoot.dfdl.xsd
stevedlawrence: Yeah, this isn't clear. The specific bug with link16 requires 1) the choice dispatch to be on the same line number with the same contents 2) in separate schemas (and thus a different parent element) 3) and a unique sibling element. Without these three things, the bug would not be triggered. I'll change "differentiator" to "a different sibling element".

PR #345, opened 2020-04-01
Adds signing keys for release

olabusayoT:


KEYS
```diff
@@ -146,3 +146,51 @@ qIpenMsTA9TFms/EVMkvvyqllMVjjbLRlg0UMhnf/DzgrErp646LkAU0m4jSzGVn
 QDM28gGyQYTRoMudTseM7Q2//F9LF9pz4tzK0g==
 =09Ch
 -----END PGP PUBLIC KEY BLOCK-----
+pub   rsa3072 2020-04-01 [SC] [expires: 2022-04-01]
```
stevedlawrence, line 149: Does that 3072 mean this is an RSA key with 3072 bit length? Apache recommends using 4096: https://www.apache.org/dev/openpgp.html#generate-key

You might want to check https://www.apache.org/dev/openpgp.html#sha-defaults so that a slightly stronger key is created.

Also, note that it looks like your key expires in a couple of years. I don't think apache requires that keys expire, but make things a little easier so you don't have to update the key in a couple years.

olabusayoT: Ah that makes sense. Can we update the file to instruct people to use "gpg --full-generate-key" instead and maybe include your link above? "gpg --gen-key" from the file selects all the defaults for you

stevedlawrence: Definitely. I'm pretty sure we took this file from some other Apache project, but if we can make the directions less error prone for our contribututors we definitely should. Feel free to include that in this PR.

PR #346, opened 2020-04-02
Fix TDML runner losing external var assignments in middle of a round …

mbeckerle:
…trip.

DAFFODIL-2311

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -429,6 +429,17 @@ class DFDLTestSuite private[tdml] (
 abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
   extends Logging {
 
+  /**
+   * Test case execution is strongly sequentialized due to this central var.
+   *
+   * Whenever we compute a new processor we just side-effect this state.
+   *
+   * Note getting rid of this var would require substantial code restructuring, which is why
+   * that wasn't done when this code was updated to call the processor.withXYZ methods which return a
+   * new processor.
+   */
+  protected final var processor: TDMLDFDLProcessor = null
+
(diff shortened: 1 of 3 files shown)


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -429,6 +429,8 @@ class DFDLTestSuite private[tdml] (
 abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
   extends Logging {
```
mbeckerle, line 431: Add comment that TestCase execution is strongly sequential and single-thread because whenever we create a new processor with new settings we just assign this var. 

I did things this way rather than restructure the code more functionally, which would have been a lot of churn. So this code still reflects imperative style code of the API it was using before.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -804,8 +815,10 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
       case (Some(_), None) => {
         compileResult.left.foreach { diags => throw TDMLException(diags, implString) }
         compileResult.right.foreach {
-          case (_, processor) =>
-            runParseExpectSuccess(processor, dataToParse, nBits, optExpectedWarnings, optExpectedValidationErrors, validationMode, roundTrip, implString)
+          case (_, proc) => {
+            processor = proc
+            runParseExpectSuccess(dataToParse, nBits, optExpectedWarnings, optExpectedValidationErrors, validationMode, roundTrip, implString)
+        }
```
stevedlawrence, line 821: indentation wrong for this curly brace.

mbeckerle: I have been hesitant to turn on auto-formatting on IDEA thusfar. At some point I may try it. Right now I have to decide to use it manually.

PR #349, opened 2020-04-06
Upgrades sbt to 1.3.9, scala to 2.12.11, genjavadoc to 0.16.

mbeckerle:
sbt, depending on the terminal you are displaying it in, will issue ANSI
colorization escapes. You may need to add the color=false option to your
environment: e.g., in my .bash_aliases I have

export SBT_OPTS=" -Dsbt.color=false "

Had to fix a number of deprecations.
junit.framework.Assert is deprecated. Changed to org.junit.Assert.

Assert.assertEquals for float/double now requires a delta argument. Supplied
1e-15 where needed.

Since scala 2.12.7, shifting of values by quantities expressed as Long is deprectated. Had to insert ".toInt" calls a number of places to silence these.

DAFFODIL-2313

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,10 +25,11 @@ jobs:
       fail-fast: false
       matrix:
         java_version: [ '8.x', '9.x', '11.x' ]
-        scala_version: [ '2.12.6', '2.11.12' ]
+        scala_version: [ '2.12.11', '2.11.12' ]
         os: [ 'ubuntu-latest', 'windows-latest' ]
     env:
       SBT: sbt -J-Xms1024m -J-Xmx5120m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m ++${{ matrix.scala_version }} coverage
+      SBTNOCOV: sbt -J-Xms1024m -J-Xmx5120m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m ++${{ matrix.scala_version }}
     steps:
 
       ############################################################
@@ -63,15 +64,15 @@ jobs:
         shell: bash
 
       - name: Build Documentation
-        run: $SBT daffodil-japi/genjavadoc:doc daffodil-sapi/doc
+        run: $SBTNOCOV daffodil-japi/genjavadoc:doc daffodil-sapi/doc
         shell: bash
 
       - name: Package Zip & Tar
-        run: $SBT daffodil-cli/universal:packageBin daffodil-cli/universal:packageZipTarball
+        run: $SBTNOCOV daffodil-cli/universal:packageBin daffodil-cli/universal:packageZipTarball
         shell: bash
 
       - name: Package RPM
-        run: $SBT daffodil-cli/rpm:packageBin
+        run: $SBTNOCOV daffodil-cli/rpm:packageBin
         if: runner.os == 'Linux'
         shell: bash
 
@@ -80,7 +81,7 @@ jobs:
       ############################################################
 
       - name: Run Rat Check
-        run: $SBT ratCheck || (cat target/rat.txt; exit 1)
+        run: $SBTNOCOV ratCheck || (cat target/rat.txt; exit 1)
         shell: bash
 
       - name: Run Unit Tests
--- a/build.sbt
+++ b/build.sbt
@@ -105,8 +105,8 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "2.6.0",
-  scalaVersion := "2.12.6",
-  crossScalaVersions := Seq("2.12.6", "2.11.12"),
+  scalaVersion := "2.12.11",
+  crossScalaVersions := Seq("2.12.11", "2.11.12"),
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation",
--- a/daffodil-japi/build.sbt
+++ b/daffodil-japi/build.sbt
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-addCompilerPlugin("com.typesafe.genjavadoc" %% "genjavadoc-plugin" % "0.11" cross CrossVersion.full)
+addCompilerPlugin("com.typesafe.genjavadoc" %% "genjavadoc-plugin" % "0.16" cross CrossVersion.full)
 
 
 lazy val JavaDoc = config("genjavadoc") extend Compile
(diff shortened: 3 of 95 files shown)


daffodil-japi/build.sbt
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-addCompilerPlugin("com.typesafe.genjavadoc" %% "genjavadoc-plugin" % "0.11" cross CrossVersion.full)
+addCompilerPlugin("com.typesafe.genjavadoc" %% "genjavadoc-plugin" % "0.16" cross CrossVersion.full)
```
stevedlawrence, line 18: Looks like something is wrong with our java docs now. GitHub actions are failing at generating the docs.

build.sbt
```diff
@@ -105,8 +105,8 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "2.6.0",
-  scalaVersion := "2.12.6",
-  crossScalaVersions := Seq("2.12.6", "2.11.12"),
+  scalaVersion := "2.12.11",
+  crossScalaVersions := Seq("2.12.11", "2.11.12"),
```
stevedlawrence, line 109: We also reference the scala version in daffodil-cli/bin.LICENSE

project/build.properties
```diff
@@ -15,4 +15,4 @@
  * limitations under the License.
  */
 
-sbt.version=1.2.7
+sbt.version=1.3.9
```
stevedlawrence, line 18: We also reference the sbt version in ``containers/release-candidate/setup-container.sh``

olabusayoT:
I think an update to .github/workflows files, as they include the scala versions as well

daffodil-japi/build.sbt
mbeckerle: So, our sbt publishLocal has no issues, so the options to the github actions are somehow different and less tolerant.

build.sbt
mbeckerle: Got that one.

project/build.properties
mbeckerle: ok

daffodil-japi/build.sbt
mbeckerle: I am reverifying that sbt publishLocal does not produce these errors when creating the doc jars.

stevedlawrence: I'm not sure if publishLocal will generate javadocs. The command to do that is ``sbt daffodil-japi/genjavadoc:doc``. But that works fine on my machine. It's not clear to me what github actions is doing differently to break this.

mbeckerle: Confirmed. The sbt publishLocal constructs javadoc jars without error. So github actions is doing this differently than our sbt build does.

olabusayoT: I think the key difference is the "coverage". We run sbt coverage ... which changes things. I ran into this before.

mbeckerle: we have a whole build.sbt in daffodil-japi which is very much about the javadoc, though I don't see any specific options there. But it does explicitly request genjavadoc-plugin 0.16 now.

stevedlawrence: Good call Lola, [DAFFODIL-2111](https://issues.apache.org/jira/browse/DAFFODIL-2111) is probably what's goin on here. Maybe we can just add ``coverageOff`` for the doc part of the test?
```
run: $SBT coverageOff daffodil-japi/genjavadoc:doc daffodil-sapi/doc
```

mbeckerle: The main.yml github action script just invokes daffodil-japi/genjavadoc:doc. 

When I do this manually it works fine. 

I have no clue where the difference in behavior comes from here.

mbeckerle: Ah.... It's because I hadn't updated the main.yml initially. I will push the updates where main.yml will have the right scala version number to go with the genjavadoc 0.16 and we'll see if it works then.

stevedlawrence: The "coverage" comes from the SBT environment variable at the top of the main.yml. It's so that every command is run with coverage enabled to ensure all the compmilation and test steps have the run thing. We should verify that sbt won't try to recompile everything if we disable coverage for just the docs.

mbeckerle: I will add the coverageOff thing.

I also found this comment in doc on genjavadoc:

    On recent version of the JDK you will have to pass the --ignore-source-errors flag to make the javadoc tool accept our output.

.github/workflows/main.yml
```diff
@@ -59,19 +60,19 @@ jobs:
       ############################################################
 
       - name: Compile
-        run: $SBT compile test:compile it:compile
+        run: $SBTNOCOV compile test:compile it:compile
```
stevedlawrence, line 63: I think we need to be careful how we do no coverage to prevent unnecessary compilation. For example, if you do the folowoing:
```
sbt compile
sbt coverage test
```
then I *think* that the first will compile everything. The second will then have to recompile everything with coverage enabled and then run the test. So we end up compiling everything twice. That was the reason why "coverage" was added to the SBT command so that we dont' do a bunch of unnessary compilation. If we switch back and forth between cov and nocov, we might end up recompiling things a bunch.

Maybe we should jus do all things that don't do any kindof publishing first with coverage enabled, then at then do all the things that do publishing and require no coverage at the en. Then we at least only duplicate compilation once.

This is assuming that there'es a noticable different in time it takes to finish all the github actions. If it still takes about thes same amount of time, it might not be worth the effort.

mbeckerle: So the compile step still uses the coverage option. It's just the doc, and packaging steps that use the no-coverage option.

stevedlawrence: Ah you're right, looks like you do all the nocov stuff first. This should be fine.

mbeckerle: No I'm not sure. The basic compile step should be SBT with coverage I believe. Only the doc, package steps should use SBTNOCOV.

stevedlawrence: Yeah, maybe that's more correct. Maybe just move the doc and package steps to the end with nocov?

olabusayoT: I think moving it to the end might cause issues for the publishing of the coverage results. That's one of the last things to run because it wants to ensure everything passes before we get to that step, correct?

mbeckerle: Let's leave it the way it is now. It does all build steps, including packaging and doc gen, then it does all testing. I think that is sensible given how time consuming the testing is.

PR #351, opened 2020-04-06
Update gpg-agent timeout

olabusayoT:
-- timeout updated to 30000s

DAFFODIL-2312

### DIFF
--- a/containers/release-candidate/daffodil-release-candidate
+++ b/containers/release-candidate/daffodil-release-candidate
@@ -61,6 +61,12 @@ read -s -p "Apache Password: " APACHE_PASSWD
 echo
 echo
 
+gpg-agent --daemon --default-cache-ttl 3000 --max-cache-ttl 3000
+if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


containers/release-candidate/daffodil-release-candidate
```diff
@@ -61,6 +61,12 @@ read -s -p "Apache Password: " APACHE_PASSWD
 echo
 echo
 
+gpg-agent --daemon --default-cache-ttl 30000 --max-cache-ttl 30000
+if [ $? -ne 0 ]; then
+   echo -e "\n!!! Unable to change timeout of the gpg-agent !!!\n"
+   echo -e "\n!!! Try updating it manually at ~/.gnupg/gpg-agent.conf for 30000 seconds each !!!\n"
+   exit
+fi
```
stevedlawrence, line 69: Just to be safe, it might be a good idea to shut the agent down at the end of the script. That way if someone starts a build and walks away a malicious user won't be able to sign something when this finishes.

PR #354, opened 2020-04-08
Add license header to .gitignore

stevedlawrence:
DAFFODIL-2319

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


mbeckerle:
This is getting silly. A .gitignore file doesn't need a license banner does it? I have never ever seen a license banner in a .gitignore file anywhere. This is not a precedent that I think we should set. If other people are doing this, then fine I will cave. But otherwise I'd prefer to not put these banners in.

stevedlawrence:
I can't find a single apache project that has a license header in the .gitignore file. I agree that in this case it does feel excessive. I'll close and resolve as won't fix unless Apache comes out with guidance that this file in particular should inclue a license.

stevedlawrence:
Actually, I was able to find a repo with a header, and it's one the current chair of the Apache Incubator contributes to:

https://github.com/apache/mynewt-core/blob/master/.gitignore 

Perhaps it is worth adding. It doesn't hurt anything, and does remove any further discussions about it.

PR #355, opened 2020-04-08
Remove SHA256 hash, SHA512 is sufficient

stevedlawrence:
DAFFODIL-2316


stevedlawrence:
Also see https://github.com/apache/incubator-daffodil-site/pull/22 for changes to the site

PR #356, opened 2020-04-08
Define sbt options required by the project

stevedlawrence:
Rather than providing instructions on how to increase sbt memory
options, which may be different depending on a users operating system,
just create an .sbtopts file in the project root with an appropriate
amount of memory. Now when sbt is run from this directory, it will
always be started with enough memory. 4GB of memory seems to be plenty
to compile and run all tests and is a reasonable amount to expect a user
to have.

DAFFODIL-2217

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,16 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


.sbtopts
```diff
@@ -0,0 +1,16 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
mbeckerle, line 1: Having a banner in this file is actually humorous.

stevedlawrence: I don't disagree. But this squashes any chance of people bringing up license issues. With all the project Apache oversees, I think it's too time consuming for them to discuss all the edge cases and eal with grey areas. So I think their mantra is along the lines of "if it can have have a license in it, then it should", even when it's kindof ridiculous like this.

PR #358, opened 2020-04-08
Require grouping and decimal separators for non-integer types

stevedlawrence:
Even if the textNumberPattern does not include decimal or grouping
separators, ICU still looks for those separators when the primType is
not an integer. If we do not require and apply the grouping and decimal
separators for these prim types, ICU will get them from the users
LANG, which means our behavior is not system environment interdependent.

To fix this, this requires and applies the decimal and grouping
separator properties for float primTypes, even if textNumberPattern does
not specify the separators.

DAFFODIL-2320

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
@@ -109,15 +109,23 @@ case class ConvertTextNumberPrim(e: ElementBase)
       case _ => (Nope, Nope)
     }
 
+    val isInt = e.primType match {
+      case PrimType.Double | PrimType.Float | PrimType.Decimal => false
+      case _ => true
+    }
+
     // If the pattern contains any of these characters, we need to set both
     // group and decimal separators, even if the pattern doesn't contain the
     // associated character. This is because even when the pattern does not
     // contain the grouping/decimal separators, ICU stills seems to take the
     // separators into account. And since ICU provides defaut values based on
-    // locales, not setting them can cause subtle locale related bugs.
+    // locales, not setting them can cause subtle locale related bugs. We must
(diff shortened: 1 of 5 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesZoned.scala
```diff
@@ -132,6 +132,7 @@ case class ConvertZonedNumberPrim(e: ElementBase)
       roundingMode,
       roundingIncrement,
       Nil,
+      true,
```
mbeckerle, line 135: Something that IDEA has been suggesting to me is to pass args using the scala convention where you can supply an argument name.  When the args have no defaults like this, and there's a flock of them, it just adds "checked documentation" to the code, because you still have to supply them all, and in the right order. But it improves readability to write 

`isInt = true,`

rather than just true floating out there.

stevedlawrence: Fixed.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -109,15 +109,23 @@ case class ConvertTextNumberPrim(e: ElementBase)
...
+    }
+
     // If the pattern contains any of these characters, we need to set both
     // group and decimal separators, even if the pattern doesn't contain the
     // associated character. This is because even when the pattern does not
     // contain the grouping/decimal separators, ICU stills seems to take the
     // separators into account. And since ICU provides defaut values based on
-    // locales, not setting them can cause subtle locale related bugs.
+    // locales, not setting them can cause subtle locale related bugs. We must
+    // also required the separators if the prim type is not an integer type,
```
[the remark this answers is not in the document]
stevedlawrence: Fixed

PR #361, opened 2020-04-17
RuntimeData objects should not carry namespaces

tuxji:
Ensure RuntimeData and derived classes get their namespaces from their
DPathCompileInfo objects, preventing redundant storage of the
DPathCompileInfo objects' namespaces.  Make sure callers like
ChoiceTermRuntine1Mixin, ElementBaseRuntime1Mixin,
SequenceTermRuntime1Mixin pass just DPathCompileInfo objects, not
namespaces as well.

DAFFODIL-1326


mbeckerle:
Unless we expect the release to be delayed extensively, we're better off not merging.

If things drag out, then a temporary branch so work can go forward on a "alternate master" branch would be a way to go, but for now I'd say let's just defer merging a few days.

PR #366, opened 2020-04-18
Converted many pass-by-name into pass-by-value for RuntimeData struct…

mbeckerle:
Converted many pass-by-name into pass-by-value for RuntimeData structures.

Other minor mods that eliminate some circularities that allow for more
by-value passing.

One of the major problems preventing more by-value passing is described in DAFFODIL-2326, which is about the passing of RuntimeData to the constructor of DPathCompileInfo.

This branch changes things that can be changed without fixing DAFFODIL-2326.

This set of changes should make incremental progress on DAFFODIL-1879.

DAFFODIL-2325

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -19,22 +19,20 @@ package org.apache.daffodil.dsom
 
 import org.apache.daffodil.equality._
 import org.apache.daffodil.processors._
-import org.apache.daffodil.infoset._
 import org.apache.daffodil.schema.annotation.props._
 import org.apache.daffodil.xml._
-import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.grammar.ElementBaseGrammarMixin
 import org.apache.daffodil.schema.annotation.props.gen._
-import org.apache.daffodil.dsom._
 import org.apache.daffodil.util.Misc
+
 import scala.xml.NamespaceBinding
 import org.apache.daffodil.util.MaybeULong
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.api.WarnID
-import java.lang.{ Integer => JInt }
-import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveOrUseNilForDefaultOrNull
+import java.lang.{Integer => JInt}
+
 import org.apache.daffodil.infoset.DataValue
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveOrUseNilForDefaultOrNull
 
@@ -420,7 +418,7 @@ trait ElementBase
       case Representation.Binary =>
         if (isComplexType || (primType != PrimType.HexBinary && primType != PrimType.AnyURI)) byteOrderEv // ensure defined
       case _ =>
-        charsetEv // ensure defined
+        encodingRaw // ensure defined
     }
     rep
   }
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
@@ -19,21 +19,16 @@ package org.apache.daffodil.dsom
 
 import scala.xml.Node
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.xml.GetAttributesMixin
 import org.apache.daffodil.xml.NS
 import org.apache.daffodil.xml.XMLUtils
-import org.apache.daffodil.processors.NonTermRuntimeData
 import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.VariableMap
 import org.apache.daffodil.processors.NonTermRuntimeData
 import org.apache.daffodil.xml.ResolvesQNames
 import org.apache.daffodil.schema.annotation.props.LookupLocation
-import org.apache.daffodil.schema.annotation.props.PropTypes
-import org.apache.daffodil.oolag.OOLAG._
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.xml.GetAttributesMixin
 import org.apache.daffodil.schema.annotation.props.PropTypes
-import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Misc
 import org.apache.daffodil.BasicComponent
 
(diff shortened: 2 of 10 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -422,7 +419,7 @@ trait ElementBase
       case Representation.Binary =>
         if (isComplexType || (primType != PrimType.HexBinary && primType != PrimType.AnyURI)) byteOrderEv // ensure defined
       case _ =>
-        charsetEv // ensure defined
+        encodingRaw // charsetEv // ensure defined
```
mbeckerle, line 422: fix comment

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -19,21 +19,10 @@ package org.apache.daffodil.dsom
...
-import org.apache.daffodil.xml.XMLUtils
-import org.apache.daffodil.processors.NonTermRuntimeData
-import org.apache.daffodil.processors.RuntimeData
-import org.apache.daffodil.processors.VariableMap
-import org.apache.daffodil.processors.NonTermRuntimeData
-import org.apache.daffodil.xml.ResolvesQNames
-import org.apache.daffodil.schema.annotation.props.LookupLocation
-import org.apache.daffodil.schema.annotation.props.PropTypes
-import org.apache.daffodil.oolag.OOLAG._
+import org.apache.daffodil.processors.{RuntimeData, VariableMap, NonTermRuntimeData}
```
mbeckerle, line 22: Revert. This is just IntelliJ IDEA folding the imports into the scala { style } instead of line by line. 

It also may be  removing unneeded ones, which is ok, but we're not consciously converting everything to the other style.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/TermEncodingMixin.scala
```diff
@@ -31,7 +31,7 @@ import org.apache.daffodil.schema.annotation.props.gen.YesNo
  */
 trait TermEncodingMixin extends KnownEncodingMixin { self: Term =>
 
-  requiredEvaluationsAlways(encodingInfo.preSerialization)
+  // requiredEvaluationsAlways(encodingInfo.preSerialization)
```
mbeckerle, line 34: Consider putting this back. Shouldn't hurt.  If not delete commented code.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
```diff
@@ -17,17 +17,10 @@
...
-import org.apache.daffodil.dsom.PrefixLengthQuasiElementDecl
-import org.apache.daffodil.schema.annotation.props.gen.AlignmentUnits
-import org.apache.daffodil.schema.annotation.props.gen.LengthKind
-import org.apache.daffodil.schema.annotation.props.gen.LengthUnits
+import org.apache.daffodil.schema.annotation.props.gen.{AlignmentUnits, LengthKind, LengthUnits}
 import org.apache.daffodil.util.Math
-import org.apache.daffodil.dsom.Root
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.dsom.QuasiElementDeclBase
+import org.apache.daffodil.dsom.{Term, ElementBase, Root, ModelGroup, QuasiElementDeclBase}
```
mbeckerle, line 23: revert import changes

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -34,10 +34,10 @@ import org.apache.daffodil.util.LogLevel
 
 trait SchemaSetRuntime1Mixin { self : SchemaSet =>
 
-  requiredEvaluationsAlways(variableMap)
+  // requiredEvaluationsAlways(variableMap)
```
mbeckerle, line 37: put back or delete commented code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -224,75 +198,36 @@ sealed class NonTermRuntimeData(
...
-    unionMemberTypes
-    repTypeRuntimeData
-    repValueSet
-    typeCalculator
-    optRepPrimType
-  }
-
-  @throws(classOf[java.io.IOException])
-  final private def writeObject(out: java.io.ObjectOutputStream): Unit = serializeObject(out)
+//  @throws(classOf[java.io.IOException])
```
mbeckerle, line 229: delete commented code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -993,22 +912,22 @@ final class ChoiceRuntimeData(
    * hook these objects into a parent-child tree without
```
mbeckerle, line 912: Fix comment. Not really about parent-child tree, but the fact that these provide access to objects which direclty or indirectly refer to this object.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestMiddleEndAttributes.scala
```diff
@@ -116,7 +115,8 @@ class TestMiddleEndAttributes {
   @Test def testStaticallyFirstWithChoice {
     val testSchema = SchemaUtils.dfdlTestSchema(
       <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
-      <dfdl:format representation="text" occursCountKind="parsed" lengthUnits="bytes" encoding="US-ASCII" initiator="" terminator="" separator="" ignoreCase="no"/>,
+      <dfdl:format representation="text" occursCountKind="parsed" lengthUnits="bytes" encoding="US-ASCII" initiator="" terminator="" separator="" ignoreCase="no"
+        alignment="1" alignmentUnits="bytes" leadingSkip="0" trailingSkip="0"/>,
```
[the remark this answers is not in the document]
mbeckerle: Well these properties are allowed to be required "from the get go" by the DFDL spec. Daffodil was not requiring them because the by-name parameters for the RunitmeData was allowing the error of not having them to be delayed until later, and at that point it got masked by some other error. 

When I removed the by-name from this, now the test fails earlier with lack of alignment and leading/strailing skip, rather than getting the expected error. Putting in these properties enables the test to continue to test what it was intended to illustrate.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
mbeckerle: I think so. One reason for the cycles was, in the past parent/child doubly-linked structure. We don't have that now so that reason isn't there any more, which is why the comment is no longer relevant.

mbeckerle:
Oops. I force-pushed this, which means it's harder to see updates where comments were made.

All the comments were fixed. All the commented code lines removed.

PR #368, opened 2020-04-22
Fix setValidation.

mbeckerle:
DAFFODIL-2328


mbeckerle:
I will modify the JIRA ticket to 3.0.0, and hold off on merging this until after 2.6.0 release. 

Though, we are getting to where we have several PRs now backed up on waiting for the release. I'm wondering if we should allow merging with the understanding that those changes are not going into 2.6.0.

olabusayoT:
> Though, we are getting to where we have several PRs now backed up on waiting for the release. I'm wondering if we should allow merging with the understanding that those changes are not going into 2.6.0.

We heard from one of the mentors, he's planning on reviewing our rc-2 soon, so I think we're in the final lap at the point

PR #369, opened 2020-04-22
Change Variables API to be mutable and implemented newVariableInstance

jadams-tresys:
I have left these as two separate commits for now to make it easier to differentiate changes.

This is still somewhat work in progress in that I have a fair amount of tests for newVariableInstance within sequences, but none for group references or choices. In theory those should just work, but I would definitely like to add some tests to verify.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
@@ -18,13 +18,23 @@
 package org.apache.daffodil.grammar
 
 import org.apache.daffodil.dsom.Term
+import org.apache.daffodil.dsom.{ DFDLNewVariableInstance, DFDLDiscriminator, DFDLAssert }
+import org.apache.daffodil.schema.annotation.props.gen.TestKind
 
 trait HasStatementsGrammarMixin extends GrammarMixin { self: Term =>
 
-  private lazy val statementGrams = statements.map { _.gram(self) }
-  // TODO: statements (but specifically not newVariableInstance) can appear on simple type definitions as well as terms.
+  // Includes setVariable as well as assert/discriminator statements that
+  // are not testKind="pattern"
+  private lazy val lowPriorityStatementGrams = lowPriorityStatements.map { _.gram(self) }
 
-  final lazy val dfdlStatementEvaluations = prod("dfdlStatementEvaluations", statementGrams.length > 0) {
-    statementGrams.fold(mt) { _ ~ _ }
+  final lazy val dfdlLowPriorityStatementEvaluations = prod("dfdlStatementEvaluations", lowPriorityStatementGrams.length > 0) {
+    lowPriorityStatementGrams.fold(mt) { _ ~ _ }
+  }
+
+  // assert/discriminator statements with testKind="pattern"
+  private lazy val patternStatementGrams = patternStatements.map { _.gram(self) }
+
+  final lazy val dfdlPatternStatementEvaluations = prod("dfdlPatternStatementEvaluations", patternStatementGrams.length > 0) {
+    patternStatementGrams.fold(mt) { _ ~ _ }
   }
 }
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
@@ -42,7 +42,11 @@ trait ModelGroupGrammarMixin
   private lazy val groupRightFraming = prod("groupRightFraming") { TrailingSkipRegion(this) }
 
   final override lazy val termContentBody = prod("termContentBody") {
-    dfdlStatementEvaluations ~ groupLeftFraming ~ groupContentWithInitiatorTerminator ~ groupRightFraming
+    // See 9.5 Evaluation Order for Statement Annotations
+    dfdlPatternStatementEvaluations ~ // Assert and Discriminator statements with testKind="pattern"
+    dfdlScopeBegin ~ // newVariableInstance
+    dfdlLowPriorityStatementEvaluations ~ // setVariable and the rest of the Assert and Discriminator statements
+    groupLeftFraming ~ groupContentWithInitiatorTerminator ~ groupRightFraming ~ dfdlScopeEnd
   }
 
   private lazy val groupContentWithInitiatorTerminator = prod("groupContentWithInitiatorTerminator") {
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
@@ -28,9 +28,10 @@ import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
 import org.apache.daffodil.infoset.RetryableException
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe.Nope
+import org.apache.daffodil.util.MStackOf
 import org.apache.daffodil.xml.{GlobalQName, UnspecifiedNamespace, NamedQName, QNameBase, RefQName}
 
-import scala.collection.immutable
+import scala.collection.mutable.Map
 
 sealed abstract class VariableState extends Serializable
 
@@ -53,22 +54,49 @@ case object VariableRead extends VariableState
 case object VariableInProcess extends VariableState
 
 /**
- * Core tuple of a pure functional "state" for variables.
+ * Core tuple of a state for variables.
  */
-case class Variable(state: VariableState, value: DataValuePrimitiveNullable, rd: VariableRuntimeData, defaultValueExpr: Maybe[CompiledExpression[AnyRef]]) extends Serializable
+case class VariableInstance(
+  var state: VariableState,
+  var value: DataValuePrimitiveNullable,
+  rd: VariableRuntimeData,
+  defaultValueExpr: Maybe[CompiledExpression[AnyRef]],
+  var priorState: VariableState = VariableUndefined,
+  var priorValue: DataValuePrimitiveNullable = DataValue.NoValue)
+  extends Serializable {
+
+  def copy(
+    state: VariableState = state,
+    value: DataValuePrimitiveNullable = value,
+    rd: VariableRuntimeData = rd,
+    defaultValueExpr: Maybe[CompiledExpression[AnyRef]] = defaultValueExpr,
+    priorState: VariableState = priorState,
+    priorValue: DataValuePrimitiveNullable = priorValue) =
+      new VariableInstance(state, value, rd, defaultValueExpr, priorState, priorValue)
+
+  def setState(s: VariableState) = {
+    this.priorState = this.state
+    this.state = s
+  }
+
+  def setValue(v: DataValuePrimitiveNullable) = {
+    this.priorValue = this.value
+    this.value = v
+  }
+
+  def reset() = {
+    Assert.invariant(this.state != VariableUndefined)
+    this.value = this.priorValue
+    this.state = this.priorState
+    this.priorValue = DataValue.NoValue
+    this.priorState = VariableUndefined
+  }
+}
 
 object VariableUtils {
 
   def setExternalVariables(currentVMap: VariableMap, bindings: Seq[Binding], referringContext: ThrowsSDE) = {
-    var newVMap = currentVMap
-    bindings.foreach { b =>
-      val vm = newVMap
-      val vqn = b.varQName
-      val vv = b.varValue
-      val res = vm.setExtVariable(vqn, vv, referringContext)
-      newVMap = res
-    }
-    newVMap
+    bindings.foreach { b => currentVMap.setExtVariable(b.varQName, b.varValue, referringContext) }
   }
 
   def convert(v: String, rd: VariableRuntimeData): DataValuePrimitive =
@@ -114,8 +142,6 @@ final class VariableBox(initialVMap: VariableMap) {
 }
 
 /**
- * Pure functional data structure for implementing DFDL's variables.
- *
  * Key concepts: DFDL variables are single-assignment. Once they have been set, they may not be set again.
  * Furthermore, they have default values, and if the default value has been read, then they may not
  * subsequently be set.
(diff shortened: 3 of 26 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -374,16 +509,111 @@
...
+
+    <tdml:document />
+
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>newVariableInstance may only be used on group reference, sequence or choice</tdml:error>
+    </tdml:errors>
+
+  </tdml:parserTestCase>
+
```
mbeckerle, line 605: I don't see tests which verifies that unwinding side-effects on newVariableInstances works properly in the presence of backtracking. 

I suggest a test with two defined variables.

You enter a scope that does NVI on each of them.

You enter another scope that does another NVI on each of them.

Then you enter a choice which sets one and reads the default of the other, then fails, backtracks, and chooses the
next branch, which does the same thing.
The final branch can succeed.

then you unwind to the enclosing scope (different variable instances), and you shoudl be alble to set and read-default-value of those instances, which weren't affected by the backtracking on the inner instances.

Then when you finally unwind all the scopes, the global instances of the variables should be able to set one and read-default from the other, as the global ones should definitely not have been affected.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -129,54 +156,66 @@ final class VariableBox(initialVMap: VariableMap) {
...
-    val optLists = optVrd.flatMap { vrd => vTable.get(vrd.globalQName) }
-    val variab = optLists.flatMap { lists => lists.flatMap {
-      _.toStream
-    }.headOption
+    val optStack = optVrd.flatMap { vrd => vTable.get(vrd.globalQName) }
+    val variab = {
+      if (optStack.isDefined)
+        Some(optStack.get.top)
+      else
+        None
```
stevedlawrence, line 193: Can this just instead just do something like:
```scala
val optStack = vTable.get(qName)
val variab = ...
```
If this uses getVariableRuntimeData it's going to do two looks up. Once in getVariableRuntimeData to find the right stack and get the rd of the the thing on the top, then another to find that stack that getVariableRuntimeData already found.

And then getVariableRuntimeData can just be
```
val optVariable = find(qName)
if (optVariable.isDefined) Some(optVariable.get.rd) else None
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -129,54 +156,66 @@ final class VariableBox(initialVMap: VariableMap) {
...
       }
     }
   }
 
-  lazy val context = Assert.invariantFailed("unused.")
-
-  private def mkVMapForNewVariable(newVar: Variable, firstTier: List[Variable], enclosingScopes: List[List[Variable]]) = {
-    val newMap = vTable + ((newVar.rd.globalQName, (newVar :: firstTier) :: enclosingScopes))
-    new VariableMap(newMap)
+  def getVaraibleStack(qName: GlobalQName): Option[MStackOf[Variable]] = {
```
stevedlawrence, line 209: Or find() and getVariableRUntimeData() could use this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -185,96 +224,84 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
...
-            (converted, vmap)
-          }
-
-          case _ => {
-            // Fix DFDL-766
-            throw new VariableHasNoValue(varQName, referringContext)
-            // Runtime error:
-            // referringContext.SDE(msg, varQName)
-          }
+    val stack = vTable.get(varQName)
```
stevedlawrence, line 230: Should we use the new getVariableStack()? Or should that function be removed? Not sure if anything uses it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -185,96 +224,84 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
...
-      case Some(Nil) => Assert.invariantFailed()
-
-      case None => {
-        // Compile time error:
-        referringContext.SDE("Variable map (compilation): unknown variable %s", varQName)
+        case _ => throw new VariableHasNoValue(varQName, referringContext)
       }
-    }
+    } else
+      referringContext.SDE("Variable map (compilation): unknown variable %s", varQName) // Fix DFDL-766
```
stevedlawrence, line 242: Not sure this comment adds anything. I'd just remove it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -185,96 +224,84 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
...
 
+  /**
+   * Creates a new instance of a variable
+   */
+  def newVariableInstance(vrd: VariableRuntimeData) = {
+    val varQName = vrd.globalQName
+    val stack = vTable.get(varQName)
+    if (stack.isDefined) {
+      stack.get.push(vrd.newVariableInstance)
+    }
```
stevedlawrence, line 280: In what case is stack not defined? Should this be an Assertion that stack.isDefined instead

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -185,96 +224,84 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
...
+  }
+
+  def removeVariableInstance(vrd: VariableRuntimeData) {
+    val varQName = vrd.globalQName
+    val stack = vTable.get(varQName)
+    if (stack.isDefined) {
+      stack.get.pop
+    }
```
stevedlawrence, line 288: Same here, should there be an assertion that stack is defined?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -185,96 +224,84 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
...
+
   private lazy val externalVarGlobalQNames: Seq[GlobalQName] =
     vTable.map { pair => pair match {
-      case (gqn, firstTier :: _) => firstTier match {
-        case (variable@Variable(_, v, ctxt, _)) :: _ if (ctxt.external) => variable.rd.globalQName
+      case (gqn, stack) if (stack.top.rd.external) => {
+        stack.top.rd.globalQName
       }
-      case (_, Nil) => Assert.invariantFailed("empty tier")
+      case (_, stack) if (stack.isEmpty) => Assert.invariantFailed("empty tier")
```
stevedlawrence, line 297: Does stack.top throw an exception if the stack is empty? Seems like it's never possible to hit this case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -340,6 +393,6 @@ class VariableMap private(vTable: Map[GlobalQName, List[List[Variable]]])
       }
       case x => Assert.invariantFailed("variables data structure not as expected. Should not be " + x)
     }
-  }
+  }*/
```
stevedlawrence, line 396: Remove comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -158,25 +158,24 @@ class SetVariableParser(expr: CompiledExpression[AnyRef], decl: VariableRuntimeD
...
-  override val context: RuntimeData)
+class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
   extends PrimParser {
-  context.notYetImplemented("newVariableInstance")
   override lazy val runtimeDependencies = Vector()
 
-  def parse(pstate: PState) = {
-    context.notYetImplemented("newVariableInstance")
+  def parse(start: PState): Unit = {
+    log(LogLevel.Debug, "This is %s", toString) // important. Don't toString unless we have to log.
```
stevedlawrence, line 166: Remove ebug message. I don't think this provides anything extra.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -158,25 +158,24 @@ class SetVariableParser(expr: CompiledExpression[AnyRef], decl: VariableRuntimeD
...
+    if (start.processorStatus.isInstanceOf[Failure]) return
+    // val vmap = start.variableMap
+    start.newVariableInstance(context)
```
stevedlawrence, line 169: In what case does the previous processor fail but we continue to call this one? If we do need thi heck, can you remove the return, so it's something like
```
if (start.processorStatus == Success) {
  start.newVariableInstance(contxt)
}
```

Can you remove the comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -166,6 +168,9 @@ final class PState private (
   private val discriminatorStack = MStackOfBoolean()
   discriminatorStack.push(false)
 
+  private val changedVariablesStack = new MStackOf[mutable.MutableList[GlobalQName]]()
```
stevedlawrence, line 171: Can you add a comment describing what this is? It's not clear what a stack of lists of qnames is used for.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -265,17 +276,29 @@ final class PState private (
   }
 
   def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: VariableRuntimeData, pstate: PState) {
-    this.setVariableMap(variableMap.setVariable(vrd, newValue, referringContext, pstate))
+    variableMap.setVariable(vrd, newValue, referringContext, pstate)
+    changedVariablesStack.top += vrd.globalQName
+  }
```
stevedlawrence, line 281: Do we also need to modify changedVariableStack if a variable is read, since that does change the state of the variable? Or is resetting that handled somehow else?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -53,28 +54,54 @@ case object VariableRead extends VariableState
...
+  extends Serializable {
+
+  def copy(
+    state: VariableState = state,
+    value: DataValuePrimitiveNullable = value,
+    rd: VariableRuntimeData = rd,
+    defaultValueExpr: Maybe[CompiledExpression[AnyRef]] = defaultValueExpr,
+    priorState: VariableState = priorState,
+    priorValue: DataValuePrimitiveNullable = priorValue) =
+      new Variable(state, value, rd, defaultValueExpr, priorState, priorValue)
```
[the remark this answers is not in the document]
mbeckerle: Some of the slots are mutable. Can't have a var in a case class I believe.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -205,6 +210,12 @@ final class PState private (
     m.restoreInto(this)
     m.clear()
     markPool.returnToPool(m)
+    changedVariablesStack.top.foreach { v => {
+      val variable = variableMap.find(v)
+      if (variable.isDefined)
+        variable.get.reset
+    }}
+    changedVariablesStack.top.clear
```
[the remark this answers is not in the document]
jadams-tresys: You are right, this clear is unnecessary, will remove

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: I think I'm in favour of just removing it as nothing ended up using it and it isn't really providing any additional functionality beyond the vTable.get() call.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
```diff
@@ -18,10 +18,11 @@
 package org.apache.daffodil.grammar
 
 import org.apache.daffodil.dsom.Term
+import org.apache.daffodil.dsom.DFDLNewVariableInstance
 
 trait HasStatementsGrammarMixin extends GrammarMixin { self: Term =>
 
-  private lazy val statementGrams = statements.map { _.gram(self) }
+  private lazy val statementGrams = statements.filter{ st => !st.isInstanceOf[DFDLNewVariableInstance] }.map { _.gram(self) }
```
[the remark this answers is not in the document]
mbeckerle: I think of a nve as a statement. So we should either rename statementGrams, or figure out whether NVE should now be included. I think it was filtered out as part of not being implemented before. We still had these combinators (grammar) being generated before, just no parsers/unparsers beyond that.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -129,54 +156,66 @@ final class VariableBox(initialVMap: VariableMap) {
...
+        val stack = new MStackOf[Variable]
+        stack.push(variab)
+        (vrd.globalQName, stack)
+    }: _*))
 
   override def toString(): String = {
     "VariableMap(" + vTable.mkString(" | ") + ")"
   }
 
+  def copy(): VariableMap = {
```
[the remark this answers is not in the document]
jadams-tresys: I've updated this to now perform a full deep copy of the entire stack for each variable. With how it is currently being called from the original DataProcessor's variable map there should only be one variable in each stack, but having a full copy function could prove useful in the future anyways.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
jadams-tresys: Good catch. I had missed this and it wasn't as easy as I expected to keep track of these variable reads. The code handling variable reads is in DState, which doesn't have direct access to things in PState, but I think I made a relatively clean fix for that.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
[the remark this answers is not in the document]
jadams-tresys: So, the other possible statements are assert, discriminator, and setVariable. The reason NVI is filtered out is because they need to be processed before the other statements, or at least before setVariable statements. The test that Mike suggested elsewhere in the review actually triggered a bug as I was processing NVI after other statements.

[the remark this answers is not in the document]
mbeckerle: So then I suggest renaming these to the beforehandStatementGrams, and/or afterwardsStatementGrams to make the timing distinction clearer. A comment citing that part of the spec is always good.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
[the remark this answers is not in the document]
jadams-tresys: I've currently left this in using MStackOf as pretty much all of the rest of the state tracking done in PState is also using stacks, so it felt a little more cohesive.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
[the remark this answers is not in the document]
jadams-tresys: Actually, according to the spec we probably aren't doing the evaluation order of statement annotations correctly. Before NVI are supposed to be processed we are supposed to evaluate discriminators/asserts with test="pattern", but I think that may require significant changes and is outside the scope of this pull request.
I'll make the suggested changes of adding more clear names and comments.

jadams-tresys:
I believe I have addressed all comments in the review now, and fixed a bug that some of the new tests exposed.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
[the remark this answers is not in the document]
stevedlawrence: I think the parser order is right. In ElementCombinator1.scala, the parse() method first does pattern discrim/asserts. Then it does parseBegin which just creates the infoset element. Then it does eBeforeParser which is the NVE (and framing), and so on. So I think we're doing pattern stuff first.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
```diff
@@ -42,7 +42,7 @@ trait ModelGroupGrammarMixin
   private lazy val groupRightFraming = prod("groupRightFraming") { TrailingSkipRegion(this) }
 
   final override lazy val termContentBody = prod("termContentBody") {
-    dfdlStatementEvaluations ~ groupLeftFraming ~ groupContentWithInitiatorTerminator ~ groupRightFraming
+    dfdlScopeBegin ~ dfdlStatementEvaluations ~ groupLeftFraming ~ groupContentWithInitiatorTerminator ~ groupRightFraming ~ dfdlScopeEnd
```
stevedlawrence, line 45: Mabye this is the ordering you're talking about is incorrect? Do we need to split up dfdlStatementEvaluations into pattern and not pattern here? So the issue is on model groups and not elements?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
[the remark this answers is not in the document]
jadams-tresys: Ok, good to know.  I was just going off of the order the grammar seemed to be getting laid out in.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ModelGroupGrammarMixin.scala
jadams-tresys: Yes, this is where I was looking at things. So I guess we do need a small fix here to do discriminators/asserts with test="pattern" first.

jadams-tresys: Okay, so I've made a change to make this reflect what is described in the spec and process statements with patterns before newVariableInstance. I added a test that I think demonstrates this, but I am struggling to think of a case where an assert or discriminator statement could affect a newVariableInstance, especially if you can't use a variable to store the pattern used by the assert/discriminator. I tried to add a test doing that, but was unsuccessful.

mbeckerle: Hmm. An assert has a pattern and a message. Suppose the pattern fails. The message could be an expression that reads a variable defined by NVI with a default value. That would make it an error to subsequently set the variable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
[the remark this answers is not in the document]
mbeckerle: I think it's fine this way. It was a minor observation really.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/HasStatementsGrammarMixin.scala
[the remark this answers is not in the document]
mbeckerle: Yeah, at one time I thought the grammar would implicitly orchestrate the order of all these things, and that was the original intent as you've recognized. 

But over time we've moved toward a style of just passing a bunch of Gram objects to a combinator which orchestrates them internally, differentiates parse from unparse behavior accordingly, etc.  This has the downside of every combinator having to deal with optimizing away in case of empty Gram objects for the various args, but the advantage of the resulting stack of combinators being better motivated than a bunch of sequencing operations only.

PR #371, opened 2020-04-24
Sonarcloud Fixes: Refactors duplicated blocks

olabusayoT:
-- fixes duplicated case statements
-- fixes duplicated functions
-- excludes Scala/Java API files from duplication analysis

DAFFODIL-2275

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
@@ -124,16 +124,7 @@ abstract class SequenceChild(protected val sq: SequenceTermBase, child: Term, gr
     import SeparatedSequenceChildBehavior._
     child match {
       case m: ModelGroup => {
-        if (child.isPotentiallyTrailing) {
-          ssp match {
-            case AnyEmpty => NonPositional
-            case TrailingEmpty => PositionalTrailingLax
-            case TrailingEmptyStrict => PositionalTrailingStrict
-            case Never => Positional
-          }
-        } else {
-          Positional
-        }
+        handleIfPotentiallyTrailingElseDefaultBehavior
       }
       case eb: ElementBase if (eb.isArray || eb.isOptional) => {
         import OccursCountKind._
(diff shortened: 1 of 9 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
```diff
@@ -123,7 +123,7 @@ abstract class SequenceChild(protected val sq: SequenceTermBase, child: Term, gr
     import SeparatorSuppressionPolicy._
     import SeparatedSequenceChildBehavior._
     child match {
-      case m: ModelGroup => {
+      case ch @  (_: ModelGroup | _: ElementBase) if ch.isInstanceOf[ModelGroup] || ch.isScalar => {
```
stevedlawrence, line 126: I personally find this case more difficult to parse out. Maybe move the content of these two cases in a lazy val or def and referece it in the cases?

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LineFoldedTransformer.scala
```diff
@@ -296,10 +296,6 @@ class LineFoldedInputStream(mode: LineFoldMode, jis: InputStream)
         case GotCRLF => {
           c = jis.read()
           c match {
-            case -1 => {
-              state = Buf2 // buffering up the LF
-              return '\r'
-            }
```
stevedlawrence, line 302: Although the code here is the same as the default case at the end, the comment is different. Can you adjust the comment of the default case so it mentions what this case is checking for?, something like "CRLF followed by other not sp/tab, or end of data. Buffer up to the LF"

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
```diff
@@ -271,11 +271,7 @@ trait EncoderDecoderMixin
...
                 truncString
               }
-              case _ if cr.isMalformed() => {
-                cr.throwException()
-                ""
-              }
-              case _ if cr.isUnmappable() => {
+              case _ if cr.isMalformed() || cr.isUnmappable=> {
                 cr.throwException()
                 ""
```
mbeckerle, line 276: cr.throwException's type is wrong if these clauses need this "" at the end. cr.throwException() should have return type Nothing, which should be fine as the return type.

sonar-project.properties
```diff
@@ -19,6 +19,8 @@ sonar.projectKey=apache_incubator-daffodil
 sonar.modules=daffodil-cli,daffodil-core,daffodil-io,daffodil-japi,daffodil-lib,daffodil-macro-lib,daffodil-propgen,daffodil-runtime1,daffodil-runtime1-unparser,daffodil-sapi,daffodil-tdml-lib,daffodil-tdml-processor,daffodil-test,daffodil-test-ibm1,daffodil-udf
 sonar.sources=src/main
 sonar.tests=src/it,src/test
+# Scala/Java APIs contain a lot of duplication by their nature. Best to exclude them.
```
mbeckerle, line 22: There's a JIRA ticket DAFFODIL-1747 which, when fixed, will want to revisit this statement.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
olabusayoT: I didn't find this to be the case. This match is used to build a string, so not having every case return a string leads to s Type Mismatch Error (Required String, Found Unit).

stevedlawrence: The problem is that ``cr`` is a Java [CoderResult](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CoderResult.html), which just returns void, so Scala can't know that throwException will always throw and exception and whatever follows it doesn't matter. Returning an empty string is a codepath that is impossible to hit. Maybe instead of returning "" it should be Assert.impossible() or something? Assert.impossible does return Nothing so it should make the scala compiler happy.

olabusayoT: That did make the compiler happy!

PR #373, opened 2020-04-27
Implement fn:namespace-uri function

tuxji:
Add two new cases to FunctionCallExpression.functionObject (in file
Expression.scala) to allow expressions to call fn:namespace-uri with
zero and one arguments and get NodeInfo.AnyURI values.

Implement these calls with FNNamespaceUri0 and FNNamespaceUri1 case
classes in FNFunctions.scala.  These classes implement the calls using
the run() method instead of the computeValue() method since the
sibling case classes FNLocalName0 and FNLocalName1 (already
implemented by someone else) also do that.  Some parts tricky to
figure out were having to handle run() calls passing
DStateForConstantFolding (that's why we call dstate.fnExists() to
terminate such calls), how to get the namespace uri from the
currentNode, figuring out how to return NodeInfo.AnyURI instead of
NodeInfo.String, and figuring out what to do when fn:namespace-uri is
called with a currentValue instead of a currentNode.  (Please review
these new lines of code to make sure they seem correct.)

Add a new test case namespace_uri_07 to Functions.tdml to find out
what happens when we call fn:namespace-uri with a currentValue instead
of a currentNode.  Not sure how to test the similar situation where
fn:namespace-uri is called without a valid current node to use as "."
though.

Make sure to run the namespace_uri_03 to namespace_uri_06 test cases
in TestDFDLExpressions.scala now that fn:namespace-uri is implemented.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -1504,11 +1504,19 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
 
       case (RefQName(_, "local-name", FUNC), args) if args.length == 0 =>
         FNZeroArgExpr(functionQNameString, functionQName,
-          NodeInfo.String, NodeInfo.AnyAtomic, FNLocalName0(_, _))
+          NodeInfo.String, NodeInfo.Exists, FNLocalName0(_, _))
 
       case (RefQName(_, "local-name", FUNC), args) if args.length == 1 =>
         FNOneArgExpr(functionQNameString, functionQName, args,
-          NodeInfo.String, NodeInfo.AnyAtomic, FNLocalName1(_, _))
+          NodeInfo.String, NodeInfo.Exists, FNLocalName1(_, _))
+
+      case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 0 =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.String, NodeInfo.Exists, FNNamespaceUri0(_, _))
+
+      case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 1 =>
+        FNOneArgExpr(functionQNameString, functionQName, args,
+          NodeInfo.String, NodeInfo.Exists, FNNamespaceUri1(_, _))
 
       case (RefQName(_, "string", XSD), args) =>
         FNOneArgExpr(functionQNameString, functionQName, args,
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
@@ -43,11 +43,11 @@ import org.apache.daffodil.infoset.DataValue.DataValueBigDecimal
 import org.apache.daffodil.infoset.DataValue.DataValueBigInt
 import org.apache.daffodil.infoset.DataValue.DataValueBool
 import org.apache.daffodil.infoset.DataValue.DataValueCalendar
-import org.apache.daffodil.infoset.DataValue.DataValueInt
 import org.apache.daffodil.infoset.DataValue.DataValueLong
 import org.apache.daffodil.infoset.DataValue.DataValueNumber
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.infoset.DataValue.DataValueString
+import org.apache.daffodil.infoset.DataValue.DataValueURI
 import org.apache.daffodil.infoset.InfosetArrayIndexOutOfBoundsException
 import org.apache.daffodil.infoset.InfosetNoSuchChildElementException
 import org.apache.daffodil.infoset.InfosetNodeNotFinalException
@@ -683,26 +683,96 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
   }
 
   override def run(dstate: DState) {
-    // Save off original state, which is the original
-    // element/node that calls inputValueCalc with fn:local-name
-    //
-    val origState = dstate
+    // Save off original node, which is the original
+    // element/node that calls fn:local-name
+    val savedNode = dstate.currentNode
 
     // Execute the recipe/expression which should
-    // return a node/element whose local-name we want.
-    //
+    // return a node/element whose local-name we want
     recipe.run(dstate)
 
     val localName = dstate.currentElement.name
+    dstate.setCurrentNode(savedNode)
 
     if (localName.contains(":"))
       throw new IllegalArgumentException("fn:local-name failed. " + localName + " is not a valid NCName as it contains ':'.")
 
-    // The original state contains the node/element upon which
-    // fn:local-name was called.  This is where we should set
-    // the value.
-    //
-    origState.setCurrentValue(localName)
+    dstate.setCurrentValue(localName)
+  }
+}
+
+/**
+ * Returns the namespace URI of the expanded-QName of the context
+ * node as an xs:string value.
+ *
+ * If the context node is in no namespace, then the function returns
+ * the zero-length xs:string value instead.
+ *
+ * The following error may be raised:
+ * - If the context node is not an element, type error [err:XPTY004]
+ *
+ * This function is called when 0 arguments are provided.  We
+ * treat this as if the argument passed was "." to denote self.
+ */
+case class FNNamespaceUri0(recipe: CompiledDPath, argType: NodeInfo.Kind)
+  extends RecipeOpWithSubRecipes(recipe) {
+  override def run(dstate: DState) {
+    // Insist this is non-constant at compile time (to avoid a NPE)
+    if (dstate.isCompile)
+      throw new IllegalStateException()
+
+    // Check that the context node is really present
+    if (dstate.currentNode eq null)
+      dstate.SDE("Context node for fn:namespace-uri is not an element")
+
+    // Same as using "." to denote self.
+    val value = dstate.currentNode.namedQName.namespace.optURI match {
+      case Nope => ""
+      case uri => uri.get.toString
+    }
+    dstate.setCurrentValue(value)
+  }
+}
+
+/**
+ * Returns the namespace URI of the expanded-QName of \$arg as an
+ * xs:string value.
+ *
+ * If the element identified by \$arg is in no namespace, then the
+ * function returns the zero-length xs:string value instead.
+ *
+ * The following error may be raised:
+ * - If \$arg is not an element, type error [err:XPTY004]
+ */
+case class FNNamespaceUri1(recipe: CompiledDPath, argType: NodeInfo.Kind)
+  extends FNOneArg(recipe, argType) {
+  override def computeValue(value: DataValuePrimitive, dstate: DState) = {
+    Assert.usageError("not to be called. DPath compiler should be answering this without runtime calls.")
+  }
+
+  override def run(dstate: DState) {
+    // Insist this is non-constant at compile time (to avoid a NPE)
+    if (dstate.isCompile)
+      throw new IllegalStateException()
+
+    // Save original node that calls fn:namespace-uri and execute the
+    // recipe/expression which should return a node/element whose
+    // namespace-uri we want
+    val savedNode = dstate.currentNode
+    recipe.run(dstate)
+    val resultNode = dstate.currentNode
+    dstate.setCurrentNode(savedNode)
+
+    // Check that the expression returned an element, not a value
+    if (resultNode eq null)
+      dstate.SDE("Argument %s for fn:namespace-uri is not an element", recipe)
+
+    // Find and return the namespace-uri we want
+    val value = resultNode.namedQName.namespace.optURI match {
+      case Nope => ""
+      case uri => uri.get.toString
+    }
+    dstate.setCurrentValue(value)
   }
 }
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
@@ -11370,13 +11370,55 @@
       </tdml:dfdlInfoset>
     </tdml:infoset>
   </tdml:parserTestCase>
-  
-<!--
-    Test Name: nilled_01
-       Schema: booleanFunctions
-         Root: nilled01
-      Purpose: This test demonstrates that the fn:nilled function returns true if arg is a node with xsi:nil = true
--->
+
+  <!--
+      Test Name: namespace_uri_07
+         Schema: home_schema.dfdl.xsd
+           Root: e5
+        Purpose: This test demonstrates the namespace-uri() function causes a SDE if given a non-element arg
+  -->
+
+  <tdml:parserTestCase name="namespace_uri_07" root="e5"
+                       model="home_schema.dfdl.xsd" description="Section 23 - Functions - fn:namespace-uri() - DFDL-23-126R">
+
+    <tdml:document>
+      <tdml:documentPart type="text"></tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>fn:namespace-uri</tdml:error>
+      <tdml:error>is not an element</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+  <!--
+      Test Name: namespace_uri_08
+         Schema: home_schema.dfdl.xsd
+           Root: e6
+        Purpose: This test demonstrates the namespace-uri() function on an element with no namespace
+  -->
+
+  <tdml:parserTestCase name="namespace_uri_08" root="e6"
+                       model="home_schema.dfdl.xsd" description="Section 23 - Functions - fn:namespace-uri() - DFDL-23-126R">
+
+    <tdml:document>
+      <tdml:documentPart type="text"></tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <e6>
+          <uri xmlns=""></uri>
+        </e6>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <!--
+      Test Name: nilled_01
+         Schema: booleanFunctions
+           Root: nilled01
+        Purpose: This test demonstrates that the fn:nilled function returns true if arg is a node with xsi:nil = true
+  -->
 
   <tdml:parserTestCase name="nilled_01" root="nilled01"
     model="booleanFunctions" description="Section 23 - Functions - fn:nilled - DFDL-23-129R">
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
@@ -63,4 +63,14 @@
     </xs:complexType>
   </xs:element>
 
+  <xs:element name="e5" type="xs:string" dfdl:inputValueCalc="{ fn:namespace-uri(0) }"/>
+
+  <xs:element name="e6">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="uri" type="xs:string" dfdl:inputValueCalc="{ fn:namespace-uri(/home:e6/uri) }"/>
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+
 </xs:schema>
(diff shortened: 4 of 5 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -706,6 +706,92 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+  override def run(dstate: DState) {
+    // Hook so we can insist this is non-constant at compile time
+    dstate.fnExists()
+    // Check that we have a current node, otherwise return empty string(URI)
+    if (dstate.currentNode eq null) {
+      dstate.setCurrentValue("")
+    }
+    else {
+      // Same as using "." to denote self.
+      val value: DataValueURI = dstate.currentNode.namedQName.namespace.optURI.get
```
stevedlawrence, line 745: I think optURI is an optional because if the namespace of an element is NoNamspace then optURI is None. According to the DFDL spec, if the namespace is NoNamespace, then this should return empty string. So we probably need a match/case here, somethign like:
```scala
val value = dstate.currentNode.namedQName.namespace.optURI match {
  case Some(uri) => uri
  case None => new URI("")
}
```
Would be good to add a test case for a no namespace schema.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1510,6 +1510,14 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
         FNOneArgExpr(functionQNameString, functionQName, args,
           NodeInfo.String, NodeInfo.AnyAtomic, FNLocalName1(_, _))
 
+      case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 0 =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.AnyURI, NodeInfo.AnyAtomic, FNNamespaceUri0(_, _))
+
+      case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 1 =>
+        FNOneArgExpr(functionQNameString, functionQName, args,
+          NodeInfo.AnyURI, NodeInfo.AnyAtomic, FNNamespaceUri1(_, _))
```
stevedlawrence, line 1519: Also the argument type of NodeInfo.AnyAtomic feels wrong to me. This is what local-name uses, and I think that's wrong too. I think this should just be NodeInfo.Exists. That should ensure that a path to a node is passed in (and not a string/int/whatever) and that the node exist when run.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -706,6 +706,92 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+  extends FNOneArg(recipe, argType) {
+  override def computeValue(value: DataValuePrimitive, dstate: DState) = {
+    Assert.usageError("not to be called. DPath compiler should be answering this without runtime calls.")
+  }
+
+  override def run(dstate: DState) {
+    // Hook so we can insist this is non-constant at compile time
+    dstate.fnExists()
+    // Save off original state, which is the original
+    // element/node that calls inputValueCalc with fn:namespace-uri.
```
stevedlawrence, line 774: Again, not your fault since this is what was in FNLocalName, but inputValueCalc probably shouldn't be mentioned in this comment. fn:namespace-ui can be called from many places other than inputValueCalc.

We also shouldn't be saving dstate, but intead just save dstate.currentNode. You'll see this pattern alot:
```scala
val savedNode = dstate.currentNode
recipe.run(dstate)
val value = dstate.getWhateverStateWeNeed
dstate.setCurrentNode(savedNode)
// do somethign with value
```
I think FNLocaleName was just buggy, or at least is old and isn't following the common patterns we've been using with more recently added functions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -706,6 +706,92 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+      origState.setCurrentValue("")
+    }
+    else {
+      // The original state contains the node/element upon which
+      // fn:namespace-uri was called.  This is where we should set
+      // the value.
+      val value : DataValueURI = dstate.currentNode.namedQName.namespace.optURI.get
+      origState.setCurrentValue(value)
+    }
+  }
```
stevedlawrence, line 792: Same comments about this as for the zero arg one.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
```diff
@@ -63,4 +63,6 @@
     </xs:complexType>
   </xs:element>
 
+  <xs:element name="e5" type="xs:string" dfdl:inputValueCalc="{ fn:namespace-uri(0) }"/>
```
stevedlawrence, line 66: I think if you pass the argType as NodeInfo.Exists instead of NodeInfo.AnyAtomic this won't even compile, which I think is the behavior we want. namespace-uri should really only work when passed a node.

daffodil-test/src/test/scala/org/apache/daffodil/section23/dfdl_expressions/TestDFDLExpressions.scala
```diff
@@ -516,10 +516,11 @@ class TestDFDLExpressions {
   //@Test def test_namespace_uri_01() { runner2.runOneTest("namespace_uri_01") }
   //@Test def test_namespace_uri_02() { runner2.runOneTest("namespace_uri_02") }
   //DFDL-1114
-  //@Test def test_namespace_uri_03() { runner2.runOneTest("namespace_uri_03") }
```
stevedlawrence, line 519: Are changes still needed to get 01 and 02 to pass?

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: yes, intentional. DFDL doesn't have xs:anyURI as a type. It's just a specialization of string.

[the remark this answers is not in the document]
mbeckerle: The return type should be xs:string

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1510,6 +1510,14 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
         FNOneArgExpr(functionQNameString, functionQName, args,
           NodeInfo.String, NodeInfo.AnyAtomic, FNLocalName1(_, _))
 
+      case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 0 =>
+        FNZeroArgExpr(functionQNameString, functionQName,
+          NodeInfo.AnyURI, NodeInfo.AnyAtomic, FNNamespaceUri0(_, _))
```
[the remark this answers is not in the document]
tuxji: I checked and FNZeroArgExpr is used by fn:local-name, fn:namespace-uri, dfdlx:outputTypeCalcNextSibling, dfdlx:repTypeValue, and dfdlx:logicalTypeValue.  That's five places so I agree a new issue makes more sense than folding the changes into this issue.  I also noticed that fn:true, fn:false, and dfdl:occursIndex take no arguments but get their own special implementation classes.  All other functions in the DFDL [spec](https://daffodil.apache.org/docs/dfdl/) take arguments.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
tuxji: Good to get clarification where the DFDL spec is and that it's normative!  I'd googled and found this XPath fn:namespace-uri [spec](https://www.w3.org/TR/xpath-functions-31/#func-namespace-uri) which'd seemed more normative than anything else.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -706,6 +706,92 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+ * - If the context item is not a node, type error [err:XPTY004]
+ *
+ * This function is called when 0 arguments are provided.  We
+ * treat this as if the argument passed was "." to denote self.
+ */
+case class FNNamespaceUri0(recipe: CompiledDPath, argType: NodeInfo.Kind)
+  extends RecipeOpWithSubRecipes(recipe) {
+  override def run(dstate: DState) {
+    // Hook so we can insist this is non-constant at compile time
+    dstate.fnExists()
```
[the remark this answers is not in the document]
tuxji: To answer your question, FNLocalName0's code, which calls dstate.currentElement.name, must be working OK even when dstate is a DStateForConstantFolding object.  I was getting a NullPointerException when trying to call dstate.currentNode.namedQName in FNNamespaceUri0 because that call goes through erd in dstate.currentNode, but DStateForConstantFolding leaves erd null.  

However, I notice that the zero-arg functions in DFDLXTypeCalcFunctions are using a clearer and more logical way than fstate.fnExists() to terminate such DStateForConstantFolding calls:
```
    if (dstate.isCompile) {
      throw new IllegalStateException()
    }
```
I like that way better so I'll use it if we still need it.  Not sure if passing the argType as NodeInfo.Exists would prevent DStateForConstantFolding calls entirely, but I'll find out.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -706,6 +706,92 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+    // Check that we have a current node, otherwise return empty string(URI)
+    if (dstate.currentNode eq null) {
+      dstate.setCurrentValue("")
```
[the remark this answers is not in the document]
tuxji: I don't know if it's possible to call FNNamespaceUri0 without a currentNode, but I created a test case that calls FNNamespaceUri1 with a literal "0" and dstate.currentNode becomes null then.  The right thing to do in that situation is to return "" whenever an uri string can't be found.  However, if you're right, setting argType to NodeInfo.Exists will prevent such a test case from compiling and passing a non-node object.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
tuxji: Yes, I agree if we add the above match/case, we also need test cases for both branches.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
tuxji: Good to know, thanks!  I'll make a similar change to FNLocalName.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
tuxji: OK, I'll try that!

daffodil-test/src/test/scala/org/apache/daffodil/section23/dfdl_expressions/TestDFDLExpressions.scala
tuxji: Yes, we need additional changes to fix DAFFODIL-1101 before 01 and 02 will pass.  I already have another branch with these changes made and tested but I will make a second pull request after this one to make these additional changes easier to review.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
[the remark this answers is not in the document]
tuxji: I have opened issue [DAFFODIL-2331](https://issues.apache.org/jira/browse/DAFFODIL-2331) to remove the argType parameter from FNZeroArgExpr.  Please edit as you see fit.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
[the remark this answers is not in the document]
tuxji: I still had to terminate DStateForConstantFolding calls by checking for dstate.isCompile.  The NPE happens when calling dstate.currentNode.namedQName because currentNode is a DISimple object with a null erd value.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
tuxji: I changed the argType to NodeInfo.Exists, although a test case still was able to pass an int to fn:namespace-uri anyway.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
tuxji: I added a match/case and test case for a node in no namespace.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
tuxji: The change didn't help the test case (it still passed an int), but at least the NodeInfo.Exists argType is better documentation.

tuxji:
I've pushed another commit to address the comments.  Thank you for the very detailed comments!

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -683,26 +683,109 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+/**
+ * Returns the namespace URI of the name of \$arg as an xs:string
+ * value.
+ *
+ * If the argument is omitted, it defaults to the context item (.).
+ * The behavior of the function if the argument is omitted is
+ * exactly the same as if the context item had been passed as
+ * the argument.
+ *
+ * If the node identified by \$arg is neither an element nor an
```
mbeckerle, line 713: DFDL doesn't have attribute nodes. 

That said, there's some value in pointing out that this doc is taken from XPath. So you can either (a) remove mention of attribute nodes herein, or (b) point out this scaladoc is taken from that of the very similar XPath function.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -683,26 +683,109 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+    dstate.setCurrentValue(value)
+  }
+}
+
+/**
+ * Returns the namespace URI of the name of \$arg as an xs:string
+ * value.
+ *
+ * If the node identified by \$arg is neither an element nor an
+ * attribute node, or it is an element or attribute node whose
```
mbeckerle, line 753: Same issue here. No attribute nodes in DFDL.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
```diff
@@ -683,26 +683,109 @@ case class FNLocalName1(recipe: CompiledDPath, argType: NodeInfo.Kind)
...
+ * expanded-QName is in no namespace, then the function returns
+ * the zero-length xs:string value.
+ *
+ * Otherwise, the result will be the namespace URI of the
+ * expanded-QName of the node identified by \$arg returned as an
+ * xs:string value.
+ *
+ * The following errors may be raised when \$arg is omitted:
+ * - If the context item is absent, dynamic error [err:XPDY002]
+ * - If the context item is not a node, type error [err:XPTY004]
```
mbeckerle, line 724: This comment seems to contradict the statement above that says if the node identified by arg is neither an element nor an attribute node... zero-length string is the value. This appears to be the implemented behavior as evidenced by your test that uses element "e5" which has fn:namespace-uri(0) which by passing an int here, makes no sense whatsoever, but you've said it returns empty string in that case, which it does.

But 0 is not a node, so this second comment suggests this should be a type error. Personally, that sounds right to me. If you pass any value to this thing you should get a type error. I think. The argument should either be absent, or be a path to a node that exists.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -11370,13 +11370,55 @@
...
+
+  <tdml:parserTestCase name="namespace_uri_07" root="e5"
+                       model="home_schema.dfdl.xsd" description="Section 23 - Functions - fn:namespace-uri() - DFDL-23-126R">
+
+    <tdml:document>
+      <tdml:documentPart type="text"></tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <e5></e5>
```
mbeckerle, line 11389: I find it surprising that this produces empty string as value.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/home_schema.dfdl.xsd
mbeckerle: So this sure seems like it should be a type error to me. If this doesn't cause a type error, then there's a separate bug, because passing any value to something that reqiures a NodeInfo.Exists should cause a type error. You can create a separate ticket for that, and add a comment here mentioning that ticket id so that we can find this test associated with that bug.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
tuxji: I'll update the comments to remove any mention of attribute nodes since you're right, DFDL 1.0 doesn't have attribute nodes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
tuxji: Yes, I agree with you about the type error XPTY004.  I was just looking through the DFDL spec this morning and I saw the following paragraph in 23. Expression Language:

> DFDL implementations MUST comply with the error code behaviour in Appendix G of the XPath 2.0 spec and map these to the correct DFDL failure type. All but one of XPath's errors map to a schema definition error. The exception is XPTY0004, which is used both for static and dynamic cases of type mismatch. A static type mismatch maps to a schema definition error, whereas a dynamic type mismatch maps to a processing error. A DFDL implementation should distinguish the two kinds of XPTY0004 error if it is able to do so, but if unable it should map all XPTY0004 errors to a schema definition error.

I'll issue a dstate.SDE if dstate.currentNode is null, saying something like (please suggest any better or more Daffodil-style way to say it):

fn:namespace-uri [ context item | argument ] '%s' is not a node

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
tuxji: I will change this test case to verify that Daffodil throws a schema definition error instead of producing an empty string value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
mbeckerle: Suggest use the term "element". The only kinds of nodes in DFDL are elements. The error in this case is passing a value, not a path to an element. So the message can say "is not an element."

stevedlawrence: One other thing worth pointing out, I *think* there is a current limiation in our DPath implementation where we can't tell if a subexpression results to a node, we only keep track of the primitive type. For example, we can't tell the difference between
```
fn:namespace-uri(0)
```
and
```
fn:namespace-uri(/path/to/int/element)
```
In both cases, I think all our implementation knows at schema compile time is that the result of the argument is an int, we can't tell that one was a constant int and the other is an element with an int type. And you can't just check if the subexpression is constant, because the expression could be something like this:
```
fn:namespace-uri(
   if (somecondition) then /path/to/int/element
   else /path/to/string/element)
)
```
I wonder if to support this check statically, our NodeInfo type system needs a distinction between node types and non-node types, e.g. NodeInfo.Int vs NodeInfo.NodeInt. That sounds like a big change so a runtime check is very reasonable for this change.

mbeckerle: Agree. Runtime check is fine for this. The NodeType.Exists was supposed to be able to make this compilation-time distinction, .... but the compiled expression result type is probably not preserving enough about whether the expression is only a value (e.g.,, output of a + b is only a value), or is a path that could be evaluated for a node or a value e.g., for fn:exists, fn:namespace-uri, etc.

tuxji:
I've addressed Mike's comments and the changes are ready to be merged.

stevedlawrence:
Looks good! Thanks for the contribution!

PR #374, opened 2020-04-29
Fixed issue with type casting on Variables

jadams-tresys:
There was an issue with how integer and decimal numbers were being
interpreted from Variables. The list of primitive types we were checking
against needed to be in order of most specific to least specific as we are
searching this list to see if the current primitive is a subtype of any of the
types in allPrimTypes. The prim type "Integer" came before UnsignedInt
so when we would try to find a subtype for an unsignedInt, it would see
that unsignedInt is a subtype of Integer first and would treat end up
treating the value as a BigInteger. Similar issues were present for decimal
numbers.

DAFFODIL-2327

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
@@ -404,21 +404,27 @@ object NodeInfo extends Enum {
   val Date = PrimType.Date
   val Time = PrimType.Time
 
+  /**
+   * This list of types must be in order of most specific to least, i.e. Byte
+   * inherits from Short, which inherits from Int etc. This is becasue
+   * fromNodeInfo does a find on this list based on isSubtypeOf, which will
+   * return the first successful result.
+   */
   val allPrims = List(
     String,
-    Int,
     Byte,
     Short,
+    Int,
     Long,
-    Integer,
-    Decimal,
-    UnsignedInt,
     UnsignedByte,
     UnsignedShort,
+    UnsignedInt,
     UnsignedLong,
     NonNegativeInteger,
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -410,15 +410,15 @@ object NodeInfo extends Enum {
...
     UnsignedInt,
     UnsignedByte,
     UnsignedShort,
     UnsignedLong,
     NonNegativeInteger,
+    Integer,
     Double,
     Float,
+    Decimal,
     HexBinary,
```
stevedlawrence, line 422: So the order of this list is important? I think it would be helpful in the commit message what the problem was so looking back on this we have a better idea of why this change was made. Might also be useful to add a comment to explain why order is important.

I'd guess this is because PrimType.fromNodeInfo searches allPrimTypes looking for the first subTypeOf, and I guess some NodeInfo's are subtypes of multiple PrimTypes? So these need to be ordred from most specific to least specific, or somethign like that?

jadams-tresys: Yes, I meant to add something along those lines to the commit message, but I pushed this up right before going to sleep and my groggy brain didn't remember.
allPrimTypes definitely needs to be in order or most specific to least specific, as we are doing a find(isSubTypeof(_)), which simply returns the first successful match.

jadams-tresys: Also, looking at this with a clearer head this morning, there are a few other changes necessary. I'm going to push up a small revision with everything in appropriate order.

PR #375, opened 2020-05-01
Fixed issues with resetting variables

jadams-tresys:
There were a few issues involving resetting variables that were
discovered in some schema projects. Issues have been fixed and test
cases have been added to the test suite to verify the fixes.

DAFFODIL-2334

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
@@ -994,6 +994,6 @@ final class VariableRuntimeData(
       }
     }
 
-  def newVariableInstance: VariableInstance = VariableInstance(state, value, this, maybeDefaultValueExpr)
+  def createVariableInstance: VariableInstance = VariableInstance(state, value, this, maybeDefaultValueExpr)
 
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
@@ -47,6 +47,7 @@ import org.apache.daffodil.processors.ParseOrUnparseState
 import org.apache.daffodil.processors.ProcessorResult
 import org.apache.daffodil.processors.TermRuntimeData
 import org.apache.daffodil.processors.VariableMap
+import org.apache.daffodil.processors.VariableRead
 import org.apache.daffodil.processors.VariableRuntimeData
 import org.apache.daffodil.processors.dfa
 import org.apache.daffodil.processors.dfa.DFADelimiter
@@ -220,6 +221,13 @@ final class PState private (
       if (variable.isDefined)
         variable.get.reset
     }}
+    /* When parsing choices or unordered sequences it is necessary to clear the
+     * list at the top of the stack because it is possible for multiple
+     * PState.reset calls to occur withing the same point of uncertainty. If we
+     * do not clear the list, the changes made in failed branches of the choice
+     * will accumulate even though their effects have already been reset.
+     */
+    changedVariablesStack.top.clear
   }
 
   def discard(m: PState.Mark) {
@@ -288,7 +296,7 @@ final class PState private (
    * Note that this function does not actually read the variable, it is used
    * just to track that the variable was read in case we need to backtrack.
    */
-  def variableRead(vrd: VariableRuntimeData) {
+  def markVariableRead(vrd: VariableRuntimeData) {
     changedVariablesStack.top += vrd.globalQName
   }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
@@ -289,7 +289,6 @@ abstract class SequenceParserBase(
       val currentPos = pstate.bitPos0b
 
       val wasDiscriminatorSet = pstate.discriminator
-      if (hasPoU) pstate.popPointOfUncertainty
       //
       // Now we handle the result of the parse attempt.
       //
@@ -331,6 +330,8 @@ abstract class SequenceParserBase(
         }
         case other => Assert.invariantFailed("Unexpected parse attempt status: " + other)
       }
+
+      if (hasPoU) pstate.popPointOfUncertainty
 
     } catch {
       // Similar try/catch/finally logic for returning marks is also used in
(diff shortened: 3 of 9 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -220,6 +221,7 @@ final class PState private (
       if (variable.isDefined)
         variable.get.reset
     }}
+    changedVariablesStack.top.clear
```
mbeckerle, line 224: So we're back to having to clear this again now. Please add comment as to why? Not obvious to me that we need this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -332,6 +331,8 @@ abstract class SequenceParserBase(
         case other => Assert.invariantFailed("Unexpected parse attempt status: " + other)
       }
 
+    if (hasPoU) pstate.popPointOfUncertainty
```
mbeckerle, line 334: This looks like the unwinding bug fix. Yes?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
jadams-tresys: I have now added a comment to this. The reason why it needs to be cleared is that when parsing a choice (or potentially an unordered sequence) there can be multiple calls to PState.reset() within the same push/popPointOfUncertainty. Which means that the list on changedVariablesStack.top will continue to accumulate changes if the list is not cleared. For example, choice branch 1 reads a variable but ultimately fails calling PState.reset(), which will reset the read on the variable, but then the choice enters branch 2 without popping the POU stack. Branch 2 then sets the variable but ultimately fails too and when PState.reset() gets called it will reset the setVariable call and then try to reset the readVariable call from branch 1 since it hasn't been cleared and we are still in the same POU.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
jadams-tresys: Yes, we were popping the POU stack before calling PState.reset() and thus the PState.reset() was resetting variables outside of the POU.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -994,6 +994,6 @@ final class VariableRuntimeData(
       }
     }
 
-  def newVariableInstance: VariableInstance = VariableInstance(state, value, this, maybeDefaultValueExpr)
+  def newVariableInstance: VariableInstance = VariableInstance(state, value, this, maybeDefaultValueExpr, state, value)
```
stevedlawrence, line 997: This isn't clear to me. Why is state and value passed in as both the current state/value and the prior state/value? Seems to me when we create a new variable instance we want the prior state/value to be undefined/no value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -289,7 +297,10 @@ final class PState private (
    * just to track that the variable was read in case we need to backtrack.
    */
   def variableRead(vrd: VariableRuntimeData) {
-    changedVariablesStack.top += vrd.globalQName
+    val maybeVar = variableMap.find(vrd.globalQName)
+    Assert.invariant(maybeVar.isDefined)
+    if (maybeVar.get.state != VariableRead)
+        changedVariablesStack.top += vrd.globalQName
```
stevedlawrence, line 303: This means that there's now strict ordering between readVariable and variableRead (confusingly named, maybe a namechange is due?), correct? This function must always be called before readVariable, otherwise readVariable will change the state to VariableRead and then this will never add anything to the stack? This feels  a bit fragile--I could easily see someone changing the order for some reason and then things break. Is there maybe a way to call variableRead from readVariable?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
jadams-tresys: So this "newVariableInstance" is confusingly what creates the initial VariableInstace. It isn't really related to the feature NVI outside of being called to create a VariableInstance object. It is called whenever a variable is created. I think renaming this to createVariableInstance may relieve the confusion. As for the prior state/value I think that may have been a change I left in while debugging some of the issues. I am removing it now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
jadams-tresys: Yes, there is strict ordering for this as currently implemented. Having readVariable be managed by DState while setVariable and newVariableInstance are handled by PState makes things a bit ugly here. Would it perhaps be better to implement a corresponding readVariable call in PState and UState and have DState.VRef() call that instead?

stevedlawrence: Yeah, I think that would clean things up a bit. Thoughts on calling the PState/UState function from readVariable, since readVariable actually knows when variable state changes to VariableRead, since a read doesnt always modify that state. That would avoid this VariableRead check.

mbeckerle: When these sorts of ordering things come up, scaladoc helps, but I always try to put in some sort of a Assert.precondition() that is true only if things are being called in the right order.

jadams-tresys: Ok, will do.

jadams-tresys: The problem with calling PState from VariableMap.readVariable() is that it doesn't know anything about PState. I thought we might be able to do a referringContext.isInstanceOf[PState], but it is always either a DPathCompileInfo or DPathElementCompileInfo. I think the call will need to remain strictly ordered and from VRef()

stevedlawrence: I don't feel strongly one way or the other, but you could modify readVariable to take in ParseOrUnparseState as a parameter, which VRef has access to from the DState. I'm not sure if that's any better or worse than other options, but that's the way to do it for that option.

jadams-tresys: Not sure why I didn' t think of that. This strange connection between DState and PState with the variableMap is confusing.

jadams-tresys:
I believe I have addressed all the comments in the review and have hopefully written a more useful commit message.

PR #376, opened 2020-05-04
Fix missing argument for format specifier and ambiguous e1

tuxji:
Two parser test cases (namespace_uri_01 and namespace_uri_02, set
aside until fn:namespace-uri() gets implemented) were failing for
other reasons that need to be fixed too.  These test cases were using
root elements e1 & e2 that could be found in two namespaces (home and
first), making the lookup ambiguous.  The lookup error handling code
tried to format an exception message with two format specifiers but
only one argument, causing an exception to be thrown that was not the
original problem.

Add the missing argument for the second format specifier %s to
SchemaSet.findRootElement.

Remove the ambiguous root element lookups and add namespace prefixes
to the elements e1 & e2 in the namespace_uri_01 and namespace_uri_02
test cases within Functions.tdml.

Allow TestDFDLExpressions to run test_namespace_uri_01 and
test_namespace_uri_02 since the reasons for them failing are fixed
now.

DAFFODIL-1101


stevedlawrence:
Thanks for your contribution!

PR #378, opened 2020-05-06
Added try/catch for potential unparse exceptions

jadams-tresys:
There are a couple unhandled exceptions resulting from attempting to
unparse a string into various primitive types that was resulting in
NumberFormatExceptions and potentially IllegalArgumentExceptions. This
simply adds a try/catch to turn these exceptions into a reasonable
UnparseError.

DAFFODIL-2124

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -29,6 +29,7 @@ import scala.xml.NamespaceBinding
 import org.apache.daffodil.util.MaybeULong
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
+import org.apache.daffodil.dpath.InvalidPrimitiveDataException
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.api.WarnID
 import java.lang.{Integer => JInt}
@@ -342,7 +343,12 @@ trait ElementBase
           // will work.
           //
           val str = defaultValueAsString
-          val value = primType.fromXMLString(str)
+          val value = try {
+            primType.fromXMLString(str)
+          } catch {
+            case ipd: InvalidPrimitiveDataException =>
+              SDE("Invalid default value: %s", ipd.getMessage)
+          }
           value
         }
       dv
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
@@ -24,6 +24,7 @@ import org.apache.daffodil.cookers.IntRangeCooker
 import org.apache.daffodil.cookers.RepValueCooker
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
+import org.apache.daffodil.dpath.InvalidPrimitiveDataException
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.IdentifyTypeCalculator
 import org.apache.daffodil.processors.RepValueSet
@@ -568,7 +569,11 @@ final class EnumerationDef(
   override lazy val optRepType = parentType.optRepType
 
   lazy val enumValueRaw: String = (xml \ "@value").head.text
-  lazy val enumValueCooked: DataValuePrimitive = parentType.primType.fromXMLString(enumValueRaw)
+  lazy val enumValueCooked: DataValuePrimitive = try {
+    parentType.primType.fromXMLString(enumValueRaw)
+  } catch {
+    case e: InvalidPrimitiveDataException => SDE("Invalid data for enumeration: %s", e.getMessage)
+  }
 
   override lazy val optRepValueSet: Option[RepValueSet] = optRepValueSetFromAttribute
   lazy val logicalValueSet: RepValueSet = RepValueSetCompiler.compile(Seq(enumValueCooked), Seq())
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
@@ -56,22 +56,14 @@ import org.apache.daffodil.util.Numbers.asLong
 import org.apache.daffodil.util.Numbers.asShort
 import org.apache.daffodil.xml.XMLUtils
 
-trait NumericRangeCheck {
-  protected val rangePrim: PrimNumeric
-  final protected def checkRange(num: Number) = {
-    if (!rangePrim.isValidRange(num)) {
-      val msg = "Value is out of range for %s type: %s".format(rangePrim.asInstanceOf[PrimType].globalQName, num)
-      throw new NumberFormatException(msg)
-    }
-  }
-}
-
 case object BooleanToLong extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = JLong.valueOf(if (asBoolean(a.getAnyRef) == true) 1L else 0L)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong =
+    JLong.valueOf(if (asBoolean(a.getAnyRef) == true) 1L else 0L)
 }
 
 case object BooleanToString extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueString = if (asBoolean(a.getAnyRef) == true) "true" else "false"
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueString =
+    if (asBoolean(a.getAnyRef) == true) "true" else "false"
 }
 
 case object DateTimeToDate extends Converter {
@@ -99,38 +91,24 @@ case object DateToDateTime extends Converter {
   }
 }
 case object DecimalToInteger extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = asBigInt(a.getBigDecimal)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt =
+    PrimType.Integer.fromNumber(a.getBigDecimal).getBigInt
 }
-case object DecimalToLong extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
-    val res = a.getBigDecimal
-    checkRange(res)
-    asLong(res)
-  }
-  override protected val rangePrim = PrimType.Long
+case object DecimalToLong extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong =
+    PrimType.Long.fromNumber(a.getBigDecimal).getLong
 }
 case object DecimalToDouble extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble = {
-    val res = a.getBigDecimal
-    // TODO: why no range check here?
-    asDouble(res)
-  }
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueDouble =
+    PrimType.Double.fromNumber(a.getBigDecimal).getDouble
 }
-case object DecimalToNonNegativeInteger extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
-    val res = a.getBigDecimal
-    checkRange(res)
-    asBigInt(res)
-  }
-  override protected val rangePrim = PrimType.NonNegativeInteger
+case object DecimalToNonNegativeInteger extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt =
+    PrimType.NonNegativeInteger.fromNumber(a.getBigDecimal).getBigInt
 }
-case object DecimalToUnsignedLong extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
-    val res = a.getBigDecimal
-    checkRange(res)
-    asBigInt(res)
-  }
-  override protected val rangePrim = PrimType.UnsignedLong
+case object DecimalToUnsignedLong extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt =
+    PrimType.UnsignedLong.fromNumber(a.getBigDecimal).getBigInt
 }
 case object DecimalToBoolean extends Converter {
   override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
@@ -143,30 +121,20 @@ case object DecimalToBoolean extends Converter {
   }
 }
 case object DoubleToDecimal extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal = JBigDecimal.valueOf(a.getDouble)
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigDecimal =
+    PrimType.Decimal.fromNumber(a.getDouble).getBigDecimal
 }
 case object DoubleToFloat extends Converter {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueFloat = {
-    val res = a.getDouble
-    // TODO: why no range check here?
-    asFloat(res)
-  }
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueFloat =
+    PrimType.Float.fromNumber(a.getDouble).getFloat
 }
-case object DoubleToLong extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong = {
-    val res = a.getDouble
-    checkRange(res)
-    asLong(res)
-  }
-  override protected val rangePrim = PrimType.Long
+case object DoubleToLong extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueLong =
+    PrimType.Long.fromNumber(a.getDouble).getLong
 }
-case object DoubleToUnsignedLong extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
-    val res = a.getDouble
-    checkRange(res)
-    asBigInt(res)
-  }
-  override protected val rangePrim = PrimType.UnsignedLong
+case object DoubleToUnsignedLong extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt =
+    asBigInt(PrimType.UnsignedLong.fromNumber(a.getDouble).getAnyRef)
 }
 case object DoubleToBoolean extends Converter {
   override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBool = {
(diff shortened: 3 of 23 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -420,7 +420,14 @@ abstract class InfosetInputter
         }
       } else if (erd.outputValueCalcExpr.isEmpty) {
         val primType = elem.erd.optPrimType.get
-        val obj = primType.fromXMLString(txt)
+        val obj = try {
+          primType.fromXMLString(txt)
```
stevedlawrence, line 424: Grepping for ``fromXMLString``, there's a handful of other places where we call this without wrapping it with try/catch. I'm wondering if we have this same issue in other places and need the try/catch?

I'm also concerned that different primitive types use different logic in their fromXMLString implementation that might throw different exceptions that we might not be accounting for here. Thoughts on modifying the ``fromXMLString`` implementations to be more thorough about checking for exceptions, and then they rethrow a single exception that callers of ``fromXMLString`` should catch so they don't have to know about particulars of different prim type implementations? This is similar to [DAFFODIL-1681](https://issues.apache.org/jira/browse/DAFFODIL-1681).

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -3981,7 +3982,27 @@
...
+  -->
   
+  <tdml:unparserTestCase name="binaryInt_unparseError" root="int02" roundTrip="false"
+      model="SimpleTypes-binary" description="Section 5 Schema types - int - DFDL-5-013R">
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:int02>7089904312036126320</ex:int02>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:document>7089904312036126320</tdml:document>
```
stevedlawrence, line 3999: There shouldn't be a document if there are errors.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
jadams-tresys: I did a look through all the different prim types and it looked like we would only see NumberFormatExceptions for all the various number primitives (int/short/float/double, etc) and IllegalArgurmentExceptions for HexBinary and Date/Time/Calendar types. Although I just double checked and realized I missed the URI type, which can potentially throw a NullPointerException or URISyntaxException. You are probably right that the right move is to modify all the primitive types to handle their own exceptions and then throw a more generic exception to be caught elsewhere.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
jadams-tresys: Yep, will remove.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
mbeckerle: Yeah, we should make fromXMLString a tighter API which throws a specific exception when unable to convert. This is a case where Scala not enforcing handling of exceptions is not doing us a favor however. 

A perhaps more Scala-ish way of doing this is to make it return a Try[String] object instead of throwing, though we tend to avoid objects like Try, Option/Some, or little crud tuples, etc. in the runtime. Unfortunately we don't have our own non-allocating version of Try[T] as yet.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
mbeckerle: This is a TDML runner bug. It's not checking for this. The XML schema for TDML could be tightened up to enforce this I think. There's already a JIRA ticket about this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
mbeckerle: Consider this:
`/**
 * Scala doesn't enforce that throws be handled. Hence APIs that involve throwing
 * an exception that callers are expected to always handle are fragile because callsers
 * might not handle them, and compilation doesn't tell you that.
 *
 * This solves that problem by putting the exception handling into the type signature of
 * the return.
 *
 * Note: ResultType cannot be a class derived from Exception, in case that wasn't obvious.
 */
class TryVal[ResultType <: AnyRef, ExceptionType <: Exception] (private val thing: AnyRef) extends AnyVal {
  @inline def isValue: Boolean = this.thing.isInstanceOf[ResultType]
  @inline private def value: ResultType = this.asInstanceOf[ResultType]
  @inline def isException : Boolean = this.thing.isInstanceOf[ExceptionType]
  @inline def toss = throw this.asInstanceOf[ExceptionType]
  @inline def get: ResultType = if (isValue) value else toss
}`

mbeckerle: Nah. Nevermind. Type tags would be needed due to type erasure, etc. etc. Too complicated.

jadams-tresys: I've added some extra try/catch blocks in a few places, but not all of them. Some of the instances that I did not add try/catch to did not seem like it would necessarily need them. For example, there are multiple calls in XMLUtils, but those I believe are only really called in testing code. One place I think I do need to add something is in VariableUtils.convert(), but if I remember correctly we don't really want to throw exceptions from the runtime. Would the appropriate action be to catch the new exception and then do an assert?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -452,9 +454,30 @@ object NodeInfo extends Enum {
...
     }
 
+    trait PrimThrowsIllegalArgument {
+      def fromString(s: String): DataValuePrimitive
+      def fromXMLString(s: String): DataValuePrimitive = {
+        try {
+          fromString(s)
+        } catch {
+          case iae: IllegalArgumentException =>
+            throw new InvalidPrimitiveDataException("%s is not a valid %s".format(iae.getMessage, this))
```
stevedlawrence, line 464: Thoughts on using just ``s`` rather than ``iae.getMessage`` or ``nfe.getMessage`` and ignore whatever is in the exception? I'm concerened that the exception string isn't always going to make sense with "is not a valid %s". For example, it looks like we just use scala's ``String.toByte`` to for string to byte conversion. If a string is out of range for a byte, it results in the exception message
```
java.lang.NumberFormatException: Value out of range. Value:"256" Radix:10"
```
So this error message would become
```
Value out of range. Value:"256" Radix:10 is not a valid byte
```
which i find a little confusing. It's maybe nice that we get "out of range" information, but the radix info is just confusing and it doesn't fit well with the "is not a valid byte".

I'd also reccomend ``this.globalQName`` for the type, instead of just ``this``. That includes the XSD namespaces so makes it clear that the restriction is based on the XSD type system.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -452,9 +454,30 @@ object NodeInfo extends Enum {
       }
     }
 
+    trait PrimThrowsIllegalArgument {
```
stevedlawrence, line 457: I'm not sure this separatation between IllegalArgument and NumberFormatException gains a whole lot. It does mean we are a little more strict on what we expect to be thrown, but at the cost of duplicating code and I think making it a little harder to read. I don't think we lose much by catching both execptions in a single fromXMLString call.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -342,7 +343,12 @@ trait ElementBase
           // will work.
           //
           val str = defaultValueAsString
-          val value = primType.fromXMLString(str)
+          val value = try {
+            primType.fromXMLString(str)
+          } catch {
+            case ipd: InvalidPrimitiveDataException =>
+              SDE("The default value (%s) is not a valid %s: %s".format(str, primType, ipd.getMessage))
```
stevedlawrence, line 350: This seems like it's going to have repetitive information. We can asume InvalidPrimiteDataException has the information about the string value and the type, so we don't need to duplicated it this SDE. The only extra informaion this SDE needs to provided is that the invalid thing is a default value. How about something like this:
```suggestion
              SDE("Invalid default value: %s".format(ipd.getMessage))
```

daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes.scala
```diff
@@ -421,6 +421,10 @@ class TestSimpleTypes {
   @Test def test_int_binary_04() { runner.runOneTest("int_binary_04") }
   @Test def test_int_binary_05() { runner.runOneTest("int_binary_05") }
 
+  @Test def test_binaryInt_unparseError() { runner.runOneTest("binaryInt_unparseError") }
+  @Test def test_hexBinary_unparseError() { runner.runOneTest("hexBinary_unparseError") }
+  @Test def test_calendar_unparseError() { runner.runOneTest("calendar_unparseError") }
+
```
stevedlawrence, line 427: Suggest adding similar tests for all of our priitive types? That would be good confirmation that DAFFODIL-1681 is resolved, in addition to the original DAFFODIL-2124 (make  sure to update the commit message to reference both tickets).

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -594,7 +617,7 @@ object NodeInfo extends Enum {
     protected sealed trait NonNegativeIntegerKind extends Integer.Kind
     case object NonNegativeInteger extends PrimTypeNode(Integer, List(UnsignedLong)) with NonNegativeIntegerKind with PrimNumeric {
       type Kind = NonNegativeIntegerKind
-      override def fromXMLString(s: String): DataValueBigInt = new JBigInt(s)
+      override def fromString(s: String): DataValueBigInt = new JBigInt(s)
```
stevedlawrence, line 620: Looks like we aren't actually validating that the string is negative? We should add some range checking.

In fact, looks like all of our unsinged primitives don't validate that result of fromXMLString is unsigned. Should we add an ``isValidRange`` check in ``fromXMLString``, or maybe just have each of the unsigned prims manually check inside fromString?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
mbeckerle: If it's in the scope of a parse call, then you have to catch it and return a failed status with the cause being the caught exception. In unparser you convert the exception to a UnparseError which is then rethrown.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
mbeckerle: I always like to just construct nested exceptions by passing the inner exception as a cause to the enclosing exception, and not constructing strings at all. 

There's an idiom for deriving exception classes so that you have both string and throwable constructors for them.  We don't do this for our exception classes, but we should so that you have string and cause constructors for them. Like this:

`class MyException protected (m: String, cause: Throwable)
extends Exception(m, cause) { // or extend an exception base class, 2-arg constructor
   def this(message: String) = this(message, null)
   def this(cause: Throwable) = this(null, cause)
  ....
}`

One can even omit the def this (m: String) constructor if the real usage pattern only involves nesting an exception.

mbeckerle: Also, since you don't know if an exception has a string message or a cause object, we have Misc.getSomeMessage(e) which returns a Some[String], i.e., you never get null back. Centralizes the logic of if you don't have a message, then check if there is a cause, and does it have a message or a cause, etc. etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -452,9 +454,30 @@ object NodeInfo extends Enum {
...
+          case iae: IllegalArgumentException =>
+            throw new InvalidPrimitiveDataException("%s is not a valid %s".format(iae.getMessage, this))
+        }
+      }
+    }
+
     trait PrimNumeric { self: Numeric.Kind =>
       def isValidRange(n: Number): Boolean
       def fromNumber(n: Number): DataValueNumber
+      def fromString(s: String): DataValueNumber
```
mbeckerle, line 472: Should  fromString be protected? It is intended to override. 
fromXMLString is more explicit about the nature of the string being converted. I.e., it came from XML; hence, strings like "&#x31;" should be equivalent to "1". I don't know that it is called in any context where the string argument isn't already converted though, such that there won't be character numeric entities or other goop to deal with.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
mbeckerle: This isValidRange is called elsewhere in number parse primitives. Arguably, it could be checked here and exception thrown at the point where we synthesize the type, and then all the places currently calling isValidRange would instead have to catch the exception . 

I think the design principle here is that we don't have context to report diagnostics properly here, so we do as little checking as possible here, and provide check calls that the calling context can use, assuming that it has better contextual information for diagnostics.

mbeckerle: I'm not saying that this is a *good* design point, I think probably it would be cleaner contract if these types succeded and produced a value of the right type in the right range, or throw an appropriate exception. But that will require quite a bit of refactoring, will change some diagnostic messages, etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -420,7 +421,14 @@ abstract class InfosetInputter
...
+        } catch {
+          case ipd: InvalidPrimitiveDataException =>
+            UnparseError(One(elem.erd.schemaFileLocation), Nope, ipd.getMessage)
+          case iae: IllegalArgumentException =>
```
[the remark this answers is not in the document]
jadams-tresys: Illegal argument exception gets thrown for invalid calendar, hexBinary, or URI, so I think we need it too, unless we want to just catch everything.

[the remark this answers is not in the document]
stevedlawrence: Isn't that caugt by your new changes and turned into an InvalidPrimitiveDataException?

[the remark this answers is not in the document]
jadams-tresys: Nevermind, wasn't paying attention to which piece of code this is. You are right, no longer need IllegalArgument here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
stevedlawrence: True, we aren't consistent about this at all right now. It's probably worth putting that change off, as you're right, it probably would be a pretty large refactor.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
jadams-tresys: So I'm currently planning on leaving this in as I am now having PrimNumeric fromNumber/XMLString perform range checks too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
jadams-tresys: I am now rolling in changes for DAFFODIL-2338 into this pull request. Ended up not needing such a large refactor.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
jadams-tresys: fromString will be protected now. As far as I can tell, I believe the string is already converted. I at least haven't run into any issues in any of our tests or any of the schema tests. I will add a test to verify this though.

daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes.scala
jadams-tresys: Done

jadams-tresys:
More test coverage has been added, as well as adding more verification to deal with DAFFODIL-2338.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -517,15 +551,15 @@ object NodeInfo extends Enum {
         }
         f
       }
-      override def fromNumber(n: Number): DataValueFloat = n.floatValue
+      override def fromNumberNoCheck(n: Number): DataValueFloat = n.floatValue
```
mbeckerle, line 554: Hmmm. Scala question - If you override a protected member, but don't say protected, did you just widen things to public, or does it inherit the protectedness?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -105,8 +105,14 @@ object VariableUtils {
     bindings.foreach { b => currentVMap.setExtVariable(b.varQName, b.varValue, referringContext) }
   }
 
-  def convert(v: String, rd: VariableRuntimeData): DataValuePrimitive =
-    rd.primType.fromXMLString(v)
+  def convert(v: String, rd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive = {
+    try {
+      rd.primType.fromXMLString(v)
+    } catch {
+      case e: Throwable => referringContext.SDE("Error processing variable %s: %s".format(rd.globalQName, e.getMessage))
```
mbeckerle, line 112: Don't catch throwable here. Use narrower set of exceptions, which you should now have a good handle on.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -168,14 +168,15 @@ abstract class BinaryIntegerBaseParser(override val context: ElementRuntimeData,
...
       }
 
-    if (!primNumeric.isValidRange(num)) {
-      PE(start, "Parsed %s is out of range for type: %s",
-        context.optPrimType.get.globalQName, num)
-      return
+    val res = try {
+      primNumeric.fromNumber(num)
+    } catch {
+      case e: Throwable => {
```
mbeckerle, line 174: Here also.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NonBaseTenTextNumberParser.scala
```diff
@@ -65,13 +65,14 @@ class ConvertNonBaseTenTextNumberParser(
...
     }
 
-    if (!primNumeric.isValidRange(bi)) {
-      PE(state, "Parsed %s is out of range for the type: %s",
-        context.optPrimType.get.globalQName, bi.toString)
-      return
+    val num = try {
+      primNumeric.fromNumber(bi)
+    } catch {
+      case e: Throwable => {
```
mbeckerle, line 71: All these catch of Throwable have to change. Basically you can never catch Throwable unless you are some top-level application structure (E.g., our CLI might have an excuse to catch throwable)

If you catch throwable, then things like Java out-of-memory problems will get masked.  And in this case, captured, turned into processing errors, and they then cause backtracking..... not good.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -413,14 +416,23 @@ abstract class InfosetInputter
...
         if (txt != null && txt != "") {
           UnparseError(One(elem.erd.schemaFileLocation), Nope, "Nilled simple element %s has content", erd.namedQName.toExtendedSyntax)
         }
       } else if (erd.outputValueCalcExpr.isEmpty) {
         val primType = elem.erd.optPrimType.get
-        val obj = primType.fromXMLString(txt)
+        val obj = try {
+          primType.fromXMLString(txt)
+        } catch {
+          case ipd: InvalidPrimitiveDataException =>
```
mbeckerle, line 431: This looks like what you should be catching everywhere you have catch Throwable in the other scala files.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -342,7 +343,12 @@ trait ElementBase
           // will work.
           //
           val str = defaultValueAsString
-          val value = primType.fromXMLString(str)
+          val value = try {
+            primType.fromXMLString(str)
+          } catch {
+            case ipd: InvalidPrimitiveDataException =>
+              SDE("Invalid default value: %s".format(str, primType, ipd.getMessage))
```
mbeckerle, line 350: I think SDE takes a string and varargs. So you don't call format yourself, it does that lazily when it computes the actual message.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -568,7 +568,11 @@ final class EnumerationDef(
   override lazy val optRepType = parentType.optRepType
 
   lazy val enumValueRaw: String = (xml \ "@value").head.text
-  lazy val enumValueCooked: DataValuePrimitive = parentType.primType.fromXMLString(enumValueRaw)
+  lazy val enumValueCooked: DataValuePrimitive = try {
+    parentType.primType.fromXMLString(enumValueRaw)
+  } catch {
+    case e: Throwable => SDE("Invalid data for enumeration: %s".format(e.getMessage))
```
mbeckerle, line 574: Dont catch throwable. You have good knowledge of the types that can be thrown here. 
SDE takes a format string and varargs. Don't call format yourself.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
```diff
@@ -291,7 +291,7 @@ abstract class Converter extends RecipeOp {
       try {
         computeValue(arg, dstate)
       } catch {
-        case e: NumberFormatException => {
+        case e @ (_: NumberFormatException | _: IllegalArgumentException) => {
```
mbeckerle, line 294: Is NumberFormatException still the right thing to catch here? In NodeInfo.scala you introduce new exception classes and I expected to see those being caught here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
mbeckerle: I think I just learned that you don't need the def this(..) in the code fragment above. That's inherited from the base.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
stevedlawrence, line 350: Use primType.globalQName to get the "xs" prefix so you don't have change the below tests.

stevedlawrence: There's also only a single %s in the format string, but 3 format parameters. I think you just want ipd.getMessage.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
```diff
@@ -59,7 +59,7 @@ import org.apache.daffodil.xml.XMLUtils
 trait NumericRangeCheck {
   protected val rangePrim: PrimNumeric
   final protected def checkRange(num: Number) = {
-    if (!rangePrim.isValidRange(num)) {
+    if (!rangePrim.isValid(num)) {
       val msg = "Value is out of range for %s type: %s".format(rangePrim.asInstanceOf[PrimType].globalQName, num)
       throw new NumberFormatException(msg)
     }
```
stevedlawrence, line 65: I'm wondering if this NumericRangeCheck trait can just go away, and then we change the converters to use the primType fromNumber(), since that now does range checking? For example, I think LongToInteger becomes:
```scala
case object LongToInt extends Converter {
  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueInt = {
    PrimType.Int.fromNumber(a.getLong)
  }
}
```
Consolidating all the conversion and range checking logic into the NodeInfo primitives seems like the right thing to me.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
stevedlawrence: Just did a test. Not the behavior I expected, but the answer is it does not inherit the protectedness. It does in fact widen the visiblity.

So these overides shoudl also include ``protected``.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -452,21 +455,52 @@ object NodeInfo extends Enum {
...
+        try {
+          val num = fromString(s)
+          if (!isValid(num.getNumber))
+            throw new OutOfRangeException("Value %s is out of range for type: %s".format(s, this))
+          num
+        } catch {
+          case nfe: NumberFormatException =>
+            throw new InvalidPrimitiveDataException("%s is not a valid %s".format(nfe.getMessage, this))
+        }
+      }
```
stevedlawrence, line 491: I recommend this this.globalQName

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -413,14 +416,23 @@ abstract class InfosetInputter
         } catch {
           case ex: NonTextFoundInSimpleContentException =>
             UnparseError(One(elem.erd.schemaFileLocation), Nope, ex.getMessage())
+          case uri: URISyntaxException =>
```
stevedlawrence, line 419: I wonder why we haven't seen this before? Maybe one of your new tests found this?

This feels odd because ``getSimpleText`` doesn't do any conversion to the primitive type, that I can see. It just gets a string value from the XML/Json/whatever infoset. Any idea where this is coming from?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
```diff
@@ -413,14 +416,23 @@ abstract class InfosetInputter
...
         }
       } else if (erd.outputValueCalcExpr.isEmpty) {
         val primType = elem.erd.optPrimType.get
-        val obj = primType.fromXMLString(txt)
+        val obj = try {
+          primType.fromXMLString(txt)
+        } catch {
+          case ipd: InvalidPrimitiveDataException =>
+            UnparseError(One(elem.erd.schemaFileLocation), Nope, ipd.getMessage)
+          case oor: OutOfRangeException =>
```
stevedlawrence, line 433: Wonder if instead of OutOfRange it would just be eaiser to alwys throw InvalidPrimitivateDataExceptino, and the message just includes range information? Then fromXMLString really only ever needs to catch a single exception?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -168,14 +168,15 @@ abstract class BinaryIntegerBaseParser(override val context: ElementRuntimeData,
...
+        PE(start, e.getMessage)
```
stevedlawrence, line 175: The second arg of PE is a format string. So if we getMessage ever returns anything with a %s or something in it, then this is going to fail because itw ill expect more var args. This should instead be:
```suggestion
        PE(start, "%s", e.getMessage)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ZonedTextParsers.scala
```diff
@@ -95,17 +95,13 @@ case class ConvertZonedNumberParser(
...
+      } catch {
+        case e: Throwable => {
+          PE(start, e.getMessage)
           return
         }
-
-        // convert to proper type
-        val asNumber = primNumeric.fromNumber(num)
-
-        asNumber
```
stevedlawrence, line 108: These is a great refactoring! Gets rid of all the random places we do range checking and moves it into a single place in NodeInfo. Good addition.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -645,7 +652,7 @@
     <tdml:document><![CDATA[70000]]></tdml:document>
     <tdml:errors>
       <tdml:error>Parse Error</tdml:error>
-      <tdml:error>xs:short</tdml:error>
+      <tdml:error>short</tdml:error>
```
stevedlawrence, line 655: Using globalQName in the errors will avoid having to change all these tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
jadams-tresys: So these exceptions do not involve calls to fromXMLString/Number and still occur with the changes I have made. The computeValue calls are primarily implemented in ConverterOps.scala, such as LongToNonNegativeInteger, which are all performing very similar operations as the fromXMLString code is. Should all of these Converter classes be changed too or do we just rely on the catch here for NumberFormatException and IllegalArgumentException (which was needed for dateTime conversions)?

stevedlawrence: I mentioend in another comment some where, but I would suggest changing the converters to all use PrimType.Foo.fromNumber(...). fromNumber will handle the validation and conversion, and will throw just your new exception that can be caught here.

jadams-tresys: I must have missed that, I will start making those changes

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
jadams-tresys: Yes, fixed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
jadams-tresys: So this came from the new blob_unparseError test, which ends up calling getSimpleText in TDMLInfosetInputter, which has a special case for xs:anyURI. If the primType is AnyURI, it simply creates a new URI object and has no try/catch for invalid URI values.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
jadams-tresys: Agreed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetInputter.scala
stevedlawrence: Ah, this feel like an issue in the TDMLInfosetInputter. The getSimpleText documentation says:
```
  /**
   * Get the content of a simple type. This will only be called when the
   * current event type is StartElement and the element is a simple type. If
   * the event contains complex data, it is an error and should throw
   * NonTextFoundInSimpleContentException. If the element does not have any
   * simple content, this should return either null or the empty string.
   */
```
So I think it's wrong for getSimpleText to throw an exception other than NonTextFoundInSimpleContentException. The TDMLInfosetInputter should probably just return ``res`` if it's not a value URI and let daffodil eventually figure out that it's not valid like it does with other primitive types.

jadams-tresys: Missed this before pushing up my last commit, will fix first thing tomorrow.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -568,7 +569,11 @@ final class EnumerationDef(
   override lazy val optRepType = parentType.optRepType
 
   lazy val enumValueRaw: String = (xml \ "@value").head.text
-  lazy val enumValueCooked: DataValuePrimitive = parentType.primType.fromXMLString(enumValueRaw)
+  lazy val enumValueCooked: DataValuePrimitive = try {
+    parentType.primType.fromXMLString(enumValueRaw)
+  } catch {
+    case e: InvalidPrimitiveDataException => SDE("Invalid data for enumeration: %s".format(e.getMessage))
```
mbeckerle, line 575: Remove call to .format. Unnecessary as SDE takes a format string and varargs for the arguments to that format string.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -105,8 +106,14 @@ object VariableUtils {
     bindings.foreach { b => currentVMap.setExtVariable(b.varQName, b.varValue, referringContext) }
   }
 
-  def convert(v: String, rd: VariableRuntimeData): DataValuePrimitive =
-    rd.primType.fromXMLString(v)
+  def convert(v: String, rd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive = {
+    try {
+      rd.primType.fromXMLString(v)
+    } catch {
+      case e: InvalidPrimitiveDataException => referringContext.SDE("Error processing variable %s: %s".format(rd.globalQName, e.getMessage))
```
mbeckerle, line 113: Remove .format call.

Line is too long also.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
```diff
@@ -179,159 +136,80 @@ case object DoubleToBoolean extends Converter {
...
 }
-case object StringToUnsignedLong extends Converter with NumericRangeCheck {
-  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = {
-    val res = new JBigInt(a.getString)
-    checkRange(res)
-    res
-  }
-  override protected val rangePrim = PrimType.UnsignedLong
+case object StringToUnsignedLong extends Converter {
+  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = PrimType.UnsignedLong.fromXMLString(a.getString).getBigInt
```
stevedlawrence, line 212: I think this is a great refactoring, but the lines are now really long. Could you maybe move the PrimType... to a new line, e.g.:
```suggestion
  override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueBigInt = 
    PrimType.UnsignedLong.fromXMLString(a.getString).getBigInt
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -462,7 +463,7 @@ object NodeInfo extends Enum {
           fromString(s)
         } catch {
           case iae: IllegalArgumentException =>
-            throw new InvalidPrimitiveDataException("%s is not a valid %s".format(iae.getMessage, this))
+            throw new InvalidPrimitiveDataException("Value '%s' is not a valid %s: %s".format(s, this, iae.getMessage))
```
stevedlawrence, line 466: Missed a this.globalQName here.

jadams-tresys:
I believe I have addressed all issues found in the review now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/ConverterOps.scala
stevedlawrence: Missed addressing this comment. I think it would be worth wrapping these long lines.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -452,21 +457,54 @@ object NodeInfo extends Enum {
...
+        try {
+          val num = fromString(s)
+          if (!isValid(num.getNumber))
+            throw new InvalidPrimitiveDataException("Value '%s' is out of range for type: %s".format(s, this.globalQName))
+          num
+        } catch {
+          case nfe: NumberFormatException =>
+            throw new InvalidPrimitiveDataException("Value '%s' is not a valid %s".format(s, this.globalQName))
+        }
+      }
```
stevedlawrence, line 495: Super minor, but feels maybe a little odd to throw an exception inside a try block? Thoughs on moving it out, so something like this:
```scala
val num = try {
  fromString(s)
} catch {
   case nfd: NumberFormatException => throw new InvalidPrimitiveDataException(...)
}
if (!isValid(num.getNumber))
  throw new InvalidPrimitiveDataException(...)
num
```
Makes it more clear that the NFE catch is really just catching the fromString conversion.

jadams-tresys: Will fix.

PR #381, opened 2020-05-12
Remove excessive log messaging from CLI tests.

mbeckerle:
LogLevel.OOLAGDebug and LogLevel.Debug were being used excessively.

DAFFODIL-2187

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
@@ -410,8 +410,6 @@ class TestCLIparsing {
       shell.sendLine(String.format("echo 0,1| %s -vvv parse -s %s -r matrix -", Util.binPath, testSchemaFile))
       shell.expect(contains("[debug]"))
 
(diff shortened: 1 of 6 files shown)


mbeckerle:
> +1 should we remove the commented out code?

I am not sure we should remove these commented out log calls. They do have some value in showing where to put logging in order to debug certain things. E.g., they have indent logic for OOLAGDebugging if needed. They can be turned on/off en-masse with a bulk replace, for example. So I think it's worth it to leave them in.

stevedlawrence:
One concern with the commented out code, this will probably show up in the sonar cloud scan. Perhaps we add some like a new SuperVerboseDebug level for if you need that extra information? If there are times we expect these logs to be useful, it seems reasonable to have a way to display them without having to recompile. Especially since not all users will be in a position to recompile and we might not have access to their schemas. If they had an issue, they could turn on this super debug and it might provide some insights on the issue.

mbeckerle:
Ok, I will remove the commented out LogLevel.Debug statements that were commented out. I those are pointless. 

I will either remove or put back the LogLevel.OOLAGDebug statements. That level really is a super-verbose-debug level itself, and the requests for that debug level are removed from the CLI tests that were generating too much output anyway.

PR #383, opened 2020-05-14
Deep copy VariableMap when DataProcessor is copied

jadams-tresys:
This copy needs to happen now that VariableMap is mutable, so sharing
the VariableMap with other processors that use/set the same variables
can cause issues. Deep copying the VariableMap resolves this issue.

DAFFODIL-2342

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestExternalVariablesNew.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestExternalVariablesNew.scala
@@ -34,6 +34,9 @@ import org.apache.daffodil.Implicits._
 import org.apache.daffodil.api.UnitTestSchemaSource
 import org.apache.daffodil.xml.QName
 import org.apache.daffodil.util.Misc
+import org.apache.daffodil.infoset.ScalaXMLInfosetOutputter
+import org.apache.daffodil.io.InputSourceDataInputStream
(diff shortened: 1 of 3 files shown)


jadams-tresys:
Test has been added and another similar potential issue was discovered by the test and fixed

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestExternalVariablesNew.scala
```diff
@@ -288,4 +301,45 @@ class TestExternalVariablesNew {
...
+    val res1 = dp1.parse(input, outputter)
+    assertTrue(outputter.getResult.toString.contains("default1"))
+
+    val res2 = dp2.parse(input, outputter)
+    assertTrue(outputter.getResult.toString.contains("value1"))
+
+    val res3 = dp1.parse(input, outputter)
+    assertTrue(outputter.getResult.toString.contains("default1"))
+
+    //checkResult(dp1.variableMap, "{http://example.com}var1", "default1")
```
mbeckerle, line 340: why keep these and the definition of the checkResult method if they're not being called?

jadams-tresys: These were actually uncommented on my file, but I didn't write the file to disk before committing. New push has them uncommented.

PR #385, opened 2020-05-24
Fix initiatedContent="yes" with empty initiator

tuxji:
An element with both initiatedContent="yes" and an initiator which
accepts the empty string should be a schema definition error.  Fix the
bug and add TDML delimiter tests to check that it is fixed.

Change ContantExpression.isKnownNonEmpty in CompiledExpression1 to
check against "%ES;" and "%WSP*;" as well as "".  Note, however, that
checking for "%WSP*;" by itself breaks three other delimiter tests
since isKnownNonEmpty gets called for dfdl:terminator as well as
dfdl:initiator. Adding a check for qn.toQName == "dfdl:initiator" as
well as value == "%WSP*;" avoids breaking the other tests, but this
fix will need review to determine if it makes sense and is complete
enough.

DAFFODIL-2199

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
@@ -68,10 +68,10 @@ trait ContentValueReferencedElementInfoMixin {
  */
 abstract class CompiledExpression[+T <: AnyRef](
   val qName: NamedQName,
-  valueForDebugPrinting: AnyRef)
+  value: AnyRef)
   extends ContentValueReferencedElementInfoMixin with Serializable {
 
-  DataValue.assertValueIsNotDataValue(valueForDebugPrinting)
+  DataValue.assertValueIsNotDataValue(value)
 
   final def toBriefXML(depth: Int = -1) = {
     "'" + prettyExpr + "'"
@@ -83,17 +83,19 @@ abstract class CompiledExpression[+T <: AnyRef](
    * particularly for `Array[Byte]`. It prints a useless thing like "[@0909280".
    * Use of `stringOf` prints "Array(....)".
    */
-  lazy val prettyExpr = stringOf(valueForDebugPrinting)
+  lazy val prettyExpr = stringOf(value)
 
   /**
-   * tells us if the property is non-empty. This is true if it is a constant non-empty expression
-   * (that is, is not ""), but it is also true if it is evaluated as a runtime expression that it is
-   * not allowed to return "".
-   *
-   * Issue: are there properties which are string-valued, and where "" can in fact be returned at run time?
-   * Assumed no. This was clarified in an errata to the DFDL spec.
+   * Tells us if the expression is the constant empty string (that is, it is "").
    */
-  def isKnownNonEmpty: Boolean
+  final lazy val isConstantEmptyString = value == ""
+
+  /**
+   * Tells us if the expression can match the empty string. We know it can if the expression
+   * is a DFDL entity like %ES; or %WSP*. We do not know whether it can if it is a more
+   * complicated constant or runtime expression.
+   */
+  final lazy val isKnownCanMatchEmptyString = value == "%ES;" || value == "%WSP*;"
 
   /**
    * used to obtain a constant value.
@@ -124,7 +126,7 @@ abstract class CompiledExpression[+T <: AnyRef](
    */
   def evaluateForwardReferencing(state: ParseOrUnparseState, whereBlockedLocation: Suspension): Maybe[T]
 
-  override def toString(): String = "CompiledExpression(" + valueForDebugPrinting.toString + ")"
+  override def toString(): String = "CompiledExpression(" + value.toString + ")"
 
 }
 
@@ -142,8 +144,6 @@ final case class ConstantExpression[+T <: AnyRef](
   def targetType = kind
 
   lazy val sourceType: NodeInfo.Kind = NodeInfo.fromObject(value)
-
-  def isKnownNonEmpty = value != ""
 
   override def evaluate(state: ParseOrUnparseState) = value
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -686,8 +686,8 @@ trait ElementBase
   private def NVDP = NilValueDelimiterPolicy
   private def EVDP = EmptyValueDelimiterPolicy
 
-  protected final lazy val hasNilValueInitiator = initTermTestExpression(initiatorParseEv, nilValueDelimiterPolicy, NVDP.Both, NVDP.Initiator)
-  protected final lazy val hasNilValueTerminator = initTermTestExpression(terminatorParseEv, nilValueDelimiterPolicy, NVDP.Both, NVDP.Terminator)
+  protected final lazy val hasNilValueInitiator = hasNonEmptyDelimiter(initiatorParseEv, nilValueDelimiterPolicy, NVDP.Both, NVDP.Initiator)
+  protected final lazy val hasNilValueTerminator = hasNonEmptyDelimiter(terminatorParseEv, nilValueDelimiterPolicy, NVDP.Both, NVDP.Terminator)
 
   /**
    * We need the nil values in raw form for diagnostic messages.
@@ -729,19 +729,19 @@ trait ElementBase
       // cause a nil value to be created.
       (isDefinedNilValue && (isSimpleType && (simpleType.primType =:= PrimType.String || simpleType.primType =:= PrimType.HexBinary) && !hasESNilValue)))
 
-  final lazy val hasEmptyValueInitiator = initTermTestExpression(initiatorParseEv, emptyValueDelimiterPolicy, EVDP.Both, EVDP.Initiator)
-  final lazy val hasEmptyValueTerminator = initTermTestExpression(terminatorParseEv, emptyValueDelimiterPolicy, EVDP.Both, EVDP.Terminator)
+  final lazy val hasEmptyValueInitiator = hasNonEmptyDelimiter(initiatorParseEv, emptyValueDelimiterPolicy, EVDP.Both, EVDP.Initiator)
+  final lazy val hasEmptyValueTerminator = hasNonEmptyDelimiter(terminatorParseEv, emptyValueDelimiterPolicy, EVDP.Both, EVDP.Terminator)
 
   // See how this function takes the prop: => Any that is pass by name (aka lazy pass).
   // That allows us to not require the property to exist at all if
-  // expr.isKnownNotEmpty turns out to be false.
-  private def initTermTestExpression(expr: DelimiterParseEv, prop: => Any, true1: Any, true2: Any): Boolean = {
+  // expr.isConstantEmptyString turns out to be true.
+  private def hasNonEmptyDelimiter(expr: DelimiterParseEv, prop: => Any, true1: Any, true2: Any): Boolean = {
     // changed from a match on a 2-tuple to if-then-else logic because we don't even want to ask for
-    // prop's value at all unless the first test is true.
-    if (expr.isKnownNonEmpty)
-      if (prop == true1 || prop == true2) true
-      else false
-    else false
+    // prop's value at all unless the first test is false.
+    if (expr.isConstantEmptyString)
+      false
+    else
+      prop == true1 || prop == true2
   }
 
   /**
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/InitiatedTerminatedMixin.scala
@@ -44,13 +44,29 @@ trait InitiatedTerminatedMixin
    * True if the term has an initiator expressed on it.
    *
    * Do not confuse with the concept of the delimiter being able to match or not match zero-length data.
-   * Whether the representation of a term in the data stream "has an initiator", as in the initator
+   * Whether the representation of a term in the data stream "has an initiator", as in the initiator
    * occupies a non-zero number of bits in the data stream, is an entirely different question.
    */
   lazy val hasInitiator = {
-    val hasOne = initiatorExpr.isKnownNonEmpty
+    val hasOne = !initiatorExpr.isConstantEmptyString
     hasOne
   }
+
+  /**
+   * True if the term's initiator cannot match zero-length data. This answers the entirely different
+   * question of whether the initiator occupies a non-zero number of bits in the data stream.
+   */
+  lazy val hasNonZeroLengthInitiator = {
+    val hasOne = !initiatorExpr.isKnownCanMatchEmptyString
+    hasOne
+  }
+
+  /**
+   * True if the term is inside an immediately enclosing model group which has the initiatedContent
+   * property set to "yes". This tells us whether we need to verify that a runtime expression defining
+   * the initiator matches a non-zero number of bits in the data stream.
+   */
+  lazy val mustMatchNonZeroData = parentSaysInitiatedContent
 
   /**
    * True if the term has a terminator expressed on it.
@@ -60,8 +76,8 @@ trait InitiatedTerminatedMixin
    * occupies a non-zero number of bits, is an entirely different question.
    */
   lazy val hasTerminator = {
-    val res = terminatorExpr.isKnownNonEmpty
-    res
+    val hasOne = !terminatorExpr.isConstantEmptyString
+    hasOne
   }
 
   private lazy val isInitiatedContentChoice: Boolean = {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
@@ -354,8 +354,13 @@ abstract class ModelGroup(index: Int)
 
   lazy val initiatedContentCheck: Unit = {
     if (initiatedContent eq YesNo.Yes) {
-      groupMembers.foreach { term =>
-        term.schemaDefinitionUnless(term.hasInitiator, "Enclosing group has initiatedContent='yes', but initiator is not defined.")
+      groupMembers.foreach {
+        term => {
+          term.schemaDefinitionUnless(term.hasInitiator,
+            "Enclosing group has initiatedContent='yes', but initiator is not defined.")
+          term.schemaDefinitionUnless(term.hasNonZeroLengthInitiator,
+            "Enclosing group has initiatedContent='yes', but initiator can match zero-length data.")
+        }
       }
     }
   }
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/SequenceGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/SequenceGrammarMixin.scala
@@ -181,7 +181,7 @@ trait SequenceGrammarMixin
    * Whether the representation of a term in the data stream "has a separator", as in a specific separator
    * occupies a non-zero number of bits, is an entirely different question.
    */
-  lazy val hasSeparator = separatorParseEv.isKnownNonEmpty
+  lazy val hasSeparator = !separatorParseEv.isConstantEmptyString
 
   lazy val sequenceSeparator = prod("separator", hasSeparator) {
     delimMTA ~ SequenceSeparator(this)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/DelimiterAndEscapeRelated.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/DelimiterAndEscapeRelated.scala
@@ -30,13 +30,13 @@ import org.apache.daffodil.util.Misc
 import org.apache.daffodil.xml.XMLUtils
 
 case class DelimiterStackCombinatorSequence(sq: SequenceTermBase, body: Gram) extends Terminal(sq, !body.isEmpty) {
-  lazy val pInit = if (sq.initiatorParseEv.isKnownNonEmpty) One(sq.initiatorParseEv) else Nope
-  lazy val pSep = if (sq.hasSeparator && sq.separatorParseEv.isKnownNonEmpty) One(sq.separatorParseEv) else Nope
-  lazy val pTerm = if (sq.terminatorParseEv.isKnownNonEmpty) One(sq.terminatorParseEv) else Nope
+  lazy val pInit = if (sq.initiatorParseEv.isConstantEmptyString) Nope else One(sq.initiatorParseEv)
+  lazy val pSep = if (sq.hasSeparator && !sq.separatorParseEv.isConstantEmptyString) One(sq.separatorParseEv) else Nope
+  lazy val pTerm = if (sq.terminatorParseEv.isConstantEmptyString) Nope else One(sq.terminatorParseEv)
 
-  lazy val uInit = if (sq.initiatorParseEv.isKnownNonEmpty) One(sq.initiatorUnparseEv) else Nope
-  lazy val uSep = if (sq.hasSeparator && sq.separatorParseEv.isKnownNonEmpty) One(sq.separatorUnparseEv) else Nope
-  lazy val uTerm = if (sq.terminatorParseEv.isKnownNonEmpty) One(sq.terminatorUnparseEv) else Nope
+  lazy val uInit = if (sq.initiatorParseEv.isConstantEmptyString) Nope else One(sq.initiatorUnparseEv)
+  lazy val uSep = if (sq.hasSeparator && !sq.separatorParseEv.isConstantEmptyString) One(sq.separatorUnparseEv) else Nope
+  lazy val uTerm = if (sq.terminatorParseEv.isConstantEmptyString) Nope else One(sq.terminatorUnparseEv)
 
   lazy val parser: DaffodilParser = new DelimiterStackParser((pInit.toList ++ pSep.toList ++ pTerm.toList).toArray, sq.runtimeData, body.parser)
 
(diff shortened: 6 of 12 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -143,7 +143,7 @@ final case class ConstantExpression[+T <: AnyRef](
 
   lazy val sourceType: NodeInfo.Kind = NodeInfo.fromObject(value)
 
-  def isKnownNonEmpty = value != ""
+  def isKnownNonEmpty = value != "" && value != "%ES;" && !(value == "%WSP*;" && qn.toQNameString == "dfdl:initiator")
```
stevedlawrence, line 146: I'm not sure this is the correct fix. Looking at ``InitatedTerminatedMixin.scala``, one place where ``isKnownNonEmpty`` is used, ``hasInitiator`` and ``hasTerminator`` look like this:
```scala
/**
 * True if the term has an initiator expressed on it.
 *
 * Do not confuse with the concept of the delimiter being able to match or
 * not match zero-length data. Whether the representation of a term in the
 * data stream "has an initiator", as in the initator occupies a non-zero
 * number of bits in the data stream, is an entirely different question.
 */
  lazy val hasInitiator = {
    val hasOne = initiatorExpr.isKnownNonEmpty
    hasOne
  }
```
Based on that comment, it think ``isKnownNonEmpty`` should only ever return true if the value is an empty string.

I'm not sure of the right fix though. Maybe the EvDelimiters or DFADelimiter needs special logic to determine if a compiled DFADelimiter can match an empty string?

Note that Delimiters can be something like ``%WSP*;%WSP*;``. It's non sensical,  but I dont' think the above check would chatch that. We do have logic in the DFA compiler to reduce things like this (in DFDLDelimiter.scala) so detecting if a delimiter could match empty string might not be too bad? Just need to check if it compiles to a single Delimiter that is either ESDelim or WSPStarDelim?

mbeckerle: An idea: just push the responsibility to answer the question "can match empty string" down to a method of the DFADelimiter, and all DFADelimiters except ESDelim and WSPStarDelim answer false, those two answer true. 

I'm not sure it's that simple though. There can be escape-schemes involved I think.

tuxji: I think both Steve's and Mike's suggestion would require more time than I want to spend and a deeper understanding of the codebase than I have now.  I do know from setting a breakpoint at isKnownNonEmpty's location and tweaking its return value in the debugger that:

1) isKnownNonEmpty is called upon initiators, separators, and terminators
2) isKnownNonEmpty is called twice on the same object with the same value and qn from two call chains, the schema compiler (in a very long call chain which eventually calls hasInitiator while checking alignment in org.apache.daffodil.dsom.ElementBase#alignmentValueInBits) and the runtime parser (in org.apache.daffodil.dsom.ElementBase#initTermTestExpression)
3) if isKnownNonEmpty returns the original value in the first call chain, two new call chains will occur before the initTermTestExpression call occurs
4) if isKnownNonEmpty returns the original value in all the previous call chains, the test still passes as long as isKnownNonEmpty returns the changed value to the initTermTestExpression call (so that is the key call chain that must be modified if you don't want to change isKnownNonEmpty here)
5) checking qn == initiator as well as value == <ES or WSP* entity> prevents other delimiter tests from breaking (they call isKnownNonEmpty with a terminator that is WSP*)

I changed the isKnownNonEmpty code to always check qn as well as value for both ES and WSP* entities:

`  def isKnownNonEmpty = value != "" && !((value == "%ES;" || value == "%WSP*;") && qn.toQNameString == "dfdl:initiator")`

However, if that's not good enough, then here's a thread dump from the key initTermTestExpression call:

```
"main@1" prio=5 tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at org.apache.daffodil.dsom.ConstantExpression.isKnownNonEmpty(CompiledExpression1.scala:148)
	  at org.apache.daffodil.processors.DelimiterEvMixin.isKnownNonEmpty(EvDelimiters.scala:36)
	  at org.apache.daffodil.processors.DelimiterEvMixin.isKnownNonEmpty$(EvDelimiters.scala:36)
	  at org.apache.daffodil.processors.DelimiterParseEv.isKnownNonEmpty(EvDelimiters.scala:54)
	  at org.apache.daffodil.dsom.ElementBase.initTermTestExpression(ElementBase.scala:741)
	  at org.apache.daffodil.dsom.ElementBase.hasEmptyValueInitiator(ElementBase.scala:732)
	  at org.apache.daffodil.dsom.ElementBase.hasEmptyValueInitiator$(ElementBase.scala:732)
	  at org.apache.daffodil.dsom.LocalElementDeclBase.hasEmptyValueInitiator$lzycompute(LocalElementDecl.scala:22)
	  - locked <0x1087> (a org.apache.daffodil.dsom.LocalElementDecl)
	  at org.apache.daffodil.dsom.LocalElementDeclBase.hasEmptyValueInitiator(LocalElementDecl.scala:22)
	  at org.apache.daffodil.dsom.ElementBase.hasEmptyValueZLSyntax(ElementBase.scala:791)
	  at org.apache.daffodil.dsom.ElementBase.hasEmptyValueZLSyntax$(ElementBase.scala:791)
	  at org.apache.daffodil.dsom.LocalElementDeclBase.hasEmptyValueZLSyntax$lzycompute(LocalElementDecl.scala:22)
	  at org.apache.daffodil.dsom.LocalElementDeclBase.hasEmptyValueZLSyntax(LocalElementDecl.scala:22)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.x$2$lzycompute(SequenceChild.scala:277)
	  - locked <0x1088> (a org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.x$2(SequenceChild.scala:267)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.isEmptyRepZeroLength$lzycompute(SequenceChild.scala:267)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.isEmptyRepZeroLength(SequenceChild.scala:267)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.unsepScalarElementHelper$lzycompute(SequenceChild.scala:473)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.unsepScalarElementHelper(SequenceChild.scala:472)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.unseparatedHelper$lzycompute(SequenceChild.scala:464)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.unseparatedHelper(SequenceChild.scala:461)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.sequenceChildParser$lzycompute(SequenceChild.scala:401)
	  at org.apache.daffodil.grammar.primitives.ScalarOrderedSequenceChild.sequenceChildParser(SequenceChild.scala:384)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.parser$lzycompute(SequenceChild.scala:78)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.parser(SequenceChild.scala:78)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.optSequenceChildParser$lzycompute(SequenceChild.scala:85)
	  at org.apache.daffodil.grammar.primitives.SequenceChild.optSequenceChildParser(SequenceChild.scala:84)
	  at org.apache.daffodil.grammar.primitives.OrderedSequence.$anonfun$parser$2(SequenceCombinator.scala:61)
	  at org.apache.daffodil.grammar.primitives.OrderedSequence$$Lambda$811.521466380.apply(Unknown Source:-1)
	  at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:280)
	  at scala.collection.TraversableLike$$Lambda$86.616207929.apply(Unknown Source:-1)
	  at scala.collection.Iterator.foreach(Iterator.scala:943)
	  at scala.collection.Iterator.foreach$(Iterator.scala:943)
	  at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	  at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	  at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	  at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	  at scala.collection.TraversableLike.flatMap(TraversableLike.scala:280)
	  at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:277)
	  at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
	  at org.apache.daffodil.grammar.primitives.OrderedSequence.parser$lzycompute(SequenceCombinator.scala:61)
	  - locked <0x1089> (a org.apache.daffodil.grammar.primitives.OrderedSequence)
	  at org.apache.daffodil.grammar.primitives.OrderedSequence.parser(SequenceCombinator.scala:54)
	  at org.apache.daffodil.grammar.SeqComp.$anonfun$parserChildren$2(Grammar.scala:86)
	  at org.apache.daffodil.grammar.SeqComp$$Lambda$806.1400582895.apply(Unknown Source:-1)
	  at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:273)
	  at scala.collection.TraversableLike$$Lambda$60.1910936570.apply(Unknown Source:-1)
	  at scala.collection.immutable.List.foreach(List.scala:392)
	  at scala.collection.TraversableLike.map(TraversableLike.scala:273)
	  at scala.collection.TraversableLike.map$(TraversableLike.scala:266)
	  at scala.collection.immutable.List.map(List.scala:298)
	  at org.apache.daffodil.grammar.SeqComp.parserChildren$lzycompute(Grammar.scala:86)
	  - locked <0x108a> (a org.apache.daffodil.grammar.SeqComp)
	  at org.apache.daffodil.grammar.SeqComp.parserChildren(Grammar.scala:86)
	  at org.apache.daffodil.grammar.SeqComp.parser$lzycompute(Grammar.scala:89)
	  at org.apache.daffodil.grammar.SeqComp.parser(Grammar.scala:88)
	  at org.apache.daffodil.grammar.Prod.parser$lzycompute(Production.scala:80)
	  - locked <0x108b> (a org.apache.daffodil.grammar.Prod)
	  at org.apache.daffodil.grammar.Prod.parser(Production.scala:76)
	  at org.apache.daffodil.runtime1.GramRuntime1Mixin.maybeParser(GramRuntime1Mixin.scala:38)
	  at org.apache.daffodil.runtime1.GramRuntime1Mixin.maybeParser$(GramRuntime1Mixin.scala:35)
	  at org.apache.daffodil.grammar.Gram.maybeParser(GrammarTerm.scala:54)
	  at org.apache.daffodil.grammar.primitives.ElementCombinatorBase.eParser$lzycompute(ElementCombinator.scala:378)
	  - locked <0x108c> (a org.apache.daffodil.grammar.primitives.ElementParseAndUnspecifiedLength)
	  at org.apache.daffodil.grammar.primitives.ElementCombinatorBase.eParser(ElementCombinator.scala:378)
	  at org.apache.daffodil.grammar.primitives.ElementParseAndUnspecifiedLength.parser$lzycompute(ElementCombinator.scala:302)
	  at org.apache.daffodil.grammar.primitives.ElementParseAndUnspecifiedLength.parser(ElementCombinator.scala:291)
	  at org.apache.daffodil.grammar.primitives.ElementCombinator.parser$lzycompute(ElementCombinator.scala:92)
	  - locked <0x108d> (a org.apache.daffodil.grammar.primitives.ElementCombinator)
	  at org.apache.daffodil.grammar.primitives.ElementCombinator.parser(ElementCombinator.scala:86)
	  at org.apache.daffodil.grammar.Prod.parser$lzycompute(Production.scala:80)
	  - locked <0x108e> (a org.apache.daffodil.grammar.Prod)
	  at org.apache.daffodil.grammar.Prod.parser(Production.scala:76)
	  at org.apache.daffodil.grammar.Prod.parser$lzycompute(Production.scala:80)
	  - locked <0x108f> (a org.apache.daffodil.grammar.Prod)
	  at org.apache.daffodil.grammar.Prod.parser(Production.scala:76)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin.$anonfun$parser$1(SchemaSetRuntime1Mixin.scala:64)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin$$Lambda$797.324579982.apply(Unknown Source:-1)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.liftedTree1$1(OOLAG.scala:669)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.valueAsAny$lzycompute(OOLAG.scala:667)
	  - locked <0x1090> (a org.apache.daffodil.oolag.OOLAG$OOLAGValue)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.valueAsAny(OOLAG.scala:664)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValue.value$lzycompute(OOLAG.scala:714)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValue.value(OOLAG.scala:714)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin.parser(SchemaSetRuntime1Mixin.scala:63)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin.parser$(SchemaSetRuntime1Mixin.scala:63)
	  at org.apache.daffodil.dsom.SchemaSet.parser$lzycompute(SchemaSet.scala:66)
	  - locked <0x1091> (a org.apache.daffodil.dsom.SchemaSet)
	  at org.apache.daffodil.dsom.SchemaSet.parser(SchemaSet.scala:66)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin.$anonfun$$init$$1(SchemaSetRuntime1Mixin.scala:37)
	  at org.apache.daffodil.runtime1.SchemaSetRuntime1Mixin$$Lambda$170.1419332030.apply(Unknown Source:-1)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.liftedTree1$1(OOLAG.scala:669)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.valueAsAny$lzycompute(OOLAG.scala:667)
	  - locked <0x1092> (a org.apache.daffodil.oolag.OOLAG$OOLAGValue)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGValueBase.valueAsAny(OOLAG.scala:664)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost.$anonfun$checkErrors$2(OOLAG.scala:414)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost$$Lambda$242.1718322084.apply$mcV$sp(Unknown Source:-1)
	  at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	  at org.apache.daffodil.oolag.OOLAG$.keepGoing(OOLAG.scala:60)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost.checkErrors(OOLAG.scala:414)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost.checkErrors$(OOLAG.scala:400)
	  at org.apache.daffodil.dsom.SchemaComponentImpl.checkErrors(SchemaComponent.scala:35)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost.isError(OOLAG.scala:471)
	  at org.apache.daffodil.oolag.OOLAG$OOLAGHost.isError$(OOLAG.scala:470)
	  at org.apache.daffodil.dsom.SchemaSet.super$isError(SchemaSet.scala:485)
	  at org.apache.daffodil.dsom.SchemaSet.$anonfun$isError$3(SchemaSet.scala:485)
	  at org.apache.daffodil.dsom.SchemaSet$$Lambda$178.270661321.apply$mcZ$sp(Unknown Source:-1)
	  at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:23)
	  at org.apache.daffodil.oolag.OOLAG$.keepGoing(OOLAG.scala:60)
	  at org.apache.daffodil.dsom.SchemaSet.$anonfun$isError$1(SchemaSet.scala:476)
	  at org.apache.daffodil.dsom.SchemaSet$$Lambda$176.1172535934.apply$mcZ$sp(Unknown Source:-1)
	  at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:23)
	  at scala.util.DynamicVariable.withValue(DynamicVariable.scala:62)
	  at org.apache.daffodil.ExecutionMode$.$anonfun$usingCompilerMode$1(ExecutionMode.scala:64)
	  at org.apache.daffodil.ExecutionMode$$$Lambda$175.2116179210.apply(Unknown Source:-1)
	  at org.apache.daffodil.dsom.SchemaSet.isError(SchemaSet.scala:476)
	  at org.apache.daffodil.compiler.ProcessorFactory.isError(Compiler.scala:122)
	  at org.apache.daffodil.compiler.Compiler.org$apache$daffodil$compiler$Compiler$$compileSourceInternal(Compiler.scala:340)
	  at org.apache.daffodil.compiler.Compiler$.org$apache$daffodil$compiler$Compiler$$compileSourceSynchronizer(Compiler.scala:378)
	  - locked <0x1093> (a org.apache.daffodil.compiler.Compiler$)
	  at org.apache.daffodil.compiler.Compiler.compileSource(Compiler.scala:325)
	  at org.apache.daffodil.tdml.processor.TDMLDFDLProcessorFactory.compileProcessor(DaffodilTDMLDFDLProcessor.scala:133)
	  at org.apache.daffodil.tdml.processor.TDMLDFDLProcessorFactory.$anonfun$getProcessor$1(DaffodilTDMLDFDLProcessor.scala:154)
	  at org.apache.daffodil.tdml.processor.TDMLDFDLProcessorFactory$$Lambda$160.1202220987.apply(Unknown Source:-1)
	  at org.apache.daffodil.tdml.SchemaCache.doCompile$lzycompute$1(SchemaCache.scala:95)
	  - locked <0x1094> (a scala.runtime.LazyRef)
	  at org.apache.daffodil.tdml.SchemaCache.doCompile$1(SchemaCache.scala:95)
	  at org.apache.daffodil.tdml.SchemaCache.$anonfun$compileAndCache$1(SchemaCache.scala:108)
	  at org.apache.daffodil.tdml.SchemaCache$$Lambda$161.1545077099.apply(Unknown Source:-1)
	  at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:86)
	  at org.apache.daffodil.tdml.SchemaCache$Cache.getOrElseUpdate(SchemaCache.scala:51)
	  - locked <0x1095> (a org.apache.daffodil.tdml.SchemaCache$Cache)
	  at org.apache.daffodil.tdml.SchemaCache.compileAndCache(SchemaCache.scala:107)
	  at org.apache.daffodil.tdml.processor.TDMLDFDLProcessorFactory.getProcessor(DaffodilTDMLDFDLProcessor.scala:154)
	  at org.apache.daffodil.tdml.TestCase.run(TDMLRunner.scala:758)
	  at org.apache.daffodil.tdml.DFDLTestSuite.runOneTest(TDMLRunner.scala:390)
	  at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:145)
	  at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:151)
	  at org.apache.daffodil.section12.delimiter_properties.TestDelimiterProperties.test_emptyInitiator3(TestDelimiterProperties.scala:85)
	  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-1)
	  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	  at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	  at java.lang.reflect.Method.invoke(Method.java:566)
	  at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	  at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	  at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	  at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	  at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	  at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	  at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	  at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	  at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	  at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	  at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	  at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	  at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	  at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	  at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	  at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	  at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	  at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	  at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
```

tuxji: Ping - I'm waiting for an answer to my question in the above comment.  The question is: Does this thread stack dump help suggest where in the codebase the responsibility to answer the question "initiator can match empty string" should be implemented? This is the key call chain where the original isKnownNonEmpty function must be changed to make my tests pass (it doesn't need to be changed for all the other calls).

stevedlawrence: Sorry about the delay.

I guess my concern here is that ``isKnownNonEmpty`` now means something different. Previously, it essentially meant the initiator was constant and was equal to the empty string, i.e. ``dfdl:initiator=""``. And it was used to test exactly this condition.  For example:
```scala
  lazy val hasInitiator = {
    val hasOne = initiatorExpr.isKnownNonEmpty
    hasOne
  }
```
hasInitiator should only be true if this condition is true (contstant with empty string).

But With this change, if we have ``dfdl:initiator="%WSP*;``, then ``isKnownNonEmpty`` will now evaluate to false, and so will ``hasInitiator``. This feels off to me. Perhaps this doesn't matter based on the order things are called, but it feels fragile. It feels to me like ``isKnownNonEmpty`` should keep the logic it currently has (perhaps with a rename like isConstantEmptyString to make it more clear what it's testing), and something else needs to have the logic to fix this issue, e.g a function call canMatchEmptyString on the DFADelimiter.

This change also prevents other properties from being able to use this. Although only delimiters use this at the momement, it seems reasonable that other properties might want to know if the property is a constant empty string, but do not care about whether the value can actually consume empty no data or not. Seems like this is sort of mixing logic of two different things.

Another thought, this restriction related to initiatedContent also applies when the initiator property is runtime evaluated, for example:
```xml
<xs:element name="bar" dfdl:initiator="{ if (../foo eq 1) 'foo' else '%WSP*;' }" ... />
```
Because isKnownNonEmpty only applies to constant expressions, the initiated content check won't apply for this runtime expression.

The right fix is probably to have some logic described above for DFADelimters to determine if it can possible match the empty string, and then have the ``InitiatorParseEv`` check that logic. This would make it so the logic would be tested on both constant and runtime expressions, and would be specific to initiators.

That's probably a somewhat significant change though...

mbeckerle: The initiator can be an expression. That expression might look at variables. So it may not be able to be evaluated at all at compile time. 

So that means the best we can do is isKnownESMatching which is a heuristic to tell us the expression (or literal) for the initiator is definitely a problem for initiatedContent. Otherwise we need a runtime check. 

I suggest not changing isKnownNotEmpty, because the meaning of that can't change. It means the literal, or return value of the expression, will not be "". We disallow delimiter expressions from returning "" because you can't change the way the schema lengthKind is compiled with a runtime change, and returning "" for a delimiter is turning off the delimited nature of the schema. 

I think what we need for the initiatedContent check is two things. A compile time check 
`
lazy val isRequiredRuntimeCheckForEmptyInitiator =
if (initiatedContent == YesNo.Yes) {
    SDEUnless(isKnownNonEmpty, "The initiator property cannot be '' (empty string) when initiatedContent='yes'.")
    SDEWhen(isKnownESMatching, "The initiator property value '%s' can match the empty string, which is incompatible with initiatedContent='yes'.",  initiator)
   true
} else 
  false 
`

I'm pseudo coding here. I'm not sure about the initiator argument to the SDEWhen above. Maybe a better way to get that for inclusion in the message string. (e.g., something like initiatorExpr.prettyString or some such). 

Given the above, isKnownESMatching can make a best effort at determining if the expression, which evaluates to a delimiter that may contain entities, can then match empty string successfully. Checking for the obvious cases syntactically is fine here.  

Then if our isRequiredRuntimeCheckForEmptyInitiator is true, we have to include that check at runtime.

mbeckerle: Given this discussion, let's add a detailed scaladoc string to isKnownNotEmpty also to specify that it is about the property value being empty, not the data stream matching empty.

tuxji:
All fixed now except for runtime expressions.  Please see `TestDelimiterProperties.test_emptyInitiator4 ` for the test and help.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
tuxji: @mbeckerle, I just saw your comments after I'd already made and pushed a new commit.  I gave isKnownNotEmpty a different name (isConstantEmptyString) but I didn't change its meaning or implementation.  I defined some new canMatchEmptyString methods and added a check for term,initiatorCanBeZeroLength to ModelGroup.initiatedContentCheck.  However, I could use help finding the right place to check initiators which are expressions (it seems to be a different place which I haven't found yet).

mbeckerle: Take a look at the Initiator class is in PrimitivesDelimiters.scala 

This class just needs another argument to the constructor to indicate whether the resulting match is allowed to be an empty string or not. That will result in that boolean (assume to be called canMatchEmptyString, for consistency) being passed down in to the DelimiterTextParser class, in DelimiterParsers.scala. I think we need to check for nChars == 0 and this new boolean arg being true around line 99 of that file, and if so it's a runtimeSDE.

tuxji: I'm not sure org.apache.daffodil.processors.parsers.DelimiterTextParser#parse is the right place.  I set a breakpoint there (DelimiterParsers.scala:71 & another at :73 just in case), ran all the TestDelimiterProperties tests, and the debugger didn't stop in that function even once.

mbeckerle: Hmmm. test_E1 in the TestDelimiterProperties.scala file hits a breakpoint at line 99 of DelimiterParsers.scala for me. That's in the parse method.

tuxji:
Fixups done.  Very useful to know that lazy val is more performative in schema compiler because of nested recursive calls.

tuxji:
Commits are squashed and ready to be merged now.

PR #386, opened 2020-06-16
Implemented dfdl:assert failureType="recoverableError"

jadams-tresys:
Simple addition of allowing recoverable errors to occur from
dfdl:assert. Right now it is simply creating a runtime Schema Definition
Warning with the assert failure message.

DAFFODIL-357

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
@@ -38,6 +38,7 @@ import org.apache.daffodil.processors.unparsers.NewVariableInstanceStartUnparser
 import org.apache.daffodil.compiler.ForParser
 import org.apache.daffodil.schema.annotation.props.PropertyLookupResult
 import org.apache.daffodil.schema.annotation.props.Found
+import org.apache.daffodil.schema.annotation.props.gen.FailureType
 import org.apache.daffodil.dsom.ExpressionCompilers
 import org.apache.daffodil.dsom.DFDLSetVariable
 import org.apache.daffodil.dsom.DFDLNewVariableInstance
@@ -59,7 +60,8 @@ abstract class AssertBase(
   scWherePropertyWasLocated: AnnotatedSchemaComponent,
   msgOpt: Option[String],
   discrim: Boolean, // are we a discriminator or not.
-  assertKindName: String)
+  assertKindName: String,
+  failureType: FailureType)
   extends ExpressionEvaluatorBase(scWherePropertyWasLocated) {
 
   def this(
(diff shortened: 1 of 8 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
```diff
@@ -98,7 +101,7 @@ abstract class AssertBooleanPrimBase(
   decl: AnnotatedSchemaComponent,
   stmt: DFDLAssertionBase,
   discrim: Boolean, // are we a discriminator or not.
-  assertKindName: String) extends AssertBase(decl, Found(stmt.testTxt, stmt, "test", false), stmt.messageAttrib, discrim, assertKindName)
+  assertKindName: String) extends AssertBase(decl, Found(stmt.testTxt, stmt, "test", false), stmt.messageAttrib, discrim, assertKindName, stmt.failureType)
```
mbeckerle, line 104: Please wrap lines before the extends keyword.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/AssertPatternParsers.scala
```diff
@@ -75,8 +77,11 @@ class AssertPatternParser(
       val isMatch = dis.lookingAt(m, start)
       if (!isMatch) {
         val message = getAssertFailureMessage(start)
-        val diag = new AssertionFailed(context.schemaFileLocation, start, message)
-        start.setFailed(diag)
+        if (failureType == FailureType.ProcessingError) {
+          val diag = new AssertionFailed(context.schemaFileLocation, start, message)
+          start.setFailed(diag)
+        } else
+          start.SDW(message)
```
stevedlawrence, line 84: Looking at the spec, it isn't really clear how recoverable error messages are expected to be presented to the user, and really what a recoverable error means. SDW doesn't seem unreasonble, but I could imagine people might skim over it thinking it's a warning? There's also no easy way to an API user to determine if a recoverable error happened, since they look idential to a SDW.

I imagine when someone does an assert with recoverable error, it's to do some sort of check that shouldn't necessarily fail the parse, which feels a lot to me like a validation error. Maybe it makes sense to make this a validation error instead?

Or maybe we should add a new error type that is RecoverableError, similar to ValidationError? Note that we have ``isProcessingError`` and ``isValidationError`` that one can check on a ParseResult. Do we need another ``isRecoverableError`` as well if we add a new error type? And some way to differentiate validation vs processing vs recoverable diagnostics in the API? And does this require adding new ``<tdml:recoverableErrors>`` to the tdml format?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -209,8 +211,11 @@ class AssertExpressionEvaluationParser(
       start.setDiscriminator(discrim)
     } else {
       val message = getAssertFailureMessage(start)
```
stevedlawrence, line 213: Also, in addition to failureType="recoverableError", the spec also states that a recoverable error should also be created if evaluting the message expression for asserts/discriminators fails. It looks like right now if the message expression fails in getAssertFailureMessage, we just generate a new message without also creating a recoverable error. However we deicded recoverable errors should be presented to the user, we should do the same for when evaluating ``messageExpr`` fails.

stevedlawrence: +1, with future fixes to address above comments

PR #387, opened 2020-06-17
DAFFODIL-2349: fixed overflow ArithmeticException when totalDigits >= 10

claudemamo:


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -455,10 +455,9 @@ final class SimpleTypeRuntimeData(
     // Per http://www.w3.org/TR/xmlschema-2/#rf-totalDigits
     // |i| < 10^totalDigits
 
-    val number = new java.math.BigDecimal(scala.math.pow(10.0, digits.doubleValue()))
-    val biNumber = new java.math.BigInteger(number.intValueExact().toString())
+    val number = new java.math.BigDecimal(scala.math.pow(10, digits)).toBigIntegerExact
     val bdData = diNode.dataValueAsBigDecimal.unscaledValue()
-    val isDataLessThanNumber = bdData.compareTo(biNumber) < 0
+    val isDataLessThanNumber = bdData.compareTo(number) < 0
     isDataLessThanNumber
```
stevedlawrence, line 461: This doesn't take into the account that ``bdData`` could be negative. This is an existing bug, but it would be nice to fix as part of this, and I think it just means adding an abs() somewhere.

I'm also thinking that we need to change how we create ``number``. scala.math.pow returns a double, so there could be precision issues.  For example, if totalDigits were 30, this would be
```scala
scala> val number = new java.math.BigDecimal(scala.math.pow(10.0, 30.0))
res0: number: java.math.BigDecimal = 1000000000000000019884624838656
```
Which is too big and would allow some values that are more than 30 digits. 30 is maybe pretty large for totalDigits and isn't too likely, but it's not out of the realm of possibility. And it looks like precision issues happen around 23. I think maybe just using BigInteger.pow should work e,g.
```scala
val number = java.math.BigInteger.TEN.pow(totalDigits)
```
That doesn't seem to have any precision issues.

Also, I don't think this takes into account the scale of the BigDecimal, it only looks at the unscaledValue. For example,
```scala
scala> (new java.math.BigDecimal("1E2")).unscaledValue
res0: java.math.BigInteger = 1
```
The total digits for 1E2 is three, but we are only comparing against the unscaledValue of 1, so this would not be caught if totalDigits were 2. I think we also need to take into account the scale or precision, though it's not immediately obvious to me the right way to do that.

stevedlawrence: I thought some more about this, and I think the above logic overcomplicates things. There's no need to compare against powers of 10. All that is needed is bd.precision and bd.scala to figure out the total digits.

If scale <= 0, that means the number is of the from XXXXXYYYYY where the number of X's is the precision and the number of Y's are the negative scale (and all zeros). For example 1234*10^2 = 123400. In this example precision is 4 and scale is -2. So for this case, the total digits is just ``bd.precision - bd.scale``.

If scale > 0, then the number is of the form (precision digits) * 10 ^ -scale, and we want whatever is larger of precision and scale. For example, if precision was 4 and scale was 2, the number would be of the form XX.XX, and so total digits is just the precision (or 4). On the other hand if precision was 2 and scale was 4, the number would be of the form 0.00XX, so we want scale this time (4 again). So in this case, the total digits is just Math.max(bd.precision, bd.scale).

I think was also need to sort of normalize the big decimal by stripping off trailing zeros. Otherwise something like 10.000 throws off the logic. This is because this is represented as a precision of 5 and scale of 3 (10000*10^-3), but those 3 trailing zeros shouldn't count towards totalDigts.

So I think the logic becomes something like this:

```scala
val bd = diNode.dataValueAsBigDecimal.stripTrailingZeros
val totalDigits =
  if (bd.scale <= 0) bd.precision - bd.scale
  else Math.max(bd.precision, bd.scale)
```

PR #388, opened 2020-06-18
Scalafix ProcedureSyntax

jw3:
Use Scalafix `ProcedureSyntax` rule to fix the deprecated procedure syntax usage.

DAFFODIL-1673

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -112,7 +112,8 @@ lazy val commonSettings = Seq(
     "-language:experimental.macros",
     "-unchecked",
     "-Xfatal-warnings",
-    "-Xxml:-coalescing"
+    "-Xxml:-coalescing",
+    "-Xfuture"
   ),
   // Workaround issue that some options are valid for javac, not javadoc.
   // These javacOptions are for code compilation only. (Issue sbt/sbt#355)
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/blob/TestBlob.scala
@@ -57,7 +57,7 @@ class TestBlob {
    * python gen_blob.py -s 1 -o 1MB.bin
    *
    ***/
-  /*@Test*/ def test_1MB_blob() {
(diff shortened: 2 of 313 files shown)


mbeckerle:
Once we incorporate this fix, are there compiler options that we can put into build.sbt to insure we don't end up adding the obsolete construct back again?

I'd like at least a warning if we add this syntax back.

jw3:
It looks like this syntax is deprecated by default starting in 2.13., however for pre 2.13 we would have to use the `-Xfuture` flag (which was noted in jira).

I just tested with `-Xfuture` locally in both this PR branch and master and get expected behavior
- compile fails in master
- compile succeeds in PR branch
- compile fails in PR branch if procedure syntax is used

Will add that option in another commit on this PR.

Offhand I am not sure what else -Xfuture could impact...

stevedlawrence:
Based on https://github.com/scala/scala/pull/3076/, it looks like the ``-Xfuture`` scalacOption enables a check for this. Not sure what other checks this flag enables, but it would probably be a good goal to enable that flag and fix whatever issues show up.

Based on this change https://github.com/scala/scala/pull/6325, it looks like scala 2.13 (which we don't support yet) changes it to the ``-Xsource:2.14`` flag. Scala 2.13 has been out for about a year, so it might be worth considering upgrading to that as well at some point.

jw3:
Looking into the failures.  It appears Scalafix skipped rewriting tests, causing failed compilation when running tests + Xfuture.

jw3:
That looks better, had to qualify test to have scalafix rewrite them

`sbt test:"scalafix ProcedureSyntax"`

That added a lot of changes, moving the total (5194) more inline with what was noted in jira (3476).

Tests pass locally with Xfuture, still getting CI failures though.  Thoughts?

jw3:
Eh, I see it skipped some integration tests too....

jw3:
It might have been easier for this task, given the complex layout of this project, to just specify the root directory rather than trying to use sbt project mappings

`sbt "scalafix ProcedureSyntax -f /home/path/to/incubator-daffodil"`

PR #393, opened 2020-06-25
Add scalac warnings for unused imports

jw3:
- cleans up unused imports using scalafix
- enables scalac warn-on-unused option
- adds support for version specific scalac options
- enables language existentials in 2.11
- does not organize imports, only removes unused

DAFFODIL-2145


jw3:
Ah, yeah will need separate options for 2.11...

daffodil-runtime1/src/main/scala/org/apache/daffodil/udf/UserDefinedFunctionService.scala
```diff
@@ -28,7 +28,6 @@ import java.lang.reflect.Method
 import org.apache.daffodil.dpath.NodeInfo
 import java.io.Serializable
 import java.io.ObjectInputStream
-import scala.language.existentials
```
jw3, line 31: Had to add a language feature flag for 2.11 due to the import being seen as unused in 2.12
  - See https://github.com/scala/bug/issues/11467

tuxji:
Looks good to me.

I followed the [scala/bug#11467](https://github.com/scala/bug/issues/11467) link, wondered who sjrd was, googled, found out he authored Scala.js, browsed his Twitter page, and learned about some interesting stuff including [scalafix-organize-imports](https://github.com/liancheng/scalafix-organize-imports).  Seems even more powerful than scalafix's RemoveUnused rule since it organizes all imports nicely into groups, although only one or the other rule should be used.

jw3:
It needs squashed some.  Do you want two commits with the build.sbt changes left separate, or squash it all into one?

stevedlawrence:
I think one commit is fine in this case. It's maybe a little difficult for the important build.sbt change to stand out, but it is usually going to be the first file listed in a diff so it's not too hidden.

tuxji:
+1  Steve encouraged me to try out my committer bit :-).

Once your squashed commit passes the checks, I'll see if I can merge your commit too.

PR #394, opened 2020-07-08
Feature Addition: API for traversing DSOM and generating NIFI RecordSchemas from DSOM

andrewjc2000:
In this fork, no code was modified other than 2 added dependencies, and several classes to define an API and test them were added.  These were all in the org.apache.daffodil.dsom package.  Here is a summary of the additions:
- An abstract class called AbstractDSOMWalker that traverses DSOM objects. As the DSOM is traversed, various event handlers that are declared abstract in the API are invoked, depending on the DSOM element encountered (Sequence, Element, Choice, etc.).
- A concrete implementation of AbstractDSOMWalker called RecordWalker that generates a NIFI RecordSchema as a result of the DSOM traversal.
- A helper class called SchemaNode for storing the relevant information about an element in the DSOM.  These are used in the intermediate stages of processing in RecordWalker.
- 3 Test classes and 3 complete DFDL Schema files for testing the output of RecordWalker.  These use the Scalatest Unit Test framework.

As a result of the above features, 2 new dependencies were also added for Core in project/Dependencies: one for NIFI Records, and another for ScalaTest

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -27,13 +27,13 @@ import java.util.zip.ZipException
 
 import scala.collection.immutable.Queue
 import scala.xml.Node
-
 import org.apache.daffodil.api.DFDL
 import org.apache.daffodil.api.DaffodilSchemaSource
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.api.URISchemaSource
 import org.apache.daffodil.api.UnitTestSchemaSource
 import org.apache.daffodil.dsom.SchemaSet
+import org.apache.daffodil.dsom.walker.RootView
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.externalvars.Binding
 import org.apache.daffodil.externalvars.ExternalVariablesLoader
@@ -98,6 +98,8 @@ final class ProcessorFactory private(
     optSchemaSet.getOrElse(
       new SchemaSet(optRootSpec, schemaSource, validateDFDLSchemas, checkAllTopLevel, tunables,
         compilerExternalVarSettings))
+
+  lazy val rootView: RootView = sset.root
 
   def elementBaseInstanceCount = sset.elementBaseInstanceCount
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
@@ -23,6 +23,8 @@ import scala.xml._
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.grammar.ModelGroupGrammarMixin
 import java.lang.{ Integer => JInt }
+
+import org.apache.daffodil.dsom.walker.ModelGroupView
 import org.apache.daffodil.schema.annotation.props.AlignmentType
 import org.apache.daffodil.schema.annotation.props.gen.AlignmentUnits
 import org.apache.daffodil.schema.annotation.props.gen.YesNo
@@ -135,7 +137,8 @@ abstract class ModelGroup(index: Int)
   extends Term
   with ModelGroupGrammarMixin
   with OverlapCheckMixin
-  with NestingLexicalMixin {
+  with NestingLexicalMixin
+  with ModelGroupView {
 
   requiredEvaluationsIfActivated(groupMembers)
   requiredEvaluationsIfActivated(initiatedContentCheck)
@@ -177,7 +180,7 @@ abstract class ModelGroup(index: Int)
    *
    * This also depends on groupMembersDef overrides all being lazy val.
    */
-  final lazy val groupMembers: Seq[Term] =
+  override final lazy val groupMembers: Seq[Term] =
     schemaSet.sharedGroupMembersFactory.getShared(shareKey, groupMembersDef)
 
   /**
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
@@ -17,7 +17,7 @@
 
 package org.apache.daffodil.dsom
 
-import java.math.{BigInteger => JBigInt}
+import java.math.{ BigInteger => JBigInt }
 
 import scala.xml.Node
 import org.apache.daffodil.cookers.IntRangeCooker
@@ -25,6 +25,7 @@ import org.apache.daffodil.cookers.RepValueCooker
 import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dpath.NodeInfo.PrimType
 import org.apache.daffodil.dpath.InvalidPrimitiveDataException
+import org.apache.daffodil.dsom.walker.SimpleTypeView
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.IdentifyTypeCalculator
 import org.apache.daffodil.processors.RepValueSet
@@ -53,9 +54,9 @@ trait TypeBase {
 trait NonPrimTypeMixin
 
 sealed trait SimpleTypeBase extends TypeBase
-  with HasOptRepTypeMixin {
+  with HasOptRepTypeMixin with SimpleTypeView {
 
-  def primType: PrimType
+  override def primType: PrimType
 }
 
 /*
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
@@ -18,16 +18,19 @@
 package org.apache.daffodil.dsom
 
 import java.util.UUID
+
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.grammar.TermGrammarMixin
 import org.apache.daffodil.schema.annotation.props.gen.YesNo
 import java.lang.{ Integer => JInt }
+
 import org.apache.daffodil.schema.annotation.props.Found
 import org.apache.daffodil.schema.annotation.props.NotFound
 import org.apache.daffodil.schema.annotation.props.gen.LengthKind
 import org.apache.daffodil.schema.annotation.props.gen.OccursCountKind
 import org.apache.daffodil.schema.annotation.props.SeparatorSuppressionPolicy
 import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.dsom.walker.TermView
 
 /**
  * Mixin for objects that are shared, but have consistency checks to be run
@@ -86,7 +89,8 @@ trait Term
   with DelimitedRuntimeValuedPropertiesMixin
   with InitiatedTerminatedMixin
   with TermEncodingMixin
-  with EscapeSchemeRefMixin {
+  with EscapeSchemeRefMixin
+  with TermView {
 
   requiredEvaluationsIfActivated(annotationObjs)
   requiredEvaluationsIfActivated(nonDefaultPropertySources)
@@ -178,7 +182,7 @@ trait Term
    *
    * The DFDL spec is not entirely consistent here either I don't believe.
    */
-  def isOptional: Boolean
+  override def isOptional: Boolean
 
   /**
    * An array can have more than 1 occurrence.
@@ -186,7 +190,7 @@ trait Term
    * An optional element (minOccurs=0, maxOccurs=1) is an array only
    * if occursCountKind is parsed, because then the max/min are ignored.
    */
-  def isArray: Boolean
+  override def isArray: Boolean
 
   def elementChildren: Seq[ElementBase]
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
@@ -1,0 +1,174 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.dsom.walker
+
+/**
+ * Below is a series of "View" Mixins defined such that particular simple attributes
+ * of the core DSOM classes are exposed as a Walk is performed.
+ *
+ * Several of them have a walkDSOM method.  By default, this method is called as the walk is performed,
+ * but a user implementing AbstractDSOMWalker can override this behavior by overriding walkFromRoot
+ * and implementing their own event handlers for each of the DSOM types.
+ */
+
+trait CommonContextView {
+  def namespaces: scala.xml.NamespaceBinding
+}
+
+trait TermView extends CommonContextView {
+  def isArray: Boolean
+  def isOptional: Boolean
+}
+
+trait TypeView
+trait SimpleTypeView extends TypeView {
+  def primType: PrimTypeView
+}
+trait ComplexTypeView extends TypeView {
+  def group: ModelGroupView
+}
+
+trait ModelGroupView extends TermView {
+  def groupMembers: Seq[TermView]
+}
+
+trait ChoiceView extends ModelGroupView
+trait SequenceView extends ModelGroupView
+trait GroupRefView extends ModelGroupView {
+  def isHidden: Boolean
+}
+
+trait ElementDeclView extends CommonContextView {
+  def isSimpleType: Boolean
+  def isComplexType: Boolean
+  def simpleType: SimpleTypeView
+  def complexType: ComplexTypeView
+}
+
+trait ElementBaseView extends ElementDeclView with TermView {
+  def name: String
+}
+
+trait RootView extends ElementBaseView
+
+/**
+ * A class designed to walk the internal representation of a DFDL Schema File.
+ *
+ * There are 2 main event handlers an implementing class has to worry about: one for Terms, and
+ * another for Types.  These are called as the DSOM is walked.
+ *
+ * Though recursion is used here to define the walk, it is not advised to use recursion between
+ * these event handlers.  Instead, consider a stack-like structure, as the DFDL Schema structure
+ * as well as the recursive method call structure can be represented by trees.
+ */
+abstract class AbstractDSOMWalker {
+
+  /**
+   * Method to be called on the beginning of the traversal.  It is recommended to add some
+   * sort of wrapper element to a stack if you're doing a typical stack-based traversal.
+   *
+   * @param root the root element of the DFDL Schema
+   */
+  protected def onWalkBegin(root: RootView): Unit
+
+  /**
+   * Method to be called when the traversal concludes.  It is recommended to put any post-processing
+   * and anything to tidy up the stack or the result here.
+   *
+   * @param root the root element of the DFDL Schema
+   */
+  protected def onWalkEnd(root: RootView): Unit
+
+  /**
+   * Method to be called whenever any element that is a Term is encountered.
+   * This applies to Sequence, Choice, GroupRef, ElementBase, etc.
+   *
+   * It is highly recommended that, when implementing this method, you pattern match
+   * some of these different sub-types at some point to handle each accordingly.
+   *
+   * @param termElement the term element
+   */
+  def onTermBegin(termElement: TermView): Unit
+
+  /**
+   * Method to be called when a Term element has finished processing
+   *
+   * @param termElement the term element
+   */
+  def onTermEnd(termElement: TermView): Unit
+
+  /**
+   * Method to be called when either a Simple or Complex Type element has been encountered.
+   * This is just the element for <simpleType> or <complexType>, so no implementation is necessary if you
+   * do not care about these wrapper portion; its children will be walked automatically.
+   *
+   * @param typeElement either a Simple or Complex type
+   */
+  def onTypeBegin(typeElement: TypeView): Unit
+
+  /**
+   * Method to be called when either a Simple or Complex Type element has been finished
+   * processing.  See onTypeBegin method description
+   *
+   * @param typeElement either a Simple or Complex type
+   */
+  def onTypeEnd(typeElement: TypeView): Unit
+
+  /**
+   * Starts a DSOM walk from the Root object.  By default, the walk is completed recursively in the helper
+   * method below, but a custom walk can be defined by overriding
+   * this method instead and calling local event handlers.
+   * @param schemaSetRoot The root element of the DFDL Schema.  This will be the starting point of the traversal
+   */
+  def walkFromRoot(schemaSetRoot: RootView): Unit = {
+    onWalkBegin(schemaSetRoot)
+    walkerHelper(schemaSetRoot)
+    onWalkEnd(schemaSetRoot)
+  }
+
+  /**
+   * Local helper method used to recursively walk the DSOM.
+   * If a non-recursive walk or a more specific recursive walk is desired, one
+   * can override this method, or simply not use it and instead override walkFromRoot
+   * and have that overridden method reference custom helper methods.
+   * @param termView the current Term element to be pattern matched
+   */
+  protected def walkerHelper(termView: TermView): Unit = {
+    onTermBegin(termView)
+    termView match {
+      case element: ElementBaseView =>
+        if (element.isComplexType) {
+          onTypeBegin(element.complexType)
+          walkerHelper(element.complexType.group)
+          onTypeEnd(element.complexType)
+        } else {
+          onTypeBegin(element.simpleType)
+          onTypeEnd(element.simpleType)
+        }
+      case groupRef: GroupRefView =>
+        if (!groupRef.isHidden) {
+          groupRef.groupMembers.foreach(walkerHelper)
+        }
+      case modelGroup: ModelGroupView =>
+        modelGroup.groupMembers.foreach(walkerHelper)
+    }
+    onTermEnd(termView)
+  }
+
+
+}
--- a/daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.dsom.walker
+
+import scala.collection.mutable.ArrayBuffer
+
+class BasicWalker(ignoreTypeWrappers: Boolean = false, onlyElements: Boolean = false, ignoreEndEvents: Boolean = true) extends AbstractDSOMWalker {
+
+  var nodeArr: ArrayBuffer[AnyRef] = ArrayBuffer()
+
+  private def addViewElement(viewElement: AnyRef): Unit = {
+    viewElement match {
+      case _: ElementBaseView => nodeArr += viewElement
+      case _: TermView => if (!onlyElements) nodeArr += viewElement
+      case _ => if (!ignoreTypeWrappers) nodeArr += viewElement
+    }
+  }
+
+  override protected def onWalkBegin(root: RootView): Unit = addViewElement(root)
+
+  override protected def onWalkEnd(root: RootView): Unit = if (!ignoreEndEvents) addViewElement(root)
+
+  override def onTermBegin(termElement: TermView): Unit = addViewElement(termElement)
+
+  override def onTermEnd(termElement: TermView): Unit = if (!ignoreEndEvents) addViewElement(termElement)
+
+  override def onTypeBegin(typeElement: TypeView): Unit = addViewElement(typeElement)
+
+  override def onTypeEnd(typeElement: TypeView): Unit = if (!ignoreEndEvents) addViewElement(typeElement)
+}
--- a/daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestDSOMWalker.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestDSOMWalker.scala
@@ -1,0 +1,177 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.dsom.walker
+
+import org.apache.daffodil.util._
+import org.apache.daffodil.compiler.{ Compiler, ProcessorFactory }
+import org.junit.Test
+import org.junit.Assert._
+
+class TestDSOMWalker {
+
+  @Test def testComplexTypesAndEndEvents(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+      <dfdl:format ref="ex:GeneralFormat"
+                   alignment="implicit" alignmentUnits="bits" occursCountKind="implicit"
+                   lengthKind="delimited" encoding="ASCII"/>,
+      <xs:element name="PersonData">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:choice>
+              <xs:element name="age" type="xs:int" minOccurs="1" maxOccurs="1"/>
+            </xs:choice>
+            <xs:group ref="testGroup" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
+      <xs:group name="testGroup">
+        <xs:sequence />
+      </xs:group>
+    )
+    val pf: ProcessorFactory = Compiler().compileNode(testSchema)
+    assertEquals(s"This basic Schema: $testSchema should compile; here are some diagnostics: ${pf.getDiagnostics}", false, pf.isError)
+    val walker: BasicWalker = new BasicWalker(ignoreEndEvents = false)
+    walker.walkFromRoot(pf.rootView)
+    val nodeStack: List[AnyRef] = walker.nodeArr.toList
+    assertEquals(s"Node Stack $nodeStack should have 16 elements", 16, nodeStack.size)
+    assertTrue("Should have received a start event for the overall traversal", nodeStack.head.isInstanceOf[RootView])
+    assertTrue("The Root element was not of type RootView", nodeStack(1).isInstanceOf[RootView])
+    assertEquals("The root element should be named 'PersonData'", "PersonData", nodeStack(1).asInstanceOf[RootView].name)
+    assertTrue("The root element should contain a complexType wrapper child", nodeStack(2).isInstanceOf[ComplexTypeView])
+    assertTrue("The complexType element should contain a Sequence child", nodeStack(3).isInstanceOf[SequenceView])
+    assertTrue("The Sequence element should contain a Choice child", nodeStack(4).isInstanceOf[ChoiceView])
+    assertTrue("The Choice element should contain an Element child", nodeStack(5).isInstanceOf[ElementBaseView])
+    assertEquals("The Element child should be named 'age'", "age", nodeStack(5).asInstanceOf[ElementBaseView].name)
+    assertEquals("The Element child should be a simple type", true, nodeStack(5).asInstanceOf[ElementBaseView].isSimpleType)
+    assertTrue("The 'age' element should have a SimpleTypeView", nodeStack(6).isInstanceOf[SimpleTypeView])
+    assertTrue("Should have received an end event for the Simple Type", nodeStack(7).isInstanceOf[SimpleTypeView])
+    assertTrue("Should have received an end event for the 'age' element", nodeStack(8).isInstanceOf[ElementBaseView])
+    assertTrue("Should have received an end event for the Choice element", nodeStack(9).isInstanceOf[ChoiceView])
+    assertTrue("The Sequence element should contain a second child that is a GroupRef", nodeStack(10).isInstanceOf[GroupRefView])
+    assertTrue("Should have received an end event for the GroupRef element", nodeStack(11).isInstanceOf[GroupRefView])
+    assertTrue("Should have received an end event for the Sequence element", nodeStack(12).isInstanceOf[SequenceView])
+    assertTrue("Should have received an end event for the complexType wrapper element", nodeStack(13).isInstanceOf[ComplexTypeView])
+    assertTrue("Should have received an end event for the Root element", nodeStack(14).isInstanceOf[RootView])
+    assertTrue("Should have received an end event for the overall traversal", nodeStack(15).isInstanceOf[RootView])
+  }
+
+  private def getSuffix(ordinal: Int): String = {
+    if ((ordinal % 100) / 10 == 1) "th"
+    else if (ordinal % 10 == 1) "st"
+    else if (ordinal % 10 == 2) "nd"
+    else if (ordinal % 10 == 3) "rd"
+    else "th"
+  }
+
+  @Test def testAllSimpleTypes(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+      <dfdl:format ref="ex:GeneralFormat"
+                   alignment="implicit" alignmentUnits="bits" occursCountKind="implicit"
+                   lengthKind="implicit" encoding="ASCII"/>,
+      <xs:element name="AllSimples">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="stringField" type="xs:string" minOccurs="1" maxOccurs="1" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="booleanField" type="xs:boolean" dfdl:lengthKind="explicit" dfdl:length="1" dfdl:textBooleanTrueRep="1" dfdl:textBooleanFalseRep="0" />
+            <xs:element name="byteField" type="xs:byte" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="shortField" type="xs:short" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="intField" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="longField" type="xs:long" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="unsignedByteField" type="xs:unsignedByte" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="unsignedShortField" type="xs:unsignedShort" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="unsignedIntField" type="xs:unsignedInt" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="unsignedLongField" type="xs:unsignedLong" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="nonNegativeIntegerField" type="xs:nonNegativeInteger" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="integerField" type="xs:integer" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="floatField" type="xs:float" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="doubleField" type="xs:double" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="decimalField" type="xs:decimal" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="hexBinaryField" type="xs:hexBinary" dfdl:lengthKind="explicit" dfdl:length="1"/>
+            <xs:element name="anyURIField" type="xs:anyURI" dfdl:lengthKind="explicit" dfdl:length="1" dfdlx:objectKind="bytes"/>
+            <xs:element name="dateTimeField" type="xs:dateTime" dfdl:calendarPattern="yyyy.MM.dd G 'at' HH:mm:ss ZZZZ"
+                        dfdl:calendarPatternKind="explicit" dfdl:lengthKind="explicit" dfdl:length="35"/>
+            <xs:element name="dateField" type="xs:date" dfdl:calendarPattern="EEEE, MMM d, ''yy" dfdl:calendarPatternKind="explicit"
+                        dfdl:lengthKind="explicit" dfdl:length="23"/>
+            <xs:element name="timeField" type="xs:time" dfdl:calendarPattern="h:mm a" dfdl:calendarPatternKind="explicit"
+                        dfdl:lengthKind="explicit" dfdl:length="8"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
+    )
+    val pf: ProcessorFactory = Compiler().compileNode(testSchema)
+    assertEquals(s"This basic Schema $testSchema should compile; here are some diagnostics: ${pf.getDiagnostics}", false, pf.isError)
+    val walker: BasicWalker = new BasicWalker(false, true)
+    walker.walkFromRoot(pf.rootView)
+    val simpleTypes: List[Class[_ <: PrimTypeView]] = List(
+      classOf[StringView], classOf[BooleanView], classOf[ByteView], classOf[ShortView], classOf[IntView],
+      classOf[LongView], classOf[UnsignedByteView], classOf[UnsignedShortView], classOf[UnsignedIntView], classOf[UnsignedLongView],
+      classOf[NonNegativeIntegerView], classOf[IntegerView], classOf[FloatView], classOf[DoubleView], classOf[DecimalView],
+      classOf[HexBinaryView], classOf[AnyURIView], classOf[DateTimeView], classOf[DateView], classOf[TimeView]
+    )
+    val nodeStack: List[AnyRef] = walker.nodeArr.toList
+    assertEquals(s"Node Stack $nodeStack did not have the expected number of elements", 2 * simpleTypes.size + 3, nodeStack.size)
+    for (index <- 1 to simpleTypes.size) {
+      val elementIndex = 2 * index + 1
+      val simpleTypeIndex = 2 * index + 2
+      assertTrue(s"The $elementIndex${getSuffix(elementIndex)} element in the stack should be an Element", nodeStack(elementIndex).isInstanceOf[ElementBaseView])
+      val className: String = simpleTypes(index - 1).getSimpleName
+      val withoutView: String = className.substring(0, className.length - 4)
+      val fieldName: String = withoutView.charAt(0).toLower + withoutView.substring(1) + "Field"
+      assertEquals(
+        s"The $elementIndex${getSuffix(elementIndex)} element in the stack should be named '$fieldName'", fieldName,
+        nodeStack(elementIndex).asInstanceOf[ElementBaseView].name
+      )
+      assertTrue(
+        s"The $simpleTypeIndex${getSuffix(simpleTypeIndex)} element in the stack should be a Simple type",
+        nodeStack(simpleTypeIndex).isInstanceOf[SimpleTypeView]
+      )
+      assertTrue(
+        s"The $simpleTypeIndex${getSuffix(simpleTypeIndex)} element in the stack should be of type '$className'",
+        simpleTypes(index - 1).isInstance(nodeStack(simpleTypeIndex).asInstanceOf[SimpleTypeView].primType)
+      )
+    }
+  }
+
+  @Test def testOptionalField(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+      <dfdl:format ref="ex:GeneralFormat"
+                   alignment="implicit" alignmentUnits="bits" occursCountKind="implicit"
+                   lengthKind="delimited" encoding="ASCII"/>,
+      <xs:element name="DataList">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="data" type="xs:string" minOccurs="0" maxOccurs="1"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
+    )
+    val pf: ProcessorFactory = Compiler().compileNode(testSchema)
+    assertEquals(s"This basic Schema: $testSchema should compile; here are some diagnostics: ${pf.getDiagnostics}", false, pf.isError)
+    val walker: BasicWalker = new BasicWalker(true, true)
+    walker.walkFromRoot(pf.rootView)
+    val nodeStack: List[AnyRef] = walker.nodeArr.toList
+    assertEquals(s"Node Stack $nodeStack did not have the expected number of elements", 3, nodeStack.size)
+    assertTrue("The 3rd element in the stack should be an Element", nodeStack(2).isInstanceOf[ElementBaseView])
+    assertEquals("The 3rd element in the stack should be named 'data'", "data", nodeStack(2).asInstanceOf[ElementBaseView].name)
+    assertEquals("The 'data' element should be optional", true, nodeStack(2).asInstanceOf[ElementBaseView].isOptional)
+  }
+
+}
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
@@ -48,6 +48,7 @@ import org.apache.daffodil.util.{ NullLogWriter => SNullLogWriter }
 import org.apache.daffodil.externalvars.Binding
 import org.apache.daffodil.externalvars.ExternalVariablesLoader
 import org.apache.daffodil.dsom.ExpressionCompilers
+import org.apache.daffodil.dsom.walker.RootView
 import org.apache.daffodil.compiler.{ InvalidParserException => SInvalidParserException }
 import org.apache.daffodil.processors.{ InvalidUsageException => SInvalidUsageException }
 import java.net.URI
@@ -372,6 +373,13 @@ class ProcessorFactory private[japi] (private var pf: SProcessorFactory)
     val dp = pf.onPath(path).asInstanceOf[SDataProcessor]
     new DataProcessor(dp)
   }
+
+  class ExperimentalWrapper(val rootView: RootView)
+  /**
+   * Exposes the RootView object corresponding to this ProcessorFactory.  This can
+   * be used to start a walk using the walkFromRoot method in a DSOM Walker.
+   */
+  lazy val experimental: ExperimentalWrapper = new ExperimentalWrapper(pf.rootView)
 
 }
 
--- a/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
+++ b/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
@@ -46,6 +46,7 @@ import org.apache.daffodil.util.{ LoggingDefaults => SLoggingDefaults }
 import org.apache.daffodil.util.{ NullLogWriter => SNullLogWriter }
 import org.apache.daffodil.externalvars.{ Binding, ExternalVariablesLoader }
 import org.apache.daffodil.dsom.ExpressionCompilers
+import org.apache.daffodil.dsom.walker.RootView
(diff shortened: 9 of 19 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,217 @@
...
+      onComplexBegin(element.complexType)
+      walkerHelper(element.complexType.group)
+      onComplexEnd(element.complexType)
+    } else {
+      onSimpleBegin(element.simpleType)
+      onSimpleEnd(element.simpleType)
+    }
+    onElementEnd(element)
+  }
+}
```
stevedlawrence, line 207: I'm not sure much is gained by spliting the walkerHelper's into different overloaded methods. Just sort of adds another layer of indirection. I think I'd rather see it as one function that accepts a Term and is something like:
```scala
term match {
 case s: Sequence => ...
 case c: Choice => ...
 case g: GroupRef => ...
 case: e: ElementBase => ...
}
```
Another option, taking my first comment a  little further, perhaps instead of the walker having the logic for walking, it all moves into the DSOM. We require that each SchemaComponent implement a ``def walkDSOM(w: Walker)`` method. So for example:
```scala
class Seqeunce {
  ...
  override def walkDSOM(w: Walker): Unit = {
    w.onSequenceBegin(...)
    groupMembers.forEach(_.walkDSOM(w))
    w.onSequenceEnd(...)
  }
```
It's sortof two sides of the same coin, but this way the DSOM elements are more in reponsible for knowing how to walk their own DSOM rather than the walker having to know how to walk everything DSOM. And the AbstractWalker essentialy just becomes a WalkerHandler of events, and doesn't do any actual walking itself.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * The sole purpose of this Companion Object is to contain this one static variable.
+ */
+object AbstractDSOMWalker {
+  /**
+   * Used to suppress all of the extra printing from running test cases
+   */
+  final val PRODUCTION_MODE: Boolean = true
+}
```
stevedlawrence, line 217: This is only used in the NiFI specific Record stuff. I would expect this to just be a flag in the RecordWalker. Presumable it would plug into NiFi's logging mechanism.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,217 @@
...
+  protected def onWalkEnd(root: Root): T
+
+  /**
+   * Method to be called when a *sequence* element is encountered.
+   * This is not really your array type; the typical "array" you'll see in an infoset
+   * is a special kind of Element, so it will be encountered
+   * in onElementBegin
+   * @param sequenceElement the sequence element
+   */
+  protected def onSequenceBegin(sequenceElement: Sequence): Unit
```
stevedlawrence, line 74: I'm a little hesitant to give a walker implementation direct access to all of our internal schema compontents. It makes it much more difficult to make changes to these SchemaComponents since we need to be concerned about backwards compatibilty. We sort of have that problem with our InfosetOutputters, I'd like to not repeat that mistake.

I'm wondering if it would be possible to come up with a reasonable list of properties that DSOM walkers actually need and only make those avaialble? I suspect this would be a pretty small list llike namespace, name, type, min/maxOccurs, and maybe a couple more properties?

Perhaps your SchemaNode is the correct abstraction? Maybe fore each walker event, the we should be creating a SchemaNode and passing that in? Then we only have to worry about backwards compatibility with SchemaNodes, which should be significantly easier than our SchemaComponents?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/RecordWalker.scala
```diff
@@ -0,0 +1,309 @@
...
+ * This class produces a NIFI RecordSchema that is intended to match the original DFDL file.
+ *
+ * The RecordSchema is built in 3 primary stages:
+ * 1) A tree of SchemaNodes is created as the DFDL file is walked; this walk is performed
+ * through the various event handlers defined in the parent abstract class.
+ * 2) The tree of SchemaNodes undergoes some post-processing, mainly to remove redundant Record wrappers.
+ * 3) The tree of SchemaNodes is converted into a RecordSchema; it is walked recursively within this class.
+ * @param schemaFile an input DFDL Schema file to parse
+ */
+class RecordWalker(schemaFile: File) extends AbstractDSOMWalker[RecordSchema](schemaFile) {
```
stevedlawrence, line 61: This is pretty great! I'm surprised at how little code is needed to hook this all together.

However, one thing we need to discuss is where this code actually lives. I think it's safe to safe it shouldn't live in the daffodil-core module. If it does, that means anyone depending on daffodil-core will now pull in a NiFi dependency, which shouldn't be necessary. The DSOM related stuff makes sense to be in core, because it needs to know how to walk the DSOM. But the NiFi specifc code needs to be in a separate module for dependency purposes.

The question, is where does that separate module live? One option would be to create an "integrations" directory, and that could contain separate directories for different kinds of integrations. The jars created from those modules can have separate dependencies so you only need to pull in those dependencies if you're using the integration jars. Something like this probably makes the most sense.

Another thing that plays a role where this lives is this RecordWalker is just one part of NiFi integration. There also needs to be a NiFi Controller/Processor that use this RecordWalker. Ideally, that too would live in the same module so that the result of this module could be a NiFi nar. One issue with that is generally maven is used to build special NiFi nars. Integrating that into the sbt build process might pose some challenges, especially since we have a couple sbt plugins. But maybe we don't want that integration so that building Daffodil doesn't require pulling in dependencies from a bunch of different frameworks? We could of course include those in our github actions, but perhaps they aren't build by default?

Another option is to put integrations in separate repos. That makes integration tests a bit more complicated. And I'm not sure how much benefit there is. The clear separate is maybe nice, but perhaps that's too much separation.

Note that there also exists Daffodil NiFi processor: https://github.com/TresysTechnology/nifi-daffodil

We probably want to determine how that might integrate in this as well. Perhaps it can all go in a single nar?

Another option, maybe we should attempt to upstream this into NiFi and the only thing that exists in this repo are the abstract walker and the API changes (plus regression tests not specific to NiFi). There is already an open pull request to add the Nif Processor here: https://github.com/apache/nifi/pull/3130. The NiFi devs didn't seem adverse to Daffodil support, but wondering if Records is the better approch. So they may be open to accepting this kind of change.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,217 @@
...
+ * @param schemaFile an input DFDL Schema file to parse
+ * @tparam T the return type of walkDSOMSchema and onWalkEnd.  This should be a class used to
+ *           represent the structure of the Schema.  You could always make it Unit if you don't
+ *           want these methods to return anything.
+ */
+abstract class AbstractDSOMWalker[T](schemaFile: File) {
+  /**
+   * The root element of the DFDL Schema.  This will be the starting point for the traversal
+   */
+  private final val schemaSetRoot: Root = Compiler().compileFile(schemaFile).sset.root
```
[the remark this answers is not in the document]
tuxji: I like Steve's suggestion to separate compilation and walking, and provide a DSOM walk method which takes a Listener class (if I understand correctly).  I have worked on someone else's Visitor class for an ANTLR v4 grammar and then written my own Listener class for the same ANTLR v4 grammar.  In my experience, the Listener class for the same use case was smaller and more maintainable since ANTLR took care of visiting every part of the parse tree (smaller chance for mistakes) and I was able to implement only a handful of listener methods that I actually needed.  ANTLR listeners can't return values directly from method calls, but I was able to attach these return values to specific parts of the parse tree (by modifying the ANTLR grammar to provide fields for these values) and have higher up listener methods collect these values from lower down parts of the parse tree.  Collecting return values may become harder if you can't store them in the tree as you walk the tree like I did, though.

In case people aren't familiar with ANTLR visitors and walkers, someone has blogged about them (<https://saumitra.me/blog/antlr4-visitor-vs-listener-pattern/>) and said:

> Good thing about listener pattern is that its all automatic and even if you don’t write any parse tree walker, ANTLR will figure out and trigger the enter and exit method for each rule. This is a huge benefit for translation type of usecases.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: The abstract walker can define a set of view mixin traits that we mix into each walked object. The traits may be specific to each walked DSOM object as it has to expose different things from them perhaps.

The walker API would be defined in terms of these traits. The traits would have public members that are the things we choose to formally expose (and support) on the walker API.

mbeckerle: I'd suggest the names should end with "View" so ComplexTypeView would be the trait mixed into ComplexTypeBase giving the walker access to its public members, which would just be pass through. 

To cheat users could always downcast to ComplexTypeBase, but I think that's glaring enough that we can live with it.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: Sometimes nothing but a recursive walk algorithm will do. A visitor struggles with state management when it has to refer to more than one tree node at a time. 

Suggest we stick for now with what makes this particular NiFI Record-Schema synthesizer work out well. If that's a visit pattern then great, but if an actual walk proves more natural we should have that.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: There needs to be a way to "register" a walker. I.e., a Daffodil API call where the walker is passed to Daffodil, so that it can be invoked. Or if the control pattern is that the walker takes control, it's really just a hook where daffodil informs the walker that the DSOM tree is ready to be walked. This should allow for many external walkers to be registered, not just one. They should be called in turn. 

Then all the NiFi-specific logic, tests, etc. all goes into distinct separate code that is not part of Daffodil. It becomes an application of Daffodil's libraries within NiFi code.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: I agree, that's the more appropriate location.  I might get rid of this flag eventually, the only reason it exists is if the Unit tests fail you can turn it on and see what the data structures look like at each stage of the processing.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * Overloaded walkerHelper method used for when some "group" element is encountered in
+   * the DSOM; the group elements are Sequences, Choices, and GroupRefs
+   * @param modelGroup the ModelGroup element that was encountered in the Schema
+   */
+  private final def walkerHelper(modelGroup: ModelGroup): Unit = {
+    modelGroup match {
+      case sequence: Sequence =>
+        onSequenceBegin(sequence)
+        for (child <- sequence.groupMembers) walkerHelper(child)
```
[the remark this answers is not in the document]
andrewjc2000: That makes sense!  I use that approach in the latest commit.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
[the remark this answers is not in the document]
andrewjc2000: This all makes a lot more sense idiomatically.  In the latest commit I opted for adding this method in ProcessorFactory:
`def walkDSOM[T](walker: AbstractDSOMWalker[T]): T = walker.walkFromRoot(sset.root)`
AbstractDSOMWalker now also does not take in any constructor parameters.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: While I did leave the fundamental recursive walk structure intact, as Mike stated might have to do anyway in the end, I did consolidate quite a few methods.  Now the user need only implement events for starting and ending the walk, and 2 handlers each for Terms and Types, where a Type could either be Simple or Complex.  See the latest commit.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: This all makes a lot of sense, hopefully I took care of it well in the latest version.

I wound up defining a series of traits ending in View as Mike suggested, such as TermView and SequenceView.  Everything in RecordWalker and the helper SchemaNode (which I renamed to RecordSchemaNode to avoid confusion, it should _only_ be used for the NIFI RecordSchema, as it has an attribute that is evaluated to a NIFI Record Simple type, now uses all of these traits.

The only "internal" item the traits can access that isn't a primitive or a String is the Enum PrimType class; I decided to keep it since it was highly useful when it came to simple types.  I hope leaving this in won't cause any issues.

Within these traits I also defined a new "walkDSOM" method; hence moving all of those walkerHelper methods out of the abstract class and into the new traits.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/RecordWalker.scala
andrewjc2000: I agree, I definitely want to separate all of the NIFI-related code from anything related to DSOM.  I am working on NIFI Processors that use this entire DSOM API but they (more so the Unparse processor) are not quite finished yet, I am still doing a great deal of refactoring.  

I am not sure that any of the NIFI code either than RecordWalker and RecordSchemaNode need exist in the main Daffodil codebase - if it were to go here, would it still go in core but just a different subdirectory, or would the japi/sapi modules be more appropriate?  

Since the base of my code largely builds off of the repo you linked, let me know if you think it is appropriate if I simply rebase my project onto that one or if we should go with a different set of processors since they are a good bit different when they utilize NIFI Records internally.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,163 @@
...
+ * Though recursion is used here to define the walk, it is not advised to use recursion between
+ * these event handlers.  Instead, consider a stack-like structure, as the DFDL Schema structure
+ * as well as the recursive method call structure can be represented by trees.
+ * @tparam T the return type of walkDSOMSchema and onWalkEnd.  This should be a class used to
+ *           represent the structure of the Schema.  You could always make it Unit if you don't
+ *           want these methods to return anything.
+ */
+abstract class AbstractDSOMWalker[T] {
+
+  /**
```
stevedlawrence, line 100: Regarding the *View stuff, this seems reasonable to me. One concern is that we probably want this to be a public stable API, that works in both Java and Scala. I *think* Scala traits get turned into interfaces that a Java class can access, but I'm not 100% sure. If we generate JavaDocs, which we probably want to do, I'm also not sure if the Scala traits turn into something useful that makes sense for a java API. Note that we use the genjavadocs plugin to generate javadocs for daffodil-japi. I'm wondering if we want to do something similar? So maybe we want to move all the walker related things that are part of the public API (i.e. everything in the dsom.walker package) to a daffodil-dsom-walker subproject? Then we can generate JavaDocs for that just like with do with the Daffodil JAPI.

Related, I'm looking into [DAFFODIL-2368](https://issues.apache.org/jira/browse/DAFFODIL-2368), which is related to how we can merge docs from multiple subprojects. I think looking at the Javadocs that results from this would be a clear way to view what parts of this API are public and what aren't.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,163 @@
...
+ * another for Types.  These are called as the DSOM is walked.
+ *
+ * Though recursion is used here to define the walk, it is not advised to use recursion between
+ * these event handlers.  Instead, consider a stack-like structure, as the DFDL Schema structure
+ * as well as the recursive method call structure can be represented by trees.
+ * @tparam T the return type of walkDSOMSchema and onWalkEnd.  This should be a class used to
+ *           represent the structure of the Schema.  You could always make it Unit if you don't
+ *           want these methods to return anything.
+ */
+abstract class AbstractDSOMWalker[T] {
```
stevedlawrence, line 98: I looks like the type parameter is only used for onWalkEnd. My conern is that this type parameter might not translate to a Java user very well. I'm not sure how Scala converts these to byte code. Perhaps it just becomes a Java generic and works fine?

But considering it's only used for the onWalkEnd method, and thuse onWalkFromRoot, I'm wondering if adds much. If a implementation of AbstractDSOMWalker needs a return value of onWalkEnd, they can implement onWalkEnd to store state and access that state after onWalkFromRoot returns? It's maybe a little uglier, but maybe avoids type parameter issues with a Java user of this API?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Method to be called whenever any element that is a Term is encountered.
+   * This applies to Sequence, Choice, GroupRef, ElementBase, etc.
+   *
+   * It is highly recommended that, when implementing this method, you pattern match
+   * some of these different sub-types at some point to handle each accordingly.
+   *
+   * @param termElement the term element
+   */
+  def onTermBegin(termElement: TermView): Unit
```
stevedlawrence, line 126: One concern with Term is that it potentially requires the user to pattern match to figure out what the term is and what to do with it. This is very natural in Scala, but in Java I think this is going to be a much of if-statements with various ``instanceOf``'s and type casting. Maybe not a big deal? I'm not sure.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,163 @@
...
+  def onTermEnd(termElement: TermView): Unit
+
+  /**
+   * Method to be called when either a Simple or Complex Type element has been encountered.
+   * This is just the element for <simpleType> or <complexType>, so no implementation is necessary if you
+   * do not care about these wrapper portion; its children will be called automatically.
+   *
+   * @param typeElement either a Simple or Complex type
+   */
+  def onTypeBegin(typeElement: Either[SimpleTypeView, ComplexTypeView]): Unit
```
stevedlawrence, line 142: Either's are a scala construct, which means this would likely be difficult, or at least uncomfortable, for a Java user to use. Perhaps we have a shared parent of SimpleTypeView and ComplexTypeView and require the user to use isntanceOf, similar to the onTerm methods?

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -99,6 +99,8 @@ final class ProcessorFactory private(
       new SchemaSet(optRootSpec, schemaSource, validateDFDLSchemas, checkAllTopLevel, tunables,
         compilerExternalVarSettings))
 
+  def walkDSOM[T](walker: AbstractDSOMWalker[T]): T = walker.walkFromRoot(sset.root)
```
stevedlawrence, line 102: Normally one would call ``processorFactory.onPath(..)``, which asserts that there are no errors and I think can result in finishing some lazy compilation? Two questions, more directed at Mike who I think knows more about the internal compialtion steps than me:
1. Should we also assert here that there are no errors?
2. Do we need to also do something to account for lazy compilation, such as ExecutionMode.usingCompilerMode? Is it possible for an SDE to get hit while walking and performing lazy compliation, or should all that be finished by the time onPath is called?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ModelGroup.scala
```diff
@@ -177,7 +180,7 @@ abstract class ModelGroup(index: Int)
    *
    * This also depends on groupMembersDef overrides all being lazy val.
    */
-  final lazy val groupMembers: Seq[Term] =
+  override final lazy val groupMembers: Seq[Term] =
```
stevedlawrence, line 183: Giving access to groupMembers probably makes it so the whole walker stuff isn't really necessary. Someone could walk the DSOM entirely with their own implementation rather than relying on all the callbacks and having to store state. Not a big deal I guess, and maybe a benefit?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
```diff
@@ -53,9 +54,9 @@ trait TypeBase {
 trait NonPrimTypeMixin
 
 sealed trait SimpleTypeBase extends TypeBase
-  with HasOptRepTypeMixin {
+  with HasOptRepTypeMixin with SimpleTypeView {
 
-  def primType: PrimType
+  override def primType: PrimType
```
stevedlawrence, line 59: Giving direct access to PrimType is potentially problematic from a backwards compatibility perspective. This is something that has changed fairly frequently. I wonder if we also want a walker specific view of PrimType's as well?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
```diff
@@ -18,16 +18,19 @@
 package org.apache.daffodil.dsom
 
 import java.util.UUID
+
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.grammar.TermGrammarMixin
 import org.apache.daffodil.schema.annotation.props.gen.YesNo
-import java.lang.{ Integer => JInt }
+import java.lang.{Integer => JInt}
```
stevedlawrence, line 25: The style we standardized on is to use spaces, though I think that's not the default in Eclipse. There should be a setting to change this though.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.daffodil.dpath.NodeInfo.PrimType
+
+trait CommonContextView {
+  def namespaces: scala.xml.NamespaceBinding
+}
+
+trait TermView extends CommonContextView {
+  def isArray: Boolean
+  def isOptional: Boolean
+  def walkDSOM[T](walker: AbstractDSOMWalker[T]): Unit = {
```
stevedlawrence, line 29: Do we want these walkDSOM function to be protected? Is there a reason that user might want to call walkDSOM directrly? We can also make them public in the future if needed, but it's much harder to switch to make them protected if they are already public.

andrewjc2000: The walk is currently initiated here:
`final def walkFromRoot(schemaSetRoot: RootView): T = {
    onWalkBegin(schemaSetRoot)
    schemaSetRoot.walkDSOM(this)
    // onWalkEnd will return the value returned by this method!
    onWalkEnd(schemaSetRoot)
  }`
As you can see, this method requires `walkDSOM` to be a public member of `TermView`.  I am not quite sure how exactly the walk would begin without this method being public.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
andrewjc2000: Sorry, I'm not sure why IntelliJ changed this!  I'll change it back in the next commit - do you all have a styling XML file that I can use for ScalaStyle or a similar tool?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
stevedlawrence: Ah, that's right. I guess the only way to do it would be to have the walkDSOM function not be on the View but be on the actual DSOM SchemaComponents, and all the walking occurs on the internal/private/unstable API. Unclear if that adds much benefit though. Perhaps there's value in allowing the implementation to change how the DSOM is walked.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Term.scala
stevedlawrence: I don't use IntelliJ, but here's our page on IntelliJ setup, Mike added some comments about some style changes:

https://cwiki.apache.org/confluence/display/DAFFODIL/IntelliJ+IDEA+Setup

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: While we do want stability, I think we should separate that from having a need for Java support. This particular API is so close to the core of the Daffodil schema compiler that I think requiring it to be scala only is fine. I don't think it is necessary for people to be able to write extensions that walk the internals parse trees in Java. 

It's not that this is not possible, it's just something I think adds little value for us to support, at least for now with this being the very first Daffodil "metadata integration".

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: There are different walk patterns yes. So I think keeping the walkers on the traits where they can be overridden or new ones added makes sense.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
stevedlawrence: One concern with requireing scala is building a nar for a NiFi controller/processer uses a maven archetype template to build the project structure. If the API requires Scala, building such a nar might be problematic, or at least this generated maven project needs to be modified support scala compilation. I assume that's possible, but I'm not sure. Other integrations might have similar issues.

And if the the utltimate goal is to upstream this into NiFi tree, it almost certainly needs to be Java friendly--as far as I know, NiFi doesn't have any Scala code in their source tree.

It we decide to only support scala (which doesn't seem unreasonable), and if there's a potential for changes to support Java, I would at least recomend we say that this API is experiemental and subject to change so that we don't have to worry as much about backwards compatibility and can more easily make such changes.

mbeckerle: Expermental until we have a second workable metadata integration sounds like the right thing.

andrewjc2000: It is possible to build nars using scala code, I previously had all of this code in a maven project that mixed NIFI-specific Java code and Scala code.  It only required a few additional packages to include.

That being said, if the NiFi developer doesn't have to code in Scala, they don't have to worry about any of that - they would only need to do this configuration if they wanted to implement AbstractDSOMWalker in Scala

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
andrewjc2000: I can understand how that could cause some issues.  What would be a more stable way to access information about a simple type without exposing the internal classes?  Would it be enough to simply add a String member of SimpleTypeBase that returns `primType.toString`?  My NiFi Record implementation is very dependent on pattern matching Simple types, and I imagine most implementations would be as well.

stevedlawrence: You could probably just use the same trait trick same as the Views (e.g.  trait that's like org.apache.daffodil.dsom.walker.IntegerView which extends dsom.walker.PrimTypeView) and then the dpath.PrimType objects extend the appropriate view. Maybe this is excessive though? I'm trying to imagine a clean API where the user only has access to things in dsom.walker.*  There's a lot of duplication though.

andrewjc2000: You mentioned that `PrimType` is changed frequently - would this list pose an issue because of that? i.e. wouldn't future developers have to go change this list of primitive Views every time `PrimType` is changed?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
stevedlawrence: Actually, these walkDSOM functions can't really be easily modified since the objects that implement the *View traits are provided by Daffodil. If we want the user to be able to implement custom walking logic, we probably need to go back something like how it was before where the Walker has the walk logic rather than the Views.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: To make this java friendly, one would typically use a visiting pattern so that what we do in scala via a match on types becomes calling a method which is overloaded on types of a parameter. I think we'd have to try it to see how it works out.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
stevedlawrence: I think list of PrimTypes isn't changed that often (and likely won't change anytime soon). The things that do change relatively frequently is the functions defined in the prim types. Because these change so often, we might not want those members to be publicly visible to a dsom walker. The View trait allows us to decide which members are made visible, which I suspect is not going to be many, if any at all. So you can still match on thes prim type views to figure out the type.

You are correct that if a new PrimType is added, we'll also need to add a PrimTypeView as well. Not much we can do about that though.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/RecordWalker.scala
stevedlawrence: In theory, the only thing that needs to be in the Daffodil codebase is the API, which is essentially the AbstractDSOMWalker and all the Views traits. Nothing depending on NiFi should go in daffodil-core because that means core depends on NiFi and thus anyone using Daffodil now needs a NiFi jar--the use of Daffodli shouldn't depend on NiFi. 

I don't think it really matters where the AbstractDSOMWalker and View traits go. They can probably just stay in core. Note that they can't go in the japi/sapi directories. This is because daffodilxXapi currently depends on daffodil-core (xapi is essentially a basic wrapper around core). If these AbstractDSOMWalker and View traits were in the daffodil-xapi, then daffodil-core would then depend on daffodil-xapi, so we'd have a circular dependency. So the AbstractDSOMWalker and View traits either stay in daffodil-core or in a new subproject. Since we are considering this experiemental, maybe it makes sense to keep it in core, and when we finalize an API and move it from experimental, we can move it to a separate dir where we can generate public java/scaladocs? I guess this is essentially what we've done with daffodil-udf. I consider it a public API so it makes sense to get it's own directory where we can generate docs for it.

So where does everything go that uses NiFi? It should probably all stick together in one place. Perhaps that place right now is the nifi-daffodil repo, just to make it very obvious what is NiFi and what is Daffodil, and to highlight what changes are Daffodil API additions and what are NiFi usages of this API? Once that's finalized and the nifi-daffodil repo is in a stable state and the experimental API is considered sufficient, it then becomes pretty trivial to either merge that into Nifi if they are willing to accept it, or merge it into a new Daffodli "integrations" directory for it to live. Or it could just remain separate--though I'd prefer it get merged somewhere.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: I opted for creating a shared parent in the latest revision - if we decide to stick with that then yes, in a Java implementation they would have to do instanceOf and casting

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: While I actually think Scala and Java Generics are pretty compatible, except for the fact that Scala supports things like Any, AnyRef, Unit, etc. while Java doesn't have those, I went ahead and removed the generic type.  It only makes things look slightly less clean for RecordWalker, but I think the fact its only purpose was to denote an awkward return type of one method it makes sense to take it out.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
andrewjc2000: I went ahead and defined an Enum type called PrimTypeView and created Enum representations for all the "PrimType" enum types within it.  I could not find a way to make the NodeInfo class import *anything* from `dsom.walker`, so I was unable to make it extend any trait I defined there.  Instead, I added a new field to `SimpleTypeBase` called `primTypeView` which just matches `primType` to my new Enum type.  Let me know if you think this approach should be changed; the pattern matching is done in `SimpleTypeBase` itself.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: I could always change `walkFromRoot` to a non-final method.  This would allow the user to override it and then instead of doing `schemaSetRoot.walkDSOM(this);` they could call their own custom helper method, similar to what we had before.  Keeping the current code intact while doing this would enable the existing walking strategy to be the default while also allowing the user to override with their own implementation.

stevedlawrence: Seems reasonable. So it sounds like basically the API becomes all the different Views and their members, plus some new public method to get the RootView. WIth the RootView, the user can traverse the DSOM tree however they want. And then we have the recursive DSOM walker implementation for users that just want to get a bunch of events and not have to worry about how things DSOM is walked. This feel like it makes the API a little larger and could make it difficult to change in the future, but as long as its experimental until it seems stable that's fine with me.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
stevedlawrence: Doesn't seem unreasonable. I had forgotten that NodeInfo is defined in daffodil-runtime1. The problem is daffodil-core depends on daffodil-runtime1, (core creates many of the objects defined in runtime1), so runtime1 can't use anything defined in core or we'd have a circular dependency. They way you handled it seems reasonable.

If you wanted to do it with mixin views, you would need to define the PrimTypeView traits in the daffodil-runtime1 class. They can still be in a dsom.walker package, just in a different subproject/jar.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: This would make the API larger due to this extra flexibility.  I went ahead and removed the `final` modifier from this method in the latest revision.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala
andrewjc2000: I went ahead with that route, it seemed a lot more elegant to integrate the View paradigm without adding any fields, just like what had been done for everything else.  I put all the Views related to PrimType in their own file in runtime-1 and most of the primitive enums now implement them

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/RecordWalker.scala
andrewjc2000: It definitely makes sense to keep the API and any NiFi-related/dependent code separate; in the latest revision I removed RecordWalker and SchemaRecordNode from core and made some general classes for testing to replace them.  I think it would make the most sense to move both of these to the nifi-daffodil repo, but as it currently stands they're both written in Scala and I'm not sure if there's necessarily a 1-1 Java conversion.

andrewjc2000: Now that the walk is initiated externally from `ProcessorFactory` rather than from the walker itself, there is currently no way to start it via the Java or Scala API.  I am going to go ahead and add the `walkDSOM` method to the `ProcessorFactory` in the JAPI and SAPI modules; let me know if you think something else should be done instead.

stevedlawrence: Re: the nifi-daffodil repo, if you're able to modify the nifi-daffodil repo to support scala, it should work fine in theory. I suspect the new controllers are mostly separate from the existing processors?

Re: walkDSOM on ProcessorFactory, I think after discussions the consensus is that while a recursive walk is useful, there may be cases where it isn't sufficient. So I would suggust that we move all the ``walkDSOM`` functions off of the Views and onto the AbstractDSOMWalker (just like you had originally), so the AbstractDSOM walker has all the logic for how to recursively walk a DSOM. Then the only new function need on the the ProcessorFactory is an ability to get the RootView. If one wants to traverse this RootView manually, they can do so. But if they just want to traverse it recrusively and get events, they can create a concrete implementation of the AbstractDSOMWalker and implement those events (and potential override the walkDSOM function if they want. So for example:
```scala
val rootView = processorFactory.rootView
val walker = new DSOMWalkerImpl()
walker.walkDSOM(rootView)
```
So the View's don't contain any logic about how they should be traversed, since they don't really know that. Recurisve makes sense, but presumable there are reasonable traversals. And technically the AbstractDSOMWalker isn't even neccessary--the views have all the information needed to traverse them. But to be helpful, it makes sense to also provide a AbstractDSOMWalker interface that contains all the logic for traversing recursively so users don't have to think about that and can just think of it as a series of events.

So this is very similar to what you originally started with, but just with Views to hide internal details of our dsom objects.

daffodil-core/src/test/resources/test/standard-xsd/bmp.xsd
```diff
@@ -0,0 +1,656 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
stevedlawrence, line 2: We generally avoid putting large schemas for file types inside the daffodil repo. Instead, we create smaller schemas that test the cases we are concerned about.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestWalker.scala
```diff
@@ -0,0 +1,335 @@
...
+}
+
+/**
+ * The sole purpose of this Companion Object is to contain this one static variable.
+ */
+object RecordWalker {
+  /**
+   * Used to suppress all of the extra printing from running test cases
+   */
+  final val PRODUCTION_MODE: Boolean = true
```
stevedlawrence, line 334: I'm concerend that this test code is fairly complex, and that it's pretty easy for there to be a small bug in here which could lead to us not realizing there's an issue with the walker. I'm wondering if there's a less complex way to test the DSOM walker to verify that it's doing the right thing?

We really just want to validate that the DSOM walker creates the correct events in the correct order. We don't really need to use these events to build anything to test that it's walking correctly. Perhaps it would be simpler if the TestDSOMWalker just received an ordered list of expected events and properties of those events. As the TestDSOMWalker walks the tree and receives events, it just ensures that they match the expected list of events.

There's probably multiple approaches, but the first thing that comes to my mind is to implement customer the walker.Views that have a custom comparision function to compare things we want to test. Note that we don't necessarily need to check all properties, for example, it might make sense to ignore group, groupMembers, simpleType, complexType, etc. since these are all implied correct if we get the correct events. So one of these might look look something like this:
```scala
case class TestRootView(
  override isSimpleType: Boolean,
  override isComplexType: Boolean,
  ...)
  extends walker.RootView {
  
  override def equals(that: AnyRef): Boolean =
    that match {
      case rv: RootView =>
        this.isSimpleType == rv.isSimpletype &&
        this.isComplexType == rv.isComplexType &&
        ...
    case _ => false
  }
}
```
We'd have a concrete TestView for each walker View we could get from the walker. Then the TestDSOMWalker could look something like this:
```scala
class TestDSOMWalker(expectedEvents: Seq[TestView]) {
  var events = expectedEvents
  def onSomeEvent(view: walker.SomeView): Unit = {
    val head :: tail = events
    assertTrue(head == view)
    events = tail 
  }
}
```
Then to test, this becomes:
```
def testWalker1() = {
  val expectedEvents = Seq(
    TestRootView(....),
    TestComplexTypeView(...),
    TestSequenceView(...),
    ...
  )
  ...
  val testWalker = new TestDSOMWalker(expectedEvents)
  testWalker.walk(pf.rootView)
}
```

There's still a little complexity here--we need to create these concrete views and ensure we're testing the right properties, including start/end events. But I think it's a little easier to verify that the walker implementation is doing the right thing with a simpler approach.

project/Dependencies.scala
```diff
@@ -29,6 +29,7 @@ object Dependencies {
     "xml-resolver" % "xml-resolver" % "1.2",
     "commons-io" % "commons-io" % "2.6",
     "jline" % "jline" % "2.14.6",
+    "org.scalatest" %% "scalatest" % "3.2.0" % "test"
```
stevedlawrence, line 32: I think it would be preferrable if these tests were changed to junit just so that all of our test uses the same mechanism.

mbeckerle: I would go farther than this. I really hate scalatest. It is a pointless DSL that enables English-centric people to use English-centric sentence structure words like "should" which is unnatural in other languages (words like should become verb conjugation in some languages), to express tests. This is unfriendly to non-native English speakers. 

I think scalatest primarily serves as an example of how to program DSLs in scala. As a test rig itself it's unnecessary, and in fact problematic. 

It encourages tests to contain text strings that are for reporting purposes, but those strings must be kept consistent with the test itself. 
It is better to just choose a good name for the test method and test class. 

This is from the scalatest web page. I've added my comments:

```
class ExampleSpec extends AnyFlatSpec with should.Matchers {

  "A Stack" should "pop values in last-in-first-out order" in { 
    //
    // This English-centric phrasing has no advantage over @Test def stackShouldPopInLIFOOrder()
    //
    val stack = new Stack[Int]
    stack.push(1)
    stack.push(2)
    stack.pop() should be (2)
    stack.pop() should be (1)
  }

  it should "throw NoSuchElementException if an empty stack is popped" in { 
    //
    // it? Referring to what? "A Stack" is conveniently right there above in this toy example. 
    // In real code it could be off screen scrolled away. This is not a strength of natural language we 
    // want to pull into formal language. 
    // 
    // The above also repeats the exception name which is 100% redundant and as that string is just for a report,
    // it can become inconsistent with the way the test evolves if the exception thrown is changed.

    val emptyStack = new Stack[Int]
    a [NoSuchElementException] should be thrownBy {
      emptyStack.pop()
    } 

   //
   // This has no advantage over @Test def stackThrowsProperException() ....
   //
  }
}

```




I think scalatest primarily serves as an example of how to program DSLs in scala. As a test rig itself it's unnecessary, and in fact problematic.

andrewjc2000: Noted, I did not realize the flaws of ScalaTest; the particular test cases I wrote happened to be helpful for representing the structure of the large schema files I wanted to verify I had the correct NiFi RecordSchema for, but I realize now how it might be more alienating for non-native English speakers, plus Unit Tests shouldn't be based around very large structures happening to work and should be more focused.  

That all said, I developed a new testing file in the latest revision that uses JUnits and while it is much simpler than what I had before. it is still a bit robust in that it checks to see if all the important Simple and Complex View types work as expected.  It is also based around a much simpler implementation of the Abstract Walker.  Feel free to comment on it.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestWalker.scala
andrewjc2000: I think I wound up with a solution that is similar to this idea; I had a testing file generate a Stack-like structure based on different View types it received, and the Tests were based around specific expected events in that List.  Since I was rewriting the Tests, I chose to automate as little as possible, I only automated the Simple Types since checking that they work was the same way for each one.  However, for each of the Complex types, I wanted to check different things like the name in some cases.  This is all in the latest revision.

daffodil-core/src/test/resources/test/standard-xsd/bmp.xsd
andrewjc2000: Yes, I noticed that - this paradigm makes more sense than what I had.  In the JUnit-based test class in the latest revision, the schemas are in the code and passed to the helper "dfdlTestSchema" I saw being used in the other test cases.  Let me know if you think the one I made for SimpleTypes is too long for a Unit Test environment; I could always break up each SimpleType into its own test

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestWalker.scala
andrewjc2000: I did not wind up ignoring things like complexType and simpleType in the testComplexTypes() test case, but I do have a flag for that in the new testing class if that is desired - let me know if you prefer that.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Starts a DSOM walk from the Root object.  By default, the walk is completed recursively in the helper
+   * method below, but a custom walk can be defined by overriding
+   * this method instead and calling local event handlers.
+   * @param schemaSetRoot The root element of the DFDL Schema.  This will be the starting point of the traversal
+   */
+  def walkFromRoot(schemaSetRoot: RootView): Unit = {
+    onWalkBegin(schemaSetRoot)
+    walkerHelper(schemaSetRoot)
+    // onWalkEnd will return the value returned by this method!
```
stevedlawrence, line 142: Can remove this comment now that walkerEnd doesn't return anything.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,177 @@
...
+          onTypeBegin(element.simpleType)
+          onTypeEnd(element.simpleType)
+        }
+      case groupRef: GroupRefView =>
+        if (!groupRef.isHidden) {
+          groupRef.groupMembers.foreach(walkerHelper)
+        }
+      case modelGroup: ModelGroupView =>
+        modelGroup.groupMembers.foreach(walkerHelper)
+      case _ =>
```
stevedlawrence, line 171: Does thid default case ever get hit? I think this is impossible?

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
```diff
@@ -0,0 +1,50 @@
...
+  override def onTermBegin(termElement: TermView): Unit = {
+    if (onlyElements) {
+      termElement match {
+        case _: ElementBaseView => nodeArr += Left(termElement)
+        case _ =>
+      }
+    } else {
+      nodeArr += Left(termElement)
+    }
+  }
```
stevedlawrence, line 39: I think maybe a cleaner or more scala-y way to write this would be something like this:
```scala
termElement match {
  case _: ElementBaseView => nodeArr += Left(termElement)
  case _=> if (!onlyElements) nodeArr += Left(termElement)
}
```

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
```diff
@@ -0,0 +1,50 @@
...
+  override def onTermEnd(termElement: TermView): Unit = {}
+
+  override def onTypeBegin(typeElement: TypeView): Unit = {
+    if (!ignoreTypeWrappers) {
+      nodeArr += Right(typeElement)
+    }
+  }
+
+  override def onTypeEnd(typeElement: TypeView): Unit = {}
+}
```
stevedlawrence, line 50: I'm a little concerned that this ignores a lot of the walker functions. It ignores all End events, and depending on the parameters even more.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestDSOMWalker.scala
```diff
@@ -0,0 +1,229 @@
...
+        }
+      case Left(someClass) => fail(s"The 'age' element did not have a SimpleTypeView, it was a ${someClass.getClass}!")
+      case Right(someClass) => fail(s"The 'age' element did not have a SimpleTypeView, it was a ${someClass.getClass}!")
+    }
+    nodeStack(6) match {
+      case Left(_: GroupRefView) =>
+      case Left(someClass) => fail(s"The Sequence element did not contain a GroupRef child, it was a ${someClass.getClass}!")
+      case Right(someClass) => fail(s"The Sequence element did not contain a GroupRef child, it was a ${someClass.getClass}!")
+    }
+  }
```
stevedlawrence, line 95: These feels a little verbose to me. All the fail messages are perhaps helpful when a test fails, but that should be rare, and it makes it difficult to visually determine that the tests are checking the right things, which I'd argue is more important.

I think part of the problem this looks so verbose is the the node stack being an Either, so you need to do a bunch of left/right checks. But that doesn't give us much. You still need to do type matcing inside the Left/Right, so the Left/Right just adds a layer of indirection. Instead, you can just make nodeArray be an Array[AnyRef].

Then, if we removed all the matching and messages, the above can be condensed down to something like this:
```scala
assertEquals(7, nodeStack.size)
assertTrue(nodeStack(0).isInstanceOf[RootView])
assertEquals(nodeStack(0).asInstanceOf[RootView].name, "PersonData")
assertTrue(nodeStack(1).isInstanceOf[ComplexTypeView])
assertTrue(nodeStack(2).isInstanceOf[SequenceView])
assertTrue(nodeStack(3).isInstanceOf[ChoiceView])
assertTrue(nodeStack(4).isInstanceOf[ElementBaseView])
assertEquals(nodeStack(4).asInstanceOf[ElementBaseView].name, "age")
assertEquals(nodeStack(4).asInstanceOf[ElementBaseView].isSimpleType, true)
assertTrue(nodeStack(5).isInstanceOf[SimpleTypeView])
assertTrue(nodeStack(5).asInstanceOf[SimpleTypeView].primType.isInstanceOf[IntView])
assertTrue(nodeStack(6).isInstanceOf[GroupRefView])
```
Which I think is much easier to visually confirm that we're checking the right things. The suggestion I previously made about having concrete implemenations of the Views and an equal function make these checks a bit more automated and maybe a bit more clear (probably gets rid of all the instance of stuff), at the expense of needing a bit more code. But something more along the lines of this doesn't seem unreasonable to me.

Note that this is much less scala-y, but readability often trumps that, especially when it comes to tests.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -373,6 +374,12 @@ class ProcessorFactory private[japi] (private var pf: SProcessorFactory)
     new DataProcessor(dp)
   }
 
+  /**
+   * Exposes the RootView object corresponding to this ProcessorFactory.  This can
+   * be used to start a walk using the walkFromRoot method in a DSOM Walker.
+   */
+  def rootView: RootView = pf.rootView
```
stevedlawrence, line 381: We probably should document that this is an experiemental API and is subject to change. Once we stabalize this and confirm we have the right functionality, we can remove the experiemtal tag and make the Views part of the public API.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,177 @@
...
+trait GroupRefView extends ModelGroupView {
+  def isHidden: Boolean
+}
+
+trait ElementDeclView extends CommonContextView {
+  def isSimpleType: Boolean
+  def isComplexType: Boolean
+  def simpleType: SimpleTypeView
+  def complexType: ComplexTypeView
+  def hasDefaultValue: Boolean
```
stevedlawrence, line 61: Do we need to provide a method to get the default value? I assume if something wants to know about default values it will also want to know what that default it?

mbeckerle: maybe not? Why do we need to know if something has a default value even? Isn't that a format issue? Maybe we need it to point out an error where a required field can be absent from the NiFi record when unparsing it? But I don't actually see why we need the value.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
andrewjc2000: It ignores all end events because I wanted the end result to be a stack of every element encountered rather than a stack of Nodes with child elements.  In all the implementations so far, End events were just used to remove the current element from the stack, which I don't want here.

stevedlawrence: Understood, but we should still have tests that verify that end events are happening correctly. If there were a bug in the AbstractDSOMWalker that caused certain end events to never happen or to happen out of order, we wouldn't know about it. Whether this test walker does those checks or if it's handled by in each test itself doesn't really seem that important, but something should be able to verify that it's correct.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: NiFi doesn't need this; I merely included it because I was trying to mimic the structure of `ElementDeclMixin`.  I notice that `ElementDeclMixin` has `defaultValueAsString` member, would that be the correct member to add?  If so, then I could add it, but I could also just remove `hasDefaultValue` from the ElementDeclView if this is not an important attribute.

stevedlawrence: If NiFi doesn't need it, I'd vote for just removing it. We can always add it in the future if some other walker implementation needs it.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/TestDSOMWalker.scala
andrewjc2000: I agree that it is robust, and I had it this way mainly for the sake of pattern matching and being able to figure out types of specific View elements on the stack when tests were failing.  That being said, for simplicity sake, I think the series of asserts based on `isInstanceOf` and `asInstanceOf` would be best.  I tried implementing a case class as you suggested above, and it turned out to be an even longer solution.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
andrewjc2000: What is the preferred way to mark an API as experimental in Daffodil?  Is there an Annotation or ScalaDoc tag, or should I simply state it's Experimental in the comment describing `rootView`?

stevedlawrence: I know the Scala language sometimes uses an ``experimental`` object that things go in, so we could do something like that:
```scala
class ProcoessorFactory(pf: SProcessorFactory) {
  ...
  object experimental {
    def rootView: RootView = pf.rootView
  }
  ...
}
```
To access the rootView, one would do something like this:
```scala
val rootView = pf.experimental.rootView
```
That makes it very obvious when you are using something that is experimental and that the API might change in the future. If we do go this route, I wonder if it also make sense to create a new sbt subproject and move all the new View traits and AbstratDSOMWalker in an experimental package (eg. org.apache.daffodil.experimental.dsom.walker), similar to how we have the daffodil-udf package split out? And we can use the new unidoc stuff to include it in public API docs?

Not sure how I feel about this though. Maybe it's too obtrusive?

I think the other option is to just have a blurb like "This API is experimental is and subject to change". I don't know if there's an annotation or standard javadoc practice for this kind of thing.

mbeckerle: I think a separate submodule is going too far unless that's necessary someday to grab the scaladoc from this and incorporate it into the public-facing scaladoc set. 
I like the experimental object technique you showed. That's sufficient to me for now.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
mbeckerle: If f NiFi doesn't need it, definitely leave it off. It's pretty good evidence that perhaps nobody will need it, and every thing we put on these views will cause someone to wonder whether or not they *need to be using it*.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
stevedlawrence: Yeah, the submodule is really only needed when we want to include View stuff in public docs. It does mean in order to use the walker/views you kindof have to know what you're doing and know where to find these views. But that's not unreasonable for experimental things. Agree that the subprojet probably isn't needed right now.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: Agreed - Resolved in the latest revision

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: As the View structure currently stands, the default case will not ever get hit; removed this case clause in the latest revision.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
andrewjc2000: Yes, that is much cleaner - I implemented this along with another case statement based on a new boolean for the tester class in the latest revision.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/walker/BasicWalker.scala
andrewjc2000: That makes sense, I did overlook that.  I wound up simply adding the element both when the begin and end events are invoked, and then the refactored test for complex types verifies the elements corresponding to those end events occur in the expected indices in the list.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
andrewjc2000: While this solution is great and simple for Scala, it doesn't seem to translate well to Java; `pf.experimental.rootView()` does not compile, and IntelliJ thinks that `ProcessorFactory.experimental$.MODULE$.rootView()` compiles but this does not work when you go to build it, either.  Could we just create a brief inner wrapper class here with its only member as `rootView` and then make `experimental` a member of `ProcessorFactory` of that wrapper type?

stevedlawrence: Seems like a reasonable approach to me.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -0,0 +1,174 @@
...
+}
+
+trait ModelGroupView extends TermView {
+  def groupMembers: Seq[TermView]
+}
+
+trait ChoiceView extends ModelGroupView
+trait SequenceView extends ModelGroupView
+trait GroupRefView extends ModelGroupView {
+  def isHidden: Boolean
```
stevedlawrence, line 53: One last look over this (looking really good, btw), I'm wondering if we want isHidden visible? We could in theory have daffodil-core present a view of groupMembers where hidden group refs are filtered out. And then API users never even have to be aware that some things are hidden because they'll never see them. Right now that logic is handled by the AbstractDSOMWalker, but if a user didn't use that they'd have to keep hidden group refs in consideration.

I think the main issue with this is that daffodil-core can't actually modify the underlying groupMembers. Daffodil still DSOM still needs to know about hidden groups, and there is no way to actually present a View different from the concrete class. The View can only *hide* members--it can't actually make changes to those members.

I guess we just don't have an issue anywhere else where a View member needs a different type/content than the underlying member.

Not sure of a good way to handle this unless we start introduce different names for View members or allocating new classes, which probably starts getting messy. So maybe this isn't worth considering, and just something to keep in mind with how this View API works.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -338,6 +339,13 @@ class ProcessorFactory private[sapi] (private var pf: SProcessorFactory)
     new DataProcessor(dp)
   }
 
+  class ExperimentalWrapper(val rootView: RootView)
+  /**
+   * Exposes the RootView object corresponding to this ProcessorFactory.  This can
+   * be used to start a walk using the walkFromRoot method in a DSOM Walker.
+   */
+  val experimental: ExperimentalWrapper = new ExperimentalWrapper(pf.rootView)
+
```
stevedlawrence, line 348: I think it's reasonable to use the ``object experimental`` technique for the scala api. I think that makes fora cleaner scaladoc and feels more natural to scala users.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/walker/AbstractDSOMWalker.scala
andrewjc2000: I think there is a solution to this, but it would break the current paradigm of every field in the `View`s referencing an existing field.  We could replace the existing `group` element in `ComplexTypeView` with something like `optVisibleGroup`, and then the implementation for `optVisibleGroup` would pattern-match `group`; if it is hidden, return `None, otherwise return the `group` wrapped in an option.  Similarly, instead of `groupMembers` being in `TermView`, we would have `visisbleGroupMembers`, and in the implementation `visibleGroupMembers` would be a filtered list of the real `groupMembers`, excluding any `TermView` that is a `GroupRef` and is also hidden.

Now, my main concern with this is it would potentially cause more confusion, leading the end user to wonder why we're only allowed to see "visible" members, and might make traversal very annoying with having to constantly unpack Options.

stevedlawrence: Yeah, I agree yoru concern, that does feel like things could get more confusing, and DSOM is alrady conufsing enough as it is.

An another approach with maybe less confusion would be to create case classes for all our Views. Accessing a View would result in allocating a new case class and copying relevant information from the base class into the View class, which would allow modifcations/filtering to what gets copied. Though, this could result in lots of allocations, which is probably best to avoided.

I'd say these types of changes probably aren't worth it unless we find more examples where it would be useful to have the View be substantially different. Probalby not worth it for just this one case.

stevedlawrence:
@andrewjc2000, with two +1's this is ready to merge. Can you please squash all of your changes into a single commit, update the commit message based on the squashed state of the code, and add the bug number to the end of the commit message? Then this will be good to merge. Thanks for the contribution!

PR #396, opened 2020-07-10
Add tests for nested choice branch cases that seemed problematic.

mbeckerle:
These tests convinced me this is all working properly.

They should be added to the suite even though they did not
find a new bug.

See DAFFODIL-2367

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/nestedChoiceDiscriminator.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/nestedChoiceDiscriminator.tdml
@@ -1,0 +1,306 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<tdml:testSuite
+  xmlns:ex="http://example.com"
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section05/facets/NulChars.tdml
```diff
@@ -36,7 +36,7 @@
           <xs:element name="foo" dfdl:lengthKind="pattern" dfdl:lengthPattern="\x{0000}{3,3}">
             <xs:simpleType>
               <xs:restriction base="xs:string">
-                <xs:pattern value="&#xE000;*"/>
+                <xs:pattern value="[&#xE000;]*"/>
```
mbeckerle, line 39: Revert. This got carried over from another experiment.

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/nestedChoiceDiscriminator.tdml
```diff
@@ -0,0 +1,306 @@
...
+              <!--
+              If capture variable not set, then just fail here, and report an error.
+              -->
+              <dfdl:discriminator message='{ fn:concat("Unable to parse message with messageID: ", ./messageID) }'
+                                  test='{ $ex:capture }'/>
+            </xs:appinfo>
+          </xs:annotation>
+        </xs:sequence>
+        <!--
+         Otherwise, create this unrecognized capture element
```
mbeckerle, line 151: Not unrecognized capture element. This is a message parse error capture element.

daffodil-test/src/test/scala/org/apache/daffodil/section07/discriminators/TestNestedChoices.scala
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.section07.discriminators
+
+/*import org.junit.Test */
```
stevedlawrence, line 20: Remove import comment.

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/nestedChoiceDiscriminator.tdml
```diff
@@ -0,0 +1,306 @@
...
+                  <xs:choice>
+                    <xs:choice dfdl:choiceDispatchKey="{ ./messageID }">
+                      <xs:sequence dfdl:choiceBranchKey="1">
+                        <!--
+                        This techique, where we discriminate the outer choice if the choice dispatch
+                        is successful, is better during DFDL schema development because we won't backtrack
+                        and create unrecognized messages that suppress errors during the parsing of a recognized
+                        message type.
+                        -->
+                        <xs:group ref="ex:discriminate"/>
```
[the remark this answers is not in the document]
mbeckerle: That point in the spec will need an editorial correction. I'll bring this up on the DFDL workgroup email list and see if there is consensus on this, but I thought this was resolved long ago.

My understanding is that there is no PoU for a choice with direct dispatch. In the same way there is no PoU for an array with a computed number of occurrences. 

That's equivalent to saying there is a PoU but it is resolved by the direct dispatch; hence, thinking of discriminators as a stack, the discriminator of the inner choice would be set true as soon as the dispatch expression resolves and matches a branch tag. 

Hence, a discriminator within a branch of a direct-dispatch-choice applies to the discriminator of the enclosing outer choice, not the inner either way. 

There really should be identifiers on PoUs so that choices can be labeled accordingly to catch mistakes.

PR #398, opened 2020-07-15
 Use sbt-unidoc plugin for generating public API documentation

stevedlawrence:
Rather than running Javadoc/Scaladoc commands separately to build our
 public API documentation, we can now just run "sbt unidoc". That
generates both Java and Scala documentation in the root target
directory, including docs combined from the japi, sapi, and udf
subprojects. This also allows for easily adding new public API
documentation for other features, such as planned DSOM walker, by just
creating a new subproject and adding it to the unidocProjectFilter sbt
setting.
    
DAFFODIL-2368


stevedlawrence:
New PR since Github was not updating the original pull request for a fixup commit. See original pull request here for comments and merge approvals.

https://github.com/apache/incubator-daffodil/pull/397

PR #399, opened 2020-07-16
Bump version for 3.0.0 development

stevedlawrence:
DAFFODIL-2370


stevedlawrence:
I'm wondering if we should modify the LICENSE/NOTICE files to not include Daffodil version information in the jar so that we don't have to remember to update these locations. There really should just be one place where the version number should need to exist. Someting like
```
(org/apache/daffodil/util/UniquenessCache.class in lib/org.apache.daffodil.daffodil-lib-*.jar)
```
IANAL, but I feel like using a glob here is pretty unambiuous about what file is being referenced.

stevedlawrence:
Agreed, <VERSION> is much more clear. Will fix.

PR #400, opened 2020-07-16
Add tests showing multiple discriminator behavior

stevedlawrence:
One test currently fails when multiple discriminators evaluate to true
followed by a parser error that requires backtracking. The multiple
discriminators do not resolve the correct points of uncertainty.

DAFFODIL-2371


daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/multipleDiscriminators.tdml
```diff
@@ -0,0 +1,171 @@
...
+    We attempt to parse the interger element, which fails, creating a Parse Error and immediately backtrack.
+    Because the inner choice PoU is resolved, we backtrack to the outer choice.
+    Because the outer choice PoU is resolved, we backtrack again, but there are
+    no more PoU's. It is a Parse Error related to the failure to parse the
+    interger element.
+  -->
+  <tdml:parserTestCase name="multipleDiscriminators5" model="multipleDiscriminators" root="root">
+    <tdml:document>tta</tdml:document>
+    <tdml:errors>
+      <tdml:error>interger</tdml:error>
```
tuxji, line 167: Is "interger" spelled correctly?  A grep shows me no occurrences of "interger" in my incubator-daffodil checkout and a find in page tells me "interger" occurs only 3 times in this page (two times in the test case comment, one time inside the test case).  Since I see an "interger" element defined in the test schema, I assume you meant "integer" instead of "interger".

stevedlawrence: Yep, that should be "integer" to match the element name related to the parser error. Good catch. Will fix.

PR #401, opened 2020-07-24
Added test case for DAFFODIL-2375

jadams-tresys:


daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
```diff
@@ -93,6 +93,8 @@ class TestVariables {
 
   @Test def test_unsignedIntVarCast(): Unit = { runner.runOneTest("unsignedIntVarCast") }
 
+  @Test def test_choiceBranchVariables(): Unit = { runner.runOneTest("choiceBranchVariables") }
```
mbeckerle, line 96: If this test fails you want to commit with the test commented out and // DAFFODIL-2375 comment.

jadams-tresys: Right, will do

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
```diff
@@ -93,6 +93,8 @@ class TestVariables {
 
   @Test def test_unsignedIntVarCast(): Unit = { runner.runOneTest("unsignedIntVarCast") }
 
+  //@Test def test_choiceBranchVariables(): Unit = { runner.runOneTest("choiceBranchVariables") }
```
tuxji, line 96: Looks good, but please insert // DAFFODIL-2375 before this line to identify which issue the failing test is for.

PR #402, opened 2020-07-27
Fix variables getting reset incorrectly

jadams-tresys:
When the variables infrastructure was changed to be mutable, a
changedVariablesStack was introduced to track what variables had been
changed in the current scope. These variables would be reset when the
PState was reset. The problem was that new lists of variables would only
be pushed when entering a point of uncertainty and popped when leaving,
however the call to PState.reset() can happen outside of points of
uncertainty, such as with nillable elements.

This change makes it so that whenever a PState.Mark is created, a new
list of changed variables is pushed on the stack. When a mark is
discarded or the PState is reset, we pop the stack.

DAFFODIL-2374

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
@@ -197,6 +197,7 @@ final class PState private (
 
   def mark(requestorID: String): PState.Mark = {
     // threadCheck()
+    changedVariablesStack.push(mutable.MutableList[GlobalQName]())
     val m = markPool.getFromPool(requestorID)
     m.captureFrom(this, requestorID)
     m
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -197,6 +197,7 @@ final class PState private (
 
   def mark(requestorID: String): PState.Mark = {
     // threadCheck()
+    changedVariablesStack.push(mutable.MutableList[GlobalQName]())
```
mbeckerle, line 200: Please add comment that there is an invariant that every mark must be either reset to, or discarded in order for this stack to be properly maintained.

tuxji: Please add the comment as Mike asked

PR #405, opened 2020-08-05
Added test to verify fix for DAFFODIL-2354

jadams-tresys:
The changes made to fix DAFFODIL-2374 also seem to have fixed the
behavior seen in DAFFODIL-2354. This commit simply adds a test to
verify that the scenario described in DAFFODIL-2354 now works.

DAFFODIL-2354


stevedlawrence:
+1

PR #406, opened 2020-08-13
Fix multiple discriminator behavior and free up marks when resolved

stevedlawrence:
Currently we maintain a separator discriminator stack, and parser would
push, pop, and modify this stack appropriately to make parse decisions.

One problem with this implementation is that discriminators only ever
modify the state on the top of the stack, which means that multiple
discriminators are essentially no-ops, rather than resolving points of
uncertainty up the stack.

Another problem is that discriminators are not tied to Marks in any way,
so resolving a discriminator does not immediately discard the Mark, and
thus memory associated with that Mark cannot be freed until the parse
unwinds.

To resolve both issues, this patch completely removes the discriminator
stack, and instead replace it with a stack of Marks, each of which
represents a points of uncertainties. Resolving a point of uncertainty
(e.g. from a discriminator) simply discards the mark on the top of the
stack. This frees up the Mark and associated memory, as well as allows
multiple discriminator behavior to discriminate multiple points of
uncertainty.

To determine if a point of uncertainty was resolved, parsers must simply
check if the associated Mark is at the top of the stack. If it is not,
then something discriminated the point of uncertainty. To simplify this
logic, multiple helper functions and methods are created to ensure
points of uncertainty are created/destroyed/handled correctly.

DAFFODIL-2371

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
@@ -47,6 +47,7 @@ import org.apache.daffodil.processors.unparsers.TypeValueCalcUnparser
 import org.apache.daffodil.processors.parsers.InitiatedContentDiscrimOnIndexGreaterThanMinParser
 import org.apache.daffodil.processors.parsers.InitiatedContentDiscrimChoiceParser
 import org.apache.daffodil.processors.parsers.InitiatedContentDiscrimChoiceAndIndexGreaterThanMinParser
+import org.apache.daffodil.processors.parsers.InitiatedContentDiscrimChoiceOnlyOnFirstIndexParser
 
 abstract class AssertBase(
   decl: AnnotatedSchemaComponent,
@@ -121,29 +122,40 @@ case class InitiatedContent(
   override val forWhat = ForParser
 
   override def parser = {
+    // Each of the parsers that this could create always appear immediately
+    // after an initiator parser. So they can all assume that the previous
+    // initiator was successfully parsed and thus resolve points of uncertainty
+    // appropriately
+ 
     t match {
-      case eb: ElementBase => {
+      case eb: ElementBase if eb.isArray => {
         (mg, eb.optPoUMinOccurs) match {
           case (sq: SequenceTermBase, Some(min)) => {
-            // no PoU if occursIndex is <= min. After that
-            // there is a PoU, so we set discriminator true
-            // if we evaluate this parser, since this parser is invoked
-            // only if the initiator's parser was successful.
+            // This is an array of elements inside a sequence with initiated
+            // content. There is no PoU if occursIndex <= min. All elements
+            // after that have a PoU, which are resolved by this parser. Note
+            // that min here is the value of optPoUMinOccurs, which is slightly
+            // different than value of minOccurs, e.g. when occursCountKind="parsed"
+            // this value is zero rather than the value of minOccurs.
             new InitiatedContentDiscrimOnIndexGreaterThanMinParser(min, eb.erd)
           }
           case (ch: ChoiceTermBase, None) => {
-            // if any array elements initiators are parsed successfully we
-            // discriminate the choice. But there is only the choice PoU
-            // to set discriminator on.
-            new InitiatedContentDiscrimChoiceParser(eb.erd)
+            // This branch of a choice is a fixed number of array elements.
+            // Because they are fixed, there is no PoU associated with the
+            // elements, only the choice. This must resolve the choice PoU only
+            // the first time so as not to resolve any other PoUs.
+            new InitiatedContentDiscrimChoiceOnlyOnFirstIndexParser(eb.erd)
           }
           case (ch: ChoiceTermBase, Some(min)) => {
-            // When occursIndex is 1, then we set the choice discriminator
-            // which is 1 below the top of the discriminator stack.
-            // Note that the min is supposed to be 1 or greater.
-            //
-            // If the occursIndex is > min then we set the array discriminator
-            // which is top of discriminator stack.
+            // This branch of a choice is for an array of elements with a
+            // minimum number of occurrences. When occursIndex <= min, there
+            // are no PoU's for the array elements, so this parser only
+            // resolves PoU's once that index has been reached. However, there
+            // is still a PoU for the choice, which must only be resolved once
+            // when we parsed the first array element. Note that min here is
+            // the value of optPoUMinOccurs, which is slightly different than
+            // the value of minOccurs, e.g. when occursCountKind="parsed" this
+            // value is zero rather than the value of minOccurs.
             new InitiatedContentDiscrimChoiceAndIndexGreaterThanMinParser(min, eb.erd)
           }
           case x => Assert.invariantFailed("Guard should exclude this case: " + x)
--- a/daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
+++ b/daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processors.parsers
+
+import scala.reflect.macros.blackbox.Context
+
+
+object PointOfUncertaintyMacros {
+
+  /**
+   * For creating a point of uncertainty and ensuring that it is cleaned up
+   * appropriately. This allows the use of standard scala-style syntax to
+   * create and use points of uncertainties, while avoiding allocating
+   * functions. Thus, this allows something like this:
+   *
+   *   pstate.withPointOfUncertainty { pou =>
+   *     // code that implements parsing, and may potentially reset to,
+   *     // discard, or resolve this pou variable
+   *   }
+   *
+   * Upon completion of the block, if the new point of uncertainty has not been
+   * discarded, reset to, or resovled, this pou will then be discarded,
+   * ensuring that Marks are always cleaned up appropriately.
+   */
+  def withPointOfUncertainty[A, B](c: Context)(pouID: c.Expr[String], context: c.Tree)(func: c.Expr[A => B]) = {
+
+    import c.universe._
+    
+    val state = TermName(c.freshName("state"))
+    val id = TermName(c.freshName("id"))
+    val pou = TermName(c.freshName("pou"))
+    val ctx = TermName(c.freshName("context"))
+
+    func.tree match {
+
+      // The func tree is something like (param => body), where param is the
+      // name the caller wants to name the point of uncertainty mark, and body
+      // is the code that uses this point of uncertainty. We transform this
+      // code to create a new point of uncertainty, evaluate the body, and then
+      // use a finally block to ensure we discard the PoU if it has not been
+      // resolved.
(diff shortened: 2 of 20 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
```diff
@@ -121,29 +122,34 @@ case class InitiatedContent(
...
           }
           case (ch: ChoiceTermBase, Some(min)) => {
-            // When occursIndex is 1, then we set the choice discriminator
-            // which is 1 below the top of the discriminator stack.
-            // Note that the min is supposed to be 1 or greater.
-            //
-            // If the occursIndex is > min then we set the array discriminator
-            // which is top of discriminator stack.
+            // This branch of a choice is for an array of elements with a
+            // minimum number of occurrences. When occursIndex <= min, there
```
mbeckerle, line 148: Just thinking we need to be sure this applies only when we're actually using the minOccurs, i.e., when occursCountKind is not 'parsed', because then we ignore the minOccurs.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
```diff
@@ -0,0 +1,97 @@
...
+
+object PointOfUncertaintyMacros {
+
+  /**
+   * For creating a point of uncertainty and ensuring that it is cleaned up
+   * appropriately. This allows the use of standard scala-style syntax to
+   * create and use points of uncertainties, while avoiding allocating
+   * functions. Thus, this allows something like this:
+   *
+   *   pstate.withPointOfUncertainty { pou =>
```
mbeckerle, line 31: So this introduces and binds the new point of uncertainty to this name pou?
Cool. I can see using this technique other places also particularly if it helps avoid allocation, which will make profiling easier.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
```diff
@@ -0,0 +1,97 @@
...
+      // code to create a new point of uncertainty, evaluate the body, and then
+      // use a finally block to ensure we discard the PoU if it has not been
+      // resolved.
+      case q"""($param: $_) => $body""" => {
+
+        // The "body" uses the "param" variable name, and scala doesn't like it
+        // if we just create a new variable with this name, with a very unclear
+        // error message. Instead, we need to create a "fresh" variable name to
+        // hold he new point of uncertainty, and then transform the body so
+        // that all instances of "param" are replaced with our new fresh name.
```
mbeckerle, line 63: This seems pretty fragile. E.g., reflection to a name will clearly break this. 
I would add a big scaladoc warning so that people keep the scope over which this transform is applied small.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1592,6 +1578,30 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
         override val name = "parser" // scala -xcheckinit reported this was uninitialized
       } with InfoProcessorBase
 
+      object InfoPointsOfUncertainty extends DebugCommand with DebugCommandValidateZeroArgs {
+        val name = "pointsOfUncertainty"
+        override lazy val short = "pou"
+        val desc = "display list of unresolved point of uncertainties"
```
mbeckerle, line 1584: grammar: points of uncertainty.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/InitiatedContentParsers.scala
```diff
@@ -47,13 +57,18 @@ final class InitiatedContentDiscrimChoiceAndIndexGreaterThanMinParser(
...
+  final def parse(start: PState): Unit = {
+    if (start.arrayPos > min) {
+      // resolve the point of uncertainty associated with array elements once
+      // we have parsed min occurrences. There should not be a point of
+      // uncertainty associated with this array until we have parsed more than
+      // min occurrences
+      start.resolvePointOfUncertainty()
+    }
+
+    if (start.arrayPos == 1) {
```
mbeckerle, line 69: So if min is 0 and start.arrayPos is 1, I think this logic calls start.resolvePointOfUncertainty() twice, which can't be right can it?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NilEmptyCombinatorParsers.scala
```diff
@@ -33,55 +27,23 @@ abstract class NilOrValueParser(ctxt: TermRuntimeData, nilParser: Parser, valueP
...
       nilParser.parse1(pstate)
 
       if (pstate.processorStatus ne Success) {
-        pstate.reset(mark)
-        mark = null
+        // did not find a nil. reset back to the pou and try to parse the value
+        pstate.resetToPointOfUncertainty(pou)
         valueParser.parse1(pstate)
       } else {
-        pstate.discard(mark)
```
mbeckerle, line 47: Loving these huge cleanups.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -304,23 +302,38 @@ final class PState private (
...
+  }
+
+  def resolvePointOfUncertainty(): Unit = {
+    if (!pointsOfUncertainty.isEmpty) {
+      val pouPop = pointsOfUncertainty.pop
+      discard(pouPop)
+    }
+  }
+
+  def isPointOfUncertaintyResolved(pou: PState.Mark): Boolean = {
```
mbeckerle, line 335: Seems odd to me. I expected this to search down the stack, it's only checking the top. Can you add scaladoc explaining why that's sufficient? The principles of operation here are quite different now, because the parsers manipulate PoUs via a local variable now, not manipulating or searching a stack at all.  So the stack is now just to identify the current unresolved PoU. This all needs to be in code comments/scaladoc not just in the intro to this PR.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesExpressions.scala
stevedlawrence: Yeah, the ``min`` variable here actually comes from ``optPoUMinOccurs above``, which is ``Some(0)`` in the case of occursCountKind="parsed", regardless of the actual value of minOccurs. I'll update this comment to make it clear that this min is not necessarily the same as minOccurs.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
stevedlawrence: Yep, it took some macro magic to figure out how to get it work, but this technique is definitely easily reusable to avoid allocations.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/processors/parsers/PointOfUncertaintyMacros.scala
stevedlawrence: Agreed. Will do.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
stevedlawrence: Will fix.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/InitiatedContentParsers.scala
stevedlawrence: If min is zero, i think there should be a point of uncertainty for the first first array element, since the entire array is optional. And there should also be a separate point of uncertainty for the enclosing choice. So this should resolve two points of uncertainty. At least that's how I understand this stuff. I'll add a test to make sure this does the right thing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/NilEmptyCombinatorParsers.scala
stevedlawrence: Yeah, I've always hated the way we handled these mark leaks, so much code to maintain and mutable variables. The macro makes this much cleaner and consistent among all uses.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
stevedlawrence: The quick description is: parsers push PoU's onto the top of the stack when one exists. So the the current in-scope PoU is always at the top of the stack. Discriminators/initated content/etc. resolve whatever is on the top of the stack. This means resolve a PoU changes the in-scope PoU, allowing multiple discriminators to resolve multiple PoUs down the stack. When parsers that created PoU's wind down, two cases are possible:
1. The parers PoU is at the top of the stack, which means nothing resolved it
2. Some previous PoU is at the top of the stack, which can only happen when a discriminator/initated content/etc. popped it off, meaning it was resolved. So this isPoUResolved function really just needs to see if the PoU the parser created is at the top of the stack.

I'll add some doc to all these things to make this clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/InitiatedContentParsers.scala
mbeckerle: Add a comment here also to save people taking it out, only to find it breaks the test, and putting it back in. Worth it to say there really are two PoUs here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -305,24 +346,86 @@ final class PState private (
...
+   * will munges with variable names in the "func" variable. Thus, this is
+   * potentially fragile, e.g. reflection on the pou variable name will fail.
+   * It is recommend to keep the func body as small as possible and avoid
+   * things like reflection that could cause breakage.
+   */
+  def withPointOfUncertainty[B](pouID: String, context: RuntimeData)(func: PState.Mark => B): B =
+    macro PointOfUncertaintyMacros.withPointOfUncertainty[PState.Mark, B]
+
+  /**
+   * This function creates a mark, which represents a point of uncertaint. This
```
tuxji, line 370: (Fix typo:) ... represents a point of **uncertainty**.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -305,24 +346,86 @@ final class PState private (
...
+    pointsOfUncertainty.push(pou)
+    pou
   }
 
-  def popPointOfUncertainty: Unit = {
-    // threadCheck()
-    discriminatorStack.pop
-    changedVariablesStack.pop
+  /**
+   * Reset to the point of uncertainty create with withPointOfUncertainty. This
```
tuxji, line 382: (Fix typo:) uncertainty **created** with

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -305,24 +346,86 @@ final class PState private (
...
+    Assert.invariant(pou == pouPop)
+    reset(pouPop)
   }
 
-  def setDiscriminator(disc: Boolean): Unit = {
-    // threadCheck()
-    discriminatorStack.pop()
-    discriminatorStack.push(disc)
+  /**
+   * Discard the point of uncertainty create with withPointOfUncertainty. Once
```
tuxji, line 395: (Fix typo:) uncertainty **created** with

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -305,24 +346,86 @@ final class PState private (
...
+   */
+  def resolvePointOfUncertainty(): Unit = {
+    if (!pointsOfUncertainty.isEmpty) {
+      val pouPop = pointsOfUncertainty.pop
+      discard(pouPop)
+    }
+  }
+
+  /**
+   * Determine if the PoU created in withPointOfUncertainty is been resolved by
```
tuxji, line 421: (Fix typo:) **has** been resolved

PR #407, opened 2020-08-14
Daffodil 2322 2210 bidi base

rsteinberger:
Addresses Daffodil-2322 and Daffodil-2210

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
@@ -76,7 +76,8 @@ final class Restriction(xmlArg: Node, val simpleTypeDef: SimpleTypeDefBase)
     val baseQNameString = baseQNameNodeSeq.text
     val tryBaseQName = QName.resolveRef(baseQNameString, xml.scope,
       tunable.unqualifiedPathStepPolicy)
-    Assert.invariant(tryBaseQName.isSuccess)
+    schemaDefinitionUnless(tryBaseQName.isSuccess,
+      "Failed to resolve base property reference for xs:restriction: " + tryBaseQName.failed.get.getMessage)
     tryBaseQName.get
   }
 
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
@@ -789,6 +789,8 @@
       type="dfdl:YesNoEnum" />
     <xsd:attribute form="qualified" name="textBidiNumeralShapes"
       type="dfdl:TextBidiNumeralShapesEnum" />
+    <xsd:attribute form="qualified" name="textBidi"
+                   type="dfdl:YesNoEnum" />
   </xsd:attributeGroup>
 
   <!-- 13.4 Properties Specific to Strings with Text representation -->
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testTextBidi.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testTextBidi.tdml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ex="http://example.com"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  suiteName="unparser tests involving bitOrder and outputValueCalc interactions"
+  defaultRoundTrip="onePass">
+
+  <!--
+  The DFDL v1.0 Specification states that:
+  Bidirectional text properties other than textBidi=”no” removed and reserved for future use.
+  -->
+
+  <tdml:defineSchema name="SchemaTextBidi" elementFormDefault="unqualified">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+    <xs:element name="root-no" type="xs:string" dfdl:textBidi="no" />
+    <xs:element name="root-yes" type="xs:string" dfdl:textBidi="yes" />
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="textBidiNo" model="SchemaTextBidi" root="root-no">
+    <tdml:document>test</tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:root-no>test</ex:root-no>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="textBidiYes" model="SchemaTextBidi" root="root-yes">
+    <tdml:document>test</tdml:document>
+    <tdml:errors>
+      <tdml:error>textBidi='yes' is not supported</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+</tdml:testSuite>
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestTextBidi.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestTextBidi.scala
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.section00.general
+
+import org.apache.daffodil.tdml.Runner
+import org.junit.{ AfterClass, Test }
+
+object TestTextBidi {
+  val testDir = "/org/apache/daffodil/section00/general/"
+  val runner = Runner(testDir, "testTextBidi.tdml")
+
+  @AfterClass def shutDown: Unit = {
+    runner.reset
+  }
+}
+
+/*
+ * This test addresses reported issue DAFFODIL-2322
+ * dfdl:textBidi property errors when used in attribute form
+ */
+class TestTextBidi {
+
+  import TestTextBidi._
+
+  @Test def test_textBidiNo(): Unit = { runner.runOneTest("textBidiNo") }
+  @Test def test_textBidiYes(): Unit = { runner.runOneTest("textBidiYes") }
+
+}
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes2.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes2.scala
@@ -99,4 +99,7 @@ class TestSimpleTypes2 {
 
   @Test def test_hexBinary_Delimited_04(): Unit = { runner.runOneTest("hexBinary_Delimited_04") }
 
+  @Test def test_restrictionBaseEmbeddedLeadlingSpace(): Unit = { runner.runOneTest("restrictionBaseEmbeddedLeadingSpace") }
+  @Test def test_restrictionBaseEmbeddedTrailingSpace(): Unit = { runner.runOneTest("restrictionBaseEmbeddedTrailingSpace") }
+
 }
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
@@ -4354,6 +4354,24 @@
     </tdml:errors>
   </tdml:unparserTestCase>
 
+  <tdml:defineSchema name="SimpleTypes-restriction-base" elementFormDefault="unqualified">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat"/>
+
+    <xs:element name="embeddedLeadingSpace">
+      <xs:simpleType>
+        <xs:restriction base=" xs:string" />
+      </xs:simpleType>
+    </xs:element>
+
+    <xs:element name="embeddedTrailingSpace">
+      <xs:simpleType>
+        <xs:restriction base="xs:string " />
+      </xs:simpleType>
+    </xs:element>
+
+  </tdml:defineSchema>
+
   <tdml:defineSchema name="SimpleTypes-padding">
     <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
     <dfdl:format ref="ex:GeneralFormat" representation="text"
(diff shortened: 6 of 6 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -73,11 +74,17 @@ final class Restriction(xmlArg: Node, val simpleTypeDef: SimpleTypeDefBase)
...
     val baseQNameNodeSeq = xml \ "@base"
-    val baseQNameString = baseQNameNodeSeq.text
-    val tryBaseQName = QName.resolveRef(baseQNameString, xml.scope,
+    val tryBaseQName = QName.resolveRef(baseQNameNodeSeq.text, xml.scope,
       tunable.unqualifiedPathStepPolicy)
-    Assert.invariant(tryBaseQName.isSuccess)
-    tryBaseQName.get
+    val result = tryBaseQName.isFailure && baseQNameNodeSeq.text.contains(" ") match {
+      case true => {
+        println("Warning: Invalid spaces in base name removed '" + baseQNameNodeSeq.text + "'")
```
olabusayoT, line 81: Also rather than a match case true/false, an if/else is preferable per Scala code style.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -73,11 +74,17 @@ final class Restriction(xmlArg: Node, val simpleTypeDef: SimpleTypeDefBase)
 
   lazy val baseQName = {
     val baseQNameNodeSeq = xml \ "@base"
-    val baseQNameString = baseQNameNodeSeq.text
-    val tryBaseQName = QName.resolveRef(baseQNameString, xml.scope,
+    val tryBaseQName = QName.resolveRef(baseQNameNodeSeq.text, xml.scope,
       tunable.unqualifiedPathStepPolicy)
-    Assert.invariant(tryBaseQName.isSuccess)
-    tryBaseQName.get
+    val result = tryBaseQName.isFailure && baseQNameNodeSeq.text.contains(" ") match {
```
olabusayoT, line 79: The assert.invariant ensured tryBaseQName was successful, however if tryBaseQName.isFailure but doesn't contain the " ", this still calls tryBaseQName rather than erroring out

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -47,6 +47,7 @@ final class Restriction(xmlArg: Node, val simpleTypeDef: SimpleTypeDefBase)
   with NestingLexicalMixin
   with TypeChecks {
 
+//  println(simpleTypeDef
```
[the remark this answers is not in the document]
rsteinberger: Do you want me to separate the 2, then we can chat about the QName issue?

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBase.tdml
```diff
@@ -0,0 +1,69 @@
...
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+    <xs:element name="foo">
+      <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:pattern value="&#xE000;*"/>
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:element>
+  </tdml:defineSchema>
```
olabusayoT, line 49: It'd be clearer to rename these schemas based on what they're testing. Ex: RestrictionBase.embedded.dfdl.xsd. The embedded.dfdl.xsd is helpful while debugging if you run into issues. I think the file should be changed to testRestrictionBase. testBase is abit ambiguous. I had to go look at the ticket to understand what was being tested.

The elements could also be named in a manner than describes the difference ex foo_noSpaceInAttribute and foo_spaceInAttribute. I had to stare hard to discover the difference

olabusayoT: I'd also like to see a test for leading space

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBase.tdml
```diff
@@ -0,0 +1,69 @@
...
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="base2" model="s2">
+    <tdml:document>foo</tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:foo>foo</ex:foo>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
```
olabusayoT, line 67: These test names should be more descriptive as well

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestBase.scala
```diff
@@ -0,0 +1,43 @@
...
+/*
+ * This test addresses reported issue DAFFODIL-2210
+ * Unhandled exception with leading/trailing space in restriction base
+ */
+class TestBase {
+
+  import TestBase._
+
+  @Test def test_base1(): Unit = { runner.runOneTest("base1") }
+  @Test def test_base2(): Unit = { runner.runOneTest("base2") }
```
olabusayoT, line 41: Same thing regarding more descriptive test names. They should be enough to describe the test without having to look through the parser test case

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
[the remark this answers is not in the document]
rsteinberger: This makes sense, I'll rework it in this direction.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBase.tdml
olabusayoT: I think you should use the description attribute here to descrive what is being tested

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBase.tdml
```diff
@@ -0,0 +1,69 @@
...
+  limitations under the License.
+-->
+
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ex="http://example.com"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  suiteName="unparser tests involving bitOrder and outputValueCalc interactions"
+  defaultRoundTrip="false">
```
olabusayoT, line 25: I think we should use onePass here. It'd have the effect of testing the unparse side as well, which can't hurt

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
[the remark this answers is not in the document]
stevedlawrence: Actually, it looks like we already have the logic for checking if a qname is valid--QName.resolveRef returns a [``Try``](https://www.scala-lang.org/api/current/scala/util/Try.html) that contains failure information. So the right fix here is rather than doing ``Assert.invariant(tryBaseQName.isSuccess)`` we should throw an SDE if the the ``Try`` failed. There are various ways to throw an SDE, but in this case you can probably do something like:
```scala
schemaDefinitionUnless(tryBaseQName.isSuccess, "some error message about invalid qname base")
```

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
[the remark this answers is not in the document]
stevedlawrence: The first is small enough that's it's not a big deal in this case. And there's lots of good comments in this PR, so let's keep it open. Just keep it mind for future PRs.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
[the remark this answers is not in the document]
rsteinberger: Sounds good. So don't try to remove the spaces, just throw the SDE.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBidi.tdml
```diff
@@ -0,0 +1,48 @@
...
+  <tdml:defineSchema name="s1" elementFormDefault="unqualified">
+
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+    <xs:element name="foo" type="xs:string" dfdl:textBidi="no" />
+
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="textBidi1" model="s1">
```
olabusayoT, line 36: This just tests the Bidi attribute, not the functionality, correct? If so I'd update the names to be more descriptive of the testing. Also as with previous, it'd be nice to have a description explaning what this is testing

tuxji: FYI.  I found out none of the textBidi* properties are implemented in Daffodil and they have been withdrawn from the DFDL 1.0 specification although Daffodil still accepts textBidi="no" for compatibility.  (See <https://issues.apache.org/jira/browse/DAFFODIL-645> and <https://issues.apache.org/jira/browse/DAFFODIL-1720>.)  I would suggest adding all the above as a comment in this textBidi.tdml file in case someone wonders why there are no tests for textBidi="yes" or any other textBidi* property, only a test for textBidi="no".

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
```diff
@@ -677,6 +677,8 @@
       type="dfdl:UTF16WidthEnum" />
     <xsd:attribute form="qualified" name="ignoreCase"
       type="dfdl:YesNoEnum" />
+    <xsd:attribute form="qualified" name="textBidi"
+      type="dfdl:YesNoEnum" />
```
tuxji, line 681: I looked for the other textBidi* properties and found they are defined further down in their own attributeGroup "TextBidiSchemeAGQualified" between lines 781-792.  If you moved this textBidi attribute further down the file to line 782 inside that group, would the test still work?  If so, please move it there to keep the textBidi* properties grouped together.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testBase.tdml
```diff
@@ -0,0 +1,69 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ex="http://example.com"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  suiteName="unparser tests involving bitOrder and outputValueCalc interactions"
```
tuxji, line 24: This suiteName is not appropriate for what these tests are doing.  However, I would go further than renaming this file; I would suggest moving the tests from section00/general to section05/simple_types and adding them to the simple-types-bases.tdml file there, which already seems to have some similar tests checking for errors involving <xs:restriction base="..."> simple types.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestBase.scala
```diff
@@ -0,0 +1,43 @@
...
+  @AfterClass def shutDown: Unit = {
+    runner.reset
+  }
+}
+
+/*
+ * This test addresses reported issue DAFFODIL-2210
+ * Unhandled exception with leading/trailing space in restriction base
+ */
+class TestBase {
```
tuxji, line 36: If you move the TDML tests to simple_type-bases.tdml, you also can move the Scala tests to TestSimpleTypes2 which already has a runner for simple-types-base.tdml.

rsteinberger: Thanks John, I believe I have responded to your comments. Following them very closely.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RestrictionUnion.scala
```diff
@@ -76,7 +76,7 @@ final class Restriction(xmlArg: Node, val simpleTypeDef: SimpleTypeDefBase)
     val baseQNameString = baseQNameNodeSeq.text
     val tryBaseQName = QName.resolveRef(baseQNameString, xml.scope,
       tunable.unqualifiedPathStepPolicy)
-    Assert.invariant(tryBaseQName.isSuccess)
+    schemaDefinitionUnless(tryBaseQName.isSuccess, "Invalid base string value.")
```
stevedlawrence, line 79: I'm not sure this error message is really clear what the problem is. Also, I think it's possible that resolveRef can fail for reasons other than being invalid, like it just couldn't find the reference. Maybe an error message that might be more clear is something like
```scala
"Failed to resolve base property reference for xs:restriction: " + tryBaseQname.failed.get.getMessage
```
That way it's more clear what "base" is, and it also includes the error message which specifically will say why it failed.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/DFDL_part2_attributes.xsd
```diff
@@ -789,6 +789,8 @@
       type="dfdl:YesNoEnum" />
     <xsd:attribute form="qualified" name="textBidiNumeralShapes"
       type="dfdl:TextBidiNumeralShapesEnum" />
+    <xsd:attribute form="qualified" name="textBidi"
+                   type="dfdl:YesNoEnum" />
```
stevedlawrence, line 793: This change, and other textbidi changes, are in the wrong commit. This is a textBidi change and so should be in the first commit. This is another downside with including two separate commits/changes in the same pull request--it's easy to accidentally put things in the wrong commit.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testTextBidi.tdml
```diff
@@ -0,0 +1,50 @@
...
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  suiteName="unparser tests involving bitOrder and outputValueCalc interactions"
+  defaultRoundTrip="false">
+
+  <!--
+  The DFDL v1.0 Specification states that:
+  Bidirectional text properties other than textBidi=”no” removed and reserved for future use.
+  As a result only the test for textBidi=no is currently implemented
+-->
```
stevedlawrence, line 31: It's probably also wroth adding a test for textBidi="yes" to make sure that creates an appropriate error.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestTextBidi.scala
```diff
@@ -1,42 +1,42 @@
...
+
+/*
+ * This test addresses reported issue DAFFODIL-2322
+ * dfdl:textBidi property errors when used in attribute form
+ */
+class TestTextBidi {
+
+  import TestTextBidi._
+
+  @Test def test_textBidi(): Unit = { runner.runOneTest("test_textBidiNo") }
```
stevedlawrence, line 40: The convention we use for naming tests is the name of the JUnit test should be thes ame as the name of the tdml test, but with a ``test_`` prefix. So this tdml test name should not have ``test_`` and the JUnit name needs to be ``test_testBidiNo`` (note the "No" at the end).

daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypes2.scala
```diff
@@ -99,4 +99,6 @@ class TestSimpleTypes2 {
 
   @Test def test_hexBinary_Delimited_04(): Unit = { runner.runOneTest("hexBinary_Delimited_04") }
 
+  @Test def test_restrictionBaseEmbeddedLeadlingSpace(): Unit = {runner.runOneTest("restrictionBaseEmbeddedLeadingSpace")}
+  @Test def test_restrictionBaseEmbeddedTrailingSpace(): Unit = {runner.runOneTest("restrictionBaseEmbeddedTrailingSpace")}
```
stevedlawrence, line 103: Add spaces inside the curly braces.

rsteinberger: Great comments, Steve. I believe I have addressed all of them.

stevedlawrence:
@tuxji 

Regarding formatting, [DAFFODIL-2133](https://issues.apache.org/jira/browse/DAFFODIL-2133) discusses one option for using scalariform to check to make sure code is formatted properly, and I think it can also make the changes. We could enable this check on GithubActions and cause a failure if it finds code that doesn't meet the guidelines. That way people submitting pull requests will see the issues and hopefully fix them before things get to review. It does mean we need to fix the entire code base to be consistent, which might be a big task.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testTextBidi.tdml
```diff
@@ -0,0 +1,60 @@
...
+  limitations under the License.
+-->
+
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ex="http://example.com"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  suiteName="unparser tests involving bitOrder and outputValueCalc interactions"
+  defaultRoundTrip="false">
```
olabusayoT, line 25: Can you make this onePass so it also tests the unparse side?

stevedlawrence: Good catch @olabusayoT. This is a valuable change to make to ensure all tests work for both parse and unparse.

Also since you're already making changes to this file, I'd also suggest combining the defineSchemas into one. In this case, it shoudn't be necessary to have two different schemas. Instead, just have a single defineSchema with two root elements (suggest changing their names from ``test`` and ``root`` to better differentiate) and then the tdml:parserTestCase's can reference the same model with a different root element.

rsteinberger: Good comments. Done.

PR #408, opened 2020-08-20
Implement SAX Parsing

olabusayoT:
Initial implementation of Daffodil SAX Parser and integration into our TDML runner.

Feedback appreciated, as I want to make sure I'm moving in the right direction.

I used multiple commits to break it up logically!

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -17,84 +17,89 @@
 
 package org.apache.daffodil
 
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.io.FileInputStream
 import java.io.FileOutputStream
 import java.io.OutputStream
-import java.io.FileInputStream
-import java.io.ByteArrayInputStream
+import java.net.URI
+import java.nio.ByteBuffer
 import java.nio.channels.Channels
 import java.nio.file.Paths
 import java.util.Scanner
-
-import scala.xml.{SAXParseException, Node}
-import org.rogach.scallop
-import org.apache.daffodil.debugger.{InteractiveDebugger, CLIDebuggerRunner, TraceDebuggerRunner}
-import org.apache.daffodil.util.Misc
-import org.apache.daffodil.util.Timer
-import org.apache.daffodil.xml._
-import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.compiler.Compiler
-import org.apache.daffodil.api.{WithDiagnostics, URISchemaSource, DFDL, DaffodilTunables}
-import org.apache.daffodil.util.Logging
-import org.apache.daffodil.util.LogLevel
-import org.apache.daffodil.util.LogWriter
-import org.apache.daffodil.util.LoggingDefaults
-import org.apache.daffodil.exceptions.NotYetImplementedException
-import java.io.File
-
-import org.apache.daffodil.tdml.DFDLTestSuite
-import org.apache.daffodil.externalvars.{Binding, BindingException}
-import org.apache.daffodil.externalvars.ExternalVariablesLoader
-import org.apache.daffodil.configuration.ConfigurationLoader
-import org.apache.daffodil.api.ValidationMode
-
-import scala.language.reflectiveCalls
-import scala.concurrent.Future
 import java.util.concurrent.Executors
 
-import scala.concurrent.ExecutionContext
-import org.rogach.scallop.ScallopOption
-
-import scala.concurrent.duration.Duration
 import scala.concurrent.Await
-import org.apache.daffodil.xml.QName
-import org.apache.daffodil.dsom.ExpressionCompilers
-import org.apache.daffodil.compiler.InvalidParserException
-import java.net.URI
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration.Duration
+import scala.language.reflectiveCalls
+import scala.xml.Node
+import scala.xml.SAXParseException
 
-import org.apache.daffodil.tdml.TDMLException
-import org.apache.daffodil.xml.RefQName
-import org.rogach.scallop.ArgType
-import org.rogach.scallop.ValueConverter
-import org.apache.daffodil.processors.DataProcessor
-import org.apache.daffodil.processors.DataLoc
-import org.apache.daffodil.processors.HasSetDebugger
-import org.apache.daffodil.exceptions.UnsuppressableException
-import org.apache.daffodil.infoset.XMLTextInfosetOutputter
-import org.apache.daffodil.infoset.NullInfosetOutputter
-import org.apache.daffodil.infoset.ScalaXMLInfosetOutputter
-import org.apache.daffodil.infoset.JsonInfosetOutputter
-import org.apache.daffodil.infoset.InfosetOutputter
-import org.apache.daffodil.infoset.JDOMInfosetOutputter
-import org.apache.daffodil.infoset.W3CDOMInfosetOutputter
-import org.apache.daffodil.infoset.XMLTextInfosetInputter
-import org.apache.daffodil.infoset.JsonInfosetInputter
-import org.apache.daffodil.infoset.ScalaXMLInfosetInputter
-import org.apache.daffodil.infoset.JDOMInfosetInputter
-import org.apache.daffodil.infoset.W3CDOMInfosetInputter
-import org.apache.daffodil.infoset.InfosetInputter
+import javax.xml.parsers.DocumentBuilderFactory
 import javax.xml.transform.TransformerFactory
 import javax.xml.transform.dom.DOMSource
 import javax.xml.transform.stream.StreamResult
-import javax.xml.parsers.DocumentBuilderFactory
 import org.apache.commons.io.IOUtils
-import org.apache.daffodil.io.InputSourceDataInputStream
-import org.apache.daffodil.tdml.TDMLTestNotCompatibleException
+import org.apache.daffodil.api.DFDL
+import org.apache.daffodil.api.DFDL.ParseResult
+import org.apache.daffodil.api.DaffodilTunables
+import org.apache.daffodil.api.URISchemaSource
+import org.apache.daffodil.api.ValidationMode
+import org.apache.daffodil.api.WithDiagnostics
+import org.apache.daffodil.compiler.Compiler
+import org.apache.daffodil.compiler.InvalidParserException
+import org.apache.daffodil.configuration.ConfigurationLoader
+import org.apache.daffodil.debugger.CLIDebuggerRunner
+import org.apache.daffodil.debugger.InteractiveDebugger
+import org.apache.daffodil.debugger.TraceDebuggerRunner
+import org.apache.daffodil.dsom.ExpressionCompilers
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.exceptions.NotYetImplementedException
+import org.apache.daffodil.exceptions.UnsuppressableException
+import org.apache.daffodil.externalvars.Binding
+import org.apache.daffodil.externalvars.BindingException
+import org.apache.daffodil.externalvars.ExternalVariablesLoader
+import org.apache.daffodil.infoset.DaffodilOutputContentHandler
+import org.apache.daffodil.infoset.InfosetInputter
+import org.apache.daffodil.infoset.InfosetOutputter
+import org.apache.daffodil.infoset.JDOMInfosetInputter
+import org.apache.daffodil.infoset.JDOMInfosetOutputter
+import org.apache.daffodil.infoset.JsonInfosetInputter
+import org.apache.daffodil.infoset.JsonInfosetOutputter
+import org.apache.daffodil.infoset.NullInfosetOutputter
+import org.apache.daffodil.infoset.ScalaXMLInfosetInputter
+import org.apache.daffodil.infoset.ScalaXMLInfosetOutputter
+import org.apache.daffodil.infoset.W3CDOMInfosetInputter
+import org.apache.daffodil.infoset.W3CDOMInfosetOutputter
+import org.apache.daffodil.infoset.XMLTextInfosetInputter
+import org.apache.daffodil.infoset.XMLTextInfosetOutputter
 import org.apache.daffodil.io.DataDumper
-import java.nio.ByteBuffer
-
 import org.apache.daffodil.io.FormatInfo
+import org.apache.daffodil.io.InputSourceDataInputStream
+import org.apache.daffodil.processors.DataLoc
+import org.apache.daffodil.processors.DataProcessor
+import org.apache.daffodil.processors.HasSetDebugger
 import org.apache.daffodil.schema.annotation.props.gen.BitOrder
+import org.apache.daffodil.tdml.DFDLTestSuite
+import org.apache.daffodil.tdml.TDMLException
+import org.apache.daffodil.tdml.TDMLTestNotCompatibleException
 import org.apache.daffodil.udf.UserDefinedFunctionFatalErrorException
+import org.apache.daffodil.util.LogLevel
+import org.apache.daffodil.util.LogWriter
+import org.apache.daffodil.util.Logging
+import org.apache.daffodil.util.LoggingDefaults
+import org.apache.daffodil.util.Misc
+import org.apache.daffodil.util.Timer
+import org.apache.daffodil.xml.QName
+import org.apache.daffodil.xml.RefQName
+import org.apache.daffodil.xml.DaffodilXMLLoader
+import org.apache.daffodil.xml.XMLUtils
+import org.rogach.scallop
+import org.rogach.scallop.ArgType
+import org.rogach.scallop.ScallopOption
+import org.rogach.scallop.ValueConverter
 
 class NullOutputStream extends OutputStream {
   override def close(): Unit = {
@@ -114,14 +119,14 @@ class NullOutputStream extends OutputStream {
   }
 }
 
-class CommandLineXMLLoaderErrorHandler() extends org.xml.sax.ErrorHandler with Logging {
+class CommandLineSAXErrorHandler() extends org.xml.sax.ErrorHandler with Logging {
 
   def warning(exception: SAXParseException) = {
-    log(LogLevel.Warning, "loading schema: " + exception.getMessage())
+    log(LogLevel.Warning, exception.getMessage())
   }
 
   def error(exception: SAXParseException) = {
-    log(LogLevel.Error, "loading schema: " + exception.getMessage())
+    log(LogLevel.Error, exception.getMessage())
     System.exit(1)
   }
 
@@ -337,7 +342,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
     val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "variables to be used when parsing. An optional namespace may be provided.")
     val tunables = props[String]('T', keyName = "tunable", valueName = "value", descr = "daffodil tunable to be used when parsing.")
     val config = opt[String](short = 'c', argName = "file", descr = "path to file containing configuration items.")
-    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to output. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', or 'null'.", default = Some("xml")).map { _.toLowerCase }
+    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to output. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'.", default = Some("xml")).map { _.toLowerCase }
     val stream = toggle(noshort = true, default = Some(false), descrYes = "when left over data exists, parse again with remaining data, separating infosets by a NUL character", descrNo = "stop after the first parse, even if left over data exists")
     val infile = trailArg[String](required = false, descr = "input file to parse. If not specified, or a value of -, reads from stdin.")
 
@@ -364,6 +369,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
       case (Some("json")) => Right(Unit)
       case (Some("jdom")) => Right(Unit)
       case (Some("w3cdom")) => Right(Unit)
+      case (Some("sax")) => Right(Unit)
       case (Some("null")) => Right(Unit)
       case (Some(t)) => Left("Unknown infoset type: " + t)
       case _ => Assert.impossible() // not possible due to default value
@@ -401,7 +407,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
     val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "variables to be used when processing. An optional namespace may be provided.")
     val tunables = props[String]('T', keyName = "tunable", valueName = "value", descr = "daffodil tunable to be used when processing.")
     val config = opt[String](short = 'c', argName = "file", descr = "path to file containing configuration items.")
-    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to parse/unparse. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', or 'null'.", default = Some("xml")).map { _.toLowerCase }
+    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to parse/unparse. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'.", default = Some("xml")).map { _.toLowerCase }
     val infile = trailArg[String](required = true, descr = "input file or directory containing files to process.")
 
     validateOpt(schema, parser, rootNS) {
@@ -417,6 +423,8 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
       case (Some("json"), _) => Right(Unit)
       case (Some("jdom"), _) => Right(Unit)
       case (Some("w3cdom"), _) => Right(Unit)
+      case (Some("sax"), Some(true)) => Left("SAX unparse is not yet implemented")
+      case (Some("sax"), _) => Right(Unit)
       case (Some("null"), Some(true)) => Left("infoset type null not valid with performance --unparse")
       case (Some("null"), _) => Right(Unit)
       case (Some(t), _) => Left("Unknown infoset type: " + t)
@@ -482,6 +490,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
       case (Some("json")) => Right(Unit)
       case (Some("jdom")) => Right(Unit)
       case (Some("w3cdom")) => Right(Unit)
+      case (Some("sax")) => Right(Unit)
       //case (Some("null")) => Right(Unit) // null is not valid for unparsing
       case (Some(t)) => Left("Unknown infoset type: " + t)
       case _ => Assert.impossible() // not possible due to default value
@@ -723,16 +732,23 @@ object Main extends Logging {
   val blobDir = Paths.get(System.getProperty("user.dir"), "daffodil-blobs")
   val blobSuffix = ".bin"
 
-  def getInfosetOutputter(infosetType: String, os: java.io.OutputStream): InfosetOutputter = {
+  def getInfosetOutputter(infosetType: String, os: java.io.OutputStream)
+  : Either[InfosetOutputter, DaffodilOutputContentHandler] = {
     val outputter = infosetType match {
-      case "xml" => new XMLTextInfosetOutputter(os, true)
-      case "scala-xml" => new ScalaXMLInfosetOutputter()
-      case "json" => new JsonInfosetOutputter(os, true)
-      case "jdom" => new JDOMInfosetOutputter()
-      case "w3cdom" => new W3CDOMInfosetOutputter()
-      case "null" => new NullInfosetOutputter()
+      case "xml" => Left(new XMLTextInfosetOutputter(os, pretty = true))
+      case "scala-xml" => Left(new ScalaXMLInfosetOutputter())
+      case "json" => Left(new JsonInfosetOutputter(os, pretty = true))
+      case "jdom" => Left(new JDOMInfosetOutputter())
+      case "w3cdom" => Left(new W3CDOMInfosetOutputter())
+      case "sax" => Right(new DaffodilOutputContentHandler(os, pretty = true))
+      case "null" => Left(new NullInfosetOutputter())
     }
-    outputter.setBlobAttributes(blobDir, null, blobSuffix)
+    if (outputter.isLeft) {
+      outputter.left.map(_.setBlobAttributes(blobDir, null, blobSuffix))
+    } else {
+      // do nothing here, we set the blobAttributes using the Blob* properties
+      // within the sax parse calls
+    }
     outputter
   }
 
@@ -840,7 +856,8 @@ object Main extends Logging {
               case Some("-") | None => System.out
               case Some(file) => new FileOutputStream(file)
             }
-            val outputter = getInfosetOutputter(parseOpts.infosetType.toOption.get, output)
+            val infosetType = parseOpts.infosetType.toOption.get
+            val eitherOutputterOrHandler = getInfosetOutputter(infosetType, output)
 
             var lastParseBitPosition = 0L
             var keepParsing = true
@@ -848,9 +865,17 @@ object Main extends Logging {
 
             while (keepParsing) {
 
-              outputter.reset() // reset in case we are streaming
-
-              val parseResult = Timer.getResult("parsing", processor.parse(inStream, outputter))
+              val parseResult = eitherOutputterOrHandler match {
+                case Right(saxContentHandler) =>
+                  // reset in case we are streaming
+                  saxContentHandler.reset()
+                  Timer.getResult("parsing",
+                    parseWithSAX(processor, inStream, saxContentHandler,
+                    new CommandLineSAXErrorHandler()))
+                case Left(outputter) =>
+                  outputter.reset() // reset in case we are streaming
+                  Timer.getResult("parsing", processor.parse(inStream, outputter))
+              }
               val finfo = parseResult.resultState.asInstanceOf[FormatInfo]
               val loc = parseResult.resultState.currentLocation.asInstanceOf[DataLoc]
               displayDiagnostics(parseResult)
@@ -859,19 +884,19 @@ object Main extends Logging {
                 keepParsing = false
                 error = true
               } else {
-                // only XMLTextInfosetOutputter and JsonInfosetOutputter write
-                // directly to the output stream. Other InfosetOutputters must manually
-                // get the result and write it to the stream
-                outputter match {
-                  case sxml: ScalaXMLInfosetOutputter => {
+                // only XMLTextInfosetOutputter, JsonInfosetOutputter and
+                // DaffodilOutputContentHandler write directly to the output stream. Other
+                // InfosetOutputters must manually get the result and write it to the stream below
+                eitherOutputterOrHandler match {
+                  case Left(sxml: ScalaXMLInfosetOutputter) => {
                     val writer = new java.io.OutputStreamWriter(output, "UTF-8")
                     scala.xml.XML.write(writer, sxml.getResult, "UTF-8", true, null)
                     writer.flush()
                   }
-                  case jdom: JDOMInfosetOutputter => {
+                  case Left(jdom: JDOMInfosetOutputter) => {
                     new org.jdom2.output.XMLOutputter().output(jdom.getResult, output)
                   }
-                  case w3cdom: W3CDOMInfosetOutputter => {
+                  case Left(w3cdom: W3CDOMInfosetOutputter) => {
                     val tf = TransformerFactory.newInstance()
                     val transformer = tf.newTransformer()
                     val result = new StreamResult(output)
@@ -1037,8 +1062,15 @@ object Main extends Logging {
                       })
                       case Right(data) => Timer.getTimeResult({
                         val input = InputSourceDataInputStream(data)
-                        val outputterForParse = getInfosetOutputter(infosetType, nullOutputStreamForParse)
-                        processor.parse(input, outputterForParse)
+                        val eitherOutputterOrHandlerForParse = getInfosetOutputter(infosetType, nullOutputStreamForParse)
+                        eitherOutputterOrHandlerForParse match {
+                          case Left(outputter) => processor.parse(input, outputter)
+                          case Right(saxContentHandler) =>
+                            val errorHandler = new CommandLineSAXErrorHandler()
+                            val parseResult = parseWithSAX(processor, input, saxContentHandler,
+                              errorHandler)
+                            parseResult
+                        }
                       })
                     }
 
@@ -1313,6 +1345,21 @@ object Main extends Logging {
     }
 
     ret
+  }
+
+  private def parseWithSAX(
+    processor: DFDL.DataProcessor,
+    data: InputSourceDataInputStream,
+    saxContentHandler: DaffodilOutputContentHandler,
+    errorHandler: CommandLineSAXErrorHandler): ParseResult = {
+    val saxXmlRdr = processor.newXMLReaderInstance
+    saxXmlRdr.setContentHandler(saxContentHandler)
+    saxXmlRdr.setErrorHandler(errorHandler)
+    saxXmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBDIRECTORY, blobDir)
+    saxXmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBSUFFIX, blobSuffix)
+    saxXmlRdr.parse(data)
+    val pr = saxXmlRdr.getProperty(XMLUtils.DAFFODIL_SAX_URN_PARSERESULT).asInstanceOf[ParseResult]
+    pr
   }
 
   def bugFound(e: Exception): Int = {
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/util/Validator.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/util/Validator.scala
@@ -17,12 +17,14 @@
 
 package org.apache.daffodil.util
 
-import javax.xml.transform.stream.StreamSource
-import javax.xml.XMLConstants
-import scala.xml.parsing.NoBindingFactoryAdapter
 import java.net.URI
-import org.apache.daffodil.xml.DFDLCatalogResolver
+
 import scala.collection.mutable
+import scala.xml.parsing.NoBindingFactoryAdapter
+
+import javax.xml.XMLConstants
+import javax.xml.transform.stream.StreamSource
+import org.apache.daffodil.xml.DFDLCatalogResolver
 import org.xml.sax.ErrorHandler
 
 /**
@@ -45,8 +47,14 @@ object Validator extends NoBindingFactoryAdapter {
     val validator = {
       val optCachedValidator = cache.get(schemaFileNames)
       optCachedValidator match {
-        case Some(validator) => {
-          validator.reset()
+        case Some(cachedValidator) => {
+          cachedValidator.reset()
+          // reset takes it back to the original state at the point we call
+          // newValidator. So we need to re-set the features, resolvers,
+          // and handlers
+          val resolver = DFDLCatalogResolver.get
+          val validator : javax.xml.validation.Validator =
+            initializeValidator(cachedValidator, errHandler, resolver)
           validator
         }
         case None => {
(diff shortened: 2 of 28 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+  override def startSimple(diSimple: DISimple): Boolean = {
+    val ns = if (diSimple.erd.thisElementsNamespace.isNoNamespace) "" else diSimple.erd.thisElementsNamespace.toString
+    val prefix = if (diSimple.erd.thisElementsNamespacePrefix == null) "" else diSimple.erd.thisElementsNamespacePrefix
+    val nsb = diSimple.namedQName
+    try {
+      if (prefix.nonEmpty && ns.nonEmpty) {
+        ch.startPrefixMapping(prefix, ns)
+      }
+
+      if (isNilled(diSimple)) { // TODO split this up into isNilled, hasValue or else (empty)
```
stevedlawrence, line 83: Agreed on the TODO, this could probably be split up to avoid duplication, e.g.
```scala
val attrs =
  if (isNilled(diSimple)) {
    // create new AttributesImpl and add xsi:nil
  } else {
    // create empty AttributesImp
  }

ch.starElement(..., attrs)

if (diSimple.hasValue) {
  // ch.characters stuff
}
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+        val attrs = new AttributesImpl()
+        attrs.addAttribute(XMLUtils.XSI_NAMESPACE, "nil", "xsi:nil", "", "true")
+        ch.startElement(ns, nsb.local, nsb.toQNameString, attrs)
+      } else if (diSimple.hasValue)  {
+        ch.startElement(ns, nsb.local, nsb.toQNameString, new AttributesImpl())
+        val text =
+          if (diSimple.erd.optPrimType.get.isInstanceOf[NodeInfo.String.Kind]) {
+            // TODO Should we use XMLInfosetOutputter remapped instead?
+            XMLUtils.remapXMLIllegalCharactersToPUA(diSimple.dataValueAsString)
```
stevedlawrence, line 92: Yeah, I think mixing in XMLInfosetOutputter and using remapped seems reasonable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+   * @return true on sucess, false if there was an error and Daffodil should stop all
+   *         future calls to the InfosetOutputter
+   */
+  override def endSimple(diSimple: DISimple): Boolean = {
+    val ns = diSimple.erd.thisElementsNamespace
+    val prefix = diSimple.erd.thisElementsNamespacePrefix
+    val nsb = diSimple.namedQName
+    try {
+      ch.endElement(ns, nsb.local, nsb.toQNameString)
+      // TODO only called endPrefix is there is a prefix
```
stevedlawrence, line 122: Yep, probably needs similar logic as mentioned above for starting prefix mapping.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+  override def startComplex(diComplex: DIComplex): Boolean = {
+    val ns = if (diComplex.erd.thisElementsNamespace.isNoNamespace) "" else diComplex.erd.thisElementsNamespace.toString
+    val prefix = if (diComplex.erd.thisElementsNamespacePrefix == null) "" else diComplex.erd.thisElementsNamespacePrefix
+    val nsb = diComplex.namedQName
+    try {
+      if (prefix.nonEmpty && ns.nonEmpty) {
+        ch.startPrefixMapping(prefix, ns)
+      }
+
+      // TODO how about empty elements
```
stevedlawrence, line 148: I would think empty complex would still have a start/endEement calls. I dont' think you need to do anything special for empty complexs. They are allowed by XML.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+      if (isNilled(diComplex)) {
+        // TODO how to set as nilled in SAX
```
stevedlawrence, line 150: I think you're doing this right. As with startSimple, it probably makes sense to split this into separate vals. Maybe even create some a new function to create the nilAttribute or empty AttributesImpl so that startSimple/startComplex can share the logic.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilContentHandler.scala
```diff
@@ -0,0 +1,57 @@
...
+class DaffodilContentHandler(dp: DataProcessor, output: OutputStream) extends ContentHandler {
+  private val inputter = new SAXInfosetInputter(new ByteArrayInputStream(output.asInstanceOf[ByteArrayOutputStream].toByteArray))
+  private var _locator: Locator = null
+
+  override def setDocumentLocator(locator: Locator): Unit = {
+    _locator = locator
+  }
+
+  override def startDocument(): Unit = {
+    output.write("<?xml version=\"1.0\" encoding=\"UTF-8\" ?>".getBytes)
```
stevedlawrence, line 34: I knwo you probably haven't put much though into this since this is all unparsing, but just note that this is all about unparsing. The OutputStream here is for daffodil to write unparsed data to, not to output the infoset to.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +721,90 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+    val sio = new SAXInfosetOutputter(_contentHandler)
+    dp.parse(isdis, sio)
+  }
+
+  def parse(arr: Array[Byte]): DFDL.ParseResult = {
+    val isdis = InputSourceDataInputStream(arr)
+    val sio = new SAXInfosetOutputter(_contentHandler)
+    dp.parse(isdis, sio)
+  }
+}
```
stevedlawrence, line 811: Instead of these other parse methods returning a ParseResult, they should probably return a Unit just the the standard XMLReader.parse method. With SAX, they way errors are determined is by setting an error handler. So probably what we need to do get the ParseResult and call appropriate functions in the errorHandler to send diagnostics via warn/error/fatalError. If they user doesn't set an errorHandler then errors a just silently ignored. We might also want to throw a SAXException or IOException as a more obvious sign that the parse failed. I'm not sure how SAX parsing is expected to handle errors. Maybe somethign to look into more.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -652,81 +653,11 @@ class DataProcessor private (
 
 }
 
-class ParseResult(dp: DataProcessor, override val resultState: PState)
```
stevedlawrence, line 652: Was this just moved? It's hard to tell if there are any changes. Reccommend moving this back to make the changes more obvious if there are any.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -874,31 +890,45 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
         actual.getBlobPaths.foreach { blobPath =>
           Files.delete(blobPath)
         }
+        actualSax.getBlobPaths.foreach { blobPath =>
+          Files.delete(blobPath)
+        }
 
         val isErr: Boolean =
-          if (actual.isProcessingError) true
+          if (actual.isProcessingError || actualSax.isProcessingError ) true
```
stevedlawrence, line 898: We want isError to be true only if both actual and actualSax errored. If one errored but the other didn't that means something went wrong.

I'm thinking all the checking logic should be split off into separate functions, then the of of this can be something like
```scala
checkParseResult(actual)
checkParseResult(actualSax)
```
That minimize duplicate code and hard to catch logic issues like this.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1129,23 +1186,32 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
         // was success. This won't get called on failure, which is fine--leave
         // blobs around for debugging
         firstParseResult.getBlobPaths.foreach { blobPath =>
           Files.delete(blobPath)
         }
+        firstParseResultSax.getBlobPaths.foreach { blobPath =>
+          Files.delete(blobPath)
+        }
       }
       case TwoPassRoundTrip => {
```
stevedlawrence, line 1216: It would be nice if we could reduce alot of this duplication into functions.

Speaking of duplication, was there a difference in amount of time needed to run all the tests by parsing everything twice, once for sax and one regular? If there is a noticable difference, we might want to put this behind a flag and only enable it on GitHub actions or before releases as a verification step.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetOutputter.scala
```diff
@@ -47,8 +49,10 @@ class TDMLInfosetOutputter() extends InfosetOutputter {
   private val w3cdomOut = new W3CDOMInfosetOutputter()
   private val jsonOut = new JsonInfosetOutputter(jsonStream, false)
   private val xmlOut = new XMLTextInfosetOutputter(xmlStream, false)
+  private val ch = new SAXHandler()
+  private val saxOut = new SAXInfosetOutputter(ch)
```
stevedlawrence, line 53: Not sure we need the SAXInfosetOutputter if we are also duplicating parse using our SAX API in the TDML runner. That shoudl sufficiently tested the infoset outputter.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -730,6 +722,14 @@ object Main extends Logging {
...
       case "jdom" => new JDOMInfosetOutputter()
       case "w3cdom" => new W3CDOMInfosetOutputter()
+      case "sax" => {
+        // we don't actually use this infoset outputter object we create below
+        // for anything other than associating the "sax" option with the
+        // SAXInfosetOutputter. The actually SAXInfosetOutputter is contained
+        // within the DaffodilXMLReader object (see newXMLReaderInstance) of
+        // Dataprocessor
+        new SAXInfosetOutputter(new SAXHandler())
+      }
```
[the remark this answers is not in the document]
olabusayoT: So it looks like I missed something on rebase, we actually do use this in the outputter match when we're outputting the result of the parse. 

                  case _: SAXInfosetOutputter => {
                    new org.jdom2.output.XMLOutputter().output(saxContentHandler.getDocument, output)
                  }

So I updated the above to specify a null content handler, so any attempts to use it will reult in a nullpointerexception but we can still use it for a match. Would that be sufficient?

        // we don't actually use this infoset outputter object we create below
        // for anything other than associating the "sax" option with the
        // SAXInfosetOutputter. So the contenthandler is set to null, and any 
        // attempts to use this will result in a NullPointerException. 
        // 
        // The actual SAXInfosetOutputter is contained
        // within the DaffodilXMLReader object (see newXMLReaderInstance) of
        // Dataprocessor
        new SAXInfosetOutputter(null)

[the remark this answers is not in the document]
stevedlawrence: I'd prefer to not create a SAXInfosetOutputter that's never actually used aside from a match case. Feels fragile since someone could accidentally use it thinking it has the output of the sax result.

Perhaps we should rethink this--relying on NPE for code correctness is probably a bad idea and I shouldn't have suggested it in the first place.

Really, we shouldn't call getInfosetOutputter at all if we aren't going to use the result. Perhaps we use something like an ``Either``, so we either have an ``InfosetOutputter``, or a ``ContentHandler``? So the code becomes something like:

```scala
val outputter: Either[InfosetOutputter, ContentHandler] =
  if (infosetType == "sax") Right(new SAXHandler())
  else Left(getInfosetOutputter(...))
```

Now later logic can use Left vs Right to determine whether to use the new SAX API or the old API, for example:
```scala
val parseResult = outputter match {
  case Left(io: InfosetOutputter) => // use io with the old API
  csae Right(ch: ContentHandler) => // use ch with the new sax API
}

 And it can also be used to match when outputting the infoset as well, something like:
```scala
output match {
  case Left(sxsml: ScalaXMLInfosetOutputtter) => ...
  case Left(jdom: JDOMInfosetOutputter) => ...
  case Right(saxHandler: SAXHandler) => ...
  case _ => // do nothing
}
```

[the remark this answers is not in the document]
olabusayoT: Oh cool, I've never used Either before. Will do. Thanks!

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -861,7 +869,7 @@ object Main extends Logging {
               } else {
                 // only XMLTextInfosetOutputter and JsonInfosetOutputter write
                 // directly to the output stream. Other InfosetOutputters must manually
-                // get the result and write it to the stream
+                // get the result and write it to the stream below
```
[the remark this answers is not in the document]
olabusayoT: [DAFFODIL-2389](https://issues.apache.org/jira/browse/DAFFODIL-2389) created

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * @return true on sucess, false if there was an error and Daffodil should stop all
+   *         future calls to the InfosetOutputter
+   */
+  override def startDocument(): Boolean = {
+    try {
+      ch.startDocument
+      true
+    } catch {
+      case _: SAXException => false
```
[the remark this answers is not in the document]
olabusayoT: [DAFFODIL-2390](https://issues.apache.org/jira/browse/DAFFODIL-2390) created

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -274,6 +285,74 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
...
+
+    if( !actual.isError && !errorHandler.isError) {
+      verifySameParseOutput(outputter, xri)
+    }
+    verifySameDiagnostics(actual, xri)
+
+    new DaffodilTDMLParseResult(actual, outputter)
+  }
+
+  def verifySameParseOutput(dpOutputter: TDMLInfosetOutputter, xri: DFDL.DaffodilXMLReader): Unit = {
```
stevedlawrence, line 321: Can this just accept the SAXHandler as the second param? I don't think it uses any other parts of the XMLReader.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -274,6 +285,74 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
...
+        throw TDMLException(
+          """SAX parse output (actual) does not match DataProcessor Parse output (expected)""" +
+            "\n" + e.getMessage, None)
+      }
+    }
+  }
+
+  private def verifySameDiagnostics(
+    actual: DFDL.ParseResult,
+    xri: DFDL.DaffodilXMLReader): Unit = {
```
stevedlawrence, line 340: Like above, can this just take the TDMLErrorHandler as a prameter?

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -802,7 +802,7 @@
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>newVariableInstances must all be distinct within the same scope</tdml:error>
-      <tdml:error>ex:myVar1</tdml:error>
+      <tdml:error>myVar1</tdml:error>
     </tdml:errors>
```
stevedlawrence, line 806: We should probably also add some tests for the Java and Scala API's to make sure that people using those API's can actually use this API. I suspect we'll need new wrapper classes for DaffodilXMLReader for japi and sapi like we have for DataProcessor.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
olabusayoT: 'sbt test' was only about 20 seconds longer when run with the 1763fb2d13eba1df5dacb44c560bd24010f35881D commit as the base. Will retest after the final rebase

daffodil-lib/src/main/scala/org/apache/daffodil/util/Validator.scala
```diff
@@ -85,5 +82,22 @@ object Validator extends NoBindingFactoryAdapter {
...
+    validator.setFeature("http://xml.org/sax/features/validation", true)
+
+    // If you enable the feature below, it seems to do no validation at all. Just passes.
+    //          validator.setFeature("http://apache.org/xml/features/validation/dynamic", true)
+
+    validator.setFeature("http://apache.org/xml/features/validation/schema", true)
+    validator.setFeature("http://apache.org/xml/features/validation/schema-full-checking", true)
+    validator.setErrorHandler(errHandler)
+    validator.setResourceResolver(resolver)
+    validator
```
mbeckerle, line 99: Is a validator a serializable object? Is it a stateful object? I am wondering if we can compute this once-only and save as a member of the SchemaSetRuntimeData, or if that isn't possible, compute it once-only at runtime and reuse it for multiple validations.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
```diff
@@ -183,27 +183,9 @@ object QName {
...
-            notNullPrefixes.foreach { p => Assert.invariant(p.length > 0) }
-            val pairs = notNullPrefixes.map { _.length } zip notNullPrefixes
-            val first = pairs.head // there has to be 1 at least. But there might be only 1
-            val minLengthPair = pairs.foldLeft(first) { case (x @ (xlen, _), y @ (ylen, _)) => if (xlen <= ylen) x else y }
-            val (n, shortest) = minLengthPair
-            Assert.invariant(n > 0)
-            Some(shortest)
-          }
-        }
+        val prefix = scope.getPrefix(targetNamespace)
```
mbeckerle, line 186: Why this change? The idea was to be deterministic rather than the prefix being just whatever scope.getPrefix happens to give us when there are multiple alternatives.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -841,21 +844,31 @@ Differences were (path, expected, actual):
...
+        val nsbACompare = nsbA.toString().trim.split(" ").sorted
+        val nsbBCompare = nsbB.toString().trim.split(" ").sorted
 
         if (labelA != labelB) {
           // different label
           List((zPath, labelA, labelB))
+        } else if (checkPrefixes && prefixA != prefixB) {
+          // different prefix
+          List((zPath, prefixA, prefixB))
+        } else if (checkNamespaces && !(nsbACompare sameElements nsbBCompare)) {
```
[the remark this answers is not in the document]
mbeckerle: Note: this was for IDE reasons, so that you can type the identifier, the period, and the IDE would then give you a choice of possible methods. I actually find the method-style syntax more understandable also, vs. infix operator style.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,266 @@
...
+    val nsb = diComplex.namedQName
+    try {
+      contentHandler.endElement(ns, nsb.local, nsb.toQNameString)
+
+      val nsbStart = diComplex.erd.minimizedScope
+      val nsbEnd = if (diComplex.isRoot) scala.xml.TopScope else diComplex.diParent.erd.minimizedScope
+      var n = nsbStart
+
+      // we store the the prefixes in the same order as we started them so we can
+      // reverse them and end them in the enclosing order
```
mbeckerle, line 219: Whyy do we want them in this specific order? Is this just to make the output more deterministic, or does it avoid outputting redundant prefix definitions? or...?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  private def createNilAttribute(): AttributesImpl = {
+    val attrs = new AttributesImpl()
+    attrs.addAttribute(XMLUtils.XSI_NAMESPACE, "nil", "xsi:nil", "", "true")
+    attrs
+  }
+
+  /**
+   * Called by Daffodil internals to signify the beginning of an array of elements.
```
mbeckerle, line 240: Is this the same scaladoc as on a base trait/class? If so is there an idiom for avoiding repetition of the comment in the code while still getting the generated HTML documentation to have the same documentation generated?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +748,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+
+  private var _propertyMap = mutable.Map[String, AnyRef]()
+  private val _featureMap = mutable.Map[String, Boolean](SAXNamespaceFeature -> false,
+    SAXNamespacePrefixFeature -> false)
+
+  override def getFeature(name: String): Boolean = {
+    name match {
+      case SAXNamespaceFeature | SAXNamespacePrefixFeature => _featureMap(name)
+      case _ => {
+        throw new SAXNotRecognizedException("Only namespace and namespace features are supported for SAX parsing")
```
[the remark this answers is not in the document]
mbeckerle: You could also display a list of the available features so as to show users the possible alternatives. As in 

"Feature unsupported: " + name + ". Supported features are: " + listOfFeatures.mkString(", ")

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetOutputter.scala
```diff
@@ -47,6 +48,7 @@ class TDMLInfosetOutputter() extends InfosetOutputter {
   private val w3cdomOut = new W3CDOMInfosetOutputter()
   private val jsonOut = new JsonInfosetOutputter(jsonStream, false)
   private val xmlOut = new XMLTextInfosetOutputter(xmlStream, false)
+  private val ch = new SAXHandler()
```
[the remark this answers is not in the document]
mbeckerle: ch is contentHandler

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -234,21 +242,24 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
...
 
-  override def parse(is: java.io.InputStream, lengthLimitInBits: Long): TDMLParseResult = {
-
-    val outputter = new TDMLInfosetOutputter()
-    outputter.setBlobAttributes(blobDir, blobPrefix, blobSuffix)
+  def parse(uri: java.net.URI, lengthLimitInBits: Long): TDMLParseResult = {
+    val url = uri.toURL
+    val dpInputStream = url.openStream()
+    val saxInputStream = url.openStream()
+    doParseWithBothApis(dpInputStream, saxInputStream, lengthLimitInBits)
```
mbeckerle, line 249: I realize for now we're trying to test everything against both SAX and prior API. But eventually we'll just obsolete the non-SAX API, or reimplement it in terms of the SAX one, so there won't be "both". So shouldn't whether we do just SAX or both SAX and prior be under control of a constant def someplace?

mbeckerle (approved):
+1 I have a few small comments, but approving as my comments add little beyond SL's change requests.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Validator.scala
stevedlawrence: There's a handful of bugs related to improving how we do full validation:

https://issues.apache.org/jira/browse/DAFFODIL-2386 - Full infoset validation should stream use SAX ValidationHandler
https://issues.apache.org/jira/browse/DAFFODIL-1749 - Store Xerces validator in DataProcessor so reloaded parsers can use full validation
https://issues.apache.org/jira/browse/DAFFODIL-1685 - Full validation should create and initialize the validator before parsing/unparsing begins

These issues can probably all be fixed as part of one change, but I'd reccomend it not being done as part of this. There's already alot going on in this PR.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
stevedlawrence: Doesn't seem unreasonable to have this configurable. We might also want to make it configurable so that when using the non-SAX API that we only generate one infoset. We currently generate like 5 infosets. For very large infosets, that could take up qutie a bit of memory.

Maybe we create a new bug to make this all configuratble some how and have some low-memory/overhead option?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Validator.scala
```diff
@@ -42,11 +44,17 @@ object Validator extends NoBindingFactoryAdapter {
...
-        case Some(validator) => {
-          validator.reset()
+        case Some(cachedValidator) => {
+          cachedValidator.reset()
+          // reset takes it back to the original state at the point we call
+          // newValidator. So we need to re-set the features, resolvers,
+          // and handlers
+          val resolver = DFDLCatalogResolver.get
+          val validator : javax.xml.validation.Validator =
+            initializeValidator(cachedValidator, errHandler, resolver)
```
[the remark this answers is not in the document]
olabusayoT: Yes, that's right. The cached validator, once reset, loses the features, error handlers and resource resolvers. I realized while running the parse a 2nd time for the sax parse...I'd noted the cached errorHandler was null even though it been set before the cache.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -18,53 +18,78 @@
...
+import org.xml.sax.XMLReader;
+
+
+object INoWarn4 {
+  ImplicitsSuppressUnusedImportWarning()
+}
+
+object EqualityNoWarn3 {
+  EqualitySuppressUnusedImportWarning()
+}
```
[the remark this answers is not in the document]
olabusayoT: It just sorted them. They used to be like the below before. I can revert if preferred

```scala
import scala.collection.immutable.Queue; object INoWarn4 { ImplicitsSuppressUnusedImportWarning() }
import org.apache.daffodil.equality._; object EqualityNoWarn3 { EqualitySuppressUnusedImportWarning() }
```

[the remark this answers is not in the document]
mbeckerle: These objects prevent Eclipse IDE (at least) from removing the import. 

That enables one to see and use the features in those imported files with IDE support. 

For example the equality one, it's an operator definition of the "=:=" and "=#=" operators which do typed equality. 

Without that object, the import is removed as unused unless the file already has uses of =:= or =#= in it. Once the import is removed, you can no longer see =:= as an available operator when typing an expression. So one tends to forget to use it, even though it provides type-safety improvements.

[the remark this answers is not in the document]
mbeckerle: Note that in some other files (not in this PR) I found that reorg of imports has converted the import of org.apache.daffodil.equality._ into import org.apache.daffodil.equality.EqualitySuppressUnusedImportWarning
which defeats the entire purpose of having the import and object definition there, which is to keep the implicit definitions available that are provided by the equality package.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -841,21 +844,31 @@ Differences were (path, expected, actual):
...
-        val Elem(_, labelB, attribsB, _, childrenB @ _*) = b
+        val Elem(prefixA, labelA, attribsA, nsbA, childrenA @ _*) = a
+        val Elem(prefixB, labelB, attribsB, nsbB, childrenB @ _*) = b
         val typeA: Option[String] = a.attribute(XSI_NAMESPACE.toString, "type").map(_.head.text)
         val typeB: Option[String] = b.attribute(XSI_NAMESPACE.toString, "type").map(_.head.text)
         val maybeType: Option[String] = Option(typeA.getOrElse(typeB.getOrElse(null)))
         val nilledA = a.attribute(XSI_NAMESPACE.toString, "nil")
         val nilledB = b.attribute(XSI_NAMESPACE.toString, "nil")
+        val nsbACompare = nsbA.toString().trim.split(" ").sorted
+        val nsbBCompare = nsbB.toString().trim.split(" ").sorted
```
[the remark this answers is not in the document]
olabusayoT: Since sameElements is order agnostic, we don't actually need to do the sort on the elements. So I removed the sort

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -334,6 +359,11 @@ class DataProcessor private (
 
   override def getDiagnostics = ssrd.diagnostics
 
+  override def newXMLReaderInstance: DFDL.DaffodilXMLReader = new DaffodilXMLReader(this)
+
+  def newContentHandlerInstance(output: OutputStream): DaffodilContentHandler =
+    new DaffodilContentHandler(this, output)
+
```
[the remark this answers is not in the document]
olabusayoT: I'll untrack the files for now!

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +748,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+  }
+
+  override def fatalError(exception: SAXParseException): Unit = {
+    error(exception)
+  }
+
+  override def getDiagnostics: Seq[Diagnostic] = _diagnostics
+
+  override def isError: Boolean = _isError
+}
```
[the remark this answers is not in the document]
olabusayoT: Moving it to the TDMLRunner means we won't be able to use it for CLI parse. Currently we use this as our errorHandler. Would we need to find a separate error handler?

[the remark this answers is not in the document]
stevedlawrence: I would consider the CLI and TDML runner as having didfferent behaviors. The TDML runner needs to gather diagnostics and then compare with expected diagnostics. But the CLI can just ``log()`` the diagnostics out as soon as it gets them and then forget about them, no need to gather them up. So it doesn't seem unreasonable to have a different error handler to me.

In fact, we actually already have a ``CommandLineXMLLoaderErrorHandler`` in Main.scala that at one point we used for loading DFDL Schemas. That code actually isn't used anymore, but maybe we can just tweak that and reuse it?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
olabusayoT: So we found that it was this implemention was at odds with getPrefix. namedQName.prefix and thisElementNamespacePrefix were returning separate things. For the particular example I was looking at, I *think* the former was returning ex and the latter tns. @stevedlawrence  and I discussed it and we found it was more accurate to return the getPrefix implementation which is based on the order of the namespaces are they defined in the schema (which ends up being null, in the cases we have xmlns="...")

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
olabusayoT: So the initial intention was to ensure the order we started the mapping of the prefix matched the order we ended them, but as Steve mentioned in another comment, this doesn't matter as per the specs:

```Note that start/endPrefixMapping events are not guaranteed to be properly nested relative to each other: all startPrefixMapping events will occur immediately before the corresponding startElement event, and all endPrefixMapping events will occur immediately after the corresponding endElement event, but their order is not otherwise guaranteed.```

File: daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala [hunk 31]
```diff
@@ -234,21 +242,24 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
...
+    val saxInputStream = new ByteArrayInputStream(arr)
+    doParseWithBothApis(dpInputStream, saxInputStream, lengthLimitInBits)
+  }
 
-    new DaffodilTDMLParseResult(actual, outputter)
+  override def parse(is: java.io.InputStream, lengthLimitInBits: Long): TDMLParseResult = {
+    val arr = IOUtils.toByteArray(is)
+    val dpInputStream = new ByteArrayInputStream(arr)
+    val saxInputStream = new ByteArrayInputStream(arr)
+    doParseWithBothApis(dpInputStream, saxInputStream, lengthLimitInBits)
```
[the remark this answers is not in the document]
olabusayoT: We don't actually use it. I had it in there to provide support for multiple parameters for the TDML parse. Similar to uri option.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
olabusayoT: Created [DAFFODIL-2392](https://issues.apache.org/jira/browse/DAFFODIL-2392) for this

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
mbeckerle: So this returns the first defined prefix for the namespace? If so that's a useful thing that people can depend on which is to define the prefix they "like" as the canonical one first, and extra prefixes like tns that are really just for local schema-author convenience later in the list.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
mbeckerle: That spec allows for non-determinstic behavior, even though it's intent is just not to over constrain implementations. 

I'm happy with any order, so long as it is deterministic - on the same data repeating a parse should result in the same order.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/QNameBase.scala
olabusayoT: Correct!

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
olabusayoT: Hmm, we use minimizedScope for it. But I don't think it's necessarily deterministic, would it be preferable to sort before we call the start/stopPrefix

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetOutputter.scala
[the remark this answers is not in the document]
olabusayoT: It wasn't used and could be removed!

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
stevedlawrence: I think it's deterministic. I think the randomness you were seeing that we talked about was because JDOM, Scala etc when converted to text output namespaces prefixes in different orders. I'm pretty sure the SAX InfosetOutputer will always create events in the same order. They just might not necessarily match how JDOM/whatever ContentHandler stores it. But that's outside the scope of Daffodil. As long was we calling things in the same deterministic order, we should be fine.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -0,0 +1,303 @@
...
+    writer.write(qName)
+    // handle attributes
+    for(i <- 0 until atts.getLength) {
+      val attsValue = atts.getValue(i)
+      val attsQName = atts.getQName(i)
+      writer.write(s""" $attsQName="$attsValue"""")
+    }
+    // handle namespaces
+    if(prefixMapping != null) {
+      val pm = prefixMapping.toString()
```
stevedlawrence, line 261: Add space after ``if` and ``for`` above. Double check the if statements, there's a couple more in this file I think.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+  private var contentHandler: ContentHandler = _
+  private var errorHandler: ErrorHandler = _
+  private var dtdHandler: DTDHandler = _
+  private var entityResolver: EntityResolver = _
+  private val saxNamespaceFeature = "http://xml.org/sax/features/namespaces"
+  private val saxNamespacePrefixFeature = "http://xml.org/sax/features/namespace-prefixes"
+  private val saxSupportedProperties = Seq(s"(${XMLUtils.DAFFODIL_SAX_URN_ROOT}:ParseResult",
+    s"${XMLUtils.DAFFODIL_SAX_URN_ROOT}:BlobDirectory",
+    s"${XMLUtils.DAFFODIL_SAX_URN_ROOT}:BlobPrefix",
+    s"${XMLUtils.DAFFODIL_SAX_URN_ROOT}:BlobSuffix")
```
stevedlawrence, line 755: All these strings are probably worth defining in XMLUtils instead of here.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+    } else {
+      throw new SAXNotRecognizedException("Feature unsupported: " + name + ".\n" +
+        "Supported features are: " + featureMap.keys.mkString(", "))
+    }
+  }
+
+  override def getProperty(name: String): AnyRef = {
+    val expectedUniqueURNRegex = s"(${XMLUtils.DAFFODIL_SAX_URN_ROOT}:)(\\w+)".r
+    val prop = name match {
+      case expectedUniqueURNRegex(_, pr) if pr == "ParseResult" => saxParseResultProperty
```
stevedlawrence, line 786: I think it would be more clear to just do string matching. E.g.
```scala
name match {
 case XMLUtils.DAFFODIL_SAX_PROPERTY_PASE_RESULT => saxParseResultProperty
...
}
```
The regex just adds extra complexity. Same with setProperty.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+          .asInstanceOf[String]
+        case _ =>
+          throw new SAXNotRecognizedException("Property unsupported: " + name + "." +
+            "Supported properties are: " + saxSupportedProperties.mkString(", "))
+      }
+    } catch {
+      case e: ClassCastException =>
+        throw new SAXNotSupportedException("Unsupported value for property: " + name + "." +
+        "Supported properties/values types are: " + saxSupportedProperties.map{
+          p => s"$p: ${p.getClass}"}.mkString(", "))
```
stevedlawrence, line 817: I'm wondering if this is useful? This is just going to be a bunch of long URI. Maybe we just assume people can look at documentation to figure out properties? Then we don't have to update that saxSupportedProperties list each time we adda  new property?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+    }
+    prop
+  }
+
+  override def setProperty(name: String, value: AnyRef): Unit = {
+    val expectedUniqueURNRegex = s"(${XMLUtils.DAFFODIL_SAX_URN_ROOT}:)(\\w+)".r
+    try {
+      name match {
+        case expectedUniqueURNRegex(_, pr) if pr == "ParseResult" => saxParseResultProperty = value
+          .asInstanceOf[ParseResult]
```
stevedlawrence, line 802: Does it make sense to let callers set the ParseResult property? That seems like something onyl Daffodil internal should set and callers can get

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+   *
+   * @return SAXInfosetOutputter object with or without blob Attributes set
+   */
+  private def createSAXInfosetOutputter(xmlReader: DaffodilXMLReader): SAXInfosetOutputter = {
+    val sioo = new SAXInfosetOutputter(xmlReader)
+    val siof = try {
+      val blobDir = try {
+        getProperty(s"${XMLUtils.DAFFODIL_SAX_URN_ROOT}:BlobDirectory").asInstanceOf[java.nio
+        .file.Path]
+      } catch {
```
stevedlawrence, line 916: Setting the propety should have done the appropriate checks. So I don't think this should need any try/catch stuff. Just needs to use the saxProperty.. variables that youve set. So I think this can all just become

```scala
sioo.setBlobAttributes(saxPropertyBlobDir, saxPropertyBlobPrefix, saxPropetyBlobSuffix)
```
The default values and checking in setProperty always ensure we have a correct value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -718,3 +741,196 @@ class UnparseResult(dp: DataProcessor, ustate: UState)
...
+
+  override def parse(systemId: String): Unit = {
+    throw new IOException("SAX parsing of systemId is unsupported")
+  }
+
+  def parse(isdis: InputSourceDataInputStream): Unit = {
+    val sio = createSAXInfosetOutputter(this)
+    val pr = dp.parse(isdis, sio)
+    handleDiagnostics(pr)
+    setProperty(s"${XMLUtils.DAFFODIL_SAX_URN_ROOT}:ParseResult", pr)
```
stevedlawrence, line 863: I'd just set the saxPropetyParserResult variable directly. No need to go through setProperty, and that way users don't have a way to change the ParseResult property. This property should be read only.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/processor/TDMLDFDLProcessor.scala
```diff
@@ -141,3 +144,12 @@ trait TDMLUnparseResult extends TDMLResult {
...
 }
+
+trait TDMLErrorHandler extends ErrorHandler with WithDiagnostics {
+  override def warning(exception: SAXParseException): Unit
+  override def error(exception: SAXParseException): Unit
+  override def fatalError(exception: SAXParseException): Unit
+
+  override def getDiagnostics: Seq[Diagnostic]
+  override def isError: Boolean
+}
```
stevedlawrence, line 155: I think the TDML Error handler is completely internal to the DaffodilTDMLDFDLProcessor, right? Note sure you need this trait. Your DaffodilSAXErrorHandler can just extend ErrorHandler without this trait.

stevedlawrence (approved):
Looks nice! Just minor updates/cleanup suggestions

+1

PR #411, opened 2020-08-28
Added SDWs for discouraged assert and discriminator placements

rsteinberger:
Modified: SequenceGroup adding checkDiscrimAndAssertPlacement
to check for miplaced assert and discriminators
Modified: discriminator.tdml
adding discrimPlacementSDW and assertPlacementSDW
Modified: TestDiscriminators.scala to add runners
Modified: daffodil-propgen dafext.xsd adding new warnings

Daffodil-2277

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
@@ -26,6 +26,7 @@ import org.apache.daffodil.xml.XMLUtils
 import org.apache.daffodil.schema.annotation.props.gen.OccursCountKind
 import org.apache.daffodil.schema.annotation.props.gen.SequenceKind
 import org.apache.daffodil.Implicits.ns2String
+import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.dsom.walker.SequenceView
 import org.apache.daffodil.grammar.SequenceGrammarMixin
 import org.apache.daffodil.schema.annotation.props.Found
@@ -40,6 +41,7 @@ import org.apache.daffodil.processors.SeparatorUnparseEv
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.runtime1.ChoiceBranchImpliedSequenceRuntime1Mixin
 import org.apache.daffodil.schema.annotation.props.FindPropertyMixin
+
 
 /**
  * Base for anything sequence-like.
@@ -91,6 +93,8 @@ abstract class SequenceGroupTermBase(
 
 
   requiredEvaluationsIfActivated(checkIfValidUnorderedSequence)
+
+  requiredEvaluationsIfActivated(checkIfNonEmptyAndDiscrimsOrAsserts)
 
   protected def apparentXMLChildren: Seq[Node]
 
@@ -146,6 +150,18 @@ abstract class SequenceGroupTermBase(
       res
     }
   }.value
+
+  /**
+   * Provides validation for assert and discriminator placement
+   */
+  protected final lazy val checkIfNonEmptyAndDiscrimsOrAsserts: Unit = {
+    val msg = "Counterintuitive placement detected. Wrap the discriminator or assert " +
+      "in an empty sequence to evaluate before the contents."
+    if (groupMembers.size > 0 && discriminatorStatements.size > 0)
+      SDW(WarnID.DiscouragedDiscriminatorPlacement, msg)
+    if (groupMembers.size > 0 && assertStatements.size > 0)
+      SDW(WarnID.DiscouragedAssertPlacement, msg)
+  }
 
   /**
    * Provides unordered sequence checks.  Will SDE if invalid.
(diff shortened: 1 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -147,6 +151,20 @@ abstract class SequenceGroupTermBase(
...
+      checkDiscrimAndAssertPlacement
+  }
+
+  private lazy val checkDiscrimAndAssertPlacement: Unit = {
+    if (groupMembers.size > 0 && discriminatorStatements.size > 0)
+      SDW(WarnID.DiscouragedDiscriminatorPlacement, "The recommended placement is after the target element.")
+    if (groupMembers.size > 0 && assertStatements.size > 0)
+      SDW(WarnID.DiscouragedAssertPlacement, "The recommended placement is after the target element.")
+  }
+
```
stevedlawrence, line 167: I think it's safe to combine these lazy vals. I'm not sure we gain anything by having two separate variables.

I'm also not sure the message makes it clear how to fix this. "After the target element" isn't obvious to me. I'm not sure how to say this with breifly, but maybe something along the lines of

> Discriminators placed directly on sequences has unintutive behavior and will evaluate after the contents of the sequence. Wrap the discriminator in an empty sequence to evaluate before the contents.

That feels horribly long and verbose, but maybe that can be reworded?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -476,6 +476,8 @@
           <xs:enumeration value="unsupportedAttributeFinalDefault" />
           <xs:enumeration value="unsupportedAttributeFormDefault" />
           <xs:enumeration value="unsupportedAttributeSchemaLocation" />
+          <xs:enumeration value="discouragedDiscriminatorPlacement" />
+          <xs:enumeration value="discouragedAssertPlacement" />
```
stevedlawrence, line 480: Can you sort these so they are at the top of the list? Makes it easier to quickly scan for warnings.

rsteinberger: All addressed.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
```diff
@@ -147,6 +151,18 @@ abstract class SequenceGroupTermBase(
     }
   }.value
 
+  /**
+   * Provides validation for assert and discriminator placement
+   */
+  protected final lazy val checkIfNonEmptyAndDiscrimsOrAsserts: Unit = {
+    val msg = "Counterintuitive placement detected. Wrap the discriminator or assert " +
+      "in an empty sequence to evaluate before the contents.\n"
```
tuxji, line 159: How does the \n help readability?  Just curious (I grep'ed the source for "SDW(WarnID" and only 3 out of 25 SDW's had \n in their messages).

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
```diff
@@ -28,6 +28,94 @@
...
+      <tdml:dfdlInfoset>
+        <root-assert>
+          <description>Hat</description>
+          <quantity>2</quantity>
+        </root-assert>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:warnings>
+      <tdml:warning>Schema Definition Warning</tdml:warning>
+      <tdml:warning>Wrap the discriminator or assert in an empty sequence</tdml:warning>
```
tuxji, line 115: No change needed since you're already checking for fewer words this time, but it's a good idea to put only a few key words you want to match against in the tdml:warning elements.  Someone rewords a SDW message and then a test breaks if it tries to match the entire SDW message word for word.  If the test checks only (for example) "Schema Definition Warning", "placement", "assert", "sequence", it'll be more robust to future changes.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
stevedlawrence: Yeah, in general we shouldn't have newlines in errors. The code the prints out diagnostics is responsible for that. There are some cases where a parse error includes multiple lines of information so newlines can help make that a little more easier to read. But usually it should be avoided. In this case, it isn't needed.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
mbeckerle: I'm good with that phrasing. It's only 2 lines of text.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SequenceGroup.scala
stevedlawrence: Looks like you still need to remove the \n.

PR #412, opened 2020-08-28
Removed ExecutionMode.scala and all references

rsteinberger:
This process has been documented as no longer being necessary.

Daffodil-1814


daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
```diff
@@ -1,485 +1,482 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
```
stevedlawrence, line 3: This entire file shows as a diff. Did you change the line endings perhaps? We try to always use unix lined endings. Make sure you IDE isn't changing things. It's not clear what changes were actually made to this and other files.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/SDE.scala
```diff
@@ -1,224 +1,221 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
```
stevedlawrence, line 3: Looks like this file had line endings messed up too.

rsteinberger: Yes, I corrected and pushed all of them.

tuxji: I think you missed SDE.scala, I still see a huge diff.

rsteinberger:
Right, so I can see that master contains SDE.scala with windows line endings. That seems to be the root of the problem. So I'm starting to wonder if we need a commit to fix that first? Or do we push this? Sorry its such a pain, but I've been elected to be the windows (always small w) guy.

stevedlawrence:
SDE.scala in master has LF line endings for me, and git on linux doesn't do any munging of line endings linke winows can, which confuses things. And I've checked out your PR branch, and then SDE.scala has CRLF line endings. So it definitely seems to me that this PR is changing from LF to CRLF, which we don't want.

Ususally git on windows has the ``core.autocrlf`` configuration option set to ``true``. This makes it so that when you checkout a branch, git wil convet the line endings to CRLF (which probably explains why it looks like a CRLF to you). And then when you make a commit, git will change them back to LF. So this all happens transparently to the user. But for some reason, it seems like git isn't converting the the CRLF back to a LF for this file. I have no idea why it wouldn't do that. All your other PR's have been fine. Maybe something is weird about this one file?

Note that it's also not a bad thing that you're using windows. It's great to have a diversity of build environments so tha we are more likely to catch environment specific issues. And it'll be good if we can figure out this issue so that we know how to fix if if more windows developers join the project and hit this problem.

rsteinberger:
Very interesting. I need to go deeper into this. I down loaded the zip of the repo to by pass git, and see that SDE.scala does not contain 0D. So git is having fun with me. Weird that it’s just this file. Stay tuned. And thanks.

tuxji:
Rick, I use a Windows 10 (winver 1909) laptop too.  I made git checkouts using Cygwin git for a long time using the following snippet inside my `~/.gitconfig`:

```ini
[core]
        safecrlf = true
        autocrlf = input
```

However, I now develop in Windows Subsystem for Linux (WSL) almost exclusively since the WSL experience (including Microsoft Terminal) has improved to the point where I feel like I'm using both Ubuntu 20.04 (inside Microsoft Terminal except for Emacs and IntelliJ IDEA which are running in their own windows using the VcXsrv X server) and Windows 10 at the same time.  The only concession I've made for Windows in my Ubuntu 20.04 environment's `~/.gitconfig` is to tell git to ignore files' executable bit in case a Windows program has created a file in my git checkout or my checkout is in a Windows filesystem instead of an Ubuntu filesystem:

```ini
[core]
        filemode = false
```

You may want to try WSL too which will allow you to mix and match your favorite development tools as you wish.  For example, I also run Visual Studio Code natively on Windows with a plugin that performs remote compilations and debugging inside WSL.

rsteinberger:
Should all be correct now. Please let me know if you still see issues. It turned out to be a combination of config setting in git vs Intellij. I now have get configured with git config --global core.autocrlf false to do no conversion. And Intellij configured to use only LF, at the project level. That should do it. But, I will certainly keep an eye on it. All documented in the New Developer doc.

PR #413, opened 2020-09-02
Various cleanups to sbt rpm configuration

stevedlawrence:
- Update sbt native-packager version. This fixes the issue where
  rpmbuild output showed up as an error in sbt even though there was no
  error. It now shows up as normal "info"
- Disable compression when building the rpm by injecting a define into
  the spec file using an sbt native-packager hack
- Change the shabang in the daffodil shell script to /usr/bin/sh. Part
  of rpmbuild mangle shebang lines so they are /usr/bin/sh instead of
  /bin/sh. This mangling makes it look like there is a difference
  between our zip binary helper and rpm content. To prevent this, just
  always use what RPM prefers.

DAFFODIL-2388


stevedlawrence:
Looks like the linux integration tests failed because ubuntu does not have sh in ``/usr/bin/sh``, but only in ``/bin/sh``. Perhaps we should be using ``#!/usr/bin/env sh`` instead.

daffodil-cli/src/templates/bash-template
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
tuxji, line 1: Bummer that you have to replace `#!/bin/sh` with `#!/usr/bin/env sh` to avoid rpmbuild changing your shebang (<https://unix.stackexchange.com/questions/588244/rpm-build-modifying-sh-files>).  Life is easier if `#!/bin/sh` just keeps working universally everywhere, even though you're restricted to plain POSIX shell syntax in that case (<https://unix.stackexchange.com/questions/496505/is-it-correct-to-use-bin-sh-in-the-hashbang-if-the-bourne-shell-isnt-available>).

However, I notice this file has the name "bash-template".  You might want to change the name to "sh-template" or change the shebang to `#!/usr/bin/env bash` so that the file's name matches its shebang.  I also ran `shellcheck bash-template` and got a few warnings about corner cases which you may want to fix too.

```text
In bash-template line 42:
    cd "$(dirname "$LNK")"
    ^--------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
    cd "$(dirname "$LNK")" || exit


In bash-template line 47:
  cd "$THISPWD"
  ^-----------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
  cd "$THISPWD" || exit


In bash-template line 53:
BINDIR=`dirname ${SCRIPT}`		# Directory script is run in
       ^-----------------^ SC2006: Use $(...) notation instead of legacy backticked `...`.
                ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
BINDIR=$(dirname "${SCRIPT}")		# Directory script is run in


In bash-template line 77:
exec java $JOPTS -cp "$CLASSPATH" "$MAINCLASS" "$@"
          ^----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
exec java "$JOPTS" -cp "$CLASSPATH" "$MAINCLASS" "$@"

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

stevedlawrence: Actually, it turns out RPM mangles the shebang line regardless. The only time it won't mangle it is if it's ``#!/usr/bin/sh``, which doesn't exist on ubuntu. Perhaps reverting back to ``#!/bin/sh`` is a better option than ``#!/usr/bin/env sh``.

The name ``bash-template`` is actually requirement from sbt native-packager. That's just the file it expects for unix-like systems. For example, it uses bat-template for windows systems.

Suggested shell changes seem reasonable.

PR #416, opened 2020-09-11
Throw SDE in the case of a choice with no branches

rsteinberger:
Section 15 of the spec satates that a choice that declares no branches
in the DFDL schema is a schema definition error.

Daffodil-1032


daffodil-core/src/main/scala/org/apache/daffodil/dsom/ChoiceGroup.scala
```diff
@@ -201,6 +202,12 @@ abstract class ChoiceTermBase(
     }
   }.value
 
+  final lazy val noBranchesFound = LV('noBranchesFound) {
+    if(groupMembers.size.equals(0)) {
```
stevedlawrence, line 206: The standard in scala is to use ``==`` instead of ``.equals()``. Also note that there should be a space between ``if`` and the open paren.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
```diff
@@ -123,9 +124,6 @@ class TestChoice {
   @Test def test_direct_dispatch_16(): Unit = { runnerCH.runOneTest("direct_dispatch_16") }
   @Test def test_direct_dispatch_17(): Unit = { runnerCH.runOneTest("direct_dispatch_17") }
 
-  //@Test def test_choice_noBranch() { runnerCH.runOneTest("choice_noBranch") } - Test consumes no data, which causes a TDMLError
-
-
```
stevedlawrence, line 128: Why did you delete the choice_noBranch test and replace it with a new one? Was something wrong with choice_noBranch? That name feels more appropriate than optionalChoice00, since this isn't testing optional choices.

rsteinberger: I didn't see the point of using the external xsd. I can changes the name if you like.

rsteinberger: I renamed it and move it to the basic model.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ChoiceGroup.scala
rsteinberger: Done

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
stevedlawrence: Gotcha. I suspect this was originally split out beause it's not a valid schema and in some cases that can cause issues. But since it doesn't, seems reasonable to put it along side the other schemas. This change is fine with me.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
```diff
@@ -38,6 +38,9 @@ class TestChoice {
 
   import TestChoice._
 
+  @Test def test_choice_noBranch (): Unit = { runnerCH.runOneTest("choice_noBranch") }
+
+  @Test def test_optionalChoice00(): Unit = { runnerCH.runOneTest("optionalChoice00") }
```
tuxji, line 43: Rick, you need to delete test_optionalChoice00 since the "optionalChoice00" test doesn't exist.  This is the reason why the PR checks are failing with the message

```
[error] Test org.apache.daffodil.section15.choice_groups.TestChoice.test_optionalChoice00 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: test optionalChoice00 was not found., took 0.001 sec
[error]     at org.apache.daffodil.tdml.TDMLException$.apply(TDMLException.scala:31)
[error]     at org.apache.daffodil.tdml.DFDLTestSuite.runOneTest(TDMLRunner.scala:388)
[error]     at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:144)
[error]     at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:150)
[error]     at org.apache.daffodil.section15.choice_groups.TestChoice.test_optionalChoice00(TestChoice.scala:43)
[error]     ...
```

tuxji (commented):
Other than the failing test, the PR looks good.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
rsteinberger: Yep, thanks John.

tuxji (approved):
+1

Looks good to me.

stevedlawrence (approved):
+1 :+1: Thanks for the contribution!

PR #417, opened 2020-09-11
Correctly expand varargs in String.format call

stevedlawrence:
Without using _ :* for the varargs, the format string is not properly
expanded and results in an IllegalFormatConversionException. This
currently only occurs when the BacktrackingException is created because
it's the only exception that uses the var args constructor with the
ThinException.

DAFFODIL-2394

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
@@ -33,7 +33,7 @@ abstract class ThinException protected (dummy: Int, cause: Throwable, fmt: Strin
   extends Exception(null, cause, false, false) {
 
   private lazy val msg_ =
-    if (fmt ne null) fmt.format(args)
+    if (fmt ne null) fmt.format(args: _*)
     else if (cause ne null) cause.getMessage()
     else Misc.getNameFromClass(this)
 
--- a/daffodil-io/src/test/scala/org/apache/daffodil/io/TestExceptions.scala
+++ b/daffodil-io/src/test/scala/org/apache/daffodil/io/TestExceptions.scala
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.io
+
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class TestExceptions {
+
+  @Test def testBacktrackingException(): Unit = {
+    val be = new BacktrackingException(123, 456)
+    val msg = be.getMessage()
+    assertTrue(msg.contains("123"))
+    assertTrue(msg.contains("456"))
+  }
+
+}


daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
```diff
@@ -33,7 +33,7 @@ abstract class ThinException protected (dummy: Int, cause: Throwable, fmt: Strin
   extends Exception(null, cause, false, false) {
 
   private lazy val msg_ =
-    if (fmt ne null) fmt.format(args)
+    if (fmt ne null) fmt.format(args: _*)
```
mbeckerle, line 36: I can see how this particular piece of list/argsList polymorphism issometimes useful, but generally, I wish this could be a scala type error so that it's caught at compile time, not run time. Oh well.

mbeckerle (approved):
+1

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
stevedlawrence: Yeah, I think it's just a genearl issue with varargs. Those functions can accept anything, and it's hard to distinguish between a a single list as being the argument or the individual fields being all the arguments.

tuxji (approved):
+1

I can't find any other places in the codebase that calls fmt.format(args), but it's nearly impossible for a grep to tell us whether a variadic function is being called with a single arg of type Any*.

```shell
interran@GH3WPL13E:~/apache/incubator-daffodil-runtime2$ rg 'fmt.format\(args\)'
daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
36:    if (fmt ne null) fmt.format(args)
```

A grep can find places where a variable of type Arg* is being used, but we would have to check each place individually after that and there are about 50 of these places....

```shell
interran@GH3WPL13E:~/apache/incubator-daffodil-asf$ rg -c 'Any\*'
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/Suspension.scala:1
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/Unparser.scala:1
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UnparseError.scala:2
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ParseErrors.scala:2
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala:2
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SchemaSetRuntimeData.scala:1
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessorStateBases.scala:5
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/ProcessingError.scala:2
daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DState.scala:1
daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/SDE.scala:18
daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala:1
daffodil-lib/src/main/scala/org/apache/daffodil/util/Logger.scala:1
daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/ThrowsSDE.scala:14
daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala:2
daffodil-lib/src/main/scala/org/apache/daffodil/api/Diagnostic.scala:3
daffodil-lib/src/test/scala/org/apache/daffodil/schema/annotation/props/TestGeneratedProperties.scala:4
daffodil-core/src/main/scala/org/apache/daffodil/grammar/GrammarTerm.scala:2
daffodil-core/src/main/scala/org/apache/daffodil/dsom/SimpleTypes.scala:1
daffodil-core/src/main/scala/org/apache/daffodil/dsom/PropProviders.scala:1
```

stevedlawrence:
With your idea of just grepping, I looked for a places in the codebase that for ``\.format(`` or ``Any\*`` and did my best to verify that all places are now doing the right thing. Based on visual inspection, I think this was the only case where we weren't handling varargs correctly, or where we were incorretly passing a Seq to format call where we intended to expand the sequence.

It's definitely not a perfect check, and it wouldn't surprise me if there are some potential placeses missed due to other things that we should grep for, but as far as I can tell this was the only incorrect usage.

PR #418, opened 2020-09-11
Add warnings for properties which look like expressions

IanCarlsonOwl:
First attempt at a pull request against the Daffodil repo - please examine carefully.

DAFFODIL-879

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
@@ -23,6 +23,7 @@ import scala.xml.NamespaceBinding
 import org.apache.daffodil.xml.NamedQName
 import java.lang.{ Long => JLong, Boolean => JBoolean }
 import org.apache.daffodil.schema.annotation.props.Found
+import org.apache.daffodil.util.DPathUtil
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.BasicComponent
 
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
@@ -21,11 +21,11 @@ import org.apache.daffodil.schema.annotation.props.PropertyMixin
 
 trait RawCommonRuntimeValuedPropertiesMixin
   extends PropertyMixin {
-  protected final lazy val optionByteOrderRaw = findPropertyOption("byteOrder")
+  protected final lazy val optionByteOrderRaw = findPropertyOption("byteOrder", expressionAllowed = true)
   protected final lazy val byteOrderRaw = requireProperty(optionByteOrderRaw)
-  protected final lazy val optionEncodingRaw = findPropertyOption("encoding")
+  protected final lazy val optionEncodingRaw = findPropertyOption("encoding", expressionAllowed = true)
   protected final lazy val encodingRaw = requireProperty(optionEncodingRaw)
-  protected final lazy val optionOutputNewLineRaw = findPropertyOption("outputNewLine")
+  protected final lazy val optionOutputNewLineRaw = findPropertyOption("outputNewLine", expressionAllowed = true)
   protected final lazy val outputNewLineRaw = requireProperty(optionOutputNewLineRaw)
 
   protected final lazy val optionFillByteRaw = findPropertyOption("fillByte")
@@ -35,9 +35,9 @@ trait RawCommonRuntimeValuedPropertiesMixin
 trait RawDelimitedRuntimeValuedPropertiesMixin
   extends RawCommonRuntimeValuedPropertiesMixin {
 
-  protected final lazy val optionInitiatorRaw = findPropertyOption("initiator")
+  protected final lazy val optionInitiatorRaw = findPropertyOption("initiator", expressionAllowed = true)
   protected final lazy val initiatorRaw = requireProperty(optionInitiatorRaw)
-  protected final lazy val optionTerminatorRaw = findPropertyOption("terminator")
+  protected final lazy val optionTerminatorRaw = findPropertyOption("terminator", expressionAllowed = true)
   protected final lazy val terminatorRaw = requireProperty(optionTerminatorRaw)
   protected final lazy val optionChoiceLengthRaw = findPropertyOption("choiceLength")
   protected final lazy val choiceLengthRaw = requireProperty(optionChoiceLengthRaw)
@@ -49,39 +49,38 @@ trait RawElementRuntimeValuedPropertiesMixin
 
   // these are almost certainly not in scope, but on the local object
   // but not always. A type might define fixed length things for example.
-  protected final lazy val optionLengthRaw = findPropertyOption("length")
+  protected final lazy val optionLengthRaw = findPropertyOption("length", expressionAllowed = true)
   protected final lazy val lengthRaw = requireProperty(optionLengthRaw)
-  protected final lazy val optionOccursCountRaw = findPropertyOption("occursCount")
+  protected final lazy val optionOccursCountRaw = findPropertyOption("occursCount", expressionAllowed = true)
   protected final lazy val occursCountRaw = requireProperty(optionOccursCountRaw)
 }
 
 trait RawSequenceRuntimeValuedPropertiesMixin
   extends RawDelimitedRuntimeValuedPropertiesMixin {
 
-  protected final lazy val optionSeparatorRaw = findPropertyOption("separator")
+  protected final lazy val optionSeparatorRaw = findPropertyOption("separator", expressionAllowed = true)
   protected final lazy val separatorRaw = requireProperty(optionSeparatorRaw)
 }
 
 trait RawLayeringRuntimeValuedPropertiesMixin
   extends PropertyMixin {
-  protected final lazy val optionLayerTransformRaw = findPropertyOption("layerTransform")
+  protected final lazy val optionLayerTransformRaw = findPropertyOption("layerTransform", expressionAllowed = true)
   protected final lazy val layerTransformRaw = requireProperty(optionLayerTransformRaw)
-  protected final lazy val optionLayerEncodingRaw = findPropertyOption("layerEncoding")
+  protected final lazy val optionLayerEncodingRaw = findPropertyOption("layerEncoding", expressionAllowed = true)
   protected final lazy val layerEncodingRaw = requireProperty(optionLayerEncodingRaw)
-  protected final lazy val optionLayerLengthRaw = findPropertyOption("layerLength")
+  protected final lazy val optionLayerLengthRaw = findPropertyOption("layerLength", expressionAllowed = true)
   protected final lazy val layerLengthRaw = requireProperty(optionLayerLengthRaw)
-  protected final lazy val optionLayerBoundaryMarkRaw = findPropertyOption("layerBoundaryMark")
+  protected final lazy val optionLayerBoundaryMarkRaw = findPropertyOption("layerBoundaryMark", expressionAllowed = true)
   protected final lazy val layerBoundaryMarkRaw = requireProperty(optionLayerBoundaryMarkRaw)
 
 }
 
 trait RawEscapeSchemeRuntimeValuedPropertiesMixin
   extends PropertyMixin {
-
   // package private because used in unit test
-  final lazy val optionEscapeCharacterRaw = findPropertyOption("escapeCharacter")
+  final lazy val optionEscapeCharacterRaw = findPropertyOption("escapeCharacter", expressionAllowed = true)
   private[dsom] final lazy val escapeCharacterRaw = requireProperty(optionEscapeCharacterRaw)
-  final lazy val optionEscapeEscapeCharacterRaw = findPropertyOption("escapeEscapeCharacter")
+  final lazy val optionEscapeEscapeCharacterRaw = findPropertyOption("escapeEscapeCharacter", expressionAllowed = true)
   protected final lazy val escapeEscapeCharacterRaw = requireProperty(optionEscapeEscapeCharacterRaw)
   protected final lazy val optionEscapeBlockStartRaw = findPropertyOption("escapeBlockStart")
   protected final lazy val escapeBlockStartRaw = requireProperty(optionEscapeBlockStartRaw)
@@ -95,19 +94,19 @@ trait RawEscapeSchemeRuntimeValuedPropertiesMixin
 trait RawSimpleTypeRuntimeValuedPropertiesMixin
   extends RawCommonRuntimeValuedPropertiesMixin {
 
-  protected final lazy val optionTextStandardDecimalSeparatorRaw = findPropertyOption("textStandardDecimalSeparator")
+  protected final lazy val optionTextStandardDecimalSeparatorRaw = findPropertyOption("textStandardDecimalSeparator", expressionAllowed = true)
   protected final lazy val textStandardDecimalSeparatorRaw = requireProperty(optionTextStandardDecimalSeparatorRaw)
-  protected final lazy val optionTextStandardGroupingSeparatorRaw = findPropertyOption("textStandardGroupingSeparator")
+  protected final lazy val optionTextStandardGroupingSeparatorRaw = findPropertyOption("textStandardGroupingSeparator", expressionAllowed = true)
   protected final lazy val textStandardGroupingSeparatorRaw = requireProperty(optionTextStandardGroupingSeparatorRaw)
 
   
-  protected final lazy val optionBinaryFloatRepRaw = findPropertyOption("binaryFloatRep")
+  protected final lazy val optionBinaryFloatRepRaw = findPropertyOption("binaryFloatRep", expressionAllowed = true)
   protected final lazy val binaryFloatRepRaw = requireProperty(optionBinaryFloatRepRaw)
-  protected final lazy val optionTextBooleanTrueRepRaw = findPropertyOption("textBooleanTrueRep")
+  protected final lazy val optionTextBooleanTrueRepRaw = findPropertyOption("textBooleanTrueRep", expressionAllowed = true)
   protected final lazy val textBooleanTrueRepRaw = requireProperty(optionTextBooleanTrueRepRaw)
-  protected final lazy val optionTextBooleanFalseRepRaw = findPropertyOption("textBooleanFalseRep")
+  protected final lazy val optionTextBooleanFalseRepRaw = findPropertyOption("textBooleanFalseRep", expressionAllowed = true)
   protected final lazy val textBooleanFalseRepRaw = requireProperty(optionTextBooleanFalseRepRaw)
-  protected final lazy val optionCalendarLanguageRaw = findPropertyOption("calendarLanguage")
+  protected final lazy val optionCalendarLanguageRaw = findPropertyOption("calendarLanguage", expressionAllowed = true)
   protected final lazy val calendarLanguageRaw = requireProperty(optionCalendarLanguageRaw)
   protected final lazy val optionBinaryBooleanTrueRepRaw = findPropertyOption("binaryBooleanTrueRep")
   protected final lazy val binaryBooleanTrueRepRaw = requireProperty(optionBinaryBooleanTrueRepRaw)
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
@@ -20,6 +20,8 @@ package org.apache.daffodil.schema.annotation.props
 import org.apache.daffodil.api.LocationInSchemaFile
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.xml.ResolvesQNames
+import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.util.DPathUtil
 
 /**
  * This file is classes and traits to implement
@@ -112,6 +114,12 @@ trait FindPropertyMixin extends PropTypes {
   def SDE(str: String, args: Any*): Nothing
 
   /**
+   * Implemented by users of the mixin so that we can
+   * report property value warnings.
+   */
+  def SDW(warnID: WarnID, std: String, args: Any*): Unit
+
+  /**
    * Implemented in various ways by users of the mixin to find a property. This
    * should never be directly called to find a property. Instead one should use
    * findPropertyOption and friends, which ensure that caching is consistent.
@@ -120,7 +128,12 @@ trait FindPropertyMixin extends PropTypes {
   
   val propCache = new scala.collection.mutable.LinkedHashMap[String, PropertyLookupResult]
 
-  def findPropertyOption(pname: String): PropertyLookupResult = {
+  /**
+   * expressionAllowed describes whether a run-time expression
+   * is valid for this option. Used to decide whether a warning
+   * should be produced if the value "looks like" an expression
+   */
+  def findPropertyOption(pname: String, expressionAllowed: Boolean = false): PropertyLookupResult = {
     val propCacheResult = propCache.get(pname)
     val propRes =
       propCacheResult match {
@@ -131,6 +144,21 @@ trait FindPropertyMixin extends PropTypes {
           lr
         }
       }
+    // if this looks like it could be an expression, issue a warning
+    // but only if expressionAllowed is false
+    if (!expressionAllowed) {
+      propRes match {
+        case Found(v, _, _, _) => {
+          if (DPathUtil.isExpression(v)) {
+            SDW(WarnID.NonExpressionPropertyValueLooksLikeExpression,
+              "Property %s looks like an expression but cannot be an expression: %s", pname, v)
+          }
+        }
+        case _ => {
+          // Do Nothing
+        }
+      }
+    } 
     propRes
   }
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testExpressionPropertyWarnings.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testExpressionPropertyWarnings.tdml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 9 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -120,7 +127,7 @@ trait FindPropertyMixin extends PropTypes {
   
   val propCache = new scala.collection.mutable.LinkedHashMap[String, PropertyLookupResult]
 
-  def findPropertyOption(pname: String): PropertyLookupResult = {
+  def findPropertyOption(pname: String, isForExpression: Boolean = false): PropertyLookupResult = {
```
mbeckerle, line 130: Add scaladoc explaining purpose of isForExpression and that only for properties that can be expressions have to pass true. 

Maybe the name should be "canBeExpressionOrNot" because many properties are optionally for expressions e.g, dfdl:byteOrder which can be literally "bigEndian" or can be an expression. Other properties like dfdl:inputValueCalc are always an expression, but we don't need this to be passed as true for them.

mbeckerle (changes requested):
Minor code style changes, and suggestions.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +138,22 @@ trait FindPropertyMixin extends PropTypes {
...
         }
       }
+    /* if this looks like it could be an epression, issue a warning */
+    /* only if isForExpression is false*/
+    /* for runtime mixins, need to supply true, other places, no change*/
+    if( ! isForExpression ) {
+      propRes match{ 
+        case Found( v, _, _, _ ) => {
+
+          if( v.startsWith( "{")) {
```
mbeckerle, line 148: We should make this test a bit more selective, like check if it begins with double curly braces (which is an escape, so don't warn), and also check that it begins and ends with a single brace.

tuxji (changes requested):
I don't have much to add to Mike's comments except as noted below.  Will approve after I see a fixup commit.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +138,22 @@ trait FindPropertyMixin extends PropTypes {
...
+            SDW( WarnID.NonExpressionPropertyValueLooksLikeExpression, "Property %s that looks like an expression cannot be an expression: %s", pname, v )
```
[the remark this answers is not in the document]
stevedlawrence: I'm definitely in favor of enabling scalariform. [DAFFODIL-2133](https://issues.apache.org/jira/browse/DAFFODIL-2133) is the issue for that. I'd like to enable the check on pull requests so it fails an PR's that don't follow the Daffodil style guidelines. The problem is I'm sure a lot of our code doesn't currently match the style guides, so we'd need to fix that first. Probably not a small task.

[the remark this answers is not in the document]
stevedlawrence: Actually, enabling sclariform probably isn't a ton of work since it autoformats everything. I guess the real work is figuring out if any of the defaults need to be changed (hopefully not, hopefully we following the Scala style guides pretty closely), and then reviewing all of the changes. And hopefully a bunch of formatting changes wouldn't be too difficult to review.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
IanCarlsonOwl: To restate - we want to warn only if the value begins and ends with { }, and doesn't begin with {{. Is that your intent?

mbeckerle: Confirmed.

stevedlawrence: DPathUtil has an ``isExpression`` function which I think has that exact logic. Probably makes sense to just reuse that so if what an expression looks like ever changes we only have to update that one function.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
```diff
@@ -21,25 +21,26 @@ import org.apache.daffodil.schema.annotation.props.PropertyMixin
 
 trait RawCommonRuntimeValuedPropertiesMixin
   extends PropertyMixin {
-  protected final lazy val optionByteOrderRaw = findPropertyOption("byteOrder")
+  final protected def forExpression = true
```
stevedlawrence, line 24: It looks like we only ever modify that mixins that we do by hand. There are also a bunch of mixins (most of the mixes) that are autogenerated by running "sbt genManaged". Do we need to update the daffodil-propgen subproject to add these forExpression things as well where appropriate?

mbeckerle: Likely not, because all the runtime-valued properties are not auto-generated by propgen, but are part of a by-hand mixin. 
The generated properties are either never expressions or always expressions. The only ones we have to deal with are the ones that are sometimes expressions.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +138,22 @@ trait FindPropertyMixin extends PropTypes {
           lr
         }
       }
+    /* if this looks like it could be an epression, issue a warning */
+    /* only if isForExpression is false*/
+    /* for runtime mixins, need to supply true, other places, no change*/
+    if( ! isForExpression ) {
+      propRes match{ 
+        case Found( v, _, _, _ ) => {
+
```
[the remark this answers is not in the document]
IanCarlsonOwl: Blank line removed, and space between ! and isForExpression removed.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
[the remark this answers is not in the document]
IanCarlsonOwl: Line wrapped before ""Property"

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +138,22 @@ trait FindPropertyMixin extends PropTypes {
           lr
         }
       }
+    /* if this looks like it could be an epression, issue a warning */
```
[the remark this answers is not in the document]
IanCarlsonOwl: Replaced /* */ with //

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
IanCarlsonOwl: Added a scaladoc. The autogenerated block that pops up in IntelliJ for the function header didn't seem to match other functions in the file. Does this scaladoc look ook?

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
IanCarlsonOwl: Moved DPathUtil from daffodil-runtime1 to daffodil-lib and invoked isExpression rather than forming the check here.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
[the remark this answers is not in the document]
IanCarlsonOwl: Fixed spelling of expression and changed comment syntax.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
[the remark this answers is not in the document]
IanCarlsonOwl: I've updated the formatting to be more in line with the rest of the file. I don't see any additional changes that need to be made in this context given the existence of DAFFODIL-2133

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
IanCarlsonOwl: My understanding of this exchange is that there's no change to be made at the moment. Is that correct?

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +144,21 @@ trait FindPropertyMixin extends PropTypes {
...
+          if(DPathUtil.isExpression(v)) {
+            SDW(WarnID.NonExpressionPropertyValueLooksLikeExpression,
+              "Property %s looks like an expression but cannot be an expression: %s", pname, v)
+          }
+        }
+        case _ => {
+          // Do Nothing
+        }
+      }
+    }
```
stevedlawrence, line 161: Needs a space after the ``if`` keyword above.

Also, this is just personal style preference, but all the indentions make this kind of hard to read to me. Anyone else get that, or is it just me? I'm wondering if something like this would be preferred?
```scala
if (!expressionAllowed) {
  val isExpression = propRes.toOption.map { DPathUtil.isExpression(_) }.getOrElse(false)
  schemaDefinitionWarningUnless(
    WarnID....,
    !isExpression,
   "Property ..."
  )
}

**Review by stevedlawrence — COMMENTED** [review]
+1 with some questions about potential clean ups.

File: daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala [hunk 8]
```diff
@@ -49,69 +50,71 @@ trait RawElementRuntimeValuedPropertiesMixin
...
   protected final lazy val layerLengthRaw = requireProperty(optionLayerLengthRaw)
-  protected final lazy val optionLayerBoundaryMarkRaw = findPropertyOption("layerBoundaryMark")
+  protected final lazy val optionLayerBoundaryMarkRaw = findPropertyOption("layerBoundaryMark", forExpression)
   protected final lazy val layerBoundaryMarkRaw = requireProperty(optionLayerBoundaryMarkRaw)
 
 }
 
 trait RawEscapeSchemeRuntimeValuedPropertiesMixin
   extends PropertyMixin {
+  final protected def forExpression = true
```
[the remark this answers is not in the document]
IanCarlsonOwl: I'm not opposed - calling out the parameter by name does seem like it'll make the purpose more transparent. I'll make that change if there's no disagreement from others.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -120,7 +128,12 @@ trait FindPropertyMixin extends PropTypes {
   
   val propCache = new scala.collection.mutable.LinkedHashMap[String, PropertyLookupResult]
 
-  def findPropertyOption(pname: String): PropertyLookupResult = {
+  /**
+   * expressionAllowed describes whether a run-time expression
+   * is valid for this option. Used to decide whether a warning
+   * should be produced if the value "looks like" an expression
+   */
+  def findPropertyOption(pname: String, expressionAllowed: Boolean = false): PropertyLookupResult = {
```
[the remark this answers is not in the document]
IanCarlsonOwl: Updated both instances to use the same expressionAllowed name. My only concern with the naming convention is that both isForExpression and expressionAllowed are a bit misleading in the case of fields which must be expressions. However, expressionAllowedButNotRequired is a bit unwieldy.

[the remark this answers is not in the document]
stevedlawrence: Another option might be ``canBeAnExpression``, which is a little less unwieldy and I think still implies it's not required.

But I don't feel strongly about the name, and I don't think ``expressionAllowed`` is terrible. I can't seem to find Mike's sugestion, maybe I made that up. But in general I usually prefer names thats a little unweildy but accurate over names that are shorter but maybe less accurate. We have plenty of variables that are basically sentences purely to make the code easier to understand and to be more meaningful.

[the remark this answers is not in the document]
mbeckerle: Note that we DONT need this argument passed as true for things that MUST be expressions. Only for things that are optionally expressions. Hence, the name should reflect that conditionality, and scaladoc should make this clear.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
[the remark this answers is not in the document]
mbeckerle: I suggest the expressionAllowed name is better than forExpression.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
mbeckerle: Looks good.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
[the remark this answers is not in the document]
mbeckerle: I like SL's suggestion that we pass this using the name.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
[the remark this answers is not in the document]
mbeckerle: expressionAllowed works for me.

mbeckerle (approved):
+1 Minor name consistency change to do. To me this looks good.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
```diff
@@ -23,6 +23,7 @@ import scala.xml.NamespaceBinding
 import org.apache.daffodil.xml.NamedQName
 import java.lang.{ Long => JLong, Boolean => JBoolean }
 import org.apache.daffodil.schema.annotation.props.Found
+import org.apache.daffodil.util.DPathUtil
```
tuxji, line 26: Is this import really used?  Is there an `import org.apache.daffodil.dpath.DPathUtil` that should be removed?

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +144,21 @@ trait FindPropertyMixin extends PropTypes {
           lr
         }
       }
+    // if this looks like it could be an expression, issue a warning
+    // but only if expressionAllowed is false
+    if(!expressionAllowed) {
```
tuxji, line 149: Formatting: should be `if (!expressionAllowed) {` not `if(!expressionAllowed) {`

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
```diff
@@ -131,6 +144,21 @@ trait FindPropertyMixin extends PropTypes {
...
+      propRes match {
+        case Found(v, _, _, _) => {
+          if(DPathUtil.isExpression(v)) {
```
tuxji, line 152: Formatting: should be `if (` not `if(`

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testExpressionPropertyWarnings.tdml
```diff
@@ -0,0 +1,46 @@
...
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="delimited"/>
+    <xs:element name="root" type="xs:double" dfdl:textNumberPattern="#####0.0#####" dfdl:textStandardInfinityRep="{../ex:thisCantBeAnExpression}" />
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="expressionPropertyWarning1" model="S1" root="root">
+    <tdml:document>1.0</tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:root xsi:type="xs:double">1.0</ex:root>
```
tuxji, line 37: Why do you have an `xsi:type="xs:double"` when the schema already declares the root element with the xs:double type?

tuxji (approved):
+1 with minor nits noted

daffodil-core/src/main/scala/org/apache/daffodil/dsom/CompiledExpression.scala
IanCarlsonOwl: The DPathUtil import is used for the new call to DPathUtil.isExpression on line 64.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testExpressionPropertyWarnings.tdml
IanCarlsonOwl: Without that decoration, the 1.0 was reduced to 1, and the round-trip parse-unparse ended up having different results on either end.

stevedlawrence: Hmm, I don't think Daffodil takes into account ``xsi:type`` except for a comparing expected vs actual results. And even then, it only uses it it for ``xs:anyURI`` and calendar types (e.g. ``xs:dateTime``). I'm pretty sure ``xsi:type="xs:double"`` shouldn't have an affect.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/PropertyScoping.scala
IanCarlsonOwl: Made a fix-up commit before rebasing that resolves this.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/testExpressionPropertyWarnings.tdml
mbeckerle: I thought the comparison did type specific equality using xsi:type information?

stevedlawrence: It does, but not for numeric types--only for xs:anyURI, xs:hexBinary, xs:time, and xs:dateTime. These are the only ones where IBM DFDL and Daffodil have different output so we needed a little extra smarts. We could add it for the remainder of types, but haven't had the need so far. 

https://github.com/apache/incubator-daffodil/blob/master/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala#L1088-L1108

mbeckerle: Ah, so for now, the only way to get a test like this to work is to make sure textNumberPattern constraints the number layout in the data, and use true values in the infoset, or just use strings as the element types. 

Honestly, I think we do need this for double and float, as they are so snarky about comparison. Without type/schema, it's all strings, so, 1 and 1.0 in XML are not equivalent.

mbeckerle: Created DAFFODIL-2402

IanCarlsonOwl:
I have squashed all commits in to one. I believe this pull request is ready to close, along with DAFFODIL-879

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
```diff
@@ -21,25 +21,25 @@ import org.apache.daffodil.schema.annotation.props.PropertyMixin
...
   protected final lazy val byteOrderRaw = requireProperty(optionByteOrderRaw)
-  protected final lazy val optionEncodingRaw = findPropertyOption("encoding")
+  protected final lazy val optionEncodingRaw = findPropertyOption("encoding", expressionAllowed = true)
   protected final lazy val encodingRaw = requireProperty(optionEncodingRaw)
-  protected final lazy val optionOutputNewLineRaw = findPropertyOption("outputNewLine")
+  protected final lazy val optionOutputNewLineRaw = findPropertyOption("outputNewLine", expressionAllowed = true)
   protected final lazy val outputNewLineRaw = requireProperty(optionOutputNewLineRaw)
 
-  protected final lazy val optionFillByteRaw = findPropertyOption("fillByte")
+  protected final lazy val optionFillByteRaw = findPropertyOption("fillByte", expressionAllowed = true)
```
stevedlawrence, line 31: fillByte is a DFDL String Literal and so cannot be an expression:

https://daffodil.apache.org/docs/dfdl/#_Toc130873645

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
```diff
@@ -21,25 +21,25 @@ import org.apache.daffodil.schema.annotation.props.PropertyMixin
...
   extends RawCommonRuntimeValuedPropertiesMixin {
 
-  protected final lazy val optionInitiatorRaw = findPropertyOption("initiator")
+  protected final lazy val optionInitiatorRaw = findPropertyOption("initiator", expressionAllowed = true)
   protected final lazy val initiatorRaw = requireProperty(optionInitiatorRaw)
-  protected final lazy val optionTerminatorRaw = findPropertyOption("terminator")
+  protected final lazy val optionTerminatorRaw = findPropertyOption("terminator", expressionAllowed = true)
   protected final lazy val terminatorRaw = requireProperty(optionTerminatorRaw)
-  protected final lazy val optionChoiceLengthRaw = findPropertyOption("choiceLength")
+  protected final lazy val optionChoiceLengthRaw = findPropertyOption("choiceLength", expressionAllowed = true)
```
stevedlawrence, line 42: choiceLength also does not allow expressions:

https://daffodil.apache.org/docs/dfdl/#_Toc398030783

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
```diff
@@ -49,69 +49,68 @@ trait RawElementRuntimeValuedPropertiesMixin
...
+  final lazy val optionEscapeEscapeCharacterRaw = findPropertyOption("escapeEscapeCharacter", expressionAllowed = true)
   protected final lazy val escapeEscapeCharacterRaw = requireProperty(optionEscapeEscapeCharacterRaw)
-  protected final lazy val optionEscapeBlockStartRaw = findPropertyOption("escapeBlockStart")
+  protected final lazy val optionEscapeBlockStartRaw = findPropertyOption("escapeBlockStart", expressionAllowed = true)
   protected final lazy val escapeBlockStartRaw = requireProperty(optionEscapeBlockStartRaw)
-  protected final lazy val optionEscapeBlockEndRaw = findPropertyOption("escapeBlockEnd")
+  protected final lazy val optionEscapeBlockEndRaw = findPropertyOption("escapeBlockEnd", expressionAllowed = true)
   protected final lazy val escapeBlockEndRaw = requireProperty(optionEscapeBlockEndRaw)
-  protected final lazy val optionExtraEscapedCharactersRaw = findPropertyOption("extraEscapedCharacters")
+  protected final lazy val optionExtraEscapedCharactersRaw = findPropertyOption("extraEscapedCharacters", expressionAllowed = true)
```
stevedlawrence, line 89: escapeBlockStart, escapeBlockEnd, and extraEscapedCharacters cannot be expressions:

https://daffodil.apache.org/docs/dfdl/#_Toc398030751

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
```diff
@@ -49,69 +49,68 @@ trait RawElementRuntimeValuedPropertiesMixin
...
+  protected final lazy val optionTextBooleanFalseRepRaw = findPropertyOption("textBooleanFalseRep", expressionAllowed = true)
   protected final lazy val textBooleanFalseRepRaw = requireProperty(optionTextBooleanFalseRepRaw)
-  protected final lazy val optionCalendarLanguageRaw = findPropertyOption("calendarLanguage")
+  protected final lazy val optionCalendarLanguageRaw = findPropertyOption("calendarLanguage", expressionAllowed = true)
   protected final lazy val calendarLanguageRaw = requireProperty(optionCalendarLanguageRaw)
-  protected final lazy val optionBinaryBooleanTrueRepRaw = findPropertyOption("binaryBooleanTrueRep")
+  protected final lazy val optionBinaryBooleanTrueRepRaw = findPropertyOption("binaryBooleanTrueRep", expressionAllowed = true)
   protected final lazy val binaryBooleanTrueRepRaw = requireProperty(optionBinaryBooleanTrueRepRaw)
-  protected final lazy val optionBinaryBooleanFalseRepRaw = findPropertyOption("binaryBooleanFalseRep")
+  protected final lazy val optionBinaryBooleanFalseRepRaw = findPropertyOption("binaryBooleanFalseRep", expressionAllowed = true)
```
stevedlawrence, line 113: binaryBooleanTrueRep and binaryBooleanFalseRep cannot be an expression:

https://daffodil.apache.org/docs/dfdl/#_Toc398030762

stevedlawrence (changes requested):
I double checked the properties that have ``expressionAllowed = true`` and I think a handful of them actually aren't allowed to be expressions. I've provided links to the DFDL Spec that describes the properties.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RawCommonRuntimeValuedPropertiesMixin.scala
IanCarlsonOwl: I've updated the commit to remove the allowedExpression=true from the lines where it was incorrect.

stevedlawrence (approved):
+1 to the new changes. Thanks for the contribution!

PR #13, opened 2017-11-30
Fix truncation bug when unparsing escapeKind="character"

stevedlawrence:
Daffodil had a logic bug when unparsing a field with an escape scheme
with escapeKind="character". In such a case, we run the field DFA
looking at each character to unparse to see if it could potentially
start a delimiter and thus may need to be escaped. If we do find such a
character, we then run each delimiter DFA on the input to determine if
the delimiter actually existed or if we just found a character the looks
like it starts a delimiter but actually doesn't. We then gather up
successful matches for a longest match check. This logic is all
correct.

The problem is that for escapeKind="character", each time a delimiter
DFA failed to match, we would move the field register to the next rule. So
if we checked for 10 delimiters and 9 failed, we would advance the field
register 9 times, which is wrong and would cause the field register to
end prematurely, leading to truncated unparsed data.

Instead, we should wait until all the delimiter DFAs are checked, and
only if none match (i.e. success.isEmpty is true) then we advance the
field register by one rule to add that character to the field. This
patch fixes the text delimited unparser to use this logic for
escapeKind="character". This is actually very similar logic that
escapeKind="block" had, so this matches that correct behavior. Comments
were also added to clarify this section of code.

DAFFODIL-1851

PR #419, opened 2020-09-14
Reduce performance degradation caused by commit b0f59ef7c8

stevedlawrence:
- Remove the DIFinalizable trait and setFinal function and replace with
  direct modification of isFinal. Scala created some pretty inefficient
  byte code for this which actually made a noticeable difference in
  performance when removed.
- Add function to DINodes to efficiently return a Maybe of the last
  child. This replaces the lastOption and avoids an Option allocation
- Remove the containerNode def and just keep track of the container node
  in the InfosetWalker with a new stack. Avoids virtual function calls
  and instanceOf/asInstanceOf.
- Skip batches of walk() calls. This avoids the overhead related to step
  logic when no progress is being made due to PoU's and allows batching
  of steps rather than stepping each time a infoset node is added.
- Replace the various infoset walker objects with functions. Rather than
  returning objects in a Maybe and calling a function in those optionst,
  just call the functions directly and return a boolean that says
  whether a step was taken. Essentially the same logic, but avoids
  object creation and virtual functions.
- Avoid accessing the top of the infoset walker stacks multiple times.
  Instead, access once and pass down as parameters. These stack accesses
  actually showed up quite a bit in profiling
- Remove match/case on types, and instead use functions to imply the
  type. Avoids instanceOf/asInstanceOf which have a noticeable
  overhead in inner loops
- Add a new setTop function to modify the top of an MStack. Avoids logic
  to pop and then push a new value.
- Remove extra logic about saved last child node, this didn't make much
  of a difference in when we called set final, and just added extra
  overhead
- Remove calls to walk() made when PoU's were resolved. Instead, only
  walk() when new elements are added. Elements are added all the time so
  these extra calls just added overhead. The new step skipping logic
  also helps with this.

DAFFODIL-2396

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
@@ -432,7 +432,7 @@ sealed trait RegularElementUnparserStartEndStrategy
           } else {
             val doc = state.documentElement
             doc.addChild(res, state.tunable) // DIDocument, which is never a current node, must have the child added
-            doc.setFinal() // that's the only child.
+            doc.isFinal = true // that's the only child.
           }
           res
         } else {
@@ -477,7 +477,7 @@ sealed trait RegularElementUnparserStartEndStrategy
       }
       val cur = state.currentInfosetNode
       if (cur.isComplex)
-        cur.asComplex.setFinal()
+        cur.isFinal = true
       state.currentInfosetNodeStack.pop
 
       move(state)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
@@ -65,7 +65,7 @@ import org.apache.daffodil.xml.XMLUtils
 import passera.unsigned.ULong
 import org.apache.daffodil.dsom.DPathCompileInfo
 
-sealed trait DINode extends DIFinalizable {
+sealed trait DINode {
 
   def asSimple: DISimple = {
     this match {
@@ -86,6 +86,16 @@ sealed trait DINode extends DIFinalizable {
     }
   }
   def isComplex: Boolean
+
+  def asArray: DIArray = {
+    this match {
+      case diArray: DIArray => diArray
+      case _ => {
+        erd.toss(new InfosetWrongNodeType("arrayType", this)) // see comment with exception class definition for why this can happen
+      }
+    }
+  }
+  def isArray: Boolean
 
   def asElement: DIElement =
     if (isSimple) asSimple
@@ -109,18 +119,15 @@ sealed trait DINode extends DIFinalizable {
   final def numChildren = contents.length
 
   /**
-   * Returns the DINode that contains this DINode. Note that this behavior is
-   * slightly different from the "parent". The parent method always returns a
-   * DIComplex, even if this DINode is in an array. This usually makes sense
-   * and is what the caller intended when they call parent.
-   *
-   * However, we sometimes want the exact DINode that contains this DINode. So
-   * if this DINode is a DIElement in an array, then this returns the DIArray.
-   * Otherwise (i.e. DIArray or non-array DIElement) then it should return the
-   * DIComplex parent. Note that if the node is a DIDocument it should return
-   * null, which mimics the behavior of .parent.
+   * If there are any children, returns the last child as a One(lastChild).
+   * Otherwise returns Nope. Note that because Daffodil may set children to
+   * null when it determines they are no longer needed, and One(null) is not
+   * allowed, this must return Nope if the last child has been freed. Thus,
+   * this is not a reliable mechanism to determine if children exist. It should
+   * only be used where we need to modify the state (e.g. isFinal) of the last
+   * child IF it exists.
    */
-  def containerNode: DINode
+  def maybeLastChild: Maybe[DINode]
 
   /**
    * Free memory associated with the child at a specified index. Note that this
@@ -129,6 +136,22 @@ sealed trait DINode extends DIFinalizable {
    * child is used in DPath expressions this should not free the child.
    */
   def freeChildIfNoLongerNeeded(index: Int): Unit
+
+  /**
+   * When unparsing, arrays and complex elements have a notion of being finalized.
+   * This is when we know that no more elements will be added to them, so things
+   * like fn:count can return a value knowing it won't change, and fn:exists can
+   * return false, knowing nobody will subsequently append the item that was being
+   * questioned.
+   */
+  var isFinal: Boolean = false
+
+  /**
+   * use to require it be finalized or throw the appropriate
+   * Array or Complex exception.
+   */
+  def requireFinal: Unit
+
 }
 
 /**
@@ -829,6 +852,7 @@ sealed trait DIElement
 
   def isSimple: Boolean
   def isComplex: Boolean
+  def isArray: Boolean
   override final def name: String = erd.name
   override final def namespace: NS = erd.targetNamespace
   override final def namedQName = erd.namedQName
@@ -911,10 +935,6 @@ sealed trait DIElement
 
   override def parent = _parent
 
-  override def containerNode =
-    if (array.isDefined) array.get.asInstanceOf[DINode]
-    else parent.asInstanceOf[DINode]
-
   def diParent = _parent.asInstanceOf[DIComplex]
 
   override def setParent(p: InfosetComplexElement): Unit = {
@@ -984,8 +1004,6 @@ final class DIArray(
 
   private lazy val nfe = new InfosetArrayNotFinalException(this)
 
-  override def containerNode = parent
-
   override def requireFinal: Unit = {
     if (!isFinal) {
       // If this DIArray isn't final, either we haven't gotten all of its
@@ -999,7 +1017,7 @@ final class DIArray(
       // isn't final then we could still be wainting for events to come in, so
       // throw an nfe.
       if (parent.isFinal) {
-        setFinal()
+        isFinal = true
       } else {
         throw nfe
       }
@@ -1008,6 +1026,7 @@ final class DIArray(
 
   override def isSimple = false
   override def isComplex = false
+  override def isArray = true
 
   // Parsers don't actually call setHidden on DIArrays, only on DIElements. But
   // DIArrays are always created when there is at least one child element, and
@@ -1056,6 +1075,17 @@ final class DIArray(
   }
 
   override def contents: IndexedSeq[DINode] = _contents
+
+  override def maybeLastChild: Maybe[DINode] = {
+    val len = _contents.length
+    if (len > 0) {
+      // note that if _contents(len - 1) has been removed (i.e. set to null)
+      // then this will by Maybe(null) which is a Nope. This is expected
+      // behavior
+      Maybe(_contents(len - 1))
+    }
+    else Nope
+  }
 
   /**
    * Note that occursIndex argument starts at position 1.
@@ -1129,6 +1159,7 @@ sealed class DISimple(override val erd: ElementRuntimeData)
 
   final override def isSimple = true
   final override def isComplex = false
+  final override def isArray = false
 
   def contents: IndexedSeq[DINode] = IndexedSeq.empty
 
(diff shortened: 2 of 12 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
```diff
@@ -477,7 +477,7 @@ sealed trait RegularElementUnparserStartEndStrategy
       }
       val cur = state.currentInfosetNode
       if (cur.isComplex)
-        cur.asComplex.setFinal()
+        cur.isFinal = true
```
mbeckerle, line 480: So the issue here is the finalizable mixin, resulting in virtual functions, when just a isFinal data member is not done by way of mixin? 

So the guidance I guess is "fewer 'thin' mixins" in the runtime? You have to really need the mixin for it to be worth the overhead?

I''m trying to figure out what lessons learned, and what guidance we put in the coding standards notes, and try to remember to suggest during code reviews.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -1458,6 +1454,12 @@ sealed class DIComplex(override val erd: ElementRuntimeData)
 
   override def children = childNodes.toStream
 
+  override def maybeLastChild: Maybe[DINode] = {
+    val len = childNodes.length
+    if (len > 0) Maybe(childNodes(len - 1))
```
mbeckerle, line 1459: Add comment here that if the childNode has been removed (set to null), then Maybe(childNodes(len - 1 )) will be Maybe(null) which is Nope, and that this is expected behavior.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
```diff
@@ -178,7 +179,11 @@ class InfosetWalker private (
    * we increment this value on the top of the stack so that the starting index
    * is correct.
    */
-  private var containerNode: DINode = startingContainerNode
+  private var containerNodeStack: MStackOf[DINode] = {
```
mbeckerle, line 182: Really? Maintaining this whole separate stack is faster than just doing a test to see if enclosing parent is a DIArray and indirecting past that to the enclosing DIComplex if so? That's kind of disappointing.   Maybe we should just have a DINode.nodeKind : Int, where 0 is DISimple, 1 is DIComplex, 2 is DIArray, etc. and then containerNode would just be { if (nodeKind == 2) parent.parent else parent }

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
```diff
@@ -201,26 +229,60 @@ class InfosetWalker private (
...
+  def walk(lastWalk: Boolean = false): Unit = {
     Assert.usage(!finished)
 
-    var maybeNextStep: Maybe[InfosetWalkerStep] = Nope
-    while ({
-      maybeNextStep = getNextStep()
-      maybeNextStep.isDefined
-    }) {
-      maybeNextStep.get.step()
+    if (walkSkipRemaining > 0 && !lastWalk) {
```
mbeckerle, line 242: I don't understand why this improves performance. 

Every node has to eventually be walked. Are we cutting down the number of times walk is called only to return without being able to make progress?  So this design is effectively polling and what we're doing is just to cut down on the polling frequency?

If so, then it suggests to me we should only be calling walk() when a there is no outstanding PoU and only after N elements are added to the infoset, for some fairly large N (like 100).  But perhaps this algorithm achieves that same thing?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
```diff
@@ -321,132 +390,114 @@ class InfosetWalker private (
...
-          case c: DIComplex => {
-            outputter.endComplex(c)
-          }
-          case _ => Assert.impossible()
-        }
+    if (containerIndex < children.size) {
+      // This block means we need to create a start event for the element in
+      // the children array at containerIndex. We then need to mutate the walker
+      // state so the next call to step() is for either the first child of
+      // this element or the next sibling.
```
mbeckerle, line 445: Don't understand "or the next sibling". What does that mean, or rather, why would that be the case?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
```diff
@@ -321,132 +390,114 @@ class InfosetWalker private (
...
-        // now want to move to the parent container, potentially free up the
-        // memory associated with this container, and then move to the next
-        // sibling of this container
-        moveToContainer()
+      val child = children(containerIndex)
+
+      if (child.isSimple) {
+        if (!child.isHidden || walkHidden) {
+          val simple = child.asInstanceOf[DISimple]
+          outputter.startSimple(simple)
```
mbeckerle, line 452: We could combine these two virtual functions into one, or provide a combined variant startEndSimple(simple). That would cut two virtual function calls down to one.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -203,31 +201,27 @@ abstract class SequenceParserBase(
...
               // other non-array elements get set as final.
-              newLastChildNode.get match {
-                case a: DIArray => {
-                  val lastArrayElem = a.contents.last
-                  if (lastArrayElem != null) {
-                    lastArrayElem.setFinal()
-                    pstate.walker.walk()
-                  }
+              val lastChild = newLastChildNode.get
+              if (!lastChild.isSimple && !lastChild.isComplex) {
```
mbeckerle, line 218: Two calls to isSimple and isComplex? Let's just add an isArray method to go with the others. It's again disappointing that x.isInstanceOf[DIArray] isn't every bit as fast as having an isArray boolean. If isSimple and isComplex are virtual functions defined on DINode and DISimple and DIComplex, then I fail to understand why they wouldn't be pretty much identical in preformance to x.isInstanceOf, why would this be? 

This really suggests we should have an Int member which we initialize to an integer enum value that identifies simple/complex/array, and then we can have inline final def isSimple = nodeKindEnum eq SIMPLE and similar for isComplex and isArray. These would then not be virtual functions at all. But I don't like having to play that sort of tricks. We're just reinventing virtual dispatch, doing so in a less general way that we know will just turn into an integer equality test.

mbeckerle (approved):
+1 I have some questions that might want to be answered by improving code comments, and small suggestions, but really this is fine to me as is.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
stevedlawrence: I think it's actually both a mixture of removeing the trait, and making things publicly modifable. For example, with a trait we get something like this:

```
public class WithFinalTrait implements Finalizable {
  private boolean Finalizable$$_isFinal;

  public void Finalizable$$_isFinal_$eq(boolean);
    Code:
       0: aload_0
       1: iload_1
       2: putfield      #15                 // Field Finalizable$$_isFinal:Z
       5: return

  public void setFinal();
    Code:
       0: aload_0
       1: invokestatic  #27                 // Method Finalizable$class.setFinal:(LFinalizable;)V
       4: return

}

public abstract class Finalizable$class {
  public static void setFinal(Finalizable);
    Code:
       0: aload_0
       1: iconst_1
       2: invokeinterface #13,  2           // InterfaceMethod Finalizable.Finalizable$$_isFinal_$eq:(Z)V
       7: return
}

```

So to set the flag we first call setFinal, which calls the static setFinal method on Finalizable, which then calls \_isFinal_$eq, which finaly sets the variable to set the value. So three function calls: setFinal, static setFinal, and isFinal_$eq.

Without the trait, but with private _isFinal, the bytecode is this:

public class WithOutFinalTrait {
  private boolean _isFinal;

  private void _isFinal_$eq(boolean);
    Code:
       0: aload_0
       1: iload_1
       2: putfield      #13                 // Field _isFinal:Z
       5: return

  public void setFinal();
    Code:
       0: aload_0
       1: iconst_1
       2: invokespecial #22                 // Method _isFinal_$eq:(Z)V
       5: return

}

So a call to setFinal directly calls _isFinal_$eq to set the variable. So we remove the one static setFinal call.

With a public final flag and direct modification, the bytecode looks like this:

public class PublicIsFinal {
  private boolean _isFinal;

  public boolean _isFinal();
    Code:
       0: aload_0
       1: getfield      #13                 // Field _isFinal:Z
       4: ireturn

  public void _isFinal_$eq(boolean);
    Code:
       0: aload_0
       1: iload_1
       2: putfield      #13                 // Field _isFinal:Z
       5: return
}


So _isFinal_$eq is called directly which modifies the field directly. So now we're down to a single function call.

So we go from three functions calls with a trait, to one function call with public member. This feels like it shouldn't make that much of a difference, but it definitely did. It wasn't huge, and I'm not sure I would reccommend we do this everywhere because it does make readability/reusability more difficult, but in performance critical situations, avoiding traits and private vars seems to help a bit.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
stevedlawrence: I think the issue was that containerNode requires a DINode, but parent is an InfosetElement, so we have to do instanceOf/asInstanceOf checks. Whatever bytecode this turns into appears to be relatively slow. If we got rid of all the InfosetElement distinction, we could probably get rid of alot of this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
stevedlawrence: That's correct. The purpose of this is to avoid performing walks() when progress wont' be made.

This algorithm is essentially donig "call only after some large N elements are created" since we now only call walk after elements are created. It's a little smarter since it increases N if we fail to make progress, and then decreases once progress is made. Currently that N starts out at 32 and we double each time progress isn't made up to a max of 2048.

We could also check it there are PoU's, but maybeDoStep essentially does that, but also takes into account  other kinds of blocks. This also allows us to increase how many calls to walk we skip when this does detect that there's somethign blocking from walking.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
stevedlawrence: The state of the infoset walker is always pointing at a child index for elements to create events for. This block is going to create an event for the containerIndex'th element in the current containerNode. After it creates that event, it needs to mutate state so the next time we take a step we are looking at the next element. If this is a complex type, the next element is the first child. If this is not a complex element, the next element is the next sibling. So that's what this is trying to say. I'll update this to be a bit more clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetWalker.scala
stevedlawrence: I actually tried that change and it seemed to not make a difference in this case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
stevedlawrence: Yeah, I'm surprised to. Of all the changes, I think this one didn't make a huge difference, it was definitely withing the margin of error for Java. But it did seem to consistent error on the positive side of that difference so I figured I'd keep it in to try to eek out every bit of performance. I'll add an isArray so this is at least more readable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
stevedlawrence: Will do

tuxji (approved):
+1

Great comments by Mike and fixes look good.

PR #420, opened 2020-09-25
Removed isDefinedForLength check for complexType elements

IanCarlsonOwl:
This change removes the isDefinedForLength check for complexType
elements as well a choice types so we don't try to look ahead
by a potentially unbounded length. This is to prevent us from
looking far enough ahead that we lose the ability to backtrack,
or lose the "beginning" of the objects being checked.

Also removed the assert on isLimitOk. This type of failure can
more correctly be caught later in parsing, such that we'll
backtrack correctly.

Also switched one of the text files from tabs to 2 space indentation.

DAFFODIL-2395

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
@@ -17,13 +17,13 @@
 -->
 
 <tdml:testSuite suiteName="ExplicitTests"
-	description="Section 12 - lengthKind=explicit" xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
```diff
@@ -32,96 +32,96 @@
...
-					<street>Ridgewood Circle</street>
-					<city>Rochester</city>
-					<state>NY</state>
-				</address>
-			</tdml:dfdlInfoset>
-		</tdml:infoset>
-	</tdml:parserTestCase>
-
-	<tdml:defineSchema name="test_ExplicitLengthBits" elementFormDefault="unqualified">
+    <xs:element name="address" dfdl:lengthKind="implicit">
```
mbeckerle, line 35: Generally it is better to do these sort of whitespace-only improvements which have no actual code change involved, in a separate commit at least, and perhaps a separate PR. That way we would still be able to see what you changed and added here instead of just seeing "a whole lot of green", and having to gestalt what's modified and what's new.

mbeckerle (approved):
+1

tuxji (approved):
+1

mbeckerle:
Merging this even though the scala 2.11 build on windows didn't pass its check.

PR #421, opened 2020-09-28
Removed invalid self referencing type definitions

rsteinberger:
Daffodil-2196


mbeckerle (approved):
+1 Just removing these fixes the issue. Great.

tuxji (approved):
+1 with some questions

1.  Does anyone know what was the original reason for adding these simple XSD type definitions to the xs namespace in the XML Schema for DFDL and whether removing them affects that reason?  Was the reason to define a subset of simple xs: types in DFDL 1.0 while omitting the rest of the simple XSD types?  I don't think the DFDL 1.0 spec forbids any of the simple xs: type definitions defined by XSD anyway and it seems these definitions were actually unnecessary, but it would be nice if someone explained that in the bug report.

2. While searching for "simpleType", I see at lines 65-71 we have the following annotation:

```xsd
  <xsd:annotation>
    <xsd:documentation>
      The simpleType element and all of its members are defined
      towards the end
      of this schema document
    </xsd:documentation>
  </xsd:annotation>
```

Should this annotation be deleted or commented out?

3. While continuing to search for "simpleType", I see at lines 131-144 we have the following group:

```xsd
  <xsd:group name="redefinable">
    <xsd:annotation>
      <xsd:documentation>
        This group is for the
        elements which can self-redefine (see &lt;redefine> below).
      </xsd:documentation>
    </xsd:annotation>
    <xsd:choice>
      <xsd:element ref="xsd:simpleType" />
      <xsd:element ref="xsd:complexType" />
      <xsd:element ref="xsd:group" />
      <!-- <xsd:element ref="xsd:attributeGroup"/> -->
    </xsd:choice>
  </xsd:group>
```

The group is used in line 124, but the xsd:redefine element is commented out in line 238 and lines 1173-1189 so the redefinable group is superfluous except as a holdover from the original XML Schema.  I simply wonder if these extra type definitions had anything to do with making these types redefinable.

4. It looks like simpleType itself is defined at lines 1529-1535 in terms of topLevelSimpleType.

5. It looks like topLevelSimpleType is defined at lines 1490-1507 as a restriction of the original xsd:simpleType type from XSD.

mbeckerle:
This XSD for DFDL is used by Daffodil when it takes in a schema, to validate that it does not use any of the XSD features that are not in the DFDL subset of XSD. 

This was done by just commenting out specific lines of the original schema which is the w3c's XML Schema for XML Schema. 
That is, we didn't write this schema ultimately, we just adapted it. 

XML Schema is itself an XML-embedded language (for better or worse), so it's possible to write an XML Schema for it, but that's a meta-circular exercise that was tripping up XML validation in IDEs working with DFDL schemas. 

There is no-doubt other "dead code" in this schema. I think the goal is to continue to be able to diff this XML Schema for DFDL against the XML Schema for XML Schema so that if XML Schema evolves, we can identify and isolate the differences.

I think this schema does not enforce everything about the subset.

PR #423, opened 2020-09-29
Replace NoRep terminology with InputValueCalc

IanCarlsonOwl:
A simple text replacement to remove the outdated terminology "NoRep" with "InputValueCalc" which actually appears in the spec.

DAFFODIL-1533


tuxji (approved):
+1

I checked for all occurrences of "NoRep" in the codebase and yes, you've replaced all of them.

stevedlawrence (approved):
+1

PR #425, opened 2020-10-01
WIP: Review Request for Daffodil-1515

rsteinberger:
This implements the functionality expressed in the ticket for DFDLCheckRangeInclusive
following the pattern of checkConstraints.
TODO: Error handling, DFDLCheckRangeExclusive, assert testing

Daffodil-1515

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -1650,6 +1650,9 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
       case (RefQName(_, "checkConstraints", DFDL), args) => {
         DFDLCheckConstraintsExpr(functionQNameString, functionQName, args)
       }
+      case (RefQName(_, "checkRangeInclusive", DFDL), args) => {
+        DFDLCheckRangeInclusiveExpr(functionQNameString, functionQName, args)
+      }
       case (RefQName(_, "decodeDFDLEntities", DFDL), args) => {
         FNOneArgExpr(functionQNameString, functionQName, args,
           NodeInfo.String, NodeInfo.String, DFDLDecodeDFDLEntities(_, _))
@@ -2411,6 +2414,25 @@ case class DFDLCheckConstraintsExpr(nameAsParsed: String, fnQName: RefQName,
     val argDPath = args(0).compiledDPath
     val c = conversions
     val res = new CompiledDPath(DFDLCheckConstraints(argDPath) +: c)
+    res
+  }
+  override def targetTypeForSubexpression(subexpr: Expression): NodeInfo.Kind = NodeInfo.AnyAtomic
+
+  override lazy val inherentType = NodeInfo.Boolean
+
+}
+
+case class DFDLCheckRangeInclusiveExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expression])
+  extends FunctionCallBase(nameAsParsed, fnQName, args) {
+
+  override lazy val children = args
+
+  override lazy val compiledDPath = {
+    checkArgCount(3)
+    val compiledDPath = args(0).compiledDPath
+    val rangeFrom: Int = args(1).text.toInt
+    val rangeTo: Int = args(2).text.toInt
+    val res = new CompiledDPath(DFDLCheckRangeInclusive(compiledDPath, rangeFrom, rangeTo))
     res
   }
   override def targetTypeForSubexpression(subexpr: Expression): NodeInfo.Kind = NodeInfo.AnyAtomic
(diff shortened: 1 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2419,6 +2422,25 @@ case class DFDLCheckConstraintsExpr(nameAsParsed: String, fnQName: RefQName,
...
+
+  override lazy val compiledDPath = {
+    checkArgCount(3)
+    val compiledDPath = args(0).compiledDPath
+    val rangeFrom: Int = args(1).text.toInt
+    val rangeTo: Int = args(2).text.toInt
+    val res = new CompiledDPath(DFDLCheckRangeInclusive(compiledDPath, rangeFrom, rangeTo))
+    res
+  }
+  override def targetTypeForSubexpression(subexpr: Expression): NodeInfo.Kind = NodeInfo.AnyAtomic
```
stevedlawrence, line 2438: I think this is where you need to handle type information about the subexpressions. The targetType for the first paramater should probably be Numeric. The targetType for the min/max parameters should be the inherentType of the first parameter.

With this type information, Daffodil will insert the appropriate conversions (or error if not allowed) to convert the min/max parameters to the type of the node. This does mean the ``DFDLCheckRangeInclusive`` class will need to have logic to evaluate those subexpressions and handle different numeric types.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions.scala
```diff
@@ -39,6 +41,31 @@ case class DFDLCheckConstraints(recipe: CompiledDPath) extends RecipeOpWithSubRe
...
 }
 
+case class DFDLCheckRangeInclusive(recipe: CompiledDPath, rangeFrom: Int, rangeTo: Int)
+  extends RecipeOpWithSubRecipes(recipe) {
+
+  override def run(dstate: DState): Unit = {
+    recipe.run(dstate)
+    if (dstate.currentElement.valid.isDefined) {
+      dstate.setCurrentValue(dstate.currentElement.valid.get)
+    } else {
```
stevedlawrence, line 51: You shouldn't need to check or set the value of ``valid`` here. That variable just says whether the the value of ``currentElement`` is valid with respect to the restrictions defined in the schema, which is unrelated to checking the if the value of node is between some range.

Ultimately, this function probably wants to look something like the below so that we run a recipe for both the data value and min/max value. Note that we need to save the currentNode when running multiple expressions:
```scala
val saved = dstate.currentNode
dataRecipe.run(dstate)
val dataVal = dstate.currentValue
dstate.setCurrentNode(saved)
minRecipe.run(dstate)
val minVal = dstate.currentValue
dstate.setCurrentNode(saved)
maxRecipe.run(dstate)
val maxVal = dstate.currentValue
val res = executeCheck(dataVal, minVal, maxVal)
dstate.setCurrentValue(res)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLFunctions.scala
```diff
@@ -39,6 +41,31 @@ case class DFDLCheckConstraints(recipe: CompiledDPath) extends RecipeOpWithSubRe
...
+    }
+  }
+
+  def executeCheck(dstate: DState, rangeFrom: Integer, rangeTo: Integer): OKOrError = {
+    val testValue: BigInt = dstate.currentValue.getBigInt
+    if (testValue.intValue() >= rangeFrom.intValue() && testValue.intValue() <= rangeTo.intValue()) {
+      OKOrError.OK
+    } else {
+      OKOrError.Error("Error...")
+    }
```
stevedlawrence, line 64: This DFDLCheckRange functions don't need to return any error information. Either a value is within range or it's not. It isn't considered an error to not be in range. So this should just return a boolean and setCurrentValue just sets the returned value.

I would also suggest we make some effort to avoid BigInts. Firstly because it's going to be somewhat slow, but secondgly because the currentValue can be one of many different types, including those that aren't integers. This function should be able to statically determine the type and perform an appropriate range change based on that type.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -27,7 +27,48 @@
...
   defaultRoundTrip="true">
-  
+
+  <tdml:defineSchema name="DFDLCheckRange">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" />
+    <xs:element name="root1" dfdl:lengthKind="explicit" dfdl:length="1" type="xs:integer">
+      <xs:annotation>
+        <xs:appinfo source="http://www.ogf.org/dfdl/">
+          <dfdl:discriminator test="{ dfdl:checkRangeInclusive(., 0, 1) }" message="Discriminator failed ..." />
```
stevedlawrence, line 37: Although discrminators works, it adds extra complexity to the test that can sometimes make it difficult to ensure if the function is behaving correctly. I would suggest just using inputValueCalc, e.g.:

```xml
<xs:element name="isInRange" ... dfdl:inputValueCalc="{ dfdl:checkRangeInclusive(...) }" />
```

This way you can get actual true/false values in the expected infoset, rather than assuming a discriminator passing/failing implies the function worked as expected.

stevedlawrence (changes requested):

mbeckerle:
Closed as inactive. DAFFODIL-1515 ticket points back to this PR for future reference.

PR #426, opened 2020-10-03
Re-enable codecov.io pull request comments

stevedlawrence:
The default behavior for codecov.io is to create a comment for each PR
to describe how it changes codecov. We previously had this disabled, but
it makes this less visiable. Delete the codecov.io config file to
re-enable PR comments to encourage new pull requests to ensure they code
has test coverage.

DAFFODIL-2403

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,5 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Prevents codecov from creating a test coverage comment in a pull request
-comment: off
+coverage:
+  status:
+    project: off
+    patch: 
+      default:
+        target: auto
+
+comment: false


daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -73,6 +73,10 @@ trait SchemaSetRuntime1Mixin { self : SchemaSet =>
 
   def onPath(xpath: String): DFDL.DataProcessor = {
       Assert.usage(!isError)
+      if (false) {
+        println("test of code cov report")
+      }
```
tuxji, line 78: Seems like it's working to me.

tuxji (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
stevedlawrence: Yep, I guess I was expected an overall PR comment to be made (It used to do this), but maybe that's been disabled in place of inline comments, which I guess is fine.

One thought, PR #427 that you opened now fails because of the codecov bot being enabled. For some reason that PR decreases coverage on the overall project. Looking at the report, it's not clear what coverage is actually missing. I'm wondering if maybe it just comes down to the fact that the PR deleted a file that maybe had full coverage, and so be removing that causes less overall coverage? I'd be surprised if there was any real change in code paths between 2.11 and 2.12. But maybe?

I'm not sure if there's a good fix. We could disable ``codevoc/project`` check and only require that PR's have coverage of what they change, but it would be nice to be alerted if a PR removes tests or something that decreases overall coverage.

tuxji: The total coverage on the project in your PR's codecov report was only `80.21% (+0.01%) compared to d3b6350` and the total coverage on the project in my PR's codecov report went down to `78.29% (+-1.93%) compared to 82028c1`.

My PR indeed deletes one file (the file that was in the Scala 2.11-only src directory).  That deletion must have pushed the project's total coverage down below 80%.  

We could add some test cases to my PR to push the total coverage above 80% again.  It might even take only 1-2 test cases if we identify a class that had no test coverage to start with.  You're more familiar with the codecov UI and the codebase than I am.  Will you please identify a class that we could add a test to?

tuxji: If adding more tests becomes too onerous to do, then we also could reduce the 80% minimum coverage limit to 75%.  That limit's probably a parameter we could configure somewhere.  But let's hope adding more tests is as easy to do.

mbeckerle: I suggest create a JIRA ticket "Increase code test coverage" and move on for now.  Time spent on improving coverage should be done with an intention to add coverage where it is most needed, not just to get back to some artificial figure of merit that honestly, was never a target. It's just where we landed when we first turned on coverage testing.

stevedlawrence: This is causing all the PR's to appear as if they are failing. It seems even the slightest change can cause a decrease in project coverage. Since we aren't actively trying to maintain some certain percentage coverage, it doesn't really make sense to have this fail PRs. I'll add a config option to disable the codecov.project. I think it's still worth keeping codecov/patch as we should probably strive to have coverage for all new code.

stevedlawrence:
So there's the comment I was expecting, but code cov edited the comment everytime one of our 12 checks passes, which resulted in a bunch of spam to the commits list. Since this is now part of the check, I think and the inline warnings now appear on the actual code, I don't think we need these comments. I'll update the config to disable comments again.

mbeckerle (approved):
+1 I recently did some reviews whcih had codecov items right in the review, and that was quite useful. However, I'm approving as we can choose to reenable that if we want. There are other downsides (email volume)

tuxji:
The email volume was due to the CI workflow running 12 compilation checks with code coverage enabled in every check.  We just dropped support for Scala 2.11 so that's 6 compilation checks now.  We don't have many Java files (just API wrappers) and they rarely change.  If Scala provides uniform features and functionality no matter which underlying Java is being used (I don't know - does it?), do we really need separate compilation checks for Java 8, 9, and 11?  Then that would leave only Linux and Windows compilation checks (definitely worth keeping), and we could modify the CI workflow to enable code coverage in only Linux, not both Linux and Windows.

stevedlawrence:
I'd prefer to keep at least some of different checks. We probably don't need all 8, 9, and 11, but I think we need at least 8 for long term support, and at least the latest version (which we're behind, 13 has been out for a while, and I think I heard there's some failure 13). And we definitely need windows vs linux, we've had a handful of bugs that only appear on one OS. So I think the minimum we need is 4 (windows + linux + 2 java version). I suspect we'll also want to enable scala 2.13 relatively soon, so that will get us back to two scala versions. So I think it's going to hard to avoid excessive comments.

I don't personally find the above codecov bot report comment all that helpful. As long as the patch checks pass that feels reasonable. If the check fails, it provides a link to the report for the details that's in the codecov report. That feels sufficient to me. And I think we will still get the inline comments (which don't acttually create emails I think), it's just the code cov report comment that is disabled.

tuxji:
Sure, let's leave the codecov bot report commented out, although we still can modify the CI workflow if we want only specific compilation checks to run codecov.  I would be surprised if the codecov reports differ for Java 8, 9, and 11 although I would expect small differences for Linux and Windows.  

Speaking of Java versions, I don't think we need to worry about non-LTS releases anyway.  Java now has a fast channel, every six months, and a slow channel, every three years.  Each third-year release is called an LTS release.  Java 8 is the previous LTS release, Java 11 is the current LTS release, and 17 will be the next LTS release.  Unless there's a reason to test non-LTS releases, I'd chuck out Java 9 and test only Java 8, 11, and then 17 when it comes out.

stevedlawrence:
Yeah, we probably can drop 9. We should definitley keep 8 as a lot of production operating systems like RHEL are still on these older LTS versions of java. Though I'll leave that to separate discussion about what OS's and Java versions we should support with CI. I'll lmerge this as is so that pull request no longer fail due to the codecov/project check.

PR #427, opened 2020-10-03
Drop support for Scala 2.11

tuxji:
Scala 2.11 is end of life and there are no plans to make a new 2.11
release.  The last time a new Scala 2.11 release came out was 2.11.12
at the end of 2017.  We kept publishing Scala 2.11 builds because
Apache Spark did not start supporting Scala 2.12 until November 2018,
but the latest Spark 3.x releases support only 2.12 and have removed
support for 2.11.

A pull request in progress (#422) depends on an open source library
(os-lib) which has not published any new Scala 2.11 builds since March
2019.  We also may need to stop building on Scala 2.11 before we can
update some of our dependencies to newer major versions
(DAFFODIL-2269).

In .github/workflows/main.yml, drop 2.11 compile checks.  Also, no
need to install sbt since it's already installed on GitHub runners.

In build.sbt, publish only 2.12 builds.

In daffodil-lib/src/main/scala/org/apache/daffodil/util/MStack.scala,
and
daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ChoiceAndOtherVariousUnparsers.scala,
delete obsolete comment.

Remove
daffodil-runtime1/src/main/scala-2.11/org/apache/daffodil/infoset/DataValue.scala
and move
daffodil-runtime1/src/main/scala-2.12/org/apache/daffodil/infoset/DataValue.scala
to main scala directory (also delete obsolete comment).

DAFFODIL-2404


stevedlawrence (approved):
+1

mbeckerle (approved):
+1

PR #428, opened 2020-10-05
RFC: Evaluate suspensions earlier

stevedlawrence:
This is part 1 to reducing memory required during unparse. The goal here is to evaluate suspensions and thus reduce buffering while unparsing rather than waiting until the very end of unparse. Split into two commits to easy reviewing.

The first commit fixes an issue where a suspension could create another suspension, which breaks assumptions about how suspensions work. Please take a look and see if this is a reasonble way to resolve this issue. I don't believe there are any other cases where a suspension could create a suspension.

The second comment is what actually evalutes suspensions ealier, using a technique similar to garbage collection, maintaining a young and old list of suspensions and evaluating them at different times.


stevedlawrence:
Pushed wrong branch. Closing pull request.

PR #429, opened 2020-10-05
RFC: Evaluate suspensions earlier

stevedlawrence:
This is part 1 to reducing memory required during unparse. The goal here is to evaluate suspensions and thus reduce buffering while unparsing rather than waiting until the very end of unparse. Split into two commits to easy reviewing.

The first commit fixes an issue where a suspension could create another suspension, which breaks assumptions about how suspensions work. Please take a look and see if this is a reasonble way to resolve this issue. I don't believe there are any other cases where a suspension could create a suspension.

The second comment is what actually evalutes suspensions ealier, using a technique similar to garbage collection, maintaining a young and old list of suspensions and evaluating them at different times.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
@@ -24,6 +24,8 @@ import org.apache.daffodil.processors.parsers._
 import org.apache.daffodil.schema.annotation.props.SeparatorSuppressionPolicy
 import org.apache.daffodil.processors.unparsers._
 import org.apache.daffodil.util.Misc
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.MaybeInt
 
 /**
  * Base class for all kinds of sequences.
@@ -45,8 +47,27 @@ abstract class SequenceCombinator(sq: SequenceTermBase, sequenceChildren: Seq[Se
 class OrderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceChild])
   extends SequenceCombinator(sq, sequenceChildrenArg) {
 
+  private lazy val sepMtaGram = sq.sequenceSeparatorMTA
+  // Note that we actually only ever use one of these depending on
+  // various factors. If there is an optional separator and a suspension is
+  // used to unparse that separator, then we cannot use the sepMtaUnparser
+  // because it results in nested suspensions, which isn't allowed. In that
+  // case, the suspension ends up handling both the optional separator and
+  // alignment using sepMtaAlignmentMaybe.
+  private lazy val (sepMtaAlignmentMaybe, sepMtaUnparserMaybe) =
+    if (sepMtaGram.isEmpty) {
+      (MaybeInt.Nope, Maybe.Nope)
+    } else {
+      (MaybeInt(sq.knownEncodingAlignmentInBits), Maybe(sepMtaGram.unparser))
+    }
+
   private lazy val sepGram = sq.sequenceSeparator
-  private lazy val sepParser = sepGram.parser
+
+  private lazy val sepParser = (sepMtaGram ~ sepGram).parser
+  // we cannot include the mtaGram in the sepUnparser. This is because the
+  // sepUnparser is run in a suspension, and the mtaGram can result in
+  // suspension, which means nested suspensions. Instead, the unparser handles
+  // the mta parser differently to avoid this
   private lazy val sepUnparser = sepGram.unparser
 
   private lazy val sequenceChildren = sequenceChildrenArg.toVector
@@ -82,7 +103,11 @@ class OrderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceChi
       sq.hasSeparator match {
         case true => new OrderedSeparatedSequenceUnparser(
           srd,
-          sq.separatorSuppressionPolicy, sq.separatorPosition, sepUnparser,
+          sq.separatorSuppressionPolicy,
+          sq.separatorPosition,
+          sepMtaAlignmentMaybe,
+          sepMtaUnparserMaybe,
+          sepUnparser,
           childUnparsers)
         case false =>
           new OrderedUnseparatedSequenceUnparser(
@@ -98,8 +123,27 @@ class UnorderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceC
 
   import SeparatedSequenceChildBehavior._
 
+  private lazy val sepMtaGram = sq.delimMTA
+  // Note that we actually only ever use one of these depending on
+  // various factors. If there is an optional separator and a suspension is
+  // used to unparse that separator, then we cannot use the sepMtaUnparser
+  // because it results in nested suspensions, which isn't allowed. In that
+  // case, the suspension ends up handling both the optional separator and
+  // alignment using sepMtaAlignmentMaybe.
+  private lazy val (sepMtaAlignmentMaybe, sepMtaUnparserMaybe) =
+    if (sepMtaGram.isEmpty) {
+      (MaybeInt.Nope, Maybe.Nope)
+    } else {
+      (MaybeInt(sq.knownEncodingAlignmentInBits), Maybe(sepMtaGram.unparser))
+    }
+
   private lazy val sepGram = sq.sequenceSeparator
-  private lazy val sepParser = sepGram.parser
+
+  private lazy val sepParser = (sepMtaGram ~ sepGram).parser
+  // we cannot include the mtaGram in the sepUnparser. This is because the
+  // sepUnparser is run in a suspension, and the mtaGram can result in
+  // suspension, which means nested suspensions. Instead, the unparser handles
+  // the mta parser differently to avoid this
   private lazy val sepUnparser = sepGram.unparser
 
   private lazy val sequenceChildren = sequenceChildrenArg.toVector
@@ -173,7 +217,11 @@ class UnorderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceC
       sq.hasSeparator match {
         case true => new OrderedSeparatedSequenceUnparser(
           srd,
-          SeparatorSuppressionPolicy.AnyEmpty, sq.separatorPosition, sepUnparser,
+          SeparatorSuppressionPolicy.AnyEmpty,
+          sq.separatorPosition,
+          sepMtaAlignmentMaybe,
+          sepMtaUnparserMaybe,
+          sepUnparser,
           childUnparsers)
         case false =>
           new OrderedUnseparatedSequenceUnparser(
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
@@ -411,8 +411,24 @@ class DirectOrBufferedDataOutputStream private[io] (
     Assert.invariant(isDirect)
   }
 
+  /**
+   * We need to keep track of what FormatInfo was used as the last write to
+   * this DOS. This way, when this DOS becomes direct and we start
+   * delivering following buffered DOS's, we are using the correct FormatInfo
+   * to check for bitOrder changes and to write data correctly. We set this
+   * value when this DOS is marked as finished with setFinished() and then use
+   * it when delivering buffered content.
+   */
+  private var finishedFormatInfo: Maybe[FormatInfo] = Nope
+
   override def setFinished(finfo: FormatInfo): Unit = {
     Assert.usage(!isFinished)
+    Assert.usage(finishedFormatInfo.isEmpty)
+
+    // this DOS is finished, save this format info so we use it when delivering
+    // buffered content
+    finishedFormatInfo = One(finfo)
+
     // if we are direct, and there's a buffer following this one
     //
     // we know it isn't finished (because of flush() above)
@@ -433,9 +449,14 @@ class DirectOrBufferedDataOutputStream private[io] (
           first.setAbsStartingBitPos0b(dabp)
         }
 
-        DirectOrBufferedDataOutputStream.deliverBufferContent(directStream, first, finfo) // from first, into direct stream's buffers
+        // from first, into direct stream's buffers. Make sure we use the
+        // format info last used for this DOS.
+        DirectOrBufferedDataOutputStream.deliverBufferContent(
+          directStream,
+          first,
+          directStream.finishedFormatInfo.get)
+
         // so now the first one is an EMPTY not necessarily a finished buffered DOS
-        //
         first.convertToDirect(directStream) // first is now the direct stream
         directStream.setDOSState(Uninitialized) // old direct stream is now dead
         directStream = first // long live the new direct stream!
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
@@ -400,6 +400,32 @@
             </xs:documentation>
           </xs:annotation>
         </xs:element>
+        <xs:element name="unparseSuspensionWaitOld" type="xs:int" default="100" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              While unparsing, some unparse actions require "suspending" which
+              requires buffering unparse output until the suspension can be
+              evaluated. Daffodil periodically attempts to reevaluate these
+              suspensions so that these buffers can be released. We attempt to
+              evaluate young suspensions shortly after creation with the hope
+              that it will succeed and we can release associated buffers. But if
+              a young suspension fails it is moved to the old suspension list.
+              Old suspensions are evaluated less frequently since they are less
+              likely to succeeded. This minimizes the overhead related to
+              evaluating suspensions that are likely to fail. The
+              unparseSuspensionWaitYoung and unparseSuspensionWaitOld
+              values determine how many elements are unparsed before evaluating
+              young and old suspensions, respectively.
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element name="unparseSuspensionWaitYoung" type="xs:int" default="5" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              See unparseSuspensionWaitOld
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
       </xs:all>
     </xs:complexType>
   </xs:element>
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
@@ -24,6 +24,8 @@ import org.apache.daffodil.processors.ModelGroupRuntimeData
 import scala.collection.mutable.Buffer
 import SeparatorSuppressionPolicy._
 import SeparatorPosition._
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.MaybeInt
 
 trait Separated { self: SequenceChildUnparser =>
 
@@ -99,9 +101,11 @@ class OrderedSeparatedSequenceUnparser(
   rd: SequenceRuntimeData,
   ssp: SeparatorSuppressionPolicy,
   spos: SeparatorPosition,
+  sepMtaAlignmentMaybe: MaybeInt,
+  sepMtaUnparserMaybe: Maybe[Unparser],
   sep: Unparser,
   childUnparsersArg: Vector[SequenceChildUnparser])
-  extends OrderedSequenceUnparserBase(rd, childUnparsersArg :+ sep) {
+  extends OrderedSequenceUnparserBase(rd, (childUnparsersArg :+ sep) ++ sepMtaUnparserMaybe.toSeq) {
 
   private val childUnparsers =
     childUnparsersArg.asInstanceOf[Seq[SequenceChildUnparser with Separated]]
@@ -117,17 +121,18 @@ class OrderedSeparatedSequenceUnparser(
     if (trd.isRepresented) {
       spos match {
         case Prefix => {
-          sep.unparse1(state)
+          unparseJustSeparator(state)
           unparser.unparse1(state)
         }
         case Infix => {
-          if (state.groupPos > 1)
-            sep.unparse1(state)
+          if (state.groupPos > 1) {
+            unparseJustSeparator(state)
+          }
           unparser.unparse1(state)
         }
         case Postfix => {
           unparser.unparse1(state)
-          sep.unparse1(state)
+          unparseJustSeparator(state)
         }
       }
     } else {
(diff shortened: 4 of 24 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
```diff
@@ -98,8 +123,27 @@ class UnorderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceC
 
   import SeparatedSequenceChildBehavior._
 
+  private lazy val sepMtaGram = sq.delimMTA
+  // Note that we actually only every use one of these depending on depending
```
mbeckerle, line 127: typo "every" => "ever"

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -137,16 +142,21 @@ class OrderedSeparatedSequenceUnparser(
   }
 
   /**
-   * Unparses the separator only.
+   * Unparses just the separator, as well as any mandatory text alignment if it necessary
    *
    * Does not deals with infix boundary condition.
    */
   private def unparseJustSeparator(state: UState): Unit = {
+    if (sepMtaUnparserMaybe.isDefined) {
```
mbeckerle, line 150: This here is the crux of the change I think. You are laying down an alignment unparser for the separator, only if the optimizer has decided one could be needed, and it is laid down unconditionally.  Probably worth inserting comment to that affect here.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceUnparserBases.scala
```diff
@@ -38,5 +38,5 @@ abstract class OrderedSequenceUnparserBase(
   // Since some of the grammar terms might have folded away to EmptyGram,
   // the number of unparsers here may be different from the number of
   // children of the sequence group.
-  Assert.invariant(srd.groupMembers.length >= childUnparsers.length - 1) // minus 1 for the separator unparser
+  //Assert.invariant(srd.groupMembers.length >= childUnparsers.length - 1) // minus 1 for the separator unparser
```
mbeckerle, line 41: Delete. With the comment above. I think this is not worth checking. It's not obvious what the number of unparsers always is.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
```diff
@@ -64,6 +64,17 @@ sealed trait Parser
   protected def parse(pstate: PState): Unit
 
   final def parse1(pstate: PState) = {
+    if (!isInitialized) {
+      println("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
```
mbeckerle, line 68: debug code? remove.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
```diff
@@ -443,7 +443,7 @@ sealed abstract class LengthState(ie: DIElement)
 
   def isEndUndef = {
     val r = maybeEndPos0bInBits.isEmpty
-    if (r) Assert.invariant(maybeStartDataOutputStream.isEmpty)
+    if (r) Assert.invariant(maybeEndDataOutputStream.isEmpty)
```
mbeckerle, line 446: Is this just a bug fix? Was this being called?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
```diff
@@ -0,0 +1,114 @@
...
+   */
+  def requireFinal(): Unit = {
+    while (suspensionsYoung.nonEmpty) {
+      suspensionsOld.enqueue(suspensionsYoung.dequeue)
+    }
+
+    evalSuspensionQueue(suspensionsOld)
+
+    if (suspensionsOld.length == 1) {
+      Assert.invariantFailed("Single suspended expression making no forward progress. " + suspensionsOld(0))
```
mbeckerle, line 79: Actually if you change this to Assert.invariant(suspensionsOld.length != 1, "Single suspended expression ...") then that would fix the code-cov warning I'm seeing.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
```diff
@@ -98,8 +123,27 @@ class UnorderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceC
...
+  // on various factors. If there is an optional separtor and a suspension is
+  // used to unaprse that separator, then we cannot use the sepMtaUnaprser
+  // because it results in nested suspensions, which isn't allowed. In that
+  // case, the suspension ends up handling both the optional separator and
+  // alignment using sepMtaAlignmentMaybe.
+  private lazy val (sepMtaAlignmentMaybe, sepMtaUnparserMaybe) =
+    if (sepMtaGram.isEmpty) {
+      (MaybeInt.Nope, Maybe.Nope)
+    } else {
+      (MaybeInt(sq.knownEncodingAlignmentInBits), Maybe(sepMtaGram.unparser))
```
mbeckerle, line 137: I think we need a test where the separator requires alignment, but ZL happens and the alignment and separator are both suppressed. A BCD number odd number of nibbles long, separated by DEL (0x7F) characters would do it.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -155,7 +165,7 @@ class OrderedSeparatedSequenceUnparser(
     state: UState,
     trailingSuspendedOps: Buffer[SuppressableSeparatorUnparserSuspendableOperation]): Unit = {
 
-    val suspendableOp = new SuppressableSeparatorUnparserSuspendableOperation(sep, trd)
+    val suspendableOp = new SuppressableSeparatorUnparserSuspendableOperation(sepMtaAlignmentMaybe, sep, trd)
```
mbeckerle, line 168: I'm seeing a coverage warning on this line also. Definitely need to see this exercised to be sure the SupressableSeparatorUnparserSuspendableOperation is being tested.  Not sure why it wouldn't have been.

mbeckerle (changes requested):
Tiny little changes needed. All in all, I am impressed how little code this took to get this working.

The CodeCov warnings I'm seeing here may or may not be meanngful, but they point at at least one important place in the code not being tested that I think needs a test.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
mbeckerle: Except,... how we get it to be zero length.... That I'm not sure.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
stevedlawrence: Yeah, I'm pretty sure this line is hit, otherwise some tests would fail. I'll look into what's going on.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
stevedlawrence: I think this is just a bug that got triggered by evaluating suspensions earlier. If the end position is empty, thent he end output stream must also be empty, not the start output stream. I suspect this was just a copy/paste error from isStartUndef.

stevedlawrence: To your second question, I think it was being called, it just always happend that the start output strema was empty if the end pos was empty due to how suspensions were evaluated.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
stevedlawrence: We don't have an invariant() function that accepts a string. I'll see if the invariant macro can be changed to allow a message that just defaults to empty string.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/Parser.scala
stevedlawrence: Ooops, yep.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
stevedlawrence: Turns out macros aren't allowed to have default arguments. But function overloading is allowed, and it's easy enough to add a new macro that accepts 2 parameters. Change has been made. This should fix this line of coverage.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.daffodil.dsom.RuntimeSchemaDefinitionError
+import org.apache.daffodil.exceptions.Assert
+
+class SuspensionTracker {
+
+  private val suspensionsYoung = new Queue[Suspension]
+  private val suspensionsOld = new Queue[Suspension]
+
+  private val suspensionWaitYoung = 5
+  private val suspensionWaitOld = 100
```
stevedlawrence, line 31: I meant to make these values tunable. It's not obvious what good values for these should be, so being able to easily tweak them could be useful.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
stevedlawrence: I have confirmed that this is in fact not hit by any tests (these new codecov warnings are nice!). The new logic for alignment is tested because the ``unparseOneWithSupression`` function also creates this same suspension.

After grepping, this ``unparseJustSeparatorWithTrailingSuppression`` function is actually just never called anywhere. I suspect this is just dead code and ``unparseOneWithSuppression`` handles all cases of suppressed separators, including the trailing suppression case? Any idea of this can just be removed, or if we're perhaps missing logic somewhere else and this function should be used?

mbeckerle: Yow. Well, the purpose of this must be for unparsing an empty element that is not trailing, so not trailing suppressed. There are some bugs about separatorSuppressionPolicy="never" which this code might be in service of. DAFFODIL-2219 is one. 

Put a nice big comment in here with a FIXME: possibly dead code, but maybe needed to fix separatorSuppressionPolicy bugs like DAFFODIL-2219

mbeckerle: Probably create a JIRA ticket about this dead code as well.

mbeckerle: To avoid the coverage thing, I suppose this code must be commented out. I'm just reluctant to delete it given the open bugs in this area.

stevedlawrence: Fair, I'll comment it out and create an issue + comment so we don't lose this if it's needed.

stevedlawrence: Fixed in latest fixup commit

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
stevedlawrence: I've confirmed that we do not have any tests that trigger the situation where we have a optional separator with MTA and a zero length field.

Maybe we can have string fields that are a non-byte size encoding, but separators with byte size encoding? That would require MTA after unparsing a field an before unparsing the separator. And we could have a field that is the empty string, which should be zero length?

mbeckerle: That is probably the best and easiest way to exercise this functionality.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
```diff
@@ -45,8 +47,27 @@ abstract class SequenceCombinator(sq: SequenceTermBase, sequenceChildren: Seq[Se
...
   extends SequenceCombinator(sq, sequenceChildrenArg) {
 
+  private lazy val sepMtaGram = sq.sequenceSeparatorMTA
+  // Note that we actually only every use one of these depending on depending
+  // on various factors. If there is an optional separtor and a suspension is
+  // used to unaprse that separator, then we cannot use the sepMtaUnaprser
+  // because it results in nested suspensions, which isn't allowed. In that
+  // case, the suspension ends up handling both the optional separator and
+  // alignment using sepMtaAlignmentMaybe.
+  private lazy val (sepMtaAlignmentMaybe, sepMtaUnparserMaybe) =
```
tuxji, line 57: Reads better as:

```
  // Note that we actually only ever use one of these depending on 
  // various factors. If there is an optional separator and a suspension is
  // used to unparse that separator, then we cannot use the sepMtaUnparser
  // because it results in nested suspensions, which isn't allowed. In that
  // case, the suspension ends up handling both the optional separator and
  // alignment using sepMtaAlignmentMaybe.
```

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
```diff
@@ -98,8 +123,27 @@ class UnorderedSequence(sq: SequenceTermBase, sequenceChildrenArg: Seq[SequenceC
...
   import SeparatedSequenceChildBehavior._
 
+  private lazy val sepMtaGram = sq.delimMTA
+  // Note that we actually only ever use one of these depending on depending
+  // on various factors. If there is an optional separtor and a suspension is
+  // used to unaprse that separator, then we cannot use the sepMtaUnaprser
+  // because it results in nested suspensions, which isn't allowed. In that
+  // case, the suspension ends up handling both the optional separator and
+  // alignment using sepMtaAlignmentMaybe.
+  private lazy val (sepMtaAlignmentMaybe, sepMtaUnparserMaybe) =
```
tuxji, line 133: Same here, reads better as:

```
  // Note that we actually only ever use one of these depending on 
  // various factors. If there is an optional separator and a suspension is
  // used to unparse that separator, then we cannot use the sepMtaUnparser
  // because it results in nested suspensions, which isn't allowed. In that
  // case, the suspension ends up handling both the optional separator and
  // alignment using sepMtaAlignmentMaybe.
```

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
tuxji: Funny, codecov check seems to say 100% of diff is covered.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -400,6 +400,32 @@
...
+              evaluated. Daffodil periodically attempts to reevaluate these
+              suspensions so that these buffers can be released. We attempt to
+              evaluate young suspensions shortly after creation with the hope
+              that it will succeed and we can release associated buffers. But if
+              a young suspension fails it is moved to the old suspension list.
+              Old suspensions are evaluated less frequently since they are less
+              likely to succeeded. This minimizes the overhead related to
+              evaluating suspensions that are likely to fail. The
+              unparseSuspensionWaitYoung and unparseSuspensionWaitOld
+              values determine how many elements unparsed before evaluating
```
tuxji, line 417: Reads better as

`values determine how many elements are unparsed before evaluating`

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -137,31 +142,56 @@ class OrderedSeparatedSequenceUnparser(
   }
 
   /**
-   * Unparses the separator only.
+   * Unparses just the separator, as well as any mandatory text alignment if it necessary
```
tuxji, line 145: Reads better: `if necessary`

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
```diff
@@ -182,10 +212,15 @@ class OrderedSeparatedSequenceUnparser(
...
-    // were ZL or not.
+    // We don't know if the unparse will result in zero length or not. We have
+    // to use a suspendable unparser here for the separator which suspends
+    // until it is known whether the unparse of the contents were ZL or not. If
+    // the suspension determines that the field is non-zero length then the
+    // suspenion must also unparser mandatory text alignment for the separator.
+    // This cannot be done with a standard MTA alignment unparser since that is
+    // a suspension and suspensions cannot create suspensions. This this
+    // suspension is also responsible for unparsing alignment if the separator
+    // should be unparsed.
```
tuxji, line 223: Reads better as

```
    // We don't know if the unparse will result in zero length or not. We have
    // to use a suspendable unparser here for the separator which suspends
    // until it is known whether the unparse of the contents were ZL or not. If
    // the suspension determines that the field is non-zero length then the
    // suspension must also unparse mandatory text alignment for the separator.
    // This cannot be done with a standard MTA alignment unparser since that is
    // a suspension and suspensions cannot create suspensions. This suspension
    // is also responsible for unparsing alignment if the separator should be
    // unparsed.
```

tuxji (approved):
+1

Looks practically ready for merging.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceCombinator.scala
stevedlawrence: I think this is a case where code coverage isn't really an accurae representation of test coverage. For example, you could have something like
```scala
if (var1) {
  doTask1()
}

if (var2) {
  doTask2()
}
```
If you have one test where ``var1 == true`` and another test where ``var2 == true``, we'll get 100% line coverage. But we won't get test coverage where both ``var1 == true`` and ``var2 == true``. I think that's what's going on here.

This is partly why we need to rely so heavily on functional tests to ensure correctness--we have so much code that does one individual thing clearly correctly (e.g. a single parser that just parses an int or aligns the bit position) but when combined can lead to incorrect results.

stevedlawrence: I've added a test that ensures that when an optional separator is determined not to be needed due to zero length data, we correctly skip unparsing the mta alignment.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SeparatedSequenceUnparsers.scala
stevedlawrence: Thanks for the sugggestions! All have been updated.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -400,6 +400,32 @@
             </xs:documentation>
           </xs:annotation>
         </xs:element>
+        <xs:element name="unparseSuspensionWaitOld" type="xs:int" default="100" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
```
mbeckerle, line 405: I like to put <![CDATA[... ]]> around blocks of doc text like this. Seems to tell tooling not to mess up the line indentation and such.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Attempts to evaluate suspensions. Old suspensions are evaluated less
+   * frequently than young suspensions. Any young suspensions that fail to
+   * evaluate are moved to the old suspensions list. If we evaluate old
+   * suspensions, we attempt to evaluate them first, with the hope that their
+   * resolution might make the young suspensions more likely to evaluate.
+   */
+  def evalSuspensions(): Unit = {
```
mbeckerle, line 43: Besides absolute measures of performance I think we should keep a count of how many times a suspension is evaluated only to be put back on a queue. A useful efficiency measure. We might want a tunable to turn on/off reporting this at end of a test.

mbeckerle (approved):
+1

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
stevedlawrence: We don't have cdata tags documentation for the other tunables. Should I add it for all of them as well? Indentation also doesn't really matter for these, since it's not test related. Do tools even modify xs:documentation tags? I would expect tools wouldn't even modify those since it feels important to not modify documentation tags--tools can't know if format is important for them.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
stevedlawrence: Good idea! Will add.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
mbeckerle: I think tools treat this as just plain old XML. They don't know that it is a specific XML-schema tag. So whitespace is fluid. 

I don't think adding more documentation blocks is necessary for this. I do like using the documentation annotations for block  comments. I think of it like javadoc, sort of.

stevedlawrence: Yeah, I gues many XML tools aren't schema aware.

I'm not suggesting more documentation blocks, I'm wondering if we want to add CDATA to all the other documentation blocks in this file for all the other tunables. We don't currently have them.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SuspensionTracker.scala
stevedlawrence: I've just pushed a change to log stats and the LogLevel.Debug level. I figure this is only really needed when debugging so probably not worth adding a new tunable.

This also found an interesting bug in the loger macro which fortunately had a pretty simple fix. Issue was that something like this didn't work:
```scala
log(LogLevel.Debug, "int: %d, float: %f", 1, 3.0)
```
The issue was that logger macro wrapped the var args in a Seq, e.g. ``Seq(1, 3.0)``. Scala tries to be helpful and casts this to a ``Seq[Double](1.0, 3.0)``, but then the arg types don't match the types specified in the printf style message. The solution is to have the macro wrap the var args in a ``Seq[Any]`` which prevents that casting. I added test to make sure this doesn't break in the future.

stevedlawrence:
I did some regression testing on a bunch of schemas and found that this causes a few schemas to fail. Needs further investigation to figure out what's going on, but I'll hold off on merging this until I figure it out.

stevedlawrence:
I found and fix one issue, I think it's an existing issue that we just never hit. I can't even figure out a way to hit it, and wasn't able to simplify schemas/data to get an easily reproducable test. However, many private schemas now test, and I think visually this code is confirmed to be wrong. See commit message in 6fe93c478d for a description of the issue.

Looks like there is still one more failure in a private schema to fix.

stevedlawrence:
Found the last issue, it actually wasn't really related to suspensions, but related to how the order of suspensions getting evaluated could change how we deliver buffered DOS to the direct DOS. Again, I'm struggling to create a simple test for this, but I think the code is fairly clear.

The issue here was that when we call setFinished on a DOS we deliver all the following buffered DOS's that have been finished so that we can release finished buffers. However, when we deliver multiple buffers, we always use the same ``FormatInfo`` as the original direct DOS. But Buffered DOS can change bitOrder, so the FormatInfo (which is where we get bitOrder) needs to change depending on the DOS. The fix is to save the FormatInfo associated with a buffered DOS when it is finished, and then use that when that DOS delivers buffered content. This allows the FormatInfo to change while delivering buffered content as we deliver different DOS.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -411,8 +411,24 @@ class DirectOrBufferedDataOutputStream private[io] (
     Assert.invariant(isDirect)
   }
 
+  /**
+   * We need to keep track of what FormatInfo was used as the last write to
+   * this DOS. This way, when we this DOS becomes direct and we start
```
tuxji, line 416: Drop extraneous "we", should be "when this DOS".

tuxji (approved):
+1 from me too

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
stevedlawrence: Fixed. Thanks!

PR #430, opened 2020-10-05
WIP: Changeover most DFDLTestSuite invocations to Runner

IanCarlsonOwl:
Tweak the interface to the Runner object, as well as
creating a RunnerOpts object to handle the multiple
combinations of parameters used by various tests without
compromising the ability to default parameters
not under test.

DAFFODIL-1300

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
@@ -19,6 +19,69 @@ package org.apache.daffodil.tdml
 
 import org.apache.daffodil.util.Misc
 import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.processors.HasSetDebugger
+import scala.xml.Node
+
+/**
+ * A wrapper to contain the optional parameters that can be used by the Runner
+ * class, so that they don't need to be redefined in multiple constructors,
+ * and can be easily defaulted by calling functions
+ */
+
+object RunnerOpts {
+  def apply(validateTDMLFile: Boolean = true,
+    validateDFDLSchemas: Boolean = true,
+    compileAllTopLevel: Boolean = false,
+    defaultRoundTripDefault: RoundTrip = RunnerOpts.defaultRoundTripDefaultDefault,
+    defaultValidationDefault: String = RunnerOpts.defaultValidationDefaultDefault,
+    defaultImplementationsDefault: Seq[String] = RunnerOpts.defaultImplementationsDefaultDefault,
+    shouldDoErrorComparisonOnCrossTests: Boolean = RunnerOpts.defaultShouldDoErrorComparisonOnCrossTests,
+    shouldDoWarningComparisonOnCrossTests: Boolean = RunnerOpts.defaultShouldDoWarningComparisonOnCrossTests): RunnerOpts = 
+      new RunnerOpts(validateTDMLFile, validateDFDLSchemas, compileAllTopLevel, 
+        defaultRoundTripDefault, defaultValidationDefault, defaultImplementationsDefault, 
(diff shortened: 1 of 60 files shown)


IanCarlsonOwl:
There are a number of calls to DFDLTestSuite still outstanding, but I wanted to get this much in front of others for a sanity-check, and to ensure it builds on all of our target systems.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
```diff
@@ -21,69 +21,94 @@ import org.apache.daffodil.util.Misc
...
+
+  def apply(dir: String, file: String, options: RunnerOpts): Runner =
+    new Runner(dir, file, options)
+
+  def apply(elem: scala.xml.Elem): Runner =
+    new Runner(elem, RunnerOpts())
+
+  def apply(elem: scala.xml.Elem, options: RunnerOpts): Runner =
+    new Runner(elem, options)
 }
```
stevedlawrence, line 112: I'm concerned that this could break existing Runners not in the daffodil repo. Other schema repos use this ``Runner`` for testing and they could currently provide options which would no longer compile once this change is made. I'm wondering if it makes sense to keep an apply the old dir/file apply method with all options defaulted and creates the new ``RunnerOpts`` using those values? We can deprecate that ``apply`` method so other repos will still work but get a message to fix them?

daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
```diff
@@ -34,69 +33,72 @@ object TresysTests {
...
+
+  lazy val runnerBC = Runner(testDir, "BC.tdml")
+
+  lazy val runnerBD = Runner(testDir, "BD.tdml")
 
   @AfterClass def shutDown: Unit = {
     runnerDelimited.reset
     runnerMD.reset
     runnerMD_NV.reset
+    runnerBB.reset
```
stevedlawrence, line 79: Since we are modifying so many test files do we also want to add the ``shutDown`` method to reset the runners? I think many of them are missing.

I'm wondering if a convenient way to do this would be to add a new mixin (i.e. DaffodilTestSuite) that implements shutDown and uses reflection to find any Runner members and automatically call reset on them?

stevedlawrence (commented):
Seems like a reasonable approach to me :+1:

tuxji (commented):
I agree with Steve's suggestions, especially the one to define a trait DFDLTestSuite with an @AfterClass shutDown method which XXXTests objects can extend to automatically call reset() on any member fields matching the type DFDLTestSuite.  Overall, your approach looks good.

daffodil-test-ibm1/src/test/scala/org/apache/daffodil/TresysTests.scala
mbeckerle: I would call this mixin AllRunnersAutoShutdownMixin.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
mbeckerle: I would second this. We need the original Runner factory method, as there are dozens of DFDL schemas we know of, and possibly others we don't know of or have access to, which are using the existing Runner factory.

IanCarlsonOwl: In commit b123d9f0ed174df22f87b564dbe38cab853e8362 I restored the old apply methods.

mbeckerle:
Closed as inactive. Ticket DAFFODIL-1300 is closed. If there is other functionality here being worked on a new JIRA ticket is needed.

PR #432, opened 2020-10-13
Remove scalac options for 2.11

jw3:
This can go away since #427

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -151,10 +151,6 @@ lazy val commonSettings = Seq(
 
 def scalacCrossOptions(scalaVersion: String) =
   CrossVersion.partialVersion(scalaVersion) match {
-    case Some((2, 11)) => Seq(
-      "-language:existentials",
-      "-Ywarn-unused-import"
-    )
     case Some((2, 12)) => Seq(
       "-Ywarn-unused:imports"
     )


build.sbt
```diff
@@ -151,10 +151,6 @@ lazy val commonSettings = Seq(
 
 def scalacCrossOptions(scalaVersion: String) =
```
jw3, line 152: Could strip out the entire function for clarity, or leave it for future convenience.

stevedlawrence: I think it's worth leaving this in and just removing the 2.11 case. I suspect we'll want to add support for 2.13 at some point and it wouldn't surprise me if scalc options changed again.

stevedlawrence (approved):
+1 :+1: Thanks for the contribution!

Only minor suggestion is to include a bug number in the commit message. I think DAFFODIL-2404 is reasonable since this is essentially an update to that bug.

build.sbt
jw3: I was leaning that way too :+1:

jw3:
> include a bug number in the commit message. I think DAFFODIL-2404

done

tuxji (approved):
+1 from me too

I suspect -Ywarn-unused:imports will move to scalacOptions if 2.13 supports it as well, but we may need a 2.13 case for options specific to 2.13 anyway.

PR #433, opened 2020-10-13
Alert user to use blobs instead of hexBinary

jadams-tresys:
The test files provided for this bug were bumping into the limits of
integer values, which we cannot create an array that is larger than
Int.MaxValue. In cases such as this, blobs should be used instead. Given
that this error is originating from the official NITF schema, we should
update the NITF schema to take advantage of blobs for large payloads.

DAFFODIL-2401

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
@@ -33,13 +33,16 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
   override final def parse(start: PState): Unit = {
     val dis = start.dataInputStream
     val currentElement = start.simpleElement
-    val nBits = getLengthInBits(start).toInt
+    val nBits = getLengthInBits(start)
     if (nBits == 0) {
       currentElement.setDataValue(zeroLengthArray)
+    } else if (nBits > Int.MaxValue) {
+      // Integer overflow will occurr, recommend using blob instead of hexBinary
+      PE(start, "Data is too large for xs:hexBinary. Consider using blobs instead: https://s.apache.org/daffodil-blob-feature")
     } else if (!dis.isDefinedForLength(nBits)) {
       PENotEnoughBits(start, nBits, dis.remainingBits)
     } else {
-      val array =  start.dataInputStream.getByteArray(nBits, start)
+      val array =  start.dataInputStream.getByteArray(nBits.toInt, start)
       currentElement.setDataValue(array)
     }
   }
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
@@ -521,7 +521,6 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
 
   def skip(nBits: Long, finfo: FormatInfo): Boolean = {
     // threadCheck()
-    Assert.usage(nBits <= Int.MaxValue)
     if (!this.isDefinedForLength(nBits)) return false
     setBitPos0b(bitPos0b + nBits)
     true
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BlobLengthParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/BlobLengthParsers.scala
@@ -47,6 +47,8 @@ sealed abstract class BlobLengthParser(override val context: ElementRuntimeData)
 
     val array = new Array[Byte](start.tunable.blobChunkSizeInBytes)
     val blobChunkSizeInBits = start.tunable.blobChunkSizeInBytes * 8
+    if (blobChunkSizeInBits > Int.MaxValue)
+      start.SDE("blobChunkSizeInBytes is too large. blobChunkSizeInBytes must be less than " + (Int.MaxValue / 8))
 
     while (remainingBitsToGet > 0) {
       val bitsToGet = Math.min(remainingBitsToGet, blobChunkSizeInBits).toInt


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
```diff
@@ -36,6 +36,9 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
     val nBits = getLengthInBits(start).toInt
     if (nBits == 0) {
       currentElement.setDataValue(zeroLengthArray)
+    } else if (nBits < 0) {
+      // Integer overflow occurred, recommend using blob instead of hexBinary
```
mbeckerle, line 40: We're depending on 32-bit integer overflow? That's very unreliable. We should do math on Long type, and check the bound, and explicitly return -1 to indicate too big. 

Or, if we are doing this right and signaling the overflow of the max length by -1 then the comment should change.

mbeckerle (changes requested):
Need to fix something around this integer overflow thing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
```diff
@@ -36,6 +36,9 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
...
+      PE(start, "Data is too large for xs:hexBinary. Consider using blobs instead: https://cwiki.apache.org/confluence/display/DAFFODIL/Proposal%3A+Binary+Large+Objects")
```
stevedlawrence, line 41: I'm a little hesitant to include this link, only because it's very possible this link might change in the future.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
```diff
@@ -90,7 +90,9 @@ sealed abstract class SpecifiedLengthParserBase(
     Assert.invariant(dis eq pState.dataInputStream)
     val bitsToSkip = finalEndPos0b - dis.bitPos0b
     Assert.invariant(bitsToSkip >= 0) // if this is < 0, then the parsing of children went past the limit, which it isn't supposed to.
-    if (bitsToSkip > 0) {
+    if (bitsToSkip > Int.MaxValue) {
+      PE(pState, "Data too large, consider using blobs instead: https://cwiki.apache.org/confluence/display/DAFFODIL/Proposal%3A+Binary+Large+Objects")
+    } else if (bitsToSkip > 0) {
```
stevedlawrence, line 95: I'm concerned there are other places we might need to deal with these kinds of overflow issues?

For example, all the binary number parsers have a getBitLength funciton that returns an int. I would never expect a binary number to have such a large length that it overflows, but if there were bad data or a bad dfdl:length expression, it could definiately happen. If we're lucky it overflows to a negative number and hopefully we get an error message, but if it overflows to some positive number we're just going to get unexpected behavior.

I'm not sure of a good way to handle this cleanly, since we deal with lengths all over the place. Just grepping for getBitLength shows something lke 50 different implementations across different parsers and unparsers. I suspect we're not handling overflow properly in many of them. And getLengthInBits shows up as well. There are probably other length calcualtions that don't use either of these.

jadams-tresys: Hm, would a better fix be to modify the various get functions to use Longs, but then throw a ParseError inside the get functions if the value ends up being larger than Int.MaxValue?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
jadams-tresys: I understand the hesitancy, but I figured that most people wouldn't be familiar with blobs and thought that the link might be the best way to point them in the right direction.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
jadams-tresys: I pushed up changes that I think cover all the getByteArray issues. As for the binaryNumberParsers, I think it might be worth making its own ticket that can be dealt with later as it is a much less likely situation to occur.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
jadams-tresys: This has been addressed in my most recent commit

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
mbeckerle: If you want to include a link, we should create an Apache short link that can be kept up to date. 
I just created: https://s.apache.org/daffodil-blob-feature for this if we want to keep the link.

mbeckerle (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SpecifiedLengthParsers.scala
stevedlawrence: Yeah, agreed this is too much to fix for now. I think we really do need to come up with a comprehensive way to ensure we dont' have any overflow issues, but it's fine to create a new bug and handle that more completely later.

stevedlawrence (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/HexBinaryLengthParsers.scala
stevedlawrence: Agreed, this would be a better link.

jadams-tresys: Done

PR #436, opened 2020-10-14
SAX Unparse Implementation

olabusayoT:
- add legacy coroutine class with 1 event ABQ
- add DaffodilInputContentHandler to collect SAXEvents from parsed XML
and convert to InfosetEvent for the SAXInfosetOutputter and the
dp.unparse
- add SAX unparsing to TDML
- Add errorHandler to unparse
- add features for namespace to xmlReader
- add DaffodilUnparserSAXException abnd DaffodilUnhandledSAXException case classes and error handling for errors during unparse
- add error info to events for communication b/w handler and inputter
- rename inputterEvent to infosetEvent
- add uri resolving for SAX simpleText (useful for TDML)
- some refactoring to isolate coroutine switching w/i contentHandler
- added comments for clarity
- change isNilled to nilValue so we can account for non-boolean values
- if getSimpleText has no value (i.e None), that is an error. Some(null)
or Some("") is indicative of an empty element
- add error handling so errors don't cause hanging

DAFFODIL-2383

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -38,6 +38,7 @@ import scala.xml.Node
 import scala.xml.SAXParseException
 
 import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.parsers.SAXParserFactory
 import javax.xml.transform.TransformerFactory
 import javax.xml.transform.dom.DOMSource
 import javax.xml.transform.stream.StreamResult
@@ -46,7 +47,9 @@ import org.apache.commons.io.IOUtils
 import org.apache.commons.io.output.NullOutputStream
 
 import org.apache.daffodil.api.DFDL
+import org.apache.daffodil.api.DFDL.DaffodilUnparseErrorSAXException
 import org.apache.daffodil.api.DFDL.ParseResult
+import org.apache.daffodil.api.DFDL.UnparseResult
 import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.api.URISchemaSource
 import org.apache.daffodil.api.ValidationMode
@@ -64,7 +67,7 @@ import org.apache.daffodil.exceptions.UnsuppressableException
 import org.apache.daffodil.externalvars.Binding
 import org.apache.daffodil.externalvars.BindingException
 import org.apache.daffodil.externalvars.ExternalVariablesLoader
-import org.apache.daffodil.infoset.DaffodilOutputContentHandler
+import org.apache.daffodil.infoset.DaffodilParseOutputStreamContentHandler
 import org.apache.daffodil.infoset.InfosetInputter
 import org.apache.daffodil.infoset.InfosetOutputter
 import org.apache.daffodil.infoset.JDOMInfosetInputter
@@ -408,7 +411,6 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
       case (Some("json"), _) => Right(Unit)
       case (Some("jdom"), _) => Right(Unit)
       case (Some("w3cdom"), _) => Right(Unit)
-      case (Some("sax"), Some(true)) => Left("SAX unparse is not yet implemented")
       case (Some("sax"), _) => Right(Unit)
       case (Some("null"), Some(true)) => Left("infoset type null not valid with performance --unparse")
       case (Some("null"), _) => Right(Unit)
@@ -443,7 +445,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
     val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "variables to be used when unparsing. An optional namespace may be provided.")
     val tunables = props[String]('T', keyName = "tunable", valueName = "value", descr = "daffodil tunable to be used when parsing.")
     val config = opt[String](short = 'c', argName = "file", descr = "path to file containing configuration items.")
-    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to unparse. Must be one of 'xml', 'scala-xml', 'json', 'jdom', or 'w3cdom'.", default = Some("xml")).map { _.toLowerCase }
+    val infosetType = opt[String](short = 'I', argName = "infoset_type", descr = "infoset type to unparse. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom' or 'sax'.", default = Some("xml")).map { _.toLowerCase }
     val stream = toggle(noshort = true, default = Some(false), descrYes = "split the input data on the NUL character, and unparse each chuck separately", descrNo = "treat the entire input data as one infoset")
     val infile = trailArg[String](required = false, descr = "input file to unparse. If not specified, or a value of -, reads from stdin.")
 
@@ -718,14 +720,14 @@ object Main extends Logging {
   val blobSuffix = ".bin"
 
   def getInfosetOutputter(infosetType: String, os: java.io.OutputStream)
-  : Either[InfosetOutputter, DaffodilOutputContentHandler] = {
+  : Either[InfosetOutputter, DaffodilParseOutputStreamContentHandler] = {
     val outputter = infosetType match {
       case "xml" => Left(new XMLTextInfosetOutputter(os, pretty = true))
       case "scala-xml" => Left(new ScalaXMLInfosetOutputter())
       case "json" => Left(new JsonInfosetOutputter(os, pretty = true))
       case "jdom" => Left(new JDOMInfosetOutputter())
       case "w3cdom" => Left(new W3CDOMInfosetOutputter())
-      case "sax" => Right(new DaffodilOutputContentHandler(os, pretty = true))
+      case "sax" => Right(new DaffodilParseOutputStreamContentHandler(os, pretty = true))
       case "null" => Left(new NullInfosetOutputter())
     }
     if (outputter.isLeft) {
@@ -801,34 +803,46 @@ object Main extends Logging {
           }
         }
       }
+      case "sax" => data match {
+        case Left(bytes) => bytes
+        case Right(is) => is
+      }
     }
   }
 
-  def getInfosetInputter(infosetType: String, anyRef: AnyRef): InfosetInputter = {
+  def getInfosetInputter(
+    infosetType: String,
+    anyRef: AnyRef,
+    processor: DFDL.DataProcessor,
+    outChannel: DFDL.Output): Either[InfosetInputter, DFDL.DaffodilUnparseContentHandler] = {
     infosetType match {
       case "xml" => {
         val is = anyRef match {
           case bytes: Array[Byte] => new ByteArrayInputStream(bytes)
           case is: InputStream => is
         }
-        new XMLTextInfosetInputter(is)
+        Left(new XMLTextInfosetInputter(is))
       }
       case "json" => {
         val is = anyRef match {
           case bytes: Array[Byte] => new ByteArrayInputStream(bytes)
           case is: InputStream => is
         }
-        new JsonInfosetInputter(is)
+        Left(new JsonInfosetInputter(is))
       }
       case "scala-xml" => {
-        new ScalaXMLInfosetInputter(anyRef.asInstanceOf[scala.xml.Node])
+        Left(new ScalaXMLInfosetInputter(anyRef.asInstanceOf[scala.xml.Node]))
       }
       case "jdom" => {
-        new JDOMInfosetInputter(anyRef.asInstanceOf[org.jdom2.Document])
+        Left(new JDOMInfosetInputter(anyRef.asInstanceOf[org.jdom2.Document]))
       }
       case "w3cdom" => {
         val tl = anyRef.asInstanceOf[ThreadLocal[org.w3c.dom.Document]]
-        new W3CDOMInfosetInputter(tl.get)
+        Left(new W3CDOMInfosetInputter(tl.get))
+      }
+      case "sax" => {
+        val dp = processor
+        Right(dp.newContentHandlerInstance(outChannel))
       }
     }
   }
@@ -917,7 +931,7 @@ object Main extends Logging {
                 error = true
               } else {
                 // only XMLTextInfosetOutputter, JsonInfosetOutputter and
-                // DaffodilOutputContentHandler write directly to the output stream. Other
+                // DaffodilParseOutputStreamContentHandler write directly to the output stream. Other
                 // InfosetOutputters must manually get the result and write it to the stream below
                 eitherOutputterOrHandler match {
                   case Left(sxml: ScalaXMLInfosetOutputter) => {
@@ -1095,8 +1109,17 @@ object Main extends Logging {
                   val task: Future[(Int, Long, Boolean)] = Future {
                     val (time, result) = inData match {
                       case Left(anyRef) => Timer.getTimeResult({
-                        val inputterForUnparse = getInfosetInputter(infosetType, anyRef)
-                        processor.unparse(inputterForUnparse, nullChannelForUnparse)
+                        val inputterForUnparse = getInfosetInputter(infosetType, anyRef, processor, nullChannelForUnparse)
+                        inputterForUnparse match {
+                          case Left(inputter) =>
+                            processor.unparse(inputter, nullChannelForUnparse)
+                          case Right(contentHandler) =>
+                            val is = anyRef match {
+                              case bytes: Array[Byte] => new ByteArrayInputStream(bytes)
+                              case is: InputStream => is
+                            }
+                            unparseWithSAX(is, contentHandler)
+                        }
                       })
                       case Right(data) => Timer.getTimeResult({
                         val input = InputSourceDataInputStream(data)
@@ -1218,8 +1241,19 @@ object Main extends Logging {
                 }
 
               val inputterData = infosetDataToInputterData(unparseOpts.infosetType.toOption.get, eitherBytesOrStream)
-              val inputter = getInfosetInputter(unparseOpts.infosetType.toOption.get, inputterData)
-              val unparseResult = Timer.getResult("unparsing", processor.unparse(inputter, outChannel))
+              val inputterOrContentHandler = getInfosetInputter(unparseOpts.infosetType.toOption
+                .get, inputterData, processor, outChannel)
+              val unparseResult = inputterOrContentHandler match {
+                case Left(inputter) =>
+                  Timer.getResult("unparsing", processor.unparse(inputter, outChannel))
+                case Right(contentHandler) =>
+                  val is = inputterData match {
+                    case bytes: Array[Byte] => new ByteArrayInputStream(bytes)
+                    case is: InputStream => is
+                  }
+                  Timer.getResult("unparsing", unparseWithSAX(is, contentHandler))
+              }
+
               displayDiagnostics(unparseResult)
 
               if (unparseResult.isError) {
@@ -1393,10 +1427,27 @@ object Main extends Logging {
     ret
   }
 
+  private def unparseWithSAX(
+    is: InputStream,
+    contentHandler: DFDL.DaffodilUnparseContentHandler): UnparseResult = {
+    val xmlReader = SAXParserFactory.newInstance.newSAXParser.getXMLReader
+    xmlReader.setContentHandler(contentHandler)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACES_FEATURE, true)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACE_PREFIXES_FEATURE, true)
+    try {
+      xmlReader.parse(new org.xml.sax.InputSource(is))
+    } catch {
+      case _: DaffodilUnparseErrorSAXException => // do nothing, unparseResult has error info
+    }
+
+    val ur = contentHandler.getUnparseResult
+    ur
+  }
+
   private def parseWithSAX(
     processor: DFDL.DataProcessor,
     data: InputSourceDataInputStream,
-    saxContentHandler: DaffodilOutputContentHandler,
+    saxContentHandler: DaffodilParseOutputStreamContentHandler,
     errorHandler: CommandLineSAXErrorHandler): ParseResult = {
     val saxXmlRdr = processor.newXMLReaderInstance
     saxXmlRdr.setContentHandler(saxContentHandler)
--- a/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXParseAPI.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXParseAPI.scala
@@ -1,0 +1,177 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processor
+
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.IOException
+
+import scala.xml.SAXParseException
+
+import org.apache.daffodil.Implicits.intercept
+import org.apache.daffodil.infoset.DaffodilParseOutputStreamContentHandler
+import org.apache.daffodil.processors.ParseResult
+import org.apache.daffodil.xml.XMLUtils
+import org.jdom2.input.sax.BuilderErrorHandler
+import org.jdom2.input.sax.SAXHandler
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.xml.sax.InputSource
+import org.xml.sax.SAXNotRecognizedException
+import org.xml.sax.SAXNotSupportedException
+
+class TestSAXParseAPI {
+  import TestSAXParseUnparseAPI._
+
+  @Test def testDaffodilParseXMLReader_setFeatureUnsupported(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val snr = intercept[SAXNotRecognizedException] {
+      xmlReader.setFeature("http://xml.org/sax/features/validation", true)
+    }
+    assertTrue(snr.getMessage.contains("Feature unsupported"))
+    assertTrue(snr.getMessage.contains("Supported features are:"))
+  }
+
+  @Test def testDaffodilParseXMLReader_get_setFeature(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val feature = "http://xml.org/sax/features/namespace-prefixes"
+    val origValue = xmlReader.getFeature(feature)
+    assertFalse(origValue)
+    xmlReader.setFeature(feature, true)
+    val newValue = xmlReader.getFeature(feature)
+    assertTrue(newValue)
+  }
+
+  @Test def testDaffodilParseXMLReader_setProperty_unsupported(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val property: AnyRef = "Hello"
+    val snr = intercept[SAXNotRecognizedException] {
+      xmlReader.setProperty("http://xml.org/sax/properties/xml-string", property)
+    }
+    assertTrue(snr.getMessage.contains("Property unsupported"))
+  }
+
+  @Test def testDaffodilParseXMLReader_setProperty_badValue(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val property: String = XMLUtils.DAFFODIL_SAX_URN_BLOBDIRECTORY
+    val propertyVal: AnyRef = "/tmp/i/am/a/directory"
+    val sns = intercept[SAXNotSupportedException](
+      xmlReader.setProperty(property, propertyVal)
+    )
+    assertTrue(sns.getMessage.contains("Unsupported value for property"))
+  }
+
+  @Test def testDaffodilParseXMLReader_get_setProperty(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val property: String = XMLUtils.DAFFODIL_SAX_URN_BLOBPREFIX
+    val propertyVal: AnyRef ="testing-blobs"
+    val origValue = xmlReader.getProperty(property)
+    assertNotEquals(propertyVal, origValue)
+    xmlReader.setProperty(property, propertyVal)
+    val newValue = xmlReader.getProperty(property)
+    assertEquals(propertyVal, newValue.asInstanceOf[String])
+  }
+
+  @Test def testDaffodilParseXMLReader_get_setContentHandler(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val parseContentHandler = new SAXHandler()
+    val origValue = xmlReader.getContentHandler
+    assertNull(origValue)
+    xmlReader.setContentHandler(parseContentHandler)
+    val newValue = xmlReader.getContentHandler
+    assertTrue(newValue.isInstanceOf[SAXHandler])
+  }
+
+  @Test def testDaffodilParseXMLReader_get_setErrorHandler(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val eh = new BuilderErrorHandler()
+    val origValue = xmlReader.getErrorHandler
+    assertNull(origValue)
+    xmlReader.setErrorHandler(eh)
+    val newValue = xmlReader.getErrorHandler
+    assertTrue(newValue.isInstanceOf[BuilderErrorHandler])
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_inputSource_no_backing_stream(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val input = new InputSource()
+    val ioe = intercept[IOException](
+      xmlReader.parse(input)
+    )
+    assertTrue(ioe.getMessage.contains("InputSource must be backed by InputStream"))
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_inputSource_with_backing_stream(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val baos = new ByteArrayOutputStream()
+    val parseOutputStreamContentHandler = new DaffodilParseOutputStreamContentHandler(baos)
+    xmlReader.setContentHandler(parseOutputStreamContentHandler)
+    val inArray = testData.getBytes()
+    val bais = new ByteArrayInputStream(inArray)
+    val input = new InputSource(bais)
+    xmlReader.parse(input)
+    val pr = xmlReader.getProperty(XMLUtils.DAFFODIL_SAX_URN_PARSERESULT).asInstanceOf[ParseResult]
+    assertTrue(!pr.isError)
+    assertEquals(testInfoset, scala.xml.XML.loadString(baos.toString))
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_inputStream(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val baos = new ByteArrayOutputStream()
+    val parseOutputStreamContentHandler = new DaffodilParseOutputStreamContentHandler(baos)
+    xmlReader.setContentHandler(parseOutputStreamContentHandler)
+    val inArray = testData.getBytes()
+    val bais = new ByteArrayInputStream(inArray)
+    xmlReader.parse(bais)
+    val pr = xmlReader.getProperty(XMLUtils.DAFFODIL_SAX_URN_PARSERESULT).asInstanceOf[ParseResult]
+    assertTrue(!pr.isError)
+    assertEquals(testInfoset, scala.xml.XML.loadString(baos.toString))
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_byteArray(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val baos = new ByteArrayOutputStream()
+    val parseOutputStreamContentHandler = new DaffodilParseOutputStreamContentHandler(baos)
+    xmlReader.setContentHandler(parseOutputStreamContentHandler)
+    val inArray = testData.getBytes()
+    xmlReader.parse(inArray)
+    val pr = xmlReader.getProperty(XMLUtils.DAFFODIL_SAX_URN_PARSERESULT).asInstanceOf[ParseResult]
+    assertTrue(!pr.isError)
+    assertEquals(testInfoset, scala.xml.XML.loadString(baos.toString))
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_errorHandler_empty_byteArray(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val baos = new ByteArrayOutputStream()
+    val parseOutputStreamContentHandler = new DaffodilParseOutputStreamContentHandler(baos)
+    xmlReader.setContentHandler(parseOutputStreamContentHandler)
+    val eh = new BuilderErrorHandler
+    xmlReader.setErrorHandler(eh)
+    val inArray = "".getBytes()
+    val spe = intercept[SAXParseException](
+      xmlReader.parse(inArray)
+    )
+    val pr = xmlReader.getProperty(XMLUtils.DAFFODIL_SAX_URN_PARSERESULT).asInstanceOf[ParseResult]
+    assertTrue(pr.isError)
+    assertTrue(spe.getMessage.contains("Insufficient bits in data"))
+  }
+}
--- a/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processor
+
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+
+import javax.xml.parsers.SAXParserFactory
+import org.apache.daffodil.xml.XMLUtils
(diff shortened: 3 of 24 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,180 @@
...
+    sendToInputter
+  }
+
+  // TODO not sure what to do with the prefix mappings, technically we don't need them since
+  // we have the attributes with startElement to look up any prefixes we need
+  override def startPrefixMapping(prefix: String, uri: String): Unit = {
+    val pre = if (prefix == "") null else prefix
+    infosetEvent.prefixMapping = One(NamespaceBinding(pre, uri, infosetEvent.prefixMapping
+      .orNull))
+  }
```
stevedlawrence, line 75: Yeah, seem like we don't need it. Unless there are XMLReaders don't prvide certain parametrs in startELement? LIke maybe uri is always null, and they expect you to figure out the namespace uri based on the qName string?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,180 @@
...
+      // reset since it was whitespace only
+      characterData.setLength(0)
+    }
+
+    if (infosetEvent == null) infosetEvent = new DFDL.SaxInfosetEvent
+    else if (!infosetEvent.isEmpty && infosetEvent.localName.isDefined) {
+      // we started another element while we were in the process of building another
+      // this most likely means the first element was complex, either ways, it is ready
+      // for the inputter queue
+      sendToInputter
```
stevedlawrence, line 96: > most likely means

I think this definitely means, right? I'm not sure I understand the isEmpty or localName.isDefined checks? can you explain those?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,180 @@
...
+    if (localName.nonEmpty && uri.nonEmpty) {
+      infosetEvent.namespaceURI = One(uri)
+      infosetEvent.localName = One(localName)
+    } else if (qName.nonEmpty) {
+      val qNameArr = qName.split(":")
+
+      if (qNameArr.length > 1) {
+        infosetEvent.localName = qNameArr.lift(1)
+        val qNamePrefix = qNameArr.lift(0).get
+        val in = atts.getIndex(s"xmlns:$qNamePrefix")
```
stevedlawrence, line 116: This implies that the namespace for this prefix is defined on this element. But it could have been defined on some parent element. I think perhaps this is where the prefix mapping comes into place. Perhaps we need to build a NamespaceBuilding, and if we do not get uri + local name, then we have to look up the prefix (which could be empty string) in the NamespaceBinding?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,180 @@
...
+      sendToInputter
+    }
+    infosetEvent.eventType = One(EndElement)
+    sendToInputter
+  }
+
+  override def characters(ch: Array[Char], start: Int, length: Int): Unit = {
+    if (infosetEvent == null) {
+      infosetEvent = new DFDL.SaxInfosetEvent
+    }
```
stevedlawrence, line 148: Having a single infosetEvent avoid this kind of stuff. We could just always have an infoset event.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -311,15 +323,62 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
...
+    // kick off SAX Unparsing
+    try {
+      xmlReader.parse(new InputSource(saxInputStream))
+    } catch {
+      case e: DaffodilUnhandledSAXException =>
+        // In the case of an unexpected errors, catch and throw as TDMLException
+        throw TDMLException("Unexpected error during SAX Unparse:" + e, None)
+      case _: DaffodilUnparseErrorSAXException =>
+        // TODO not sure what else to do with this error, unparseResult and its diagnostics
+        // will be handled below
```
stevedlawrence, line 361: Yeah, I think this is safe to ignore since you get the parse result from the ContentHandler.

stevedlawrence (commented):
Looks really nice. Great feature to have. Just a handful of mostly minor.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -0,0 +1,157 @@
...
+    _isNilled
+  }
+
+  //called without changing any state
+  override def hasNext(): Boolean = {
+    if (isDone) {
+      false
+    } else {
+      val event = this.resume(ch, Try(currentEvent))
+      nextEvent = copyEvent(event.getOrElse(null))
```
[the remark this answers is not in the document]
olabusayoT: Yes, the contentHandler has just 1 event that it modifies, which is why I chose to do the copy way. But I updated to have just 1 instance of the 3 events and they we copy between the fields rather than the allocating a new infosetEvent on each copy

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -0,0 +1,157 @@
...
+      nextEvent != null
+    }
+  }
+
+  // done with this current event, move on to the next event
+  override def next(): Unit = {
+    currentEvent = copyEvent(nextEvent)
+    if (currentEvent.eventType.contains(EndDocument)) {
+      isDone = true
+    }
```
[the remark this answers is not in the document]
olabusayoT: I decided to stick with copying the fields and have only 1 allocation of each infosetEvent which happens at it's definition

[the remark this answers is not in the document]
stevedlawrence: Makes sense. Agreed that this is probably the best approach to minimze allocations. Only concern is it looks like the ContentHandler is allocating new instances? Can that be avoided?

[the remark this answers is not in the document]
olabusayoT: Yes, I took those out as well. Now we only allocate a new InfosetEvent at the declaration of the each infoset val

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
olabusayoT: The expectation is for the readers to support the features, which a user can set to either get the uri and localName or just the qualified qName. So I think we can build this mapping and use it together with the Attributes in startElement to ensure we all the info to get a URI

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+    infosetEvent.namespaceURI =
+      if (uri.nonEmpty) {
+        One(uri)
+      } else if (qNameArr.length > 1) {
+        One(prefixMapping.getURI(qNameArr.head))
+      } else {
+        try {
+          One(prefixMapping.getURI(null))
+        } catch {
+          case _: NullPointerException => Nope
```
mbeckerle, line 203: If prefixMapping.getURI(null) can throw a NPE, then can the call to prefixMapping.getURI above also? So should this try/catch surround that also?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  override def characters(ch: Array[Char], start: Int, length: Int): Unit = {
+    characterData.appendAll(ch, start, length)
+  }
+
+  private def sendToInputter(): Unit = {
+    val infosetEventWithResponse = this.resume(inputter, Try(infosetEvent))
+    infosetEvent.clear
+    // if it is failure, we will not have an unparseResult, so we only set unparseResults for
+    // events wrapped in Success, which can include events wrapped in Success, that have
```
mbeckerle, line 218: "events wrapped in Success, which can include events wrapped in Success"...?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+    // expected errors.
+    if (infosetEventWithResponse.isSuccess && infosetEventWithResponse.get.unparseResult.isDefined) {
+      unparseResult = infosetEventWithResponse.get.unparseResult.get
+    }
+    // the exception from events wrapped in failures and events wrapped in Success with an unparse
```
mbeckerle, line 223: Can you please clarify what "wrap in Success" or "wrap in failure" means?

mbeckerle (changes requested):
Requesting just some small changes and clarifying comments.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -779,24 +781,33 @@ object Main extends Logging {
...
+      case "sax" => data
     }
   }
 
-  def getInfosetInputter(infosetType: String, anyRef: AnyRef): InfosetInputter = {
+  def getInfosetInputter(
+    infosetType: String,
+    anyRef: AnyRef,
+    processor: Option[DFDL.DataProcessor] = None,
+    outChannel: Option[DFDL.Output] = None): Either[InfosetInputter, DFDL.DaffodilInputContentHandler] = {
```
tuxji, line 792: Hmm, I don't see any benefit from declaring `x: Option[X] = None` at lines 791 and 792 only to call `x.get` afterwards at lines 809 and 810 below.  You're still going to get an exception, only it will be called `Predef.NoSuchElementException` instead of `java.lang.NullPointerException`.  If you don't want an exception, you would have to use getOrElse, a match expression, or flatten, flatMap, foreach, etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
```diff
@@ -202,6 +212,66 @@ object DFDL {
...
+        causeError.isEmpty &&
+        unparseResult.isEmpty
+    }
+
+    override def toString: String = {
+      val errorState = if (isError) "ERROR" else ""
+      s"$errorState ${if (eventType.isDefined) eventType.get else "No_EventType"}:" +
+        s"{${if (namespaceURI.isDefined) namespaceURI.get else ""}}" +
+        s"${if (localName.isDefined) localName.get else "No_Name"}:" +
+        s"(${if (simpleText.isDefined) simpleText.get else "No_content"})"
```
tuxji, line 262: You can simplify these lines by using `.getOrElse` instead of an if expression.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -0,0 +1,192 @@
...
+        throw new InvalidInfosetException("xsi:nil property is not a valid boolean: '" + nilValue +
+          "' for element " + getLocalName())
+      }
+    } else {
+      MaybeBoolean.Nope
+    }
+    _isNilled
+  }
+
+  //called without changing any state
```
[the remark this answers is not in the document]
olabusayoT: Hmm, it thought it was the opposite, that hasNext was more akin to a peek (i.e can be called multiple times w/o affecting state), and next was like top (changes state each time it is called). This was Steve's comment on the motivation for changing to the method: https://github.com/apache/incubator-daffodil/pull/436#discussion_r505646529

Also this way is more inline with how the other infosetinputters implement their hasNext/next

tuxji (changes requested):
I'm a little concerned about how many times Option.get is called.  Every call of Option.get is a potential exception waiting to be thrown and I've already run out of stamina/time before I could finish reviewing every such place.  There are better idioms for using Option as shown in <https://alvinalexander.com/scala/best-practice-option-some-none-pattern-scala-idioms/>.   You could use match expressions, getOrElse, or other functions in some places instead of using isDefined in if expressions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
olabusayoT: Unfortunately Maybes do not have .getOrElse; I'm told this is to prevent allocations, which is the point of using Maybes over Options. So in place of .getOrElse, we have to use 'if x.isDefined x else ...'

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
mbeckerle: Ah, so wrapped in a Try Success/Failure.  Got it. Thanks.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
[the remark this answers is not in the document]
mbeckerle: Well, every iterator I've ever written, the action all takes place in hasNext(). The logic is always:
```
def hasNext() = {
  if (isPreBuffered) true
  else {
    tryToFillPreBuffer() // does all the work, sets isPreBuffered boolean accordingly
    isPreBuffered
  }
}

def next() = 
  if (hasNext) {
    val result = preBuffer
    clearPreBuffer()
    result
  } else {
    .... exception ...
  }
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
mbeckerle: Yup. This is the runtime system. We need to avoid scala-constructs that are heavyweight in favor of lightweight java-ish idioms unfortunately.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -0,0 +1,192 @@
...
+  extends InfosetInputter with DFDL.ConsumerCoroutine {
+  // allows support for converting relative URIs in data to absolute URIs, this is mainly use
+  // for TDML as the tests allow relative URIs. This can be set to true by calling the
+  // inputContentHandler.enableInputterUriAbsolutization()
+  var enableUriAbsolutization: Boolean = false
+
+  var endDocumentReceived = false
+  val currentEvent: DFDL.SaxInfosetEvent = new DFDL.SaxInfosetEvent
+  val nextEvent: DFDL.SaxInfosetEvent = new DFDL.SaxInfosetEvent
+
```
stevedlawrence, line 75: Can these vals be made private? Should help IDE autocompletion if we get better at marking things private that can be private. I think scala compiler can also make some optimizations if it knows that nothign outside this class can access a variable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+      }
+    infosetEvent.namespaceURI =
+      if (uri.nonEmpty) {
+        One(uri)
+      } else if (qNameArr.length > 1) { // has a prefix
+        // get the prefix off the qname
+        val qNamePrefix = qNameArr.lift(0).get
+        // look up prefix with and without xmlns prefix
+        val in = atts.getIndex(s"xmlns:$qNamePrefix")
+        val otherIn = if (in == -1) atts.getIndex("qNamePrefix") else in
```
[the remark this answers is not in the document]
stevedlawrence: I think this is looking for the XML attribute on the element named ``xmlns:foo="..."`` to figure out the namespace. Though, it isn't clear to me why this is needed.. Shouldn't ``startPrefixMapping`` be called for each of the xmlns mappings in the infoset, so our namepace bindings are all available there? So this could be somethign like this?
```scala
prefixMapping.getURI(qNamePrefix)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+    sendToInputter()
+  }
+
+  override def startPrefixMapping(prefix: String, uri: String): Unit = {
+    val pre = if (prefix == "") null else prefix
+    prefixMapping = NamespaceBinding(pre, uri, prefixMapping)
+  }
+
+  override def endPrefixMapping(prefix: String): Unit = {
+    // do nothing
```
stevedlawrence, line 99: I think we need to do something here. If startPrefixMapping is adding a new mapping endPrefixMapping needs to remove that, i.e.
```sclaa
prefixMapping = prefixMapping.parent
```
Note that according to SAX the order of endPrefixMapping doesn't necessarily match the order of startPrefixMapping. So using the ``prefix`` variable to check we're removing the right thing would require looking up the namespace mapping and removing it. And since NamespaceMappings are immutable, it makes it hard to remove a previous one. So I'd just recommend we remove the top, and as long as the XMLReader is doing the right thing we'll remove the right number of mappings.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
tuxji: Gotcha, thought Option was being used, sorry.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -0,0 +1,192 @@
...
+  // done with current event, move on to the next event
+  override def next(): Unit = {
+    copyEvent(source = nextEvent, dest = currentEvent)
+    if (currentEvent.eventType.contains(EndDocument)) {
+      endDocumentReceived = true
+      nextEvent.clear
+    } else {
+      val event = this.resume(inputContentHandler, Try(currentEvent))
+      copyEvent(source = event.getOrElse(null), dest = nextEvent)
```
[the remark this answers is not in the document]
olabusayoT: That makes sense. The event is actually a Try[SaxInfosetEvent], so I removed the getOrElse and passed it in directly, and did "isFailure" as the check in the function

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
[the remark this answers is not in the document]
olabusayoT: Ah good catch. That was a typo!

@stevedlawrence , the XMLReader implementation we use doesn't set the namespaces feature to be true by default, so we have no uris/localNames, and start/endPrefixMappings don't get called and we must depend instead on the AttributesImpl object. So this implementation is for the use case where the user's XMLReader implementation also has non-standard feature implementation. So AttributeImpl usually looks like "xmlns:prefix", so the prefix is stored with the namespaceURI, and we can look it up this way. 

We do check if the prefixMapping has the qNamePrefix though, we just check if after we check the atts.

```scala
name = "localpart="xmlns:ex",rawname="xmlns:ex""
type = "CDATA"
value = "http://example.com"
nonNormalizedValue = "http://example.com"
```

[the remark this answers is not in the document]
stevedlawrence: Gotcha, it would be worth adding comment explaining why we have to do all this logic.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilInputContentHandler.scala
```diff
@@ -0,0 +1,250 @@
...
+    // use Attributes to determine xsi:nil value
+    val nilIn = atts.getIndex(XMLConstants.W3C_XML_SCHEMA_INSTANCE_NS_URI, "nil")
+    infosetEvent.nilValue = if (nilIn >= 0) {
+      val nilValue = atts.getValue(nilIn)
+      One(nilValue)
+    } else {
+      Nope
+    }
+    // set localName and namespaceURI
+    val qNameArr = qName.split(":")
```
stevedlawrence, line 130: I'm wondering for performance reasons we could avoid this split since it isn't always necessary? I belive split is implemtend via regex, so it's going to be relatively slow. If I'm reading the SAX API, we should either get uri + llocalName or qName. So if we get uri + localName, we can just ignore qName and never have to do a split. Or perhaps just make the qNameArr a lazy val so we only evaluate it if it's needed.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
olabusayoT: The "= None" was intended as the default for the arguments so the nonSAX infosetTypes wouldn't need to pass in anything for the processor and outChannel params. But in both places I call it, we're passing in Some(_), so I removed the option and default.

I could change it so we pass in these 2 variables in a tuple via the anyRef, but I thought this way would be clearer since the anyRef is intended to represent inputterData

mbeckerle (approved):
+1

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -217,6 +228,49 @@
...
+ * implementation. The call to the
+ * {@link org.apache.daffodil.japi.DataProcessor#newContentHandlerInstance(java.nio.channels.WritableByteChannel)} method must be provided with the {@link java.nio.channels.WritableByteChannel},
+ * where the unparsed data ought to be written to. Any XMLReader implementation is permissible, as
+ * long as they allow support for the namespace and namespace-prefixes features of XMLReader. The
+ * namespaces feature MUST be set to true, and namespace-prefixes is only optional if the former is true.
+ *
+ * <pre>
+ * {@code
+ *  ByteArrayInputStream is = new ByteArrayInputStream(data);
+ *  DaffodilInputContentHandler inputContentHandler = dp.newContentHandlerInstance(wbc);
```
stevedlawrence, line 244: I'm wondering if there's a better name than DaffodilInputContentHandler? I'm not really sure what the Input here references since the XMLReader is what reads the input. The ContentHandler just outputs the unparsed data.

Maybe just DaffodilContentHandler which mimics the existing DaffodilXMLReader. I could also maybe see something like DaffodilUnparseContentHandler and DaffodilParseXMLReader if we want a more clear distinction between parse/unparse.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -217,6 +228,49 @@
...
+ *   xmlReader.setContentHandler(inputContentHandler)
+ *  } catch (ParserConfigurationException | SAXException e) {
+ *   ...
+ * `}
+ * }
+ * </pre>
+ *
+ * The call to the XMLReader.parse method must be wrapped in a try/catch, as
+ * {@link org.apache.daffodil.japi.DaffodilInputContentHandler} relies on throwing an exception to
+ * stop the coroutine and the XMLReader.parse in the case of any errors/failures.
```
stevedlawrence, line 256: Not sure mentioning the coroutine here adds anything since that's purely an implementation details. People using this, which is what this doc is intended for, dont' really care too much about that fact that we use a coroutine behind the scenes.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -217,6 +228,49 @@
...
+ * {@link org.apache.daffodil.japi.DaffodilUnparseErrorSAXException}, for any other errors.
+ *
+ * <pre>
+ * {@code
+ *  try {
+ *    xmlReader.parse(new InputSource(is));
+ *  } catch (DaffodilUnparseErrorSAXException | DaffodilUnhandledSAXException e) {
+ *    ...
+ *  }
+ *  UnparseResult ur = inputContHandler.getUnparseResult();
```
stevedlawrence, line 269: If we get a UnhandledSAXException, then what does getUnparseResult return? Can it be null?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -835,6 +845,18 @@ class InvalidParserException private[japi] (cause: org.apache.daffodil.compiler.
...
 class InvalidUsageException private[japi] (cause: org.apache.daffodil.processors.InvalidUsageException) extends Exception(cause.getMessage(), cause.getCause())
 
+/**
+ * This exception will be thrown when when the unparseResult.isError during a SAX Unparse
+ */
+class DaffodilUnparseErrorSAXException(exception: SDaffodilUnparseErrorSAXException)
+  extends org.xml.sax.SAXException(exception.getMessage)
+
+/**
+ * This exception will be thrown when an unexpected error occurs during the SAX unparse
```
stevedlawrence, line 855: Might be worth documentation that if this happens there is no UnparseResult and getUnparseResult from teh content handler may return null, or something to that effect.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -953,3 +975,68 @@ class DaffodilXMLReader private[japi] (xmlrdr: SDaffodilXMLReader) extends org.x
    */
   def parse(arr: Array[Byte]): Unit = xmlrdr.parse(arr)
 }
+
+/**
+ * Read in an infoset from any SAX Reader for unparsing
+ */
```
stevedlawrence, line 981: I'm not sure I would say "read in". Since this doesn't really do that. The XMLReader handles parsing the XML. This just unparses data base on SAX events from an XMLReader.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -953,3 +975,68 @@ class DaffodilXMLReader private[japi] (xmlrdr: SDaffodilXMLReader) extends org.x
...
+class DaffodilInputContentHandler private[japi] (sContentHandler: SDaffodilInputContentHandler)
+  extends ContentHandlerProxy {
+  override val contentHandler: org.xml.sax.ContentHandler = sContentHandler
+
+  def getUnparseResult: UnparseResult =
```
stevedlawrence, line 986: Needs scaladoc for getUnparseResult.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -953,3 +975,68 @@ class DaffodilXMLReader private[japi] (xmlrdr: SDaffodilXMLReader) extends org.x
...
+    new UnparseResult(sContentHandler.getUnparseResult.asInstanceOf[SUnparseResult])
+  def enableInputterBlobUriAbsolutization(): Unit= sContentHandler.enableInputterBlobUriAbsolutization()
```
stevedlawrence, line 988: Needs scala doc, and should mention that this should note be used in production. I'm also not sure about "Absolutization". Maybe something like "resolveRelativeInfosetBlobURIs()

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -913,3 +935,70 @@ class DaffodilXMLReader private[sapi] (xmlrdr: SDaffodilXMLReader) extends org.x
...
   def parse(arr: Array[Byte]): Unit = xmlrdr.parse(arr)
 }
+
+/**
+ * Read in an infoset from any SAX Reader for unparsing
+ */
+class DaffodilInputContentHandler private[sapi] (sContentHandler: SDaffodilInputContentHandler)
+  extends ContentHandlerProxy {
+
+  override val contentHandler: org.xml.sax.ContentHandler = sContentHandler
```
stevedlawrence, line 945: I think this should be private/protected. I think even though the base class defines it as protected, I think not making it protected here will make it public, which we probably don't want.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
olabusayoT: Decided to take it out, since this is used specifically for TDML. Also renamed!

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
olabusayoT: It will be null, updated documentation to reflect

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -0,0 +1,309 @@
...
+    // set localName and namespaceURI
+    lazy val qNameArr = qName.split(":")
+    infosetEvent.localName =
+      if (localName.nonEmpty) {
+        One(localName)
+      } else if (qNameArr.length > 1) {
+        One(qNameArr.last)
+      } else if (qNameArr.nonEmpty) {
+        One(qNameArr.head)
+      } else {
```
tuxji, line 180: What is the difference between the `qNameArr.length > 1` and `qNameArr.nonEmpty` cases which requires returning `One(qNameArr.last)` on one hand and `One(qNameArr.head)` on the other hand?  Could you merge both cases into `qNameArr.nonEmpty` and return `One(qNameArr.last)` since `last == head` when `qNameArr.length == 1`?  I suspect there must be a reason since I see the simpler logic used in endElement at lines 232-239 below.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/package.scala
```diff
@@ -134,36 +139,37 @@ package org.apache.daffodil
  *
  * <h5>SAX Parse</h5>
  *
- * The [[DaffodilXMLReader.parse(isdis:org\.apache\.daffodil\.sapi\.io\.InputSourceDataInputStream* DaffodilXMLReader.parse]] method accepts input data to parse in the form of a
+ * The [[DaffodilParseXMLReader.parse(isdis:org\.apache\.daffodil\.sapi\.io\.InputSourceDataInputStream* DaffodilParseXMLReader.parse]] method accepts input data to parse in the form of a
```
tuxji, line 142: Very long line here.

tuxji (approved):
+1

Looks like this PR's almost finished!  Only minor questions or suggestions below.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -217,6 +227,52 @@
...
  *
+ * <h5>SAX Unparse</h5>
+ *
+ * In order to kick off an unparse via the SAX API, one must register the
+ * {@link org.apache.daffodil.japi.DaffodilUnparseContentHandler} as the contentHandler for an XMLReader
+ * implementation. The call to the
+ * {@link org.apache.daffodil.japi.DataProcessor#newContentHandlerInstance(java.nio.channels.WritableByteChannel)} method must be provided with the {@link java.nio.channels.WritableByteChannel},
+ * where the unparsed data ought to be written to. Any XMLReader implementation is permissible, as
+ * long as they allow support for the namespace and namespace-prefixes features of XMLReader. The
+ * namespaces feature MUST be set to true, and namespace-prefixes is only optional if the former is true.
```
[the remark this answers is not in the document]
olabusayoT: The caller has to set the namespaces feature, since it is called on the XMLReader, and that is outside of the Daffodil unparse work. The places you'll see Daffodil doing this is from the CLI and tests, but we don't set it for the caller. As far as if there are bad things that could happen, i would say no primarily because the namespaces feature should be set to true on default by the XMLReader, but also because we have support for when the namespaces feature isn't true.

According to the [docs](http://www.saxproject.org/namespaces.html), when the namespaces feature is set to true, the XMLReader should send the namespace/prefix mappings to the ContentHandler and include the uri and localName when calling start/endElement. If it is false, then the namespaces-feature has to be true, otherwise it is an illegal state. When the latter feature is true, the namespace/prefix mappings are sent only through startElement, and its AttributeImpl param (namespaces are treated like attributes). The good news is we have support for extracting the namespace/prefix mappings through either of these routes, so I'll update the documentation to make it more generic and say "as long as they have XML Namespace support."

The rest of the information should only matter for the parse side, once [DAFFODIL-2422](https://issues.apache.org/jira/browse/DAFFODIL-2422) is completed

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -953,3 +989,72 @@ class DaffodilXMLReader private[japi] (xmlrdr: SDaffodilXMLReader) extends org.x
...
+  extends ContentHandlerProxy {
+
+  override protected val contentHandler: org.xml.sax.ContentHandler = sContentHandler
+
+  /**
+   * Returns the result of the SAX unparse containing diagnostic information. In the case of an
+   * DaffodilUnhandledSAXException, this will return null.
+   */
+  def getUnparseResult: UnparseResult =
+    new UnparseResult(sContentHandler.getUnparseResult.asInstanceOf[SUnparseResult])
```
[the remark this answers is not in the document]
olabusayoT: Ah great catch, the implementation should actually change. In the case of  an unhandleException sUnparseResult will be null, so we should not create an UnparseResult object from that and should return null instead

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
olabusayoT: Ah you're right, it was a left over from when I was using options, I was returning either lift(1) or lift(0) then since the former case represented when there was a prefix and a localName and the latter, when there is only a localName, but it can be simplified to .nonEmpty and .last

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -0,0 +1,309 @@
...
+  def mapPrefixMappingFromAttributesImpl(atts:Attributes): Unit = {
+    var i = 0
+    while (i < atts.getLength) {
+      val qName = atts.getQName(i)
+      if (qName.startsWith("xmlns")) {
+        val uri =  atts.getValue(i)
+        val prefix = if(qName.contains(":")) {
+          val pref = qName.split(":").last
+          pref
+        } else {
```
stevedlawrence, line 129: I think there's potential for very subtle, though rare, bugs. Which are some of the worst. For example, what if we have an attribute thats ``xmlnsblah:foo=whatever``. That's not an xmlns attribute, but this logic would think it is. I think maybe instead we need something more like
```scala
if (qname == "xmlns") {
  prefixMapping = NamespaceBinding(null, uri, prefixMapping) // no prefix
} else if (qname.startsWith("xmlns:")) {
  prefixMapping = NamespaceBinding(qname.substring(6), uri, prefixMapping)
} else {
  // no-op, not a namespace mapping attribute
}
```
This also avoids ``contains`` and ``split``, which I think are relatively expensive as far as string functions go--I think they both treat the string parameter as a regex behind the scenes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -0,0 +1,309 @@
...
+    sendToInputter()
+  }
+
+  override def endDocument(): Unit = {
+    infosetEvent.eventType = One(EndDocument)
+    sendToInputter()
+  }
+
+  override def startPrefixMapping(prefix: String, uri: String): Unit = {
+    contentHandlerPrefixMappingUsed = true
```
stevedlawrence, line 101: Does the SAX API say anythign about not mixing use of startPrefixMapping and xmlns in the Attributes? It seems reasonable to only have one or the other, but if they are allowed to be mixed, we might not want to disable the the mapPrefixMapping stuff if we see startPrefixMapping. It's defnitely nice if they can't be mixed, because then we can avoid that stack logic and parsing Attributes for xmlns.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -0,0 +1,309 @@
...
+        One(qNameArr.head)
+      } else {
+        Nope
+      }
+    infosetEvent.namespaceURI =
+      if (uri.nonEmpty) {
+        One(uri)
+      } else if (qNameArr.length > 1) { // has a prefix
+        // this use case is for the situation where the XMLReader doesn't pass in uri/localname,
+        // so prefix information must be determined via AttributeImpl attribute. This can usually
```
stevedlawrence, line 188: Do you mean Attribute instead of AttributeImpl here? The parameter passed in in just Attribute.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -0,0 +1,309 @@
...
+      }
+    infosetEvent.eventType = One(EndElement)
+
+    if (!contentHandlerPrefixMappingUsed) {
+      // always pops
+      prefixMapping = prefixMappingTrackingStack.pop
+    }
+    sendToInputter()
+  }
+
```
stevedlawrence, line 264: Can the logic involved uri, localname and qname be factored out into a helper function? It looks like the logic in startElement and enedElement is the same?

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXParseAPI.scala
```diff
@@ -0,0 +1,218 @@
...
+    val ioe = intercept[IOException](
+      xmlReader.parse(input)
+    )
+    assertTrue(ioe.getMessage.contains("Inputsource must be backed by Inputstream"))
+  }
+
+  @Test def testDaffodilParseXMLReader_parse_inputSource_with_backing_stream(): Unit = {
+    val xmlReader = dp.newXMLReaderInstance
+    val baos = new ByteArrayOutputStream()
+    val ch = new DaffodilOutputContentHandler(baos)
```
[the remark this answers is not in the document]
olabusayoT: The OutputContentHandler is the internal contentHandler we use on the parse side, so we don't have a dp method that creates it. I will change it to DaffodilParseContentHandler to make it more clear!

Will do!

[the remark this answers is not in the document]
stevedlawrence: Ah that's right. Maybe it's worth renaming this to DaffodilOutputStreamContentHandler?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
olabusayoT: So unfortunately it appears to be implementation dependent when namespaces=false and namespace-prefixes=true (it is listed as 'unknown' behavior...which seems to me to be up for interpretation). In the true/true and true/false scenarios, they are explicitly not mixed. In the SAXParserFactory implementation we use for CLI and tests, it doesn't mix them in any scenarios, though. I just thought to play it safe since we are using a shared variable for both. 

Because if they are mixed, I think it would be duplicated information. There would be no reason to have it be separate information.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -38,44 +38,48 @@
...
+ *
  * <pre>
  * {@code
- * xmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBDIRECTORY(), "/tmp/");
- * xmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBPREFIX(), "daffodil-sax-");
- * xmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBSUFFIX(), ".bin");
+ * xmlReader.setProperty(xmlReader.DAFFODIL_SAX_URN_BLOBDIRECTORY(),
+ *  Paths.get(System.getProperty("java.io.tmpdir"))); // value type: java.nio.file.Paths
+ * xmlReader.setProperty(xmlReader.DAFFODIL_SAX_URN_BLOBPREFIX(), "daffodil-sax-"); // value type String
+ * xmlReader.setProperty(xmlReader.DAFFODIL_SAX_URN_BLOBSUFFIX(), ".bin"); // value type String
```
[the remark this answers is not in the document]
olabusayoT: You're right. For some reason, intellij adds it in during autocomplete! Fixed!

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
stevedlawrence: Seems reasonable then to make the assumption. Probably worth adding a comment somewhere (maybe documenting the val contentHandlerPrefixMappingUsed) how this variable is used and that it's making this assumption that these things aren't mixed.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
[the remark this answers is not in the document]
olabusayoT: Ah I remember now why I did it like this! The code actually requires it. I'm not sure if it's because the api is implemented behind the scenes in scala. But I wanted to have the documentation match the expectation in the code, because the code won't compile without the (). So I've unfixed this, because if someone is trying to write code in Java using the API, they will need the parens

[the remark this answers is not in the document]
olabusayoT: Reference: https://alvinalexander.com/scala/how-to-access-scala-var-val-fields-from-java/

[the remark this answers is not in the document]
stevedlawrence: Yeah, this is a downside with using Scala in Java. Much of what Scala generates in bytecode is actually public functions that get/set private variables, without a way to access the actual variables themselves. I'm not sure if there's away around this.

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXParseAPI.scala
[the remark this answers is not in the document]
olabusayoT: I also added tests directly connecting SAX parse/unparse and Dataprocessor parse/unparse, as was mentioned in original comment

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
```diff
@@ -0,0 +1,123 @@
...
+class TestSAXUnparseAPI {
+  import TestSAXUnparseAPI._
+
+  @Test def testUnparseContentHandler_unparse(): Unit = {
+    val bao = new ByteArrayOutputStream()
+    val wbc = java.nio.channels.Channels.newChannel(bao)
+    val unparseContentHandler = dp.newContentHandlerInstance(wbc)
+    xmlReader.setContentHandler(unparseContentHandler)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACES_FEATURE, true)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACE_PREFIXES_FEATURE, true)
```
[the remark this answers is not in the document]
olabusayoT: You are right, XMLReaders are not threadsafe, and neither are SAXParserFactorys! Fixed

stevedlawrence (approved):
+1 looks great!

PR #437, opened 2020-10-15
Free infoset element during unparse

stevedlawrence:
- This refactors how we figure out when DINodes are final during
  unparsing, mainly for arrays. Previously, arrays were marked final
  when their parent became final, but that meant arrays could
  potentially stick around for a long time. Instead, we set arrays as
  final when a new element is added after it, or if it's the last thing
  in a complex type. In both cases, we know no more elements will be
  added and the array is final
- After marking DINodes as final we free them if they are not used in an
  expression to reduce memory requirements
- Changes dfdlx:outputTypeCalcNextSibling function to mark possible next
  siblings and itself as used in an expression, since both are needed in
  the function implementation so we cannot free those elements.
- Add new releaseUnneededInfoset tunable which allows one to disable
  this freeing of infoset elements, for both parse and unparse. This is
  helpful when running unit tests that do manual inspect of the infoset
  after it is complete to ensure correctness. We can't have these
  elements being freed. Refactor existing variable names to match this
  tunable
- Adds areDebugging logic to UState, similar to PState, so we can
  disable freeing infoset nodes while debugging.

DAFFODIL-1272

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfoset.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfoset.scala
@@ -74,12 +74,17 @@ object TestInfoset {
   }
 
   /**
-   * Returns the root element of the infoset, along with
-   * the schema compiler Root object for examining schema-compiler
-   * computations for unit testing them.
+   * Returns the root element of the infoset, along with the schema compiler
+   * Root object for examining schema-compiler computations for unit testing
+   * them. Because this assumes tests will be inspecting the internal infoset
+   * for correctness, it sets the releaseUnusedInfoset tunable to false so that
+   * the infoset elements are not freed
    */
   def testInfoset(testSchema: scala.xml.Elem, infosetAsXML: scala.xml.Elem): (DIElement, Root, DaffodilTunables) = {
-    val schemaCompiler = Compiler().withTunable("allowExternalPathExpressions", "true")
+    val schemaCompiler =
+      Compiler()
+        .withTunable("allowExternalPathExpressions", "true")
+        .withTunable("releaseUnneededInfoset", "false")
     val pf = schemaCompiler.compileNode(testSchema).asInstanceOf[ProcessorFactory]
     if (pf.isError) {
       val msgs = pf.getDiagnostics.map { _.getMessage() }.mkString("\n")
--- a/daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
@@ -1,0 +1,282 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.infoset
+
+import java.nio.channels.Channels
+
+import org.apache.commons.io.output.NullOutputStream
+
+import org.junit.Assert._
+import org.junit.Test
+
+import org.apache.daffodil.compiler.Compiler
+import org.apache.daffodil.io.InputSourceDataInputStream
+import org.apache.daffodil.processors.parsers.PState
+import org.apache.daffodil.processors.unparsers.UStateMain
+import org.apache.daffodil.util.SchemaUtils
+
+object TestInfosetFree {
+
+  /**
+   * Compiles an infoset with infoset releasing disabled. By disabling the
+   * releasing, the Scala infoset is marked with attributes signifying which
+   * elements *would* have been freed if we didn't disable it. This allows for
+   * a reliable way to validate which elements were freed.
+   *
+   * This first parses and unparses data, without freeing any infoset
+   * nodes--only marking as described above. It then walks both infosets,
+   * configured to add information to the infoset about which elements were
+   * freed. We validate that both the parse and unparse results are the
+   * same--any infoset elements freed during a parse should also be freed
+   * during unparse. We then return the infoset for the test to compare against
+   * the expected value.
+   */
+  def test(
+    schema: scala.xml.Elem,
+    bytes: Array[Byte]): scala.xml.Node = {
+
+    val compiler = Compiler()
+      .withTunable("releaseUnneededInfoset", "false")
+
+    val pf = compiler.compileNode(schema)
+    if (pf.isError) {
+      val msgs = pf.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("pf compile errors: " + msgs)
+    }
+    pf.sset.root.erd.preSerialization // force evaluation of all compile-time constructs
+    val dp = pf.onPath("/")
+    if (dp.isError) {
+      val msgs = dp.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("dp compile errors: " + msgs)
+    }
+
+    // Parse the data, note that we do not set showFreedInfo here because
+    // DINodes aren't freed until *after* the infoset walker walks them. So the
+    // maybeFreed state hasn't been set yet when the infoset outputter gets the
+    // events. We must walk the infoset again after the parse is complete
+    val parseInput = InputSourceDataInputStream(bytes)
+    val parseOutputter = new ScalaXMLInfosetOutputter()
+    val parseResult = dp.parse(parseInput, parseOutputter)
+    if (parseResult.isError) {
+      val msgs = parseResult.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("parse errors: " + msgs)
+    }
+
+    val unparseInputter = new ScalaXMLInfosetInputter(parseOutputter.getResult)
+    val unparseOutput = Channels.newChannel(NullOutputStream.NULL_OUTPUT_STREAM)
+    val unparseResult = dp.unparse(unparseInputter, unparseOutput)
+    if (unparseResult.isError) {
+      val msgs = unparseResult.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("unparse errors: " + msgs)
+    }
+
+    // now walk the parse and unparse infosets and convert them to Scala XML
+    // with the showFreedInfoset set 
+
+    def docToXML(doc: DIDocument): scala.xml.Node = {
+      val detailedOutputter = new ScalaXMLInfosetOutputter(
+        showFormatInfo = false,
+        showFreedInfo = true)
+
+      val infosetWalker = InfosetWalker(
+        doc,
+        detailedOutputter,
+        walkHidden = true, // let's ensure any hidden elements are free
+        ignoreBlocks = true, // there should be no blocks, but ignore them just to be sure
+        releaseUnneededInfoset = false) // do not free the infoset
+      infosetWalker.walk(lastWalk = true)
+
+      detailedOutputter.getResult()
+    }
+
+    val parseDoc = parseResult.resultState.asInstanceOf[PState].infoset.asInstanceOf[DIDocument]
+    val unparseDoc = unparseResult.resultState.asInstanceOf[UStateMain].documentElement
+  
+    val parseXML = docToXML(parseDoc)
+    val unparseXML = docToXML(unparseDoc)
+
+    if (parseXML.toString != unparseXML.toString) {
+      fail("parse and unparse XML did not match, infoset not freed the same")
+    }
+
+    parseXML 
+  } 
+}
+
+class TestInfosetFree {
+
+  @Test def testInfosetFree1(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />,
+      <dfdl:format ref="tns:GeneralFormat" />,
+      <xs:element name="root">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:element name="e" type="xs:int" maxOccurs="unbounded"
+              dfdl:length="1" dfdl:lengthKind="explicit" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+
+    val actualXML = TestInfosetFree.test(testSchema, "123".getBytes)
+
+    // all elements freed, including array elements and the array itself
+    val expectedXML =
+      <root freed="self" xmlns="http://example.com">
+        <e freed="self+array">1</e>
+        <e freed="self+array">2</e>
+        <e freed="self+array">3</e>
+      </root>
+
+    assertEquals(scala.xml.Utility.trim(expectedXML).toString, actualXML.toString)
+  }
+
+  @Test def testInfosetFree2(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />,
+      <dfdl:format ref="tns:GeneralFormat" />,
+      <xs:group name="hidden">
+        <xs:sequence>
+          <xs:element name="e" type="xs:int" dfdl:length="1" dfdl:lengthKind="explicit"
+            dfdl:outputValueCalc="{ 1 }" />
+        </xs:sequence>
+      </xs:group>
+      <xs:element name="root">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:sequence dfdl:hiddenGroupRef="tns:hidden" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+
+    val actualXML = TestInfosetFree.test(testSchema, "1".getBytes)
+
+    // all elements freed, including hidden elements
+    val expectedXML =
+      <root freed="self" xmlns="http://example.com">
+        <e freed="self">1</e>
+      </root>
+
+    assertEquals(scala.xml.Utility.trim(expectedXML).toString, actualXML.toString)
+  }
+
+  @Test def testInfosetFree3(): Unit = {
+    val testSchema = SchemaUtils.dfdlTestSchema(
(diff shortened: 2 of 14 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
```diff
@@ -417,39 +418,56 @@ sealed trait RegularElementUnparserStartEndStrategy
...
+          // cannot add content to a nilled complex element
+          UnparseError(One(erd.schemaFileLocation), Nope, "Nilled complex element %s has content from %s",
+            parentComplex.erd.namedQName.toExtendedSyntax,
+            newElem.erd.namedQName.toExtendedSyntax)
         }
 
+        // We are about to add a child to this complex element. Before we do
+        // that, if the last child added to this complex is a DIArray, and this
+        // new child isn't part of that array, that implies that the DIArray
+        // will have no more children added and should be marked as final, and
```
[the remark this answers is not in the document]
stevedlawrence: > So there's some confusion here around the role of isFinal in my mind.

They do feel separate, but are actually closely related. isFinal is only set for elements when unparseEnd is called, which means the element must have been fully unparsed and can be removed. The place where this sort feels question is with suspensions. The unparse process looks like this:

1. parseBegin - create an Element and add to infoset
2. Unparse body, potentially creating a Suspension
3. parseEnd - mark element as isFInal and free

So I think a concern is that we might unparse the body and create a suspension, but then immediately call parseEnd and free the element. However, creating a suspension clones the current infoset, so while the element is freed from the "main" infoset, the suspension still has a pointer to it and can do the normal processing.

> If we wait for an entire array to be isFinal before we free any of its child elements, then we cannot truly stream a large array. Freeing element with index N-1 upon unparse of element N seems essential to obtain streaming behavior.

We do not wait for an array to be final before freeing any of it's children. We free the children as soon as they are done unparsing, because those children will call unparseEnd, get marked as final, and they will be freed. Only when all those children are finished and we add something after the array is the arrary marked as isFInal and the array can be freed.

> I would expect we would pass the array index down to the freeChildIfNoLongerNeeded call, since without that it can't decide to release things based on index.

We do not currently free anything based on index. We free things as soon as possible, including children of arrays.

The one case were don't free things (in both parse and unparse) is when something is used in an expression. So large arrays are only a problem if they are use in expressions. And even then, we still only keep the things in those arrays that are used in expressions. So it's still wasteful, and large arrays can cause problems, but it's certainly less of an issue.

There is a proposal to add a new dfdlx property that gives hints to Daffodil to say that if an array with this property is used in an expression (and thus we can't free it's contents) then we only need to keep the most recent X elements. This isn't implemented yet. Perhaps this can be postponed for a future release?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
```diff
@@ -515,35 +567,64 @@ trait OVCStartEndStrategy
...
-
     // if an OVC element existed, the start AND end events were consumed in
     // unparseBegin. No need to advance the cursor here.
 
+    // ovcElem is finished, free it if possible. OVC elements are not allow in
+    // arrays, so we can directly get the diParent to get the container DINode
+    if (state.releaseUnneededInfoset) {
+      val ovcElem = state.currentInfosetNodeStack.pop
+      val ovcContainer = ovcElem.get.diParent
+      ovcContainer.freeChildIfNoLongerNeeded(ovcContainer.numChildren - 1)
```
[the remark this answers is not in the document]
stevedlawrence: Correct, that logic has not been imlemented in either parse or unparse. I don't think we can do that on a global basis because there are many different array uses where sometimes we can't get away with that. I think it needs to be per array. Mentioned above, but a proposal is to add a new dfdlx: property that can be placed on arrarys that are used in expression and Daffodil can use it has a hint only keep X previous array elements.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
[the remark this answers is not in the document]
mbeckerle: So, if I have a format which is a giant array, e.g., PCAP format, the elements of the array are all referenced by expressions because the header of each packet needs the length of the packet contents. So we'll never free anything (pretty much. We might be able to discard a few header elements in each packet, but we won't be able to discard the packet payload contents) when parsing or unparsing PCAP? If that's the case, then this isn't really solving people's streaming issue until this dfdlx extension property telling it to prune arrays is implemented? 
Or.... since the suspension to fill in the length of the packet in the packet header is evaluated, at that point do we mark everything final and so that we now know we can release it?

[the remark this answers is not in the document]
stevedlawrence: I'd have to do some testing, but I think this would actualy stream somewhat okay. The PCAP schema essentially looks like this:
```xml
<PCAP>
  <GlobalHeader>...<GlobalHeader>
  <Packet>...</Packet>
  ...
  <Packet>...</Packet>
</PCAP>
```
Named down steps are the only thing that cause elements to be marked as "referenced by an expression" and thus not freeable. And the PCAP schema, nothing actually references "Packet" in a down step. Lots of children of the Packet element do get referenced, but not the Packet itself. So when a Packet element is marked as isFinal, we will immediately free it. So I suspect we'll only ever have one Packet (and all of it's children) in memory at a time. Which should be very reasonble.

This does exlucding things like suspensions which might hold on to packets, but hopefully the earlier suspension evaluation will mitigate that.

This also doesn't handle the fact the there there is going to be a DIArray with size N where N is the number of packets. Most of the elements in that array will be null because they'll be freed, but it will still take up a good chunk of memory for large arrays. Note that a schema could minimize this a bit by "chunking" the array. E.g.
```xml
<xs:element name="PacketChunk" maxOccurs="unbounded">
  <xs:complexType>
    <xs:sequence>
      <xs:element name="Packet" maxOccurs="10000" ... />
    </xs:sequence>
  </xs:complexType>
</xs:element>
```
It's a bit ugly, but is a potential work around for very large arrays without having to compact them.

Note that we could add a feature to compact them after they grow to some size and have enough elements freed, but there are things like fn:count() and index offsets that need to be considered, so if we do want to do that, it's probably best left for another change.

I'm also not argueing that the extension to hint to clean up arrays isn't necessary. But I do think that this change does resolve some streaming issues without requiring additional extension properties be added.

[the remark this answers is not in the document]
mbeckerle: Ok. Thanks. This is what I didn't understand. I do understand the array chunking trick, and how that would perhaps work to minimize the array. But the real problem is not a large spine of an array, but the contents of the array, at least for now.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfoset.scala
```diff
@@ -74,12 +74,17 @@ object TestInfoset {
...
+   * them. Because this assumes tests will be inspecting the internal infoset
+   * for correctness, it sets the releaseUnusedInfoset tunable to false so that
+   * the infoset elements are not freed
    */
   def testInfoset(testSchema: scala.xml.Elem, infosetAsXML: scala.xml.Elem): (DIElement, Root, DaffodilTunables) = {
-    val schemaCompiler = Compiler().withTunable("allowExternalPathExpressions", "true")
+    val schemaCompiler =
+      Compiler()
+        .withTunable("allowExternalPathExpressions", "true")
+        .withTunable("releaseUnneededInfoset", "false")
```
mbeckerle, line 87: Do we have specific tests for the unparser's release mechanism that set this to true, exercise unparsing, and verify that the infoset doesn't have specific children in it at the end? Seems to me without that sort of a test, we could break this and not know it. 

At the end of unparsing, the infoset rooted in the initial UState should still be there, but all its children that aren't referenced in expressions should be gone. We need a test or a few to verify that things are dropped properly.

stevedlawrence: It's actually kindof hard to write a test to make sure everything is getting cleaned up appropriately. At the end of unparsing, the DIDocument should have a single child, the root element. If that root element isn't used in an expression it should get set to null. So at the end of a parse, the DIDocument child(0) should be null if not used in an expression. But that only tests that we were able to free the root element. It's possible things were broken so that no a single other element was freed and the only thing actually freed was the root element. And since the root elemen was freed, there's no way to know if any of it's children were freed--we no longer have any access to it or it's children.

Maybe the best way to ensure this is working correctly is to just develop tests that would fail with OutOfMemory errors if we weren't cleaning up infoset correctly. Those tests would likely be too large/take too long to keep in the Daffodil repo, but I believe @dfthompson is working on some tests that live in Owl's internal test infrastructure. Once those are done, we might want to find a place where those can be made public so others can test them.

mbeckerle:
How about a test mode? A settable flag in pstate/ustate and a mutable list.

Instead of just nilling the refs to things being released when the flag is set you also push them onto the list. At end of a test run we inspect the list to insure the right things have been put on the list?

Seems like small effort for a big increase in testability maintainability of this feature. Could be done for both parse and unparse.

stevedlawrence:
Doesn't seem totally unreasonable. I guess individual tests would then need to manually maintain a list of elements that are expected to be freed and compare against this built-in list? I can't think of a good way to automate it the process.

mbeckerle:
The infoset "chunks" on the list of released ones might still convert to xml elements via toXML. If so you can take the whole array, convert to an XML doc by wrapping an element around them all, and a test can then contain a corresponding expected XML chunk to compare to it.

stevedlawrence:
Not a bad idea, but I guess it still requires someone to manually create the expected XML. I was hoping for some other way things to just know what shoudl be release. But that's probably too difficult and fragile. I'll come up with a way to easy create the expected results. XML might be a good approach.

stevedlawrence:
Another thought, what if instead we make it so that if we would have freed a DINode but didn't then we modify some new state in the DNIode that says that it would have been freed. And when we output the infoset as XML we have to ability to configure it to include an attribute saying it would have been freed. Then we just have these tests disable freeing output as XML with this attribute, and then we can easily inspect the XML to determine which elements would have been freed. No concerns with elemetn children being null, and the resulting XML should look exactly like the infoset, just with extra attributes. Thoughts?

mbeckerle:
That seems like a better choice. We just mark the nodes and convert to an XML attribute.

stevedlawrence:
Ok, I've just added a commit that adds some state to track when an infoset element would have been released if it's disabled, adn the ability to output this state with the ScalaXMLinfosetOutputter. Also added some tests that verify that DINodes are being released as expected.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+    val actualXML = TestInfosetFree.test(testSchema, "1123".getBytes)
+
+    // all elements freed, execpted for fieldLen because it is used in an expression
+    val expectedXML =
+      <root freed="self" xmlns="http://example.com">
+        <fieldLen>1</fieldLen>
+        <field freed="self+array">1</field>
+        <field freed="self+array">2</field>
+        <field freed="self+array">3</field>
+      </root>
```
stevedlawrence, line 199: Note that this test is someone analagous to the PCAP situation. There's a header and an unbounded number of things that follow. And the header never directly references the array elements. So individual array elements are freed as they are completed, and the array is freed at the end of the parse/unparse. The header is never freed, but presumably it is small enough that it's okay. It certainly is in the case of pcap.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+    // are not currently freed. Our usedInAnExpression logic is not yet
+    // sophisticated enough to differentiate between how elements are used in
+    // an expression and how that affect releasability
+    val expectedXML =
+      <root freed="self" xmlns="http://example.com">
+        <fieldCount>3</fieldCount>
+        <field>1</field>
+        <field>2</field>
+        <field>3</field>
+      </root>
```
stevedlawrence, line 234: This test is maybe more analagous to some more common formats, where we have lots of expression, and output value calc's that can make streaming difficult. In this particular case, nothing can be freed. In this case, we would need a feature that allows for compacting arrays and handling count of elements as different type of acces than element indicies.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
mbeckerle: This is great. Exactly what I was hoping to have.

mbeckerle (approved):
+1

stevedlawrence:
Still need one more +1 before this can be merged. @jadams-tresys @olabusayoT @tuxji Any chance one of you could take a look at this?

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+
+    val actualXML = TestInfosetFree.test(testSchema, "3123".getBytes)
+
+    // Most elements are not freed. Both fieldCount and field are used in an
+    // expression. Even though only the count of the field array is needed, and
+    // we could theoretically free the field elements but no the array, they
+    // are not currently freed. Our usedInAnExpression logic is not yet
```
olabusayoT, line 225: "but not the array"*

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+   * the expected value.
+   */
+  def test(
+    schema: scala.xml.Elem,
+    bytes: Array[Byte]): scala.xml.Node = {
+
+    val compiler = Compiler()
+      .withTunable("releaseUnneededInfoset", "false")
+
+    val pf = compiler.compileNode(schema).asInstanceOf[ProcessorFactory]
```
tuxji, line 54: I don't think the .asInstanceOf[ProcessorFactory] is necessary, but I could be wrong.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+      fail("pf compile errors: " + msgs)
+    }
+    pf.sset.root.erd.preSerialization // force evaluation of all compile-time constructs
+    val dp = pf.onPath("/").asInstanceOf[DataProcessor]
+    if (dp.isError) {
+      val msgs = dp.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("dp compile errors: " + msgs)
+    }
+
+    // Parse the data, note that we do no set showFreedInfo here because
```
tuxji, line 66: *no -> not

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+              dfdl:lengthKind="explicit" dfdl:length="1" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+
+    val actualXML = TestInfosetFree.test(testSchema, "3123".getBytes)
+
+    // Most elements are not freed. Both fieldCount and field are used in an
+    // expression. Even though only the count of the field array is needed, and
+    // we could theoretically free the field elements but no the array, they
```
tuxji, line 224: *no -> not

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ElementUnparser.scala
```diff
@@ -515,35 +561,60 @@ trait OVCStartEndStrategy
...
+    state.currentInfosetNodeStack.push(One(ovcElem))
   }
 
   protected final override def unparseEnd(state: UState): Unit = {
-    state.currentInfosetNodeStack.pop
-
     // if an OVC element existed, the start AND end events were consumed in
     // unparseBegin. No need to advance the cursor here.
 
+    // ovcElem is finished, free it if possible. OVC elements are not allow in
```
tuxji, line 603: *allow -> allowed

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+            <xs:element name="field" type="xs:int" maxOccurs="unbounded"
+              dfdl:occursCountKind="expression" dfdl:occursCount="{ ../tns:fieldCount }"
+              dfdl:lengthKind="explicit" dfdl:length="{ ../tns:fieldLen[dfdl:occursIndex()] }" />
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+
+    val actualXML = TestInfosetFree.test(testSchema, "3123122333".getBytes)
+
+    // fieldCount and filedLen are not free, both are used in expressions.
```
olabusayoT, line 260: typo filedLen -> fieldLen

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
```diff
@@ -0,0 +1,280 @@
...
+    val compiler = Compiler()
+      .withTunable("releaseUnneededInfoset", "false")
+
+    val pf = compiler.compileNode(schema).asInstanceOf[ProcessorFactory]
+    if (pf.isError) {
+      val msgs = pf.getDiagnostics.map { _.getMessage() }.mkString("\n")
+      fail("pf compile errors: " + msgs)
+    }
+    pf.sset.root.erd.preSerialization // force evaluation of all compile-time constructs
+    val dp = pf.onPath("/").asInstanceOf[DataProcessor]
```
tuxji, line 60: Likewise, is the asInstanceOf really needed here?

tuxji (approved):
+1

Looks good to me.

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
olabusayoT, line 66: not*

olabusayoT (approved):
+1 looks great! Just a few typos in comments

daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetFree.scala
stevedlawrence: We did need it because this DataProcessor has an additional helper def unparse() method that accepts an OutputStream instead of the normal WritableByteChannel. But it's better to change this test to use the WritableByteChannel instead. All unnecessary asInstanceOf's are removed and typos fixed.

PR #438, opened 2020-10-15
Fix Legacy/SAX Discrepancy with NamespaceBindings

olabusayoT:
Fixes nullpointer exception error as can be seen in [this DAFFODIL-2383 comment](https://issues.apache.org/jira/browse/DAFFODIL-2383?focusedCommentId=17210164&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17210164) by making sure that n stops when it is null, and ensure we stop when the current mapping is equal to the TopScope, as there is an issue with mininmizeScope (DAFFODIL-2282) and nsbStart doesn't always have a mapping that'd equal nsbEnd

DAFFODIL-2383


mbeckerle (approved):
+1

stevedlawrence (approved):
+1

PR #439, opened 2020-10-15
Allow CLI unparse to unparse from a stream of data

stevedlawrence:
The CLI currently reads in the entire infoset into memory and unparses
from that. This means that the CLI cannot be be used to unparse very
large infosets due to memory limitations.

In some cases we do want to read the infoset into memory, such as with
the performance command to avoid IO latency, or with the --stream
option, which must scan the input looking for the end of each stremamed
infoset. We also must do this if we are unparsing from something like a
Scala XML Node, since we must read the entire infoset to convert it to a
Node.

But in cases where possible, we should not read the entire infoset into
memory, and instead just treat is as an InputStream and let the
InfosetInputter read in the infoset as needed.

This changes the infosetDataToInputterData function to accept either an
Array[Byte] or an InputStream and return an InputStream where possible.
The getInfosetInputter function then uses the appropriate InputStream
for the InfosetInputter, thus avoiding reading the entire infoset into
memory in some cases.

DAFFODIL-2412


mbeckerle (approved):
+1

jadams-tresys (approved):
+1

stevedlawrence:
Added some CLI tests should add more coverage to Main.scala, including this new code. I hate seeing that little red :x:

tuxji (approved):
+1

FYI, the codecov report (<https://codecov.io/gh/apache/incubator-daffodil/commit/ddcc7e7e7b15a01ca8211edcd893392f4ab3ccbf>) for this particular diff is a bit strange.  If I read the report correctly, tests aren't calling infosetDataToInputterData and getInfosetInputter with certain parameter values ("scala-xml" and "jdom" for both functions and "json" for the second function).  I don't see a reason to be concerned since the match logic is simple and IO classes are well tested but I'd like to know if my understanding of the report is correct.  It's strange that lines 795, 803, and 835 are green since they can be executed only inside code paths that are marked uncovered (red), however.  My interpretation is that these lines were incorrectly marked covered (if these lines were correctly marked covered, then lines 797, 804, and 837 should have been marked covered too).

stevedlawrence:
Yeah, I think you're reading the codecov right. It was correct that parts of this diff weren't covered, but how it showed that, as you say, doesn't feel right. If I run the commands to generate a coverage report locally, it feels a little more correct, but still not quite right. I think the challenge is just that Scala doesn't a bunch of stuff behind the scense, and the coverge tools just can't make an accurate represetnation of what's going on.

But it was correct that we didn't have CLI tests for different infoset types, and so no actual coverage for some of these changes I just made, and I think it was worth adding just to make sure we don't have any regressions in the future. The way this infosetDataToInputterData and getInfosetInputter is sort of fragile since they retun/accept AnyRef, so it's important that they have a little coverage.

PR #440, opened 2020-10-16
Added warning when multiple choice branches are empty

jadams-tresys:
It may not be obvious to the user that when a choice has multiple empty
branches only the first will ever be selected even if there is a
choiceBranchKey associated with each branch, as choiceDispatchKey is not
evaluated on unparse. In the test that has been added for this, this
behavior of always defaulting to the first empty branch results in a
multiple set error on a variable.

DAFFODIL-2375

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
@@ -86,11 +86,15 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
     // in the arriving infoset for unparse.
     //
     val optDefaultBranch = {
-      val optEmpty: Option[Term] =
-        groupMembers.find { gm =>
+      val optEmpty: Option[Term] = {
+        val emptyBranches = groupMembers.filter { gm =>
           val ies = gm.identifyingEventsForChoiceBranch
           ies.pnes.isEmpty // empty event list makes it the default, not simply isOpen
         }
+        if (emptyBranches.length > 1)
+          SDW(WarnID.MultipleChoiceBranches, "Multiple choice branches with no required elements detected and the infoset does not specify a branch, selecting the first branch for unparsing")
+        emptyBranches.headOption
+      }
       val optOpen: Option[Term] =
         optEmpty.orElse {
           groupMembers.find { gm =>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
@@ -1735,9 +1735,15 @@
 
   </tdml:defineSchema>
 
-  <tdml:parserTestCase name="choiceBranchVariables" root="root"
+  <!-- This test describes a situation where a choice is used to simply set a variable and nothing else.
+    In the choice of this test the branches only contain empty sequences that call setVariable on different
+    variables, depending on the branch. This works as expected during parsing, but because nothing gets
+    added to the infoset for these setVariable calls, on unparsing only the first branch of the choice is
+    ever used, which if the choice is part of an array, as it is in this test, results in the setVariable
+    call in the first branch being called multiple times, resulting in a double set error.
+  -->
+  <tdml:unparserTestCase name="multipleBranchesWithNoElementsSetVariableError" root="root"
     model="choice_branch_variables" description="Section 7 - setVariable - relative path expression - DFDL-7-126R">
-    <tdml:document>1,2</tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
         <root>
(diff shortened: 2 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -86,11 +86,15 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
...
     val optDefaultBranch = {
-      val optEmpty: Option[Term] =
-        groupMembers.find { gm =>
+      val optEmpty: Option[Term] = {
+        val emptyBranches = groupMembers.filter { gm =>
           val ies = gm.identifyingEventsForChoiceBranch
           ies.pnes.isEmpty // empty event list makes it the default, not simply isOpen
         }
+        if (emptyBranches.length > 1)
+          SDW(WarnID.MultipleChoiceBranches, "Multiple choice branches are empty, defaulting to first empty branch")
```
stevedlawrence, line 95: I think this error message could be more clear. First, "defaulting" has a meaning in the DFDL spec that is different from what this is saying, so I think we should avoid using that word. The word "empty" feels a bit ambiguous as well. It also doesn't poitn out which other empty branches are being ignored. That could be helpful information. From Mike's bug, mentioning branches that can't be unparsed might be a more helpful approach, maybe something like:

```scala
emptyBranches.tail.foreach { branch =>
  branch.SDW("Choice branch with no elements can never be unparsed")
}
```
So we get a warning for each branch that can never be used to unparse for each branch. It means there will be multiple warnings if there are multiple empty branches, but it makes it easier for the user to identify which branches are empty and fix the issue.

jadams-tresys: So its not that a branch with no elements won't be unparsed. What happened originally was I had this choice that had empty sequences that would simply call setVariable on different variables, depending on the branch. What I didn't realize was that on unparse, there is nothing in the infoset to determine which branch was taken and the choiceBranchKey is not evaluated, so it defaults to the first empty branch every time. This resulted in the multiple set variable error.

stevedlawrence: So there is only one empty branch in your schema?

stevedlawrence: Sounds to me like you're saying what would have made your problem more clear would have been to get warning that choiceDispatch is ignored on unparsing? I feel like that's type of thing is going to create too many warnings. Just because the DFDL spec might not be intuitive doesn't necessarily mean we should create warnings.

Or is the issue that you had multiple empty branches, and you thought in the case of empty branch choicesDispatch was used to disambiguate? If that were the case, I wonder if the suggested warnings would have led you to the realization that choiceDispatch is ignored?

jadams-tresys: There are multiple empty branches in the schema.  choiceDispatch not getting evaluated was definitely unexpected to me, but I understand that may generate too many warnings to be practical.  Using choiceDispatch to disambiguate would have been nice, but I the spec isn't really clear what expected behaviour is when unparsing a choice with empty branches (or at least I can't find anything about it in the spec, its definitely not mentioned in 15.1.3 Unparsing Choices section).

I'm not sure what a good warning message would be in this rather particular case, maybe a warning whenever we are defaulting to the first empty branch of the choice.

stevedlawrence: According to [DAFFODIL-2324](https://issues.apache.org/jira/browse/DAFFODIL-2324), there is an erratum that clarifies the right behavior for how empty branches should be unparsed. That might not be folded into https://daffodil.apache.org/docs/dfdl/ yet. I don't think the spec on our website has been updated in a while.

My concer is that that "defaulting to the first empty branch" might not be clear, especially since "defaulting" means something different. 

Looking to the Erratum for inspriation, 5.60 says:
>Section 15.1.3 Insert at end of 1st paragraph. 
"If the next element to unparse does not identify any branch of the choice, or there is no next element to unparse, then there must be a choice branch with no required elements and the first such branch would be selected for unparsing. A choice branch could consist only of a nest of model groups with no actual element content or only optional element content."
>
>Section 9.4.3.2 Insert at end of final paragraph.
"If no choice branch is selected, then there must be a choice branch with no required elements, and the first such branch would be selected."

So rather than "default" you really need something like "element does not identify a branch" or something. I'm not sure of a way to say that cleanly. Which is way rather than say a branch is the "default" (which isn't accurate and potentially confusing), I just suggested saysing X branches can't be unparsed. Is sort of like one is the negative of the other. Open to other suggestions though. Succinct but descriptive diagnostic messages aren't easy.

jadams-tresys: I've pushed up a change with what I think is a more clear warning message:
Multiple choice branches with no required elements detected, selecting the first branch for unparsing

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -1752,7 +1751,18 @@
...
       </tdml:dfdlInfoset>
     </tdml:infoset>
-  </tdml:parserTestCase>
+    <tdml:warnings>
+      <tdml:warning>Multiple choice branches</tdml:warning>
+      <tdml:warning>no required elements</tdml:warning>
+      <tdml:warning>selecting the first branch</tdml:warning>
+    </tdml:warnings>
+
+    <tdml:errors>
```
mbeckerle, line 1760: Please add block comment to this TDML test because this is an obscure enough case that it needs some explanation. 
Suggest name of test should be "multipleBranchesWIthNoElementsDoubleWrite" or something conveying the complexity of the situation.

mbeckerle (approved):
+1. Just requesting a comment explaining the test a bit more.

tuxji (approved):
+1 with Mike's suggestions too

stevedlawrence:
I'm still a little concerned this error message could be confusing. "selecting the first branch for unparsing" could be interpreted as it's always going to unparse the first branch of a choice, regardless of anything else. Which isn't the case. It's going to unparse the first branch with no elements, and it's only going to do that if the infoset does not identify which branch to unparse.

jadams-tresys:
I've changed the warning message to be a little most specific and updated the test name and added a comment describing the test case.

stevedlawrence (approved):
+1

PR #441, opened 2020-10-18
Fix thread-unsafe Logging trait usage

chlin501:
Based on the comment in the ticket to remove the Logging mixin from
DataProcessor, and delete the two calls to log from the DataProcessor
class.

DAFFODIL-2216


mbeckerle (approved):
+1 Thanks for this fix.

tuxji (approved):
+1

Looks good.

PR #442, opened 2020-10-21
Enable pattern facet to use PUA characters

mbeckerle:
Enables patterns to validate the DFDL infoset
characters that are otherwise XML-Illegal such as
0xB or NUL (0x0), by expressing them as &amp;#xE00B; or
&amp;#xE000;

Test test_nulPattern1 verifies this.

DAFFODIL-2363

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
@@ -21,6 +21,7 @@ import java.math.BigInteger
 import scala.xml.Node
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.dpath.NodeInfo.PrimType
+import org.apache.daffodil.xml.XMLUtils
 
 trait Facets { self: Restriction =>
   import org.apache.daffodil.dsom.FacetTypes._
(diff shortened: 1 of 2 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
```diff
@@ -18,9 +18,11 @@
 package org.apache.daffodil.dsom
 
 import java.math.BigInteger
+
```
mbeckerle, line 21: Remove blank line. Not sure at all how this got inserted.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
```diff
@@ -129,10 +131,45 @@ trait Facets { self: Restriction =>
...
+          // Before processing a regex of these characters in Daffodil's pattern facet
+          // validation, we must remap these PUA characters back to the originally
+          // intended code points, since that's what the Infoset strings will contain.
+          //
+          // Consider the character code 0xB. This is illegal in XML v1.0 documents.
+          // A DFDL Schema is an XML Schema, which is an XML document; hence, we cannot
+          // use the character with code 0xB directly, nor can we use an XML numeric
+          // character entity like &#xB; for it. The character is simply disallowed in
+          // XML, including DFDL schemas. Hence, we mention 0xB by using a remapping of it
+          // to the PUA area character 0XE00B, which we express by &#xE00B;
```
mbeckerle, line 156: Lower case "x"

olabusayoT (approved):
Looks great, +1! Comment about additional line insertion.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
```diff
@@ -129,10 +131,45 @@ trait Facets { self: Restriction =>
...
+          // Hence a pattern regex like "[&#xE00B;&#x20;0-9a-zA-Z]" will match
+          // the character with char code 0xB (remapped from E00B), as well as spaces,
+          // and alphanumeric characters.
```
stevedlawrence, line 159: Just so I'm clear, ``&#x0B;`` is disallowed in an XSD pattern regex because XSD is just a special XML. And based on Daffodil's PUA handling, one must instead use ``&xE00B;``? Then this transforms that into a regex that matches 0x0B? And our internal infoset value that we check this pattern against hasn't been mapped to PUA at the time we run this regex, because we only transform infoset values to PUA right before converting to XML with an XML infoset outputter?

stevedlawrence (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
mbeckerle: Yeah, it's a bit odd, but in any XML context, including the pattern regex, if you want to talk about an XML-illegal character you have to use the PUA character instead. A DFDL schema is an XML document, so the pattern facet regex has to do this also.

PR #444, opened 2020-10-21
Use AbsentRep instead of MissingSeparator

IanCarlsonOwl:
In cases where a element in a sequence's prefix separator is not found
use the AbsentRep error instead of MissingSeparator. This causes
us to reset to a PoU rather than just continuing - throwing
out the error from finding out of scope separators.

DAFFODIL-2399


jadams-tresys (approved):
+1 Assuming this change doesn't break any of the schemas. I can't remember if they are tested by the build bots or not.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceChildParseResultHelper.scala
```diff
@@ -279,7 +279,7 @@ class PositionalTrailingGroupSeparatedSequenceChildParseResultHelper(
     if (isModelGroupRepPossiblyZeroLength &&
       !isModelGroupRepNonZeroLength) {
       pstate.setSuccess()
-      ParseAttemptStatus.MissingSeparator
+      ParseAttemptStatus.AbsentRep
```
mbeckerle, line 282: The JIRA ticket has a small test that reproduces this problem. That test needs to be part of this change set.

mbeckerle (changes requested):
Please add the test.

stevedlawrence:
Does this also mean that there is a bug in the TDML Runner where it wasn't detecting tests that returned both an infoset *and* diagnostics containing errors? This would be good to fix as part of this to ensure we don't accidentally add back this kind of issue.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceChildParseResultHelper.scala
mbeckerle: We need to run HL7 regression tests for this, as well as EDIFACT and other DFDL schemas that exercise this aspect of the code more thoroughly. 

I'm going to need the test, breakpoint this line, and walk backwards to where this is used to figure this out.

IanCarlsonOwl: I have added the test I used to vet this change under the UserSubmitted tests as DFDL_2399.

IanCarlsonOwl:
> Does this also mean that there is a bug in the TDML Runner where it wasn't detecting tests that returned both an infoset _and_ diagnostics containing errors? This would be good to fix as part of this to ensure we don't accidentally add back this kind of issue.

I scanned through the TDMLRunner, and I don't see any logic preventing the test from returning both errors and an infoset. There is logic to prevent the TDMLRunner from being started while _expecting_ both however.

stevedlawrence:
> I scanned through the TDMLRunner, and I don't see any logic preventing the test from returning both errors and an infoset. There is logic to prevent the TDMLRunner from being started while expecting both however.

It would probably be a good idea to at that, that would show if there are any other tests that have this same issue.

IanCarlsonOwl:
> > I scanned through the TDMLRunner, and I don't see any logic preventing the test from returning both errors and an infoset. There is logic to prevent the TDMLRunner from being started while expecting both however.
> 
> It would probably be a good idea to add that check, that would show if there are any other tests that have this same issue.

Adding a check to enforce no diagnostics - i.e. Assert.invariant(diagObjs.length == 0) in TDMLRunner.scala line 938 or so - causes 76 test failures. It catches a lot of ValidationError type messages, but also a few parse errors.

stevedlawrence:
I think checking that there are no diagnostics is to strict of a check. Things like validation errors and warnings are diagnostics that should be allowed. We really want to ensure that there are no error diagnostics. Each diagnostics has an isError that says whether it's warning vs error.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SeparatedSequenceChildParseResultHelper.scala
mbeckerle: I've traced this in the code. I'm convinced AbsentRep is the right return value here. 

I think this code does need future work to clean up some confusion. There are code paths being walked here that when written clearly had elements in mind, but in this test case, the schema component in consideration is a sequence group, so that suggests a strong possibility cases are missing. 

So long as a full regression on all known DFDL schemas works, I'm happy with this.

mbeckerle (approved):
+1 if full regression on all known schemas works.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -272,6 +277,10 @@ abstract class SequenceParserBase(
               // the sequence succesfully
               isDone = true
               pstate.setSuccess()
+              // If we're masking the failure, we don't want the error dianostics
+              // to flow up. Restore the diagnostics from before the parse
+              // attempt
+              pstate.diagnostics = diagnosticsBeforeAttempt
```
stevedlawrence, line 283: I've added an additional test that I think shows that discriminated unordered sequences are just broken, even before these changes. I discovered this when trying to figure out why there's a single test failure with these changes.

I think the issue is that unordred sequences are implemented by repeatedly parsing a ChoiceParser. When the ChoiceParser fails to parse a branch, then the sequence parser assume we are done with with the unordered sequence. When that happens, we hit this failure case, throw away that failure and the "all alternatives failed" diagnostics, and continue parsing like normal. The ChoiceParser will have reset us back to the the PoU, so things just work.

However, say we have a discriminated unordred sequence (e.g. initiated content). Now let's assume an initiator was found. This will resolve the PoU created in the ChoiceParser. Now say there is a failue parsing that discriminated branch. The PoU was resolved so we will not attempt any alternatives of the unordered sequence/ChoiceParser and we immediately create a parse error. We then return back to here, and subsequently ignore that error and associated diagnostics. And because the PoU was resolved, we don't even backtrack and will just continue parsing whereever the choice branch left off.

The new test shows this behavior. Even though an assertion is created, it is ignored and the parse finishes. This is incorrect.

I'm not sure yet what the right fix is, but I think things are just totally broken with discriminated unordred sequences. It feels to me like maybe we can't reuse the ChoiceParser. I think the sequence parser needs to be the one to create the PoU associated with each alternative of the unordered sequence rather than relying on the ChoiceParser. That way it can know if the PoU was resolved or not, and then correctly end the sequence or propogate an error upwards. Right now, the ChoiceParser does everything so we don't have enough information in the SequeneParser to act correctly.

Note that the SequenceParser does have some logic around UnorderedDiscriminatedFailure, but I think that's just broken. The case above on line 273 is never actually hit in codecov. I'm guessing that this stuff broke when we change how PoU's work.

Thoughts?

mbeckerle: One thought is we should split this into two issues. 

One is the fact that diagnostics are being created and retained despite backtracking, and that one is important to fix. 

The other is a bug with unordered sequences, and the importance of that may not be as high - depends on whether there are DFDL Schemas depending on it and are broken by this. 

VCard is the only published DFDL schema I can find that uses unordered sequences.

An unordered sequence is a lot like an array of choice, but yeah, my guess based on your note above is we can't directly reuse choice parser, but must make a variant of it (sharing traits/bases perhaps) that has somewhat different behavior for discriminators since logically an unordered sequence does not have two nested PoUs, just one per child element instance.

stevedlawrence: I'll do a regression test and see if there are any failures with these changes. If not, I agree that it's probably worth merging this and open a new issue for discriminated unordred sequences. I think that combination is rare enough that it's unlikely that any schemas use the feature.

stevedlawrence: All my regression tests pass except for vCard, which has one test that runs into this exact issue. vCard has this exact issue--unordered sequences with initated content

So if a unordered sequence branch fails in vCard, we lose the diagnostic about why that initiated content branch failed. We will then continue trying to parse after the unorderd sequence fails (which is is this second pre-existing bug). In this case, we run into another parse error. Because we threw away the error we don't get the old helpful error message.

So this fixes the bug related to incorrectly showing diagnostics, at the expense of sometimes not showing diagnostics, and sometimes not even failing a parse when we should. I'll create a bug for this new issue. Seems pretty important to fix, but considering it only affects this one case, I'm not sure it's a blocker.

stevedlawrence: Created [DAFFODIL-2512](https://issues.apache.org/jira/browse/DAFFODIL-2512) for this issue.

stevedlawrence:
I've commented out the tests failing due to DAFFODIL-2512, so all tests now pass. I've removed the WIP tag. Since it's been a while since, I'd like for this to get a re-review before squash/merge.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -261,6 +262,10 @@ abstract class SequenceParserBase(
               // So we mask the failure, and exit the sequence successfully
               pstate.setSuccess()
               isDone = true
+              // If we're masking the failure, we don't want the error dianostics
+              // to flow up. Restore the diagnostics from before the parse
+              // attempt
+              pstate.diagnostics = diagnosticsBeforeAttempt
             }
 
             // We successfully parsed a discriminator, but failed to parse the discriminated content.
```
tuxji, line 271: I noticed that this match expression has 4 cases and I looked at each of them.  The first 3 cases set isDone = true, but we've added pstate.diagnostics = diagnosticsBeforeAttempt to only the 1st and 3rd cases.  Should the 2nd case (which begins on line 271) also have pstate.diagnostics = diagnosticsBeforeAttempt added after isDone = true?

tuxji (approved):
+1 with a question below

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -261,6 +262,10 @@ abstract class SequenceParserBase(
               // So we mask the failure, and exit the sequence successfully
               pstate.setSuccess()
               isDone = true
+              // If we're masking the failure, we don't want the error dianostics
+              // to flow up. Restore the diagnostics from before the parse
+              // attempt
+              pstate.diagnostics = diagnosticsBeforeAttempt
```
mbeckerle, line 268: This would seem to be needed everywhere we do pstate.setSuccess().  So I did a search for these. These are in fact the only two places where this is needed. :+1:

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
stevedlawrence: I think the only place it's needed is when we do ``pstate.setSuccess()``. That's where we are essentially ignoring any previous errors, and so we also need to remove diagnostics that came from those errors.

I think part of the fix to DAFFODIL-2512 might want to change how this works. Something like rather than overriding failed state and diagnostics, maybe we want a PoU and we just reset back to it. But that might undo other state, like infoset stuff, that we do want to keep around. So it's not so obvious what the right thing to do is here.

mbeckerle (approved):
+1

PR #446, opened 2020-10-23
Update all dependencies to latest versions

stevedlawrence:
- Updated all silences to match the latest and verified that no
  dependencies has significant license changes
- Fix deprecation warnings with Apache Commons NullOutputStream
- New ICU version fixed some issues and changes some behaviors. It now
  correctly unparses INF, -INF, and NaN according to the inf/nan rep
  properties so we do not need to manually handle it. It now also
  correctly obeys optional decimal points in some patterns. When
  unparsing, it also now rounds floating point numbers if the pattern
  does not contain decimal separators. Tests added to verify this
  behavior in case it changes again.
- Updates sbt to latest version, which now requires testOptions to
  maintain current test output behavior

DAFFODIL-2269


build.sbt
```diff
@@ -216,7 +218,7 @@ lazy val libManagedSettings = Seq(
       schemas.map { schema =>
         val out = outdir / "org" / "apache" / "daffodil" / "xsd" / schema.getName
         IO.copyFile(schema, out)
-        stream.log.info("Generated %s".format(out))
+        stream.log.info("generated %s".format(out))
```
stevedlawrence, line 221: This was done just to make output more consistent with sbt output. They've starting using lowercase in lots of places.

daffodil-cli/bin.NOTICE
```diff
@@ -106,18 +113,11 @@ Scala Parser Combinators (lib/org.scala-lang.modules.scala-parser-combinators_2.
...
   scala-xml includes software developed at
   LAMP/EPFL (https://lamp.epfl.ch/) and
   Lightbend, Inc. (https://www.lightbend.com/).
-
-Woodstox (lib/com.fasterxml.woodstox.woodstox-core-5.1.0.jar)
-  This product currently only contains code developed by authors
-  of specific components, as identified by the source code files.
-
-  Since product implements StAX API, it has dependencies to StAX API
-  classes.
```
stevedlawrence, line 123: This appears to be a notice that was only applied to old version of woodstox. Newer versions do not include a notice that I can find.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -102,24 +104,6 @@ import org.rogach.scallop.ArgType
...
-    //do nothing
-  }
-  override def write(b: Array[Byte], off: Int, len: Int): Unit = {
-    //do nothing
-  }
-  override def write(b: Int): Unit = {
-    //do nothing
-  }
-}
-
```
stevedlawrence, line 122: Use Apache IO NullOutputStream instead of creating our own.

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
```diff
@@ -4288,4 +4288,26 @@
...
+    <tdml:document>
+      <tdml:documentPart type="text">11</tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <tnp96>10.9</tnp96>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+
+  </tdml:unparserTestCase>
```
stevedlawrence, line 4311: ICU changed the behavior for how floating point numbers are formatted/unparsed when the textNumberPattern does not contain a decimal separator. The previous behavior was to unparse the number with the separator. New behavior rounds the floating point number. Note that when parsing it will still accept floating point numbers if the pattern does not contain a decimal separator. This means patterns without decimal separators and the data contains a deciaml point may not round trip correct. See textStandardFloatPatternNoSeparators1 above for the parse case.

daffodil-test/src/test/resources/org/apache/daffodil/section17/calc_value_properties/outputValueCalc2.tdml
```diff
@@ -494,7 +494,7 @@
       </tdml:dfdlInfoset>
     </tdml:infoset>
     <tdml:document>
-      <tdml:documentPart type="text">XYZ5.0X038.0Y41.23Z008</tdml:documentPart>
+      <tdml:documentPart type="text">XYZ5X038.0Y41.23Z008</tdml:documentPart>
```
stevedlawrence, line 497: ICU change behabior here, the text pattern here has option decimal point for the first number (5.0e03). The pattern says the .0 is optional, and so is now correctly left off when the value is zero.

tuxji (approved):
+1

Looks great!

Oops, just noticed that all checks failed with at least one test failure.

```
[info] Test org.apache.daffodil.dsom.TestPolymorphicUpwardRelativeExpressions.testPoly2 started
Error:  Test org.apache.daffodil.dsom.TestPolymorphicUpwardRelativeExpressions.testPoly2 failed: java.lang.ClassCastException: cannot assign instance of scala.collection.immutable.List$SerializationProxy to field org.apache.daffodil.dsom.DPathCompileInfo.elementCompileInfos of type scala.collection.Seq in instance of org.apache.daffodil.dsom.DPathCompileInfo
```

stevedlawrence:
Looks like a single test failing in github actions (not on my local tests) with the exception:
```
020-10-23T17:23:16.9625895Z [error] Test org.apache.daffodil.dsom.TestPolymorphicUpwardRelativeExpressions.testPoly2 failed: java.lang.ClassCastException: cannot assign instance of scala.collection.immutable.List$SerializationProxy to field org.apache.daffodil.dsom.DPathCompileInfo.elementCompileInfos of type scala.collection.Seq in instance of org.apache.daffodil.dsom.DPathCompileInfo, took 0.264 sec
2020-10-23T17:23:16.9629730Z [error]     at java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2201)
```
Seems like some weird serialization error. Found similar reports in sbt, and they closed the issue as "won't fix", claiming the solution is to set ``fork in Test := true``. I'm not sure we want to fork our tests, so perhaps the solution is to stick with sbt 1.3 for now. I'll undo the sbt upgrade change.

stevedlawrence:
Ok, looks like on Scala 2.12.12 there is some serialization issue, that doesn't appear to have an easy fix. Something about different classloaders, likely caused by SBT.

On sbt 1.4.1, there is some race condition with TestDaffodilXMLLoader. Skimming the tests, the way it creates/deletes files looks somewhat fragile and doesn't surprise me that there is a race condition. I'll look more into the TestDaffodilXMLLoader issue next week.

stevedlawrence:
I've removed the tests since they are clearly not thread safe. We can add new tests if we can determine a reasonable way to test the same functionality.

daffodil-test/src/test/scala/org/apache/daffodil/xml/test/unit/TestDaffodilXMLLoader.scala
```diff
@@ -1,382 +0,0 @@
...
-   * This test insures that a user catalog doesn't get priority over
-   * our internal catalog by defining a catalog and schema that uses
-   * one of our namespaces.
-   *
-   * Our dafint.xsd defines testElement and the line and col attributes
-   * must be type xs:int. The data here has dafint:line with value 'A',
-   * so if the built-in schema is used, it will find this error. So the
-   * test passes if the error is detected.
-   */
-  @Test def testThatUserCatalogCannotOverrideDaffodilInternalCatalog1(): Unit = {
```
mbeckerle, line 123: I think we do need a way to test this, or XML catalogs are just going to stop working at some point. By deleting these tests, instead of commenting them out with // TODO type comments around them, they are out of site/out of mind, and will be forgotten by the time things break. 

I'm fine with removing these tests if we either book a JIRA ticket to add testing of this and include a git hash that still has these tests in it, so they're easily retrieved, OR, just comment these out and put //TODO. Either way.

mbeckerle (approved):
+1 with resolution about not losing track of the tests per the one comment I added.

daffodil-test/src/test/scala/org/apache/daffodil/xml/test/unit/TestDaffodilXMLLoader.scala
stevedlawrence: Agreed. I'll create a ticket to add CatalogManager.properties tests, and reference a commit hash that includes these tests as good candidates once we determine a reasonable way to test this.

stevedlawrence: Created [DAFFODIL-2424](https://issues.apache.org/jira/browse/DAFFODIL-2424) to track the issue of adding back tests.

PR #448, opened 2020-10-27
Fix issue unparsing large blobs

jadams-tresys:
With large blob files it is possible that the blob size will be greater
than Int.MaxValue, which was resulting in an overflow and causing the
check to ensure that we have written out all bytes to fail. Simply
moving the byte length int conversion to the section of code for
non-blob files prevents this issue from occurring.

DAFFODIL-2425

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
@@ -845,107 +845,70 @@ class DirectOrBufferedDataOutputStream private[io] (
   }
 
   /**
-   * Convenience methods that temporarily set and (reliably) restore the bitLimit.
-   * The argument gives the limit length. Note this is a length, not a bit position.
-   *
-   * This is added to the current bit position to get the limiting bit position
-   * which is then set as the bitLimit when
-   * the body is evaluated. On return the bit limit is restored to its
-   * prior value.
-   * <p>
-   * The return value is false if the new bit limit is beyond the existing bit limit range.
-   * Otherwise the return value is true.
-   * <p>
-   * The prior value is restored even if an Error/Exception is thrown. (ie., via a try-finally)
-   * <p>
-   * These are intended for use implementing specified-length types (simple or complex).
-   * <p>
-   * Note that length limits in lengthUnits Characters are not implemented
-   * this way. See fillCharBuffer(cb) method.
+   * Delivers the bits of this DOS's buffer into the given directDOS's output stream.
+   * Deals with the possibility that the directDOS ends with a fragment byte, or
+   * the bufDOS does, or both. Handles file based buffers as well.
    */
-  // private def withBitLengthLimit(lengthLimitInBits: Long)(body: => Unit): Boolean = macro IOMacros.withBitLengthLimitMacroForOutput
-
-  protected def deliverContent(directDOS: DirectOrBufferedDataOutputStream, finfo: FormatInfo) = {
+  protected def deliverContent(directDOS: DirectOrBufferedDataOutputStream, finfo: FormatInfo): Unit = {
     val bufferNBits = this.relBitPos0b // don't have to subtract a starting offset. It's always zero in buffered case.
     val bufOS = this.bufferingJOS
+    val nBytes = (bufferNBits / 8)
+    val nFragBits = this.fragmentLastByteLimit
 
-    if (directDOS.isEndOnByteBoundary && this.isEndOnByteBoundary) {
-      // no fragment bytes anywhere - just take the bytes
-
-      val nBytes = (bufferNBits / 8).toInt
-      val nBytesPut = {
-        if (bufOS.isFile) {
-          bufOS.close
-          val nBitsPut = try {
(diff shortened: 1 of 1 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -885,8 +885,10 @@ class DirectOrBufferedDataOutputStream private[io] (
               bufOS.getFile.delete()
           }
           nBitsPut / 8
-        } else
-          directDOS.putBytes(bufOS.getBuf, 0, nBytes, finfo)
+        } else {
+          Assert.usage(nBytes <= Int.MaxValue)
```
stevedlawrence, line 889: I think this shoudl be an ``Assert.invariant``? My understanding is that ``Assert.usage`` has the implication that we're just not following our own assumptions about a function or something. For example, a function can only be called if isError is false, so we might have ``Assert.usage(!isError)``. It's still possible to call a function when things are an error, but we shouldn't do that.

But ``Assert.invariant`` implies that it shouldn't even be possible, not matter what we do. In this case, the size of a non-file backed buffer should never be larger than IntMax because a non-file backed buffer is backed by an ArrayBufffer, which can never be larger than IntMax. So nBytes being larger than IntMax should be impossible.

Also, below this in the else block which I think starts with the exact same code with just different variable names (and missing the new Assert.usage you've added). While we are here, I wonder it it makes sense to move this chunk of code that writes whole bytes out of the if-else blocks and have it always run so we remove this duplicate logic. Then there's just the if-statement forhandling fragment bytes if they exist?

stevedlawrence (approved):
+1, with suggestion to maybe remove duplicate code

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: So funnily enough, I originally had it as an Assert.invariant, but then I saw elsewhere in the code where we were doing the same Int.MaxValue check, it was using Assert.usage and I changed my code to match.  I'll update my code to be invariant, but would you want me to update the other assert too?

I'll also remove the duplicate code too, missed that when I was working on this yesterday.

stevedlawrence: I assume you're talking about:
```scala
abstract class LocalBuffer[T <: java.nio.Buffer] {

  def getBuf(length: Long) = {
    Assert.usage(length <= Int.MaxValue)
    ...
  }
}
```
In this case, I think Assert.usage is correct because it's enforcing an assumption about the parameter value, and ensuring uses of this function don't break this assumption. So Assert.usage is sortof saying "XYZ shouldn't be used this way", whereas Assert.invariant says "XYZ shouldn't be possible". It's fairly pretty subtle difference, and I think these two checks are slightly different.

mbeckerle: Assert.usage is for incorrect usage of the API of a method/function, i.e., the caller is not "using" it correctly. Generally this is for checking args, but also checking that state is proper for the use of the method. A good example is next/hasNext. You are not supposed to call next() unless hasNext() is true. So a next() implementation can do Assert.usage(this.hasNext()).  Assert.invariant is just stating that at this point in the code this condition is an invariant, i.e., it is supposed to be true. It is more things that are deeper than arg or pre-condition checking of a call. One writes Assert.invariant in order to make clear the assumptions of the code, particularly those that are subtle, and so might get changed accidently as code is maintained.

tuxji (approved):
+1 with a documentation request

The deliverContent function you're changing is preceded by a long Scaladoc comment that seems to apply to a commented out one-line function, so there is no comment which actually describes the deliverContent function itself.  This function is one of the longest functions in the file so it should have a comment explaining what it's used for and how it works.  Can reviewers like Steve and Mike help draft a comment, and can we delete the preceding Scaladoc comment and commented out one-line function since it looks like dead code that should be deleted?  Also, deliverContent should have a Unit type.

jadams-tresys:
I've pushed up changes regarding the logic of the deliverContent function, including the use of Assert.invariant and added the Unit type to the function.  I have not added a Scaladoc comment, as this function is now pretty straightforward and removing the duplicated code makes it fairly small.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -865,87 +865,66 @@ class DirectOrBufferedDataOutputStream private[io] (
    */
   // private def withBitLengthLimit(lengthLimitInBits: Long)(body: => Unit): Boolean = macro IOMacros.withBitLengthLimitMacroForOutput
```
mbeckerle, line 866: Remove commented code and prior block comment no longer relevant.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
jadams-tresys: Makes sense.

PR #449, opened 2020-10-28
Tweaks to release candidate script, mainly to fix building Windows MS…

stevedlawrence:
…I installer

- Commit 5061ed004f updated the SBT native packager plugin we use to
  build artifacts like the RPM and MSI installers. That plugin update
  changed where it looks for the candle.exe and light.exe executables
  used to build the MSI, so we need to change how we modify WiX so that
  it runs with wine.
- Removes the installation of .NET 4.5. It is unclear why that was
  originally needed, but it now causes the MSI build to fail with it
  installed. Perhaps building the container now installs an updated
  version of wine that comes with a working .NET version
- Fixes a typo when displaying a git clone command in the release
  candidate script
- Adds sbt-native-packager "maintainer" option globally rather than just
  for RPM and Windows to remove "missing maintainer" warning when
  building zip/tar
- Modifies release candidate script to display available signing keys
  for ease of use. Required moving the gpg timeout change to earlier in
  the script

DAFFODIL-2428

### DIFF
--- a/containers/release-candidate/Dockerfile
+++ b/containers/release-candidate/Dockerfile
@@ -22,8 +22,12 @@ WORKDIR /root
 COPY setup-container.sh /root/
 RUN /root/setup-container.sh
 
-COPY wix_wine.sh /root/wix311/\\bin\\candle.exe
-COPY wix_wine.sh /root/wix311/\\bin\\light.exe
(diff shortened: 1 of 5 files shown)


mbeckerle (approved):
+1

tuxji:
+1 as a top-level "Conversation" comment

PR #450, opened 2020-10-30
Improve newVariableInstace and defineVariable defaultValue expressions

jadams-tresys:
This commit enables newVariableInstance's to have non-constant default
value expressions.  These expressions are evaluated when the
newVariableInstance is created.

DAFFODIL-2352

This is WIP because dfdl:defineVariable defaultValue's also allow for expressions referencing other variables with default values (as long as they evaluate to a constant) and that currently causes a cirularity error to occur (see the commented out test case). This may involve so more low-level changes to 1) Get rid of the circularity error, and 2) Perform more thorough checking to see if the referenced variable's defaultValue expression evaluates to a constant.  I feel this may warrant a separate ticket, but wanted to bring this up for discussion and possibly allow the newVariableInstance fix to get into the 3.0.0 release.

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/util/MStack.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/util/MStack.scala
@@ -182,7 +182,7 @@ object MStackOfAnyRef {
  * things.
  */
 protected abstract class MStack[@specialized T] private[util] (
-  arrayAllocator: (Int) => Array[T], nullValue: T) extends Serializable {
+  arrayAllocator: (Int) => Array[T], nullValue: T) {
 
   private var index = 0
   private var table: Array[T] = null
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
@@ -491,6 +491,10 @@ class DataProcessor private (
       try {
         Assert.usageErrorUnless(state.dataInputStreamIsValid, "Attempted to use an invalid input source. This can happen due to our position in the input source not being properly reset after failed parse could not backtrack to its original position")
 
+        // Force the evaluation of any defineVariable's with non-constant default
+        // value expressions
+        state.variableMap.forceExpressionEvaluations(state)
+
         this.startElement(state, p)
         p.parse1(state)
         this.endElement(state, p)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
@@ -29,11 +29,12 @@ import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
 import org.apache.daffodil.infoset.RetryableException
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe.Nope
-import org.apache.daffodil.util.MStackOf
+import org.apache.daffodil.util.PreSerialization
+import org.apache.daffodil.util.TransientParam
 import org.apache.daffodil.xml.{GlobalQName, UnspecifiedNamespace, NamedQName, RefQName}
 import org.apache.daffodil.processors.parsers.PState
 
-import scala.collection.mutable.Map
+import scala.collection.mutable.{Map, ArrayBuffer}
 
 sealed abstract class VariableState extends Serializable
 
@@ -56,16 +57,103 @@ case object VariableRead extends VariableState
 case object VariableInProcess extends VariableState
 
 /**
- * Core tuple of a state for variables.
+ * Class for maintaining the state of a variable
+ *
+ * Note: stateArg, valueArg, and defaultValueExprArg are pass by name/lazy
+ * values as it is necessary to postpone their evaluation beyond when the
+ * VariableInstance is created. Attempting to evaluate these arguments will
+ * likely trigger an expression to be evaluated, which will query the
+ * VariableMap if the expression references another variable. If this is
+ * occurring within a defineVariable call, the VariableMap hasn't been fully
+ * created and attempting to evaluate such an expression will result in an OOLAG
+ * Circular Definition Exception
  */
-case class VariableInstance(
-  var state: VariableState,
-  var value: DataValuePrimitiveNullable,
-  rd: VariableRuntimeData,
-  defaultValueExpr: Maybe[CompiledExpression[AnyRef]],
+object VariableInstance {
+  def apply(
+    stateArg: => VariableState,
+    valueArg: => DataValuePrimitiveNullable,
+    rd: VariableRuntimeData,
+    defaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
+    priorState: VariableState = VariableUndefined,
+    priorValue: DataValuePrimitiveNullable = DataValue.NoValue) = {
+
+    new VariableInstance(stateArg, valueArg, rd, defaultValueExprArg, priorState, priorValue)
+  }
+}
+
+/**
+ * See documentation for object VariableInstance
+ */
+class VariableInstance private (
+  @TransientParam stateArg: => VariableState,
+  @TransientParam valueArg: => DataValuePrimitiveNullable,
+  val rd: VariableRuntimeData,
+  @TransientParam defaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
   var priorState: VariableState = VariableUndefined,
   var priorValue: DataValuePrimitiveNullable = DataValue.NoValue)
-  extends Serializable {
+  extends Serializable
+  with PreSerialization {
+
+  lazy val defaultValueExpr = defaultValueExprArg
+
+  private var _value: Option[DataValuePrimitiveNullable] = None
+  private var _state: VariableState = null
+
+  /**
+   * Returns the current value of the VariableInstance
+   *
+   * This function allows value to be set while also enabling valueArg to be
+   * lazily evaluated. This allows the VariableInstance to support setting the
+   * value later and allows the construction of the VariableMap containing the
+   * VariableInstance before evaluating the default value expression
+   */
+  def value = {
+    if (_value.isEmpty) {
+      _value = Some(valueArg)
+    }
+
+    _value.get
+  }
+
+  /**
+   * Returns the current state of the VariableInstance
+   *
+   * This function allows state to be set while also enabling stateArg to be
+   * lazily evaluated. This allows the VariableInstance to support setting the
+   * state later and allows the construction of the VariableMap containing the
+   * VariableInstance before evaluating the default value expression
+   */
+  def state = {
+    if (_state == null) {
+      _state = stateArg
+    }
+
+    _state
+  }
+
+  def setState(s: VariableState) = {
+    this.priorState = this.state
+    this._state = s
+  }
+
+  def setValue(v: DataValuePrimitiveNullable) = {
+    this.priorValue = this.value
+    this._value = Some(v)
+  }
+
+  override def preSerialization: Unit = {
+    defaultValueExpr
+    value
+    state
+  }
+
+  override def toString: String = "VariableInstance(%s,%s,%s,%s,%s,%s".format(
+                                                    state,
+                                                    value,
+                                                    rd,
+                                                    defaultValueExpr,
+                                                    priorState,
+                                                    priorValue)
 
   def copy(
     state: VariableState = state,
@@ -73,27 +161,22 @@ case class VariableInstance(
     rd: VariableRuntimeData = rd,
     defaultValueExpr: Maybe[CompiledExpression[AnyRef]] = defaultValueExpr,
     priorState: VariableState = priorState,
-    priorValue: DataValuePrimitiveNullable = priorValue) =
-      new VariableInstance(state, value, rd, defaultValueExpr, priorState, priorValue)
-
-  def setState(s: VariableState) = {
-    this.priorState = this.state
-    this.state = s
+    priorValue: DataValuePrimitiveNullable = priorValue) = {
+      val inst = new VariableInstance(state, value, rd, defaultValueExpr, priorState, priorValue)
+      inst.init()
+      inst
   }
 
-  def setValue(v: DataValuePrimitiveNullable) = {
-    this.priorValue = this.value
-    this.value = v
-  }
+  def init() = preSerialization
 
   def reset() = {
     Assert.invariant(this.state != VariableUndefined)
     (this.state, this.priorState, this.defaultValueExpr.isDefined) match {
-      case (VariableRead, VariableSet, _) => this.state = VariableSet
-      case (VariableRead, _, true) => this.state = VariableDefined
+      case (VariableRead, VariableSet, _) => this.setState(VariableSet)
+      case (VariableRead, _, true) => this.setState(VariableDefined)
       case (VariableSet, _, _) => {
-        this.state = this.priorState
-        this.value = this.priorValue
+        this.setState(this.priorState)
+        this.setValue(this.priorValue)
       }
       case (_, _, _) => Assert.impossible("Should have SDE before reaching this")
     }
@@ -127,6 +210,14 @@ abstract class VariableException(val qname: NamedQName, val context: VariableRun
 class VariableHasNoValue(qname: NamedQName, context: VariableRuntimeData) extends VariableException(qname, context,
   "Variable map (runtime): variable %s has no value. It was not set, and has no default value.".format(qname))
   with RetryableException
+
+/**
+ * This expression can be thrown either at the start of parsing is the
+ * expressions in a defineVariable are circular, or later during parsing if
+ * newVariableInstance contains a circular expression
+ */
+class VariableCircularDefinition(qname: NamedQName, context: VariableRuntimeData) extends VariableException(qname, context,
+  "Variable map (runtime): variable %s is part of a circular definition with other variables".format(qname))
 
 /**
  * Provides one more indirection to the variable map.
@@ -166,17 +257,21 @@ final class VariableBox(initialVMap: VariableMap) {
  *
  * The DPath implementation must be made to implement the
  * no-set-after-default-value-has-been-read behavior. This requires that reading the variables causes a state transition.
+ *
+ * The implementation of the VariabeMap uses ArrayBuffers essentially as a
+ * stack, as they allow for easy serialization unlike the custom MStack classes
+ * we use elsewhere. Scala's mutable Stack is deprecated in 2.12
  */
-class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
+class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]])
   extends Serializable {
 
   def this(topLevelVRDs: Seq[VariableRuntimeData] = Nil) =
     this(Map(topLevelVRDs.map {
       vrd =>
         val variab = vrd.createVariableInstance()
-        val stack = new MStackOf[VariableInstance]
-        stack.push(variab)
-        (vrd.globalQName, stack)
+        val abuf = new ArrayBuffer[VariableInstance]
+        abuf += variab
+        (vrd.globalQName, abuf)
     }: _*))
 
   override def toString(): String = {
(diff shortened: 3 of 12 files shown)


jadams-tresys:
This should be ready for a proper review.  Open to input on how best to handle the negative test cases that don't want to work in TDMLRunner.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -222,7 +300,7 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
   /**
    * For testing mostly.
    */
-  def getVariableBindings(qn: GlobalQName): MStackOf[VariableInstance] = {
+  def getVariableBindings(qn: GlobalQName): Queue[VariableInstance] = {
```
mbeckerle, line 303: We used to use a stack data structure. Why is this Queue and not Stack? It would seem the most natural replacement for MStack would be Stack. 

This would eliminate a bunch of confusion below where we get a "queue" from the vtable using a "qname", i.e., where "Q" gets overloaded about Qualified (as in names) and Queue (as in data structure). 

Can we use Stack?

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
```diff
@@ -59,6 +59,16 @@ class TestVariables {
   @Test def test_varInstance_09(): Unit = { runner.runOneTest("varInstance_09") }
   @Test def test_varInstance_10(): Unit = { runner.runOneTest("varInstance_10") }
   @Test def test_varInstance_11(): Unit = { runner.runOneTest("varInstance_11") }
+  @Test def test_varInstance_12(): Unit = { runner.runOneTest("varInstance_12") }
+
+  @Test def test_defineVariable_nonConstantExpression(): Unit = { runner.runOneTest("defineVariable_nonConstantExpression") }
+  // The following tests don't seem to play nicely with TDMLRunner and the errors are not detected/checked against
+  //@Test def test_circular_defineVariable_err(): Unit = { runner.runOneTest("circular_defineVariable_err") }
```
mbeckerle, line 66: Is there still a reason for these tests to be commented out? If so please create JIRA tickets for these.

mbeckerle (changes requested):
Stack vs.. Queue issue, and need some scaladoc and comments added.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -222,7 +300,7 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
   /**
    * For testing mostly.
```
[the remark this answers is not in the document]
jadams-tresys: I've changed it to "Used only for testing" as that seems to be the only place this was used.  I think this was from the original implementation of variables.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: So mutable Stack is deprecated in Scala 2.12.  It's back in 2.13, but I believe we are still primarily targetting 2.12

mbeckerle: Well, a cheap collection class is ArrayBuffer, if we don't care about it really being stack-like or queue-like. I guess I don't like using a Queue as a bag if mutable set, or just array buffer is sufficient. What is the access pattern needed here? It used to be MStack, so I'm surprised Queue works.

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
jadams-tresys: I meant to leave a comment about those tests in the pull request, but must have forgotten.

test_circular_defineVariable_err produces the error as expected with the given test case, but for some reason TDMLRunner fails to recognize it and the test just fails and prints the stack trace.  I'm guessing that it may be due to the fact that the test fails after schema compilation but before any infoset is generated.  Same issue applies for test_defineVariable_ref_noDefault_err.  I'm open to creating a ticket for it, but wanted to get opinions if I am simply doing something wrong in the test or if this sort of test should be moved to a CLI test instead (which shouldn't have any problem detecting the error).

test_defineVariable_ref_infoset_err runs into issues in the expression path handling code where it assumes that there will always be an infoset.  I will create a ticket for this bug.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: A stack is the ideal structure for this as with newVariableInstance we are pushing and popping VariableInstance's as they go in and out of scope.  That being said, you are right that an ArrayBuffer will work just fine and I'm sure would be more efficient. I'll make the change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -128,6 +184,10 @@ class VariableHasNoValue(qname: NamedQName, context: VariableRuntimeData) extend
   "Variable map (runtime): variable %s has no value. It was not set, and has no default value.".format(qname))
   with RetryableException
 
+class VariableCircularDefinition(qname: NamedQName, context: VariableRuntimeData) extends VariableException(qname, context,
+  "Variable map (runtime): variable %s is part of a circular definition with other variables".format(qname))
+  with RetryableException
```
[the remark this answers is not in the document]
jadams-tresys: I think I had retryable here simply because I copy and pasted from the exception above it. Will add scaladoc

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -232,19 +310,37 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
+
+            // Need to update the variable's value with the result of the
+            // expression
+            variable.setState(VariableRead)
+            variable.setValue(res)
+
+            res
+          } else {
+            variable.setState(VariableRead)
+            variable.value.getNonNullable
```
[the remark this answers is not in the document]
jadams-tresys: So I added this else section thinking that it would get hit during unparsing, but after thinking it over, this will never get called during unparse, so I restructured this to not check to see if the given state is PState and removed this if/else.

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
mbeckerle: I think the bug that is failing and producing a stack trace that's got to be a TDML-runner issue. Do you have a negative test that defines two variables that are circular. That test should be able to expect an SDE and get it. If that's getting masked somehow, it's a TDML runner issue, or perhaps the error is being thrown of wrong type somehow. Wants to be a compile time SDE.

jadams-tresys: Both the circular definition error and the test with an expression referencing a variable without a default value produce runtime schema definition errors, but they are not picked up by TDMLRunner for some reason.  I'll create a bug for the issue and point to the tests that trigger the issue.

mbeckerle (approved):
+1 with JIRA tix for the TDML runner issues where runtime SDEs are being masked.

jadams-tresys:
DAFFODIL-2443 has been created for the TDMLRunner issues
DAFFODIL-2444 has been created for the unhandled exceptions when evaluating paths before the infoset exists.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -232,19 +345,33 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
-      variable.state match {
-        case VariableRead if (variable.value.isDefined) => variable.value.getNonNullable
-        case VariableDefined | VariableSet if (variable.value.isDefined) => {
+    val abuf = vTable.get(varQName)
+    if (abuf.isDefined) {
+      val variable = abuf.get.last
+      (variable.state, variable.value.isDefined, variable.defaultValueExpr.isDefined) match {
+        case (VariableRead, true, _) => variable.value.getNonNullable
+        case (VariableDefined | VariableSet, true, _) => {
           if (maybePstate.isDefined && maybePstate.get.isInstanceOf[PState])
```
stevedlawrence, line 354: While we're renaming variables, can we change maybePstate to just maybeState, since it seems it can be either parse or unparse? I thought this was specifically for parse, and was confused about what happens during unparse, but that doesn't seem to be the case. Or should this always be PState, and maybe the type of maybePState should be Maybe[PState] in which case you only need th isDefined check?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -232,19 +345,33 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
 
           variable.setState(VariableRead)
           variable.value.getNonNullable
         }
-        case _ => throw new VariableHasNoValue(varQName, vrd)
+        case (VariableDefined, false, true) => {
+          Assert.invariant(maybePstate.isDefined)
+          variable.setState(VariableInProcess)
+          val pstate = maybePstate.get
+          val res = DataValue.unsafeFromAnyRef(variable.defaultValueExpr.get.evaluate(pstate))
```
stevedlawrence, line 364: I guess this means maybePState really could be a UState or a PState?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -232,19 +345,33 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
+
+          // Need to update the variable's value with the result of the
+          // expression
+          variable.setState(VariableRead)
+          variable.setValue(res)
+
+          res
```
stevedlawrence, line 371: Is this correct? This is evaluating the defaultValue and setting the state when the variable is read. But the spec says this about newVariableInstance:

> If a default value is specified this initial value of the instance will be created when the instance is created

So it seems this evaluation should not happen during read, but when the instance of the variable is created. And then that default could be overridden with a setVariableCal.

Looks like you do evaluate the default expression below in newVariableInstance, so maybe this just needs to use what that sets?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -232,19 +345,33 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
+        }
+        case (VariableInProcess, _, _) => throw new VariableCircularDefinition(varQName, vrd)
+        case (_, _, _) => throw new VariableHasNoValue(varQName, vrd)
```
stevedlawrence, line 374: Should the test we have add coverage for this, but it's currently failing due to DAFFODIL-2444? This seems like a really important path to cover.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -286,23 +413,28 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
     val varQName = vrd.globalQName
-    val stack = vTable.get(varQName)
-    Assert.invariant(stack.isDefined)
-    stack.get.push(vrd.createVariableInstance)
+    val abuf = vTable.get(varQName)
+    Assert.invariant(abuf.isDefined)
+
+    if (vrd.maybeDefaultValueExpr.isDefined) {
+      val defaultValue = DataValue.unsafeFromAnyRef(vrd.maybeDefaultValueExpr.get.evaluate(state))
+      abuf.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, referringContext))
```
stevedlawrence, line 423: This ``VariableUtils.convert`` seems incorrect to me. I would expect that evaluating the defaultValue expression, or any setVariable expression as well, would know the target type and the expression evaluation would handle the conversion, just like it does with all other expression evaluations. Looks like this isn't something you added, so probably out of scope for this change. But we should create a bug to fix this.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -1884,4 +1976,63 @@
...
+
+  <tdml:parserTestCase name="defineVariable_ref_noDefault_err" root="root"
+    model="defineVariable_ref_noDefault" description="Section 7 - ">
+    <tdml:document/>
+    <tdml:errors>
+      <tdml:error>Runtime Schema Definition Error</tdml:error>
+      <tdml:error>has no value. It was not set</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
```
stevedlawrence, line 2037: An intersting test case that I'm not sure is covered but would be worth having: What if you have two variables where they both have defaults, and one refers to the other, like what you have above:
```xml
    <dfdl:defineVariable name="var1" type="xs:int" defaultValue="5" />
￼    <dfdl:defineVariable name="var2" type="xs:int" defaultValue="{ $var1 + 1 }" />
```
Now what if during parse we have a dfdl:setVariable that sets the value of both of these variables? The spec says this:
> If a defaultValue expression references another variable then the single-assignment nature of variables prevents the referenced variable's value from ever changing, that is, it is considered to be a read of the variable's value, and once read, a variable's value cannot be changed.

So we shoudl be able to setVariable var2, but not setVariable var1. I think having a test that checks for this would be useful.

Another test I think would be useful, if it doesn't already exist, would be to allow var1 and var2 to be externally set, and make sure that those values override the default values. With the change with the forceEvaluations and delay evaluates of things, I'm not 100% confident that works as expected.

daffodil-lib/src/main/scala/org/apache/daffodil/util/MStack.scala
```diff
@@ -182,7 +182,7 @@ object MStackOfAnyRef {
  * things.
  */
 protected abstract class MStack[@specialized T] private[util] (
-  arrayAllocator: (Int) => Array[T], nullValue: T) extends Serializable {
+  arrayAllocator: (Int) => Array[T], nullValue: T) {
```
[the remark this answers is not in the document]
jadams-tresys: It was no longer needed since I ended up switching data structures.  The Serializable here was only added for Variables during the newVariableInstance redesign, so it is no longer necessary.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -435,6 +435,11 @@ class DataProcessor private (
       state.notifyDebugging(true)
     }
     state.dataProc.get.init(ssrd.parser)
+
+    // Force the evaluation of any defineVariable's with non-constant default
+    // value expressions
+    state.variableMap.forceExpressionEvaluations(state)
+
```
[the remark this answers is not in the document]
jadams-tresys: Makes sense, will investigate

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: There is some book keeping that is specific to parsing only (see the call to markVariableRead) that doesn't apply to unparsing, hence the check.  I'll go ahead with the rename though.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Yes, renamed to just state

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: So this case here is really only for the case where we have a defineVariable with a default value expression that references another variable with a default value expression.  So, what happens is when we call the forceExpressoinEvaluations at the start of parsing, it evaluates the top level variable default value expression, which then calls readVariable for the variable that it references, which then needs it's expression evaluated at that time.  I'll add a comment to this case explaining what is happening when this case is hit.

This isn't an issue for newVariableInstance, as we do no need to postpone the evaluation of expressions because the VariableMap alreadys exists by the time we would run into any newVariableInstance expressions.

stevedlawrence: Ah, makes sense. A comment explaining this would definitely be useful.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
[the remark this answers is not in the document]
jadams-tresys: You are correct, the problem was indeed having this expression evaluations outside of the try block.  This resolved the issues mentioned in DAFFODIL-2443.  DAFFODIL-2444 still applies though.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: This is covered by the tests that were mentioned in DAFFODIL-2443 which, thanks to the changes you suggested for moving the forceExpressionEvaluations inside the doParse try block, now work as expected and hit these cases.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
jadams-tresys: This should work, but you are right, definitely should have a test to ensure that it does.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: I've pushed up changes addressing all your other comments and I've created DAFFODIL-2445 to track this issue.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -188,24 +283,42 @@ class VariableMap private(vTable: Map[GlobalQName, MStackOf[VariableInstance]])
...
-      val newStack= new MStackOf[VariableInstance]()
+    val table = Map[GlobalQName, ArrayBuffer[VariableInstance]]()
+    vTable.foreach { case (k: GlobalQName, abuf: ArrayBuffer[VariableInstance]) => {
+      val newBuf = new ArrayBuffer[VariableInstance]()
 
       // toList provides a list in LIFO order, so we need to reverse it to
       // maintain order
-      s.toList.reverse.foreach { case v: VariableInstance => newStack.push(v.copy()) }
-      table(k) = newStack
+      abuf.foreach { case v: VariableInstance => { newBuf += v.copy() }}
```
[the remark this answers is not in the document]
stevedlawrence: Doesn't look like you changed foreach to map, was this just missed or does it not work?

stevedlawrence (approved):

stevedlawrence:
+1 just a couple minor changes for comments that maybe you missed?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -343,8 +343,8 @@ final class PState private (
     changedVariablesStack.top += vrd.globalQName
   }
 
-  def newVariableInstance(vrd: VariableRuntimeData): Unit = {
-    variableMap.newVariableInstance(vrd)
+  def newVariableInstance(vrd: VariableRuntimeData, pstate: PState): Unit = {
+    variableMap.newVariableInstance(vrd, pstate)
```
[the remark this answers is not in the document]
jadams-tresys: Made the change and also updated the setVariable function/calls to follow sut

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
jadams-tresys: I must have missed this one.  I wonder what the logic github has for deciding what things should be hidden

PR #451, opened 2020-11-06
Clean up .gitignore and ratCheck excludes list

stevedlawrence:
- Update the .gitignore file so that the .bsp directory generated by sbt
  for the Build Server Protocol is correctly ignored by ratCheck--the
  forward slash caused problems for the sbt-rat plugin. While at it,
  remove all .gitignore entries that are related to the OS environment
  and IDE. These types of entries should go in the $HOME/config/ignore
  or $GIT_DIR/info/exclude files (man gitignore for more info).
- Removes .classpath and .keep cruft
- Updates the ratExcludes file based on the above changes. Also adds
  missing binary files for tests and windows installation that ratCheck
  doesn't alert about but should be ignored since they are expected to
  be in he repo. Also sorts the ratExcludes list for easier scanning

DAFFODIL-2430


tuxji (approved):
+1

I like the idea of putting all OS/IDE-related .gitignore entries in $XDG_CONFIG_HOME/git/ignore or $GIT_DIR/info/exclude.  Note we will have to rely more on pull request reviews to prevent developers from inadvertently committing OS/IDE-related files.  It also would be a good idea to add a step to the [Code Contributor Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow) asking developers to create either of these files and put their OS/IDE-related entries there.

stevedlawrence:
Good call on updating the workflow. Will do. Maybe it's also worth adding a note in the .gitignore file that it should not include OS/IDE related files and to use the other options instead?

Regarding accidental commits of OS/IDE files, I'd guess that in the majority of cases, if not all, those files won't have an Apache license, which means they will fail the ratCheck. So that should hopefully catch the cases where contributors just do something like ``git add . && git commit``. But still something to be aware of for sure, it's definitely possible for these files to include licenses that rat won't have a problem with. We also need to keep an eye out for .gitignore or ratExcludes changes that include OS/IDE files.

stevedlawrence:
Pushed two changes, one that adds a note to .gitignore to prevent OS/IDE changes, and another so that plugins.sbt is not ignored by the *.sbt ignore. This fixes DAFFODIL-2360, figured it make sense to just combine the change into a single PR since it's so small.

mbeckerle (approved):
+1

PR #454, opened 2020-11-11
Update ratCheck to latest version, and check fail on unexpected binaries

stevedlawrence:
ASF policy is to not allow unexpected binaries in source/releases.
Unfortunately, ratCheck doesn't currently fail if binaries are
committed, so it is possible for them to unintentionally slip into the
code base. To ensure this doesn't happen, this updates the sbt-rat
plugin to the latest version and enables the new ratFailBinaries setting
so it fails if binaries are found in the repo. This now requires that
any binaries that are expected and allowed to be in the repo (e.g.
images, test data) are explicitly excluded in project/Rat.scala

DAFFODIL-2434

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -232,12 +232,11 @@ lazy val ratSettings = Seq(
   ratLicenses := Seq(
     ("BSD2 ", Rat.BSD2_LICENSE_NAME, Rat.LICENSE_TEXT_PASSERA)
   ),
-
   ratLicenseFamilies := Seq(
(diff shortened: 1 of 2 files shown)


stevedlawrence:
Note: This PR will fail until #451 is merged.

olabusayoT (approved):

olabusayoT:
+1

tuxji (approved):
+1

PR #456, opened 2020-11-12
Support all LTS and one non-LTS Java versions

stevedlawrence:
- Drop support for Java 9 since it is no longer supported by upstream.
  Now only support the two LTS versions (Java 8 and 11) and the latest
  non-LTS version (Java 15)
- Javadoc versions newer than Java 12 are a bit more strict about some
  aspects of Javadoc. Update the JAPI/SAPI documentation so that there
  are no errors, and to minimze warnings. The genjavadoc plugin we use
  to convert Scala code to Java API isn't perfect so we can't get rid of
  all warnings. This is caused by missing constructor documentation that
  doesn't convert from scaladoc to javadoc.
- Add the -quiet option to javadoc so it only shows if there are issues.

DAFFODIL-2463, DAFFODIL-2332

### DIFF
--- a/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/infoset/Infoset.scala
+++ b/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/infoset/Infoset.scala
@@ -205,7 +205,7 @@ abstract class InfosetOutputter extends SInfosetOutputter {
 
  
 /**
- * Output the infoset as a scala.xml.Node
+ * [[InfosetOutputter]] to build an infoset represented as a scala.xml.Node
  *
  * @param showFormatInfo add additional properties to each scala.xml.Node for debug purposes
  */
@@ -214,9 +214,17 @@ class ScalaXMLInfosetOutputter(showFormatInfo: Boolean = false)
 
   override val infosetOutputter = new SScalaXMLInfosetOutputter(showFormatInfo)
 
+  /**
+   * Get the scala.xml.Node representing the infoset created during a parse
+   *
+   * This function shuld only be called if ParseResult.isError() returns false
+   */
   def getResult(): scala.xml.Node = infosetOutputter.getResult()
 }
 
+/**
+ * [[InfosetOutputter]] to build an infoset represented as XML written to a java.io.OutputStream
+ */
 class XMLTextInfosetOutputter private (outputter: SXMLTextInfosetOutputter)
   extends InfosetOutputterProxy {
 
@@ -252,6 +260,9 @@ class XMLTextInfosetOutputter private (outputter: SXMLTextInfosetOutputter)
   override val infosetOutputter = outputter
 }
 
+/**
+ * [[InfosetOutputter]] to build an infoset represented as JSON written to a java.io.OutputStream
+ */
 class JsonInfosetOutputter private (outputter: SJsonInfosetOutputter)
   extends InfosetOutputterProxy {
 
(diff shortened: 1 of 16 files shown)


daffodil-udf/src/main/java/org/apache/daffodil/udf/UserDefinedFunctionIdentification.java
```diff
@@ -28,11 +28,21 @@
...
 public @interface UserDefinedFunctionIdentification {
+  /**
+   * Get the name of the user defined function
+   *
+   * @return the name of the user defined function
+   */
   String name();
+
+  /**
+   * Get the namespace URI of the user defined function
```
mbeckerle, line 43: Gaaak. My job is at the redundancy department is where I work.  :-)

I hate having to have 6 lines All of which basically say the same thing, and add no value over just the function signature itself.  To me this is just code pollution. Might as well say:
```
/**
 * Read the f***ing method name.  (RTFM)
 */
```

What happens if you have no javadoc for this at all? The javadoc generated will still show this as a public member right?

mbeckerle (approved):
+1 Though I question whether some javadoc is just code pollution. I will have to get Java 15 now.

daffodil-udf/src/main/java/org/apache/daffodil/udf/UserDefinedFunctionIdentification.java
stevedlawrence: I agree, this feels very unnecessary. Javadoc shows up just fine, the only downside is this will output a warning until you add some javadoc. I was trying to minimize the number of warnings so that it's easier to see when we are missing actually important javadoc that we're actually missing.

mbeckerle: Ok then. Will one-liners work. 
```
/** @returns the value */
```
Might work. 1 line vs. 5.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/infoset/Infoset.scala
```diff
@@ -295,6 +304,9 @@ class JDOMInfosetOutputter()
 
   override val infosetOutputter = new SJDOMInfosetOutputter()
 
+  /**
+   * Get the jdom2 Document representing the infoset created during a parse
```
tuxji, line 308: This comment isn't 100% the same as the comment that was added to the japi file.  The other comment said "Get the jdom Document" instead of "Get the jdom2 Document".  The two sets of APIs do have to be different where it's needed but it's a maintenance burden to have to make sure unwanted differences don't cause problems and this is an example of how easily differences can happen.

tuxji (approved):
+1

I agree with Mike about 1) the redundant verbosity of some javadocs, and 2) the error-prone duplication of code between daffodil-japi and daffodil-sapi.  For 1) let's use 1-liners where possible and 2) let's explore the feasibility of automatically generating daffodil-japi and daffodil-sapi or getting rid of them somehow.

daffodil-udf/src/main/java/org/apache/daffodil/udf/UserDefinedFunctionIdentification.java
stevedlawrence: I agree with removing redundancy, but I don't see much value in trying to save lines or even removing descriptions. It maybe feels verbose, but this is standard practice in Javadocs. And perhaps the real issue is this docs are just not good docs. For example, maybe this should be more along the lines of:

```scala
/**
 * Get the local name of the user defined function, excluding a namespace or prefix
 *
 * @returns the local name
 */
```

Which I think does provide some added value that just "name()" doesn't necessarily imply

stevedlawrence:
I've made some updates that I improve the javadocs consistency between scala and java api's, remove some redundancy, and adds additional detail in some places that I think validates the docs existence.

Agreed that we really need to figure out a solution to get rid of the maintenence burdon of these two similar but not quiet the same APIs. [DAFFODIL-1747](https://issues.apache.org/jira/browse/DAFFODIL-1747) talks about some of the minor differences.

PR #457, opened 2020-11-16
Correctly detect zero length status for "dead" DataOutputStreams

stevedlawrence:
When unparsing postfix separators, we create split regions to determine
if a data region is zero length or not, which is used to help decided
whether or not to output a postfix separator. If these regions are
"finished" without anything ever writing to them, then we know the
region is zero length. But if one of these regions is finished but
subsequently content delivered and it becomes "dead", it is possible for
the zero length status to never be set. This leaves the zero length
status in an "unknown" state which can lead to locked suspensions.

This change fixes this so if a DOS is either finished *or dead* and the
zero length status has never been set, then we can assume that region
had zero length, thus allowing suspensions to resolve.

DAFFODIL-2431

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
@@ -84,13 +84,14 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
       case Zero => // ok
       case NonZero => // ok
       case Unknown => {
-        if (this.isFinished) {
+        if (this.isFinished || this.isDead) {
+          // this DOS was finished or this DOS is dead (i.e. was finished,
+          // became direct and then the following DOS become direct). In either
+          // case, we know that nothing has written to this DOS or the
+          // zlStatus_ would have been set by setNonZeroLength() call, which is
+          // required to be called by everything that writes.
           //
-          // nothing has written to this, or the
-          // zlStatus would have been set by setNonZeroLength() call,
-          // which is required to be called by everything that writes.
-          //
-          // So regardless of whether this stream has been merged/collapsed, etc.
+          // So regardless of whether this stream has been merged/collapsed/dead, etc.
           // we know that no actual writes occurred to this DOS, so it is zero length.
           // And now that it is finsihed, that can never change.
           zlStatus_ = Zero
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-csv.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/test-csv.tdml
@@ -1,0 +1,150 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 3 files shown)


tuxji (approved):
+1

Nice tests!

mbeckerle:
What does "finished" but subsequently content is delivered to it.... what is the scenario where that can happen?

mbeckerle (approved):
+1 Looks fine, and good tests. I just don't understand the transition to state isDead, and how we can be asking for zlStatus of something that isDead. So maybe there's a reason to put a comment in the code about that.

stevedlawrence:
Say we have a direct DOS that is immediately followed by some buffered DOS's, let's call them DOS A and DOS B. The way the postifx suspension stuff works is it depends on some previous DOSs and cannot resolve until their ``zeroLengthStatus`` is determined. Let's assume our postfix suspension only cares about the zeroLengthStatus of the buffered DOS A.

Now let's assume DOS B is finished via a call to ``setFinished``. Because DOS B is not direct, nothing happens except setting its state to Finished.

Now let's assume DOS A is finished via a call to ``setFinished``. Because DOS A is not direct, nothing happens except setting its state to Finished.

Let's also assume at this point nothing asks for the zeroLengthStatus of DOS A. So if nothing wrote to DOS A at this point, we can assume DOS A is a ``Zero`` zeroLengthStatus. But that isn't set yet.

Now let's assume the Direct DOS is finished via a call to ``setFinished``. Because it is direct and it's finished, we then deliver the content of A to the OutputStream in our DirectDOS, and the A DOS becomes the new direct DOS. The state of the old direct DOS is set to "Uninitialized", i.e. ``isDead == true``. Because DOS B is also finished, we can deliver the contents of B to the new direct DOS (which is DOS A). Now DOS A is marked as uninitialized, ``isDead == true``, and B becomes the new direct DOS.

So at this point we have two "dead" DOS's (the original direct DOS and DOS A) and DOS B is now the direct DOS.

But there is still a suspension that needs to know the zeroLengthStatus of this dead DOS A so that it can determine if the postfix separator should be unparsed. When the suspension unsuspends (whenever that may be), it will ask for the zeroLengthStatus of this dead DOS and we must give it an answer. Even though this DOS is dead, it still is needed to provide zero length information.

So the crux of the issue is that a DOS that a suspension depends on could be finished, made direct, and then made dead, all before the suspension that needs information about that DOS unsuspends and asks for the zeroLengthStatus. Fortunately, the logic for finished vs dead is exactly the same. Suspensions don't care.

stevedlawrence:
I've updated the commit message and clarified a comment about what it means to be dead. The commit wasn't wasn't entirely clear.

mbeckerle:
This invariant that "dead" doesn't mean really dead, because a suspension can still care whether it is ZL or not, has implications also for resource management. E.g., suppose we wanted to pool these DOS objects for reuse. The state we call "dead" for them is not equivalent to "reusable" so should be put back into the pool. This suggests the state dead and method isDead should be renamed to outputComplete or some other name that suggests notDeadYet, but that one aspect of usage is over with, but there can still be suspensions. I think the suspension has to be attached to the DOS while it is live. Hence, we could add a hasAssociatedSuspension bit to the DOS object. Then things that are outputComplete but not hasAssociatedSuspension are truly "dead" and could be recycled. Otherwise they would be in state hasAssociatedSuspension until that suspension is resolved. It probably isn't worth it to do all this. I'd just like to somehow capture this very subtle invariant that dead doesn't mean fully-dead.

stevedlawrence:
Agreed, "dead" is perhaps misleading considering there things might still care about the DOS.

I've added [DAFFODIL-2439](https://issues.apache.org/jira/browse/DAFFODIL-2439) to rename variables so that we can make it more clear when a DOS really is no longer needed.

PR #458, opened 2020-11-18
TDML Runner provides namespace. Disallows dup test names.

mbeckerle:
Set out to fix the fact that the TDML Runner needs to pass the
namespace of the root to the TDML processor, as some processors can't
(from their current APIs) search and resolve this themselves.

This fixes a problem with vcard and the IBM cross tester rig.

Picked off DAFFODIL-1299, because I was there in the code, and it was easy to do. 

DAFFODIL-1299, DAFFODIL-2440

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -304,10 +304,9 @@ class DFDLTestSuite private[tdml] (
     _.tcName
   }.filter { case (name, seq) => seq.length > 1 }
   if (duplicateTestCases.nonEmpty) {
-    duplicateTestCases.foreach {
-      case (name, _) =>
-        System.err.println("TDML Runner: More than one test case for name '%s'.".format(name))
-    }
+    val listOfDups = duplicateTestCases.map{ case(name, _) => name}
+    throw new TDMLExceptionImpl(
+      "More than one test for names: " + listOfDups.mkString(","), None)
   }
   val testCaseMap = testCases.map { tc => (tc.tcName -> tc) }.toMap
   val suiteName = (ts \ "@suiteName").text
@@ -380,7 +379,7 @@ class DFDLTestSuite private[tdml] (
     }
     if (isTDMLFileValid) {
       loadingExceptions.foreach { le =>  log(LogLevel.Warning, le.toString) }
-      val testCase = testCases.find(_.tcName == testName)
+      val testCase = testCaseMap.get(testName)
       testCase match {
         case None => throw TDMLException("test " + testName + " was not found.", None)
         case Some(tc) => {
@@ -511,6 +510,7 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
   lazy val tcID = (testCaseXML \ "@ID").text
   lazy val id = tcName + (if (tcID != "") "(" + tcID + ")" else "")
   lazy val rootAttrib = (testCaseXML \ "@root").text
+  lazy val rootNSAttrib = (testCaseXML \ "@rootNS").text
 
   lazy val (infosetRootName, infosetRootNamespaceString) =
     if (this.optExpectedOrInputInfoset.isDefined) {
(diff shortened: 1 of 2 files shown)


mbeckerle:
Works in conjunction with changes to IBM DFDL Cross Tester changes here:
https://github.com/OpenDFDL/ibmDFDLCrossTester/pull/12

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -528,13 +532,39 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
+        }
+        node
+      }
+      val tns = (schemaNode \ "@targetNamespace").text
+      val nsURIString = {
+        if (tns != "") tns
+        else null
+      }
+      nsURIString
+    }
```
stevedlawrence, line 567: It's not clear to me why all this logic is needed for he TDML Runner, when the TDML format originally came from. Shouldn't this not be necesary?

Or could we instead change root to a QName so that the namespace is implied from that? Perhaps that breaks alot of our tests?

stevedlawrence (approved):
+1, but not entirely clear why we need this but IBM doesn't

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -308,6 +308,9 @@ class DFDLTestSuite private[tdml] (
       case (name, _) =>
         System.err.println("TDML Runner: More than one test case for name '%s'.".format(name))
     }
+    val listOfDups = duplicateTestCases.map{ case(name, _) => name}
+    throw new TDMLExceptionImpl(
+      "More than one test for names: " + listOfDups.mkString(","), None)
   }
```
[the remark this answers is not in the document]
tuxji: Yes, and it probably looks better to print all the duplicate tests on one line than to print them one by one unless the test names are very long.

tuxji (approved):
+1

I don't see any conflicts with Runtime2TDMLDFDLProcessor, so looks good to me.

build.sbt
```diff
@@ -104,7 +104,7 @@ lazy val testStdLayout    = Project("daffodil-test-stdLayout", file("test-stdLay
 
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
-  version := "3.0.0",
+  version := "3.1.0-SNAPSHOT",
```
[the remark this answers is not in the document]
mbeckerle: Only had to do this, because I needed the multi-schema test rig to grab this newer Daffodil. It should merge (in the future) without conflict.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: I like the idea of just specifying the root as a QName vs. having added rootNS attribute. 

But ..... probably not worth it.

Turns out this only matters for one situation - TDML negative parser test cases. All other kinds of test cases have an infoset, and the root element's name and namespace can be taken from that (and do not have to be specified at all otherwise).  Any negative parser test case using the ibm cross tester would fail with being unable to find the root element for lack of a namespace. 

So this will only matter such time as we try to run TDML test suites against IBM DFDL including the negative parser tests. 

Turns out there is one such negative parser test case in the vcard DFDL schema. That's what I wanted to fix immediately so we could have clean regression of the supposed portable DFDL schemas on both Daffodil and IBM DFDL working. 

Ultimately we'd like to run daffodil-test against the ibm cross tester (ticket DAFFODIL-2028).

 In daffodil-test there are many many such negative parser tests. So having a heuristic figure out the namespace in most of those will reduce the need to add these rootNS attributes to many of these negative parser test cases.

stevedlawrence: Do we know how IBM's TDML runner handles this negative case and how it finds the root element? They can do this without the rootNS attribute. I know our TDML has diverged quite a bit from the original TDML format, but it would be nice it we could minimize the differences and keep them somewhat compatible.

I'm not a against the hueristic (I did give this a +1), but it feels unsatisfying to me to rely on this heursitic when we could just require that tests are unambigous and error if there are any ambiguities. Unambiguous tests are good thing.

mbeckerle: Their internal TDML implementation does not depend on the same API we have to depend on. 
They can search for elements. We can't.

PR #459, opened 2020-11-20
Ensure src/main directories exist to make Sonarcloud happy

stevedlawrence:
Sonarcloud requires that the src/main directory exists for its scans,
even if there are no sources. Commit 064d7b8655 removed all the .keep
files which caused the src/main directory in daffodil-test and
daffodil-test-ibm1 to be deleted since those modules have no actual
sources. But this caused Sonarcloud failures.

To fix this, add back the .keep files for only these two directories so
that sonarcloud scan will work. It's also possible some IDE's might be
unhappy without src/main directories, so this isn't necessarily
sonarcloud specific.

DAFFODIL-2430

### DIFF
--- a/daffodil-test-ibm1/src/main/.keep
+++ b/daffodil-test-ibm1/src/main/.keep
@@ -1,0 +1,17 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


stevedlawrence:
Thanks to @olabusayoT for testing that this is all that's needed to get sonarcloud fixed.

olabusayoT:
+1

olabusayoT (approved):

tuxji (approved):
+1

PR #460, opened 2020-11-20
Add validation support for tunables

stevedlawrence:
- Adds xs:restrictions to tunables for which there is only a valid range
  within the range of the primitive type. Modifies the tunable generator
  to convert these restrictions to code checks and throw an
  IllegalArgumentException if outside the valid range.
- Also sorts the tunable by name so they are organized.

DAFFODIL-2432

### DIFF
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
@@ -79,6 +79,25 @@
     </xs:simpleContent>
   </xs:complexType>
 
+  <!--
+    This file is used by TunableGenerator.scala to generate code that parses and
+    validates tunable values. The following schema elements define these tunables
+    to be generated. Although this schema is verbose and repetitive, it makes it
+    relatively straightforward to parse and ensure we generate correct code. As
+    such, one must follow a strict pattern to ensure correct code is generated.
+
+    Each element representing a tunable must have an schema primitive as its type,
+    as determined by the tunables allowed value space. Each tunable must also have
+    a default value that is valid for the type. If an element needs additional
+    limitations beyond the value space of the type, then an xs:simpleType with
+    xs:restriction must be used. The base attribute of the restriction must be a
+    schema primitive type. Only the following restrictions are currently supported:
+
+      - minInclusive
+      - maxInclusive
+      - minExclusive
+      - maxExclusive
+  -->
   <xs:element name="tunables">
     <xs:complexType>
       <xs:all>
@@ -110,46 +129,17 @@
             </xs:documentation>
           </xs:annotation>
         </xs:element>
-        <xs:element name="blobChunkSizeInBytes" type="xs:int" default="4096" minOccurs="0">
+        <xs:element name="blobChunkSizeInBytes" default="4096" minOccurs="0">
           <xs:annotation>
             <xs:documentation>
               When reading/writing blob data, the maximum number of bytes to read/write at a time.
             </xs:documentation>
           </xs:annotation>
-        </xs:element>
-        <xs:element name="outputStreamChunkSizeInBytes" type="xs:int" default="65536" minOccurs="0">
-          <xs:annotation>
-            <xs:documentation>
-              When writing file data to the output stream during unparse, this
-              is the maximum number of bytes to write at a time.
-            </xs:documentation>
-          </xs:annotation>
-        </xs:element>
-        <xs:element name="maxByteArrayOutputStreamBufferSizeInBytes" type="xs:long" default="2097152000" minOccurs="0">
-          <xs:annotation>
-            <xs:documentation>
-              When unparsing, this is the maximum size of the buffer that the
-              ByteArrayOutputStream can grow to before switching to a file based
-              output stream.
-            </xs:documentation>
-          </xs:annotation>
-        </xs:element>
-        <xs:element name="tempFilePath" type="xs:string" default="" minOccurs="0">
-          <xs:annotation>
-            <xs:documentation>
-              When unparsing, use this path to store temporary files that may be genrated.
-              The default value (empty string) will result in the use of the java.io.tmpdir
-              property being used as the path.
-            </xs:documentation>
-          </xs:annotation>
-        </xs:element>
-        <xs:element name="defaultInitialRegexMatchLimitInChars" type="xs:int" default="32" minOccurs="0">
-          <xs:annotation>
-            <xs:documentation>
-              Deprecated. This tunable no longer has any affect and is only kept for
-              backwards compatability.
-            </xs:documentation>
-          </xs:annotation>
+          <xs:simpleType>
+            <xs:restriction base="xs:int">
+              <xs:minInclusive value="1" />
+            </xs:restriction>
+          </xs:simpleType>
         </xs:element>
         <xs:element name="defaultEmptyElementParsePolicy" type="daf:TunableEmptyElementParsePolicy" default="treatAsEmpty" minOccurs="0">
           <xs:annotation>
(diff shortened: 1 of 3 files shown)


daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -401,6 +471,11 @@
               frequency of context switching, but increase the memory footprint.
             </xs:documentation>
           </xs:annotation>
+          <xs:simpleType>
```
mbeckerle, line 474: A defined simple type daf:OneOrGreater or some such would be more compact here as these would all remain one-liners with a type="daf:oneOrGreater" attribute.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -197,16 +224,53 @@ class PrimitiveTunable(name: String, schemaType: String, schemaDefault: String)
...
 
+  private def restrictionCheck(rCheck: String, rValue: String): Option[String] = {
+    if (rValue != "") {
+      Some(s"""  if (!(v ${rCheck} ${rValue})) throwInvalidTunableValue(tunable, value)""")
+    } else {
+      None
+    }
+  }
+
+  private val minInclusive = node \\ "minInclusive" \@ "value"
```
mbeckerle, line 235: So it appears this is depending on the XML schema of the tunables being structured not as just an XML schema, but as one that always uses immediate simple types definitions having exactly these attributes. 

If so, you need comments in the tunables.xsd file because looking at it, first thing I'd do is factor out all the redundant type definitions which would break this scraper/generator code.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -110,55 +110,39 @@
...
           </xs:annotation>
         </xs:element>
-        <xs:element name="blobChunkSizeInBytes" type="xs:int" default="4096" minOccurs="0">
+        <xs:element name="blobChunkSizeInBytes" default="4096" minOccurs="0">
           <xs:annotation>
             <xs:documentation>
               When reading/writing blob data, the maximum number of bytes to read/write at a time.
             </xs:documentation>
           </xs:annotation>
+          <xs:simpleType>
```
mbeckerle, line 119: Your scraper generator depends on this being an immediate simpletype with a restriction with min/max/exclusive/inclusive in it manifestly. 

You need to document that here, because changing this to say type="daf:oneOrGreater" with that defined elsewhere would break this I think.

mbeckerle (commented):
+1 A few suggestions for comments needed to enhance maintainability is all.

tuxji (approved):
+1

Hope the TunableGenerator doesn't become more fragile, but any problems should show up pretty easily.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
stevedlawrence: My only concern here, and why I went this route, is it makes the generator quite a bit more complicated. It doesn't currently have any logic to find other types or support restrictions not defined directly as a child of the tunable. It's still pretty dumb. I'd prefer that we just require this schema be verbose in favor a being easily parsable. There's plenty of ways one might want to refactor this file (e.g. imports/includes, unions), that I don't really want to have to support. I'd rather add comments in the schema file that specifiy that the generator expects a certain form. Thoughts?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -200,20 +194,44 @@
...
+              When unparsing, this is the maximum size of the buffer that the
+              ByteArrayOutputStream can grow to before switching to a file based
+              output stream.
+            </xs:documentation>
+          </xs:annotation>
+          <xs:simpleType>
+            <xs:restriction base="xs:int">
+              <xs:minInclusive value="0" />
+            </xs:restriction>
+          </xs:simpleType>
```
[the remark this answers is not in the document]
stevedlawrence: I think this field actually must have a max size of IntMax, so using a long was incorrect.

This tunable sets the max size for an unparse buffer before we switch the buffer from memory based via a ByteArrayOutputStream to file based via a FileOutputStream. When the buffer is backed by a ByteArrayOutputStream, the max length must be IntMax since byte arrays can't be larger than an int. If we allowed this value to be a long, we could overfill the ByteArrayOutputStream instead of switching to a FileOutputStream and I think an exception will be thrown.

I think zero is also a valid value for this tunable. When this value is zero, it just means that we'll always use a FileOutputStream for buffers and never use an in-memory byte array. Perhaps this could be useful on a system with extremely limited memory?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
tuxji: Adding a comment that the tunable generator needs dafext.xsd to be easily parseable (with some explanation what the certain form needs to be) is fine with me.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -265,28 +302,48 @@
...
             <xs:documentation>
               The smallest allowed value of the dfdl:binaryDecimalVirtualPoint property.
             </xs:documentation>
           </xs:annotation>
+          <xs:simpleType>
+            <xs:restriction base="xs:int">
+              <xs:maxInclusive value="-1" />
+            </xs:restriction>
+          </xs:simpleType>
         </xs:element>
```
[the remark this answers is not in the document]
stevedlawrence: The min/maxBinaryDecimalVirtualPoint tunables describe the smallest and largest allowed values of the dfdl:binaryDecimalVirtualPoint property. The value of this property is essentially used by doing:
```scala
val bigDec = new JBigDecimal(bigInt, binaryDecimalVirtualPoint)
```
I believe we ran into a bug where very large and very small values of binaryDecimalVirtualPoint didn't work as expected so we put in placed an arbitrary limit of -200 to 200. And I think the tunable was added as a way to easily change this arbitrary limit if needed.

My thinking with min/maxBinaryDecimalVirtual point was that minBinaryDecimalVirtualPoint should always be negative so has a max value of -1. And maxBinaryDecimalVirtualPoint should be positive, so has a min value of 1.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
stevedlawrence: I've added a new comment that describes the required format.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -134,7 +138,9 @@ class TunableGenerator(schemaRootConfig: scala.xml.Node, schemaRootExt: scala.xm
     val tunables =
       tunableNodes.map { tunableNode =>
         val schemaName = tunableNode \@ "name"
-        val schemaType = tunableNode \@ "type"
+        val schemaType =
+          if (tunableNode \@ "type" != "") tunableNode \@ "type"
+          else tunableNode \\ "restriction" \@ "base"
```
mbeckerle, line 143: This code is not going to give any sort of useful diagnostic if the XSD just happens to have constructs that aren't recognized in it. 

Add a check here if there is no type attribute, then there must be a restriction, or throw a useful diagnostic stating that these are the only allowed options.

mbeckerle (approved):
+1 with diagnostic if XSD has neither type attribute nor immediate simple type with restriction.

stevedlawrence:
I've added a new commit to add additional validation so that we get more helpful errors if the schema doesn't follow our strict requirements.

This found one tunable that didn't have a default because the default was hardcoded. I changed it so the default in the XSD is the actual java code so that it the tunable has a default, but I'm not sure I like that. This means all properties have a default and we atually use it, but it feels wrong having java code in the XSD. Thoughts?

```diff
- new FileTunable("tempFilePath", """System.getProperty("java.io.tmpdir")""")
+ new FileTunable("tempFilePath", schemaDefault)
```

```diff
- <xs:element name="tempFilePath" type="xs:string" default="" minOccurs="0">
+ <xs:element name="tempFilePath" type="xs:string" default="System.getProperty(&quot;java.io.tmpdir&quot;)" minOccurs="0">
 ```

mbeckerle:
So this thunk of Java code (or is it scala code) gets somehow executed?
That has to be an exceptional case in the generator anyway, so why not just put the code there, and document this in the XSD by giving this string a value like "This string is ignored. Value is taken from java.io.tmpdir property" . 

We probably don't want that to be generally done for more tunables. So I'd rather take this out of the XSD, document that this is hard-coded.

stevedlawrence:
That's a much better solution. Fixup commit added.

PR #461, opened 2020-11-23
Merge asf/master into asf/runtime2-2202

tuxji:
Bring the runtime2-2202 branch up to date with the main branch.  Also,
fix .github/workflows/sonarcloud.yml so that sbt will be able to
compile the C source files too.  Make main.yml and sonarcloud.yml
setup their jobs the same way for consistency, with one exception: use
$SBTNOCOV instead of $SBT since I don't think sonarcloud needs to run
sbt with coverage enabled, does it?

Rest of new changes are just to fix merge conflicts in these 6 files:
```
.github/workflows/main.yml
daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
project/Dependencies.scala
project/Rat.scala
```


tuxji:
Everything looked fine until I created the pull request.  Now GitHub says this repository will not allow a merge commit to be created and this branch cannot be rebased due to conflicts.

mbeckerle:
We don't allow merge commits onto master. 

My theory: I suppose that is a repository-wide setting which is why it is preventing you from doing this on runtime2-2202 branch.

Maybe git checkout master; git checkout -b master-2020-11-12; git checkout runtime2-2202; git rebase master-2020-11-12 ;

mbeckerle (approved):
+1

tuxji:
I can't merge this PR so I'll close it and do a force-push instead after waiting for +1 from two devs on this PR or in the dev list where I've sent an email as well.

mbeckerle:
I would like @stevedlawrence  to weigh in on the proper git workflow for this long-lived branch that we want to keep current with master while remaining on a separate branch.  

In principle, this seems simple. Periodically rebase this runtime2-2202 branch on top of master, but I can see how that requires a force-push each time and a force-push on a fork dev branch is fine (and part of our standard workflow), but on a branch like this runtime2-2202 that others will be fetching and using, it's pretty likely to cause trouble. 

 The alternative maybe is to rebase this runime2-2202 branch on top of master, then git checkout runtime2-2202-N (where N = 1, 2, ....) and push that new branch each time. That clutters the repo with new branches though.

stevedlawrence:
For development branches like this, the workflow I've ususally seen and preferred is to just periodically rebase onto master when it seems appropriate. Eventually this branch will need to be rebased and committed, so doing it that way makes things a bit easier once that time comes. I think merge commits created by merging master into this branch could make that process more difficult.

The downside as mentioned, is that it can cause some headaches if multiple people are working on this dev branch and it's constantly being rebased. I would suggest the solution here is to only rebase periodicaly (e.g once every month or two) and ensure there is good public communication when a rebase is needed/planned. If it needs to be rebased due to some new feature in master for example, send out an email to dev@ and give the usual 72 hours to make sure there are no outstanding commits for this dev branch. At that point, perform the rebase and send out an email saying it's been rebased and for people to update their branch.

Note that we don't need a new PR for this dev branch rebase, we just trust that whoever does the rebase verified there were no issues before doing the force push. Also, any open PRs for this branch should still remain open and github should do the right thing. Though, ideally those PRs would be merged before the rebase to avoid any potential issues.

tuxji:
Thanks, Steve.  Since I've already sent an email to dev@ much longer than 72 hours before, I'll do the force-push now from my rebased branch.  I'll close this PR since we won't need it, but my force-push's contents will be identical to this PR (although my changes so far will be squashed into a single commit).

PR #463, opened 2020-11-25
Embedded Schematron

jw3:
Enable validation using schematron embedded in dfdl xsd.

DAFFODIL-2446


jw3:
Capturing the initial poc here for review as it progresses.  

The TestEmbeddedSchematron demonstrates the capability, but still a bit of work to do.

Based on #452 

### draft todo
- [ ] detect sch or xsd
- [ ] doc in readme
- [ ] tests with complex schematron
- [ ] rebase after #452

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
```diff
@@ -0,0 +1,24 @@
...
+           elementFormDefault="qualified">
+
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <xs:annotation>
+        <xs:appinfo source="http://www.ogf.org/dfdl/">
+            <dfdl:format ref="GeneralFormat" />
+        </xs:appinfo>
+    </xs:annotation>
+
+    <xs:element name="uuid" type="xs:string" dfdl:lengthKind="delimited" dfdl:terminator="%NL;">
```
mbeckerle, line 13: This test is not able to distinguish "." from "/", which is a key thing for tests to cover. Nest this uuid element inside another root element and then this would have more coverage, or add a second such test which would test exactly the issue of whether "." reflects the right infoset location in the schematron rules.

mbeckerle:
A good thing to do with this (and the prior) PR that provide schematron, is to fix the JPEG schema that is published on github to use this new feature.  Version 1 would incorporate that JPEG schema's existing ".sch" schematron rules file. Version 2 would migrate those rules up into the DFDL schema itself, where IMHO, they belong.  

Without the schematron validation, that DFDL JPEG schema is known to be very poor at rejecting bad data. It's the specific example where we learned it is very important to make DFDL schemas not only parse correct data, but reject invalid and/or malformed data. Without the schematron rules, this JPEG schema accepts all sorts of malformed stuff. This more powerful validator in this case enforces, not only validity, but well-formedness even.

daffodil-schematron/src/main/resources/iso-schematron-xslt2/ExtractSchFromXSD-2.xsl
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+        based on an original transform by Eddie Robertsson
+        2001/04/21      fn: added support for included schemas
+        2001/06/27      er: changed XMl Schema prefix from xsd: to xs: and changed to the Rec namespace
+        2010/04/14    	rj: Update for ISO Schematron using xslt2
+-->
+
```
stevedlawrence, line 8: Remember to add this new file to the three LICENSE files in LICENSE, daffodil-cli/LICENSE, and in daffodil-schematron/src/main/resources/....

jw3:
### reference

- https://reference.niem.gov/niem/specification/sch-in-xsd/1.0alpha4/niem-schematron-in-xsd-1.0alpha4-2016-03-16.html
- https://journal.code4lib.org/articles/13546
- https://www.mulberrytech.com/papers/schematron-Philly.pdf
- 

(some of these might make the readme, might not, here for posterity)

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
jw3: Updated.  Is that change enough to cover that the rule is properly scoped to the context?

Working on improving my daffodil schema understanding at the same time as adding more complex tests and working with the JPEG schema...  :turtle:

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
```diff
@@ -30,13 +30,19 @@
                     <sch:assert test="substring(.,9,1)='-' and substring(.,14,1)='-' and substring(.,19,1)='-' and substring(.,24,1)='-'">malformed uuid</sch:assert>
                 </sch:rule>
             </sch:pattern>
+            <sch:pattern name="test-short-uuid">
+                <sch:rule context="short-uuid">
+                    <sch:assert test="string-length(.)=8">short uuid must be 8 chars</sch:assert>
```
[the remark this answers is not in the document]
jw3: > I was expecting to see this rule as an annotation on the element declaration of short-uuid below.

The first pass had it embedded and it looked like it was going to get noisy.

I can see advantages of both..  will add an example back in for demonstration.  

> W.r.t this test, I was hoping for a test using a sch rule that would use some relative paths like ../../foo/bar to reach around in the infoset and constrain disparate things together.

I see.  Will add examples.

jw3:
@rthomas320 

> If SVRL reports were generated would it be possible to commit those to git also?

I can post the SVRL generated by a unit test or two here in the PR.

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
```diff
@@ -24,25 +24,26 @@
...
+            <sch:pattern name="validate-uuid">
                 <sch:rule context="uuid">
-                    <sch:assert test="string-length(.)=36">uuid must be 36 chars</sch:assert>
-                    <sch:assert test="substring(.,9,1)='-' and substring(.,14,1)='-' and substring(.,19,1)='-' and substring(.,24,1)='-'">malformed uuid</sch:assert>
-                </sch:rule>
-            </sch:pattern>
-            <sch:pattern name="test-short-uuid">
-                <sch:rule context="short-uuid">
-                    <sch:assert test="string-length(.)=8">short uuid must be 8 chars</sch:assert>
+                    <sch:assert test="string-length(time_low)=8"><sch:name/> wrong length</sch:assert>
```
mbeckerle, line 29: So you don't need package names on these functions? Why? 

I would expect to see "fn:string-length(.) eq 8"  

Is schematron's expression language actually not XPath based?

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/TestBasicValidation.scala
```diff
@@ -68,7 +68,7 @@ class TestBasicValidation {
     val factory = new TransformerFactoryImpl()
     factory.setURIResolver(resolver)
     val sch = getClass.getClassLoader.getResourceAsStream("sch/schematron-1.sch")
-    val rules = Transforms.fromSch(sch, factory)
+    val rules = Transforms.from(sch, Sch, factory)
```
mbeckerle, line 71: Might be a separate ticket, but it seems to me we want to be able to write TDML tests with validation settings that will cause them to invoke schematron as part of the runtime. E.g., define a tdml:testSuite with defaultValidation="schematron" or a specific tdml:parserTestCase with validation="schematron"

daffodil-schematron/src/test/resources/xsd/extends-1.dfdl.xsd
```diff
@@ -0,0 +1,59 @@
...
+
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <xs:annotation>
+        <xs:appinfo source="http://www.ogf.org/dfdl/">
+            <dfdl:format ref="GeneralFormat"/>
+        </xs:appinfo>
+        <xs:appinfo>
+            <sch:pattern>
+                <sch:rule abstract="true" id="blank">
+                    <sch:assert test="string-length(.) > 0"><sch:name/> is blank.</sch:assert>
```
mbeckerle, line 30: I expected the test expression to have prefixes like fn:string-length(.)

Is the schematron expression language not XPath based? I am also surprised to see ">" as an operation.

mbeckerle (approved):
Test aren't passing, otherwise this looks to be coming along nicely. Just a couple questions.

daffodil-schematron/src/test/resources/xsd/extends-1.dfdl.xsd
rthomas320: XPath allows the greater than '>'  symbol, but not the less than '<' symbol

mbeckerle: Thanks. What about the namespaces for the function names? 

The reason I focus on this is that DFDL expressions are XPath 2.0 derived, though not identically those, but DFDL expressions the function names are in a namespace, which usually requires a namespace prefix like "fn:". 

This inconsistency is manageable, but unfortunate if expressions like these that could, in principle be either DFDL expressions or schematron expressions, can't be cut/pasted between schematron rules and say, DFDL asserts, because the expression languages are that different. 

This really only matters for very simple expressions. DFDL expressions basically can't do anything that can't execute in constant time. If it involves searching, counting, or really any sort of iteration over the infoset tree, then DFDL expressions don't allow it for the most part, and Daffodil doesn't implement it. Those are the kinds of things that you need schematron for.  So DFDL expressions can already check lengths, but a query like "all FOO elements must be beneath a BAR element" or "there must be 3 or more FOO children of each QUUX element" can't be expressed.

jw3:
Tests fixed, going to work through my checklist above.  BMP, GIF, and JPEG appear to be working... I need a generator for each of them to created mutated versions of each... :christmas_tree: :gift: 

For now though things are close and I can reshape/add examples based on the feedback above.

mbeckerle:
If you are looking for negative tests, I recall the DFDL for JPEG without the schematron enforcement will literally accept /etc/termcap as a JPEG. Basically, any piece of line-noise bigger than some minimum is classified as a JPEG blob segment, and only the schematron rules enforce the proper positioning of these blob segments in a proper jpeg segment nest. I.e., the DFDL part does nothing to enforce the "grammar" rules of JPEG. So at least this one class of complete-line-noise negative examples for JPEG are easy.

daffodil-schematron/src/test/resources/xsd/extends-1.dfdl.xsd
jw3: All the references I see say there is a full set of operators in both xpath 1 and 2.  Here is one of several without looking too hard https://www.w3schools.com/xml/xpath_operators.asp

Curious as to what I am missing in this discussion about limited xpath operators?

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
jw3: Ive saw it done both ways.  Will try to pin down why with a reference.

daffodil-schematron/src/test/resources/xsd/extends-1.dfdl.xsd
mbeckerle: XPath 1.0 seems to not require a correct namespace on function names. XPath 2.0 does. 

per: https://www.w3.org/2006/xpath-functions/

DFDL's expression language is similar to XPath 2.0. E.g., you cannot just type string-length(../some/path) in DFDL nor XPath 2.0 and expect it to work. You must bind a prefix (usually "fn" to namespace http://www.w3.org/2005/xpath-functions/ and use the prefix when calling the functions.

If the current schematron is limited to XPath 1.0, then so be it for now. But there's a chance it is using XPath 2.0 but turning on the XPath 1.0 compatibility mode. In which case, if we could *not* do that, then the expressions would be more harmonious with the DFDL expression language.

jw3: Setting `queryBinding="xslt2"` on the root element will enable xslt+xpath 2.0 with our current version of schematron.

jw3: I cant yet generate a failing test even after enabling xslt 2.0.

> E.g., you cannot just type string-length(../some/path) in DFDL nor XPath 2.0 and expect it to work. You must bind a prefix (usually "fn" to namespace http://www.w3.org/2005/xpath-functions/ and use the prefix when calling the functions.

In context of xpath, are you saying "must"  because there is enforcement or because its good practice?

Perhaps a counter example example is the GIF schema that uses unprefixed functions and enables xslt 2.0.

With all that said, I think the tools are all there, the included version of schematron supports either way.

jw3:
Squashed and rebased

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
jw3: > So you don't need package names on these functions? Why?

The default namespace is searched if the function is unprefixed, at least that is what I have gleaned from a couple random sources.

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/TestBasicValidation.scala
jw3: added DAFFODIL-2448

jw3:
@rthomas320 

> If SVRL reports were generated would it be possible to commit those to git also?

A few examples of svrl output

```
<svrl:schematron-output title="" schemaVersion="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:schold="http://www.ascc.net/xml/schematron" xmlns:saxon="http://saxon.sf.net/" xmlns:iso="http://purl.oclc.org/dsdl/schematron" xmlns:svrl="http://purl.oclc.org/dsdl/svrl">
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="time_low"/>
   <svrl:failed-assert location="/uuid[1]/time_low[1]" test="string-length(.)=8">
      <svrl:text>time_low wrong length</svrl:text>
   </svrl:failed-assert>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="time_mid"/>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="time_hi_and_version"/>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="clock_seq_hi_and_res"/>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="node"/>
</svrl:schematron-output>
```

```
<svrl:schematron-output title="" schemaVersion="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:schold="http://www.ascc.net/xml/schematron" xmlns:saxon="http://saxon.sf.net/" xmlns:iso="http://purl.oclc.org/dsdl/schematron" xmlns:svrl="http://purl.oclc.org/dsdl/svrl">
   <svrl:active-pattern document=""/>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="first"/>
   <svrl:failed-assert location="/name[1]/first[1]" test="string-length(.) &gt; 0">
      <svrl:text>first is blank.</svrl:text>
   </svrl:failed-assert>
   <svrl:fired-rule context="last"/>
   <svrl:failed-assert location="/name[1]/last[1]" test="string-length(.) &gt; 0">
      <svrl:text>last is blank.</svrl:text>
   </svrl:failed-assert>
</svrl:schematron-output>
```

```
<svrl:schematron-output title="" schemaVersion="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:schold="http://www.ascc.net/xml/schematron" xmlns:saxon="http://saxon.sf.net/" xmlns:iso="http://purl.oclc.org/dsdl/schematron" xmlns:svrl="http://purl.oclc.org/dsdl/svrl">
   <svrl:active-pattern document=""/>
   <svrl:active-pattern document=""/>
   <svrl:fired-rule context="first"/>
   <svrl:fired-rule context="last"/>
</svrl:schematron-output>
```

daffodil-schematron/src/test/resources/xsd/embedded-1.dfdl.xsd
jw3: from [iso_schematron_skeleton_for_saxon.xsl#L66-L70](https://github.com/Schematron/schematron/blob/master/trunk/schematron/code/iso_schematron_skeleton_for_saxon.xsl#L66-L70
)
```
   Another tip is that Schematron provides an
   element <iso:ns> for declaring the namespaces and prefixes used in Xpaths in 
   attribute values; it does not extend the XML Namespaces mechanism: if a name
   in an XPath has a prefix, there must be an <iso:ns> element for that prefix; if
   a name in an XPath does not have a prefix, it is always in no namespace.
```

jw3:
Any remaining concerns with this?  Work is complete unless there are changes requested.

daffodil-schematron/src/main/scala/org/apache/daffodil/validation/schematron/ClassPathUriResolver.scala
```diff
@@ -31,9 +31,15 @@ final class ClassPathUriResolver(rulesDir: String, fallback: Option[URIResolver]
     Option(getClass.getClassLoader.getResourceAsStream(path)) match {
       case Some(is) => new StreamSource(is)
       case None =>
-        fallback.map(_.resolve(href, base)).getOrElse(
-          throw ValidatorInitializationException(s"schematron resource not found at $path")
-        )
+        // fallback #1;; try the raw classpath without the prefix
+        Option(getClass.getClassLoader.getResourceAsStream(href)) match {
```
stevedlawrence, line 35: Just curious, what's the case were we need to resolve without the template dir? Is this because a DFDL schema might xs:include/xs:import other schemas that might have embedded schematron annotations?

daffodil-schematron/src/test/resources/xsd/always-fails-1.dfdl.xsd
```diff
@@ -0,0 +1,41 @@
...
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <xs:annotation>
+        <xs:appinfo source="http://www.ogf.org/dfdl/">
+            <dfdl:format ref="GeneralFormat"/>
+        </xs:appinfo>
+    </xs:annotation>
+
+    <xs:element name="always-fails" type="xs:string" dfdl:lengthKind="delimited" dfdl:terminator="">
+        <xs:annotation>
+            <xs:appinfo>
```
stevedlawrence, line 31: Interesting, so embdded schematron rules don't require a special appinfo like DFDL does? I guess the XSLT that extracts these just looks for any any elements that have the schematron namespace?

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/EmbeddedTesting.scala
```diff
@@ -0,0 +1,96 @@
...
+  case class Validation(dp: DataProcessor) {
+    def parse(str: String, verbose: PrintInfosetMode = Quiet): PR = withBytes(str.getBytes, verbose)
+
+    def withBytes(bytes: Array[Byte], verbose: PrintInfosetMode = Quiet): PR = {
+      val bos = new ByteArrayOutputStream()
+      val r1 = dp.parse(
+        new InputSourceDataInputStream(new ByteArrayInputStream(bytes)),
+        new XMLTextInfosetOutputter(bos, true))
+
+//      println(bos.toString)
```
stevedlawrence, line 62: Remove println

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/EmbeddedTesting.scala
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.Assert.assertTrue
+
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.File
+
+trait EmbeddedTesting {
+  def withJpgSchema(f: Validation => Unit): Unit = withSchema("xsd/jpeg.dfdl.xsd")(f)
+  def withBmpSchema(f: Validation => Unit): Unit = withSchema("xsd/bmp.dfdl.xsd")(f)
+  def withGifSchema(f: Validation => Unit): Unit = withSchema("xsd/gif.dfdl.xsd")(f)
```
stevedlawrence, line 38: We really shouldn't rely on external schemas for testing. I'd prefer we just remove this and ensure our included test files provide sufficient test coverage. Though I assume this means embedded schematron does work with work with those schemas, which is good to hear.

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/TestEmbeddedSchematron.scala
```diff
@@ -0,0 +1,73 @@
...
+    assertFalse(f.parse("2;1").validated)
+    assertFalse(f.parse("0;0").validated)
+  }
+
+  @Test def testWithNs1(): Unit = withSchema("xsd/with-ns-1.dfdl.xsd") { f =>
+    assertTrue(f.parse("0;1", Always).validated)
+    assertFalse(f.parse("2;1").validated)
+    assertFalse(f.parse("0;0").validated)
+  }
+}
```
stevedlawrence, line 73: I would also like to see testing of the CLI for this feature. First so that it ensures the CLI works and can find things the same way the API can, but also because it acts as a good example of expected usage for the CLI. Could you add some tests in daffodil-cli to test this functionality?

stevedlawrence (approved):

stevedlawrence:
+1 Looks nice! Just a couple minor comments about the licenses and maybe add some CLI testing?

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/EmbeddedTesting.scala
jw3: Yep, good catch.  Those schemas got copied into here to help a while and were deleted, missed this bit of cleanup.

daffodil-schematron/src/test/resources/xsd/always-fails-1.dfdl.xsd
jw3: Importing the ns is the only requirement I have saw.  In this case schematron is imported from the top level schema tag, there are a couple tests that import from an appinfo.  Otherwise nothing noteworthy.

daffodil-cli/bin.LICENSE
```diff
@@ -876,3 +876,28 @@ subcomponents is subject to the terms and conditions of the following licenses.
   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
   SOFTWARE.
+
+  This product bundles content from the Schematron converters, including
+  the following files:
+    - daffodil-schematron/src/main/resources/iso-schematron-xslt2/ExtractSchFromXSD-2.xsl
```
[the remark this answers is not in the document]
jw3: ok.  looks like the other schematron files above need corrected also, agree?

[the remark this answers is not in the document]
stevedlawrence: Yep. Looks like those have the same issues.

daffodil-schematron/src/main/resources/META-INF/LICENSE
```diff
@@ -237,3 +237,28 @@ subcomponents is subject to the terms and conditions of the following licenses.
   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
   SOFTWARE.
+
+  This product bundles content from the Schematron converters, including
+  the following files:
+    - daffodil-schematron/src/main/resources/iso-schematron-xslt2/ExtractSchFromXSD-2.xsl
```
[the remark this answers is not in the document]
jw3: :ok_hand:

mbeckerle (approved):
+1 As long as there are no remaining license-related issues I'm good with the changes.

daffodil-schematron/src/test/scala/org/apache/daffodil/validation/schematron/TestEmbeddedSchematron.scala
jw3: Done.  This was a good exercise.  :+1:

daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
```diff
@@ -400,7 +400,12 @@ trait AnnotatedMixin { self: AnnotatedSchemaComponent =>
       {
         ai.attribute("source") match {
           case None => {
-            this.SDW(WarnID.AppinfoNoSource, """xs:appinfo without source attribute. Is source="http://www.ogf.org/dfdl/" missing?""")
+            // if a child node in the dfdl namespace exists we will provide a warning about using the source property
+            ai.child.flatMap(n => Option(n.namespace)).find{_.contains("dfdl")}.foreach { _ =>
```
stevedlawrence, line 404: In the ``case Some`` below, we use contains("ogf") || contains("dfdl") to consider a DFDL namespace "acceptable". Should we use the same logic here?

daffodil-test/src/test/resources/org/apache/daffodil/section02/schema_definition_errors/SchemaDefinitionErrors.tdml
```diff
@@ -203,9 +203,11 @@
...
     <tdml:document><![CDATA[test]]></tdml:document>
 
-    <tdml:errors>
-      <tdml:error />
-    </tdml:errors>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <elem>test</elem>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
```
stevedlawrence, line 210: Just want to point out that if there is not a ``<tdml:warning>`` element in a test, the TDML runner doesn't care if there are warnings or not. The TMDL Runner only requires that warnings exist if the ``<tdml:warnigns>`` element exists. So unfortunately there's no way test that no warnings were output. Probably still worth having this, since some point we will want to add that feature and if we are ignoring a warning here, that would be caught some day.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestValidating.scala
```diff
@@ -0,0 +1,45 @@
...
+import net.sf.expectit.matcher.Matchers.regexp
+import org.junit.Test
+
+class TestValidating {
+  val data = "input/uuid.txt"
+  val uuid = "xsd/string.dfdl.xsd"
+  val never = "sch/never-fails.sch"
+  val always = "sch/always-fails.sch"
+
+  val alwaysResult = regexp("<.+-fails>2f6481e6-542c-11eb-ae93-0242ac130002</.+-fails>\n")
```
stevedlawrence, line 29: Tests in windows are failing, perhaps due to this \n? Is this \r\n on windows? Can we just make this tests loser so it doesn't care about the newline?

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
```diff
@@ -0,0 +1,105 @@
...
+    val schema = "xsd/with-ns-1.dfdl.xsd"
+    s"parse --validate schematron={{$schema}} -s {{$schema}} $data" -> contains("</myns:interval>\n")
+  }
+
+  @Test def testWithNs3(): Unit = withShell(FailureErrorCodeResult) {
+    val data = mktmp("0;0")
+    val schema = "xsd/with-ns-1.dfdl.xsd"
+    s"parse --validate schematron={{$schema}} -s {{$schema}} $data" -> contains("</myns:interval>\n")
+  }
+}
```
stevedlawrence, line 105: Lot's of \n's in this file. Likely going to cause issues on windows.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
jw3: done

daffodil-test/src/test/resources/org/apache/daffodil/section02/schema_definition_errors/SchemaDefinitionErrors.tdml
jw3: Ah, ok.  Handling the diagnostics output on stderr in the cli tests must have had me thinking this was checked, but I see what you mean now.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
```diff
@@ -0,0 +1,105 @@
...
+    s"parse --validate schematron={{$schema}} -s {{$schema}} $data" -> contains("</name>\n")
+  }
+
+  @Test def testWithNs1(): Unit = withShell() {
+    val data = mktmp("0;1")
+    val schema = "xsd/with-ns-1.dfdl.xsd"
+    s"parse --validate schematron={{$schema}} -s {{$schema}} $data" -> contains("</myns:interval>\n")
+  }
+
+  @Test def testWithNs2(): Unit = withShell(FailureErrorCodeResult) {
```
stevedlawrence, line 94: Some of these tests check for FailureErrorCodeResults, but looked like they are checking for an infoset? Does that mean these tests fail validation? Is it worth also checking for a validation error in that case?

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestValidating.scala
jw3: Ill try that.  Looked like my paths were mangled in the logs, but maybe a red herring... I dont have a windows box to test so just going to take a few stabs at it and see what happens  :boom: :hammer:

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
jw3: Roger.  Will clean this up.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
```diff
@@ -0,0 +1,109 @@
...
+   */
+  def mktmp(d: Array[Byte]): Path = {
+    val f = File.createTempFile("schval", "data")
+    val os = new FileOutputStream(f)
+    os.write(d)
+    os.close()
+    f.toPath
+  }
+  def mktmp(str: String): Path = mktmp(str.getBytes)
+}
```
stevedlawrence, line 109: This is great! But it fells like it belongs in ``CLI/Util.scala`` for other tests to use. Though, I guess there are lots of little bits that at are very specific to schematron tests, so maybe it's not quite ready? For example, single curly braces means the resource is in daffodil-schematron is very specific to this (and honestly, not very obvious).

Something like this is definitely needed, and is almost certainly how DAFFODIL-2381 should be resolved.

However, I don't really want there to be two different ways to write CLI tests long term, so we need to make sure we document that this is written differently from others CLI tests, and that that should probably be fixed as part of DAFFODIL-2381. Perhaps add a comment to DAFFODIL-2381 to specifically point out this code, and explain that this should become the basis for those changes, and maybe mention anything that is specific to schematron that likely needs to be replaced.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
stevedlawrence, line 105: Can you delete these tmp files when a test is finished? Polluting tmp and not cleaning it up is something Daffodil tests are really bad at, and we have dozens of different places and ways we create temp files. Some we need to be better at.

stevedlawrence:
:+1: Looks really nice! Loving the recent improvements to the CLI tests, hoping that can be the basis for cleaning up all of our CLI test boilerplate in the future. Just need to get things passing on windows, and squash all the commits and I think this ready to merge.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
jw3: that made for a nice refinement

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/TestEmbedded.scala
jw3: Yes. What strted the entire abstraction of temp file creation was to set deleteOnExit() for all these, and I ended up forgetting it :laughing: 

done

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
jw3: Yeah, needs more work, consider it a reference implementation.  I added a comment for DAFFODIL-2381 on the package object, will comment on the jira ticket.

> (and honestly, not very obvious)

Noted this behavior in the doc on the package object.

jw3:
Windows issues fixed up, squashed and rebased.  Ready to go I believe.

stevedlawrence:
Looks good! I think you've got all the requisite +1's. Feel free to merge and resolve the Jira ticket.

PR #464, opened 2020-12-03
Fixes sonarcloud duplication error

olabusayoT:
Sonarcloud is currently failing due to the sapi/japi and Main.scala code exceeding the duplication threshold. We add sapi/japi directories and Main.scala to cpd exclusions list so duplication checks are not done on them.

I was not able to replicate the failure state on my local instance, so this PR technically was not tested for the fix. It was tested to ensure a valid check could be run.


stevedlawrence (approved):

stevedlawrence:
+1 Only suggestion is to add a bug number to the commit message, I think DAFFODIL-2275 is reasonable--not sure this needs a new bug.

mbeckerle (approved):
+1

olabusayoT:
This change does nothing to affect codecoverage so we are merging it despite the failing codecov/project check

PR #465, opened 2020-12-07
Reorganize C source files

tuxji:
As requested on the dev list, reoganize C source files into two
distinct groups.  Separate these groups from each other by moving them
into two subdirectories, "c/libcli" and "c/libruntime".

In build.sbt, ensure C compiler test-builds these subdirectories.

In TestCLIGenerateC.scala, update tests to keep them passing (some
tests were checking the location and existence of the
"generated_code.c" file).

In resources/c, move the C source files into libcli and libruntime.

In CodeGenerator.scala, make CodeGenerator write generated_code.[ch]
into libruntime, find C source files in any subdirectory, and compile
them with two include flags (one for each subdirectory).

In TestCodeGenerator.scala, make all tests remove temporary output dir
after using it, and update tests to keep them passing (some tests were
checking the location and existence of the "generated_code.c" file).


build.sbt
```diff
@@ -57,13 +57,15 @@ lazy val runtime2         = Project("daffodil-runtime2", file("daffodil-runtime2
                                 Compile / ccTargets := ListSet(runtime2CFiles),
                                 Compile / cSources  := Map(
                                   runtime2CFiles -> (
-                                    ((Compile / resourceDirectory).value / "c" * GlobFilter("*.c")).get() ++
+                                    ((Compile / resourceDirectory).value / "c" / "libcli" * GlobFilter("*.c")).get() ++
+                                    ((Compile / resourceDirectory).value / "c" / "libruntime" * GlobFilter("*.c")).get() ++
```
stevedlawrence, line 61: Just an FYI, you can use ** to get a recusrive glob, for example, you could instead do something like:
```scala
((Compile / resourceDirectory).value / "c" / ** GlobFilter("*.c")).get()
```
Though, being explicit might have some advantages?

Another question, since the libcli and libruntime directories are both added to runtime2CFiles, does that mean both libs end up in the same .so/.a files? Same goes for the examples file? If we're separating out source, do we also want to separate the resulting compiled objects? Or is this purley organizational?

Another related question, I see we are also compiling the examples files into this same .a/so file? Should that be separate as well? And chance of conflicts between examples, libcli, and libruntime?

daffodil-runtime2/src/test/scala/org/apache/daffodil/runtime2/TestCodeGenerator.scala
```diff
@@ -26,18 +26,31 @@ import org.apache.daffodil.compiler.Compiler
...
- * functionality of CodeGenerator's generated code for as many DFDL
- * schemas as you could want.
+ * to [[CodeGenerator]] for a single test DFDL schema.  Running TDML
+ * tests with daffodil-runtime2 is a more effective way to check that
+ * CodeGenerator can generate appropriate code for as many DFDL schemas
+ * as you could want.
  */
 class TestCodeGenerator {
+  // Ensure all tests remove tmpOutputDir after using it
+  val tmpOutputDir: os.Path = os.pwd/"generate_tmp"
```
stevedlawrence, line 44: Any particular reason to not use ``Files.createTempDirectory``? That has the advantage that it goes in /tmp, and ensures theat you have a unique, empty directory. Avoids the need for the before function to clean it out. I also noticed one of the windows tests failed in here somewhere with an error about this "generate_tmp" directory. I wonder if there is some kind of race condition here?

daffodil-runtime2/src/test/scala/org/apache/daffodil/runtime2/TestCodeGenerator.scala
```diff
@@ -78,35 +91,29 @@ class TestCodeGenerator {
     val cg = pf.forLanguage("c")
 
     // Generate code from the test schema successfully
-    val outputDir = cg.generateCode(None, "./generateCode_tmp")
+    val outputDir = cg.generateCode(None, s"$tmpOutputDir")
```
stevedlawrence, line 94: Thoughts on having each test create it's own temp directory? That way there's no possibility for one test to interfere with another test?

I've also wished for a while that we had a function that all our tests could use so we could have a standard way to create temp directories. Right now lots of tests do it differently and we pollute /tmp with a bunch of files/directories. It would be nice if had something like this instead:

```scala
object DaffodilUtil {
  private val rootTempDir = Paths.get(System.getProperty("java.io.tmpdir"), "daffodil")
  def withDaffodilTempDir[T](prefix: String = "")(body: Path => T): T = {
      Files.createDirectories(rootTempDir)
      val tmpDir = Files.createTempDirectory(rootTempDir, prefix)
      try {
        body(tmpDir)
      } finally {
        // recurisively delete tmpDir
      }
  }
}
```

Then, we can update individual tests, and maybe even things like the TDML runner, that need a temp directory can just do something like:
```scala
DaffodilUtil.withTempDir() { tmpDir => 
  // do test using tmpDir, it will automatically be deleted when complete
}
```

Might be out of scope for this change though. If you don't want to add somethign like this to this PR, I'll create a bug so we don't forget. I've had this in the back of my head for a while.

stevedlawrence (approved):
Just a couple questions, looks good.

stevedlawrence:
+1

daffodil-runtime2/src/test/scala/org/apache/daffodil/runtime2/TestCodeGenerator.scala
tuxji: Yes, I also noticed that one of the Windows builds failed because this directory couldn't be deleted, probably because both the daffodil-cli and daffodil-runtime2 tests were running at the same time and trying to use this directory at the same time.  Since I've been using os-lib calls for all my file operations, I've replaced the os.pwd/"generate_tmp" with calls to os.temp.dir() in the test methods.

daffodil-runtime2/src/test/scala/org/apache/daffodil/runtime2/TestCodeGenerator.scala
tuxji: Nice idea, although I've simply added calls to os.temp.dir() so each test method creates its own temp directory and kept using `@After` to remove the temp directory.  (The os-lib documentation says the directory created by os.temp.dir() is supposed to be removed when the JVM exits, but I've never found that to happen in practice, bummer.)

tuxji (commented):
I'll push a commit to address the feedback and wait for another +1 before squashing the commits and merging.

mbeckerle (approved):
+1

PR #466, opened 2020-12-09
Expand C code generator to handle more integer types

tuxji:
Add another example schema to runtime2 tests and make C code generator
able to handle more integer types.  Starting to run into alignment
issues currently causing different behavior between runtime1 and
runtime2 and will need to handle these issues later.

In xml_reader.c, rename xmlInt32Elem to xmlIntegerElem and expand it
to read 8, 16, 32, and 64-bit signed/unsigned integer numbers from XML
elements.  Have not tested all cases yet.

In xml_writer.c, rename xmlInt32Elem to xmlIntegerElem and expand it
to write 8, 16, 32, and 64-bit signed/unsigned integers as XML element
values.  Have not tested all cases yet.

In infoset.c, expand walkInfosetNode to call visitIntegerElem method
for 8, 16, 32, and 64-bit signed/unsigned integer type codes.

In infoset.h, rename VisitInt32Elem type to VisitIntegerElem and
visitInteger32Elem field to visitIntegerElem.  Add 8, 16, 32, and
64-bit signed/unsigned integer type codes to TypeCode.

In ex_int32.c, update to reflect changes to generated code actually
generated from ex_int32 root element in TestRuntime2.tdml.

In ex_int32.h, update to reflect changes to generated code actually
generated from ex_int32 root element in TestRuntime2.tdml.

In Runtime2DataProcessor.scala, make parse and unparse methods remove
temporary directories after using them and make ParseResult convert
output file to XML immediately instead of waiting until called later
since output file will no longer be around.

In BinaryIntegerKnownLengthCodeGenerator.scala, remove some
restrictions from when code generator could handle only 32-bit signed
integers.  Modify generated code more flexible and capable of handling
different integer lengths.

In CodeGeneratorState.scala, expand CodeGeneratorState to handle 8,
16, 32, and 64-bit signed/unsigned integers.  Modify generated code to
define dummy be8toh and htobe8 functions since <endian.h> doesn't
define any endian byte-swapping functions for 8-bit integers.

In ElementParseAndUnspecifiedLengthCodeGenerator.scala, allow
alignment fills to precede elements in order to allow code to be
generated, although we still need to generate some extra code to
adjust for alignment fills later.

In TestRuntime2.dfdl.xsd, expand test schema to add another schema
example coming from somewhere else (an orion_command element using
8-bit and 16-bit signed/unsigned integers).  This element has
inadvertently required an alignment fill of 1 byte before the checksum
field that probably should be removed, but we may need to generate
code to handle alignment fills anyway.

In TestRuntime2.tdml, rename parse_int32 & unparse_int32 tests to
ex_int32_parse & ex_int32_unparse and add two new tests
orion_command_parse & orion_command_unparse.  Specify explicit root
elements for every test since we have multiple root elements now.

Rename parse_int32 & unparse_int32 data files to ex_int32_parse &
ex_int32_unparse.  Generated code doesn't expect XML elements to have
an ex prefix anymore (need to check why) so remove ex prefixes from
ex_int32_unparse elements.

Add orion_command_parse && orion_command_unparse data files for new
tests.

In TestRuntime2.scala, call ex_int32 & orion_command tests.

In Rat.scala, exclude renamed & new binary data files from checks.


daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryIntegerKnownLengthCodeGenerator.scala
```diff
@@ -36,50 +38,47 @@ trait BinaryIntegerKnownLengthCodeGenerator {
...
-      e.SDE("Only signed integers are supported.")
-
-    val initStatement = s"    instance->$fieldName = 0xCDCDCDCD;"
+    // Start generating code snippets
+    val initialValue = lengthInBits match {
+      case 8 => "0xCD"
+      case 16 => "0xCDCD"
+      case 32 => "0xCDCDCDCD"
+      case 64 => "0xCDCDCDCDCDCDCDCD"
+      case _ => e.SDE("Lengths other than 8, 16, 32, or 64 bits are not supported.")
```
stevedlawrence, line 59: What is this 0xCD magic number?

project/Rat.scala
```diff
@@ -114,7 +114,8 @@ object Rat {
     file("daffodil-sapi/src/test/resources/test/sapi/myData16.dat"),
     file("daffodil-sapi/src/test/resources/test/sapi/myData19.dat"),
     file("daffodil-sapi/src/test/resources/test/sapi/myDataBroken.dat"),
-    file("daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/parse_int32"),
+    file("daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_int32_parse"),
+    file("daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion_command_parse"),
```
stevedlawrence, line 118: Can you add .bin, .dat or something just so it makes it more clear that these are binary files? Makes it more obvious the reason why we need to exclude them from rat.

Also, can these runtime-2 excludes be sorted so they are before daffodil-sapi?

stevedlawrence (approved):

stevedlawrence:
Looks nice! Just had some licesning/dependency questions that I missed in previous reviews.

daffodil-runtime2/src/main/resources/c/libcli/xml_reader.c
```diff
@@ -184,10 +184,50 @@ xmlInt32Elem(XMLReader *reader, const ERD *erd, int32_t *location)
     {
         if (strcmp(name_from_xml, name_from_erd) == 0)
         {
-            // Check for any errors getting the 32-bit integer
+            // Check for any errors getting the integer number
             const char *errstr = NULL;
-            *location = (int32_t)strtonum(number_from_xml, INT32_MIN, INT32_MAX,
-                                          &errstr);
+
+            // Need to handle varying bit lengths and signedness
```
mbeckerle, line 190: Is this a TODO? If so add "TODO", otherwise remove "Need to"

daffodil-runtime2/src/main/resources/c/libcli/xml_writer.c
```diff
@@ -90,29 +90,64 @@ xmlEndComplex(XMLWriter *writer, const InfosetBase *base)
...
+    // Set namespace declaration if necessary
+    const char *xmlns = get_erd_xmlns(erd);
     if (xmlns)
     {
         const char *ns = get_erd_ns(erd);
         mxmlElementSetAttr(simple, xmlns, ns);
     }
-    return (simple && text) ? NULL : "Error making new int32 simple element";
+
+    // Need to handle varying bit lengths and signedness
```
mbeckerle, line 110: "need to" again

daffodil-runtime2/src/main/resources/examples/ex_int32.c
```diff
@@ -99,42 +99,45 @@ static const ERD c2_ERD = {
...
+    static ex_int32    instance;
     InfosetBase *root = &instance._base;
-    c1_ERD.initSelf(root);
+    ex_int32_ERD.initSelf(root);
     return root;
 }
 
 // Methods to initialize, parse, and unparse infoset nodes
 
+static inline uint8_t be8toh(uint8_t be8b) { return be8b; }
```
mbeckerle, line 138: Add a comment. I don't understand the point of these functions that just return the arg. They appear to be degenerate cases of endian-sensitive integer transformations for the 1-byte case. Do you use these?

In this naming schema "beNtoh" what's the "h" stand for?

daffodil-runtime2/src/main/resources/examples/ex_int32.c
```diff
@@ -25,15 +25,15 @@
 static void        c2_initSelf(c2 *instance);
 static const char *c2_parseSelf(c2 *instance, const PState *pstate);
 static const char *c2_unparseSelf(const c2 *instance, const UState *ustate);
-static void        c1_initSelf(c1 *instance);
-static const char *c1_parseSelf(c1 *instance, const PState *pstate);
-static const char *c1_unparseSelf(const c1 *instance, const UState *ustate);
+static void        ex_int32_initSelf(ex_int32 *instance);
```
mbeckerle, line 28: is this example for signed int 32 or unsigned. There's some mixing below.

daffodil-runtime2/src/main/resources/examples/ex_int32.c
```diff
@@ -149,7 +152,7 @@ c2_parseSelf(c2 *instance, const PState *pstate)
     const char *error_msg = NULL;
     if (!error_msg)
     {
-        char   buffer[4];
+        char   buffer[sizeof(uint32_t)];
```
mbeckerle, line 155: Why does this use uint32_t instead of int32_t ? For consistency's sake, we should use int32_t I think.

daffodil-runtime2/src/main/resources/examples/ex_int32.c
```diff
@@ -206,20 +209,20 @@ c2_unparseSelf(const c2 *instance, const UState *ustate)
...
 
 static const char *
-c1_parseSelf(c1 *instance, const PState *pstate)
+ex_int32_parseSelf(ex_int32 *instance, const PState *pstate)
 {
     const char *error_msg = NULL;
     if (!error_msg)
     {
-        char   buffer[4];
+        char   buffer[sizeof(uint32_t)];
```
mbeckerle, line 225: same issue in this file. SHould be int32_t ?

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryIntegerKnownLengthCodeGenerator.scala
mbeckerle: If this is just a noise pattern so that it is not zero but visibly recognizable on dumps, just add a comment to that effect.

I think this does *not* need to be the dfdl:fillByte as these integers are written atomically, all bytes; hence, the pattern will only be visible in a debugger for the time between this initialization and when the location is written with the integer value.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -260,6 +272,9 @@ class CodeGeneratorState {
          |
          |// Methods to initialize, parse, and unparse infoset nodes
          |
+         |static inline uint8_t be8toh(uint8_t be8b) { return be8b; }
```
mbeckerle, line 275: So are these here because they aren't in the standard library for this sort of thing, yet we generate calls to a converter with this naming convention even for the 8-bit case?

mbeckerle (approved):
+1 I just suggest some comments be added here and there.

tuxji (commented):
Update pull request in response to reviewer feedback.  Lift byteOrder
restriction so both bigEndian and littleEndian integers can be tested.
Expand test schema to test all integer sizes, signs, and byte orders.

tuxji:
Coincidentally, Scala Times just included a blog post about os-lib (<https://mungingdata.com/scala/filesystem-paths-move-copy-list-delete-folders/>) in their latest [newsletter](https://mailchi.mp/softwaremill/scala-times-issue-2609176).

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala
```diff
@@ -108,6 +108,8 @@ class Runtime2DataProcessor(executableFile: os.Path) extends DFDL.DataProcessorB
         val parseResult = new ParseResult(outfile, Failure(parseError))
         parseResult.addDiagnostic(parseError)
         parseResult
+    } finally {
+      os.remove.all(tempDir)
```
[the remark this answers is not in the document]
stevedlawrence: OS-lib does look interesting. We should consider updates to use it throughout Daffodil it we think it's worth using in one part. I'd really like to keep the codebase as consistent as possible.

Considering MPL has extra licensing requirements, and if we're only using directories-jvm for getting a per OS cache directory, I've vote for just implementing that logic ourselves.

THe NOTICE requirement ususally only comes from Apache licensed things. So probably no change needs to be made for the bin.NOTICE. The MPL rquirement suggests adding some line in the REAMDE. The apache licensing page says:

> An appropriate and prominent label is a label the user will read while learning about the distribution - for example in a README, and it should identify the third-party product, its licensing, and a url to the its homepage. Please also ensure to comply with any attribution/notice requirements in the specific license in question.

So if we do include this, we probably need to include that information the daffodil-cli README file.

PR #468, opened 2020-12-14
Fix abort when viewing prefixed length elements in the debugger

stevedlawrence:
The InfosetWalker, which is used for displaying the infoset, expects
elements to have a container element. When creating a detached element
for prefixed elements, we did so without a DIDocument. This causes
assertions to fail or null pointers to be hit. To ensure the detached
element used for prefixed parsing/element is appropriate the
InfosetWalker, create a "detached" DIDocument and the new detached
DIElement to that.

DAFFODIL-2348


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
```diff
@@ -385,8 +388,11 @@ class SpecifiedLengthPrefixedUnparser(
   override lazy val childProcessors = Vector(prefixedLengthUnparser, eUnparser)
 
   override def unparse(state: UState): Unit = {
-    // create a "detached" element that the prefix length will be used to unparse
-    val plElem = Infoset.newElement(prefixedLengthERD).asInstanceOf[DISimple]
+    // Create a "detached" DIDocument with a single child element that the
+    // prefix length will be parsed to. This creates a completely new
+    // infoset and parses to that, so care is taken to ensure this infoset
+    // is only used for the prefix length parsing and is removed afterwards
+    val plElem = Infoset.newDetachedElement(state, prefixedLengthERD).asInstanceOf[DISimple]
```
mbeckerle, line 395: I'm a bit concerned about the overhead of this added node, but we should deal with that via profiling. There are many many overheads in Daffodil, no reason to think this one will be the most important.

mbeckerle (approved):
+1

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SpecifiedLengthUnparsers.scala
stevedlawrence: Perhaps an alternative would be to get the current DIDocument and add this detached element as a child of that? That avoids the additional allocation, but then cleanup becomes a bit trickier? I guess it shouldn't be hard to remove that new child? There might also be considerations with the InfosetWalker, it might have some assumptions that the DIDocument only has a single node. Probably surmountable, but does add some complexity.

jadams-tresys (approved):
+1

PR #469, opened 2021-01-08
Add tests confirming round-trip behavior of date/times

stevedlawrence:
Due to canonicalization, we may sometimes lose location specific time
zone information, resulting in unparing time zones to generic offsets.
This means they do not round trip. Some tests simply disabling
roundtripping, others use two-pass which hid this issue. For these
tests, this disables round tripping and adds unparse specific tests to
ensure we unparse these location specific patterns correctly. Other
tests are tweaked so that they can round trip where possible, usually
due to optional zero-padding in calendar pattern.

DAFFODIL-1580


mbeckerle (approved):
+1

tuxji (approved):
+1

PR #470, opened 2021-01-13
Add support for displaying variables in the debugger

stevedlawrence:
Adds a new info subcommand called "variables" to show the current
in-scope state and value of variables. For example:

    (debug) info variables
      variables:
        ex:var1:  (undefined)
        {http://www.ogf.org/dfdl/dfdl-1.0/}binaryFloatRep: ieee (default)
        {http://www.ogf.org/dfdl/dfdl-1.0/}byteOrder: bigEndian (default)
        {http://www.ogf.org/dfdl/dfdl-1.0/}encoding: UTF-8 (default)
        {http://www.ogf.org/dfdl/dfdl-1.0/}outputNewLine: %LF; (default)

Also allows specifcying one or more variable names to only output those
if you don't want to see all variables, for example:

    (debug) info varibles var1 byteOrder
      variables:
        ex:var1:  (undefined)
        {http://www.ogf.org/dfdl/dfdl-1.0/}byteOrder: bigEndian (default)

This also modifies the info command so that you can provide parameters
to subcommands, which is needed to allow variable names after the "info
variables".

Fixes the "info data" command which at one point
had optional parameters to control how data should be output (either
text or binary). This was broken at some point, but is now available.

Allows the "clear" command to be used in the "display" command, e.g.

  (debug) display clear

This can be useful to clear the screen every time the debugger pauses,
sometimes making it easier to see what changed.

DAFFODIL-2453

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
@@ -940,12 +940,12 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
       val desc = "show value of expression each time program stops"
       val longDesc = """|Usage: di[splay] <debugger_command>
                         |
-                        |Execute a debugger command (limited to eval or info) every time a debugger
-                        |console is displayed to the user.
+                        |Execute a debugger command (limited to eval, info, and clear) every time a
+                        |there is a pause in the debugger.
                         |
                         |Example: display info infoset""".stripMargin
       override lazy val short = "di"
-      override val subcommands = Seq(Eval, Info)
+      override val subcommands = Seq(Eval, Info, Clear)
 
       def act(args: Seq[String], prestate: StateForDebugger, state: ParseOrUnparseState, processor: Processor): DebugState.Type = {
         DebuggerConfig.displays += new Display(DebuggerConfig.displayIndex, args)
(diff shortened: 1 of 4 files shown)


mbeckerle (approved):
+1 Pretty important feature now, with variables being used more now. 

Does Trace dump the variables every time? Or no change to trace behavior?

tuxji (approved):
+1

Coverage report looks good, although a CLI "info data text/binary" test would make almost all of the newly added code covered by tests.

stevedlawrence:
> Does Trace dump the variables every time? Or no change to trace behavior?

Currently no change in trace behavior. We could make it so trace dumps all variables every step, but that feels abit too verbose to me (but maybe that's expected with trace?). Another option might be to modify the "diff" command to support printing variable changes. That's could probably be useful, but it's maybe a bit of an effort--so maybe it makes sense to be part of a separate PR?

stevedlawrence:
> Coverage report looks good, although a CLI "info data text/binary" test would make almost all of the newly added code covered by tests.

Agreed, will add.

stevedlawrence:
The diff approach was actually not as complicated as I thought it might be. I've added it to this commit. Let me know if that seems like a reasonable approach to have variables show up during trace mode, which does include the info diff command.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1404,6 +1469,24 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
               if (prestate.arrayPos != state.arrayPos) { debugPrintln("occursIndex: %d -> %d".format(prestate.arrayPos, state.arrayPos), "  "); diff = true }
               if (prestate.groupPos != state.groupPos) { debugPrintln("groupIndex: %d -> %d".format(prestate.groupPos, state.groupPos), "  "); diff = true }
               if (prestate.childPos != state.childPos) { debugPrintln("childIndex: %d -> %d".format(prestate.childPos, state.childPos), "  "); diff = true }
+              prestate.variableMap.qnames.foreach { qname =>
```
mbeckerle, line 1472: Yeah, this is good. diff shows variables that changed. Great. 
+1

PR #471, opened 2021-01-13
Support arrays, real numbers, and improve TDML processing

tuxji:
Expand the C code generator to support both binary real numbers and
fixed occurrence arrays of binary numbers.  Reduce the size of the
generated C code by emitting one-line function calls instead of 14-16
lines of low level parsing & unparsing C code for each binary number.
Move the 14-16 lines of low level code into functions within parsers.c
and unparsers.c in the c/libruntime directory instead.

Remove the orion-command element from TestRuntime2.dfdl.xsd and add it
back as a new test schema (orion-command.dfdl.xsd), tests
(orion-command.tdml), and data files (command_parse.dat/unparse.xml,
video_settings_parse.dat/unparse.xml).  Rename the ex_ints element in
TestRuntime2.dfdl.xsd to ex_nums and add arrays and real numbers to
its schema and data files.

Fix file-writing/reading collision problems when both TestOrionCommand
and TestRuntime2 try to compile and run C code at the same time.  Get
rid of dev.dirs % directories-jvm and make the TDML processor write to
a temporary unique directory for each test which gets deleted after
each test.  Thankfully, the "zig cc" command now caches previously
built files in zig's own global cache directory instead of a local
zig_cache directory so the TDML processor no longer needs to write to
a daffodil cache directory anymore.

Changelog:

In main.yml, install pkgconf instead of pkg-config in msys2 setup
since pkgconf is newer and better.

In README.md, add new build requirements so developers will know they
will need a C compiler and Mini-XML library to build Daffodil.

In bin.LICENSE, remove license for dirs-dev / directories-jvm.

In TestCLIGenerateC.scala, rename ex_ints to ex_nums.

In c/Makefile, change CFLAGS for easier debugging and remove unneeded
libraries.

Run include-what-you-use on C source files in verbose mode to update
comments saying exactly what is used from each include in these files:

 - daffodil_argp.c
 - daffodil_main.c
 - xml_reader.c
 - xml_writer.c
 - infoset.c
 - parsers.[ch]
 - unparsers.[ch]
 - ex_nums.[ch]

In daffodil_main.c, update the parseSelf and unparseSelf calls since
they now return void instead of const char *error_msg.

In xml_reader.c, define new strtofnum and strtodnum functions to
support reading floating point numbers from XML data.  Rename
xmlIntegerElem to xmlNumberElem and make it call strtofnum and
strtodnum as well as strtounum and strtonum.

In xml_writer.c, rename xmlIntegerElem to xmlNumberElem and make it
write floating point numbers as well as integers.  Use G format
specifier and call fixNumberIfNeeded on number's text afterwards to
ensure that [-]INF and NaN will be capitalized correctly according to
the XML Schema's xsd:float specification.  We still need to make
runtime1 and runtime2's XML floating point output match each other
even more closely or make TDML Runner compare real numbers
numerically, not textually, so it can ignore minor textual differences
without ignoring numerical differences.

In infoset.c, make walkInfosetNode call visitNumberElem for primitive
floats/doubles as well as primitive integers.

In infoset.h, change signatures of parseSelf and unparseSelf.  Rename
VisitIntegerElem to VisitNumberElem.  Add enumerations for primitive
floats/doubles.  Add const char *error_msg to PState and UState
structs so parser/unparser functions can set it when needed.  Add
(void) to rootElement() to fix -Weverything warning.

In parsers.{c,h}, define functions to parse binary real and integer
numbers.  Use macros to achieve DRY (Don't Repeat Yourself).  Note
that binary floating point numbers can be little-endian or big-endian
just like binary integers.

In unparsers.{c,h}, define functions to unparse binary real and
integer numbers.  Use macros to achieve DRY (Don't Repeat Yourself).

In ex_nums.{c,h}, rename from ex_ints.{c,h} to ex_nums.{c,h}.  Verify
real numbers work, then implement and test code to support a fixed
occurrence array, propagate changes to C code generator, and verify C
code generator produces same code.  Add a second array to ex_nums'
schema and verify C code generator generates appropriate code for it
too.  Make generator add (void) to rootElement() and (const char*) to
ERD offset computations to fix -Weverything warnings.

In CodeGenerator.scala, make generateCode return codeDir rather than
outputDir, pass codeDir rather than outputDir to compileCode, and
perform compilation in codeDir rather than outputDir since "zig cc"
will now cache previously built files in zig's global cache directory,
not a local zig_cache directory.

In Runtime2CodeGenerator.scala, add BinaryFloat and BinaryDouble cases
within the generateCode function and make them call
BinaryFloatCodeGenerator.  Also add and call a noop function to
facilitate setting a breakpoint within the generateCode function.

In BinaryFloatCodeGenerator.scala, initialize float fields with NaN
macro (NAN) since some platforms don't define signaling NaN macros
(SNAN & SNANF).  Emit one-line function calls to parse and unparse
32-bit, 64-bit big-endian & little-endian binary floats.  Generate
code to initialize a floating point array, parse data into the array,
and unparse data from the array too.

In BinaryIntegerKnownLengthCodeGenerator.scala, remove minimum 8-bit
alignment restriction to allow schemas to say they want 1-bit
alignment although C runtime can't handle binary numbers smaller than
8 bits right now.  Emit one-line function calls to parse and unparse
binary 8-bit, 16-bit, 32-bit, 64-bit big-endian & little-endian binary
integers.  Generate code to initialize an integer array, parse data
into the array, and unparse data from the array too.  Call
e.schemaDefinitionUnless instead of e.SDE inside an if expression to
avoid uncovered line of code warning as Mike suggested.

In CodeGeneratorState.scala, add private qualifiedName and localName
methods to generate unique C file-scope names for ERD objects and
plain local names for structs/fields although we probably will need to
prepend namespace prefixes to plain local names to disambiguate some
structs/fields from each other later.  Make changes to support arrays
and binary floats as well as binary integers.  Include arrays in
struct declarations and ERD objects' count of children and offset
computations.  Change parseSelf and unparseSelf to return void instead
of const char *error_msg.  Shorten ${C}_compute_ERD_offsets to
${C}_compute_offsets.

In Runtime2TDMLDFDLProcessor.scala, make getProcessor create a
temporary unique directory per call instead of using the same cache
directory from dev.dirs.ProjectDirectories.  Get codeDir from
generateCode and pass codeDir to compileCode instead of outputDir.
Pass temporary directory to Runtime2TDMLDFDLProcessor,
Runtime2TDMLParseResult, and Runtime2TDMLUnparseResult and implement
cleanUp() in Runtime2TDMLParseResult and Runtime2TDMLUnparseResult to
allow us to delete temporary directory later.

In TestRuntime2.dfdl.xsd, rename ex_ints to ex_nums and make it define
floating point numbers as well as integer numbers.  Use nested
elements again to ensure more code is executed by tests according to
coverage report.  Add two fixed occurrence arrays (integer and float)
to ex_nums.  Move orion_command and its simple types to
orion-command.xml.

In TestRuntime2.tdml, rename ex_ints tests to ex_nums.  Move
orion_command tests to orion-command.tdml.  Use ex_nums_unparse1.xml
with runtime1 and ex_nums_unparse2.xml with runtime2 to demonstrate
differences in XML floating point output between both runtimes.

In command_parse.dat, rename from orion_command_parse.dat to
command_parse.dat.  Change tilt field's value from -21231 to -1231 to
fit new constraint in orion-command.dfdl.xsd.

In command_unparse.xml, rename from orion_command_unparse.xml to
command_unparse.xml.  Change tilt field's value from -21231 to -1231
to fit new constraint in orion-command.dfdl.xsd.

In ex_nums_parse.dat, rename from ex_ints_parse.dat to
ex_nums_parse.dat.  Add arrays and real numbers.  Make two of the real
numbers INF and NaN to ensure fixNumberIfNeeded gets tested.

In ex_nums_unparse1.xml, rename from ex_nums_unparse.xml, add arrays
and real numbers, and split into ex_nums_unparse1.xml and
ex_nums_unparse2.xml to demonstrate differences in XML floating point
output between both runtimes.

In orion-command.dfdl.xsd, add new test schema with Command,
CameraState, and VideoSettings elements.

In orion-command.tdml, add new test cases for Command and
VideoSettings elements.

In video_settings_parse.dat, add data for VideoSettings element.

In video_settings_unparse.xml, add infoset for VideoSettings element.

In TestCodeGenerator.scala, rename outputDir to tempDir, get codeDir
from generateCode, and pass codeDir to compileCode instead of
outputDir.

In TestOrionCommand.scala, run test cases in orion-command.tdml as
part of sbt tests.  Note that tests might run in parallel (that is,
TestOrionCommand.scala and TestRuntime2.scala can overlap each other
concurrently).

In TestRuntime2.scala, rename ex_ints to ex_nums.  Move orion_command
test cases to TestOrionCommand.scala.  Change names and add 2 more
methods to match renamed & new test cases in TestRuntime2.tdml.

In TDMLRunner.scala, replace duplicate lines of code deleting blob
files with cleanUp() calls.  Add two new cleanUp() calls in
UnparserTestCase to ensure unparse tests clean up temporary files too.

In TDMLDFDLProcessor.scala, add a new cleanUp() abstract function to
TDMLResult.

In DaffodilTDMLDFDLProcessor.scala, make DaffodilTDMLParseResult and
DaffodilTDMLUnparseResult implement cleanUp() to remove blob files or
do nothing respectively.

In Dependencies.scala, remove "dev.dirs" % "directories" since we
don't need it anymore.

In Rat.scala, tell Rat to ignore new/renamed test data files.


daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -217,10 +232,12 @@ class CodeGeneratorState {
         case PrimType.Int => "int32_t    "
         case PrimType.Short => "int16_t    "
         case PrimType.Byte => "int8_t     "
+        case PrimType.Float => "float      "
+        case PrimType.Double => "double     "
         case x => context.SDE("Unsupported primitive type: " + x)
       }
     } else {
-      child.namedQName.local + "         "
+      localName(child)
```
mbeckerle, line 240: Feels like you want qualfiedName(child) here, but you want to convert that into something you can use as a field name.

At some point we'll have to pick a preferred prefix for a namespace from the available prefixes for a namespace, and then construct a field name by prefix_localName or some such formula.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -247,10 +264,12 @@ class CodeGeneratorState {
     val erds = this.erds.mkString("\n")
     val finalImplementation = this.finalImplementation.mkString("\n")
     val code =
-      s"""#include "generated_code.h" // for generated code structs
-         |#include <endian.h>         // for be32toh, htobe32, etc.
-         |#include <stddef.h>         // for ptrdiff_t
-         |#include <stdio.h>          // for NULL, fread, fwrite, size_t, FILE
+      s"""#define __STDC_WANT_IEC_60559_BFP_EXT__
```
mbeckerle: Comment needed to explain what WANT_IEC_60559_BFP means. I think BFP stands for binary floating point?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
```diff
@@ -0,0 +1,92 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<tdml:testSuite
+  defaultConfig="config-runtime1"
```
mbeckerle, line 20: So if I want to run these tests on runtime2 I change this defaultConfig?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
```diff
@@ -0,0 +1,92 @@
...
+
+  <tdml:defineConfig name="config-runtime2">
+    <daf:tunables>
+      <daf:tdmlImplementation>daffodil-runtime2</daf:tdmlImplementation>
+    </daf:tunables>
+  </tdml:defineConfig>
+
+  <tdml:parserTestCase
+    description="orion-command Command parse test"
+    model="orion-command.xml"
```
mbeckerle, line 42: file name convention is ".dfdl.xml" or ".dfdl.xsd"

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0"?>
```
mbeckerle, line 1: file name convention ".dfdl.xsd" or ".dfdl.xml"

mbeckerle (approved):
+1  Minor things really, and note taking for the future. I.e., things not so much for this change set, but things I thought of when reading it. 

This is coming out quite well.

daffodil-runtime2/src/main/resources/c/libcli/xml_reader.c
```diff
@@ -97,6 +98,72 @@ strtounum(const char *numptr, uintmax_t maxval, const char **errstrp)
     return value;
 }
 
+// Convert an XML element's text to a float (call strtof with our own
+// error checking)
+
+static float
+strtofnum(const char *numptr, const char **errstrp)
```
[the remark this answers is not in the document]
tuxji: I see; that makes sense.  Instead of checking whether a const char *error_msg was returned, we can pass an error_info struct and check whether it has some error messages afterwards.  I don't think Daffodil has internationalized messages right now; has anyone asked for that support soon?

daffodil-runtime2/src/main/resources/c/libruntime/infoset.h
```diff
@@ -102,14 +100,16 @@ typedef struct InfosetBase
 
 typedef struct PState
 {
-    FILE *stream; // input to read from
+    FILE *      stream;    // input to read from
+    const char *error_msg; // to stop if an error happens
```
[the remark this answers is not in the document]
tuxji: I'm not sure when we will need to support backtracking in a binary representation with only a limited subset of DFDL 1.0 supported.  A collaborator my group is working with has been using `dfdl:discriminator test="{../../tagt eq 111}"` on message elements to perform discrimination from a choice of equally valid message elements, but I think his schema should be using dfdl:choiceDispatchKey instead which doesn't require backtracking.  When and where might we start to need backtracking, I'm wondering?

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -139,26 +159,18 @@ class CodeGeneratorState {
...
-         |    {
-         |        error_msg = ${C}_parseSelf(&instance->$e, pstate);
-         |    }""".stripMargin
-    val unparseStatement =
-      s"""    if (!error_msg)
-         |    {
-         |        error_msg = ${C}_unparseSelf(&instance->$e, ustate);
-         |    }""".stripMargin
+    val parseStatement = s"    ${C}_parseSelf(&instance->$e, pstate);"
+    val unparseStatement = s"    ${C}_unparseSelf(&instance->$e, ustate);"
```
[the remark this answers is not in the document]
tuxji: Actually, I spotted that CodeCov check myself and realized that I needed to add nested elements back to my ex_nums test element.  I added a second commit after your review which should ensure these lines are covered by the ex_nums tests now.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -139,26 +159,18 @@ class CodeGeneratorState {
...
     structs.top.initStatements += initStatement
     structs.top.parserStatements += parseStatement
     structs.top.unparserStatements += unparseStatement
   }
 
   def pushComplexElement(context: ElementBase): Unit = {
-    val C = context.namedQName.local
+    val C = localName(context)
```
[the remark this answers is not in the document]
tuxji: I haven't been qualifying C struct names yet.  I didn't think we would need to qualify them like we need to qualify C struct field names to avoid name collisions.  Now that I think about it, however, we could have two complex elements with the same local name in two different namespaces.  We would need to qualify these two complex elements' C struct names with their namespace prefixes to avoid a name collision.  Right now qualifiedName(context) would only qualify a complex element's local name with the local name of any parent complex element the complex element is nested inside.  Let's address how to qualify C struct names another time; we may want to treat them differently than C struct field names anyway.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
tuxji: Hmm, I didn't think we would need to qualify C struct field names since their names only need to be unique inside each C struct's scope.  Are you thinking somebody might define a complex element with two children (either simple or complex elements) having the same local name in two different namespaces too?  We would need a prefix_localName in that case too.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
tuxji: Yes.  I've been thinking that I'd like to simplify this boilerplate and do away with the config-runtime1 and config-runtime2.  Initially I thought Daffodil could look at defaultImplementations="daffodil daffodil-runtime2" and know it implements both of these, so it might as well run both of them.  However, I've found out that runtime1 prints floating point numbers to XML in a way that runtime2 can't print to XML compatibly without a lot of extra effort.  Therefore the same XML test data file won't pass in both runtime1 and runtime2.    Still, we might simplify the boilerplate by adding a new attribute that has the same effect as defaultConfig="config-runtime1" or defaultConfig="config-runtime2".

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
tuxji: Yeah.  This file came from a collaborator and I'd wanted to keep the same name, but I'll change the name.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.xml
tuxji: Ditto.

tuxji (commented):
My responses to Mike's comments.

Still waiting for a second +1, someone else please review when you can.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
tuxji: I needed this WANT_IEC_60559_BFP to make <math.h> provide the signaling NaN macros (SNAN, SNANF), but I ended up removing that define in my second commit because it didn't work on Windows.  I decided it wasn't worth trying to initialize fields with SNAN/SNANF (too unportable) and initialized them with NAN instead (a more portable macro).

daffodil-runtime2/src/main/resources/c/libcli/xml_reader.c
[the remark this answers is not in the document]
mbeckerle: Internationalized message support will eventually come up. It's very painful unless we put the support in early. This is also an issue of performance for backtracking (eventually someday)

daffodil-runtime2/src/main/resources/c/libruntime/infoset.h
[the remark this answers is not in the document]
mbeckerle: Not soon. My preference is create the deterministic DFDL subset first. Eventually people will want a fully functioning back-end.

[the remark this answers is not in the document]
mbeckerle: The other reason to do this is so that programmers won't write code assembling error messages from strings. They'll just pass a bunch of static pointers. This is less code clutter.

mbeckerle (approved):
+1

daffodil-runtime2/src/main/resources/c/libruntime/parsers.h
```diff
@@ -0,0 +1,52 @@
...
+extern void parse_le_uint32(uint32_t *number, PState *pstate);
+extern void parse_le_uint16(uint16_t *number, PState *pstate);
+extern void parse_le_uint8(uint8_t *number, PState *pstate);
+
+extern void parse_le_int64(int64_t *number, PState *pstate);
+extern void parse_le_int32(int32_t *number, PState *pstate);
+extern void parse_le_int16(int16_t *number, PState *pstate);
+extern void parse_le_int8(int8_t *number, PState *pstate);
+
+#endif // PARSERS_H
```
stevedlawrence, line 52: There's so much duplication in {paresrs,unparsers}.{c,h}. Any thoughts on using macros to reduce this?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
stevedlawrence: One option might be to add type awareness to the TDML runner. We already have that for calendars, hex binary, and blobs (see ``def textIsSame`` in ``XMLUtils.scala``). We could make it so if an infoset includes ``type="xs:float"`` or ``xs:double``, then we consider float values to be the same if they are within some small delta.

Or I wonder if the DFDL spec provides any guidance on how floating point values should be converted to base-10 strings for infoset? Maybe that's outside the scope of DFDL. Perhaps we should create a bug to make sure different runtimes output same floating point string reps? I assume runtime1 just uses float .toString() so we're just relying on Java to do the right thing, but maybe it's not doing the best thing in all cases.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
stevedlawrence: You mention that this orion files do not have a license? Can we get the contributor to license them as Apache? If not, I'd prefer we try to keep non-Apache compliant licenses out of the repo, even if they are on branches--it's too easy to forget about them once things get merged. At the very least, we need a scary comment somewhere (maybe Rat.scala) that these files are only temporarily ignored and do not follow apache license guidelines.

I'd also rather strive to get those binary files in TDML files. It's just easier to visiual the data that way and verify a test without having to fire up a hex editor to view the bin files.

stevedlawrence (approved):

stevedlawrence:
+1 with minor comments. Would like to also see the windows issues fixed, but I think based on another comment you fixed the portability already.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/orion-command.tdml
tuxji: I'm still waiting for a reply to an email I sent our collaborator last week whether we can license orion-command.xml as Apache.  I see the point about moving binary data into TDML files,  On the other hand having binary data files makes it easier to debug C code generated by the C code generator (compile with -g and run under the debugger with the binary data file as a command line argument).

tuxji (commented):
I've pushed a commit to address a comment or two and add support for fixed occurrence arrays too.  The Windows compilation problem is gone but I'm seeing the TestRuntime2 unit tests hang for some reason when running under Windows.  I'll try to address some more comments and that problem next.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryIntegerKnownLengthCodeGenerator.scala
```diff
@@ -18,70 +18,52 @@
...
     val e = g.e
-    val fieldName = e.namedQName.local
     val lengthInBits: Long = {
       e.schemaDefinitionUnless(e.elementLengthInBitsEv.isConstant, "Runtime dfdl:length expressions not supported.")
       val len = e.elementLengthInBitsEv.constValue.get
       len
     }
-
+    if (e.bitOrder ne BitOrder.MostSignificantBitFirst)
+      e.SDE("Only dfdl:bitOrder 'mostSignificantBitFirst' is supported.")
```
mbeckerle, line 36: If you want to get rid of this codecov annotation about this line, you can use 
e.schemaDefinitionWhen(....)
That embeds the predicate into the same line of code, so the predicate gets executed, which will satisfy codecov. 

We need a shorter alias though, should be SDEWhen(predicate, msgTemplate, args*)

tuxji (commented):
Most of feedback has been addressed, commits squashed, and all checks passed, so merging now.  Remaining feedback will be addressed later: 1) errorinfo objects instead of error strings; 2) struct / field names qualified with namespace prefixes when needed; 3) numerical instead of textual comparison of floating point numbers in TDML runner (def textIsSame in XMLUtils.scala) after discussion on dev list.

PR #473, opened 2021-01-20
Implement variable direction property

jadams-tresys:
This adds a 'direction' property to defineVariable statements with the
following options:
 - both (default)
 - parseOnly
 - unparseOnly

Current behavior is to create NadaParser/Unparser's for setVariable and
newVariableInstance statements when we are processing in the opposite
direction of the direction property while attempting to read a variable
while going the opposite direction will trigger a warning.

DAFFODIL-2429

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
@@ -27,10 +27,11 @@ import org.apache.daffodil.infoset.Infoset
 import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.processors.Evaluatable
 import org.apache.daffodil.processors.NonTermRuntimeData
-import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.Success
 import org.apache.daffodil.processors.TypeCalculator
 import org.apache.daffodil.processors.VariableRuntimeData
+import org.apache.daffodil.processors.{ VariableInProcess, VariableDefined }
+//import org.apache.daffodil.processors.SuspendableOperation
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.MaybeULong
 
@@ -77,9 +78,24 @@ final class SetVariableUnparser(
 
 }
 
+final class NewVariableInstanceSuspendableExpression(
+  override val expr: CompiledExpression[AnyRef],
+  override val rd: VariableRuntimeData)
+  extends SuspendableExpression {
+
+  override protected def processExpressionResult(ustate: UState, v: DataValuePrimitive): Unit = {
+    val vi = ustate.variableMap.find(rd.globalQName)
+    Assert.invariant(vi.isDefined)
+    vi.get.setState(VariableDefined)
+    vi.get.setValue(v)
+  }
+
+  override protected def maybeKnownLengthInBits(ustate: UState) = MaybeULong(0)
+}
+
 // When implemented this almost certainly wants to be a combinator
 // Not two separate unparsers.
-class NewVariableInstanceStartUnparser(override val context: RuntimeData)
+class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData)
   extends PrimUnparserNoData {
 
   override lazy val runtimeDependencies = Vector()
@@ -87,21 +103,25 @@ class NewVariableInstanceStartUnparser(override val context: RuntimeData)
   override lazy val childProcessors = Vector()
 
   override def unparse(state: UState) = {
-    val vrd = context.asInstanceOf[VariableRuntimeData]
-    state.newVariableInstance(vrd)
+    state.variableMap.newVariableInstance(context)
+    if (context.maybeDefaultValueExpr.isDefined) {
+      val maybeVar = state.variableMap.find(context.globalQName)
+      Assert.invariant(maybeVar.isDefined)
+      maybeVar.get.setState(VariableInProcess)
+      val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
+      suspendableExpression.run(state)
+    }
   }
 }
 
-class NewVariableInstanceEndUnparser(override val context: RuntimeData)
+class NewVariableInstanceEndUnparser(override val context: VariableRuntimeData)
   extends PrimUnparserNoData {
 
   override lazy val runtimeDependencies = Vector()
 
   override lazy val childProcessors = Vector()
 
-  override def unparse(state: UState) = {
-    state.removeVariableInstance(context.asInstanceOf[VariableRuntimeData])
-  }
+  override def unparse(state: UState) = state.variableMap.removeVariableInstance(context)
 }
 
 class TypeValueCalcUnparser(typeCalculator: TypeCalculator, repTypeUnparser: Unparser, e: ElementRuntimeData, repTypeRuntimeData: ElementRuntimeData)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
@@ -1745,6 +1745,7 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
             case VariableRead => "read"
             case VariableSet => "set"
             case VariableUndefined => "undefined"
+            case VariableBeingDefined => "being defined"
             case VariableInProcess => "in process"
           }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
@@ -33,6 +33,8 @@ import org.apache.daffodil.util.PreSerialization
 import org.apache.daffodil.util.TransientParam
 import org.apache.daffodil.xml.{GlobalQName, UnspecifiedNamespace, NamedQName, RefQName}
 import org.apache.daffodil.processors.parsers.PState
+import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.schema.annotation.props.gen.VariableDirection
 
 import scala.collection.mutable.{Map, ArrayBuffer}
 
@@ -41,6 +43,11 @@ sealed abstract class VariableState extends Serializable
 case object VariableUndefined extends VariableState
 
 case object VariableDefined extends VariableState
+
+/**
+ * Used to detect circular definitions of variables
+ */
+case object VariableBeingDefined extends VariableState
 
 case object VariableSet extends VariableState
 
@@ -141,19 +148,22 @@ class VariableInstance private (
     this._value = Some(v)
   }
 
+  def setDefaultValue(v: DataValuePrimitiveNullable) = this._value = Some(v)
+
   override def preSerialization: Unit = {
     defaultValueExpr
     value
     state
   }
 
-  override def toString: String = "VariableInstance(%s,%s,%s,%s,%s,%s".format(
+  override def toString: String = "VariableInstance(%s,%s,%s,%s,%s,%s,%s)".format(
                                                     state,
                                                     value,
                                                     rd,
                                                     defaultValueExpr,
                                                     priorState,
-                                                    priorValue)
+                                                    priorValue,
+                                                    this.hashCode)
 
   def copy(
     state: VariableState = state,
@@ -211,6 +221,10 @@ class VariableHasNoValue(qname: NamedQName, context: VariableRuntimeData) extend
   "Variable map (runtime): variable %s has no value. It was not set, and has no default value.".format(qname))
   with RetryableException
 
+class VariableSuspended(qname: NamedQName, context: VariableRuntimeData) extends VariableException(qname, context,
+  "Variable map (runtime): variable %s is currently suspended".format(qname))
+  with RetryableException
+
 /**
  * This expression can be thrown either at the start of parsing is the
  * expressions in a defineVariable are circular, or later during parsing if
@@ -244,6 +258,8 @@ final class VariableBox(initialVMap: VariableMap) {
   def setVMap(newMap: VariableMap): Unit = {
     vmap_ = newMap
   }
+
+  def cloneForSuspension(): VariableBox = new VariableBox(vmap.topLevelInstances)
 }
 
 /**
@@ -269,9 +285,9 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
     this(Map(topLevelVRDs.map {
       vrd =>
         val variab = vrd.createVariableInstance()
-        val abuf = new ArrayBuffer[VariableInstance]
-        abuf += variab
-        (vrd.globalQName, abuf)
+        val variableInstances = new ArrayBuffer[VariableInstance]
+        variableInstances += variab
+        (vrd.globalQName, variableInstances)
     }: _*))
 
   override def toString(): String = {
@@ -283,8 +299,18 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
    * VariableInstances are mutable and cannot safely be shared across threads
    */
   def copy(): VariableMap = {
-    val table = vTable.map { case (k: GlobalQName, abuf: ArrayBuffer[VariableInstance]) => {
-      val newBuf = abuf.map { _.copy() }
+    val table = vTable.map { case (k: GlobalQName, variableInstances: ArrayBuffer[VariableInstance]) => {
+      val newBuf = variableInstances.map { _.copy() }
+      (k, newBuf)
+    }}
+
+    new VariableMap(table)
+  }
+
+  def topLevelInstances(): VariableMap = {
+    val table = vTable.map { case (k: GlobalQName, variableInstances: ArrayBuffer[VariableInstance]) => {
+      val newBuf = new ArrayBuffer[VariableInstance]()
+      newBuf.append(variableInstances.last)
       (k, newBuf)
     }}
 
@@ -297,7 +323,7 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
   // must also ensure that the expressions only reference other variables that
   // have default value expressions or are defined externally.
   def forceExpressionEvaluations(state: ParseOrUnparseState): Unit = {
-    vTable.foreach { case (_, abuf) => { abuf.foreach { inst => {
+    vTable.foreach { case (_, variableInstances) => { variableInstances.foreach { inst => {
       (inst.state, inst.value, inst.defaultValueExpr.isDefined) match {
         // Evaluate defineVariable statements with non-constant default value expressions
         case (VariableDefined, DataValue.NoValue, true) => {
@@ -342,16 +368,24 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
    * Returns the value of a variable and sets the state of the variable to be
    * VariableRead.
    */
-  def readVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE, maybeState: Maybe[ParseOrUnparseState]): DataValuePrimitive = {
+  def readVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE, pstate: ParseOrUnparseState): DataValuePrimitive = {
     val varQName = vrd.globalQName
-    val abuf = vTable.get(varQName)
-    if (abuf.isDefined) {
-      val variable = abuf.get.last
+    vrd.direction match {
+      case VariableDirection.ParseOnly if (!pstate.isInstanceOf[PState]) =>
+        pstate.SDE("Attempting to read variable %s which is marked as parseOnly during unparsing".format(varQName))
+      case VariableDirection.UnparseOnly if (!pstate.isInstanceOf[UState]) =>
+        pstate.SDE("Attempting to read variable %s which is marked as unparseOnly during parsing".format(varQName))
+      case _ => // Do nothing
+    }
+
+    val variableInstances = vTable.get(varQName)
+    if (variableInstances.isDefined) {
+      val variable = variableInstances.get.last
       variable.state match {
         case VariableRead if (variable.value.isDefined) => variable.value.getNonNullable
         case VariableDefined | VariableSet if (variable.value.isDefined) => {
-          if (maybeState.isDefined && maybeState.get.isInstanceOf[PState])
-            maybeState.get.asInstanceOf[PState].markVariableRead(vrd)
+          if (pstate.isInstanceOf[PState])
+            pstate.asInstanceOf[PState].markVariableRead(vrd)
 
           variable.setState(VariableRead)
           variable.value.getNonNullable
@@ -362,10 +396,8 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         // forceExpressionEvaluations function after which all variables should
         // have a defined value
         case VariableDefined if (!variable.value.isDefined && variable.defaultValueExpr.isDefined) => {
-          Assert.invariant(maybeState.isDefined)
-          variable.setState(VariableInProcess)
-          val state = maybeState.get
-          val res = DataValue.unsafeFromAnyRef(variable.defaultValueExpr.get.evaluate(state))
+          variable.setState(VariableBeingDefined)
+          val res = DataValue.unsafeFromAnyRef(variable.defaultValueExpr.get.evaluate(pstate))
 
           // Need to update the variable's value with the result of the
           // expression
@@ -374,7 +406,8 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
 
           res
         }
-        case VariableInProcess => throw new VariableCircularDefinition(varQName, vrd)
+        case VariableBeingDefined => throw new VariableCircularDefinition(varQName, vrd)
+        case VariableInProcess => throw new VariableSuspended(varQName, vrd)
         case _ => throw new VariableHasNoValue(varQName, vrd)
       }
     } else
@@ -386,9 +419,9 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
    */
   def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: ThrowsSDE, pstate: ParseOrUnparseState) = {
     val varQName = vrd.globalQName
-    val abuf = vTable.get(varQName)
-    if (abuf.isDefined) {
-      val variable = abuf.get.last
+    val variableInstances = vTable.get(varQName)
+    if (variableInstances.isDefined) {
+      val variable = variableInstances.get.last
       variable.state match {
         case VariableSet => {
           referringContext.SDE("Cannot set variable %s twice. State was: %s. Existing value: %s",
@@ -407,6 +440,21 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         }
 
         case _ => {
+          vrd.direction match {
+            /**
+             * Due to potential race conditions regarding the setting of
+             * variables via setVariable and default values in cominbation with
+             * suspesions during unparsing, we only allow the use of either
+             * setVariable statements or a default value when unparsing a
+             * variable.
+             */
+            case VariableDirection.UnparseOnly | VariableDirection.Both if (vrd.maybeDefaultValueExpr.isDefined && variableInstances.get.size > 1) => {
+              // Variable has an unparse direction, a default value, and a
+              // newVariableInstance
+              pstate.SDE("Variable %s has an unparse direction and a default value, setting the variable may cause race conditions when combined with a forward referencing expression.", varQName)
+            }
+            case _ => // Do nothing
+          }
           variable.setValue(VariableUtils.convert(newValue.getAnyRef.toString, variable.rd, referringContext))
           variable.setState(VariableSet)
         }
@@ -415,30 +463,38 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
   }
 
   /**
-   * Creates a new instance of a variable
+   * Creates a new instance of a variable with default value
    */
-  def newVariableInstance(vrd: VariableRuntimeData, state: ParseOrUnparseState) = {
+  def newVariableInstance(vrd: VariableRuntimeData, defaultValue: DataValuePrimitive) = {
     val varQName = vrd.globalQName
-    val abuf = vTable.get(varQName)
-    Assert.invariant(abuf.isDefined)
+    val variableInstances = vTable.get(varQName)
+    Assert.invariant(variableInstances.isDefined)
 
-    if (vrd.maybeDefaultValueExpr.isDefined) {
-      val defaultValue = DataValue.unsafeFromAnyRef(vrd.maybeDefaultValueExpr.get.evaluate(state))
-      abuf.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, vrd))
-    } else
-      abuf.get += vrd.createVariableInstance()
+    variableInstances.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, vrd))
+  }
+
+  /**
+   * Creates a new instance of a variable without default value
+   */
+  def newVariableInstance(vrd: VariableRuntimeData) = {
+    val varQName = vrd.globalQName
+    val variableInstances = vTable.get(varQName)
+    Assert.invariant(variableInstances.isDefined)
+
+    val v = vrd.createVariableInstance()
+    variableInstances.get += v
   }
 
   def removeVariableInstance(vrd: VariableRuntimeData): Unit = {
     val varQName = vrd.globalQName
-    val abuf = vTable.get(varQName)
-    Assert.invariant(abuf.isDefined)
-    abuf.get.trimEnd(1)
+    val variableInstances = vTable.get(varQName)
+    Assert.invariant(variableInstances.isDefined && variableInstances.get.nonEmpty)
+    variableInstances.get.trimEnd(1)
   }
 
 
   private lazy val externalVarGlobalQNames: Seq[GlobalQName] =
-    vTable.map { case (_, abuf) if (abuf.last.rd.external) => abuf.last.rd.globalQName }.toSeq
+    vTable.map { case (_, variableInstances) if (variableInstances.last.rd.external) => variableInstances.last.rd.globalQName }.toSeq
 
   /**
    * Assigns an external variable and sets the variables state to VariableSet
@@ -465,11 +521,11 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         bindingQName
       }
 
-    val abuf = vTable.get(varQName.toGlobalQName)
-    if (!abuf.isDefined)
+    val variableInstances = vTable.get(varQName.toGlobalQName)
+    if (!variableInstances.isDefined)
       referringContext.schemaDefinitionError("unknown variable %s", varQName)
     else {
-      val variable = abuf.get.last
+      val variable = variableInstances.get.last
       variable.state match {
         case VariableDefined if (!variable.rd.external) => {
           referringContext.SDE("Cannot set variable %s externally. State was: %s. Existing value: %s.",
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
@@ -166,6 +166,11 @@ class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
 
   def parse(start: PState): Unit = {
     start.newVariableInstance(context)
+    if (context.maybeDefaultValueExpr.isDefined) {
+      val v = start.variableMap.find(context.globalQName).get
+      val res = DataValue.unsafeFromAnyRef(context.maybeDefaultValueExpr.get.evaluate(start))
+      v.setDefaultValue(res)
+    }
   }
 }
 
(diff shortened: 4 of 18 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +76,52 @@ final class SetVariableUnparser(
...
+class NewVariableInstanceEndUnparser(override val context: VariableRuntimeData)
   extends PrimUnparserNoData {
 
   override lazy val runtimeDependencies = Vector()
 
   override lazy val childProcessors = Vector()
 
   override def unparse(state: UState) = {
-    state.removeVariableInstance(context.asInstanceOf[VariableRuntimeData])
+    if (state.variableMap.find(context.globalQName).isDefined)
```
stevedlawrence, line 123: In what case coudl the variable not exist?

Is what's going on here is the newVariableInstance suspends and so hasn't been created yet. We continue unparsing but because that instance wasn't created so we have nothing to create so nothing to delete?

But, say we had two nested new variable instances. Say the outer one doesn't suspend and the inner one does. In that case, the outer NewVariableInstanceStartUnparser adds a new variable instance just fine. Then we call NewVariableInstanceStartUnparser for the inner variable instance. That inner NVI suspends and so doesn't create the variable isntance until much later. Unparse will continue and we'll call NewVariableInstanceEndUnparser for the inner one. Note that we still haven't created the outer new variable instance. But we do have a new variable instance from the outter NVI. So this find() will work, and we'll reomve a variable instance. But we'll remove the outer NVI, which this should only remove the inner NVI (which doesn't exist yet). So now variable access will likely fail because we've removed our instance.

Does that all sound like what would happen, or am I missing something subtle? If that is right, then it probably leads to my previous comment that we do need to always create an NVI without an value, and then suspensions can modify those values.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -312,7 +314,7 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
   def find(qName: GlobalQName): Option[VariableInstance] = {
     val optBuf = vTable.get(qName)
     val variab = {
-      if (optBuf.isDefined)
+      if (optBuf.isDefined && !optBuf.get.isEmpty)
```
stevedlawrence, line 317: This feels like it shouldn't be needed and is maybe a sign that we're doing somethign wrong?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -344,6 +346,16 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
    */
   def readVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE, maybeState: Maybe[ParseOrUnparseState]): DataValuePrimitive = {
     val varQName = vrd.globalQName
+    if (maybeState.isDefined) {
+      vrd.direction match {
+        case VariableDirection.ParseOnly if (!maybeState.get.isInstanceOf[PState]) =>
+          maybeState.get.SDW("Attempting to read variable %s which is marked as parseOnly during unparsing".format(varQName))
+        case VariableDirection.UnparseOnly if (!maybeState.get.isInstanceOf[UState]) =>
+          maybeState.get.SDW("Attempting to read variable %s which is marked as unparseOnly during parsing".format(varQName))
+        case _ => // Do nothing
```
stevedlawrence, line 355: Should these be SDEs?

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -2078,4 +2080,268 @@
...
+                </xs:annotation>
+                <xs:element name="byte4" type="xs:unsignedByte" dfdl:outputValueCalc="{ xs:unsignedByte($ex:remainingAddr) }" />
+              </xs:sequence>
+            </xs:sequence>
+          </xs:sequence>
+          <xs:element name="IPsrc" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="7"
+            dfdl:inputValueCalc="{ fn:concat(../ex:byte1, '.', ../ex:byte2, '.', ../ex:byte3, '.', ../ex:byte4) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
```
stevedlawrence, line 2127: I think it might be worth adding  anew test cases. SImilar to this where we have nested NVI's, but where we access the variable after the each instance goes out of scope. That way we can ensure that suspensions are happening correctly and that we are properly removing NVI's from scopes correctly.

stevedlawrence (commented):
Comments about how we create/remove NVI's, I'm not positive but if feels like there might be cases where the way this suspends NVI's might cause ordering issues.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +76,52 @@ final class SetVariableUnparser(
...
+  override val rd: VariableRuntimeData)
+  extends SuspendableExpression {
+
+  override protected def processExpressionResult(ustate: UState, v: DataValuePrimitive): Unit = {
+    ustate.variableMap.newVariableInstance(rd, v)
+  }
+
+  override protected def maybeKnownLengthInBits(ustate: UState) = MaybeULong(0)
+}
+
```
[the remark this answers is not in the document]
jadams-tresys: Hm, I think you might be right.  What probably needs to be done is to add an additional variable state, VariableSuspended, which is then changed once the expression evaluates.  Attempting to read or set a variable that is suspended would then also need to trigger a suspension.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: This is the one thing I'm not sure about.  It's an easy change to make, but given that there is no spec for this feature I wasn't sure what we would want to happen here.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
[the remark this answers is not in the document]
stevedlawrence: One thing to that jumped to my mind, does a setVariable even need to suspend if the defaultValue is suspending (i.e. the vairables is in the new VariableSuspended state)? In this case, we'll never use the defaultValue, so why wait for it to become evaluated if setVariable doesn't need to suspend?

I think the answer here is the defaultValue could access other variables which changes the state of those variables to read, so it probably is important to wait for the defaultValue to evaluate, even if it's value will be overwriten by a setVariable. 

Another unrelated thought about how this all needs to work, say we had something like this:
* NVI variable1, defaultValue = 1
* NVI variable2, defaultValue = "{ /some/future/path/that/requires/suspension + $variable1 }"
* setVariable, variable1 = 2
* element A dfdl:inputValueCalc="{ $variable1 }"

So we create NVI variable1 with an immediate value of defaultValue1. All good.

Then we create a NVI variable2, but the defaultValue must suspend because it accesses some future path. In suspending, it clones the UState (including the current "1" value of variable1). Note that it hasn't read variable1 until the path comes into existance.

We continue the unparse and evaluate the setVariable and change variable1 to a value of 2. This is fine because variable1 hasn't been read yet, because NVI variable2 defaultValue suspended.

Continuing the parse, we evaluate the inputValueCalc and set element A to variable1, which now has a value of 2.

At some point we unsuspend the defaultValue, and set defalutValue to 1. I *think* this works because this suspension is working on a cloned UState. But this uses the 1 value from variable1 because it was cloned. This feels wrong?

And we're done, and it's not clear what the right behavior was.

Note that if variable2 didn' need to suspend (say that long path existed), then variable1 is immedately read when setting variable2's defaultValue, and so the setVariable for variable1 should be an SDE.

So it seems we could get two different behaviors depending if defaultValue suspends or not, and it's not immediately clear to me what the right behavior is. Hopefully I'm missing something?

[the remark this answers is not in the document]
mbeckerle: The right behavior  is that suppose a DFDL implementation did not do streaming unparsing at all. There would be no-suspending when reaching forward for element values. 

The behavior one gets in that situation is the correct answer.  Streaming unparsing has to achieve this same behavior, or it is a bug. 

The correct behavior in the above example in steve lawrence's comment  is a runtime SDE due to setVariable variable1 occuring after the variable1 has already been read for its default value. 
 
In Daffodil's streaming implementation, the forward referencing expression causing a suspension should make a copy/clone of all variable state, so such time as it does read variable1, it will get the default value of 1. 

So far so good.

As unparsing continues, the setVariable statement will be encountered, and it will set variable1's value to 2.  It does not know that the expression has suspended but will eventually read variable1.  This is the bug. 

A new state on variable1 indicating it is pending a read by a suspension won't help. This doesn't work in general, because expressions can have if-then-else logic such that some conditional branches evaluate certain variables, not others. So until we evaluate an expression we can't know what variables it would read. 

I believe we have a situation close to a classic compiler anti-dependence here. 

Fact is, this setVariable should be itself suspended until after the variable-read in the expression for variable2. 

The current daffodil expression implementation won't even attempt to read variable1's value until the suspension completes; hence, variable1 wouldn't even get marked as 'pending read' state until the forward reference has been resolved.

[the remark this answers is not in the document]
mbeckerle: Doing the full dependency analysis of every read/write of a variable across the DFDL schema is one way to solve this, albeit a complex one.

I think a "workaround" fix might be this: in variable direction both or unparseOnly, disallow setVariable after a NVI with a default value. This requires the user to either remove the default value from the NVI, or remove the setVariable. This requires analysis of whether there are setVariable statements in scope of an unparse-direction NVI having default value, and that's itself somewhat tricky, but it is far easier than a full dependency analysis.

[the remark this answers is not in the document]
mbeckerle: An easier check could be if a given variable is used anywhere in the schema by a NVI statement having default value, and the variable has unparse-direction, that no setVariable is allowed anywhere for that variable. This eliminates the need for a precise roll-up of which setVariable statements are in scope of a particular NVI or not. 

Such a variable is what we might call "bindable only" or non-settable. I think this fixes many things, and DFDL schemas that actually try to use setVariable with such variables can split a variable into two variables. Use one for binding via NVI with defaults, and the other for the setVariable purposes, it's NVIs would not have any default values.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
mbeckerle: Definitely make them SDE. We can always weaken them to SDW in the future if we find SDE too restrictive. But I think SDE is right.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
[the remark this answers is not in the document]
mbeckerle: As this is a dfdlx experimental feature, I think we need a wiki page description of it, and that would have a section detailing these implementation-specific restrictions on how to use variables at unparse time due to the streaming unparser behavior.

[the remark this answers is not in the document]
mbeckerle: Lastly this scenario SL came up with needs a test that either shows the problem, or shows it not occurring.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
mbeckerle: I think SL is correct. This isn't right. We can't suspend creation of the variable instance, we have to suspend the calculation of its value sometimes, and we can't read it until that calculation is complete, but the instance has to be there so that the expressions can try to read and determine that they've blocked on a pending variable value.

mbeckerle (changes requested):
Need more tests to characterize some of the flaws we suspect the implementation will have. Whether we fix those issues this change set or not, I don't know. It may make sense to fix it this much, and open tickets and fix those other issues later.

I think there is one serious algorithmic issue related to when NVI instances are created relative to suspensions.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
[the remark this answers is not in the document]
stevedlawrence: Thinking more on this, I think there's maybe some other complexities related to variables, and doesn't even include NVI.

Imagine we having a variable with a constant static default, and then a setVariable to set that variable but that must suspend due to a forward lookahead. Now if an an element tries to read that variable, it must also suspend until that setVariable completes. Something to consider with this is that setVariable suspension is working on a cloned ustate. So the cloned ustate must do a shallow copy of the variable map so that when it does modify the variable map, other suspensions will see that change. I think this may be how it works, but worth having some tests to consider.

But also note that since this is a shallow copy, it means other suspensions/parsers could modify this variable map, and could maybe potentially affect the suspension once it unsuspends? Maybe something like this:

* variable1 defaultValue = 1
* variable2 defaultValue = 2
* setVariable variable1, value = /forward/path/forcing/suspension + variable2
* setVariable variable2, value = 3

So setVariable variable1 must suspend. Then setVariable varible2 changes the value of variable2 to 3. Then at some point when setVariable variable1 unsuspends, it gets the value of 3 for variable2 (assuming shallow variable map copy in the ustate clone). The correct result is setVariable2 should cause an SDE because setVariable variable1 should have read variable2. So this too has this order dependency stuff going on and doesn't involve NVI. I think this is basically the same thing as my previous scenario, but doesn't use NVI.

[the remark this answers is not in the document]
mbeckerle: This may ultimately just be a flaw in the DFDL language notion of variables. They're supposed to be single-assignment variables, but this mixture of defaultValue and setVariable where a read locks down the default value was always a sketchy design point because things become very order dependent, and since the unparser does things out of order due to suspensions and forward reference, the tenuous interactions of these things become problematic. 

A way to fix this is to say that for any variable that has unparse direction (so both or unparseOnly) you have to pick whether you use *only* defaultValue expressions to bind values to it, or whether you will use setVariable statements, in which case the variable cannot have any default value in the base defineVariable nor in any newVariableInstance statements.  This I think fixes all these issues. A variable is either defaultable, and not settable, or it is settable and not defaultable.  I think these restrictions are very livable.

jadams-tresys:
Got things working with the one scoping issue I was running into.  Still need to add a test for the newVariableInstance/setVar scenario that Steve was describing.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +78,83 @@ final class SetVariableUnparser(
...
+    if (context.maybeDefaultValueExpr.isDefined) {
+      val maybeVar = state.variableMap.find(context.globalQName)
+      Assert.invariant(maybeVar.isDefined)
+
+      // Set the variable's state to VariableInProcess while waiting for the NVI
+      // expression to complete. When the expression completes, this original
+      // variable instance will have its state reset to its original value.
+      maybeVar.get.setState(VariableInProcess)
+      val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
+      suspendableExpression.run(state)
```
stevedlawrence, line 116: I'm not sure if this is right. The NewVariableInstanceSuspendableExpression doesn't create the actual newVariableInstance until the deafult expression successfully evaluates. Isn't it possible that this default requires a suspension, then we continue and then an expression tries to reference that new instance. But since it's not created yet it's still trying to access the old instance?

Seems like this should be something like
```scala
state.variableMap.newVariableInstance(context)
val maybeVar = state.variableMap.find(...) // get the new instsance
maybeVar.get.setState(VariableInProcess)
```
So we create a new instance, and then set its state so that anything referencing this new instance needs to wait until the default is evaluated. Then the NVISuspendableExpression processExpressionResult function sets the value of this new instance and sets the value to VariableSet. Or something along those lines?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +78,83 @@ final class SetVariableUnparser(
...
 }
 
-class NewVariableInstanceEndUnparser(override val context: RuntimeData)
-  extends PrimUnparserNoData {
+final class NewVariableInstanceSuspendableOperation(override val rd: VariableRuntimeData)
+  extends SuspendableOperation {
+
+    override val isReadOnly = true
+
+    override def toString = "Suspension for removal of variable instance: " + rd.globalQName
```
stevedlawrence, line 127: I'd prefer we just stick with the default toString for suspendable operations so that the strings are consistent. I'm working on a debugger patch to output suspension information and this string is going to stand out as very different.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +78,83 @@ final class SetVariableUnparser(
...
+    override def test(ustate: UState): Boolean = {
+      // If there is no default value expression, simply return true to
+      // immediately remove the variable instance
+      if (!rd.maybeDefaultValueExpr.isDefined)
+        true
+      else {
+        val v = ustate.variableMap.find(rd.globalQName)
+        Assert.invariant(v.isDefined)
+        v.get.state != VariableInProcess
+      }
```
stevedlawrence, line 138: I think if you make the suggestion above, then you don't need this suspension. NewVarialbeInstanceStartUnparser will always create a newVariableInstance, and so NewVariableInstanceEndUnparser can always remove that new instance. The only thing we need to make sure is that we correctly clone UState for suspensions that occurr inbetween this Start and End so that they have access to this NVI and that they can mutate it so that other suspensions can see the mutation.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,30 +78,83 @@ final class SetVariableUnparser(
...
+  with SuspendableUnparserNoData {
 
   override lazy val runtimeDependencies = Vector()
 
   override lazy val childProcessors = Vector()
 
-  override def unparse(state: UState) = {
-    state.removeVariableInstance(context.asInstanceOf[VariableRuntimeData])
+  override def suspendableOperation = {
+    new NewVariableInstanceSuspendableOperation(context)
```
stevedlawrence, line 157: As mentioned above, I think what you had before is correct. I think this Unparser unconditinoally remove the NVI, without any thought about suspensions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1711,6 +1711,7 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
             case VariableRead => "read"
             case VariableSet => "set"
             case VariableUndefined => "undefined"
+            case VariableBeingDefined => "being defined"
```
stevedlawrence, line 1714: How is VariableBeingDefined different from VariableInProcess? My reading of the scaladoc for VariableInProcess sounds alot like what we need. When a NVI is suspended, what if we just put that instance into VariableInProcess? This says that any read should suspend. And If we implement Mike's suggestion, set's aren't even allowed if there is a default value. So basically VariableInProcess means this variable is being set by either a defaultValue or setVariable (but only via one of them) and that suspension hasn't finished yet. Once that's finished, reads can unsuspend.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -42,6 +44,11 @@ case object VariableUndefined extends VariableState
 
 case object VariableDefined extends VariableState
 
+/**
+ * Used to detect circular definitions of variables
+ */
```
stevedlawrence, line 49: Need a more thurough description of this. Though, I'm wondering if we can get away with VariableInProcess and don't need a new state?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -363,7 +386,7 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         // have a defined value
         case VariableDefined if (!variable.value.isDefined && variable.defaultValueExpr.isDefined) => {
           Assert.invariant(maybeState.isDefined)
-          variable.setState(VariableInProcess)
+          variable.setState(VariableBeingDefined)
```
stevedlawrence, line 389: This makes me thing that this two states are actually basically the same thing? If not, the distinction isn't very clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -415,24 +439,32 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
     val abuf = vTable.get(varQName)
     Assert.invariant(abuf.isDefined)
 
-    if (vrd.maybeDefaultValueExpr.isDefined) {
-      val defaultValue = DataValue.unsafeFromAnyRef(vrd.maybeDefaultValueExpr.get.evaluate(state))
-      abuf.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, vrd))
-    } else
-      abuf.get += vrd.createVariableInstance()
+    abuf.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, vrd))
   }
```
stevedlawrence, line 450: I wonder if this could be made more clear if the defaultValue wasn't passed into this, and add a new function that's something like setDefaultValue()? That way it's very clear when the NVI is actually created, and also how the default value is set, and what state changes when it's set? Not sure off hand, but just a suggestion.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -165,7 +165,11 @@ class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
   override lazy val runtimeDependencies = Vector()
 
   def parse(start: PState): Unit = {
-    start.newVariableInstance(context)
+    if (context.maybeDefaultValueExpr.isDefined) {
+      val res = context.maybeDefaultValueExpr.get.evaluate(start)
+      start.newVariableInstance(context, DataValue.unsafeFromAnyRef(res))
+    } else
+      start.newVariableInstance(context)
```
stevedlawrence, line 172: Benefit of above suggestion about splitting up NVI and default value, this becomes something like
```scala
start.newVariableInstance(context)
if (context.maybeDefaultVlueExpr.isDefined) {
  val res = context.maybeDefaultValueExpr.get.evaluate(start)
￼  start.setVariableDefault(...)
}
```
Benefit here is it decouples creating a new variable instance and setting the default. Granted the two are almost always going to be coupled, but it makes it easier to set what state setting a default value changes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -343,14 +342,6 @@ abstract class UState(
 
   def documentElement: DIDocument
 
-  def newVariableInstance(vrd: VariableRuntimeData): Unit = {
-    variableMap.newVariableInstance(vrd, this)
-  }
-
-  def removeVariableInstance(vrd: VariableRuntimeData): Unit = {
-    variableMap.removeVariableInstance(vrd)
-  }
```
stevedlawrence, line 352: I was expecting to a see a change in this file when we clone the ustate for suspensions. I think based on a discussion on the mailing list, the NVIUnparser should always create a newVariableInstance and then create a suspension if a defautlValue is needed to be set (since that could suspend).

Then if anything else suspends (like reading this NVI default value), then when clongin the ustate for suspension, we need to create a shallow copy of the top of the variable map. This way the suspensions created while this NVI was in scope have pointer to that instance. But because it's a shallow copy if one of those suspensions mutates the state (i.e. marking the variable as read) other suspension will see this new state.

stevedlawrence (changes requested):

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
jadams-tresys: So I had been trying to get the approach you described to work for quite a while, but kept running into SuspensionDeadlock issues.  Perhaps I wasn't copying things appropriately when the UState gets cloned for suspension.  I'll take another look at it, but this commit was the first thing that I could get to work and handle scoping properly.

mbeckerle (commented):
Is this change implementing the restrictions previously described, where an unparse-direction (so unparseOnly or both) is either settable, or default-value bindable, but not a mixture of the two?

Without that I think with streaming unparsing we are going to have unresolvable race-conditions between the read of a default value, and a read of the same variable after a setVariable statement.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -374,7 +397,8 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
 
           res
         }
-        case VariableInProcess => throw new VariableCircularDefinition(varQName, vrd)
+        case VariableBeingDefined => throw new VariableCircularDefinition(varQName, vrd)
+        case VariableInProcess => throw new VariableSuspended(varQName, vrd)
```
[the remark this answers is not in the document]
jadams-tresys: That is correct. VariableBeingDefined is used when a defineVariable statement references another variable in its default value. Its state gets set to VBD while we are resolving the variable reference and if we hit another variable that is set to VBD it means there is a circular definition.

VariableInProcess is for variables that are suspended but should complete once the suspension is done.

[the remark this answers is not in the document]
stevedlawrence: Ah, ok, this makes sense. One thing that is maybe unclear to me then. Above we have this:
```scala
  case VariableDefined if (!variable.value.isDefined && variable.defaultValueExpr.isDefined) => {
    Assert.invariant(maybeState.isDefined)
    variable.setState(VariableBeingDefined)
```
So it seems like a variable goes from VariableDefined to VariableBeingDefined when it's read but has no value (and has a defaultValue). My intuition would say that these pre-runtime variables should have started in the state of VariableBeingDefined if they don't have a value and then switch to VariableDefined when they actually do get a value. Since this doesn't appear to be the case, it would probably be really helpful to add a comment explaining why this state change happens.

jadams-tresys:
I reverted back to the original style of immediate creation/removal of NVI statements and only suspending when attempting to evaluate the expression.

I added some commented out code for only allowing setVariable to be used during unparsing on variables without a default value.  This code causes several existing test cases to break (haven't checked against the various schema projects yet), so I wasn't sure if we wanted to go through with this change.  Maybe a warning about race conditions/potential undefined behavior would be a better approach?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -407,32 +446,54 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
-          variable.setState(VariableSet)
+          vrd.direction match {
+            /**
+             * Due to potential race conditions regarding the setting of
+             * variables via setVariable and default values in cominbation with
+             * suspesions during unparsing, we only allow the use of either
+             * setVariable statements or a default value when unparsing a
+             * variable.
+             */
+            //case VariableDirection.UnparseOnly | VariableDirection.Both if (vrd.maybeDefaultValueExpr.isDefined) => {
```
mbeckerle, line 457: So I think we can enable this check, but we can make it check for something narrower that will not cause existing tests to fail

If a variable has no NVIs associated with it. Then it is just a global variable, and I think this SDW/SDE is not needed at all. 

If a variable does have NVI statements, let's call that a scoped variable. If any of those NVI statements have default values, or the global declaration has a default value, then I think this warning is important if the variable direction is "both", and should be an SDE if the direction is "unparse". 

This requires us to compute whether a variable is global, or scoped with any default values supplied, if direction is both or parsed.

mbeckerle (changes requested):
I think we should try the narrower SDW/SDE approach on the variables only that have NVI statements for them as suggested in my specific comment.

jadams-tresys:
Made a small change to enable the check and only when it is a newVariableInstance.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -407,6 +446,21 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
         case _ => {
+          vrd.direction match {
+            /**
+             * Due to potential race conditions regarding the setting of
+             * variables via setVariable and default values in cominbation with
+             * suspesions during unparsing, we only allow the use of either
+             * setVariable statements or a default value when unparsing a
+             * variable.
+             */
+            case VariableDirection.UnparseOnly | VariableDirection.Both if (vrd.maybeDefaultValueExpr.isDefined && abuf.get.size > 1) => {
```
mbeckerle, line 457: abuf is the structure that accumulates the related NVIs for this variable? 

This is fine, I'm just wondering if the name "abuf" can be improved. 

How many of these warnings do we get anyway? (None I would guess, except from tests designed to cause it?

If so then I'd suggest making this an SDE. That's a more conservative design point because we can always weaken it to a warning, but once people start depending on schemas working where it's a warning we would break their schema by changing it to an SDE.

mbeckerle (approved):
+1  With minor change SDW to SDE, and improve the name "abuf".

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
```diff
@@ -60,6 +60,17 @@ class TestVariables {
...
+
+  @Test def test_varDirection_1(): Unit = { runner.runOneTest("varDirection_1") }
+  @Test def test_varDirection_2(): Unit = { runner.runOneTest("varDirection_2") }
+  @Test def test_varDirection_3(): Unit = { runner.runOneTest("varDirection_3") }
+  @Test def test_varDirection_err1(): Unit = { runner.runOneTest("varDirection_err1") }
+  @Test def test_varDirection_err2(): Unit = { runner.runOneTest("varDirection_err2") }
+  @Test def test_varDirection_setVar1(): Unit = { runner.runOneTest("varDirection_setVar1") }
+  @Test def test_varDirection_setVar2(): Unit = { runner.runOneTest("varDirection_setVar2") }
+  @Test def test_varDirection_nvi1(): Unit = { runner.runOneTest("varDirection_nvi1") }
+  @Test def test_varDirection_nvi2(): Unit = { runner.runOneTest("varDirection_nvi2") }
```
stevedlawrence, line 74: Was playing around with this NVI stuff and testing various variable things. Was testing this language in the spec:
> If the instance is not assigned a new default value then it inherits the default value specified by dfdl:defineVariable or externally provided by the DFDL processor.
I don't think we support the "or externaly provided" part of this. Here's the example schema I used:

```xml
  <xs:annotation>
      <dfdl:format ref="ex:GeneralFormat" />
      <dfdl:defineVariable name="var1" type="xs:int" defaultValue="1" external="true" />
    </xs:appinfo>
  </xs:annotation>

  <xs:element name="root">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="beforeNVI" type="xs:int" dfdl:inputValueCalc="{ $ex:var1 }" />
        <xs:sequence>
          <xs:annotation>
              <dfdl:newVariableInstance ref="ex:var1" />
            </xs:appinfo>
          </xs:annotation>
          <xs:element name="afterNVI" type="xs:int" dfdl:inputValueCalc="{ $ex:var1 }" />
        </xs:sequence>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
```
If I run the following:
```
echo -n | daffodil parse -Dvar1=2 -s src/main/resources/com/tresys/csv/xsd/csv.dfdl.xsd
```
Then I get the infoset:
```xml
<root>
  <beforeNVI>2</beforeNVI>
  <afterNVI>1</afterNVI>
</root>
```
So the newVariableInstance is using the defalut value from the defineVariable, but my reading of that line in the spec says the default value for the NVI should be the value that was externally provided, so both beforeNVI and afterNVI should be 2.

Is my reading correct? If so, I don't think that needs to be fixed as part of this, but we should open a bug.

stevedlawrence (approved):
+1 with a few minor comments.

I personally find all this variable stuff hard to follow and reason about its correctness. That's not because of your changes, but just because of the complexity of variables with all different states/directions/scopes/defaultValues/etc. I'm not sure if there's a way to make it all more clear but, at the very least I think it would be helpful to add more comments explaining what's doing what and more importantly, why.

daffodil-test/src/test/scala/org/apache/daffodil/section07/variables/TestVariables.scala
mbeckerle: Definitely a bug. Externally set variables that have a default value, the default value is 100% discarded and the externally provided one is used for all cases.

stevedlawrence: Ok, I've created [DAFFODIL-2481](https://issues.apache.org/jira/browse/DAFFODIL-2481) for that issue. I don't think it needs to be fixed as part of the PR, which is only about the new direction stuff.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -77,31 +78,52 @@ final class SetVariableUnparser(
...
 
+final class NewVariableInstanceSuspendableExpression(
+  override val expr: CompiledExpression[AnyRef],
+  override val rd: VariableRuntimeData)
+  extends SuspendableExpression {
+
+  override protected def processExpressionResult(ustate: UState, v: DataValuePrimitive): Unit = {
+    val vi = ustate.variableMap.find(rd.globalQName)
+    Assert.invariant(vi.isDefined)
+    // Reset original variable's state from VariableInProcess
```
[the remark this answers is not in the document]
jadams-tresys: This really should always just be VariableDefined and setting to the priorState was a round about way of doing just that. Making the change now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -291,6 +305,18 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
     new VariableMap(table)
   }
 
+  def topLevelInstances(): VariableMap = {
+    val table = vTable.map { case (k: GlobalQName, abuf: ArrayBuffer[VariableInstance]) => {
+      val newBuf = new ArrayBuffer[VariableInstance]()
+      if (abuf.length > 1)
+        newBuf.append(abuf(abuf.length - 2))
+      newBuf.append(abuf.last)
+      (k, newBuf)
```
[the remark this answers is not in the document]
jadams-tresys: Ah, this was something that should have been removed.  It was necessary to support the self-referencing NVI statements, but since we aren't allowing those any more it ls no longer needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -344,11 +370,23 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
    */
   def readVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE, maybeState: Maybe[ParseOrUnparseState]): DataValuePrimitive = {
     val varQName = vrd.globalQName
+    if (maybeState.isDefined) {
```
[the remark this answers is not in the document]
jadams-tresys: So this actually seems like it might be an artifact of dstate.parseOrUnparseState returning a maybe. For the time being I've modified the call Dstate makes when reading a variable to do an Assert.invariant(dstate.parseOrUnparserState.isDefined)

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: We had one existing case that caused the warning.  I'm going to create an additional test case specifically to trigger this now that it will be an SDE as the existing test case does provide important scoping testing for newVariableInstances.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
jadams-tresys: After spending too much time trying to figure out how to make a fake PState (couldn't find any existing code and PState seems to require a lot of scaffolding to get set up) I decided to modify the test slightly to simply call vmap.find() and check the value of the returned variable instance, which doesn't require a PState.

PR #477, opened 2021-01-21
Removed test looking for compile time check.

mbeckerle:
A run time check for negative indexes is sufficient.

DAFFODIL-1865


stevedlawrence:
Looks like there was a test failure unrelated to the change.I've created ticket [DAFFODIL-2461](https://issues.apache.org/jira/browse/DAFFODIL-2461) to investigate that.

Looks fine otherwise, +1

stevedlawrence (approved):

jw3:
Ill take a look at that failure.

tuxji (approved):
+1

Yeah, humans are unlikely to write a literal -1 or negative expression in an array index :).

PR #479, opened 2021-01-22
Fix schematron integration test ec parsing

jw3:
Fixes random failures parsing the error code on windows by using getBefore to get the ec string instead of getInput.

DAFFODIL-2461

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
@@ -71,7 +71,7 @@ package object schematron {
     try {
       shell.sendLine(cmd).expect(expectation)
 
-      val actualEc = shell.sendLine(echoEC).expect(matchEC).getInput.trim.split(eol).last
+      val actualEc = shell.sendLine(echoEC).expect(matchEC).getBefore.trim.split(eol).last
       Try(actualEc.toInt) match {
         case Success(`ec`) => // good
         case Success(v) => fail(s"wrong ec, $v")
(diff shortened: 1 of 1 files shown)


jw3:
It appeared that getInput would allow additional characters to be returned after the match position.

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
```diff
@@ -94,7 +94,7 @@ package object schematron {
   private val mustache2 = """\{\{(.+?)}}""".r.unanchored
 
   private val echoEC = s"echo ${if(Util.isWindows) "%errorlevel%" else "$?"}"
-  private val matchEC = regexLine("""\d+""")
+  private val matchEC = regexLine(s"""(?<=\\d+)${if(Util.isWindows) eol else "$"}""")
```
jw3, line 97: Look behind to find the so it is not consumed and is available in `getBefore`

Had to match the end of line differently, not entirely sure why `$eol` (`\n`) doesnt match on linux.

mbeckerle: Put this discussion into a comment in front of the offending line, maybe somebody will figure it out in the future. Probably CRLF vs. LF related. I think of '\n' as being LF, not CRLF. I think modern regex has \R which is \n or \r\n.

mbeckerle (approved):
+1

jw3:
> Probably CRLF vs. LF related. I think of '\n' as being LF, not CRLF. I think modern regex has \R which is \n or \r\n.

Yeah I wonder if it is the filter that is applied to the shell

`.withInputFilters(replaceInString("\r\n", "\n"))`

where the regex `$` would be platform specific `\r\n` and that filter changes the stream before it gets to the match.

Nothing to do here right now.  It can be addressed in the CLI integrations ticket.  Ill document that thought.

stevedlawrence:
+1 thanks for looking into this so quickly!

jw3:
Sure thing.  Hate to break the build, especially with intermittent things, now gun-shy of CI runs  :fearful: 

Speaking of... looks like we are at 14 minutes of generating a coverage report...

stevedlawrence:
Not to worry, this isn't the first time we've had intermitting errors cause build failures. Glad this one was at least reproducable. That's often difficult with this CI failures.

Though, looks like something is stuck in either github actions or in the coverage report. I have no idea what the issue could be, but it's almost certainly not a Daffodil issue. Unfortunately, these errors aren't all that uncommon either. I usually just cancel the workflow and re-run.

PR #481, opened 2021-01-25
ADD test for "%%" to bypass existing test for "%NL"

regetom:
Add an if test to the noCharClassEntities function to not throw an SDE when the passed string
contains "%%NL"

daffodil-774

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
@@ -63,7 +63,7 @@ object InitiatorCooker extends DelimiterCooker("initiator")
 
 object TerminatorCooker extends DelimiterCooker("terminator")
 
-object TerminatorCookerNoES extends DelimiterCookerNoES("terminator")
+object TerminatorDelimitedCooker extends DelimiterCookerNoSoleES("terminator")
 
 object SeparatorCooker extends DelimiterCookerNoES("separator")
 
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
@@ -491,7 +491,7 @@ abstract class UpperCaseToken(propNameArg: String = null)
 sealed abstract class StringLiteralBase(propNameArg: String,
   protected val allowByteEntities: Boolean)
   extends AutoPropNameBase(propNameArg)
-  with StringLiteralCookerMixin {
+    with StringLiteralCookerMixin {
 
   private val xmlEntityPattern = new Regex("""&(quot|amp|apos|lt|gt);""", "entity")
 
@@ -560,48 +560,64 @@ sealed trait SingleCharacterMixin { self: StringLiteralBase =>
 class SingleCharacterLiteral(pn: String, allowByteEntities: Boolean)
   extends StringLiteralBase(pn, allowByteEntities) with SingleCharacterMixin
 
-trait NoCharClassEntitiesMixin {
+trait DisallowedCharClassEntitiesMixin {
 
   protected def propName: String
+  protected def disallowedCharClassEntities: Seq[String]
 
   /**
-   * Override if set of prohibited class entities is different
+   * The raw string to test is supplied as the first parameter
+   * A list of disallowed character class entities is supplied as the 2nd parameter
+   * This list is added to the regex string below
+   * Ex regex: (?:^|[^%])(?:%%)*(%(?:NL|ES);)
+   * (?:^|[^%]) look for the beginning of the string or a non-% character
+   * (?:%%)* look for an even number of %'s
+   * (%(?:NL|ES);) look for any single % left after previous match along with and NL or ES followed by a ;
+   *
    */
-  protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-    context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%ES;"), "Property dfdl:%s cannot contain %%ES;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP;"), "Property dfdl:%s cannot contain %%WSP;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP+;"), "Property dfdl:%s cannot contain %%WSP+;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP*;"), "Property dfdl:%s cannot contain %%WSP*;", propName)
+  private lazy val disallowedRegex = {
+    val disallowedCharClassMatch = disallowedCharClassEntities.map(Pattern.quote).mkString("|")
+    val regexString = "(?:^|[^%])(?:%%)*(%(?:" + disallowedCharClassMatch + ");)"
+    regexString.r
   }
 
   protected def testRaw(raw: String, context: ThrowsSDE): Unit = {
-    noCharClassEntities(raw, context)
+    val matchedGroups = disallowedRegex.findAllMatchIn(raw).map {_.group(1)}
+    context.schemaDefinitionUnless(
+      matchedGroups.isEmpty,
+      "Property dfdl:%s contains disallowed character class(es): %s" ,
+      propName,
+      matchedGroups.mkString(", "))
   }
 }
 
 class StringLiteralNoCharClassEntities(pn: String, allowByteEntities: Boolean)
   extends StringLiteralBase(pn, allowByteEntities)
-  with NoCharClassEntitiesMixin {
+  with DisallowedCharClassEntitiesMixin {
 
+  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
   override def testRaw(raw: String, context: ThrowsSDE) =
-    super[NoCharClassEntitiesMixin].testRaw(raw, context)
+    super[DisallowedCharClassEntitiesMixin].testRaw(raw,context)
 }
 
 class SingleCharacterLiteralNoCharClassEntities(pn: String, allowByteEntities: Boolean)
   extends StringLiteralBase(pn, allowByteEntities)
-  with NoCharClassEntitiesMixin
+  with DisallowedCharClassEntitiesMixin
   with SingleCharacterMixin {
 
-  override protected def testRaw(raw: String, context: ThrowsSDE) = super.testRaw(raw, context)
+  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
+  override protected def testRaw(raw: String, context: ThrowsSDE) =
+    super.testRaw(raw, context)
 }
 
 class StringLiteralNonEmptyNoCharClassEntitiesNoByteEntities(pn: String = null)
   extends StringLiteralBase(pn, false)
-  with NoCharClassEntitiesMixin
+  with DisallowedCharClassEntitiesMixin
   with NonEmptyMixin {
 
-  override def testRaw(raw: String, context: ThrowsSDE) = super[NoCharClassEntitiesMixin].testRaw(raw, context)
+  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
+  override def testRaw(raw: String, context: ThrowsSDE) =
+    super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
 
   override protected def testCooked(cooked: String, context: ThrowsSDE) = super[NonEmptyMixin].testCooked(cooked, context)
 }
@@ -616,18 +632,12 @@ class SingleCharacterLiteralNoCharClassEntitiesWithByteEntities(pn: String = nul
   extends SingleCharacterLiteralNoCharClassEntities(pn, allowByteEntities = true)
 
 class StringLiteralESEntityWithByteEntities(pn: String)
-  extends StringLiteralNoCharClassEntities(pn, allowByteEntities = true) {
+  extends StringLiteralNoCharClassEntities(pn, allowByteEntities = true)
+  with DisallowedCharClassEntitiesMixin {
 
-  /**
-   * Override Because ES is allowed.
-   */
-  override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-    context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-    // ES is allowed
-    context.schemaDefinitionUnless(!raw.contains("%WSP;"), "Property dfdl:%s cannot contain %%WSP;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP+;"), "Property dfdl:%s cannot contain %%WSP+;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP*;"), "Property dfdl:%s cannot contain %%WSP*;", propName)
-  }
+  override val disallowedCharClassEntities = Seq("NL", "WSP", "WSP+", "WSP*")
+  override def testRaw(raw: String, context: ThrowsSDE) =
+    super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
 }
 
 class SingleCharacterLiteralESEntityWithByteEntities(pn: String = null)
@@ -738,9 +748,11 @@ class ListOfStringLiteralNonEmptyNoCharClassEntitiesNoByteEntities(pn: String = 
   extends ListOfStringLiteralBase(pn, false) {
 
   override protected val oneLiteralCooker =
-    new StringLiteralNonEmpty(propName, allowByteEntities = false) with NoCharClassEntitiesMixin {
+    new StringLiteralNonEmpty(propName, allowByteEntities = false) with DisallowedCharClassEntitiesMixin {
 
-      override def testRaw(raw: String, context: ThrowsSDE) = super[NoCharClassEntitiesMixin].testRaw(raw, context)
+      override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
+      override def testRaw(raw: String, context: ThrowsSDE) =
+        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
     }
 }
 
@@ -748,26 +760,22 @@ class ListOfStringLiteralNoCharClass_NL_ES_EntitiesNoByteEntities(pn: String = n
   extends ListOfStringLiteralBase(pn, false) {
 
   override protected val oneLiteralCooker =
-    new StringLiteral(propName, allowByteEntities = false) with NoCharClassEntitiesMixin {
+    new StringLiteral(propName, allowByteEntities = false) with DisallowedCharClassEntitiesMixin {
 
-      /**
-       * fewer prohibited char class entities
-       */
-      override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-        context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-        context.schemaDefinitionUnless(!raw.contains("%ES;"), "Property dfdl:%s cannot contain %%ES;", propName)
-      }
-
-      override def testRaw(raw: String, context: ThrowsSDE) = super[NoCharClassEntitiesMixin].testRaw(raw, context)
+      override val disallowedCharClassEntities = Seq("NL", "ES")
+      override def testRaw(raw: String, context: ThrowsSDE) =
+        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
     }
 }
 
 class SingleCharacterLineEndingOrCRLF_NoCharClassEntitiesNoByteEntities(pn: String = null)
-  extends StringLiteral(pn, allowByteEntities = false) with NoCharClassEntitiesMixin {
+  extends StringLiteral(pn, allowByteEntities = false) with DisallowedCharClassEntitiesMixin {
 
   private val validNLs: List[Char] = List('\u000A', '\u000D', '\u0085', '\u2028')
 
-  override def testRaw(raw: String, context: ThrowsSDE) = super[NoCharClassEntitiesMixin].testRaw(raw, context)
+  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
+  override def testRaw(raw: String, context: ThrowsSDE) =
+    super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
 
   /**
    * Check that length is 1 (single char) except for CRLF case, and that it's a line ending char.
@@ -784,18 +792,11 @@ class NonEmptyListOfStringLiteralCharClass_ES_WithByteEntities(pn: String)
   extends ListOfStringLiteralBase(pn, true) {
 
   override protected val oneLiteralCooker =
-    new StringLiteral(propName, allowByteEntities = true) with NoCharClassEntitiesMixin {
+    new StringLiteral(propName, allowByteEntities = true) with DisallowedCharClassEntitiesMixin {
 
-      override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-        context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-        context.schemaDefinitionUnless(!raw.contains("%WSP;"), "Property dfdl:%s cannot contain %%WSP;", propName)
-        context.schemaDefinitionUnless(!raw.contains("%WSP+;"), "Property dfdl:%s cannot contain %%WSP+;", propName)
-        context.schemaDefinitionUnless(!raw.contains("%WSP*;"), "Property dfdl:%s cannot contain %%WSP*;", propName)
-      }
-
-      override def testRaw(raw: String, context: ThrowsSDE) = {
-        super[NoCharClassEntitiesMixin].testRaw(raw, context)
-      }
+      override val disallowedCharClassEntities = Seq("NL", "WSP", "WSP+", "WSP*")
+      override def testRaw(raw: String, context: ThrowsSDE) =
+        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
     }
 
   override def testCooked(cookedList: List[String], context: ThrowsSDE) = {
@@ -806,16 +807,27 @@ class NonEmptyListOfStringLiteralCharClass_ES_WithByteEntities(pn: String)
 class DelimiterCookerNoES(pn: String) extends ListOfString1OrMoreLiteral(pn, true) {
 
   override val oneLiteralCooker: StringLiteralBase =
-    new StringLiteralNoCharClassEntities(propName, true) {
+    new StringLiteralNoCharClassEntities(propName, true) with DisallowedCharClassEntitiesMixin {
 
-      override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-        // TODO: this isn't quite right, as it will allow combined delimiters
-        // that still match the empty string, e.g. "%ES;%WSP*;". We could check
-        // if raw.contains("%WSP*;"), but that is too general, preventing valid
-        // delimiters like "foo%WSP*;bar". Although the below matches the
-        // specification, it's probably not the intended behavior.
-        context.schemaDefinitionUnless(raw != "%WSP*;", "Property dfdl:%s cannot contain %%WSP*; when dfdl:lengthKind=\"delimited\".", propName)
-        context.schemaDefinitionUnless(raw != "%ES;", "Property dfdl:%s cannot contain %%ES; when dfdl:lengthKind=\"delimited\".", propName)
+      // Disallow "%ES" in the string raw. Disallow "%WSP*" when it is
+      // the only value in the string raw
+      override val disallowedCharClassEntities = Seq("ES")
+      override def testRaw(raw: String, context: ThrowsSDE) = {
+        context.schemaDefinitionUnless(raw != "%WSP*;", """For dfdl:%s the WSP* entity cannot appear on it's own when dfdl:lengthKind="delimited".""", propName)
+        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
+      }
+    }
+}
+
+class DelimiterCookerNoSoleES(pn: String) extends ListOfString1OrMoreLiteral(pn, true) {
+
+  override val oneLiteralCooker: StringLiteralBase =
+    new StringLiteralBase(propName, true) {
+
+     override def testRaw(raw: String, context: ThrowsSDE): Unit = {
+        // "Neither %ES or %WSP*;" can appear on their own when lenghtKind is delimited.
+        context.schemaDefinitionUnless(raw != "%ES;", """For dfdl:%s the ES entity cannot appear on its own when dfdl:lengthKind="delimited".""", propName)
+        context.schemaDefinitionUnless(raw != "%WSP*;", """For dfdl:%s the WSP* entity cannot appear on its own when dfdl:lengthKind="delimited".""", propName)
       }
     }
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala
@@ -23,7 +23,7 @@ import org.apache.daffodil.processors.dfa.CreateDelimiterDFA
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.processors.parsers.DelimiterTextType
-import org.apache.daffodil.cookers.TerminatorCookerNoES
+import org.apache.daffodil.cookers.TerminatorDelimitedCooker
 import org.apache.daffodil.cookers.SeparatorCooker
 import org.apache.daffodil.cookers.InitiatorCooker
 import org.apache.daffodil.cookers.TerminatorCooker
@@ -117,13 +117,13 @@ class InitiatorUnparseEv(expr: CompiledExpression[String], outputNewLine: Output
 class TerminatorParseEv(expr: CompiledExpression[String], isLengthKindDelimited: Boolean, ignoreCase: Boolean, tci: DPathCompileInfo)
   extends DelimiterParseEv(DelimiterTextType.Terminator, expr, ignoreCase, tci) {
 
-  override val converter = if (isLengthKindDelimited) TerminatorCookerNoES else TerminatorCooker
+  override val converter = if (isLengthKindDelimited) TerminatorDelimitedCooker else TerminatorCooker
 }
 
 class TerminatorUnparseEv(expr: CompiledExpression[String], isLengthKindDelimited: Boolean, outputNewLine: OutputNewLineEv, tci: DPathCompileInfo)
   extends DelimiterUnparseEv(DelimiterTextType.Terminator, expr, outputNewLine, tci) {
 
-  override val converter = if (isLengthKindDelimited) TerminatorCookerNoES else TerminatorCooker
+  override val converter = if (isLengthKindDelimited) TerminatorDelimitedCooker else TerminatorCooker
 }
 
 class SeparatorParseEv(expr: CompiledExpression[String], ignoreCase: Boolean, tci: DPathCompileInfo)
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml
@@ -1236,7 +1236,7 @@ is multiple bytes in UTF-8 encoding that is used"
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>dfdl:terminator</tdml:error>
-      <tdml:error>cannot contain %ES;</tdml:error>
+      <tdml:error>ES entity cannot appear on its own</tdml:error>
       <tdml:error>dfdl:lengthKind="delimited"</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/entities_01.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/entities_01.tdml
@@ -46,8 +46,8 @@
     </tdml:document>
     <tdml:errors>
       <tdml:error>Separator</tdml:error>
-      <tdml:error>cannot</tdml:error>
-      <tdml:error>contain</tdml:error>
+      <tdml:error>contains</tdml:error>
+      <tdml:error>disallowed</tdml:error>
       <tdml:error>ES</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
@@ -267,8 +267,8 @@
     <tdml:errors>
       <tdml:error>Schema definition error</tdml:error>
       <tdml:error>nilValue</tdml:error>
-      <tdml:error>cannot</tdml:error>
-      <tdml:error>contain</tdml:error>
+      <tdml:error>contains</tdml:error>
+      <tdml:error>disallowed</tdml:error>
       <tdml:error>NL</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterPropertiesUnparse.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/DelimiterPropertiesUnparse.tdml
@@ -253,7 +253,7 @@
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>dfdl:terminator</tdml:error>
       <tdml:error>cannot</tdml:error>
-      <tdml:error>contain</tdml:error>
+      <tdml:error>own</tdml:error>
       <tdml:error>ES</tdml:error>
     </tdml:errors>
   </tdml:unparserTestCase>
@@ -331,8 +331,8 @@
   <tdml:errors>
     <tdml:error>Schema Definition Error</tdml:error>
     <tdml:error>Separator</tdml:error>
-    <tdml:error>cannot</tdml:error>
-    <tdml:error>contain</tdml:error>
+    <tdml:error>contains</tdml:error>
+    <tdml:error>disallowed</tdml:error>
     <tdml:error>ES</tdml:error>
   </tdml:errors>
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section13/nillable/literal-character-nils.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section13/nillable/literal-character-nils.tdml
@@ -230,8 +230,8 @@
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>nilValue</tdml:error>
-      <tdml:error>cannot</tdml:error>
-      <tdml:error>contain</tdml:error>
+      <tdml:error>contains</tdml:error>
+      <tdml:error>disallowed</tdml:error>
       <tdml:error>WSP*</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
@@ -1459,7 +1459,7 @@
     </tdml:document>
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStandardDecimalSeparator cannot contain %WSP;</tdml:error>
+      <tdml:error>textStandardDecimalSeparator contains disallowed character class(es): %WSP;</tdml:error>
     </tdml:errors>
 
   </tdml:parserTestCase>
@@ -1889,7 +1889,7 @@
     </tdml:document>
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStandardGroupingSeparator cannot contain %NL;</tdml:error>
+      <tdml:error>textStandardGroupingSeparator contains disallowed character class(es): %NL;</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
   
@@ -3621,8 +3621,8 @@
   <tdml:errors>
     <tdml:error>Schema Definition Error</tdml:error>
     <tdml:error>textStandardDecimalSeparator</tdml:error>
-    <tdml:error>cannot</tdml:error>
-    <tdml:error>contain</tdml:error>
+    <tdml:error>contains</tdml:error>
+    <tdml:error>disallowed</tdml:error>
     <tdml:error>%NL;</tdml:error>
   </tdml:errors>
   </tdml:parserTestCase>
@@ -3645,8 +3645,8 @@
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
       <tdml:error>textStandardGroupingSeparator</tdml:error>
-      <tdml:error>cannot</tdml:error>
-      <tdml:error>contain</tdml:error>
+      <tdml:error>contains</tdml:error>
+      <tdml:error>disallowed</tdml:error>
       <tdml:error>%WSP+;</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
@@ -3816,7 +3816,7 @@
 
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStandardExponentRep cannot contain %NL;</tdml:error>
+      <tdml:error>textStandardExponentRep contains disallowed character class(es): %NL;</tdml:error>
     </tdml:errors>
 
   </tdml:parserTestCase>
@@ -4090,7 +4090,7 @@
 
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStandardNaNRep cannot contain %NL;</tdml:error>
+      <tdml:error>textStandardNaNRep contains disallowed character class(es): %NL;</tdml:error>
     </tdml:errors>
 
   </tdml:parserTestCase>
@@ -4120,7 +4120,7 @@
 
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStandardInfinityRep cannot contain %NL;</tdml:error>
+      <tdml:error>textStandardInfinityRep contains disallowed character class(es): %NL;</tdml:error>
     </tdml:errors>
 
   </tdml:parserTestCase>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberPropsUnparse.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberPropsUnparse.tdml
@@ -470,7 +470,7 @@
     </tdml:infoset>
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>textStringPadCharacter cannot contain %WSP+;</tdml:error>
+      <tdml:error>textStringPadCharacter contains disallowed character class(es): %WSP+;</tdml:error>
     </tdml:errors>
   </tdml:unparserTestCase>
 
@@ -489,8 +489,8 @@
   <tdml:errors>
     <tdml:error>Schema Definition Error</tdml:error>
     <tdml:error>textStringPadCharacter</tdml:error>
-    <tdml:error>cannot</tdml:error>
-    <tdml:error>contain</tdml:error>
+    <tdml:error>contains</tdml:error>
+    <tdml:error>disallowed</tdml:error>
     <tdml:error>NL</tdml:error>
   </tdml:errors>
   </tdml:unparserTestCase>
@@ -510,8 +510,8 @@
   <tdml:errors>
     <tdml:error>Schema Definition Error</tdml:error>
     <tdml:error>textStringPadCharacter</tdml:error>
-    <tdml:error>cannot</tdml:error>
-    <tdml:error>contain</tdml:error>
+    <tdml:error>contains</tdml:error>
+    <tdml:error>disallowed</tdml:error>
     <tdml:error>ES</tdml:error>
   </tdml:errors>
   </tdml:unparserTestCase>


regetom:
I am a newbie to Daffodil and Scala.  I'm not sure if this was the best way to handle this request.  Also not sure if I should have had the if statement wrap the other 4 strings in the function.  Any feedback or suggestions are welcome.

stevedlawrence:
Welcome!

Though, it looks like this pull request only includes ``.idea/`` files created by IntelliJ. Did you forget to add your changes to the scala files?

Also, as a rule, we don't commit files created by IDEs so we don't have to keep track of everyone's unique editor configurations. Instead, we recommend that you globally ignore those files so you don't accidentally commit them. The default git ignore files to do this are ``$XDG_CONFIG_HOME/git/ignore`` or ``$HOME/.config/git/ignore``. So you could just add ``.idea`` to one of those files.

regetom:
Not sure why the Run Rat Check ran 5 hours 48 minutes and 49 seconds and had to be canceled for Java15, Scala2.12.11, Windows-latest test. All other Run RAT Check tests completed in less than 34 seconds.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -568,7 +568,9 @@ trait NoCharClassEntitiesMixin {
    * Override if set of prohibited class entities is different
    */
   protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-    context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
+    if(!raw.contains("%%")){
+      context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
+    }
```
stevedlawrence, line 573: This is actually a really tricky and subtle situtation. Because of this, I think this isn't quite right for the check to fix DAFFODIL-774.

As an example, textStandardZeroRep property is not allowed to contain the NL and ES character classes. So something like this is not allowed:
```xml
<xs:element ... dfdl:textStandardZeroRep="0_%NL;" />
```
In this example we have a zero rep of the string ``0``, followed by an underscore, followed by a newline. I believe this works as intented right now and creates an SDE because this contains an unallowed ``%NL;``

However, someone could theoretically do something like this (note the escaping `%%`` before NL):
```xml
<xs:element ... dfdl:textStandardZeroRep="0_%%NL;" />
```
So the zero rep in this case is the string ``0``, followed by an underscore, followed by the literal string ``%NL;`` instead of a newline. We would currently consider that an SDE because the raw string contains ``%NL;``, even though that's not really a ``%NL;`` because the ``%`` is escaped via a double ``%``. Note that it is very unlikely for a user to ever want a literal ``%NL;`` as a property value, so one might even argue making this check better as DAFFODIL-744 describes isn't worth it. Though it is an interesting excercise, and would make Daffodil more correct if we did add it.

However, what your check would do is make it so that if the raw property value contains ``%%`` then we don't even bother checking for ``%NL;`` at all. So for example, say we had something like this:
```xml
<xs:element ... dfdl:textStandardZeroRep="0%%_%NL;" />
```
So in this case, the user specified that the zero rep is the string ``0``, followed by a single perect sign (%% escape means single perect), followed by an underscore, followed by a newline. This shouldn't be allowed because the zero rep contains ``%NL;``. But because it contains %%, your new check would skip the ``%NL;`` check and it won't error as it should. Again, this is a contrived example, but that's what this edge case is talking about.

To properly fixed DAFFODIL-744, I don't think it's enough to just do simple string searches on the raw value like we do now. How we detect %NL; and other disallowed character classes must change, because it's too difficult to take into account all the different escape combinations. I think instead of looking at the raw value, we need to inspect the cooked value (or what we do with the cooked value). I'm not too familar with that code, but I think cooking the strings is supposed to deal with things like escaped percents, character classes, etc. and so the cooked result should have that escaping done and provide information to actually know if the property actually contains an unescaped %NL; or not. I'm not positive about this though.

If it's not clear where to get starting on making that change, it might be helpful to start a discussion on the dev list and see if there are any other ideas on how to best fix this, or maybe we'll decide it's not even worth the effort. It is a very rare edge case.

Whatever the solution, we also must make sure it works for ES, WSP, WSP+ and other character classes. Although the bug only lists NL as a problem, this same escaping issue exists for those properties.

Also, if you do continue working on this, I would reccommend starting with writing some TDML tests (maybe using the examples I've mentioned above as a starting point) to show we have the incorrect behavior, and what the correct behavior should be for some more complex cases with lots of escaping going on. It we determine the fix isn't worth the effort, we would at least have tests showing the expected behavior and how Daffodil doesn't do the right thing in case we ever do decide to fix it.

mbeckerle: I think you want to modify NoCharClassEntitiesMixin to make the checks do more than just search for %NL; or similar, because that does not exclude "%%NL;" which is allowed. 

The correct predicate test really has to use a regex to find an "unescaped %". The regex for this is quite tricky. It has to prevent an odd number of consecutive % characters. E.g., "%%%NL;" is %% then %NL;, so should cause an SDE. But "%%%%" is %%, %%, NL; which is allowed. A regex for "unescaped %" is the key concept, and this odd/even number of % is the core idea, and that can definitely be done with a regex.

BTW: my regex goto links are:
http://myregextester.com/
http://rick.measham.id.au/paste/explain.pl
https://www.regular-expressions.info/lookaround.html

stevedlawrence: Ah yeah, that's much simpler than what I was thinking. A even/odd check should do it just fine.

regetom:
I wasn't sure how to accomplish the even/odd requirement with regex, but I did come up with this:  context.schemaDefinitionUnless((!raw.contains("%NL;")) && !(raw.count(_ == '%') % 2 ==0)), str = "Property dfdl:%s cannot contain %%NL;", proprName)

I tested the results with this simple scala snippet:
object DfdlTest {
  def main(args: Array[String]) {
    val raw1 = "%%%NL";
    val raw2 = "%%%%NL";
  
    if(!(raw1.count(_== '%') % 2 == 0) && !(raw1.contains("%NL;")))
      println("raw1 = " + raw1 + " allowed")
    else
      println("raw1 = " + raw1 + " not allowed")

    if(!(raw2.count(_== '%') % 2 == 0) && !(raw2.contains("%NL;")))
      println("raw2 = " + raw2 + " allowed")
    else
      println("raw2 = " + raw2 + " not allowed")
  }
}
Let me know if this would be an acceptable solution or if I should do some further research toward a regex solution.

stevedlawrence:
I think an issue with that approach is that you can have multiple character classes in these property strings. For example:
```xml
<xs:element ... dfdl:textStandardZeroRep="%SP;%NL;" />
```
This is a space character followed by a newline. This shouldn't be allowed because ``%NL;`` is an unescaped newline, but in this case the count of ``%``'s is even so will be incorrectly allowed. So it's not just the count of ``%``'s that matters, but the count of all ``%``'s immediately preceeding an ``%NL;``. I think the following regex will do that (though, not heavily tested):
```regex
(^|[^%])(%%)*%NL;
```
So first we look for either the beginning of the string or a non-``%`` character with ``(^|[^%])``. Following that, we look for an even number of ``%``'s with ``(%%)*``. And finally following that we look for a ``%NL;``. I think that should match any ``%NL;`` not immediately preceeded by an even number of ``%``'s, in other words, an unescaped ``%NL;``.

mbeckerle:
Yeah, regex are really hard without myregextester, and the explain regex site tools. 

Here's a regex that matches even number of % followed by single % followed by either NL or ES followed by ";"
That is, it matches an "unescaped NL" or "unescaped ES".

```
(?:%%)*(%(?:NL|ES)\;)
```
You can of course generalize the NL|ES to the set of DFDL entities you want it to exclude. 

Here's a test string:
```
%NL;abcd%%ES;efgh%NLijkl%%%%%NL;mnop
```
And here's the matchAll output as described by myregextester.com
```
$matches Array:
(
    [0] => Array
        (
            [0] => %NL;
            [1] => %ES;
            [2] => %%%%%NL;
        )

    [1] => Array
        (
            [0] => %NL;
            [1] => %ES;
            [2] => %NL;
        )

)
```
And here's the explanation of this per 
http://rick.measham.id.au/paste/explain.pl?regex=%28%3F%3A%25%25%29*%28%25%28%3F%3ANL%7CES%29%5C%3B%29
```
NODE                     EXPLANATION
--------------------------------------------------------------------------------
  (?:                      group, but do not capture (0 or more times
                           (matching the most amount possible)):
--------------------------------------------------------------------------------
    %%                       '%%'
--------------------------------------------------------------------------------
  )*                       end of grouping
--------------------------------------------------------------------------------
  (                        group and capture to \1:
--------------------------------------------------------------------------------
    %                        '%'
--------------------------------------------------------------------------------
    (?:                      group, but do not capture:
--------------------------------------------------------------------------------
      NL                       'NL'
--------------------------------------------------------------------------------
     |                        OR
--------------------------------------------------------------------------------
      ES                       'ES'
--------------------------------------------------------------------------------
    )                        end of grouping
--------------------------------------------------------------------------------
    \;                       ';'
--------------------------------------------------------------------------------
  )                        end of \1
```

regetom:
Results from regex: (^|[^%])(%%)*%NL
image

Results from regex:  (?:%%)*(%(?:NL|ES)\;)
image

Possible change:
context.schemaDefinitionUnless(("(^|[^%])(%%)*%NL".r findFirstIn raw) != None)), str = "Property dfdl:%s cannot contain %%NL;", proprName)

Scala Test:
import scala.util.matching.Regex

object DfdlTest {
  def main(args: Array[String]) {
    val rawA = Array("%SP,%NL","%%SP,%%NL","%SP,%%NL","%%SP,%%%NL","%SP,%%%NL","%%SP,%%%%NL","%NL","%%NL","%NL;abcd%%ES;efgh%NLijkl%%%%%NL;mnop")

    for( raw <- rawA ){
      println(raw + " FFI Result = " + ("(^|[^%])(%%)*%NL".r findFirstIn raw))

      if(("(^|[^%])(%%)*%NL".r findFirstIn raw) != None)
        println(raw + " send to SDE")
      else
        println(raw + " allowed")
    }
  }
}

stevedlawrence:
Mike, nice improvement, the non-capture group stuff is nice and makes it easy to find and extract all unescaped character classes in a single regex. Could avoid a lot of duplication. Though, I think in your example the ``%ES;`` shouldn't be matched since that percent is escaped? The ``(^|[^%])`` trick should fix that I think? Adding that in and making it a non-capture group gives this regex:
```
(?:^|[^%])(?:%%)*(%(?:NL|ES)\;)
```
It's a bit ugly, but a single regex to capture all relevant information is nice, and documenting this isn't too terrible.

stevedlawrence:
@regetom yep, I think something like that is the right approach. Note that with Mike's technique of finding all illegal character classes in a single regex, I think we could use ``findAllIn`` instead of just finding the first with ``findFirstIn``. That way we can gather all the capture groups and output an error that includes all of them rather than just erroring on the first one found, maybe something like this (untested):

```scala
val matchedGroups = regex.findAllIn(raw).map { _.group(1) }
schemaDefinitionUnless(
  matchedGroups.isEmpty,
  "Property dfdl:%s contains disallowed character classes: %s",
  propName,
  matchedGroups.mkString(", "))
```

regetom:
when  I use (?:^|[^%])(?:%%)*(%(?:NL|ES)\;) I get this:
image
This works though ("(?:^|[^%])(?:%%)*(%(?:NL|ES))
image

stevedlawrence:
I think that semicolon just shouldn't be escaped. That's not a special character in regex so doesn't need escaping:
```
(?:^|[^%])(?:%%)*(%(?:NL|ES);)
```

regetom:
If I use this (?:^|[^%])(?:%%)*(%(?:NL|ES);), only the last item in the test comes back with Some().
image
Is that the desired behavior? Will there always be a ';'?

stevedlawrence:
Yes, there will also be a semicolon for a character class to be seen as a character class. If you just had something like ``%NLfoo``, you'll get a different error somewhere else that's something like ``found a ` that isn't followed by a valid character class``. That check happens elsewhere, so you shouldn't have to worry about that in this check. For this regex, you can assume the semicolon always exists since we're only looking for known character classes that aren't allowed.

mbeckerle:
Suggestion for not only this PR, but in the future, I think we should cut/paste the result of the regex explain into comments in the code to document the intention of the regex. 

The regex explanation is from http://rick.measham.id.au/paste/explain.pl

stevedlawrence:
I'd actually prefer to see a more human description, for example the description I gave above

> So first we look for either the beginning of the string or a non-% character with (^|[^%]). Following that, we look for an even number of %'s with (%%)*. And finally following that we look for a %NL;. I think that should match any %NL; not immediately preceeded by an even number of %'s, in other words, an unescaped %NL;.

This could definitely use improvment, but I think it's more clear what the regex is actually testing, and can give additional information about why we are looking for things (e.g. we need even number of %'s). If that explanation isn't clear, someone can always copy/paste the regex into a regex explainer tool and get a symbol by symbol breakdown.

Also, there might be some potential licensing issues? Not sure if those explains have a license for that output (or if it's even licensable). But always somethign to consider when copy/pasting something off the web.

regetom:
I get this error:
image
The compile is successful if I remove ".map { _.group(1) }"
In testing (with ".map { _.group }" removed) I get results like:
image
The original function calls hardcoded the character classes like "%%NL;" or "%%ES;" in the parameters.
In the test result instances smoetimes there is only one '%". Other times there can be more than two.  Should I format the results to take only the last three character (ex "NL;") and prepend two '%'?

mbeckerle:
I think you need to call findAllMatchIn, not findAllIn. I think findAllIn has to be used with an iterator pattern because there is just one Match object that is being advanced. Whereas findAllMatchIn provides separate match objects with disjoint state.

regetom:
findAllMatchIn works with .map { _.group(1) }
test results now look like this:
Property dfdl:test contains disallowed character classes: %NL;, %NL;
Property dfdl:test contains disallowed character classes: %ES;
Property dfdl:test contains disallowed character classes: %NL;, %WSP;, %NL;
by changing:
  "Property dfdl:%s contains disallowed character classes: %s", propName, matchedGroups.mkString(", "))
to
  "Property dfdl:%s contains disallowed character classes: %%%s", propName, matchedGroups.mkString(", %"))
I was able to get this:
Property dfdl:test contains disallowed character classes: %%NL;, %%NL;
Property dfdl:test contains disallowed character classes: %%ES;
Property dfdl:test contains disallowed character classes: %%NL;, %%WSP;, %%NL;

stevedlawrence:
What you have first looks right to me. I'm not sure we need the double %% in the error message. The issue is that there is an invalid character class. That fact that there are a zero or more escaped percents preceeding it doesn't really matter. The ``%NL;`` is the only invalid part.

There is maybe an argument that if there were a bunch of extra percents, that maybe the user just messed up their escaping, so we could give a warning to that effect, but I'm not sure that's really necessary and adds more complication--we would then need to capture the extra percents as well.

Also, it might make sense to use ``.distinct`` so we don't show the same character class multiple times in the warning, e.g.:
```scala
r.findAllMatchIn(raw).map { _.group(1) }.distinct
```
I'm not sure the duplication adds value and is potential confusing.

regetom:
I thought it was there so the SDE prints properly.  
context.schemaDefinitionUnless(!raw.contans("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
It was in the string passed to schemDefinitionUnless in the current code.

stevedlawrence:
Yep, in that case you do need to escape the % because the second paramater of schemaDefinitionUnless is a format string. But when you this:
```scala
schemaDefinitionUnless(..., "Property dfdl:%s contains disallowed character classes: %s", propName, matchedGroups.mkString(", "))
```
The format string only contains ``%s``, it doesn't contain ``%NL;`` anymore. The result of ``matchedGroups.mkString(", ")`` is just one giant string that does happen to contain %'s, but they aren't anything special because they aren't in the format string. schemaDefinitionUnless does a strait replace of %s with that string, and ignores the fact that the string might have perects in it.

regetom:
OK

regetom:
There are a couple of places in EntityReplacer.scala where the noCharClassEntities() is overridden in another class.  The class StringLiteralESEntityWithByteEntities() has the same original functions calls except for 'ES'.  The class ListOfStringLiteralNoCharClass_NL_ES_EntitiedsNoByteEntities() also overrides the function but only has function calls for "NL' and "ES".  The classes NonEmptyListOfStringLiteralCharClass_ES_WithByteEntities() and DelimiterCookerNoES() also override the functions.  Should all of these be modified to follow the new logic?

regetom:
It might be better to only have one function, but pass in a different regex string to control which entities (NL|ES|WSP|WSP+|WSP*) can throw SDE in the various classes?

stevedlawrence:
Definitely agreed. We should definitely try to factor out as much duplication as possible into a single function. Maybe something like each of these cookers that have invalid entities must implement a variable that's says what's disallows, e.g.
```scala
val disallowedCharClassEntities = Seq("NL", "ES", ...)
```
This can then be used to build the appropriate regex that the noCharClassEntities() would then use. No need to override that function anymore, and hides the complex regex so you don't have to dicipher it just to change/verify which entities are allowed.

mbeckerle:
Title of this PR no longer matches the nature of the change. This is "a good thing (tm)".

 Probably name of the corresponding JIRA ticket too. Suggest change to "Improve detection of disallowed entities in property values" or something like that.

regetom:
I will look at making the suggested changes (including to Jira ticket) and re-commit the changes when done.  Should I also look at adding additional tests in TDML?  I am not familiar with TDML is there somewhere you could point me to to get the basics?

mbeckerle:
https://daffodil.apache.org/tdml/
(I am happy to report that google found that for me.)

In the daffodil tree, the daffodil-test module is basically a big project containing almost entirely TDML tests. 

There are probably existing tests related to this feature and you can add tests to that test file, or copy the scala code driver file (which are just a JUInit thing that drives a TDML test as a JUnit test) and modify to make a new one to just drive a single TDML file with just your tests in it. 

Tests of textStandardZeroRep are in the scala driver file
src/test/scala/org/apache/daffodil/section13/text_number_props/TestTextNumberProps.scala
e.g.,  test_standardZeroRep01 to test_standardZeroRep11.

The corresponding TDML file is in the analogous
src/test/resources/..../section13/text_number_props/TextNumberProps.tdml

This file is pretty big, but I think it makes sense to just make sure the tests here cover the corner cases. 

I don't know what IDE you use, but in both Eclipse and IntelliJ IDEA, you can right click on the scala def of a test, and run exadtly that one test or debug exactly that one test. This is the primary way I debug and test daffodil.

regetom:
The overriden examples of noCharClassEntities((), have comments like:
  /**
   * Override Because ES is allowed.
   */
I am replacing the override functions with calls to the new function.  I am thinking I should modify these comments, but I am not sure what to say.  Something like?
*/
 * Disallow %NL, %WSP, %WSP+, WSP* character class entities
 * /
 Or should I just remove the comments altogether?

mbeckerle:
If the code is self explanatory, because the lists of allowed/disallowed entities are explicitly sitting there, then yeah, no comments are needed. Given a list of them though, it may not be obvious which one is not there. So a comment like // Not ES
may be helpful.
Alternatively you can do something like:
```
val entitiesWithoutES = Seq("%NL;", ....)
```
To make it self-documenting without using comments.

But in general, push your changes and let's see how it is coming out. 
I often push changes and immediately put some comments on the PR myself to ask people what they think about this or that aspect. Just like your question here.

regetom:
I added the parameter disallowedCharClassEntities: String to noCharClassEntities().  I was able to eliminate all the overrides of noCharClassEntities except for class DelimiterCookerNoES.  The error string built there also include this phrase'when dfdl:lengthKind=\"delimited\"."'.  This is the only class that includes extra text in the error string.  I was thinking about adding an additional parameter extraText: String to the noCharClassEntities() funtion, but I wasn't sure if that was the best solution.  I getting an error compiling this class because the override no longer matches the parameters in the noCharClassEntities() funtion.   Once I have a solution for that I will check the code in.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -810,7 +809,7 @@ class DelimiterCookerNoES(pn: String) extends ListOfString1OrMoreLiteral(pn, tru
   override val oneLiteralCooker: StringLiteralBase =
     new StringLiteralNoCharClassEntities(propName, true) {
 
-      override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
+      override protected def noCharClassEntities(raw: String, disallowedCharClassEntities: String, context: ThrowsSDE): Unit = {
         // TODO: this isn't quite right, as it will allow combined delimiters
```
regetom, line 813: I added the parameter disallowedCharClassEntities here to match the def above, but I am not using the function because of the additional error message test for these items. Perhaps there is a better way to handle this?

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -565,43 +565,59 @@ trait NoCharClassEntitiesMixin {
...
+    context.schemaDefinitionUnless(
+      matchedGroups.isEmpty,
+      "Property dfdl:%s contains disallowed character classes: %s" ,
+      propName,
+      matchedGroups.mkString(", "))
+  }
+
+  protected def testRaw(raw: String, disallowedCharClassEntities: String, context: ThrowsSDE): Unit = {
+    noCharClassEntities(raw, "NL|ES|WSP|WSP+|WSP*", context)
   }
```
stevedlawrence, line 592: I would avoid passing in things to testRaw, that should really only passing in the raw string, and then implementations of testRaw need to have the information for how to perform the raw test provided some other way. I'm also not a huge fan of the parameter being a string of entities separated by pipes, since it sort of implies it's being inserted into a regex. I think a Seq of strings is a little easierto visually scan, and is easier to change it the regex changes in the future.

What if instead we do something to how propName is used in this trait:

```scala
trait DisallowedCharClassEntitiesMixin {
  protected def propName: String
  protected def disallowedCharClassEntities: Seq[String]

  private lazy val disalllowedRegex = {
    val disallowedCharClassMatch = disallowedCharClassEntities.mkstring("|")
    val regexString = "(?:^|[^%])(?:%%)*(%(?:" + disallowedCharClassMatch + ");)"
    regexString.r
  }

  protected def testRaw(raw:String, context: ThrowsSDE): Unit = {
    val matchedGroup = disallowedRegex.finaAllMatchIn(raw).map { _.group(1) }
    conext.schemaDefinitionUnless(...)
  }
}
```
So this renames the trait to be a bit more descriptive about what it's really doing, since it isn't about "No" entities, but "Disallowed" entities. It also adds adds an abstract def ``disallowedCharClassEntities`` member that things that mix in this trait must implement to say what the disallowed entities are. Using this member, it builds and saves a regex (so it doesn't have to rebuild and compile it every time it's used). And the testRaw function has all the logic for actually testing and creating SDE's. The noCharClassEntities function isn't really needed anymore.

Then things that implement this trait just have to set that one member, e.g.

```scala
class StringLiteralNoCharClassEntities
  extends ...
  with DisallowedCharClassEntitiesMixin {
  
  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")

  ...
}
```

regetom:
Ii took at look at why all the checks failed.  Unit test are expecting results back like:
"cannot contain %ES;"
With the modifications, the wording for the new results has changed and the entities are combined:
Property dfdl:test contains disallowed character classes: %NL;, %ES;
I tried to determine where the unit tests are in the code base, but the paths in the failed check details don't seem to match the paths in the code base.

regetom:
Some of the problems maybe due to duplicate entities ex: %NL;, %ES:, %ES;
I removed "distinct" from the code because I was getting this compile error:
value distinct is not a member of Iterator[String]
       val matchedGroups = disallowedRegex.findAllMatchIn(raw).map { _.group(1) }.distinct

stevedlawrence:
> I removed "distinct" from the code because I was getting this compile error:
> value distinct is not a member of Iterator[String]

Scala 2.12 doesn't have the ``distinct`` function on the ``Iterator`` trait. The easiest way to get around this to to just convert the ``Iterator`` to a ``Seq``. This list will always be small enough (or even zero length in most cases) that this conversion shouldn't matter performance-wise.

``scala
val matchedGroups = disallowedRegex.findAllMatchIn(raw).map { _.group(1) }.toSeq.distinct
```

**Comment by stevedlawrence** [comment]
> I took a look at why all the checks failed. Unit tests are expecting results back like:
> "cannot contain %ES;"
> With the modifications, the wording for the new results has changed and the entities are combined:
> Property dfdl:test contains disallowed character classes: %NL;, %ES;

I think this is very likely part of the issue. We probably just need to update the TDML files for these negative tests to check for different keywords in the error message.

Though, I think we may need to double check which entities are disallowed. For example, one of the failing tests has this error message:
```
Test org.apache.daffodil.general.TestPrimitives.testLengthKindDelimited2 failed: java.lang.Exception: Schema Definition Error: Property dfdl:separator contains disallowed character classes: %WSP;, %WSP;
```
but the dfdl:separator property is allowed to have WSP as a separator. So I think something is off in the latest changes.

> I tried to determine where the unit tests are in the code base, but the paths in the failed check details don't seem to match the paths in the code base.

Are you seeing paths in ``/tmp`` in the error messages? If so, you're probably running into [DAFFODIL-2159](https://issues.apache.org/jira/browse/DAFFODIL-2159) where we show tmp paths instead of actual paths. That's something we just need to get fixed, it's definitely a pain.

To find tests, I usually just use grep. If you use eclipse or IntelliJ there might be an easier way to do it. Bur for example, in the above test error message, we know that the test class is called ``TestPrimitives`` and the test inside that class is called ``testLengthKindDelimited2``. If we grep for the test name:
```
$ grep -R testLengthKindDelimited2 *
daffodil-core/src/test/scala/org/apache/daffodil/general/TestPrimitives.scala:  @Test def testLengthKindDelimited2: Unit = {
```
We see that this test is defined in ``TestPrimitives.scala``. So we can open that file and find the function to see what it's testing.

Note that this particular test is in daffodil-core, so it's doesn't use any TDML tests. This test is just implemented in code, much like a unit test. Finding TDML tests can require a few more steps. As an example of and how to find a failing TDML test, you also have this as a failing test:

```
Error:  Test org.apache.daffodil.section13.nillable.TestNillable.test_entity_fail_06 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) Did not find diagnostic message "cannot" in any of the actual diagnostic messages: 
Error:  Schema Definition Error: Property dfdl:nilValue contains disallowed character classes: %NL;
```
Note that this particular test does look like a case where we just need to update the expected error message. In this case ``TestNillable`` is the class name and the test name is ``test_entity_fail_06``. Grepping for the test name, we get:
```
$ grep -R test_entity_fail_06 *
daffodil-test/src/test/scala/org/apache/daffodil/section13/nillable/TestNillable.scala:  @Test def test_entity_fail_06(): Unit = { runnerEntity.runOneTest("entity_fail_06") }
```
So this is in the ``TestNillable.scala`` file. We can see it is calling ``runOneTest`` using ``runnerEntity``, with the name of the TDML test case of ``entity_fail_06``. Opening up that file and looking at ``runnerEntity`` and how it's created, we can see that the TDML tests for that runner are in the ``entities_01.tdml`` file in ``/org/apache/daffodil/section06/entities/``. So we can then look in ``daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/entities_01.tdml`` add look for the parserTestCase with name ``entity_fail_06`` to find the inputs and expected outputs/errors of this test.

Note that a shortcut that works in the majority of cases (at least for tests in the daffodil-test project) is to just grep tdml files for the test name with the ``test_`` prefix removed--Scala tests have the prefix, associated TMDL test names do not. So for example, if the test name that failed was "test_fooBarTest_01", then you can just grep for "fooBarTest_01" in all tdml files, for example:
```
$ grep -R fooBarTest_01 --include *.tdml
```
More often than not, that will return one result which is the tdml file the test is in. If it doesn't return anything, you maybe have to take the longer approach mentioned before.

File: daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala [hunk 4]
```diff
@@ -560,48 +560,67 @@ sealed trait SingleCharacterMixin { self: StringLiteralBase =>
...
 }
 
 class StringLiteralNoCharClassEntities(pn: String, allowByteEntities: Boolean)
   extends StringLiteralBase(pn, allowByteEntities)
-  with NoCharClassEntitiesMixin {
+  with DisallowedCharClassEntitiesMixin {
 
+  // Check for all NL|ES|WSP|WSP+|WSP*
+  override val disallowedCharClassEntities = Seq("NL", "ES", "WSP", "WSP+", "WSP*")
```
stevedlawrence, line 599: I'm guessing the cause of some of the errors you're seeing (e.g. WSP not allowed in the dfdl:separator property) is that ``WSP+`` and ``WSP*`` contain regex characters. So if we blindly throw those in a regex, it will mess with our matches. We want to match the literal string ``WSP+``, but treating that as a regex will match ``WS`` followed by one or more ``P``'s, which we definitely don't want. Smilar issues for ``WSP*``.

Fortunately, it looks like the ``java.util.regex.Pattern`` class has a function just for escaping literal strings that are inserted into regex's. So I think we can just change the above to something like this, so that we escape all the disallowed char class entities before putting them in the regex.
val disallowedCharClassMatch = disallowedCharClassEntities.map(Pattern.quote).mkString("|")
```
Hopefully that will get rid of some of the errors.

**Comment by stevedlawrence** [comment]
I think I have an idea of what's going on here. The separator in this case is is ``%WSP;%WSP*;\%NL;%WSP;%WSP*;``. The previous test in ``DelimiterCookerNoES`` looked like this:
```scala
context.schemaDefinitionUnless(raw != "%WSP*;", "Property dfdl:%s cannot contain %%WSP*;...")
context.schemaDefinitionUnless(raw != "%ES;", "Property dfdl:%s cannot contain %%ES;...")
```
So although the error messages is "Property cannot contain %WSP* or %ES", it was really only checking if the entire property value is ``%WSP*;`` or ``%ES;``. In this test case, the entire property value is neither of those (it's a bunch of combined character classes) and so it was allowed.

But with your new changes you have made it so ``ES`` and ``WSP*`` are disallowed if they show up **anywhere**, which is the case in this example. So you've changed the behavior to be much more strict than it previously was.

The DFDL spec says about the dfdl separator property:
* DFDL Character Class ES is not allowed.
* The WSP* entity cannot appear on its own as one of the string literals in the list when determining the length of a component by scanning for delimiters.

So we are doing the correct thing for ES by disallowing it, but WSP* shouldn't be blanket disallowed. It should only be disallowed if it's the sole separator.

So for DelimiterCookerNoES, ``disallowedCharClassEntities`` should contain only ``ES``. And we need an additional check for a sole WSP*. So what we probably for this cooker is something like this:

```scala
override val disallowedCharClassEntities = Seq("ES")
override def testRaw(raw: String, context: ThrowsSDE) = {
  context.schemaDefinitionUnless(raw != "%WSP*;", "The WSP* entitty canont appear on it's own ...")
  super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
}
```
This way we manually test for the sole WSP* and use the disallowed logic to check for ES.

regetom:
Previously, I had 23 testing failing.  after applying Steve's suggested change above, only 16 tests failed. Of the 16 tests still failing, many were failing to a verbiage change in the error message.  Specifically 9 test failed because they were looking for the "cannot" in the diagnostic message.  I could not find the code that set "cannot" requirement, so I changed the error message verbiage from "... contains disallowed character classes ..." to "... cannot contain character classes ..." this change corrected 9 of the test failures.  The remaining test failures require more specific diagnostic meesages.  For example: 
[error] Test org.apache.daffodil.section13.text_number_props.TestTextNumberProps.test_infCharClasses failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) Did not find diagnostic message "textStandardInfinityRep cannot contain %NL;" in any of the actual diagnostic messages: 
[error] Schema Definition Error: Property dfdl:textStandardInfinityRep cannot contain character classes: %NL;
I modified the diagnostic message again, removing "cahracter classes" and all the tests passed. I'll check the code in with these changes.  If the diagnostic verbiage  "... contains disallowed character classes ..." is preferred, please point me to where in the code this would need to be modified.

mbeckerle:
In general it is better for tests to be relatively insensitive to wording of specific diagnostic messages. 
Instead of tests expecting "textStandardInfinityRep cannot contain %NL;" I claim it is better for the test to insist on finding 3 separate things in the diagnsotic message "textStandardInfiinityRep" "cannot", and "%NL;". That's sufficient evidence that the message conveys what is needed without being overly sensitive to changes in wording. 
In your case "cannot contain" vs. "disallowed"  however, .... the test must check for some word indicating that the message is about disallowed things being used. Changes of wording in messages of this sort, from "disallowed" to "cannot contain" or vice versa should be expected to break negative tests no matter what. But correcting the tests to match the new wording is routinely done as diagnostic messages are improved.

regetom:
I have one more test that is failing.  It appears to also occur in class DelimiterCookerNoES.  Previous to the modification, the error message was coded as:
//context.schemaDefinitionUnless(raw != "%ES;", "Property dfdl:%s cannot contain %%ES; when dfdl:lengthKind=\"delimited\".", propName)
The modification produces the error message: Property dfdl:terminator cannot contain %ES;
The resulting failure is:
[error] Test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) Did not find diagnostic message "dfdl:lengthKind="delimited"" in any of the actual diagnostic messages:
[error] Schema Definition Error: Property dfdl:terminator cannot contain %ES;
The dfdl:lengthKind=terminator is the actual error message appears to come from here:
<xs:schema dafint:file="file:/C:/Users/regis.thomas/incubator-daffodil/daffodil-test/target/scala-2.12/test-classes/org/apache/daffodil/section06/entities/Entities.tdml" targetNamespace="http://example.com" elementFormDefault="qualified" xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions" xmlns="http://example.com" xmlns:dafint="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int" xmlns:fn="http://www.w3.org/2005/xpath-functions" xmlns:tns="http://example.com" xmlns:ex="http://example.com" xmlns:ct="http://w3.ibm.com/xmlns/dfdl/ctInfoset" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData">
  <xs:include dafint:line="1208" dafint:col="6" schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
  <xs:annotation>
      <dfdl:format dafint:line="1209" dafint:col="6" terminator="" initiatedContent="no" initiator="" textNumberRep="standard" ignoreCase="no" byteOrder="bigEndian" separator="" encoding="US-ASCII" representation="text" ref="ex:GeneralFormat"/>
    </xs:appinfo>
  </xs:annotation>
  <xs:element dafint:line="1213" dafint:col="6" dfdl:lengthKind="implicit" name="root">
    <xs:complexType dafint:line="1214" dafint:col="8">
      <xs:sequence dafint:line="1215" dafint:col="10">
        <xs:element dafint:line="1216" dafint:col="12" dfdl:length="1" dfdl:lengthKind="explicit" type="xs:string" name="key"/>
        <xs:element dafint:line="1217" dafint:col="12" dfdl:terminator="{ if (../ex:key eq '1') then '%ES;' else 'END' }" dfdl:lengthKind="delimited" type="xs:string" name="value"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
I'm not sure how to fix this without reverting the code in DelimiterCookerNoES back to hardcode the dfdl:lengthKind to delimited.

regetom:
Here is the modified class definition:class DelimiterCookerNoES(pn: String) extends ListOfString1OrMoreLiteral(pn, true) {

  override val oneLiteralCooker: StringLiteralBase =
    new StringLiteralNoCharClassEntities(propName, true) with DisallowedCharClassEntitiesMixin {

      //override protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
      // TODO: this isn't quite right, as it will allow combined delimiters
      // that still match the empty string, e.g. "%ES;%WSP*;". We could check
      // if raw.contains("%WSP*;"), but that is too general, preventing valid
      // delimiters like "foo%WSP*;bar". Although the below matches the
      // specification, it's probably not the intended behavior.
      override val disallowedCharClassEntities = Seq("ES")
      override def testRaw(raw: String, context: ThrowsSDE) = {
        context.schemaDefinitionUnless(raw != "%WSP*;", """The WSP* entity cannot appear on it's own when dfdl:lengthKind="delimited".""")
        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
      }
      //context.schemaDefinitionUnless(raw != "%WSP*;", "Property dfdl:%s cannot contain %%WSP*; when dfdl:lengthKind=\"delimited\".", propName)
      //context.schemaDefinitionUnless(raw != "%ES;", "Property dfdl:%s cannot contain %%ES; when dfdl:lengthKind=\"delimited\".", propName)
      //}
    }
}
I tried to check in my changes but I need to check them into the new non-incubator branch and I didn't want to loose the previous comments.

stevedlawrence:
If we look at Cookers.scala, we see these lines:
```scala
object TerminatorCookerNoES extends DelimiterCookerNoES("terminator")

object SeparatorCooker extends DelimiterCookerNoES("separator")
```
So ``TerminatorCookerNoES`` (which is used for terminators when lengthKind="delimited") and ``SeparatorCooker`` both use ``DelimiterCookerNoES``. But if we look at the specification, terminators and separators actually have slightly different restrictions:

**dfdl:separator**
> * DFDL Character Classes NL, WSP, WSP+, and WSP* are allowed.
> * The WSP* entity cannot appear on its own as one of the string literals in the list when determining the length of a component by scanning for delimiters.

(Note, I'm not sure why dfdl:separator allows ES on it's own if WSP* isn't allowed, I think  that's a bug in the spec and that's just missing?)

**dfdl:terminator**
> * DFDL Character Classes NL, WSP, WSP+, WSP*, and ES are allowed.
> * Neither the ES entity nor the WSP* entity may appear on their own as one of the string literals in the list when the parser is determining the length of a component by scanning for delimiters.

So separator needs a cooker that always disallows ES and only disallows WSP* (and probably ES too) when it's alone (what you have now).

But terminator needs a cooker that allows all character classes (), and disallows both WSP* and ES when they're alone.

So I think we just need two separator Cookers. Maybe the one for separator is called ``DelimiterCookerNoES`` (what you have now but with I think the missing ES added), and one called ``DelimterCookerNoSoleES`` which is old DelimiterCookerNoES (has no restriction on character class, with the only restriction being sole things.

Note, I wonder if it's also a bug that separator doesn't allow ES? I feel like terminator and separator should have the same restrictions when when length kind is delimited? They are both terminating markup in that case, and I would expect to behave the same. Maybe the originl DleimiterCookerNoES is the right thing?

Going to need @mbeckerle to weigh in on this.

mbeckerle:
Hmmm. There is no such thing as lengthKind terminator. The lengthKind would be delimited with a non-empty terminator property.



Get Outlook for Android<https://aka.ms/ghei36>

stevedlawrence:
I think lengthKind="terminator" was a typo. The actual message does correctly have lengthKind="delimited".

The question I had was why does the DFDL spec say ``ES`` can never be used in the ``dfdl:separator`` property, but it can be used in the ``dfdl:terminator`` property as long as it isn't alone as one of the string literals when lengthKind="delimited"? What's the usecase for ever needing ES in ``dfdl:terminator``?

But thinking about it, I guess the usecase is an optional terminator with with non-delimited lengths. So something like ``dfdl:terminator="FOO %ES;"`` seems reasonable when lengthKind is not delimited. So I guess I answered my own question.

So @regetom, I think this does mean we need two separate Cookers. One is needed for cooking separators that is called DelimiterCookerNoES (same as what you have now), and one called something like DelimterCookerNoSoleES which is like the old DelimiterCookerNoES (has no restriction on character class, with the only restriction being sole ES/WSP*).

mbeckerle:
I can comment from the DFDL specification and workgroup perspective.

The use case is indeed that where the length kind is something like 'pattern', and depending on the length, a terminator somewhere needs to be present or absent.

This comes up in formats where a string can be say, up to 50 characters long, where if it is exactly 50 characters long, there is no terminator (%ES; case), but if it is less than 50 characters long, a specific character like DEL (x7F) is the terminator. The VMF format (mil-std-6017) is an example of a format that has this kind of strings.

Here's an example:

  <xs:group name="tString50">
    <xs:sequence>
      <xs:element name="value" type="xs:string" dfdl:lengthKind="pattern"
        dfdl:lengthPattern="[^\x7F]{0,49}(?=\x7F)|.{50}" />
      <xs:sequence
        dfdl:terminator="{if (fn:string-length(./value) eq 50) then '%ES;' else '%DEL;'}" />
    </xs:sequence>
  </xs:group>

Note that the regex for the pattern is using look-ahead to see if it finds the DEL terminator or not without consuming it as part of the length. That's the trick here.

The dfdl:separator property does not offer this same possibility of using %ES; because a separator doesn't necessarily act like a terminator, (e.g., separatorPosition 'prefix' is a blend of initiator and terminator) doesn't have many of the associated properties like emptyValueDelimiterPolicy controlling it, but does have entirely different properties related to separator suppression. Disallowing %ES; was just a conservative design choice. In the future it is possible to relax this restriction if all the nuances of property interaction could be worked out. But separators are just so different from terminators that without a compelling use case that cannot be equivalently modeled as terminators, the complexity did not seem worth it.

stevedlawrence:
Thanks for the background Mike. Definitely make sense. So that confirms that we need two separate cookers for the slight differences between separators and terminator requirements.

regetom:
I mis-stated the diagnostic message here is a portion of the actual message:
[info] Test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 started
[error] Test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) Did not find diagnostic message "dfdl:lengthKind="delimited"" in any of the actual diagnostic messages:
[error] Schema Definition Error: Property dfdl:terminator cannot contain %ES;
the test is looking ofr the text "dfdl:lengthKind="delimited" but the diagnotic contains the text "dfdl:terminator".  This is coming from the attempt to create a generic message format.  Perhaps my changes to EntityReplacer.scala has caused some other side effect or possibly there is an error in the setup of test_emptyStringEntityTermInExpressionDelimited_01.  Unfortunately I am not familiar enough with the daffodil code base to make a determination.  I tried to check my code chnages into the new repository so everyone is looking at the same code, but I received a permission denied error.
remote: Permission to apache/daffodil.git denied to regetom.
fatal: unable to access 'https://github.com/apache/daffodil.git/': The requested URL returned error: 403
regetom@NTELVD4:/mnt/c/Users/regis.thomas/incubator-daffodil/daffodil$ git push origin daffodil-774-SDE_esc_char_entities

stevedlawrence:
To make your code visible to us, make sure to push it to the branch on your fork (github.com/regetom) rather than apache (github.com/apache). That will automatically update this pull request.

mbeckerle:
If your changes are pushed to this branch of your fork, any other developer can check them out to look at your code. You need not push them elsewhere. Github specifically supports the ability to checkout a PR. 

I have a .gitconfig file in my homedir with these contents:
`[user]
	name = Michael Beckerle
	email = ....my email....
[alias]
	pr = "!f() { git fetch $1 refs/pull/$2/head:pr/$1/$2; } ; f"

[core]
	excludesfile = /home/mbeckerle/.gitignore
`
That alias allows me to type
`git pr asf 481`
and that will checkout your branch of this PR. The "asf" is my git remote name for the central daffodil repository location. (My own fork is named "origin")

The error message about terminator cannot be %ES; really should also mention lengthKind="delimited" because when length kind is NOT delimited, %ES; is allowed for terminator.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -281,7 +281,7 @@ final class EntityReplacer {
    * this form, but allow the others.
    */
   private def replaceEntity(proposedEntity: String, orig: String, context: Option[ThrowsSDE], forUnparse: Boolean,
-    allowByteEntity: Boolean): String = {
+                            allowByteEntity: Boolean): String = {
```
stevedlawrence, line 284: Check your IDE settings for indentation. You've got a few places wher indentation was changed, but shouldn't be.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -560,48 +560,68 @@ sealed trait SingleCharacterMixin { self: StringLiteralBase =>
...
   }
 
   protected def testRaw(raw: String, context: ThrowsSDE): Unit = {
-    noCharClassEntities(raw, context)
+    val matchedGroups = disallowedRegex.findAllMatchIn(raw).map {_.group(1)}
+    context.schemaDefinitionUnless(
+      matchedGroups.isEmpty,
+      "Property dfdl:%s cannot contain %s" ,
+      propName,
+      matchedGroups.mkString(", "))
```
stevedlawrence, line 591: If there are multiple disallowed character class found, the error message would look something like ``Property dfdl:foo cannot contain ES, WSP*, WSP``, which I personally find a little odd and maybe inconsistent. Suggest rewording it, I personaly liked what you had before, and we can just update the expected error messages in the TDML files.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -806,17 +816,22 @@ class NonEmptyListOfStringLiteralCharClass_ES_WithByteEntities(pn: String)
...
+      override val disallowedCharClassEntities = Seq("ES")
+      override def testRaw(raw: String, context: ThrowsSDE) = {
+        context.schemaDefinitionUnless(raw != "%WSP*;", """The WSP* entity cannot appear on it's own when dfdl:lengthKind="delimited".""")
+        super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
       }
+      //context.schemaDefinitionUnless(raw != "%WSP*;", "Property dfdl:%s cannot contain %%WSP*; when dfdl:lengthKind=\"delimited\".", propName)
+      //context.schemaDefinitionUnless(raw != "%ES;", "Property dfdl:%s cannot contain %%ES; when dfdl:lengthKind=\"delimited\".", propName)
+      //}
     }
 }
```
stevedlawrence, line 837: I think what you have here for DelimiterCookerNoES is correct for the dfdl separator property, but is not correct for dfdl:terminator. ES is allowed in the dfdl:terminator property, so we need a different cooker that has different checks.

For the dfdl:terminator proeprty, there are no disallowed character class entities, and both WSP* and ES canot appear on their own when lengthKind is delimited. This logic is different from this ``DelimiterCookerNoES``. So you need a new class that is similar to this cooker (maybe call it ``DelimiterCookerNoSoleES``), but it should not mixin the ``DisallowedCharClassEntitiesMixin`` and it should not have ``disallowedCharClassEntities``. All it has is ``testRaw`` that checks for the same WSP* check, but also does the same check for ES.

Then, in Cookers.scala, ``object TerminatorCookerNoES extends DelimiterCookerNoES("terminator")`` extends this new ``DelimiterCokerNoSoleES``.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -835,4 +850,4 @@ class DelimiterCooker(pn: String) extends ListOfStringLiteralBase(pn, true) {
    */
   override protected def oneLiteralCooker: StringLiteralBase = Assert.usageError("not to be used.")
   override protected def cook(raw: String, context: ThrowsSDE, forUnparse: Boolean): List[String] = Assert.usageError("not to be used")
-}
+}
```
stevedlawrence, line 853: Check your IDE settings to make sure it ends files with a newline.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -560,48 +560,68 @@ sealed trait SingleCharacterMixin { self: StringLiteralBase =>
...
+   *
    */
-  protected def noCharClassEntities(raw: String, context: ThrowsSDE): Unit = {
-    context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%ES;"), "Property dfdl:%s cannot contain %%ES;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP;"), "Property dfdl:%s cannot contain %%WSP;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP+;"), "Property dfdl:%s cannot contain %%WSP+;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP*;"), "Property dfdl:%s cannot contain %%WSP*;", propName)
+  private lazy val disallowedRegex = {
+//    val disallowedCharClassMatch = disallowedCharClassEntities.mkString("|")
```
stevedlawrence, line 579: Makesure to remove commented out code.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -560,48 +560,68 @@ sealed trait SingleCharacterMixin { self: StringLiteralBase =>
...
 
 class SingleCharacterLiteralNoCharClassEntities(pn: String, allowByteEntities: Boolean)
   extends StringLiteralBase(pn, allowByteEntities)
-  with NoCharClassEntitiesMixin
-  with SingleCharacterMixin {
+    with DisallowedCharClassEntitiesMixin
+    with SingleCharacterMixin {
 
-  override protected def testRaw(raw: String, context: ThrowsSDE) = super.testRaw(raw, context)
+  // Check for all NL|ES|WSP|WSP+|WSP*
```
stevedlawrence, line 610: These ``// CHeck for `` comments dont' really add much. The code ``disalloweCharClassEntites = ...`` is self documenting enough. Suggest removing these // Check comments.

regetom:
I created DelimiterCookerNoSoleES I didn't change Cookers.scala. The test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 still fails. From the results it seems like the logic flows through DelimiterCookerNoES instead of the new DelimiterCookerNoSoleES. If I modify Cooker.scala and change the line
object TerminatorCookerNoES extends DelimiterCookerNoES("terminator")
to
object TerminatorCookerNoES extends DelimiterCookerNoSoleES("terminator")
then the test hangs here:
[info] Test run started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.testNearestEnclosingSequenceElementRef started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.testImmediatelyEnclosingModelGroup1 started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.testStaticallyFirstWithChoice started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.testRequiredSiblings started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.test_hasStaticallyRequiredOccurrencesInDataRepresentation_1 started
[info] Test org.apache.daffodil.dsom.TestMiddleEndAttributes.test_hasStaticallyRequiredOccurrencesInDataRepresentation_2 started
[info] Test run finished: 0 failed, 0 ignored, 6 total, 0.162s
| => org.apache.daffodil.dpath.TestDPath 829s
| => org.apache.daffodil.infoset.TestInfosetInputter1 829s
| => org.apache.daffodil.dsom.TestDsomCompilerUnparse1 832s
| => org.apache.daffodil.infoset.TestInfosetInputterFromReader 832s
It seems to hang on this test group:
[info] Test run started
[info] Test org.apache.daffodil.dpath.TestDPath.test_oneUpwardStepsAcrossElementReference started
[info] Test org.apache.daffodil.dpath.TestDPath.test_twoUpwardSteps started
[info] Test org.apache.daffodil.dpath.TestDPath.test_twoUpwardStepsAcrossElementReference started

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -818,20 +819,27 @@ class DelimiterCookerNoES(pn: String) extends ListOfString1OrMoreLiteral(pn, tru
...
+class DelimiterCookerNoSoleES(pn: String) extends ListOfString1OrMoreLiteral(pn, true) {
+
+  override val oneLiteralCooker: StringLiteralBase =
+    new StringLiteralNoCharClassEntities(propName, true) {
+
+      // "Neither %ES or %WSP*;" can appear on their own when lenghtKind is delimited.
+      override def testRaw(raw: String, context: ThrowsSDE) = {
+        context.schemaDefinitionUnless(raw != "%ES;", """The ES entity cannot appear on its own when dfdl:lengthKind="delimited".""")
+        context.schemaDefinitionUnless(raw != "%WSP*;", """The WSP* entity cannot appear on its own when dfdl:lengthKind="delimited".""")
+        testRaw(raw, context)
```
stevedlawrence, line 841: This is the line where things are probably hangning.

This line is calling ``testRaw``, which is itself, which creates an infinite loop and make things hang. Previously, this called ``super[DisallowedCharClassEntitiesMixin].testRaw``, which is a different ``testRaw`` function in a different trait that has additional logic for checking for disallowed entities. But this class has no disallowed entities, so we don't need to call it, or any other testRaw function. This function contains all the logic we need.

So removing this line should prevent the hang, and hopefully allow the tests to pass.

regetom:
I removed call to testRaw from DelimiterCookerNoSoleES.  I had to define noCharClassEntities() otherwise the compile complained that "contexct" and "raw" did not exist.  The test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 still fails.  The error message seems to indicate the logic is still flowing through DelimeterCookerNoES instead of DeilitimerCookerNoSoleEs.  Here is the relevant error message from the log;
[info] Test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 started
[error] Test org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) Did not find diagnostic message "dfdl:lengthKind="delimited"" in any of the actual diagnostic messages: 
[error] Schema Definition Error: Property dfdl:terminator contains disallowed character class(es): %ES;
[error] Schema context: ex:value Location line 1217 column 12 in file:/tmp/EmptyStringEntityInExpressionDelimited5351926126395074609.dfdl.xsd, took 0.116 sec
[error]     at org.apache.daffodil.tdml.TDMLException$.apply(TDMLException.scala:31)
[error]     at org.apache.daffodil.tdml.VerifyTestCase$.$anonfun$verifyAllDiagnosticsFound$4(TDMLRunner.scala:1564)
[error]     at org.apache.daffodil.tdml.VerifyTestCase$.$anonfun$verifyAllDiagnosticsFound$4$adapted(TDMLRunner.scala:1553)
[error]     at scala.collection.immutable.List.foreach(List.scala:392)
[error]     at org.apache.daffodil.tdml.VerifyTestCase$.verifyAllDiagnosticsFound(TDMLRunner.scala:1553)
[error]     at org.apache.daffodil.tdml.TestCase.checkDiagnosticMessages(TDMLRunner.scala:811)
[error]     at org.apache.daffodil.tdml.ParserTestCase.runParseExpectErrors(TDMLRunner.scala:928)
[error]     at org.apache.daffodil.tdml.ParserTestCase.$anonfun$runProcessor$4(TDMLRunner.scala:856)
[error]     at org.apache.daffodil.tdml.ParserTestCase.$anonfun$runProcessor$4$adapted(TDMLRunner.scala:853)
[error]     at scala.util.Either$RightProjection.foreach(Either.scala:652)
[error]     at org.apache.daffodil.tdml.ParserTestCase.runProcessor(TDMLRunner.scala:853)
[error]     at org.apache.daffodil.tdml.TestCase.run(TDMLRunner.scala:796)
[error]     at org.apache.daffodil.tdml.DFDLTestSuite.runOneTest(TDMLRunner.scala:386)
[error]     at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:144)
[error]     at org.apache.daffodil.tdml.Runner.runOneTest(RunnerFactory.scala:150)
[error]     at org.apache.daffodil.section06.entities.TestEntities.test_emptyStringEntityTermInExpressionDelimited_01(TestEntities.scala:98)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]     at java.lang.reflect.Method.invoke(Method.java:566)
[error]     ...

stevedlawrence:
> I removed the call to testRaw from DelimiterCookerNoSoleES. I had to define noCharClassEntities() otherwise the compiler complained that "context" and "raw" did not exist.

I'm not sure why you get this error, but nothing calls ``noCharClassEntities``. You have removed all calls of it, which I think is correct. The ``testRaw`` function is the only that that does anything now. So adding the ``noCharClassEntities`` probably isn't the right solution, and we need to bring the ``testRaw`` function back, since that's what actually does tests.

In your current pull request that you've pushed, if I change
```scala
protected def noCharClassEntities(...)
```
to
```scala
override def testRaw(...)
```
then it compiles without any errors. And if I update ``Cookers.scala`` to use this new cooker:
```scala
object TerminatorCookerNoES extends DelimiterCookerNoSoleES("terminator")
```
and run the test you mention, then I do get the error error:

> The ES entity cannot appear on its own when dfdl:lengthKind="delimited

Which seems right to me. Unfortunately, the test still fails because the TDML test expects ``dfdl:terminator`` in the error message, but that's expected since the new error message no longer includes that information (we should probably add that back to the error message). But with those changes I mentioned, it does seem to be doing the right test and detecting the sole ES.

One additional suggestion that I just noticed: In the new ``DelimiterCookerNoSoleES``, the ``oneLiteralCooker`` member is a ``StringLiteralNoCharClassEntities``. That class implies that there are some char class entities that are never allowed. But that isn't really the case with our new ``NoSoleES`` cooker. That new cooker allows all entities, but just has a special restriction on how some can be used. Since we don't need any of the logic associated with ``StringLiteralNoCharClassEntities``, we should just make it a ``StringLiteralBase``, for the new ``NoSoleES`` cooker:

```scala
  override val oneLiteralCooker: StringLiteralBase =
    new StringLiteralBase(propName, true) {
```

regetom:
I changed object TerminatorCookerNoES extends DelimiterCookerNoES("terminator") to
object TerminatorCookerNoES extends DelimiterCookerNoSoleES("terminator") in cookers.scala .
I also made thes uggested changes to DelimiterCookerNoSoleEs in EntityReplacer.scala.
2 tests failed, but the error seemed to indicate that DelimiterCookerNoSoleES was called, so I changed the tdml files for those tests and now no tests fail.  Not sure if changing the tdml file was the correct fix.

stevedlawrence (approved):

stevedlawrence:
+1 :+1:  looks good to me

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala
```diff
@@ -63,7 +63,7 @@ object InitiatorCooker extends DelimiterCooker("initiator")
 
 object TerminatorCooker extends DelimiterCooker("terminator")
 
-object TerminatorCookerNoES extends DelimiterCookerNoES("terminator")
+object TerminatorCookerNoES extends DelimiterCookerNoSoleES("terminator")
```
tuxji, line 66: Is it appropriate for `TerminatorCookerNoES` to keep its original name or should it be renamed to `TerminatorCookieNoSoleES`?

stevedlawrence: Good point. Maybe a more accurate name would be ``TerminatorDelimitedCooker``? Since that is what differentiates this cooker from the normal ``TerminatorCooker``--this one applies when lengthKind="delimited".

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -616,18 +632,12 @@ class SingleCharacterLiteralNoCharClassEntitiesWithByteEntities(pn: String = nul
...
-    context.schemaDefinitionUnless(!raw.contains("%NL;"), "Property dfdl:%s cannot contain %%NL;", propName)
-    // ES is allowed
-    context.schemaDefinitionUnless(!raw.contains("%WSP;"), "Property dfdl:%s cannot contain %%WSP;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP+;"), "Property dfdl:%s cannot contain %%WSP+;", propName)
-    context.schemaDefinitionUnless(!raw.contains("%WSP*;"), "Property dfdl:%s cannot contain %%WSP*;", propName)
-  }
+  override val disallowedCharClassEntities = Seq("NL", "WSP", "WSP+", "WSP*")
+  override def testRaw(raw: String, context: ThrowsSDE) =
+    super[DisallowedCharClassEntitiesMixin].testRaw(raw, context)
 }
```
tuxji, line 641: We got code not covered by tests warnings for lines 638 and 640.  That means this class (StringLiteralESEntityWithByteEntities) isn't covered by any tests.  In fact, StringLiteralESEntityWithByteEntities isn't used anywhere in the codebase as far as I can see.  Should we remove it?

tuxji (approved):
+1

Fine with me too.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
stevedlawrence: Skimming through these, it actually looks like we have a few more these classes that aren't ever used.

We might want to just create a new cleanup ticket to go through all those and remove the ones that aren't used or figure out why some aren't used. It's possible we're just missing some cases where these were expected to be used and never implemented them.

stevedlawrence:
@regetom, i think this is close to being ready to merge. Can you rename TerminatorCookerNoES, and then squash all your changes into a single commit, update that commit to describe the changes, include ``DAFFODIL-774`` at the end, and force push to your fork? Once that's complete, we can merge it. Thanks!

regetom:
Yes, I'll try to get that done today.

regetom:
The name change from TerminatorCookerNoES to TerminatorDelimitedCooker seems to have caused a bunch check fails, should I revert the name back?

stevedlawrence:
I think the new name makes more sense. The issue is that you changed the name where the class is defined, but didn't change the name where it's actually used, which is causing it to fail to compile. Take a look at the compile errors in the check failures. The error messages will point out the file and line numbers where the old name is still being used.

regetom:
I modified the code below to reflect the new name but now 2 tdml tests have failed, I will replace the expected results, but it doesn't seem this change should have had an effect on the tests if they passed with the previous function name.
**Error:  /home/runner/work/daffodil/daffodil/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala:26:8: object TerminatorCookerNoES is not a member of package org.apache.daffodil.cookers
Error:  import org.apache.daffodil.cookers.TerminatorCookerNoES
Error:         ^
Error:  /home/runner/work/daffodil/daffodil/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala:120:55: not found: value TerminatorCookerNoES
Error:    override val converter = if (isLengthKindDelimited) TerminatorCookerNoES else TerminatorCooker
Error:                                                        ^
Error:  /home/runner/work/daffodil/daffodil/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala:126:55: not found: value TerminatorCookerNoES
Error:    override val converter = if (isLengthKindDelimited) TerminatorCookerNoES else TerminatorCooker
Error:                                                        ^
Error:  /home/runner/work/daffodil/daffodil/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvDelimiters.scala:26:36: Unused import**

stevedlawrence:
Yeah, that's surprising to me too that this class rename would change in TDML results. Once you push your changes we can see if the change makes sense or not.

regetom:
Is this request still good to go?  Or do I need re-approval after the changes?

stevedlawrence:
Looks good to me. Please rebase your commis to the latest master and squash into a single commit with a descriptive commit message, and this should be good to merge.

regetom:
I seem to have lost my last changes during the rebase. I am adding them back.  Do I need to do another rebase?

stevedlawrence:
Yes, once you've added back your changes please squash them all into a single commit and force push.

stevedlawrence:
Looks good! Thank for the contribution!

PR #482, opened 2021-01-27
Update debugger support of found delimiters/fields/diffs

stevedlawrence:
- Refactor how diff's are calculated so that all the logic is moved into
  the individual command objects that care about that piece of
  information. Now, any info command that is "diffable" must implement
  the InfoDiffable trait and the 'diff' function. The "info diff"
  command now automatically finds all commands that implement this trait
  and will call this diff function. This simplifies the info diff
  command and makes it so it doesn't need to know anything about other
  info commands or anything about the state.
- Instead of "info foundDelimiter" showing both the found delimiter and
  the found field, it now only shows the found delimiter. And add a new
  "info foundField" command to show the found field that "info
  foundDelimiter" used to show. This makes the code less complex and
  gives the user more control over what information they want to see.
  These new commands are also made diffable so they now show up in "info
  diff". When diffs are found, this command outputs something like this:
  ```
    diff:
      foundDelimiter: none -> ,
      foundField: none -> smith
  ```
  The above shows that a comma delimiter was found and that the field
  proceeding that delimiter was "smith".

DAFFODIL-758

> Note this doesn't fully resolve DAFFODIL-758, this is just an incremental step in improving trace output.

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
@@ -17,39 +17,40 @@
 
 package org.apache.daffodil.debugger
 
-import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.schema.annotation.props.gen.Representation
-import org.apache.daffodil.processors._
-import org.apache.daffodil.infoset._
-import org.apache.daffodil.processors.parsers._
-import org.apache.daffodil.xml.XMLUtils
-import org.apache.daffodil.xml.GlobalQName
-import org.apache.daffodil.xml.QName
 import java.io.File
+
+import scala.collection.JavaConverters._
+
+import jline.console.completer.AggregateCompleter
 import jline.console.completer.Completer
 import jline.console.completer.StringsCompleter
-import jline.console.completer.AggregateCompleter
-import org.apache.daffodil.util.Enum
-import org.apache.daffodil.dsom.ExpressionCompilerClass
-import org.apache.daffodil.dpath.NodeInfo
+
+import org.apache.daffodil.BasicComponent
+import org.apache.daffodil.api.DaffodilTunables
 import org.apache.daffodil.dpath.ExpressionEvaluationException
-import org.apache.daffodil.util.DPathUtil
-import org.apache.daffodil.util.Misc
-import org.apache.daffodil.oolag.ErrorsNotYetRecorded
-import org.apache.daffodil.processors.unparsers.UState
-import org.apache.daffodil.processors.unparsers.Unparser
+import org.apache.daffodil.dpath.NodeInfo
+import org.apache.daffodil.dsom.ExpressionCompilerClass
 import org.apache.daffodil.dsom.RelativePathPastRootError
-import org.apache.daffodil.exceptions.UnsuppressableException
-import scala.collection.mutable
 import org.apache.daffodil.dsom.RuntimeSchemaDefinitionError
-import org.apache.daffodil.util.Misc
+import org.apache.daffodil.exceptions.Assert
+import org.apache.daffodil.exceptions.UnsuppressableException
 import org.apache.daffodil.infoset.InfosetElement
 import org.apache.daffodil.infoset.XMLTextInfosetOutputter
-import org.apache.daffodil.processors.parsers.ConvertTextCombinatorParser
+import org.apache.daffodil.infoset._
+import org.apache.daffodil.oolag.ErrorsNotYetRecorded
 import org.apache.daffodil.oolag.OOLAG._
-import scala.collection.JavaConverters._
-import org.apache.daffodil.api.DaffodilTunables
-import org.apache.daffodil.BasicComponent
+import org.apache.daffodil.processors._
+import org.apache.daffodil.processors.parsers.ConvertTextCombinatorParser
+import org.apache.daffodil.processors.parsers._
+import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.processors.unparsers.Unparser
+import org.apache.daffodil.schema.annotation.props.gen.Representation
+import org.apache.daffodil.util.DPathUtil
+import org.apache.daffodil.util.Enum
+import org.apache.daffodil.util.Misc
+import org.apache.daffodil.xml.GlobalQName
+import org.apache.daffodil.xml.QName
+import org.apache.daffodil.xml.XMLUtils
 
 abstract class InteractiveDebuggerRunner {
   def init(id: InteractiveDebugger): Unit
@@ -136,12 +137,16 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
 
   var debugState: DebugState.Type = DebugState.Pause
 
-  override def init(parser: Parser): Unit = {
+  var previousProcessorState: StateForDebugger = _
+
+  override def init(state: PState, parser: Parser): Unit = {
     runner.init(this)
+    previousProcessorState = state.copyStateForDebugger
   }
 
-  override def init(unparser: Unparser): Unit = {
+  override def init(state: UState, unparser: Unparser): Unit = {
     runner.init(this)
+    previousProcessorState = state.copyStateForDebugger
   }
 
   override def fini(parser: Parser): Unit = {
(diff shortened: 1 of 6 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1528,37 +1542,91 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
       }
 
-      object InfoFoundDelimiter extends DebugCommand with DebugCommandValidateZeroArgs {
+      object InfoFoundDelimiter extends DebugCommand with DebugCommandValidateZeroArgs with InfoDiffable {
         val name = "foundDelimiter"
         override lazy val short = "fd"
         val desc = "display the current found delimiter"
         val longDesc = desc
+
+        private def getDelimOpt(s: StateForDebugger): Option[String] = {
```
bsloane1650, line 1551: Since we are in runtime1, we want to use Maybe.

This is just the debugger, so Option is probably fine if you actually need the extra functionality.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1528,37 +1542,91 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          val delimOptPre = getDelimOpt(prestate)
+          val delimOptPost = getDelimOpt(state)
+
+          (delimOptPre, delimOptPost) match {
+            case (None, None) => false
+            case (Some(pre), Some(post)) if pre == post => false
+            case (_, _) => {
+              val strPre = delimOptPre.getOrElse("none")
+              val strPost = delimOptPost.getOrElse("none")
```
bsloane1650, line 1577: Could just do: case (Some(strPre), Some(strPost)) instead of deconstructing again.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1528,37 +1542,91 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
+          DebugState.Pause
+        }
+
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          val delimOptPre = getDelimOpt(prestate)
+          val delimOptPost = getDelimOpt(state)
+
+          (delimOptPre, delimOptPost) match {
+            case (None, None) => false
+            case (Some(pre), Some(post)) if pre == post => false
```
bsloane1650, line 1574: These cases are the same: case (a, b) if a == b => false

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1528,37 +1542,91 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
+
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          val fieldOptPre = getFieldOpt(prestate)
+          val fieldOptPost = getFieldOpt(state)
+          (fieldOptPre, fieldOptPost) match {
+            case (None, None) => false
+            case (Some(pre), Some(post)) if pre == post => false
+            case (_, _) => {
+              val strPre = fieldOptPre.getOrElse("none")
+              val strPost = fieldOptPost.getOrElse("none")
```
bsloane1650, line 1616: See InfoFoundDelimiter

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1567,6 +1635,15 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
           }
           DebugState.Pause
         }
+
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          if (prestate.groupPos != state.groupPos) {
+            debugPrintln("%s: %d -> %d".format(name, prestate.groupPos, state.groupPos), "  ")
```
bsloane1650, line 1641: We are doing this for bassically every diff.

Would it make more sense for the interface to instead be:

```
getValue(state: State): Option[Any]
```
Then let the trait itself implement the diff() function?

bsloane1650 (approved):
+1.

Some minor improvements suggested.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1320,21 +1337,40 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
           }
           DebugState.Pause
         }
+
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          if (prestate.bitLimit0b != state.bitLimit0b) {
+            debugPrintln("%s: %d -> %d".format(name, prestate.bitLimit0b, state.bitLimit0b), "  ")
```
mbeckerle, line 1343: codecov says this is uncovered. Probably should have a test that outputs ALL diffs.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1528,37 +1542,91 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
-                debugPrintln("foundDelimiter: %s".format(Misc.remapStringToVisibleGlyphs(pr.matchedDelimiterValue.get)), "  ")
-              } else {
-                debugPrintln("%s: nothing found".format(name))
-              }
-            }
-            case ustate: UState => {
-              // TODO
+          val delimOpt = getDelimOpt(state)
+          val str = delimOpt.getOrElse("nothing found")
+          debugPrintln("%s: %s".format(name, str))
```
mbeckerle, line 1564: Codecov is also complaining about this. Seems like this may be spurious? You showed examples of this. Tests of this would have to be under the daffodil-cli/src/test directory.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1638,6 +1716,15 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
           }
           DebugState.Pause
         }
+
+        def diff(prestate: StateForDebugger, state: ParseOrUnparseState): Boolean = {
+          if (prestate.arrayPos != state.arrayPos) {
+            debugPrintln("%s: %d -> %d".format(name, prestate.arrayPos, state.arrayPos), "  ")
```
mbeckerle, line 1722: I see now that this change set does not include any CLI tests that drive this new info-diff system, and that's why codecov is complaining. You should add at least one test that just does info diff on a contrived example that exercises all of these diffs.

mbeckerle (changes requested):
This is a great improvement. A CLI test that exercises the new diff codes is needed. I also concur with a few of bsloane's suggestions for refactoring the diff code to do some of the work in the shared trait.

stevedlawrence:
I've added a new commit that adds much more refactoring, including some of the suggestions above. This is almost an RFC level of changes. The message in the new commit describes what changed. I haven't yet added tests to improve coverge, curious what others think of these changes before doing that.

mbeckerle:
I like the concept of eliminating the before-capture, and using the prior after-state as the incoming before state. Code does look cleaned up a bit. 

I think it will be nearly equivalent when used. Instead of showing diffs for only one element you'd have to stop before that element, then single step then info diff to see only the deltas for that one step. Seems fine. 

I think the real proof of this will be how well it works out for unparsing. 

btw I think when debugging, it's highly likely we want to retry all suspensions at each stop (repeating until no more of them successfully evaluate).  At least that should be one mode which is to evaluate suspensions as soon as possible. Perhaps another mode would evaluate them "normally", but I fear this will just be viewed as non-deterministic by users. They won't have any idea why things are being evaluated at various points. 

I'm just waiting for some tests to one-plus this.

bsloane1650:
New changes look good.

stevedlawrence:
I've added a couple tests to make sure all the diff code paths are covered. I'm not sure if codecov is going to update this PR, but the results can be seen here:

https://app.codecov.io/gh/apache/incubator-daffodil/compare/482/diff

There's still probably room for improvement in debugger coverage, but this at least covers all the new/existing diff logic.

mbeckerle (approved):
+1

PR #483, opened 2021-01-29
Enable 3 tests showing apostrophes work properly now in TDML

mbeckerle:
Also many tests differing only by the test config had identical test names in the TDML.
These were modified so the ones that specify a config have a suffix added to their names.

DAFFODIL-1928, DAFFODIL-1420

### DIFF
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestFileBuffering.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestFileBuffering.scala
@@ -33,36 +33,36 @@ class TestUnparserFileBuffering {
 
   import TestUnparserFileBuffering._
 
-  @Test def test_puaInfosetChars_01(): Unit = { runner.runOneTest("puaInfosetChars_01") }
-  @Test def test_puaInfosetChars_02(): Unit = { runner.runOneTest("puaInfosetChars_02") }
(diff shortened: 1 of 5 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestFileBuffering.scala
```diff
@@ -33,36 +33,36 @@ class TestUnparserFileBuffering {
 
   import TestUnparserFileBuffering._
 
-  @Test def test_puaInfosetChars_01(): Unit = { runner.runOneTest("puaInfosetChars_01") }
-  @Test def test_puaInfosetChars_02(): Unit = { runner.runOneTest("puaInfosetChars_02") }
+  @Test def test_puaInfosetChars_01(): Unit = { runner.runOneTest("puaInfosetChars_01_ffb") }
```
bsloane1650, line 36: Do we want to add the _ffb suffix to the JUnit testname as well?

mbeckerle: Good idea. Will do.

tuxji (approved):
+1

Looks good to me!

bsloane1650 (approved):
+1

stevedlawrence (approved):
+1

PR #484, opened 2021-02-01
updateEclipseClasspaths failing silently for eclipse IDE setup

mbeckerle:
Added println so proper operation of UpdateEclipseClasspaths is visible. 

This was failing silently due to DAFFODIL_HOME being set incorrectly 
so I added System.err.println to the code and assertion checks to 
make it clear it is operating correctly. 

The bug here was just that if DAFFODIL_HOME is incorrect it was failing
silently. 

DAFFODIL-2454

### DIFF
--- a/project/UpdateEclipseClasspaths.scala
+++ b/project/UpdateEclipseClasspaths.scala
@@ -23,6 +23,8 @@ import scala.language.reflectiveCalls
 /**
  * After 'sbt eclipse' creates the projects, you still have to update every .classpath file
  * so that all but daffodil-macro-lib have a dependency on daffodil-macro-lib.
+ * 
+ * That is what this does.   
  */
 
 object UpdateEclipseClasspaths extends App {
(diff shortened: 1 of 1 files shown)


daffodil-tdml-processor/src/test/resources/test/tdml/tdmlQuoting.tdml
```diff
@@ -93,7 +93,7 @@
 
 	<tdml:parserTestCase name="apos_test2"
 		root="apostrophe" model="apostrophe"
-		description="Can't deal properly with apostrophe (&apos;) marks within the document data in TDML " roundTrip="true">
+		description="Shows proper handling of apostrophe (&apos;) marks within the document data in TDML " roundTrip="true">
```
stevedlawrence, line 96: Can you rebase this change onto master and remove the apostrophy changes that are in a separate commit?

mbeckerle: Yes. Mistake.

project/UpdateEclipseClasspaths.scala
```diff
@@ -63,6 +65,7 @@ object UpdateEclipseClasspaths extends App {
     //
     val s = System.getenv("DAFFODIL_HOME")
     assert(s ne null, "$DAFFODIL_HOME undefined")
+    System.err.println("DAFFODIL_HOME environment variable: " + s)
```
mbeckerle, line 68: Probably these should all just be println. Not system.err.println.

stevedlawrence: I'm wondering if we should change this to use sbt logging rather than println? Then we can give appropriate info/error diagnostics, so errors will really jump out? It might take a bit of refactoring to passing in the ``stream.log`` from sbt. And it more greatly ties this tool to sbt, but that doesn't seem unreasonable. 

Also related, I just opened [DAFFODIL-2465](https://issues.apache.org/jira/browse/DAFFODIL-2465). If it's okay to add assume this is always run via sbt, then we can also easily get access to baseDirectory.value and I think get rid of DAFFODIL_HOME completely.

project/UpdateEclipseClasspaths.scala
```diff
@@ -32,15 +32,17 @@ object UpdateEclipseClasspaths extends App {
...
     val classPathFiles = epl.flatMap { f =>
       if (f.isDirectory())
         f.listFiles().filter { _.getName == ".classpath" }
       else Nil
     }.filterNot { _.toString.contains("daffodil-macro-lib") }
 
+    assert(!classPathFiles.isEmpty, "No subproject directories found. Check DAFFODIL_HOME environment variable.")
     classPathFiles.foreach { f =>
-      println("updating " + f.toString)
+      println("Updating " + f.toString)
```
stevedlawrence, line 45: SBT standards are actually to start message with lowercase, so I think "updating" is correct if were following that.

stevedlawrence (approved):
+1, though I suggest we make this tool specific to SBT and so we can use sbt logging and baseDirectory variables. Can be done as a separate pull request some other day since that might be a bit more involved and maybe not high priority?

mbeckerle:
I just spent 2 hours debugging by print statement and in fact I gave up and just debugged the problem by moving the code into daffodil-macro-lib as a top-level application, and then debugging it in Eclipse. 

I think sbt plugin debugging has to be done the way embedded debugging usually works which is by attaching a debugger to another jvm started with debug flags/ports specified. Which is a pain in the neck.

By keeping this thing as ordinary code, no-sbt-specific things in it, I can debug it as ordinary code. I don't need a debug system that can cope with debugging sbt plugins. 

Having now got that debugged.... I've moved the code back to the project/ directory so it will run from sbt.

stevedlawrence:
I discovered that the sbt eclipse plugin has a built-in way to make changes to these files. Take a look at PR #486 and see if that seems like a reasonable alternative to maintaining updateEclipseClasspaths.

tuxji (approved):
+1 

Changes look good if you still want to merge them.  However, I also looked at PR #486 and it adds only a few extra lines to build.sbt & plugins.sbt.  If these ~10 new lines work, they'll be much easier to maintain than the ~100 old lines in in UpdateEclipseClasspaths.scala.

mbeckerle:
I'm merging this anyway, because at least then the code in the history is not flawed should we want to grab it again for any reason. (E.g., to use with some other IDE which doesn't have sbt support for example.)

PR #485, opened 2021-02-01
Confirm behavior when compiling schemas without global elements

stevedlawrence:
If a primary schema does not contain any global elements, and no root
element is given when compiling the schema, the correct behavior is a
schema definition error. If a root element is provided when compiling a
schema, we are able to search included/imported schemas for this
element. This adds/modifies tests to confirm this behavior.

Note that TDML runner always provides a root element, so testing the
negative case requires a CLI test. This also slightly modifies related
tests that were not quite correct based on the description of the test.

DAFFODIL-913

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
@@ -206,27 +206,19 @@ class TestCLIparsing {
   }
 
   @Test def test_1593_CLI_Parsing_MultifileSchema_noGlobalElem(): Unit = {
-    val tmp_filename: String = (System.currentTimeMillis / 1000).toString()
-    val file = new File(tmp_filename)
-    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/TopLevel.dfdl.xsd")
-    val inputFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/02nine_headers.txt")
-    val (testSchemaFile, testInputFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile)) else (schemaFile, inputFile)
-
-    val shell = Util.start("")
+    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/multi_base_21.dfdl.xsd")
+    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+    val shell = Util.startIncludeErrors("")
 
     try {
-      val cmd = String.format("%s parse -s %s -o %s %s", Util.binPath, testSchemaFile, tmp_filename, testInputFile)
+      val cmd = String.format(Util.echoN("does not matter") + " | %s parse -s %s", Util.binPath, testSchemaFile)
       shell.sendLine(cmd)
-
-      //TODO:update
-      //val err = shell.getCurrentStandardErrContents()
-      //assertEquals(err, "")
+      shell.expectIn(1, contains("No global elements"))
+      shell.expectIn(1, contains("multi_base_21.dfdl.xsd"))
       shell.sendLine("exit")
       shell.expect(eof)
     } finally {
       shell.close()
-      //Remove the temporary output file
-      assertTrue("Failed to remove temporary file: %s".format(file), file.delete)
     }
   }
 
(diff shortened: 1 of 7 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -206,27 +206,19 @@ class TestCLIparsing {
   }
 
   @Test def test_1593_CLI_Parsing_MultifileSchema_noGlobalElem(): Unit = {
-    val tmp_filename: String = (System.currentTimeMillis / 1000).toString()
-    val file = new File(tmp_filename)
-    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/TopLevel.dfdl.xsd")
```
stevedlawrence, line 211: This schema has global elements and doesn't actaully test anything. I think this was originally meant to be a stub test or something. Modified it to test what the name implies it should test.

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -206,27 +206,19 @@ class TestCLIparsing {
...
-    val (testSchemaFile, testInputFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile)) else (schemaFile, inputFile)
-
-    val shell = Util.start("")
+    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/multi_base_21.dfdl.xsd")
+    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+    val shell = Util.startIncludeErrors("")
 
     try {
-      val cmd = String.format("%s parse -s %s -o %s %s", Util.binPath, testSchemaFile, tmp_filename, testInputFile)
+      val cmd = String.format(Util.echoN("does not matter") + " | %s parse -s %s", Util.binPath, testSchemaFile)
```
tuxji, line 214: Nice use of echo to avoid creating/deleting a temporary input file.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
```diff
@@ -2415,4 +2417,23 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+  <!--
+    Test name: schemaNoGlobalElems_03
+    Schema: multi_base_21.dfdl.xsd
+    Root: base2
+    Purpose: This test demonstrates that you can have a schema if there are
+             no global elements in the primary schema, as long as the root element is
+             specified (the TDML Runner always specifies the root)
```
tuxji, line 2426: I searched TDMLRunner.scala for "root" to understand what you meant by "the TDML Runner always specifies the root".  What seems to happen is if a TDML test has an infoset, the TDML Runner uses that infoset's root element (XML documents have a single root element) as the root.  The root="name" attribute is not even needed and in fact the TDML Runner throws an exception if the root="name" attribute does not match the name of the root element of the infoset.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
```diff
@@ -2415,4 +2417,23 @@
...
+  -->
+
+  <tdml:parserTestCase name="schemaNoGlobalElems_03"
+    root="distant_element" model="multi_base_21.dfdl.xsd" description="">
```
tuxji, line 2430: Therefore, the root="distant_element" attribute here is redundant since the test case has an infoset with the distant_element as well.

tuxji (approved):
+1

I learned that the root attribute on TDML test cases can be omitted most of the time even when a schema has multiple global elements.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
mbeckerle: TDML runner used to require the root attribute. When we made it smarter to look at the infoset, we decided not to make it illegal to have the root element, but to check consistency if it is present. Just to avoid editing hundreds of tests for a non-functional reason.

mbeckerle (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
stevedlawrence: Yeah, the root attribute is technically redundant. I don't think that was always the case, not sure when that feature was added. I wonder if we should have have a daffodil standard that the root attribute shouldn't be provided for positive test cases? I do kindof like it just for the sanity check. I know I sometimes copy the infoset from another test and forget to update the root element. In that case I get an error about the root attribute not matching.

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
stevedlawrence: Ah yeah, I missed this comment. I do like the consistency check.

PR #488, opened 2021-02-07
Support choices in runtime2 code generator with restrictions

tuxji:
Extend the C code generator in daffodil-runtime2 to support choices in
DFDL schemas with restrictions (no backtracking or discrimination
needed).  An xs:choice complex type must have a dfdl:choiceDispatchKey
expression pointing to a preceding infoset element which must contain
an integral value.  Each alternative in the choice must be labeled
with a dfdl:choiceBranchKey attribute having only integer literals.
The C code generator makes some assumptions to simplify converting the
choiceDispatchKey expression to C code accessing the preceding infoset
element:

 - the expression starts with '{xs:string( and ends with )}'
 - the expression returns the value of a previous element without
   changing the value in any way (except converting it to xs:string)
 - both the expression and the C code use only local names (for now...)
 - we can map the context node's path to a Unix-like slash path
 - all dpath operations look like Unix-like relative paths (../tag)
 - we can normalize the new path and convert it to C struct dot notation
 - we can store the accessed value in an int64_t local variable safely

Also add some DFDL schemas with choice elements to make sure C code
generator keeps working for these schemas.

Changelog:

In build.sbt, trim unnecessary C compile flags (note we don't use the
.o files built by sbt-cc, we only want C compilation error/warning
messages).

In c/Makefile, put *.h files in HEADERS (not SOURCES) and trim
-Wno-missing-field-initializers from CFLAGS since we've fixed all
these warnings in the C code now.

In daffodil_argp.c, fix warnings about missing field initializers.

In daffodil_main.c, fix warnings about missing field initializers.

In infoset.c, store offsets as size_t instead of ptrdiff_t since
they're always nonnegative numbers.  Handle CHOICE typeCode
appropriately (i.e., call ERD member function initChoice) when walking
infoset with visitor methods.

In infoset.h, add initChoice to ERD member functions and CHOICE to
TypeCode enumerations.  Store offsets as size_t instead of ptrdiff_t
since they're always nonnegative numbers.  Define static constant
NO_CHOICE as value to determine when _choice fields are uninitialized.

In parsers.c/unparsers.c, reorder macro definitions slightly.

Add NestedUnion.[ch] to show how C code can support choice complex
types.  C code was generated from nested.dfdl.xsd schema using
NestedUnion as the root element.

In ex_nums.[ch], replace old generated code with new generated code.

In CodeGenerator.scala, add scaladoc comments and remove -lpthread
since we don't need to link C executable with it.

In Runtime2CodeGenerator.scala, add braces to generateCode method.
Note that I tried to detect choice complex types when walking grammar
objects in generateCode method but found out it was very difficult to
detect them in that place while it was easier to detect them in
ElementParseAndUnspecifiedLengthCodeGenerator.scala instead.

In CodeGeneratorState.scala, rename qualifiedName function to erdName
and improve it to qualify erdName with global complex type's name and
append _ERD before returning erdName.  Update calls in rest of file
accordingly.  Make addImplementation method generate new ERD member
function initChoice.  Add choiceDispatchField function to convert a
choiceDispatchKey expression into C struct dot notation for accessing
the choiceDispatchKey expression's runtime value.  We'll probably
revise the choiceDispatchField function a few times to handle more
cases as they come up.  Add new methods addBeforeSwitchStatements and
addAfterSwitchStatements to generate declarations, ERD definitions,
offset computations, ERD computations, init statements, initChoice
statements, parse statements, and unparse statements needed to handle
choices with several alternative branches.  Make addComplexTypeERD
method include correct count and initChoice member function in ERDs
containing alternative branches.  Make addComplexTypeStatements method
generate C code (switch statement cases) for initChoice, parse, and
unparse functions of complex types with alternative branches.  Make
minor changes to addFieldDeclaration, generateCodeFile and
generateCodeHeader methods to improve the generated C code.

In ElementParseAndUnspecifiedLengthCodeGenerator.scala, call
cgState.addBeforeSwitchStatements and cgState.addAfterSwitchStatements
in right places to generate C code for complex types having choices.

In TestRuntime2.tdml, change defaultRoundTrip to "none" since every
parse test is accompanied by an unparse test, making the extra work
done by "onePass" redundant.  Or maybe we should keep
defaultRoundTrip="onePass" and delete all the unparse tests, but it's
nice to have debuggable entry points for both parse and unparse tests.

Add egress_xdcc_bw.dfdl.xsd and ingress_xdcc_bw.dfdl.xsd from a
collaborator who already gave permission for Apache license.  Add
egress_xdcc_bw.tdml and ingress_xdcc_bw.tdml with corresponding parse
binary data files and unparse XML files to test all choice
alternatives in both schemas.

Add nested.dfdl.xsd to provide a much smaller schema also using
dfdl:choiceDispatchKey and dfdl:choiceBranchKey for development and
testing.  Add nested.tdml with corresponding binary data files and
unparse XML files to test NestedUnion in nested schema.

In orion-command.dfdl.xsd, incorporate latest changes from
collaborator to define global elements using global complex type
declarations instead of nested complex type declarations.

In orion-command.tdml, change defaultRoundTrip to "none" to avoid
redundancy and speed up tests.

Add TestEgressXdccBw.scala, TestIngressXdccBw.scala, and
TestNested.scala to provide entry points for calling all of the new
test cases.

In Rat.scala, tell ratCheck to ignore newly added binary data files.


daffodil-runtime2/src/main/resources/examples/ex_nums.c
```diff
@@ -660,10 +559,10 @@ littleEndian_unparseSelf(const littleEndian *instance, UState *ustate)
 static void
 ex_nums_initSelf(ex_nums *instance)
 {
+    instance->_base.erd = &ex_nums_ERD;
     array_initSelf(&instance->array);
     bigEndian_initSelf(&instance->bigEndian);
     littleEndian_initSelf(&instance->littleEndian);
-    instance->_base.erd = &ex_nums__ERD;
 }
```
stevedlawrence, line 567: What are these src/main/resources/examples files for? Are these examples of what the code generate will create? Are these for testing purposes? Perhaps they should be in src/test/resources?

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -105,41 +129,178 @@ class CodeGeneratorState {
...
+          expr.substring(before.length, expr.length - after.length) else expr
+        val normalizedURI = new URI(absoluteSlashPath + "/" + relativePath).normalize
+        normalizedURI.getPath.substring(1)
+      case _ => ""
+    }
+    // Strip namespace prefixes since C code uses only local names (for now...)
+    val localDispatchSlashPath = dispatchSlashPath.replaceAll("/[^:]+:", "/")
+    val res = localDispatchSlashPath.replace('/', '.')
+    res
+  }
```
stevedlawrence, line 172: As this runtime2 matures, I wonder if eventually we would want to add code generate support to the DPath. This way you can take advantage all of the DPath expression capabilities like type checking/conversion. But then rather than creating our runtime1 objects, you can generate code. This likely requires a decent amount of refactoring of the DPath, but probably somethign worth keeping in the back of our minds. As is, this feels pretty fragile, and I'm not even sure it will error if a dispatch expression doesn't match the supported format.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ingress_xdcc_bw.tdml
```diff
@@ -0,0 +1,196 @@
...
+    root="GapsPDU">
+    <tdml:infoset>
+      <tdml:dfdlInfoset type="file">ingress_xdcc_bw_unparse_116.xml</tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:document>
+      <tdml:documentPart type="file">ingress_xdcc_bw_parse_116.dat</tdml:documentPart>
+    </tdml:document>
+  </tdml:unparserTestCase>
+
+</tdml:testSuite>
```
stevedlawrence, line 196: Does it make sense to put all these tests in daffodil-tests, or somewhere not runtime2 specific? It makes sense that they are in here because they're made to ensure that runtime2 works, but these schemas aren't runtime2 specific. They should run just fine in DFDL runtime, or maybe even the IBM DFDL runtime.

I would think that maybe runtime2 should only contain things like unit tests that are truely specific to runtime2. For example, tests that generate code and then examing the generated code to make sure the right thing is generated. Not sure we need that kind of test, but that's the kind of tests that should be in runtime2 rather than daffodil-test.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -162,10 +322,31 @@ class CodeGeneratorState {
   def addComplexTypeStatements(child: ElementBase): Unit = {
     val C = localName(child)
     val e = child.name
+    val hasChoice = structs.top.initChoiceStatements.nonEmpty
```
stevedlawrence, line 325: If I'm reading this right, it seems this only supports direct dispatch choices that are the model group of a complex type, e.g.:
```xml
<xs:element name="foo">
  <xs:complexType>
    <xs:choice dfdl:choiceDispatchKey="...">
      ...
    </xs:choice>
  </xs:complexType>
</xs:element>
```
So sequences with "anonymous" choices are not suppported? I guess supporting those adds more complications since you currently use ``_choice`` to figure out which union member is used, and anonmous choices means you could have multiple unions and _choice's, which someone need a distint name?

Does this distint name issue also arise in sequence with multiple elements with the same name? For example:
```xml
<xs:element name="foo">
  <xs:complexType>
    <xs:sequence>
      <xs:element name="a" ... />
      <xs:element name="b" ... />
      <xs:element name="a" ... />
    </xs:choice>
  </xs:complexType>
</xs:element>
```
Would this currently result in to struct members with the name "a" and cause an error?

stevedlawrence (approved):
+1

Couple minor questions just to improve my understanding of this generator. Really nice how relatively little code is needed for this!

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
mbeckerle: I think for runtime2 it would be acceptable to rule out this sort of schema - requiring all children to have unique names seems like a sensible restriction given all the other restrictions in runtime2.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -105,41 +129,178 @@ class CodeGeneratorState {
     qnameInit
   }
 
+  /**
+   * We want to convert a choiceDispatchKey expression into C struct dot
+   * notation (rootElement->[subElement.field]) which will access the C
+   * struct field containing the choiceDispatchKey's runtime value.
+   *
```
mbeckerle, line 136: This is ok  temporarily. Eventually we want to modify the DPath expression compiler to generate C code from expressions. That shouldn't be too hard, but will require refactoring code a bit.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
mbeckerle: Yes, the right way to do this eventually is have DPath generate C-code from expressions.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -105,41 +129,178 @@ class CodeGeneratorState {
...
+           |        childNode->erd = childERD;
+           |        return true;
+           |    }
+           |    else
+           |    {
+           |        return false;
+           |    }""".stripMargin
+      val parseStatement =
+        s"""    default:
+           |        pstate->error_msg = "node's _choice field was not initialized during parsing";
```
mbeckerle, line 249: I already made this comment in the example code. But this should be a different kind of error. We need to distinghish errors that come from users making mistakes writing schemas, and failures of the implementation to correctly compile or run the schema.
The latter are fatal errors and the right thing is probably to fail and stop processing. (Fail in simplest way possible.)

mbeckerle (changes requested):
I am not that sure about my comments. The tests suggest far more is already working than I suspected from reviewing the code changes.

tuxji (commented):
Thanks for the review comments.  I've replied with my own comments.  Once I push a new commit, may I have a second +1 (assuming the changes look good) so I can squash, merge, and start working on another pull request?  I want to support xs:boolean and fix a C scope name collision in a new collaborator schema.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
mbeckerle: This kind of error - choice not getting initialized properly. Is not a DFDL "processing error" in the formal DFDL sense that it should cause parser backtracking or result in a "Parse Error" message.
It's an internal code invariant failure. It indicates the runtime2 system has a bug and is not initializing things properly. 
These kinds of errors are strongly distinguished throughout the Daffodil code base. We should continue that in the runtime2 code. 
I think you should just call a global invariantFailed("bug - choice field not initialized") function. That should probably just abort/throw/longjmp out and end processing. In principle, once the runtime2 code is correct, this code path should not be even be reachable.

tuxji: I did the following tests:

```shell
$ cd daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2
$ daffodil generate c -s nested.dfdl.xsd -r NestedUnion
$ cp nested.dfdl.xsd c/
$ cp nested_union_parse_4.dat c/parse.dat
$ cp nested_union_unparse_4.xml c/unparse.xml
$ cd c
$ make tests
./daffodil parse parse.dat -o test_unparse.xml
xmldiff unparse.xml test_unparse.xml

./daffodil unparse unparse.xml -o test_parse.dat
diff parse.dat test_parse.dat
$ : Also check Daffodil runtime1...
$ daffodil parse -s nested.dfdl.xsd -r NestedUnion -o test_unparse.xml parse.dat # No errors
$ xmldiff unparse.xml test_unparse.xml # No errors
$ daffodil unparse -s nested.dfdl.xsd -r NestedUnion -o test_parse.dat unparse.xml # No errors
$ diff parse.dat test_parse.dat # No errors
$ : Now change the choice dispatch key in parse.dat...
$ emacs parse.dat # change 4th byte from ^D to ^@
$ make parse-test
./daffodil parse parse.dat -o test_unparse.xml
./daffodil: Parse error: no match between choice dispatch key and any branch key
make: *** [Makefile:42: parse-test] Error 1
$ daffodil parse -s nested.dfdl.xsd -r NestedUnion -o test_unparse.xml parse.dat
[error] Parse Error: Choice dispatch key (0) failed to match any of the branch keys.
Schema context: choice[1] Location line 71 column 12 in file:/home/interran/apache/incubator-daffodil-runtime2/daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/c/nested.dfdl.xsd
Data location was preceding byte 4
$ : Now change the choice dispatch key in unparse.xml as well....
$ emacs unparse.xml # change <tag>4</tag> to <tag>0</tag>
$ make unparse-test
./daffodil unparse unparse.xml -o test_parse.dat
./daffodil: Walk error: no match between choice dispatch key and any branch key
make: *** [Makefile:46: unparse-test] Error 1
$ daffodil unparse -s nested.dfdl.xsd -r NestedUnion -o test_parse.dat unparse.xml # No errors
$ diff parse.dat test_parse.dat # No errors, 4th byte is zero in both files
```

My point is that this line in CodeGeneratorState which we are talking about makes sure c-daffodil behaves the same as scala-daffodil.  Both report a processing error saying there is no match between the choice dispatch key and any branch key when parsing the parse.dat file.  You can say that c-daffodil shouldn't be more strict than scala-daffodil when unparsing (it reports a no match processing error on the modifed unparse.xml while scala-daffodil silently writes out the same modified parse.dat) but that's a different part of code than this one.

daffodil-runtime2/src/main/resources/c/libruntime/infoset.c
```diff
@@ -159,6 +159,13 @@ walkInfosetNode(const VisitEventHandler *handler, const InfosetBase *infoNode)
             error_msg =
                 handler->visitNumberElem(handler, childERD, numLocation);
             break;
+        case CHOICE:
+            if (!infoNode->erd->initChoice(infoNode, rootElement()))
```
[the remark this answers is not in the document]
mbeckerle: So, if a choice is an element, we put a typecode in the corresponding ERD and we have everything we need to cope with choices. 
But if a choice isn't an element, then.... we need a scheme that works for having more than one choice appearing in the sequence of a complex type element. 
One approach is anonymous (but uniquely named) Quasi-elements. These are elements with ERDs and all, but with a bit set in the ERD indicating that it's a quasi-element, and so when traversing, it is elided, so that the user (e.g., a path expression refering into it) can ignore its presence. This eliding becomes pervasive in path expressions, in traversal walks, etc. 
The other approach is inlining these, in which case we have multiple kinds of children - simple element, complex element, and choice-group. 
This latter approach is, honestly, more what people expect if they are creating C structs. Having to have a whole node for a choice,... in some sense the whole point of a choice is to avoid having a node. 
To the extent that elements of complex type as children get inlined into the enclosing parent structure, these two approaches may not be very different really. 

All this work is on a branch, so I'm not going to hold off on approving this change set to resolve this, so long as we recognize that it's not really reasonable for Runtime2's subset of DFDL to require every choice to be its own element in the long run.

What you have here is still good progress, and worth integrating.

daffodil-runtime2/src/main/resources/c/libruntime/infoset.c
```diff
@@ -160,10 +160,11 @@ walkInfosetNode(const VisitEventHandler *handler, const InfosetBase *infoNode)
                 handler->visitNumberElem(handler, childERD, numLocation);
             break;
         case CHOICE:
+            // Point next ERD to choice of alternative elements' ERDs
             if (!infoNode->erd->initChoice(infoNode, rootElement()))
             {
                 error_msg =
-                    "node's _choice field was not initialized during walk";
+                    "Walk error: no match between choice dispatch key and any branch key";
```
mbeckerle, line 167: This is another example of something that should be impossible. Getting this error means "runtime 2 has a bug". We should abort here.

daffodil-runtime2/src/main/resources/examples/NestedUnion.c
```diff
@@ -361,7 +361,8 @@ data_parseSelf(data *instance, PState *pstate)
         bar_parseSelf(&instance->bar, pstate);
         break;
     default:
-        pstate->error_msg = "node's _choice field was not initialized during parsing";
+        pstate->error_msg =
+            "Parse error: no match between choice dispatch key and any branch key";
```
mbeckerle, line 365: I think this should just abort(msg)

daffodil-runtime2/src/main/resources/examples/NestedUnion.c
```diff
@@ -379,7 +380,8 @@ data_unparseSelf(const data *instance, UState *ustate)
         bar_unparseSelf(&instance->bar, ustate);
         break;
     default:
-        ustate->error_msg = "node's _choice field was not initialized during unparsing";
+        ustate->error_msg =
+            "Unparse error: no match between choice dispatch key and any branch key";
```
mbeckerle, line 384: Another just abort() case.

mbeckerle (approved):
+1

daffodil-runtime2/src/main/resources/examples/NestedUnion.c
mbeckerle: Ah. I think I'm incorrect here. This is parse, so we're determining if there is a choice branch key match here. So this is a regular Processing/Parse error.

daffodil-runtime2/src/main/resources/examples/NestedUnion.c
mbeckerle: I think I'm wrong here also. This is unparse, I am not sure why we're looking at branch keys here though. Unparsing isn't symetric with parsing. It never needs to look at branch keys. The unparse problem is given an element in the infoset, what branch does that indicate exists. 
In Runtime1 it's called "choice branch resolver".

daffodil-runtime2/src/main/resources/c/libruntime/infoset.c
mbeckerle: This one, we're walking the infoset. So the parser has already dealt with the choice. At this point, if the choice dispatch key isn't one of the choice branch keys, then the parser didn't do its job. So I think this is an abort()/impossible situation that in theory one should not even be able to exercise.

tuxji: Actually, daffodil_main.c calls walkInfoset() in two places (from both the parse and unparse use cases).  You're right that when we parse a binary data file, the parseSelf() function has already dealt with the choice (initChoice() is just re-initializing the _choice member field at this time).  However, when we read an unparse XML file into an infoset in memory, we also walk the uninitialized infoset before we call unparseSelf() to write the unparsed binary data to the output file.  The walk validates each new XML element matches the schema's next infoset element and calls the correct number conversion function (float, integer, etc.).  The _choice member field is uninitialized at this time and I didn't know I could write a choice branch resolver which looks at the next XML element and figures out which branch does that element indicate exists inside the choice group.  I thought I would just call initChoice() from walkInfosetNode(), it would use the choice dispatch key to select the right branch, and then walkInfosetNode() would know what the next infoset element should be.  This behavior causes c-daffodil to behave more strictly than scala-daffodil; if the unparse XML file has no match between the choice dispatch key and any branch key, c-daffodil reports a no match processing error at this very place while scala-daffodil reports no error at all:

```shell
$ emacs unparse.xml # change <tag>4</tag> to <tag>0</tag>
$ make unparse-test
./daffodil unparse unparse.xml -o test_parse.dat
./daffodil: Walk error: no match between choice dispatch key and any branch key
make: *** [Makefile:46: unparse-test] Error 1
$ daffodil unparse -s nested.dfdl.xsd -r NestedUnion -o test_parse.dat unparse.xml # No errors
$ diff parse.dat test_parse.dat # No errors, 4th byte is zero in both files
```

Perhaps c-daffodil should load such an XML file without a no match processing error and unparse the infoset to a binary data file without a no match processing error.  We would have to code and call a choice branch resolver in C which peeks at the next XML element, figures out which branch does that element indicate exists inside the choice group, and initializes the choice and element runtime data (_choice and childNode->erd member fields) accordingly.  We probably would replace the initChoice() call in walkInfosetNode() with a call to that choice branch resolver and we might or might not call initChoice() in unparseSelf().  The reason I called initChoice() in all these places is that I thought we might want to remove the _choice member field and just call initChoice() to tell us the correct element to visit next in parse, walk, and unparse.

daffodil-runtime2/src/main/resources/examples/NestedUnion.c
tuxji: If we implement a choice branch resolver in c-runtime2, I think we would call it in walkInfosetNode() because the resolver needs to peek at the next XML element and walkInfosetNode() reads the XML file's data into infoset member fields.  We probably should call a getChoice() function in unparseSelf() instead of accessing the _choice member field in case we want to remove the _choice member field from C structs.

PR #490, opened 2021-02-11
Unsupported characterset

susmita-m145:
Jira- DAFFODIL-2049


susmita-m145:
Kindly review my changes

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/DaffodilCharsetProvider.scala
```diff
@@ -51,7 +51,8 @@ object DaffodilCharsetProvider {
     BitsCharsetUTF16LE,
     BitsCharsetUTF32BE,
     BitsCharsetUTF32LE,
-    BitsCharsetUTF8)
+    BitsCharsetUTF8,
+    BitsCharsetIBM1047 )
```
stevedlawrence, line 55: I would prefer that this be added to this list in alphabetical order so things are a bit more organized. So move this right after BitsCharsetIBM037

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package org.apache.daffodil.processors.charset
+
+import org.apache.daffodil.io.InputSourceDataInputStream
+import org.apache.daffodil.io.FormatInfo
+
+object BitsCharsetIBM1047 extends {
+  override val name = "IBM1047"
+  override val aliases = Seq("X-IBM-1047-S390")
```
stevedlawrence, line 25: Why the ``X-``? Can this just be ``IBM-1047-S390``?

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
```diff
@@ -0,0 +1,57 @@
...
+} with BitsCharsetJava {
+
+  val decodeStringTable =
+    "\u0000\u0001\u0002\u0003\u009C\u0009\u0086\u007F\u0097\u008D\u008E\u000B\u000C\u000D\u000E\u000F" +
+    "\u0010\u0011\u0012\u0013\u009D\u0085\u0008\u0087\u0018\u0019\u0092\u008F\u001C\u001D\u001E\u005E" +
```
stevedlawrence, line 30: According to https://www.compart.com/en/unicode/charsets/IBM1047, the ``\u005E`` at the end of this line is incorrect and should be ``\u001F``. I think everything else looks correct.

stevedlawrence (approved):
Thanks for submitting things. I suggest a couple minor changes

stevedlawrence:
It would also be really helpful to add a test so that this new code has some test coverage to make sure we don't accidentally break this in the future.

It looks like we have similar tests in ``daffodil-test/src/test/resources/org/apache/daffodil/section10/representation_properties/encodings.tdml``. The easiest way to add this test would be to:
1. Copy the ``eString`` element to make a new element with this new encoding
2. Copy the ``ebcidc1`` parserTestCase to create a new parserTestCase with a unique name and that uses your new element as the root

With that done, you just need to modify ``daffodil-test/src/test/scala/org/apache/daffodil/section10/representation_properties/TestRepProps2.scala`` and copy ``def test_ebcdic1`` to run your new parser test case.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
mbeckerle: Should not have "-s390" suffix. Name of this charset should just be "IBM-1047"

mbeckerle: To clarify, (maybe add this as a code comment for future developers). Charset names should be taken from: https://www.iana.org/assignments/character-sets/character-sets.xhtml

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
```diff
@@ -0,0 +1,57 @@
...
+    "\\\u00F7\u0053\u0054\u0055\u0056\u0057\u0058\u0059\u005A\u00B2\u00D4\u00D6\u00D2\u00D3\u00D5" +
+    "\u0030\u0031\u0032\u0033\u0034\u0035\u0036\u0037\u0038\u0039\u00B3\u00DB\u00DC\u00D9\u00DA\u009F"
+
+  override def newDecoder() = new BitsCharsetDecoderIBM1047()
+}
+
+class BitsCharsetDecoderIBM1047
+  extends BitsCharsetDecoderByteSize {
+
+  protected override def decodeOneChar(dis: InputSourceDataInputStream, finfo: FormatInfo): Char = {
```
mbeckerle, line 52: Suggest since the finfo argument is unused that it be named "unused" instead of "finfo". Makes it explicit that you intended to not use it, rather than you forgot something.

mbeckerle: Also, this method seems generic, and seems like it should be the default implementation, and every single-byte-character-set (SBCS) would share it. 

If the BitsCharsetDecoderByteSize had a constructor that took the BitsCharsetJava object, then is could be a one-liner:
```
class BitsCharsetDecoderIBM1047 extends BitsCharsetDecoderByteSize(BitsCharset1047)
```

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
mbeckerle: I think a unit test should parse and unparse a string containing every character, all 256 of them, perhaps in groups of 16 to match these lines of code. That would make it absolutely positive that there are no typos in these, at least for the characters that are not control-codes/whitespace.

mbeckerle (changes requested):
Small changes requested only. 

I think this is a good addition. Thank you for adding it.  

Once we add the last missing COBOL/Mainframe data features to Daffodil (JIRA DAFFODIL-853 may be the only one left we don't have), then I expect many other people will want us to add more EBCDIC/IBM charsets, as there are many variants.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
stevedlawrence: > Suggest since the finfo argument is unused that it be named "unused" instead of "finfo". Makes it explicit that you intended to not use it, rather than you forgot something.

Scala 2.13 adds an @unused annotation for this purpose. This way the variable name can be kept, which can sometimes be self documenting (though it doesn't add much in this case). This is also supposed to silence element unused warnings. We dont' get those right now, but maybe that's because we don't have that warning turned on.

It doesn't look like Scala 2.12 has anything similar. But maybe this is the kindof thing we just wait until we bump up to 2.13 to fix find all the unused parameters and annotate them correctly?

mbeckerle: I'm fine with that approach, we have a separate JIRA ticket about re-enabling warnings for things like unused params.

mbeckerle:
@susmita-m145 we're moving towards release 3.1.0 soon. I'd like to get this charset in. 
Can you fix the review-comments issues?

mbeckerle (approved):
+1. Thank you.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM1047.scala
```diff
@@ -0,0 +1,57 @@
...
+    "\u00D8\u0061\u0062\u0063\u0064\u0065\u0066\u0067\u0068\u0069\u00AB\u00BB\u00F0\u00FD\u00FE\u00B1" +
+    "\u00B0\u006A\u006B\u006C\u006D\u006E\u006F\u0070\u0071\u0072\u00AA\u00BA\u00E6\u00B8\u00C6\u00A4" +
+    "\u00B5\u007E\u0073\u0074\u0075\u0076\u0077\u0078\u0079\u007A\u00A1\u00BF\u00D0\u005B\u00DE\u00AE" +
+    "\u00AC\u00A3\u00A5\u00B7\u00A9\u00A7\u00B6\u00BC\u00BD\u00BE\u00DD\u00A8\u00AF\u005D\u00B4\u00D7" +
+    "\u007B\u0041\u0042\u0043\u0044\u0045\u0046\u0047\u0048\u0049\u00AD\u00F4\u00F6\u00F2\u00F3\u00F5" +
+    "\u007D\u004A\u004B\u004C\u004D\u004E\u004F\u0050\u0051\u0052\u00B9\u00FB\u00FC\u00F9\u00FA\u00FF" +
+    "\\\u00F7\u0053\u0054\u0055\u0056\u0057\u0058\u0059\u005A\u00B2\u00D4\u00D6\u00D2\u00D3\u00D5" +
+    "\u0030\u0031\u0032\u0033\u0034\u0035\u0036\u0037\u0038\u0039\u00B3\u00DB\u00DC\u00D9\u00DA\u009F"
+
+  override def newDecoder() = new BitsCharsetDecoderIBM1047()
```
mbeckerle, line 46: You have tests that use this charset. So the only way this could not be covered is if there is no driver executing those tests.  I don't see any scala test class that is invoking runOneTest("IBM1047") anywhere, so that's why this isn't covered. 

The tests need to be called from 
src/test/scala/org/apache/daffodil/section10/representation_properties/TestRepProps2.scala

mbeckerle (changes requested):
Missing test driver calls. That should fix CodeCov issues.

daffodil-test/src/test/scala/org/apache/daffodil/section10/representation_properties/encodings.tdml
```diff
@@ -0,0 +1,275 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
stevedlawrence, line 2: It looks like this file is a duplicate of the same file in src/test/resources, but with the addition of the sigle IBM1047 element and test case. The new element/parserTestCase should be moved to the encodings.tdml file in src/test/resources and this file deleted.

tuxji: I verified by comparison inside Emacs that this file not only duplicates `daffodil-test/src/test/resources/org/apache/daffodil/section10/representation_properties/encodings.tdml` but also has unwanted changes at line 54 (a stray backtick character was inserted) and lines 239 & 255 (a no-break space, \u00A0, was replaced with a regular space, \u0020).  @susmita-m145 , please delete this duplicate file and use an UTF8-compliant text editor to carefully insert your new lines into the original file without changing any other lines.

At line 55, insert:

```xml
  <xs:element name="eString1047" type="xs:string" dfdl:encoding="IBM-1047"/>
```

At line 119, insert:

```xml
  <tdml:parserTestCase name="IBM1047" root="eString1047" model="enc1" 
    description="IBM-1047 basic test">
    <tdml:document>
      <tdml:documentPart type="text" encoding="IBM-1047">12345</tdml:documentPart>
    </tdml:document>
    <tdml:infoset>
      <tdml:dfdlInfoset>
        <tns:eString>12345</tns:eString>
      </tdml:dfdlInfoset>
    </tdml:infoset>
  </tdml:parserTestCase>
```

Then please run `sbt test` and verify that the tests pass before you push the new commit.

tuxji (changes requested):
I think this PR needs only one more change to pass the tests.  Please delete a duplicate file and edit the original file instead.

daffodil-test/src/test/scala/org/apache/daffodil/section10/representation_properties/encodings.tdml
mbeckerle: Wow, I'm not the only guy who uses emacs? How great is that?  :-) 

@susmita-m145  This team is *really good* and I depend on this team as well as our automated continuous-integration runs, to catch things. I have fallen into the trap of running all the tests, seeing success, then trying to slide in one more little inconsequential change, like just updating a code-comment, and then pushing it, only to find *blam* I made a typo and things are broken. 

As you are one of the people working on EBCDIC/IBM stuff, I am also very interested in your use case for Daffodil and the priority you give to finishing the tickets like the P, V text number pattern features (for COBOL data) etc. 

I am hoping you have a bit more time to push this PR over the finish line. Thanks -mikeb

mbeckerle:
Charsets are now pluggable as of daffodil git hash c81b213b02ac414a39290787cf2eea14d83fdc26, so this can be done as a dynamically loaded charset extension now. 

I'm not at all opposed to building this into the charsets that come with Daffodil. I just want to point out that the sands have shifted a little under this.

mbeckerle:
Closing as inactive. The ticket points back to this work if useful.

PR #493, opened 2021-02-24
Add debugger "info suspensions" and "set diffExcludes" commands

stevedlawrence:
- Add new trait for info commands that are Seq's rather than simple
  values, implementing act() and diff() functions
- Use this new trait to create a new "info suspensions" command to list
  the current suspensions. For example:

      (debug) info suspensions
        suspensions:
          SuppressableSeparatorUnparserSuspendableOperation for title
          RegionSplitSuspendableOperation for title
          SuppressableSeparatorUnparserSuspendableOperation for title
      (debug) info diff
        suspensions:
          + SuppressableSeparatorUnparserSuspendableOperation for title

- Add a new "set diffExcludes" command to exclude info commands from
  showing up in "info diff". This can be useful if there are certain
  diffs that happen all the time that the user doesn't care about. For
  example:

      (debug) set diffExcludes bitPosition bitLimit suspensions

- Skip debug steps during suspensions. This currently causes an abort
  (some suspension state should never be accessed), but also can cause
  confusing during debugging, especially related to diffs, since
  Daffodil temporarily time travel when evaluating suspensions. More
  thought is needed to figure out best how to debug suspensions so the
  time travel aspect is obvious and understandable.

DAFFODIL-758

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
@@ -43,6 +43,7 @@ import org.apache.daffodil.processors._
 import org.apache.daffodil.processors.parsers.ConvertTextCombinatorParser
 import org.apache.daffodil.processors.parsers._
 import org.apache.daffodil.processors.unparsers.UState
+import org.apache.daffodil.processors.unparsers.UStateForSuspension
 import org.apache.daffodil.processors.unparsers.Unparser
 import org.apache.daffodil.schema.annotation.props.gen.Representation
 import org.apache.daffodil.util.DPathUtil
@@ -118,6 +119,9 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
     /* whether to remove hidden elements when displaying the infoset */
     var removeHidden: Boolean = false
 
+    /* list of info commands to exclude when running 'info diff' */
+    var diffExcludes: Seq[String] = Seq.empty
+
     /* stores the last actual command (i.e. not a "") that was executed */
     var lastCommand: String = ""
 
@@ -158,6 +162,11 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
   }
 
   def debugStep(state: ParseOrUnparseState, processor: Processor, ignoreBreakpoints: Boolean): Unit = {
+      // ignore debug steps called during suspensions, those must be handled differently
+      if (state.isInstanceOf[UStateForSuspension]) {
+        return
+      }
+
       debugState = debugState match {
         case _ if ((state.processorStatus ne Success) && DebuggerConfig.breakOnFailure) => DebugState.Pause
         case DebugState.Continue | DebugState.Trace if !ignoreBreakpoints => {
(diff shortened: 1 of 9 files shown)


mbeckerle (approved):
+1
There is another place besides "info sus" which needs to display the suspensions readably, which is when unparsing fails with a suspensions deadlock. 

Will these changes also change that? 

I have a suggestion as well. Perhaps we want to call a method like toStringForDebugger which defaults to just toString, but can be overridden to do something like convert SuppressableSeparatorSuspendedOperation into "Suppressable Separator".  Right now the class names of the suspension objects are really long, and some words are in all such like "SuspendedOperation" or "Suspension" which add nothing but clutter. 

This could come later however.

stevedlawrence:
This doesn't change how suspension deadlocks are displayed, only let's you see what suspensions are active during an unparse. All that deadlock output logic is unchanged. Yeah, a toStringForDebugger, or something similar, definitely seems needed to make this output a big more useful. What we currently have is useful to get an idea of what's going on, but if things are going wrong, and things like suspensions are blocked that you wouldn't expect to be, or for longer than you would expect, the current output doesn't help with that. This definitely isn't the last change for suspension info output.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1842,6 +1908,32 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
         }
       }
 
+      object SetDiffExcludes extends DebugCommand {
+        val name = "diffExcludes"
+        val desc = "set info commands to exlude in the 'info diff' commanad"
```
tuxji, line 1913: Fix typo in "exlude".

tuxji (approved):
+1

Looks good.

I didn't see any code coverage report; does it show up only if a new line isn't covered by any test?

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
stevedlawrence: Will fix.

Codecov seems a bit flakey to me. Sometimes comments/inline code shows up, sometimes they don't. And looking at the pull request there are a few lines that aren't covered:

https://app.codecov.io/gh/apache/daffodil/compare/493/diff

I'll see if I can tweak this test or add a new one to increase the coverage.

tuxji: Hmm, I couldn't view any report data in your 493 link (the page said Unauthorized Github API: Forbidden Check on Codecov's status or see our docs for common support: Error 403).  I tried a report link from one of the build jobs' logs but the diff tab said "No files changed" so I couldn't see per-line coverage there either.  Not sure what's going on.  Here's the report link I tried.

https://codecov.io/gh/apache/daffodil/commit/ba2d93f012bdff346c6f380a180c7ea3253bb5fb

stevedlawrence:
You might need to be logged in to codecov (proably via their Github auth method) for those links to work? I don't know why that's required but if I go to incognito mode or log out I get the same errors you see.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
tuxji: Yes, you're right, I had to log into codecov on their home page.  Now both links work and show me line by line coverage changes although one aggregates all build jobs and the other is from only one build job.

PR #495, opened 2021-03-04
Patterns to match strings with newlines.

mbeckerle:
Tests added illustrating this.

Diagnostic messages about pattern facets improved
to include things like &#xE000; when that's part of the pattern facet.

Some cleanup of XMLUtils which was unclear to me whether it was doing the right thing. (It was)

DAFFODIL-2474 

https://issues.apache.org/jira/browse/DAFFODIL-2474

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
@@ -44,7 +44,8 @@ trait Facets { self: Restriction =>
 
   private def retrieveFacetValuesFromRestrictionBase(xml: Node, facetName: Facet.Type): Seq[String] = {
     val res = xml \\ "restriction" \ facetName.toString() \\ "@value"
-    if (res.length > 0) res.map(n => n.text).toList else List.empty
+    val ret = if (res.length > 0) res.map(n => n.text).toList else List.empty
+    ret
   }
   private def enumeration(xml: Node): Seq[String] = { retrieveFacetValuesFromRestrictionBase(xml, Facet.enumeration) }
   private def fractionDigits(xml: Node): String = { retrieveFacetValueFromRestrictionBase(xml, Facet.fractionDigits) }
@@ -56,7 +57,8 @@ trait Facets { self: Restriction =>
   private def minLength(xml: Node): String = { retrieveFacetValueFromRestrictionBase(xml, Facet.minLength) }
   private def pattern(xml: Node): Seq[String] = {
     // Patterns are OR'd locally, AND'd remotely
-    retrieveFacetValuesFromRestrictionBase(xml, Facet.pattern).map(p => p)
+    val res = retrieveFacetValuesFromRestrictionBase(xml, Facet.pattern).map(p => p)
+    res
   }
   private def totalDigits(xml: Node): String = { retrieveFacetValueFromRestrictionBase(xml, Facet.totalDigits) }
   private def whitespace(xml: Node): String = { retrieveFacetValueFromRestrictionBase(xml, Facet.whiteSpace) }
@@ -128,10 +130,10 @@ trait Facets { self: Restriction =>
   final lazy val hasFractionDigits: Boolean = (localFractionDigitsValue != "") || (getRemoteFacetValues(Facet.fractionDigits).size > 0)
 
   final lazy val patternValues: Seq[FacetValueR] = {
-    val values = combinedBaseFacets.filter { case (f, _) => f == Facet.pattern }
+    val values: Seq[(Facet.Type, Values)] = combinedBaseFacets.filter { case (f, _) => f == Facet.pattern }
     if (values.size > 0) {
       val res: Seq[FacetValueR] = values.map {
-        case (f, v) => {
+        case (f, v: String) => {
           //
           // The DFDL Infoset can contain strings which hold characters
           // that are not allowed in XML at all.
@@ -160,7 +162,7 @@ trait Facets { self: Restriction =>
           // The XSD numeric character entity &#xE000; can be used to match ASCII NUL
           // (char code 0).
           //
-          val remapped = XMLUtils.remapPUAToXMLIllegalCharacters(v)
+          val remapped: String = XMLUtils.remapPUAToXMLIllegalCharacters(v)
           (f, remapped.r)
         }
       }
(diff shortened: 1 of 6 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -702,6 +706,17 @@ object XMLUtils {
...
+    // #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
+    i match {
+      case 0x9 | 0xA | 0xD => false
+      case z if (z >= 0x20 && z <= 0xD7FF) => false
+      case z if (z >= 0xE000 && z <= 0xFFFD ) => false
+      case z if (z >= 0x10000 && z <= 0x10FFFF) => false
+      case _ => true
+    }
+  }
+
```
stevedlawrence, line 719: Doesn't look like this is used anywhere? Is this jus a helpful tool that we may need in the future, or was the intentino to may replace some of the above code with it?

Might also be worth modifying the comment to say "legal characters: ..." or something. Since the function is called isXMLIllegalChars I assumed that comment was the list of illegal charactes and was a bit confused.

Also, just an FYI, I was curious to see how this match/catch compiled to bytecode compared to if-statements, and they end up exactly the same. Not too surprised, but good to know.

stevedlawrence (approved):
=1

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
```diff
@@ -160,7 +162,7 @@ trait Facets { self: Restriction =>
           // The XSD numeric character entity &#xE000; can be used to match ASCII NUL
           // (char code 0).
           //
-          val remapped = XMLUtils.remapPUAToXMLIllegalCharacters(v)
+          val remapped: String = XMLUtils.remapPUAToXMLIllegalCharacters(v)
```
tuxji, line 165: I'm curious about these explicit types (lines 133, 136, 165).  What was the reason you added these explicit types?  I thought the Scala compiler inferred types for local values fairly well although sometimes IntelliJ IDEA gets a little confused itself (is that why?).

daffodil-test/src/test/resources/org/apache/daffodil/section05/facets/PatternRanges.tdml
```diff
@@ -0,0 +1,123 @@
...
+        Using E000 remapping for XML illegal characters 0E to 1F
+        20 to 7F are legal.
+        -->
+        <!--
+          We cannot depend on Daffodil's mapping from E009 to 9, E00A to A, and E00D to D.
+          Because that won't work in Xerces with "full" validation.
+          We can't use numeric entities for the tab, LF, or CR, because those aren't allowed in attribute
+          values. So we use \t, \n, and \r for those control characters.
+          -->
+        <xs:pattern value="[&#xE000;-&#xE008;\t\n&#xE00B;&#xE00C;\r&#xE00E;-&#xE01F;&#x20;-&#x7F;]*"/>
```
tuxji, line 50: You have two redundant comments here - did you plan to keep both comments or merge them?

daffodil-test/src/test/resources/org/apache/daffodil/section05/facets/PatternRanges.tdml
```diff
@@ -0,0 +1,123 @@
...
+      </xs:restriction>
+    </xs:simpleType>
+
+
+    <xs:element name="str" type="ex:singleByteCharsWithCodepointLessThan1F"
+       dfdl:lengthKind="delimited"/>
```
tuxji, line 56: I was a bit puzzled at seeing dfdl:lengthKind `delimited` on a solitary root element until I read the DFDL spec and realized the end of the data stream delimits an element as well.  However, the spec says we also can use the dfdl:lengthKind `endOfParent` to allow the last element to consume data up to the end of the data stream; might `endOfParent` be more self-documenting too?

daffodil-test/src/test/scala/org/apache/daffodil/section05/facets/TestPatternRanges.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  @AfterClass def shutDown = {
+    runner.reset
+  }
+}
+
+import org.apache.daffodil.section05.facets.TestPatternRanges._
+
+class TestPatternRanges {
+
```
tuxji, line 34: Most TDML tests would put a relative import at this line instead of a absolute-package import at line 31.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
tuxji: Yes, I'm curious too.

tuxji (approved):
+1

Looks good, only curious about a couple things.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Facets.scala
mbeckerle: Yes, I needed these in order to figure out what was going on here. Type aliases were confusing IDEA. I tend to leave in that sort of thing when it's needed to improve understanding, why take it back out.

daffodil-test/src/test/resources/org/apache/daffodil/section05/facets/PatternRanges.tdml
mbeckerle: I'll take a look and fix.

daffodil-test/src/test/resources/org/apache/daffodil/section05/facets/PatternRanges.tdml
mbeckerle: Yes, but endOfParent isn't supported yet by Daffodil.

mbeckerle: I'll add a comment about this not having a delimiter, so uses end-of-file.

daffodil-test/src/test/scala/org/apache/daffodil/section05/facets/TestPatternRanges.scala
mbeckerle: I'll hoist this in.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
mbeckerle: I will remove this.

This is me once again not remembering that I already wrote something long ago. I started writing it again, then found what I needed. 

I may add some scaladoc to these escapers, pua mappers, etc. to explain their usage, what they do (and don't do).

PR #510, opened 2021-03-15
New variable instances should inherit external values

jadams-tresys:
There was a bug where variables that were given external values were not
having those values passed on by default to newVariableInstance's. This
commit addresses this issue by setting NVI's to the externally provided
value by default if no other default value is specified for the NVI.

DAFFODIL-2481


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -482,6 +474,15 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
     val v = vrd.createVariableInstance()
+
+    // If the variable is external, the original variable instance should have
+    // already been set to the externally provided value. Use this value as the
+    // default value
+    if (vrd.external && !vrd.hasLocalDefault) {
+      val value = variableInstances.get.last.externalValue
+      v.externalValue = value
+      v.setValue(value)
+    }
```
stevedlawrence, line 485: This sort of splits the NVI logic into two separate places, which I think makes it a bit harder to reason about. Part is here, and part is the actual parser/unparse? Thoughts on moving this logic into to NVI parser/unparse so it's all consolidated in one place?

Also, thoughts on changing the ``externalValue`` to be ``maybeExternalValue``, and anytime we create a new variable instance we just copy that value from the previous instance to the new instance (you may already do this). This way, we always have the external value if we need it, and you never need to look at previous onces. So the NVI logic just becomes something like

```scala
if (needToEvaluateDefaultExpression) {
  // create suspension to set default value
} else if (v.maybeExternalValue.isDefined) {
  v.setValue(v.maybeExerternalValue.get)
} else {
  // no default anywhere and no external value, new instance remains undefined
}
```

This way, if logic about NVI's ever changes, we only have to change the one parser/unparser and don't also have to think about what other places might deal with external values/defaultExpressions/state/etc

Also note that this logic doesn't care if a variable is external or not. Only setExtVariable should care about that. Everything else can just assume if it has an external variable then it was allowed to be set.

daffodil-test/src/test/scala/org/apache/daffodil/section07/external_variables/TestExternalVariables.scala
```diff
@@ -56,6 +56,10 @@ class TestExternalVariables {
     runner.runOneTest("override_define_vars_05")
   }
 
+  @Test def test_override_define_vars_06(): Unit = {
+    runner.runOneTest("override_define_vars_06")
```
stevedlawrence, line 60: Can we also add this same test  but with the config no set so we test the logic that if a variable is external, but no external variable is set then it uses the default value from define variable?

Might also be worth adding the test I mentioned in a comment to ensure we have the right logic for whether to evaluate a defineVariable defaultValue, or just use the value it evaluated to when parse/unaprse started.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -953,6 +953,7 @@ final class VariableRuntimeData(
   pathArg: String,
   namespacesArg: NamespaceBinding,
   val external: Boolean,
+  val hasLocalDefault: Boolean,
   val direction: VariableDirection,
   @TransientParam maybeDefaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
```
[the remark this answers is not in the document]
jadams-tresys: You interpreted the situation correctly. maybeDefaultValueExprArg will inherit from the variable definition if there isn't one local to the newVariableInstance.  I can make the name change to try to clear this up a little more.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/DFDLDefineVariable.scala
```diff
@@ -137,6 +140,11 @@ final class DFDLNewVariableInstance(node: Node, decl: AnnotatedSchemaComponent)
     case (Some(str), v) => schemaDefinitionError("Default value of variable was supplied both as attribute and element value: %s", node.toString)
   }
 
+  final lazy val hasLocalDefault = (defaultValueAsAttribute, defaultValueAsElement) match {
+    case (None, "") => false
+    case _ => true
+  }
```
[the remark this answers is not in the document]
jadams-tresys: That's a good idea. I hadn't thought about treating it/returning a tuple, which is why I had it separate. Will make the change.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -104,7 +104,10 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
 
   override def unparse(state: UState) = {
     state.variableMap.newVariableInstance(context)
-    if (context.maybeDefaultValueExpr.isDefined) {
+
+    // Only process the default value expression if it is defined and only if
+    // the variable is non-external or has a local default
+    if (context.maybeDefaultValueExpr.isDefined && (!context.external || context.hasLocalDefault)) {
```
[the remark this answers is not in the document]
jadams-tresys: You are right, I missed that logic bug there.  Will also add a test to check for this situation.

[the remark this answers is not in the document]
mbeckerle: I think given an NVI without a default value, and a defineVariable with an expression for the default value, the assumption should be that the defineVariable default value is computed once, at start of processing, and that value is reused. The expression would not be re-evaluated when the NVI instance is created. 

Suppose your var2 was also external="true", so that the default value can be overridden externally. In that case the default value expression would never be evaluated. 

There is no place in DFDL where "inherit the expression" semantics is used I think. 

E.g., suppose you use a property from a dfdl:format default format and there is an expression used there. That expression is evaluated once only. Not every time that property is referenced. If that expression references variables then the values of those variables are only the global values of them. 

The "inherit the expression" semantics can be obtained by creating new variable instances of both your var1 and var2 where var2 has its own default value expression of "{ $var1 }".  

`<dfdl:newVariableInstance ref="var1" />
<dfd:setVariable ref="var1" value="Y" />
<dfdl:newVariableInstance ref="var2" defaultValue="{ $var1 }"/>`

Here var2's default value would be Y.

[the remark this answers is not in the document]
stevedlawrence: Seems a reasonable argument to me.

So in that case, a decent amount of this changes. I think we no longer need the isLocalDefault because the existance of maybeDefaultValueExpr gives us that. And we now need to capture the original defineVariable value (whether it be set externally or via a defaultValue expression) and make that available to NVIs for use when then maybeDefaultValueExpr isn't set.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -104,7 +104,10 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
 
   override def unparse(state: UState) = {
     state.variableMap.newVariableInstance(context)
-    if (context.maybeDefaultValueExpr.isDefined) {
+
+    // Only process the default value expression if it is defined and only if
+    // the variable is non-external or has a local default
```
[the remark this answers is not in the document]
mbeckerle: This comment is actually incorrect I think. For a newVariableInstance, we should always process the defaultValue expression if it is provided and the direction includes unparsing. If it is not provided, then the default value is that of the global defineVariable, which would have already been evaluated (at processor start time), or provided externally at start of processing for the global variable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -97,7 +98,8 @@ class VariableInstance private (
   val rd: VariableRuntimeData,
   @TransientParam defaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
   var priorState: VariableState = VariableUndefined,
-  var priorValue: DataValuePrimitiveNullable = DataValue.NoValue)
+  var priorValue: DataValuePrimitiveNullable = DataValue.NoValue,
```
mbeckerle, line 101: priorValue... I'm struggling with this. Why do we need that? When we create a NVI, the default value does not use that of an enclosing NVI. The DFDL spec says "The tuple is initialized based on the specifics of the dfdl:defineVariable annotation." The "tuple" being the state of a variable instance. So why do we need priorValue?

daffodil-test/src/test/resources/org/apache/daffodil/section07/external_variables/external_variables.tdml
```diff
@@ -338,9 +338,72 @@
...
 
+			</tdml:dfdlInfoset>
+		</tdml:infoset>
+  </tdml:parserTestCase>
+
+	<tdml:defineSchema name="v5">
+		<dfdl:defineVariable name="v_with_default" type="xsd:int"
+			defaultValue="1" external="true" />
+		<xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" />
```
mbeckerle, line 349: Indentation

mbeckerle (changes requested):
Added a question on top of the issues SL observed, about scoping of default values of variables. Just want to be sure we're taking the defaults from a NVI defaultValue expression, or the default value of the define-variable, not an enclosing NVI's value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Prior value is used in one place only right now.  When we are backtracking from a point of uncertainty where a variable was changed, when the PState is reset we also reset all the changed variables back to their previous state/value.

mbeckerle: Ok. Thanks

daffodil-test/src/test/resources/org/apache/daffodil/section07/external_variables/external_variables.tdml
jadams-tresys: Darn spaces vs. tabs

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -148,7 +154,11 @@ class VariableInstance private (
     this._value = Some(v)
   }
 
-  def setDefaultValue(v: DataValuePrimitiveNullable) = this._value = Some(v)
+  /* This is used to set a default value without assigning a priorValue */
```
stevedlawrence, line 157: I noticed forceExpressionEvaluations uses setValue. Should it be using setDefaultValue? Or is this something that should realy only be used at runtime?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -465,12 +488,12 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
    */
-  def newVariableInstance(vrd: VariableRuntimeData, defaultValue: DataValuePrimitive) = {
+  def newVariableInstance(vrd: VariableRuntimeData, defaultValue: DataValuePrimitiveNullable) = {
     val varQName = vrd.globalQName
     val variableInstances = vTable.get(varQName)
     Assert.invariant(variableInstances.isDefined)
 
-    variableInstances.get += vrd.createVariableInstance(VariableUtils.convert(defaultValue.getAnyRef.toString, vrd, vrd))
+    variableInstances.get += vrd.createVariableInstance(defaultValue)
   }
```
stevedlawrence, line 497: I would actually prefer we drop this function all together. If I want to find all places where we set a defaultValue, I want to be able to grep for ``setDefaultValue``. Having to also grep for newVarialbeInstance where we pass in an extra paramter is more difficult. Is there any reason why uses of this can't be replaced with
```scala
val v = newVariableInstance(...)
v.setDefaultValue(...)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -547,7 +569,9 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         }
 
         case _ => {
-          variable.setValue(VariableUtils.convert(newValue.getAnyRef.toString, variable.rd, referringContext))
+          val value = VariableUtils.convert(newValue.getAnyRef.toString, variable.rd, referringContext)
+          variable.externalValue = value
```
stevedlawrence, line 573: I think this ``variable.externalValue = value`` goes away if the logic I mentioned previously makes sense to switch to. Again, one less place where we need to remember to change state.

stevedlawrence (commented):
Some thoughts on slight tweaks that I *think* might make it more easy to reason about state changes. Thoughts?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -103,17 +103,23 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
+    // If the NVI has a default value expression, process it to replace the
+    // default value inherited from the global defineVariable statement.
+    if (context.maybeDefaultValueExpr.isDefined) {
+      v.setState(VariableInProcess)
       val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
       suspendableExpression.run(state)
     }
+
+    if (v.value != DataValue.NoValue)
+      v.setState(VariableDefined)
```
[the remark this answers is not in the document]
jadams-tresys: You are right, the state should have been set when we call setDefaultValue on the variable instance. Will make the change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: You're right, forceExpressionEvaluations should be using setDefaultValue.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Makes sense to me, will change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -105,6 +105,9 @@ class VariableInstance private (
 
   lazy val defaultValueExpr = defaultValueExprArg
 
+  // This represents the default value at the start of processing, provided by
+  // either the defaultValue expression or by an external binding
+  var defaultValue: DataValuePrimitiveNullable = DataValue.NoValue
```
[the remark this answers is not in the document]
jadams-tresys: I think my most recent push addresses these changes, but it still feels a little awkward copying around the firstInstanceInitialValue as a NVI doesn't start with a copy of the original variable instance.  Let me know what you think.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -328,10 +337,11 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         // Evaluate defineVariable statements with non-constant default value expressions
         case (VariableDefined, DataValue.NoValue, true) => {
           val res = inst.defaultValueExpr.get.evaluate(state)
-          inst.setValue(DataValue.unsafeFromAnyRef(res))
+          inst.setDefaultValue(DataValue.unsafeFromAnyRef(res))
         }
         case (_, _, _) => // Do nothing
       }
+      inst.firstInstanceInitialValue = inst.value
```
stevedlawrence, line 344: Personally, I'd rather see a function called after ``forceExprssionEvaluations`` is called that does this copy. That way there is only a single place where this variable is changed and it's easier to verify it's happening at the right time. Functionally it's exactly the same, but I think eaiser to read and understand what's going on.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -462,27 +484,16 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
    */
   def newVariableInstance(vrd: VariableRuntimeData) = {
     val varQName = vrd.globalQName
     val variableInstances = vTable.get(varQName)
     Assert.invariant(variableInstances.isDefined)
-
-    val v = vrd.createVariableInstance()
-    variableInstances.get += v
+    val nvi = vrd.createVariableInstance()
+    nvi.firstInstanceInitialValue = getFirstInstance(varQName).get.firstInstanceInitialValue
```
stevedlawrence, line 495: Note that this doesn't necessarily need to come from the first instance. Since each instance gets a copy from the first, we can just copy the value from any previous instance. But could this just be
```scala
nvi.firstInstanceInitialValue = variableInstance(0).firstInstanceInitialValue
```
We don't really need ``getFirstInstance`` which is duplicating some effort, like the vtable lookup.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -166,10 +166,16 @@ class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
...
+    val v = start.variableMap.find(context.globalQName).getOrElse(Assert.invariantFailed("Unable to find variable %s".format(context.globalQName)))
+
     if (context.maybeDefaultValueExpr.isDefined) {
-      val v = start.variableMap.find(context.globalQName).get
       val res = DataValue.unsafeFromAnyRef(context.maybeDefaultValueExpr.get.evaluate(start))
       v.setDefaultValue(res)
+    } else {
+      // The NVI will inherit the default value of the original variable instance
+      // This will also inherit any externally provided bindings.
+      v.setDefaultValue(v.firstInstanceInitialValue)
```
stevedlawrence, line 178: Same comments as unparse.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
stevedlawrence: I agree, the copying of the ``fisrtInstanceInitialValue`` thing does seem a bit weird. What I was trying to avoid by suggesting this was a member variable that only has a value in the first instance. If we didn't do this copy and someone accidentally accessed this member from an instance other than the first one, then they'll probably get a null value or some unexpected behavior. By copying the value from the first instance to all following instances, it doesn't matter from which instance you access this value, you will always get the initial value of the first instance.

Maybe a way to avoid that would be do have another class, e.g.
```scala
class InitialVariableInstance extends VariableInstance {
  var initialValue = ...
}
```
This way only the ``IntialVariableInstance`` has this initialValue variable, and so you can never accidentally access it from some other instance, because other instances don't even have this variable. I'm not sure if that's worth it though. You end up needing something like this to get the initial value:
```scala
instances(0).asInstanceOf[InitialVariableInstance].initialValue
```

I'm sure we could come up with additionl member variables/functions to get around this and make it look cleaner, but I'm not convinced it's entirely worth it for just one member.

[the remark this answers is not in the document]
stevedlawrence: Note, it would probably be worth added a comment to this affect, explaing why we have this initial thing and why we are copying. Assuming it stays and we don't come up with a better alternative.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -104,12 +102,17 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
-      val maybeVar = state.variableMap.find(context.globalQName)
-      Assert.invariant(maybeVar.isDefined)
-      maybeVar.get.setState(VariableInProcess)
+      v.setState(VariableInProcess)
       val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
       suspendableExpression.run(state)
+    } else {
+      // The NVI will inherit the default value of the original variable instance
+      // This will also inherit any externally provided bindings.
+      v.setDefaultValue(v.firstInstanceInitialValue)
```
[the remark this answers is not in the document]
jadams-tresys: I'm pretty sure that a variable instance with value=NoValue and state=VariableDefined is normal and expected for a defineVariable without a default value.  I believe only time a variable's state should be undefined is when the variable is first defined it's priorState is set to VariableUndefined. This comes into play when we are resetting a variable back to priorState/Value and we check to make sure that the variable is no undefined.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -84,10 +84,8 @@ final class NewVariableInstanceSuspendableExpression(
   extends SuspendableExpression {
 
   override protected def processExpressionResult(ustate: UState, v: DataValuePrimitive): Unit = {
-    val vi = ustate.variableMap.find(rd.globalQName)
-    Assert.invariant(vi.isDefined)
-    vi.get.setState(VariableDefined)
-    vi.get.setValue(v)
+    val vi = ustate.variableMap.find(rd.globalQName).getOrElse(Assert.invariantFailed("Unable to find variable %s".format(rd.globalQName)))
+    vi.setDefaultValue(v) // This also sets variable state to VariableDefined
```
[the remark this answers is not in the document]
jadams-tresys: I think a little more verbosity is fine. Change made.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -1,4 +1,5 @@
-/*
+/* This represents the default value at the start of processing, provided by
+ * either the defaultValue expression or by an external binding
  * Licensed to the Apache Software Foundation (ASF) under one or more
```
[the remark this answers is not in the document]
jadams-tresys: As much as I love vim, it makes it easy for an inadvertent keystroke to do something like this. Missed it when I was reviewing the diff.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -103,6 +104,9 @@ class VariableInstance private (
 
   lazy val defaultValueExpr = defaultValueExprArg
 
+  // This represents the default value at the start of processing, provided by
```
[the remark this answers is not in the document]
jadams-tresys: There are some instances where we are using the defaultValueExpression in places without access to a variable runtime data. For example, we check to see if a variable has a defaultValueExpression defined when resetting the instance to determine the correct state to reset to. We also evaluate the expressions in forceExpressionEvaluations without access to a VRD.

It may be possible to restructure these things, but I'm not sure its worth the effort.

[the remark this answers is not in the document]
stevedlawrence: Yeah, I think you're right that there are places where we only have access to the instance.

That said, I think anything that can simplify this variable stuff is worth the effort. It's taken me multiple reviews to get an idea of how it all works, and I still don't understand it all (that isn't saying anything about yoru code, this stuff is just complicated).

For example, I noticed that each ``VariableInstance`` already has a pointer to its VariableRuntimeData, so could we remove these defaultValueExpr/Arg members and access the expression on the RuntimeData when needed?

Also, I noticed in the VRD that it stores state and value depending on if the if there's a defaultValue expression, and then that is used to initialize the VariableInstance. This feels unnecessary and wrong to me. We shouldnt' be creating a VariableInstance with some state based on whether or not it has a defaultValue, since things like external values also play a role. It seems like we should always create an empty VariableInstance with no value/state, and the all the new you've added should be setting default value/state as appropriate. In fact, my guess is this is where the value=NoValue but state=VariableDefined comes from.

All that said, I wonder if instead, a VariableInstance should take no parameters except for VariableRuntimeData. So the value, state, priorValue, priorState, and defaultValue become member variables. When we create a VariableInstance, it's value is initialize to None and state to VariableUndefined. Then all the logic you've added will get triggered during parse to figure out default values and set state appropriately.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -328,10 +337,11 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
         // Evaluate defineVariable statements with non-constant default value expressions
         case (VariableDefined, DataValue.NoValue, true) => {
           val res = inst.defaultValueExpr.get.evaluate(state)
-          inst.setValue(DataValue.unsafeFromAnyRef(res))
+          inst.setDefaultValue(DataValue.unsafeFromAnyRef(res))
```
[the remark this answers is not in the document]
jadams-tresys: So VariableDefined with value=NoValue is end result of a defineVariable statement without a default value expression.  In this instance, the variable is defined (hence the defineVariable statement) but without a value.  I believe VariableUndefined is only used/checked when resetting a variable we check to make sure it is not undefined as that would mean we are trying to reset a variable that has already been reset.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: You are right that we don't need getFirstInstance anymore.  I needed it before when I was setting the initial default from the NVI parser, but that has since been refactored.

That being said, I'm not sure you realized that newVariableInstance does not copy any data from the initial variable instance, so this copy is necessary.

stevedlawrence: If we don't do the copy, then we get variables that have unset values--that feels dangerous to me. I'd rather have variables that we never use but have the right value, rather than variables that we might accientally use and have the wrong value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -117,6 +121,7 @@ class VariableInstance private (
   def value = {
     if (_value.isEmpty) {
       _value = Some(valueArg)
+      firstInstanceInitialValue = valueArg
```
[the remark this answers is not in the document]
jadams-tresys: You are right, this isn't needed.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -104,12 +102,17 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
 
   override def unparse(state: UState) = {
     state.variableMap.newVariableInstance(context)
+
+    val v = state.variableMap.find(context.globalQName).getOrElse(Assert.invariantFailed("Unable to find variable %s".format(context.globalQName)))
```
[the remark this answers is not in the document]
jadams-tresys: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
stevedlawrence: That feels like a bug to me, but maybe I just don't understand what the different states mean (we need to document them). My understanding of the states is:

<dl>
<dt>VariableUndefined</dt>
<dd>The variable has no value, a variable read is an error, setVariable is allowed</dd>
<dt>VariableBeingDefined</dt>
<dd>We are in the process of evaluting and setting the value to a default value, used for circularity detection of defineVariable defaulValue's</dd>
<dt>VariableDefined</dt>
<dd>The variable has a value (either default or externally set), a variable read of that value is allowed, set variable is allowed</dd>
<dt>VariableSet</dt>
<dd>the variable has a value that has been set via setVariable, variable read is allowed, setVariable is no longer allowed</dd>
<dt>VariableRead</dt>
<dd>the variable has been read with a value, that value is now set, variable read is allowed, setVariable is no longer allowed</dd>
<dt>VariableInProcess</dt>
<dd>A suspension to set a variable (via either defaultValue or setVariable?) during unparsing has been created. This variable does not yet have a value. All reads/sets of this variable should block until this state changes.
</dl>

This is my understanding, but it sounds like that's not the case?

Note that another comment I made suggested a change to so that VariableInstance's aren't initialize with NoValue, maybe that would affect some of this?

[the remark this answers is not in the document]
jadams-tresys: Honestly, that is probably what the different states should be (only VariableDefined/Undefined are different than what you have). I don't think I've changed the current usage patterns of any of these states (other than adding VariableBeingDefined) while I've been making changes to variables, but it would probably be good to add clear documentation on what each state means and make the change to VariableDefined/Undefined.  I don't think it will take much and shouldn't break anything.

I'll take a look at it tomorrow and try to add it to this commit, assuming it doesn't break things or require a large refactor.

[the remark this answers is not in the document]
jadams-tresys: After taking another look at this, I think we may both have been confused here.  This case of VariableDefined with DataValue.NoValue is used specifically when the defineVariable has a non-constant default value expression.  The value is temporarily set to DataValue.NoValue until forceExpressionEvaluation actually evaluates the non-constant expression and sets the correct value. A more clear solution to this may be to have variables that have a non-constant default value expression remain Undefined until the expression has been evaluated. I'm testing how that change will look now.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -103,13 +102,22 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
+
     if (context.maybeDefaultValueExpr.isDefined) {
-      val maybeVar = state.variableMap.find(context.globalQName)
-      Assert.invariant(maybeVar.isDefined)
-      maybeVar.get.setState(VariableInProcess)
-      val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
-      suspendableExpression.run(state)
+      val dve = context.maybeDefaultValueExpr.get
+      dve match {
+        case constExpr: ConstantExpression[_] => // Do nothing, value has already been computed at schema compile time
```
stevedlawrence, line 110: Not sure I understand this. If the defaultValue for the NVI is constant, we still need to get that constant value and set it as the defalutValue, right? Seems like this should be something like
```scala
v.setDefalutValue(constExpr.evaluate(state))
```
Though, this is essentially what the NewVariableInstanceSuspendableExpression will do, so I'm not sure if we need a special case for constant. I guess it does avoid allocating this suspendable thing, but that's probably not too big of a deal? And since the expression is constant, it shouldn't even suspend, so we will avoid that overhead regardless.

Can we remove this special case and just always do the below?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -103,13 +102,22 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
+        case _ => {
+          v.setState(VariableInProcess)
+          val suspendableExpression = new NewVariableInstanceDefaultValueSuspendableExpression(dve, context)
```
stevedlawrence, line 113: Can we pass in the variable instance instance the NewVariableInstanceDefaultValue.... so that we don't have to do a variableMap.find in processExpressionResult?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -978,17 +978,12 @@ final class VariableRuntimeData(
...
-      val defaultValueExpr = maybeDefaultValueExpr.get
-      defaultValueExpr match {
-        case constExpr: ConstantExpression[_] => DataValue.unsafeFromAnyRef(constExpr.constant)
-        case _ => DataValue.NoValue
+  private lazy val (state, value) = {
+    if (!maybeDefaultValueExpr.isDefined)
+      (VariableUndefined, DataValue.NoValue)
+    else maybeDefaultValueExpr.get match {
+        case constExpr: ConstantExpression[_] => (VariableDefined, DataValue.unsafeFromAnyRef(constExpr.constant))
+        case _ => (VariableUndefined, DataValue.NoValue)
```
stevedlawrence, line 986: Ah, so this is how you handle constants? I'm curious what others say, but I'd rather not have this special case here. It's maybe a little more performant when it is a constant, but it adds extra special cases and spreads out the NVI default logic in separate places. And I suspect NVI default values will rarely be constants, so we probably won't even hit this case very often.

I'd rather we just delete this state/value thing entirely, change the createVariableInstance function always creates a new instances with VariableUndefinedState and NoValue, and then the parsers/unparsers can mutate the state in a single place.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -103,13 +102,22 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
+          suspendableExpression.run(state)
+        }
+      }
+    } else {
+      // The NVI will inherit the default value of the original variable instance
+      // This will also inherit any externally provided bindings.
+      v.setDefaultValue(v.firstInstanceInitialValue)
```
stevedlawrence, line 120: Just had a though, what if a variable did not have a default value and was not set externaly? Then firstInstanceInitialValue is NoValue, and so we are going to set the default value to not value. Which changes the state to Defined. So we could get in a state where we are Defined but with NoValue. Do we want to make it so firstInstanceInitialValue is an Option, and this becomes something like:
```scala
} else if (v.firstInstanceInitialValue.isDefined) {
  v.setDefaultValue(v.firstInstanceInitialValue.get)
}
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -148,22 +168,26 @@ class VariableInstance private (
     this._value = Some(v)
   }
 
-  def setDefaultValue(v: DataValuePrimitiveNullable) = this._value = Some(v)
+  /* This is used to set a default value without assigning a priorValue */
+  def setDefaultValue(v: DataValuePrimitiveNullable) = {
+    Assert.invariant(this.state == VariableUndefined || this.state == VariableInProcess)
```
stevedlawrence, line 173: Should we assert the ``v != NoValue`` too?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -324,17 +348,31 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
+  /**
+   * This function is called immediately after forceExpressionEvaluations in
+   * order to set the firstInstanceInitialValue for each variable instance.
+   * These initial values will be inherited by future new variable instances if
+   * the newVariableInstance statement does not provide a default value
+   * expression
+   */
+  def setFirstInstanceInitialValues(): Unit = {
+    vTable.foreach { case (_, variableInstances) => {
+      variableInstances.foreach { inst => inst.firstInstanceInitialValue = inst.value }
```
stevedlawrence, line 372: variableInstances should always have a length of 1 here, right? Since this should only ever be called before NVI's are evaluated? Thoughts on making this
```scala
Assert.invariant(variableInstances.length == 1)
variablesInstances(0).firstInstanceInitialValue = inst._value
```
Note that if firstInstanceInitialValue is turned into an Option mentioend above, then this is just a straigt copy of the private _value. So it keeps the Some/None aspect of the value as well if a instance was never set by a default.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
jadams-tresys: So we can actually just do a check to see if firstInstanceInitialValue is defined, which checks to see if it is something other than DataValue.NoValue.  So no need for making it an Option.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Yes, done.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Makes sense and I imagine that would be more performant anyway.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
stevedlawrence: Makes sense. I wonder if then it makes sense for _value to no longer be an Option then? Maybe ``def value`` and ``def stat`` just go away entirely if stateArg and valueArg aren't neeed anymore, and then ``value`` and ``state`` are just vars that are initlaized to NoValue and Undefined just like priorValue and priorState? Seems like _value only needed to be an option to deal with the pass by name of value, but I don't think we need that anymore?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
jadams-tresys: That makes sense to me.  Most of the stuff in RuntimeData was set up to just get things working with the way we had been handling variables before when they were much simpler things without so much to keep track of.

jadams-tresys: Also, just realized that removing the state/value stuff from RuntimeData.scala means we don't need to do the complicated/weird lazy argument stuff.

jadams-tresys: One issue with this approach is that it does break some tests that relied on default values with constant expressions being evaluated at compile time (a few of the external variable loader unit tests (non-tdml).  I'm going to push up a commit with them changed to not use default values.

jadams-tresys: Actually, looking at the tests that are failing, they are all external variable tests that don't use tdml runner.  I'm wondering if these are simply relics of a time when TDMLRunner didn't support external variables.  If so, could we just remove them (after assuring that there are equivalent TDML tests)?

stevedlawrence: Doesn't seem unreasonable to me :+1:

jadams-tresys:
Note that the last commit produces 3 failures in external variable unit
    tests, ie non TDMLRunner tests. They rely on variables constant value
    defaultValueExpressions to be evaluated at schema compile time. I'm
    thinking that these tests were remnants of a time whne TDMLRunner didn't
    support external variables and we simply remove these tests after making
    sure we aren't losing any coverage by only using TDMLRunner tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -65,35 +82,26 @@ case object VariableInProcess extends VariableState
...
- * Circular Definition Exception
  */
 object VariableInstance {
   def apply(
-    stateArg: => VariableState,
-    valueArg: => DataValuePrimitiveNullable,
+    state: VariableState = VariableUndefined,
+    value: DataValuePrimitiveNullable = DataValue.NoValue,
     rd: VariableRuntimeData,
     defaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
```
stevedlawrence, line 91: Do we need to pass in defaultValueExprArg anymore? I think your recent changes made it so we always get the default expression from the VariableRuntimeData when it's need?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -65,35 +82,26 @@ case object VariableInProcess extends VariableState
...
 /**
  * See documentation for object VariableInstance
  */
 class VariableInstance private (
-  @TransientParam stateArg: => VariableState,
-  @TransientParam valueArg: => DataValuePrimitiveNullable,
+  var state: VariableState,
+  var value: DataValuePrimitiveNullable,
   val rd: VariableRuntimeData,
   @TransientParam defaultValueExprArg: => Maybe[CompiledExpression[AnyRef]],
```
stevedlawrence, line 106: Thoughts on moving state, value, priorState, and priorValue to be members of this class, rather than constructor parameters? I think everywhere we create a VariableInstance now these values are always VairableUndefined/NoValue, and with your changes I can't think of a reason why we would ever need to create one of these with something different?

If so, and if we can remove defaultValueExpr, maybe the only parameter is the VariableRunimeData?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -165,11 +166,21 @@ class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
...
-    start.newVariableInstance(context)
+    val nvi = start.newVariableInstance(context)
+
     if (context.maybeDefaultValueExpr.isDefined) {
-      val v = start.variableMap.find(context.globalQName).get
-      val res = DataValue.unsafeFromAnyRef(context.maybeDefaultValueExpr.get.evaluate(start))
-      v.setDefaultValue(res)
+      val dve = context.maybeDefaultValueExpr.get
+      dve match {
+        case constExpr: ConstantExpression[_] => // Do nothing, value has already been computed at schema compile time
```
stevedlawrence, line 174: Remove the constant thing here so it matches unparse?

stevedlawrence (commented):
Looks really good, just a coulple more small potential refactorings.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
jadams-tresys: Missed that, will fix.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: There are a few places that it is still needed where we only have access to VariableInstances without a VariableRuntimeData.  Things like VariableInstance.reset() and forceExpressionEvaluations()

stevedlawrence: But VariantInstance has a reference to VRD, right? Can't those use that?

jadams-tresys: Ah, I see what you mean now.  I can just make the defaultValueExpr a class member and not worry about passing it in.  I think that makes sense. In the VariableInstance contructor we will just set the defaultValueExpr class arg to what is in the VariableRuntimeData.

stevedlawrence: Do we even need to do that? If something needs acces to it could it just do ``vrd.defaultValueExpr``, assume defaultValueExpr is made public if it's not already? Do we need a refrence in both the VariabeInstance and the VRD?

jadams-tresys: You are right, we don't.  I've made the change.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
```diff
@@ -103,13 +101,17 @@ class NewVariableInstanceStartUnparser(override val context: VariableRuntimeData
...
-      maybeVar.get.setState(VariableInProcess)
-      val suspendableExpression = new NewVariableInstanceSuspendableExpression(context.maybeDefaultValueExpr.get, context)
+      val dve = context.maybeDefaultValueExpr.get
+      nvi.setState(VariableInProcess)
+      val suspendableExpression = new NewVariableInstanceDefaultValueSuspendableExpression(dve, context, nvi)
       suspendableExpression.run(state)
+    } else if (nvi.firstInstanceInitialValue.isDefined) {
+      // The NVI will inherit the default value of the original variable instance
+      // This will also inherit any externally provided bindings.
+      nvi.setDefaultValue(nvi.firstInstanceInitialValue)
```
stevedlawrence, line 114: Codecov says this branch isn't covered. We should add a test for this. This is an important edge case we need to make sure is covered.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -65,123 +79,70 @@ case object VariableInProcess extends VariableState
...
     priorState: VariableState = priorState,
     priorValue: DataValuePrimitiveNullable = priorValue) = {
-      val inst = new VariableInstance(state, value, rd, defaultValueExpr, priorState, priorValue)
-      inst.init()
+      val inst = new VariableInstance(rd)
+      inst.state = state
+      inst.priorState = priorState
+      inst.value = value
+      inst.priorValue = priorValue
       inst
```
stevedlawrence, line 140: Now that we no onger need init(), I wonder if we should just make ``VariableInstance`` a case class now? Then we get this copy function for free and Scala ensures all state is copied (in fact, right now, we aren't copying firstInstanceInitialValue so there's potentially a bug). That also means we can get rid of the ``VariableInstance`` companion object since case class will also give us that for free.

stevedlawrence (approved):
+1

Maybe one last bit of minor refactoring, and maybe one an additional test to get full coverage of the new code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Done

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
mbeckerle: Still need this test case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
mbeckerle: The VariableInstance constructor takes only a VRD now. So this thread seems to be resolved.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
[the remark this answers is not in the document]
mbeckerle: Please capture the comment above, clarifying the VariableDefined with DataValue.NoValue - what that means and how used, in code comments please. I suggest a comment is needed where VariableDefined is introduced.

mbeckerle (changes requested):
A test case still needs to be added to cover a corner case. SL observed this particular corner case really does need a test to cover it, and I concur. 

Beyond that, I added only suggestion for minor comments.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
jadams-tresys: Digging into why this wasn't already covered lead me to finding a bug in unparsing. I wasn't forcing the evaluation of defaultValue expressions on defineVariable statements when unparsing.  I've pushed up a change fixing that.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
jadams-tresys: Actually, I dug into this a bit more and it seems that when using a case class the copy method is only a shallow copy.  This ends up causing several tests to break (weirdly none in the TestVariables TDML tests, which is why I thought this was an easy change).  I've reverted back to using a normal class again.

mbeckerle (approved):
+1, assuming CI tests all pass (they're running right now), and codeCov is satisfied for that one case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
stevedlawrence: Just curious, what in VariableInstance needs a deep copy? Isn't all the state just enums and things that don't have a differencne between deep/shallow copy? Wonder if something else is going on?

stevedlawrence: Ah, nevermind, the default case class copy() doesn't copy member var's, only copies the constructor parameters. So all our var state isn't copied with a case class. You're right that we still need this copy method.

stevedlawrence: Might be worth copying the firstInstanceInitialValue just for correctness though.

PR #518, opened 2021-03-30
Allow custom debuggers through API

jw3:
- Moves `DebuggerRunner` setter to `withDebuggerRunner`
- Modify `withDebugger` to accept `Debugger` implementations

DAFFODIL-2491

### DIFF
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
@@ -58,6 +58,7 @@ import java.net.URI
 
 import org.apache.daffodil.api.URISchemaSource
 import org.apache.daffodil.api.Validator
+import org.apache.daffodil.debugger.Debugger
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe._
 import org.apache.daffodil.util.MaybeULong
@@ -532,7 +533,8 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
   /**
    * Enable/disable debugging.
    *
-   * Before enabling, [[DataProcessor#setDebugger(DebuggerRunner)]] must be called with a non-null debugger.
+   * Before enabling, [[DataProcessor#withDebugger]] or [[DataProcessor#withDebuggerRunner(DebuggerRunner)]] must be
+   * called with a non-null debugger.
    *
    * @param flag true to enable debugging, false to disabled
    */
@@ -544,7 +546,8 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
   /**
    * Obtain a new [[DataProcessor]] instance with debugging enabled or disabled.
    *
-   * Before enabling, [[DataProcessor#withDebugger(DebuggerRunner)]] must be called to obtain a [[DataProcessor]] with a non-null debugger.
+   * Before enabling, [[DataProcessor#withDebugger(Debugger)]] or [[DataProcessor#withDebuggerRunner(DebuggerRunner)]]
+   * must be called to obtain a [[DataProcessor]] with a non-null debugger.
    *
    * @param flag true to enable debugging, false to disabled
    */
@@ -557,7 +560,7 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
    *
    * @param dr debugger runner
    */
-  @deprecated("Use withDebugger.", "2.6.0")
+  @deprecated("Use withDebuggerRunner.", "2.6.0")
   def setDebugger(dr: DebuggerRunner): Unit = {
     val debugger = newDebugger(dr)
     dp = dp.withDebugger(debugger)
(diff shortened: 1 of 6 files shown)


daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -557,7 +558,7 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
    *
    * @param dr debugger runner
    */
-  @deprecated("Use withDebugger.", "2.6.0")
+  @deprecated("Use withDebuggerRunner.", "2.6.0")
```
jw3, line 561: So the API change causes a mismatch with the deprecated method.  Took approach of leaving deprecated method as-is and updated the message.  :man_shrugging:

jw3:
Questioned the API impact of refactoring the existing method to match the `DebuggerRunner` parameter.  Though process was

- not refactoring is a significant naming clash
- not likely a well used part of the API

:grimacing:

tuxji (changes requested):
I can't see anything wrong with these changes and I would give them a +1, but for some reason the CI build is failing on Java 8 and 11 with this error while generating the javadocs (even though the CI succeeds on Java 15):

> /home/runner/work/daffodil/daffodil/daffodil-japi/target/java/org/apache/daffodil/japi/DataProcessor.java:163:1: reference not found

We also should fix a javadoc warning: 

> [warn] /home/runner/work/daffodil/daffodil/daffodil-udf/src/main/java/org/apache/daffodil/udf/UserDefinedFunctionProvider.java:48:3: Could not find any member to link for "ReflectiveOperationException".

jw3:
> I can't see anything wrong with these changes and I would give them a +1

Thanks.  :ok_hand:

> for some reason the CI build is failing on Java 8 and 11 with this error while generating the javadocs

Yeah I kicked off a doc build locally and it worked.  Need to dig into it.  Will check out that other one too.

Looks like I need a test for the new method for codecov too.

jw3:
Ah its a doc in a generated file in the JAPI, trying to reference the refactored name

jw3:
> We also should fix a javadoc warning:
> 
  > > [warn] /home/runner/work/daffodil/daffodil/daffodil-udf/src/main/java/org/apache/daffodil/udf/UserDefinedFunctionProvider.java:48:3: Could not find any member to link for "ReflectiveOperationException".

Looked at this some, not sure what is causing it but did exist prior to this PR.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -17,47 +17,48 @@
 
 package org.apache.daffodil.japi
 
-import org.apache.daffodil.compiler.{ Compiler => SCompiler }
+import org.apache.daffodil.compiler.{Compiler => SCompiler}
```
tuxji, line 20: I don't know which style is closer to the standard Scala formatting as implemented by the unmodified defaults of the Scalariform tool, but let's make any formatting changes in a separate PR that reformats the entire codebase uniformly.  Will you please remove these formatting changes which your IDE must have made? 

Also, which IDE are you using?  I checked two pages in our Wiki (https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Style+Guidelines and https://cwiki.apache.org/confluence/display/DAFFODIL/IntelliJ+IDEA+Setup) and found nothing there which tells people how to fix this particular formatting default in IntelliJ IDEA.  If you're using IDEA, that's an oversight which we need to fix - please let me know how to find the default after you change it and I'll update the Wiki for you.

tuxji (approved):
+1 after import formatting changes are reversed

Looks great!  Just need to reverse some inadvertent formatting changes to keep the codebase consistent.  About that pre-existing javadoc warning - it would have been nice to fix it but no worries since it sounds like it's not trivial.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
jw3: Yeah, unintentional, repaired.

> Also, which IDE are you using? 

Intellij.  I use [Scalafmt](https://scalameta.org/scalafmt/) for nearly everything else, never adjusted for this project.

> ... how to fix this particular formatting default in IntelliJ IDEA

@mbeckerle's [comment here](https://cwiki.apache.org/confluence/display/DAFFODIL/IntelliJ+IDEA+Setup?focusedCommentId=148644684#comment-148644684) fixes it.

mbeckerle (commented):
These changes all look fine to me. Not sure what's going on with CI and Java 8 and 11.

jw3:
> Not sure what's going on with CI and Java 8 and 11.

_Should_ be all good now.  Those earlier failures were doc related, from a comment referencing a type that should have been refactored but was not  due to the order in which I renamed/changed parameter type.  Intellij refactor got all the code points but missed the doc.

mbeckerle (approved):
+1

PR #519, opened 2021-03-31
WIP on doctype restriction 

mbeckerle:
DAFFODIL-1422 is a ticket about restricting the XML we accept so that we do not allow DOCTYPE declarations in it.
This is a security related provision, as use of DOCTYPEs leaves XML loaders subject to a variety of problems such as documents exploding in size as the DOCTYPE declarations are expanded. 
DOCTYPEs are an old obsolete idea, and simply disallowing them entirely is the best option.

There are other jira tickets about disallowing resolvers and loaders from dereferencing URIs treating them as internet URLs. 

The upshot of all this is that "loading" XML is tricky, and needs to be done carefully via a centralized library that provides the various options different loading requires, while not exposing/allowing the security vulnerabilities. 

This change set does not yet include establishing a single central library that all XML loading goes through. 

Right now this is at the point where it is apparent such a library is needed, because there are too many places that are invoking XML loaders for them to just all be "done the right way". Under maintenance this is too likely to drift. 

A key starting point is to survey every place Daffodil does XML loading. These include loading of:

* DFDL schemas and the include/import schema files they reference
** This can include DFDL schemas, but also XSD for annotation languages (e.g., schematron annotations)
** Note that this validating loader is loading a schema, but loading it not as a schema, but as ordinary XML. This should be validated against the schema for DFDL schemas. 
* XML Infosets being unparsed
** (Currently not a validating loader)
* TDML files for testing - this can in turn lead to loading of DFDL schemas
** Test cases can load XML Infoset files. 
** validation here involves validating defineSchema elements which contain DFDL schema. 
* Config files
* daffodil-propgen loads the XSD files for DFDL annotations, tunables, etc.
* Xerces validator loads a schema, and the include/import files it references
* Xerces validator loads XML being validated 

This may not be a comprehensive list.

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -28,7 +28,6 @@ import java.nio.channels.Channels
 import java.nio.file.Paths
 import java.util.Scanner
 import java.util.concurrent.Executors
-
 import com.typesafe.config.ConfigFactory
 
 import scala.concurrent.Await
@@ -37,7 +36,6 @@ import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.xml.Node
 import scala.xml.SAXParseException
-
 import javax.xml.parsers.DocumentBuilderFactory
 import javax.xml.parsers.SAXParserFactory
 import javax.xml.transform.TransformerFactory
@@ -106,8 +104,10 @@ import org.rogach.scallop
 import org.rogach.scallop.ArgType
 import org.rogach.scallop.ScallopOption
 import org.rogach.scallop.ValueConverter
+import org.xml.sax.XMLReader
 
 import scala.util.matching.Regex
+import scala.xml.SAXParser
 
 class CommandLineSAXErrorHandler() extends org.xml.sax.ErrorHandler with Logging {
 
@@ -807,7 +807,18 @@ object Main extends Logging {
           case Left(bytes) => new ByteArrayInputStream(bytes)
           case Right(is) => is
         }
-        scala.xml.XML.load(is)
+        val parser: SAXParser = {
+          val f = SAXParserFactory.newInstance()
+          f.setNamespaceAware(false)
+          val p = f.newSAXParser()
+          p
+        }
+        //
+        // FIXME: This needs to use a Daffodil-created loader, so that we can
+        // insure that the XMLReader used from it has the setSecureDefaults treatment.
+        //
+        System.err.println("FIXME: Needs to use a Daffodil-created loader.")
+        scala.xml.XML.withSAXParser(parser).load(is)
       }
       case InfosetType.JDOM => {
         val is = data match {
@@ -1488,7 +1499,9 @@ object Main extends Logging {
   private def unparseWithSAX(
     is: InputStream,
     contentHandler: DFDL.DaffodilUnparseContentHandler): UnparseResult = {
-    val xmlReader = SAXParserFactory.newInstance.newSAXParser.getXMLReader
+    val xmlReader: XMLReader = SAXParserFactory.newInstance.newSAXParser.getXMLReader
+    // standard secure XML parsing features
+    XMLUtils.setSecureDefaults(xmlReader)
     xmlReader.setContentHandler(contentHandler)
     xmlReader.setFeature(XMLUtils.SAX_NAMESPACES_FEATURE, true)
     xmlReader.setFeature(XMLUtils.SAX_NAMESPACE_PREFIXES_FEATURE, true)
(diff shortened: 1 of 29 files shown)


mbeckerle:
We need tests that use doctype decls and general entities and verify that we reject them.

mbeckerle:
Note: if we disable DOCTYPE we are disabling general entities, as those can only be supplied in a doctype.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -807,7 +807,18 @@ object Main extends Logging {
...
+          val f = SAXParserFactory.newInstance()
+          f.setNamespaceAware(false)
+          val p = f.newSAXParser()
+          p
+        }
+        //
+        // FIXME: This needs to use a Daffodil-created loader, so that we can
+        // insure that the XMLReader used from it has the setSecureDefaults treatment.
+        //
+        System.err.println("FIXME: Needs to use a Daffodil-created loader.")
```
mbeckerle, line 820: Still to fix.

daffodil-lib/src/main/scala/org/apache/daffodil/api/DaffodilSchemaSource.scala
```diff
@@ -109,6 +113,27 @@ class URISchemaSource protected (val fileOrResource: URI) extends DaffodilSchema
   }
 }
 
+/**
+ * Convenient for testing. Allows creating XML strings for loading that contain things
+ * not supported by scala XML syntax like DOCTYPE decls.
+ *
+ * @param str - string that is loaded as XML.
+ */
+case class StringSchemaSource(str: String)
```
mbeckerle, line 122: If for testing only, can live in src/test/scala

daffodil-lib/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesXMLValidator.scala
```diff
@@ -17,22 +17,31 @@
...
   }
 
   def validate(xmlFile: File): Either[java.lang.Throwable, _] = {
     try {
-      val factory = new org.apache.xerces.jaxp.validation.XMLSchemaFactory()
+      val factory = new XMLSchemaFactory()
+      factory.setResourceResolver(DFDLCatalogResolver.get)
       val schema = factory.newSchema(new StreamSource(extVarXsd))
       val validator = schema.newValidator()
+      validator.setFeature(XMLUtils.XML_DISALLOW_DOCTYPE_FEATURE, true)
```
mbeckerle, line 44: move to XMLUtils.setSecureDefaults(v: Validator) ??

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -450,6 +450,64 @@ object XMLUtils {
...
+  val XML_EXTERNAL_GENERAL_ENTITIES_FEATURE = "http://xml.org/sax/features/external-general-entities"
+
+  /**
+   * Sets properties that disable insecure XML reader behaviors.
+   * @param xmlReader - the reader to change feature settings on.
+   */
+  def setSecureDefaults(xmlReader: XMLReader) : Unit = {
+    try {
+      xmlReader.setFeature(XMLUtils.XML_DISALLOW_DOCTYPE_FEATURE, true)
+      xmlReader.setFeature(XMLUtils.XML_EXTERNAL_PARAMETER_ENTITIES_FEATURE, false)
```
mbeckerle, line 475: Add comment that these next 2 restrictions are not strictly speaking necessary, as they are implied by disallowing doctypes.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -450,6 +450,64 @@ object XMLUtils {
...
+      // System.err.println("SAX XMLReader supports disallow properties: " + xmlReader)
+    } catch {
+      case e: SAXNotRecognizedException => {
+        // System.err.println("xmlReader: " + e.getMessage()) // good place for a breakpoint
+        throw e
+      }
+    }
+  }
+
+//  def setSecureDefaults(schemaFactory: org.apache.xerces.jaxp.validation.XMLSchemaFactory) : Unit = {
```
mbeckerle, line 486: Remove commented code. Add comment above that the disallowing of doctypes works for XMLReader and Validator, but not XMLSchemaFactory.

daffodil-runtime1/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesLoader.scala
```diff
@@ -66,13 +68,15 @@ object ExternalVariablesLoader {
...
     Assert.usage(file ne null)
-    ExternalVariablesValidator.validate(file) match {
+    val validatorResult = ExternalVariablesValidator.validate(file)
+    validatorResult match {
       case Left(ex) => Assert.abort(ex)
       case Right(_) => // Success
     }
     val enc = determineEncoding(file) // The encoding is needed for ConstructingParser
     val input = scala.io.Source.fromURI(file.toURI)(enc)
-    val node = ConstructingParser.fromSource(input, true).document.docElem
```
mbeckerle, line 75: Add comment that this was changed from the Constructing loader because of a unexpected error, and reusing the DaffodlXMLLoader works, no need for a special case loader for external variable bindings.

mbeckerle:
Closing. Will open a new PR for this.

PR #524, opened 2021-04-07
Updated test to check for both '.' or ','

jadams-tresys:
Using the following settings:

export SBT_OPTS="-Xss2M -Xmx5G -Xms5G -Duser.country=DE
-Duser.language=de"

Results in floating point numbers using a ',' instead of a '.' (ie.
'3,0' instead of '3.0'. Updated test to use a regex to check for both
symbols.

DAFFODIL-2492


stevedlawrence (approved):

stevedlawrence:
+1

Would it be worth creating a ticket to modify our GitHub actions so one or two of the jobs runs with a different locale so we can automate a bit of this testing?

jadams-tresys:
I think you are right Steve, we definitely could use some recurring locale testing as I could definitely see this sort of things continuing to pop up in the future.

tuxji (approved):
+1

PR #525, opened 2021-04-07
Make some daffodil-runtime2 improvements

tuxji:
Refactor duplicate code in Binary*CodeGenerator.scala files.

Change PState/UState member field from a string error message to a
struct error object.  Make parse/unparse functions initialize that
member field if any error happens.  Move checks for errors from
parse/unparse functions to their callers, e.g., "if (ustate->error)
return;".

Collect validation errors without stopping parsing or unparsing by
initializing another PState/UState member field pointing to a
validation struct with some fields.  As last step, update all C files
using include-what-you-use to document what's used from each include.

Add instructions to README.md how to setup required C tool-chain
elements on Linux and Windows.

Generate daffodil-runtime2's name and version into generated_code.c
automatically so we won't need to keep it up to date.

Changelog:

In README.md, fix or disable markdownlint warnings.  Expand Build
Requirements with instructions how to install C compiler, Mini-XML
library, java, and sbt on Linux and Windows.  Verify/update old
hyperlinks and add new MSYS2 and clang hyperlinks.

In build.sbt, add Compile / cCompiler and ccArchiveCommand settings as
a guide to show developers how to override these default settings if
they want sbt to call the C compiler with a different name than "cc".
The sbt-cc plugin doesn't iterate through a list (${CC}, cc, gcc,
clang, zig cc) to find a C compiler like our runtime2 code does.

In daffodil_argp.c, remove argp_program_version's definition since we
now generate it automatically within generated_code.c.

In daffodil_main.c, add fflush_or_exit and change fopen_or_exit /
fclose_or_open to call continue_or_exit with error message instead of
calling exit directly.  Make main print any validation diagnostics
before printing any errors since they've been split into separate
PState/UState member fields.  Also make main call continue_or_exit
instead of calling error directly and pass Error structs, not strings,
to all continue_or_exit calls.

In stack.c, make stack functions call continue_or_exit instead of
calling error directly and pass Error structs, not strings, to all
continue_or_exit calls.

In xml_reader.c, make conversion functions return Error structs, not
strings, to indicate errors converting strings to primitive datums.
Make reader functions return Error structs, not strings, to indicate
errors reading XML data.

In xml_writer.c, make writer functions return Error structs, not
strings, to indicate errors writing XML data.  Remove redundant stack
checking code already performed within stack fuctions.

In errors.c (new file), use switch statement inside error_message
function to centralize all error messages in one place for easier
future internationalization.  Use switch statement inside
print_mybe_stop function to allow error messages to interpolate an
extra detail (string or integer) if needed.  Define a single static
array that can be used to store validation diagnostics and define a
function to print validation diagnostics.  Implement continue_or_exit
and eof_or_error functions here as well.

In errors.h (new file), define enumerations for each type of error
which could occur in C code.  Define an Error struct to store an
enumeration and optionally an extra detail (string or integer).  Also
define a Diagnostics struct to store an array of data validation
errors which should be reported separately from parse errors.  Move
PState and UState structs here from infoset.h and give them member
fields pointing to data validation and parse errors.  Also declare
need_diagnostics, print_diagnostics, continue_or_exit, eof_or_error
prototypes and UNUSED macro in errors.h.

In infoset.c, make walkInfosetNode and walkInfoset functions return
Error struct, not string.  Remove eof_or_error_msg function (now moved
to errors.c with shorter name).

In infoset.h, change InitChoiceRD and Visit* prototypes to return
Error structs, not strings.  Remove PState & UState structs,
eof_or_error_msg prototype, and UNUSED macro (all moved to errors.h).

In parsers.c, make parse_<endian>_<type> functions stop enclosing
their code in "if (!pstate->error_msg)" statements and use Error
structs, not strings.  Callers are now expected to check for errors
after each call immediately.  Make parse_fill_bytes use
read_stream_update_position helper macro too.  Make
parse_validate_fixed report errors in separate pstate->validati field,
not pstate->error field.

In unparsers.c, make unparse_<endian>_<type> functions stop enclosing
their code in "if (!ustate->error_msg)" statements and use Error
structs, not strings.  Callers are now expected to check for errors
after each call immediately.  Make unparse_fill_bytes use
write_stream_update_position helper macro too.  Make
unparse_validate_fixed report errors in separate ustate->validati
field, not ustate->error field.

In NestedUnion.c and ex_nums.c examples, update with freshly generated
code to illustrate code generator changes (in particular, defining
argp_program_version, adding an if statement after each parse/unparse
call to check for an error and skip rest of calls immediately, and
returning errors directly from initChoice functions with choice key
that failed to match any branch key).

In BinaryAbstractCodeGenerator.scala (new file), refactor common code
that used to be duplicated in BinaryBooleanCodeGenerator.scala,
BinaryFloatCodeGenerator.scala, and
BinaryIntegerKnownLengthCodeGenerator.scala; now they pass parameters
to binaryAbstractGenerateCode instead.

In CodeGeneratorState.scala, make all necessary changes to report
errors using structs instead of strings in generated code including
initChoice functions, add if statements after each parse/unparse call,
and generate the program name and version automatically.

In build.properties, change sbt.version from 1.4.1 to 1.4.9 to ensure
successful sbt build from scratch in a new Ubuntu or Windows VM (sbt
was getting java.lang.NoClassDefFound: scala/Serializer).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@
   limitations under the License.
 -->
 
+<!-- markdownlint-disable first-line-heading -->
+<!-- markdownlint-disable line-length -->
+<!-- markdownlint-disable no-inline-html -->
 [<img src="https://daffodil.apache.org/assets/themes/apache/img/apache-daffodil-logo.svg" height="85" align="left" alt="Apache Daffodil"/>][Website]
 [<img src="https://img.shields.io/github/workflow/status/apache/daffodil/Daffodil%20CI/master.svg" align="right"/>][GitHub Actions]
 <br clear="right" />
@@ -23,89 +26,114 @@
 [<img src="https://img.shields.io/maven-central/v/org.apache.daffodil/daffodil-core_2.12.svg?color=brightgreen&label=version" align="right"/>][Releases]
 <br clear="both" />
 
-Apache Daffodil is an open-source implementation of the [DFDL specification]
-that uses DFDL data descriptions to parse fixed format data into an infoset.
-This infoset is commonly converted into XML or JSON to enable the use of
-well-established XML or JSON technologies and libraries to consume, inspect,
-and manipulate fixed format data in existing solutions. Daffodil is also
-capable of serializing or "unparsing" data back to the original data format.
-The DFDL infoset can also be converted directly to/from the data structures
-carried by data processing frameworks so as to bypass any XML/JSON overheads.
+Apache Daffodil is an open-source implementation of the [DFDL
+specification] that uses DFDL data descriptions to parse fixed format
+data into an infoset.  This infoset is commonly converted into XML or
+JSON to enable the use of well-established XML or JSON technologies
+and libraries to consume, inspect, and manipulate fixed format data in
+existing solutions.  Daffodil is also capable of serializing or
+"unparsing" data back to the original data format.  The DFDL infoset
+can also be converted directly to/from the data structures carried by
+data processing frameworks so as to bypass any XML/JSON overheads.
 
-For more information about Daffodil, see https://daffodil.apache.org/.
+For more information about Daffodil, see the [Website].
 
 ## Build Requirements
 
 * JDK 8 or higher
 * SBT 0.13.8 or higher
-* C compiler (for daffodil-runtime2 only)
-* Mini-XML Version 3.2 or higher (for daffodil-runtime2 only)
+* C compiler C99 or higher
+* Mini-XML Version 3.2 or higher
+
+Since Daffodil has a DFDL to C backend, you will need a C compiler
+([gcc] or [clang]), the [Mini-XML] library, and possibly the GNU
+[argp] library if your system's C library doesn't include it.  You can
+install gcc and libmxml as system packages on most Unix based
+platforms with distribution-specific packager commands such as (Debian
+and Ubuntu):
+
+    # Just mentioning all other packages you might need too
+    sudo apt install build-essential curl git libmxml-dev
+
+You will need the Java Software Development Kit ([JDK]) and the Scala
+Build Tool ([SBT]) to build Daffodil, run all tests, create packages,
+and more.  [SDK] offers an easy and uniform way to install both java
+and sbt on any Unix based platform:
+
+    curl -s "https://get.sdkman.io" | bash
+    sdk install java
+    sdk install sbt
+
+You can edit the Compile / cCompiler setting in build.sbt if you don't
+want sbt to call your C compiler with "cc" as the driver command.
+
+On Windows, the easiest way to install gcc and libargp is to install
+[MSYS2]'s collection of free tools and libraries although MSYS2 has no
+package for libmxml which you'll need to build from source.  First
+install [MSYS2] following its website's installation instructions,
+then run the following commands in a "MSYS2 MSYS" window:
+
+    pacman -S gcc git libargp-devel make pkgconf
+    git clone https://github.com/michaelrsweet/mxml.git
+    cd mxml
+    ./configure --prefix=/usr --disable-shared --disable-threads
+    make
+    make install
+
+You also need to install [JDK} and [SBT] from their Windows
+installation packages and define an environment variable using
+Windows' control panel for editing environment variables.  Define an
+environment variable with the name `MSYS2_PATH_TYPE` and the value
+`inherit`.  Now when you open a new "MSYS2 MSYS" window from the Start
+Menu, you will be able to type your sbt commands in the MSYS2 window
+and both sbt and daffodil will be able to call the C compiler.
 
 ## Getting Started
-
-You will need the full Java Software Development Kit ([JDK] or [SDK]),
-not the Java Runtime Environment (JRE), to build Daffodil.  You also
-will need [SBT] to build Daffodil, run all tests, create packages, and
-more.
-
-In order to build daffodil-runtime2, you will need a C compiler (for
-example, [gcc]), the [Mini-XML] library, and possibly the [argp]
-library if your system doesn't include it in its C library.
 
 Below are some of the more common commands used for Daffodil development.
 
 ### Compile
 
-```text
-$ sbt compile
-```
+    sbt compile
 
 ### Tests
 
 Run all unit tests:
 
-```text
-$ sbt test 
-```
+    sbt test
 
 Run all command line interface tests:
 
-```text
-$ sbt it:test
-```
+    sbt it:test
 
 ### Command Line Interface
 
-Create Linux and Windows shell scripts in `daffodil-cli/target/universal/stage/bin/`. See
-the [Command Line Interface] documentation for details on its usage:
+Create Linux and Windows shell scripts in
+`daffodil-cli/target/universal/stage/bin/`.  See the [Command Line
+Interface] documentation for details on its usage:
 
-```btext
-$ sbt daffodil-cli/stage
-```
+    sbt daffodil-cli/stage
 
 ### License Check
 
-Generate an [Apache RAT] license check report located in ``target/rat.txt`` and error if
-any unapproved licenses are found:
+Generate an [Apache RAT] license check report located in
+``target/rat.txt`` and error if any unapproved licenses are found:
 
-```text
-$ sbt ratCheck
-```
+    sbt ratCheck
 
 ### Test Coverage Report
 
 Generate an [sbt-scoverage] test coverage report located in
 ``target/scala-ver/scoverage-report/``:
 
-```text
-$ sbt clean coverage test it:test
-$ sbt coverageAggregate
-```
+    sbt clean coverage test it:test
+    sbt coverageAggregate
 
 ## Getting Help
 
 For questions, we can be reached at the dev@daffodil.apache.org or
-users@daffodil.apache.org mailing lists. Bugs can be reported via the [Daffodil JIRA].
+users@daffodil.apache.org mailing lists.  Bugs can be reported via the
+[Daffodil JIRA].
 
 ## License
 
@@ -113,17 +141,19 @@ Apache Daffodil is licensed under the [Apache License, v2.0].
 
 [Apache License, v2.0]: https://www.apache.org/licenses/LICENSE-2.0
 [Apache RAT]: https://creadur.apache.org/rat/
-[CodeCov]: https://codecov.io/gh/apache/daffodil/
+[CodeCov]: https://app.codecov.io/gh/apache/daffodil
 [Command Line Interface]: https://daffodil.apache.org/cli/
-[DFDL specification]: http://www.ogf.org/dfdl
-[Daffodil JIRA]: https://issues.apache.org/jira/projects/DAFFODIL
+[DFDL specification]: https://daffodil.apache.org/docs/dfdl/
+[Daffodil JIRA]: https://issues.apache.org/jira/projects/DAFFODIL/
 [Github Actions]: https://github.com/apache/daffodil/actions?query=branch%3Amaster+
-[JDK]: https://docs.oracle.com/en/java/javase/11/install/overview-jdk-installation.html
+[JDK]: https://adoptopenjdk.net/
 [Mini-XML]: https://www.msweet.org/mxml/
+[MSYS2]: https://www.msys2.org/
 [Releases]: http://daffodil.apache.org/releases/
-[SBT]: http://www.scala-sbt.org
-[SDK]: https://sdkman.io
-[Website]: https://daffodil.apache.org
+[SBT]: https://www.scala-sbt.org/
+[SDK]: https://sdkman.io/
+[Website]: https://daffodil.apache.org/
 [argp]: https://packages.msys2.org/package/libargp-devel
+[clang]: https://clang.llvm.org/get_started.html
 [gcc]: https://linuxize.com/post/how-to-install-gcc-on-ubuntu-20-04/
-[sbt-scoverage]: https://github.com/scoverage/sbt-scoverage
+[sbt-scoverage]: https://github.com/scoverage/sbt-scoverage/
--- a/build.sbt
+++ b/build.sbt
@@ -68,6 +68,8 @@ lazy val runtime2         = Project("daffodil-runtime2", file("daffodil-runtime2
                               .settings(commonSettings)
                               .settings(publishArtifact in (Compile, packageDoc) := false)
                               .settings(
+                                Compile / cCompiler := "cc",
+                                Compile / ccArchiveCommand := "ar",
                                 Compile / ccTargets := ListSet(runtime2CFiles),
                                 Compile / cSources  := Map(
                                   runtime2CFiles -> (
--- a/daffodil-runtime2/src/main/resources/c/libruntime/errors.c
+++ b/daffodil-runtime2/src/main/resources/c/libruntime/errors.c
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "errors.h"
+#include <error.h>     // for error
+#include <inttypes.h>  // for PRId64
+#include <stdio.h>     // for NULL, feof, ferror, FILE, size_t
+#include <stdlib.h>    // for EXIT_FAILURE
+
+// error_message - return an internationalized error message
+
+static const char *
+error_message(enum ErrorCode code)
+{
+    switch (code)
+    {
+    case ERR_CHOICE_KEY:
+        return "no match between choice dispatch key %" PRId64
+               " and any branch key";
+    case ERR_FILE_CLOSE:
+        return "error closing file";
+    case ERR_FILE_FLUSH:
+        return "error flushing stream to file";
+    case ERR_FILE_OPEN:
+        return "error opening file '%s'";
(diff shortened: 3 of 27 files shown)


mbeckerle (approved):
+1 I had minor comments/suggestions. In general this is a really nice change set, as it handles the whole issue of generalizing the error diagnostics system to one that is more appropriately abstract. Not string oriented.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,117 @@
...
+};
+
+// Error - specific error occuring now
+
+typedef struct Error
+{
+    enum ErrorCode code;
+    union
+    {
+        const char *s;   // for %s
```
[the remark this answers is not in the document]
tuxji: Oh, I didn't consider passing schema filename/lineno information for a diagnostic.  We'd still want to pass a string or int64 for some errors, so I would store an ERD (or a Location object) in its own separate Error member field, not as an alternative field in the union.  We currently don't store any schema filename/lineno in the ERD metadata, so I'll add this idea to the runtime2 todos for a later time.  We also might want to pass a PState/UState for some errors to convey filename/position information about where the error was detected in the data, yes?

daffodil-runtime2/src/main/resources/c/libruntime/parsers.c
```diff
@@ -142,36 +136,38 @@ define_parse_endian_integer(le, uint, 32)
...
+        Diagnostics *validati = need_diagnostics();
+        pstate->validati = validati;
+
+        if (validati->length <
+            sizeof(validati->array) / sizeof(*validati->array))
+        {
+            Error *error = &validati->array[validati->length++];
+            error->code = ERR_FIXED_VALUE;
+            error->s = element;
+        }
```
tuxji, line 171: This is the compile-time array length calculation I was talking about.  I'll replace this if statement with an add_diagnostic call as you suggested.  Do you really want add_diagnostic to stop the program as soon as the array fills up, or simply stop adding any more diagnostics to the array?

```c
        const Error error = {ERR_FIXED_VALUE, {element}};
        add_diagnostic(validati, &error);
```

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,117 @@
...
+        int64_t     d64; // for %d64
+    };
+} Error;
+
+// Diagnostics - array of validation errors
+
+typedef struct Diagnostics
+{
+    Error  array[100];
```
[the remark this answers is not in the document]
tuxji: Actually, the C standard allows us to find an array's size at compile time using two sizeof calls: `sizeof(array) / sizeof(*array)`.  Since C doesn't allow us to define a scoped compile-time constant inside a header file easily (we'd have to define a macro or an enumeration, both of which go into the global namespace), I thought it was better to use 100 in the header file and two sizeof calls to get the array's size.  However, I like your idea of defining an add_diagnostic abstraction so I'll define that function in errors.[ch] and call it in parsers.c & unparsers.c.

tuxji (commented):
Thanks for the suggestions, I'll take care of one now and add the other to the runtime2 todos.  I have a question below as well.

tuxji:
I can run sbt compile in a Windows VM fine, so I'm not sure why the Windows checks are hanging at that step.  Let's continue reviewing the PR and merge it after a second +1 since I need to rebase the branch on the latest master anyway.  Will someone else please review this PR?

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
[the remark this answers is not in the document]
mbeckerle: Yes, in the regular scala backend, the parse error method is actually a method of the PState/UState object, so that's always implicitly passed. We depend on the fact that the state has all the static information accessible. In this runtime, perhaps we have to pass the static runtime-data object as well as the state object. 

In the standard scala backend, there are runtime-data objects that are not ERD, i.e., they are associated with choices, variables, etc. We should plan on a need for that also. More than just "elements" need runtime data containing static information.  

Basically, everything that could ever need runtime context (which is everything, since this is a runtime/backend) should be passed a state, and either the state provides access to the runtime data, or the runtime data (static context) should also be passed.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,121 @@
...
+    size_t length;
+} Diagnostics;
+
+// PState - mutable state while parsing data
```
mbeckerle, line 77: We should consider whether we want to adopt one of the Java-doc like tools for C/C++ code and structure our comments that way.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,121 @@
...
+
+typedef struct UState
+{
+    FILE *       stream;   // output to write data to
+    size_t       position; // 0-based position in stream
+    Diagnostics *validati; // any validation diagnostics
+    const Error *error;    // any error which stops program
+} UState;
+
+// need_diagnostics - return pointer to validation diagnostics
```
mbeckerle, line 97: Is this a naming convention? I would have expected this to be named get_diagnostics, as it is a "getter".

daffodil-runtime2/src/main/resources/c/libruntime/errors.c
```diff
@@ -0,0 +1,200 @@
...
+
+// add_diagnostic - add a new error to validation diagnostics
+
+void
+add_diagnostic(Diagnostics *validati, const Error *error)
+{
+    if (validati && error)
+    {
+        if (validati->length <
+            sizeof(validati->array) / sizeof(*validati->array))
```
mbeckerle, line 146: I guess the only way more than one can accumulate here is if they are warnings. 

But to continue execution when this is full feels like the wrong choice. 

Maybe return a boolean which indicates whether the diagnostic was able to be recorded or not. 

Then the place issuing the warning can decide if there are "too many warnings" and exit, or not?

mbeckerle (approved):
+1 once windows CI issues are fixed, or understood to be spurious.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,121 @@
...
+    ERR_STRTONUM_RANGE,
+    ERR_XML_DECL,
+    ERR_XML_ELEMENT,
+    ERR_XML_ERD,
+    ERR_XML_GONE,
+    ERR_XML_INPUT,
+    ERR_XML_LEFT,
+    ERR_XML_MISMATCH,
+    ERR_XML_WRITE
+};
```
stevedlawrence, line 55: Runtime1 has different kinds of diagnostics, e.g. error, warning, validation error, recoverable error. Are these all just considered "errors" in runtime1? Are there any diagnostics that are fatal errors?

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -0,0 +1,121 @@
...
+{
+    FILE *       stream;   // output to write data to
+    size_t       position; // 0-based position in stream
+    Diagnostics *validati; // any validation diagnostics
+    const Error *error;    // any error which stops program
+} UState;
+
+// need_diagnostics - return pointer to validation diagnostics
+
+extern Diagnostics *need_diagnostics(void);
```
stevedlawrence, line 99: Why get diagnostics from the static variable, rather than just having the PState/UState have a Diagnostics instances? Then PState/UState can be passed around and this state can be mutated as diagnostics/errors are created. Avoids potentialy issues with global state/threading.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryAbstractCodeGenerator.scala
```diff
@@ -0,0 +1,70 @@
...
+    e.schemaDefinitionUnless(e.bitOrder eq BitOrder.MostSignificantBitFirst, "Only dfdl:bitOrder 'mostSignificantBitFirst' is supported.")
+    e.schemaDefinitionUnless(e.byteOrderEv.isConstant, "Runtime dfdl:byteOrder expressions not supported.")
+    e.schemaDefinitionUnless(e.elementLengthInBitsEv.isConstant, "Runtime dfdl:length expressions not supported.")
+
+    val fieldName = e.namedQName.local
+    val byteOrder = e.byteOrderEv.constValue
+    val conv = if (byteOrder eq ByteOrder.BigEndian) "be" else "le"
+    val arraySize = if (e.occursCountKind == OccursCountKind.Fixed) e.maxOccurs else 0
+    val fixed = e.xml.attribute("fixed")
+    val fixedValue = if (fixed.isDefined) fixed.get.text else ""
```
stevedlawrence, line 41: Would prefer we add this fixed logic to the dsom. For example, right now there is nothing pasing the fixed value and making sure it matches the type. For example
```
<xs:element name="someBool" type="xs:boolean" fixed="invalidBooleanValue" />
```
Ideally the dsom would convert the "fixed" value to match the primitive type, which the DSOM should be doing.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryAbstractCodeGenerator.scala
```diff
@@ -0,0 +1,70 @@
...
+           |    if (pstate->error) return;""".stripMargin
+      val unparseStatement =
+        s"""    unparse_${conv}_$prim(instance->$fieldName$deref, $unparseArgs);
+           |    if (ustate->error) return;""".stripMargin
+      cgState.addSimpleTypeStatements(initStatement, parseStatement, unparseStatement)
+
+      if (fixedValue.nonEmpty) {
+        val init2 = ""
+        val parse2 =
+          s"""    parse_validate_fixed(instance->$fieldName$deref == $fixedValue, "$fieldName", pstate);
```
stevedlawrence, line 56: Wondering if there's a potentially security issue here? Looks like we just inject the fixed value from teh schema directly into this C code. Which means I think a schema could look something like this to inject some code?
```xml
<xs:element name="foo" ... fixed='0, "foo", pstate); malcious C code here; //' />
```

stevedlawrence (approved):
+1

Handful of questions, but looks good! Some nice refactorings.

README.md
```diff
@@ -23,107 +26,134 @@
...
+
+Since Daffodil has a DFDL to C backend, you will need a C compiler
+([gcc] or [clang]), the [Mini-XML] library, and possibly the GNU
+[argp] library if your system's C library doesn't include it.  You can
+install gcc and libmxml as system packages on most Unix based
+platforms with distribution-specific packager commands such as (Debian
+and Ubuntu):
+
+    # Just mentioning all other packages you might need too
+    sudo apt install build-essential curl git libmxml-dev
```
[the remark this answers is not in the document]
stevedlawrence: Another thought, we could make build setup instructions for different distos collapsable: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

That way it doesn't make the README huge by default, but has the information people need if they need it. For example:

## Build Setup

<details>
  <summary>Fedora</summary>
  
```
sudo dnf install ...
```
More stuff here
</details>

<details>
  <summary>Ubuntu/Debian</summary>
  
```
sudo apt install ...
```
More stuff here
</details>

<details>
  <summary>Windows</summary>
  
Windows Instructions here
</details>

tuxji (commented):
Have addressed reviewer requested changes.  Will wait for rebase on master branch to determine whether Windows checks really still fail.

README.md
```diff
@@ -45,94 +45,46 @@ For more information about Daffodil, see the [Website].
...
+See [BUILD.md](BUILD.md) for more details.
 
 ## Getting Started
 
-Below are some of the more common commands used for Daffodil development.
+[SBT] is the officially supported tool to build Daffodil.  Here are
+some of the more commonly used commands for Daffodil development.
 
-### Compile
+* Compile source code:
```
stevedlawrence, line 55: I personally perfer the ### headings over the bullets.  It makes it easier to skim over and find the command you might need, rather than having the full sentences.

build.sbt
```diff
@@ -68,8 +68,8 @@ lazy val runtime2         = Project("daffodil-runtime2", file("daffodil-runtime2
                               .settings(commonSettings)
                               .settings(publishArtifact in (Compile, packageDoc) := false)
                               .settings(
-                                Compile / cCompiler := "cc",
-                                Compile / ccArchiveCommand := "ar",
+                                Compile / cCompiler := sys.env.getOrElse("CC", "cc"),
+                                Compile / ccArchiveCommand := sys.env.getOrElse("AR", "ar"),
```
stevedlawrence, line 72: Is it worth mentiong the AR enironment variable in the BUILD setup readme?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -352,6 +353,38 @@ trait ElementBase
     } else DataValue.NoValue
   }
 
+  /**
+   * Is either None or Some(primTypeValue).
```
stevedlawrence, line 357: This comment isn't accurent. This is always a pimitive, but might be the special NoValue.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
```diff
@@ -18,8 +18,16 @@
...
 
-#include <stdio.h>    // for FILE, size_t
+#include <stdbool.h>  // for bool
 #include <stdint.h>   // for int64_t
+#include <stdio.h>    // for FILE, size_t
+
+enum Limits
+{
+    LIMIT_DIAGNOSTICS = 100,  // limits how many diagnostics can accumulate
+    LIMIT_NAME_LENGTH = 9999, // limits how long infoset names can become
```
stevedlawrence, line 28: This limit seems pretty huge. I would think we could make this smaller and reduce memory usage?

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryAbstractCodeGenerator.scala
stevedlawrence: I think this only supports fixed values where the primitive is a numeric? Are string primitives supported yet, or do we need a check here that SDE's if hasFixedValue but primitive type is not a numeric?

stevedlawrence (commented):
Some additional minor questions/comments on the updates.

BUILD.md
```diff
@@ -0,0 +1,63 @@
...
+prefer.
+
+Since Daffodil now has a C backend as well as a Scala backend, you
+will need a C compiler ([gcc] or [clang]), the [Mini-XML] library, and
+possibly the GNU [argp] library if your system's C library doesn't
+include it already.  The recommended way is to install them using your
+operating system's package manager (for example, `apt` on Debian or
+Ubuntu):
+
+    sudo apt install build-essential curl git libmxml-dev
```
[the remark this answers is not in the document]
tuxji: Mini-XML is mxml-dev in Fedora and libmxml-dev in Ubuntu; libxml is something completely different on both Fedora and Ubuntu.  If Fedora's mxml-dev is only 3.1, the API is missing a function which we expect to call so we still need to install Mini-XML from source there too.  I'll separate the build instructions for Fedora, Ubuntu, and Windows and make them more clear when I rebase the runtime2-2204 branch.

BUILD.md
```diff
@@ -0,0 +1,63 @@
...
+driver command.
+
+On Windows, the easiest way to install gcc and libargp is to install
+[MSYS2]'s collection of free tools and libraries although MSYS2 has no
+package for libmxml which you'll need to build from source.  First
+install [MSYS2] following its website's installation instructions,
+then run the following commands in a "MSYS2 MSYS" window:
+
+    pacman -S gcc git libargp-devel make pkgconf
+    git clone https://github.com/michaelrsweet/mxml.git
```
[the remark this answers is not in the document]
tuxji: Yes, I will put `-t v3.2` in the git clone command both here and in the CI workflow when I rebase the runtime2-2204 branch.

README.md
tuxji: OK, fair enough.  I'll put the ### headings back.

build.sbt
tuxji: Yes, if someone says CC=true, they probably need to say AR=true too.  I'll mention AR too.  Can you think of a clean way to skip the tests too?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
tuxji: I had noticed that the comment seemed inaccurate when I copied it too.  Yes, I'll update both places.

daffodil-runtime2/src/main/resources/c/libruntime/errors.h
tuxji: If you look at infoset.c which uses LIMIT_NAME_LENGTH, it's just a couple of functions with static char[9999] buffers and their storage are used only temporarily by the callers long enough to hand off to the Mini-XML library (which copies the actual, usually less than 20 characters long, strings immediately) or compare against strings from the Mini-XML library using strcmp.  Setting the limit to 9999 characters only uses static storage in a couple of places and reduces the chance that an infoset namespace or name might not fit in the static buffer.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/BinaryAbstractCodeGenerator.scala
tuxji: Runtime2 doesn't support strings yet, and we do need to consider how to handle fixed values that are strings (we'd have to SDE or call strcmp instead of using value == fixed).  I would expect that strings in binary data could have length fields as well as be terminated by a null byte.

tuxji (commented):
Thanks, I've updated my todos accordingly.

build.sbt
stevedlawrence: Not sure if there's a simple way to do it. If all that's really needed is to install a compiler and mini xml, I dont't think it's worth it. I think maybe if we get more runtimes or more devs focusing on different runtimes, it might be worth figuring out how to make things more pluggable so people can really only focus on runtimes they care about (as long as CI continues to test all runtimes). But for now, I dont' think it's a big deal. And it will also force people to test it, which is probably important until we get more people working on runtime2.

PR #527, opened 2021-04-12
Fix breakages due to JLine3 upgrade

stevedlawrence:
Commit 66dda260ce upgraded from JLine2 to JLine3, which resulted in some
subtle breakages:

- In JLine2, the readLine() function returned null when a user pressed
  Ctrl-C or Ctrl-D. JLine3 changed this behavior so that the function
  never returns null, but instead throws exceptions. This changes the
  CLIDebuggerRunner to not check for null and instead catch the
  appropriate exceptions.
- When the command completion code was rewritten, it compared against
  the "name" field of DebuggerCommand's to figure out which command a
  user had typed. This meant that if a user used short commands, e.g.
  "i <TAB>", then we would not list any potential completions. This
  fixes this by using == for the comparision, which compares against
  both the long and short command names.

DAFFODIL-2494

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/debugger/CLIDebuggerRunner.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/debugger/CLIDebuggerRunner.scala
@@ -24,9 +24,11 @@ import scala.io.Source
 
 import org.jline.reader.Candidate
 import org.jline.reader.Completer
+import org.jline.reader.EndOfFileException
 import org.jline.reader.LineReader
 import org.jline.reader.LineReaderBuilder
 import org.jline.reader.ParsedLine
+import org.jline.reader.UserInterruptException
 
 
 class CLIDebuggerRunner(cmdsIter: Iterator[String]) extends InteractiveDebuggerRunner {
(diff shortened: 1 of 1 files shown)


mbeckerle (approved):
+1

daffodil-cli/src/main/scala/org/apache/daffodil/debugger/CLIDebuggerRunner.scala
```diff
@@ -104,8 +111,10 @@ class CLIDebuggerCompleter(id: InteractiveDebugger) extends Completer {
           // We have the name for the next command, try to find the
           // associated subcommand of the current DebugCommand. If we don't
           // find one, it just means they user typed something that's not a
-          // valid command and we no command to use for completing
-          val nextCmd = cmd.subcommands.find { _.name == nextCmdName }
+          // valid command and we no command to use for completing. Note that
```
tuxji, line 114: Typo: "we [have] no command..."

tuxji (approved):
+1 

Looks good,  I'm assuming CI is just broken for some unrelated reason.

stevedlawrence:
Typo fixed.

I wonder if the CI failures are due to bintray being deprecated soon? The download links for some of the sbt plugin dependencies in the CI are looking at repo.scala-sbt.org/... and repo.typesafe.org/..., which are redirecting to a bintray.com/... link, and getting a "forbidden" error. Hopefully this is just temporary.

stevedlawrence:
Yeah, this is related to bintray deprecration. Bintray is having a brownout today and April 26, before ultimately shutting down May 1.

https://status.bintray.com/

Looks like no PR's are going to pass today :disappointed:

PR #529, opened 2021-04-13
WIP: Changes to support Java 16

stevedlawrence:
- Update GitHub actions to replace Java 15 with 16, 15 is no longer
  supported
- Remove ByteSource from the data dumper. This requires a certain kind
  of reflection that is no longer allowed in Java 16. Fortunately, the
  generality this provided was not actualy used--all places we dump data
  are trivially replaced with ByteBuffer's
- **WIP**: Java 16 changes the GZIPOutputStream to always output 0xFF for
  the GZIP header OS field, where it used to output 0x00. See Java bug
  JDK-8244706. There is no simple way to override this value to use the
  old behavior. This means when unparsing GZIP layers, the unparse has
  one byte difference under Java 16. This causes two layer tests to fail
  on Java 16. I have included a fix to the input data so the tests pass
  on Java 16, but now these tests will fail on other Java versions. Do
  we need separate tests that are only run and expected to pass on
  certain versions of Java? This is a WIP because the best way to handle
  this is not clear.

DAFFODIL-2484

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
@@ -43,7 +43,7 @@ class GZIPTransformer(layerLengthInBytesEv: LayerLengthInBytesEv)
   }
 
   override protected def wrapLayerEncoder(jos: java.io.OutputStream): java.io.OutputStream = {
-    val s = new java.util.zip.GZIPOutputStream(jos)
+    val s = GZIPFixedOutputStream(jos)
     s
   }
 
@@ -67,5 +67,86 @@ object GZIPTransformerFactory
 
     val xformer = new GZIPTransformer(maybeLayerLengthInBytesEv.get)
     xformer
+  }
+}
+
+
+object GZIPFixedOutputStream {
+
+  private val fixIsNeeded = !scala.util.Properties.isJavaAtLeast("16")
+
+  /**
+   * Create a GZIPOutputStream that, if necessary, proxies writes through an
+   * OutputStream that fixes inconsistencies between Java versions
+   */
+  def apply(os: java.io.OutputStream) = {
+    val fixedOS = if (fixIsNeeded) new GZIPFixedOutputStream(os) else os
+    new java.util.zip.GZIPOutputStream(fixedOS)
+  }
+}
+
+/**
+ * Prior to Java 16, the java.util.zip.GZIPOutputStream wrote a value of zero for
+ * the OS field in the header (byte index 9). In Java 16, this was changed to a
+ * value of 255 to better abide by the GZIP specification. Unfortunately, this
+ * means unparsed data using a GZIP layer might have a single byte difference,
+ * depending on the Java version used. This can lead to inconsistent behavior of
+ * test failures that expect a certain byte value.
+ *
+ * To resolve this issue, we create this GZIPFixedOutputStream. This should wrap
+ * the underlying OutputStream and be passed as the OutputStream to the
+ * GZIPOutputStream. When the GZIPOutputStream writes the 9th byte to this
+ * GZIPFixedOutputStream, this will always write a value of 255, making all Java
+ * versions prior to 16 consistent with Java 16+ behavior.
+ */
+class GZIPFixedOutputStream private (os: java.io.OutputStream) extends java.io.OutputStream {
+
+  /**
+   * The next byte position that byte will be written to. If this is negative,
+   * that means we have already fixed the output and everything should just
+   * pass straight through.
+   */
+  private var bytePosition = 0
+
+  override def close(): Unit = os.close()
+  override def flush(): Unit = os.flush()
+
+  override def write(b: Array[Byte], off: Int, len: Int): Unit = {
+    if (bytePosition < 0) {
+      // The bad byte has been fixed, pass all writes directly through to the
+      // underlying OutputStream. This may be more efficient than the default
+      // OutputStream write() function, which writes the bytes from his array
+      // one at a time
+      os.write(b, off, len)
+    } else {
+      // The bad byte has not been fixed yet. Unless a newer version of Java
+      // has made changes, the GZIPOutputStreamm will have passed in a 10 byte
+      // array to this function that includes the bad byte. Let's just write
+      // that array using the default write(array) method that writes these
+      // bytes one at a time and will call the write(int) method that will fix
+      // that byte. Calling write() one at a time is maybe inefficient but for
+      // such a small array it should not have a noticeable effect.
+      super.write(b, off, len)
+    }
+  }
+
+  override def write(b: Int): Unit = {
+    if (bytePosition < 0) {
+      // The bad byte has already been fixed, simply pass this byte through to
+      // the underlying OutputStream
+      os.write(b)
+    } else if (bytePosition < 9) {
+      // The bad byte has not been fixed, and we haven't reached it yet, simply
+      // pass this byte through and increment our byte position
+      os.write(b)
+      bytePosition += 1
+    } else if (bytePosition == 9) {
+      // This is the bad byte, it is a 0 on some Java versions. Write 255
+      // instead of to match Java 16+ behavior. Also, set bytePosition to -1 to
+      // signify that we have fixed the bad byte and that all other writes
+      // should just pass directly to the underlying OutputStream
+      os.write(255)
+      bytePosition = -1
+    }
   }
 }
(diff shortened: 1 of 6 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
```diff
@@ -383,7 +383,7 @@ jones,arya,cat,1986-02-19
 
 We gzip that, and then we must prepend that with the length (as a binary 4-byte int) before we base64 encode.
   -->
-      <tdml:documentPart type="text"  replaceDFDLEntities="true"><![CDATA[AAAAcx+LCAAAAAAAAAAtyUEKgCAQheG94E1mIDWittG+M0xpaNQIo5tuX0Kb98P7LioVjiTf3sn7%CR;%LF;K8CyzlqVO9UIkrcgFTYh9pnBTOOInUPba3XmyOX7WiEGlqfxgJ1B6xpzKEDyEOxUf7JoJq1e/RI4%CR;%LF;wXIAAAA=--END--]]></tdml:documentPart>
+      <tdml:documentPart type="text"  replaceDFDLEntities="true"><![CDATA[AAAAcx+LCAAAAAAAAP8tyUEKgCAQheG94E1mIDWittG+M0xpaNQIo5tuX0Kb98P7LioVjiTf3sn7%CR;%LF;K8CyzlqVO9UIkrcgFTYh9pnBTOOInUPba3XmyOX7WiEGlqfxgJ1B6xpzKEDyEOxUf7JoJq1e/RI4%CR;%LF;wXIAAAA=--END--]]></tdml:documentPart>
```
stevedlawrence, line 386: These two tests show the difference between the GZIPOutputStream in Java 16 and older versions of Java, as mentioned in the commit message. With these changes, these tests will only pass on Java 16.

Any ideas on how to best resolve this? Do we just accept this difference and add a TDML Runner feature similar to "implementation" that skips the test it isn't the right Java version?

stevedlawrence: I was thinking one alternative would be to create something like a ``GZIPFixerOutputStream`` that acts as a proxy betwen the GZIPOutputStream and the underlying output stream. This "fixer" could change the 10th byte to 255 and pass all other bytes through unchanged, so regardless of the java version we get consistent gzip output.

Might look something like this:
```scala
class GZIPFixerOutputStream(os: OutputStream) extends OutputStream {
  private val position = 0
  override def write(byte: Byte) {
    if (position == 10) os.write(255) else os.write(byte)
    postition += 1
  }
}

class GZIPTransformer {
  ...
  override protected def wrapLayerEncoder(jos: java.io.OutputStream): java.io.OutputStream = {
    val fixer = new GZIPFixerOutputStream(jos)
    val s = new java.util.zip.GZIPOutputStream(fix)
    s
  }
}
```
It's a bit of a hack, but at least we get consistent behavior regardless of the Java version?

tuxji: I think a layer which fixes the 10th byte's value for consistency seems like a good idea, but I'm not familiar with how layers are added and used.  Would this layer slow down performance a lot because of the single byte writes?  If there is a big performance hit, can the layer's use be limited to only these two tests?  

I don't think we need to change the TDML schema to allow running tests only on certain JDK versions.  We can add an if statement directly to the Scala test runner class methods which runs these tests so it runs either the pre-Java 16 test case or the Java 16 test case.

stevedlawrence: Yeah, writing bytes one at a time is probably a bad idea, and was more just a snippet to get some discussions going. I've added a commit that adds this new GZIP fixer thing with extra logic so that it is essentially just a passthrouh to the underlying OutputStream after bad byte as been fixed. So there first 10 bytes or so might be a little slower, and there's going to be little bit of overhead to proxy calls to the underlying OutputStream, but I suspect that's minimal compared to the work needed to actually compress data.

This does mean the GZIP layer is going to create slightly different data than before. If we really care, we could add a compatability tunable, and this fixer could set that 9th byte to 0 or 255 depending if you want the old or new Java behavior. Not sure if that's worth it though. But that seems like a bit of work for one byte that most people wont' care about. But at least we are now consistent regarldess of Java version, which is probably more important.

mbeckerle: So this behavior change for Java 16 is permanent, and the old behavior was clearly wrong, so we should adopt this new behavior going forward, and the fixer should be applied to Java releases prior to this change (actually, what revision introduced this fix, it may be earlier than 16?)

So we should change everything to adopt the Java 16 corrected behavior in the tests, and use the fixer for earlier versions of Java prior to 16 conditionally, based on detecting them. 

I agree a TDML runner enhancement doesn't seem needed.

Question: Does just changing tests to twoPass fix the tests? If we don't actually compare the gzipped data, but rather compare the infosets, as a twoPass test does, maybe that is the easiest thing to do.

mbeckerle: Ah... mistake. A twoPass test won't fix this, because it requires that a onePass test NOT succeed, but a onePass test will succeed on Java16, so that's not a fix.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
```diff
@@ -69,3 +69,84 @@ object GZIPTransformerFactory
...
+      // such a small array it should not have a noticeable effect.
+      super.write(b, off, len)
+    }
+  }
+
+  override def write(b: Int): Unit = {
+    if (bytePosition < 0) {
+      // The bad byte has already been fixed, simply pass this byte through to
+      // the underlying OutputStream
+      os.write(b)
```
tuxji, line 137: We have a check warning at this line saying this line was not covered by tests.  Why didn't the tests call this line?

tuxji (commented):
I think the changes are very close to getting my +1 at this point.  All of the tests are passing but a codecov warning has me concerned that the tests aren't calling an important line which we want to be tested.

daffodil-test/src/test/resources/org/apache/daffodil/layers/layers.tdml
stevedlawrence: Yes, this change is part of Java 16. I believe it's just considered a bug prior to Java 16. Here's the release notes where the change is mentioned: https://jdk.java.net/16/release-notes#JDK-8244706

> GZIPOutputStream Sets the GZIP OS Header Field to the Correct Default Value (JDK-8244706)
core-libs/java.util.jar
> Prior to JDK 16, GZIPOutputStream set the OS field in the GZIP header to 0 (meaning FAT filesystem), which does not match the default value specified in section 2.3.1.2 of the GZIP file format specification version 4.3 (RFC 1952).
>
> As of JDK 16, the GZIP OS Header Field is set to 255, which is the default value as defined in RFC 1952.

Take a look at the GZIPFixedOutputStream as part of this PR. This does what you suggest and fixes the output so that this byte is always 255, regardless of the java version.

mbeckerle (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/GZipTransformer.scala
stevedlawrence: I think I have an idea of what's going on.

The first thing GZIPOutputStream does is write the 10 byte gzip header. To do this, it first calls ``write(array)`` with an Array of size 10. The last of these bytes (index 9) is the bad byte we need to fix. BytePosition is 0, so we call ``super.write(b, off, len)``. That will call this ``write`` function for each byte, one by one.

For bytes 0-8 we'll fall into the ``if (bytePosition < 9)`` case. In that case, we'll send the byte through to the underlying OutputStream unchanged, and increment ``bytePosition``.

For last byte at index 9 (the bad byte), we'll fall into the ``if (bytePosition == 9)`` case. In this case, we will fix the bad byte and we'll set ``bytePositio``n to -1.

And now, because this was the last index in the array we called ``super.write`` on, we won't call this ``write(int)``  function again. And because the ``bytePosition`` is now -1, all future calls to ``write(array)`` will go straight to the underlying stream. So the only way this can get hit is if the GZIPOutputStream calls ``write(int)``. And looking through the code of the GZIPOutputSTream, I think this just never happens. It always writes from an array, so we always just pass it through directly.

So, because GZIPOutputStream calls ``write(array)`` for the header, and only the last byte of that array needs to be fixed, and because it never directly calls ``write(int)`` for anything else, we never hit this case. If the GZIPOutputStream ever changes so either of these aren't the case, we will be able to hit this case. So technically this isn't possible, but only because of the current behavior of GZIPOutputStream. It's possible that could change in the future.

tuxji: OK, sounds good!

tuxji (approved):
+1

PR #533, opened 2021-04-19
Added tests that verify bug is fixed.

mbeckerle:
Roger found a bug in 3.0.0, turns out it's already fixed in 3.1.0, probably due to a change made of late by @stevedlawrence 

This just adds a test for the specific case Roger reported.

DAFFODIL-2498


stevedlawrence (approved):

stevedlawrence:
+1

tuxji (approved):
+1

PR #535, opened 2021-04-20
More tests from user examples.

mbeckerle:
Another example of trailingEmptyStrict not checking when it should detect an error.
(From recent users mailing list.)

DAFFODIL-2217


stevedlawrence (approved):
+1

jw3 (approved):
:+1:

PR #536, opened 2021-04-21
WIP: Added -Y warning check options

mbeckerle:
These 4 were added.
    "-Ywarn-infer-any",
    "-Ywarn-inaccessible",
    "-Ywarn-nullary-unit",
    "-Ywarn-unused-import"

Unwise to merge this until we're at a stable point. 

Honestly the warn-nullary-unit change was a bunch of lines of code changed for basically no value. 
It doesn't require that the calls to the functions be changed to also have the parentheses.

I.e., it wants

`def init(): Unit = { ... }`

but you can still, without warning call this by just saying `init`. No parentheses. 

DAFFODIL-2145

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -192,7 +192,12 @@ def buildScalacOptions(scalaVersion: String) = {
     "-unchecked",
     "-Xfatal-warnings",
     "-Xxml:-coalescing",
-    "-Xfuture"
+    "-Xfuture",
+    "-Ywarn-infer-any",
+    "-Ywarn-inaccessible",
+    // "-Ywarn-nullary-unit", // we cannot use this. It interferes with the Uniform Access Principle.
+    // See https://stackoverflow.com/questions/7600910/difference-between-function-with-parentheses-and-without.
+    "-Ywarn-unused-import"
   )
 
   val scalaVersionSpecificOptions = CrossVersion.partialVersion(scalaVersion) match {
@@ -211,6 +216,8 @@ def buildScalacOptions(scalaVersion: String) = {
     commonOptions ++ scalaVersionSpecificOptions ++ javaVersionSpecificOptions
 }
 
+// Workaround issue that some options are valid for javac, not javadoc.
+// These javacOptions are for code compilation only. (Issue sbt/sbt#355)
 def buildJavacOptions() = {
   val commonOptions = Seq(
     "-Werror",
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
@@ -19,18 +19,14 @@ package org.apache.daffodil.io
 
 import java.io.ByteArrayOutputStream
 import java.math.{BigInteger => JBigInt}
-import java.math.{BigInteger => JBigInt}
 import java.nio.ByteBuffer
 import java.nio.channels.Channels
 import java.nio.charset.CoderResult
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardOpenOption
-
 import passera.unsigned.ULong
-
 import org.apache.commons.io.output.TeeOutputStream
-
 import org.apache.daffodil.equality._
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.processors.charset.BitsCharsetNonByteSizeEncoder
@@ -41,7 +37,6 @@ import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe.Nope
 import org.apache.daffodil.util.Maybe.One
 import org.apache.daffodil.util.MaybeULong
-
 
 sealed trait DOSState
 private[io] case object Active extends DOSState
@@ -1079,10 +1074,6 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
   override final def resetMaybeRelBitLimit0b(savedBitLimit0b: MaybeULong): Unit = {
     Assert.usage(isWritable)
     setMaybeRelBitLimit0b(savedBitLimit0b, true)
-  }
-
-  final override def validateFinalStreamState: Unit = {
-    // nothing to validate
   }
 
   final override def isAligned(bitAlignment1b: Int): Boolean = {
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataStreamCommon.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataStreamCommon.scala
@@ -98,14 +98,4 @@ trait DataStreamCommon {
    */
   def pastData(nBytesRequested: Int): ByteBuffer
   def futureData(nBytesRequested: Int): ByteBuffer
-
-  /**
-   * Called once after each parse operation to verify final invariants for
-   * the implementation.
-   *
-   * Use to perform checks such as that data structures held in pools are
-   * all returned before end of parse.
-   */
-  def validateFinalStreamState: Unit
-
 }
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/StringDataInputStreamForUnparse.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/StringDataInputStreamForUnparse.scala
@@ -83,6 +83,4 @@ final class StringDataInputStreamForUnparse
   override def hasData: Boolean = doNotUse
   override def skip(nBits: Long, finfo: FormatInfo): Boolean = doNotUse
   override def resetBitLimit0b(savedBitLimit0b: MaybeULong): Unit = doNotUse
-  // $COVERAGE-ON$
-  override def validateFinalStreamState: Unit = {} // does nothing
 }
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/oolag/OOLAG.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/oolag/OOLAG.scala
@@ -299,7 +299,7 @@ object OOLAG {
      * Unconditionally, evaluate the LV arg in order to insure all checks for this
      * object are performed.
      */
-    private def requiredEvaluationsAlways(lv: OOLAGValueBase) : Unit = {
+    private def requiredEvaluationsAlways(lv: OOLAGValueBase): Unit = {
       val accumPoint =
         if (this.hasOOLAGRootSetup)
           oolagRoot
@@ -372,7 +372,7 @@ object OOLAG {
      * Saves the arg LV, and insures it is evaluated later only if
      * setRequiredEvaluationActive() is called for this object.
      */
-    private def requiredEvaluationsIfActivated(lv: OOLAGValueBase) : Unit = {
+    private def requiredEvaluationsIfActivated(lv: OOLAGValueBase): Unit = {
       if (requiredEvalStatus eq Active)
         if (hasOOLAGRootSetup)
           oolagRoot.requiredEvalFunctions +:= lv // active. Rooted. Accumulate centrally.
@@ -391,7 +391,7 @@ object OOLAG {
      * expression/LV, and if so, all the conditional evaluations will be captured and
      * evaluated as part of the ongoing evaluation of requiredEvaluations expressions.
      */
-    final def setRequiredEvaluationsActive() : Unit = setRequiredEvaluationsActiveOnceOnly
+    final def setRequiredEvaluationsActive(): Unit = setRequiredEvaluationsActiveOnceOnly
     private lazy val setRequiredEvaluationsActiveOnceOnly = {
       requiredEvalStatus = Active
       centralizeEvalFunctionsWhenReady()
@@ -410,7 +410,7 @@ object OOLAG {
      * onto the root, which happens automatically via the calls to
      * centralizeEvalFunctionsWhenReady().
      */
-    private def checkErrors: Unit = {
+    final def checkErrors(): Unit = {
       Assert.usage(this.isOOLAGRoot || requiredEvalFunctions == Nil)
       while (oolagRoot.requiredEvalFunctions != Nil) { // while there is an accumulated crop of eval functions
         // grab the current crop of eval functions
@@ -639,7 +639,7 @@ object OOLAG {
       value_ = Maybe(res)
     }
 
-    protected final def oolagFinalize = {
+    protected final def oolagFinalize() = {
       setIndent(indent - 2)
       Logger.log.trace(" " * indent + s"pop: ${thisThing}")
       if (oolagContext.currentOVList.nonEmpty)
--- a/daffodil-lib/src/test/scala/passera/test/UnsignedCheck.scala
+++ b/daffodil-lib/src/test/scala/passera/test/UnsignedCheck.scala
@@ -30,14 +30,12 @@ import org.scalacheck._
 // import org.scalacheck.ConsoleReporter
 import org.scalacheck.Prop._
 import passera.unsigned._
-
 import org.junit.Test
 import org.junit.Assert._
 
 import scala.language.implicitConversions
 
 class UnsignedCheck {
-  import Gen._
   import Arbitrary.arbitrary
 
   val zero = 0.toUInt
@@ -52,7 +50,7 @@ class UnsignedCheck {
   def genUInt: Gen[UInt] = for (n <- arbitrary[Int]) yield UInt(n)
   implicit def arbUInt: Arbitrary[UInt] = Arbitrary(genUInt)
 
-  @Test def testIntToString = {
+  @Test def testIntToString() = {
     assertTrue(
       forAll { n: Int => n >= 0 ==> (n.toUInt.toString == n.toString) }
     )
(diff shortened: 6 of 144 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/oolag/OOLAG.scala
```diff
@@ -308,7 +308,7 @@ object OOLAG extends Logging {
      * Unconditionally, evaluate the LV arg in order to insure all checks for this
      * object are performed.
      */
-    private def requiredEvaluationsAlways(lv: OOLAGValueBase) : Unit = {
+    private def requiredEvaluationsAlways(lv: OOLAGValueBase) (): Unit = {
```
stevedlawrence, line 311: A few of the changes in this file look incorrect--looks like theres an extra set of parens. Is this a accidental find/replace due to the space before the colon?

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/PropertyGenerator.scala
```diff
@@ -595,7 +595,7 @@ object Currency {
   def generateNonEnumStringInit(pgName: String, propInits: Seq[String]) = {
     val initFuncName = initialLowerCase(pgName)
     """
-  def """ + initFuncName + """Init() : Unit = {""" +
+  def """ + initFuncName + """Init() (): Unit = {""" +
```
stevedlawrence, line 598: This looks off too, I think this ends up generated an extra set of parens.

daffodil-propgen/src/test/scala/org/apache/daffodil/propGen/TestPropertyGenerator.scala
```diff
@@ -65,7 +65,7 @@ class TestPropertyGenerator {
     val mx = pg.genAttributeGroup(sch)
     assertTrue(mx.contains(""" = convertToBoolean(findProperty("prefixIncludesPrefixLength").value)"""))
     assertTrue(mx.contains("""LengthKindMixin"""))
-    assertTrue(mx.contains("""def lengthPropertiesAGInit() : Unit = {"""))
+    assertTrue(mx.contains("""def lengthPropertiesAGInit() (): Unit = {"""))
```
stevedlawrence, line 68: Same issue here

stevedlawrence (approved):
+1, maybe a couple accidental find/replaces that aren't quiet right?

daffodil-propgen/src/test/scala/org/apache/daffodil/propGen/TestPropertyGenerator.scala
mbeckerle: Yup. I'm counting on the CI to find the bugs for me.

stevedlawrence: Looks likethe CI passed, I guess this is technically valid syntax since scala allows multiple parameter lists for currying, and I guess it has no problem with an empty extra parameter list?

mbeckerle: Strange. But yes you can 
`def foo(): Unit = { .... }`
then invoke with just
`foo` not `foo()`
And scala supplies the implied added '()'. So it seems it can do this even for
`def foo()(): Unit = { .... }`
which invokes with 
`foo()`, again supplying the second added '()' implicitly. 

Grrrr. Feels like a bug, not a feature. 

I'll chase down all ") ()". 

I've been thinking about Scala 2.13 and Scala 3.0 eventually, but unless we cross-support them (like we did for 2.11 and 2.12 for a time) so that 2.12 continues to be supported, we're stuck because other important Scala libraries like Apache Spark are still on 2.12.

In other words, I suspect we have to turn on these other warnings that we need for code-quality again without waiting for new scala versions. I.e., in Scala 2.12, and cope with them best we can. E.g., I think the warn on unused and warn on dead-code are both pretty important things for the compiler to do for us. 

A wildcard worthy of investigation is whether the Java modules stuff (introduced in Java 9 I believe, and which seems not to be the OGSI stuff all over again) allows one to link together modules with different dependencies, so long as the crossing between them is a a Java API.

jw3:
Using Scalafix makes some of these operations less painful... but some others more painful.

https://scalacenter.github.io/scalafix/

see #388 #393

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -386,7 +386,7 @@ final class UStateForSuspension(
   override def advanceAccessor: InfosetAccessor = die
   override def inspect: Boolean = die
   override def inspectAccessor: InfosetAccessor = die
-  override def fini: Unit = {
+  override def fini(): Unit = {
```
mbeckerle, line 389: This finalize method not being called suggests it is not needed. Could be dead code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -530,7 +530,7 @@ final class UStateMain private (
   override def advanceAccessor: InfosetAccessor = inputter.advanceAccessor
   override def inspect: Boolean = inputter.inspect
   override def inspectAccessor: InfosetAccessor = inputter.inspectAccessor
-  override def fini: Unit = { inputter.fini }
+  override def fini(): Unit = { inputter.fini }
```
mbeckerle, line 533: ditto. Dead code?

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -1078,7 +1080,7 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
     setMaybeRelBitLimit0b(savedBitLimit0b, true)
   }
 
-  final override def validateFinalStreamState: Unit = {
+  final override def validateFinalStreamState(): Unit = {
```
mbeckerle, line 1083: This not being called indicates a bug or dead code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DFDLDelimiter.scala
```diff
@@ -445,7 +445,7 @@ class CharDelim(val char: Char, ignoreCase: Boolean) extends DelimBase {
   }
 
   lazy val typeName = "CharDelim"
-  def print = {
+  def print() = {
```
mbeckerle, line 448: Dead code - left over from debugging. Remove or comment out fully? Ditto for all other print() methods in this file.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DFDLDelimiter.scala
```diff
@@ -412,7 +412,7 @@ class Delimiter {
 
 abstract class DelimBase extends Base {
   def typeName: String
-  def print: Unit
+  def print(): Unit
```
mbeckerle, line 415: Might be left over debug method. Removed?

tuxji (approved):
+1

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
mbeckerle: TDML runner explicitly runs both regular and sax APIs.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: Not dead.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
mbeckerle: Not dead code.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataStreamCommon.scala
```diff
@@ -98,14 +98,4 @@ trait DataStreamCommon {
...
   def futureData(nBytesRequested: Int): ByteBuffer
-
-  /**
-   * Called once after each parse operation to verify final invariants for
-   * the implementation.
-   *
-   * Use to perform checks such as that data structures held in pools are
-   * all returned before end of parse.
-   */
-  def validateFinalStreamState: Unit
```
tuxji, line 109: Are you removing this code for a valid reason?

daffodil-io/src/main/scala/org/apache/daffodil/io/StringDataInputStreamForUnparse.scala
```diff
@@ -84,5 +84,4 @@ final class StringDataInputStreamForUnparse
   override def skip(nBits: Long, finfo: FormatInfo): Boolean = doNotUse
   override def resetBitLimit0b(savedBitLimit0b: MaybeULong): Unit = doNotUse
   // $COVERAGE-ON$
-  override def validateFinalStreamState: Unit = {} // does nothing
```
tuxji, line 87: If this removal is valid. also remove the `// $COVERAGE-ON$`.

daffodil-lib/src/test/scala/passera/test/UnsignedCheck.scala
```diff
@@ -30,14 +30,12 @@ import org.scalacheck._
...
 import org.scalacheck.Prop._
 import passera.unsigned._
-
 import org.junit.Test
 import org.junit.Assert._
 
 import scala.language.implicitConversions
 
 class UnsignedCheck {
-  import Gen._
```
tuxji, line 40: Removing an import should be safe as long as the code still compiles, but where is Gen defined?  I ran `rg -uu -w Gen` over my entire fully built checkout and it found only three lines, none of which appears to define Gen:

```
daffodil-lib/src/test/scala/passera/test/UnsignedCheck.scala
40:  import Gen._
52:  def genUInt: Gen[UInt] = for (n <- arbitrary[Int]) yield UInt(n)
61:  val nonNegLong = Gen.choose(0L, 0x00000000ffffffffL)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
```diff
@@ -223,7 +223,7 @@ class DPathCompileInfo(
     with PreSerialization
     with HasSchemaFileLocation {
 
-  def initialize: Unit = {
+  lazy val initialize: Unit = {
     parents
   }
```
tuxji, line 228: Why this change, since I don't see any computation being done?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -397,34 +397,31 @@ final class UStateForSuspension(
...
+  // $COVERAGE-OFF$
   override def advance: Boolean = die
   override def advanceAccessor: InfosetAccessor = die
   override def inspect: Boolean = die
   override def inspectAccessor: InfosetAccessor = die
-  override def fini: Unit = {
-    //do nothing
-  }
+  override def fini(): Unit = die
   override def inspectOrError = die
```
tuxji, line 406: Why does `fini()` get parentheses but `inspectOrError` doesn't get parentheses?

tuxji (changes requested):
I noticed you asked for another review and I saw some old code was carried forward in build.sbt, so you need to fix that file.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -1081,10 +1076,6 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
     setMaybeRelBitLimit0b(savedBitLimit0b, true)
   }
 
-  final override def validateFinalStreamState: Unit = {
-    // nothing to validate
-  }
```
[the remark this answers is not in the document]
mbeckerle: Dead code. Base is never called.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataStreamCommon.scala
mbeckerle: Dead code. Base is never called.

daffodil-io/src/main/scala/org/apache/daffodil/io/StringDataInputStreamForUnparse.scala
mbeckerle: dead code. Will fix.

daffodil-lib/src/test/scala/passera/test/UnsignedCheck.scala
mbeckerle: IntelliJ removed this automatically I assume. 
Gen is defined in scalacheck/scalacheck_2.12/scalacheck_2.12-1.15.4-sources.jar!/org/scalacheck/Gen.scala

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: Yeah. I think this should be without the (). Clearly scala is not warning about this. It was warning about it at one point.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala
tuxji: I found a StackOverflow [answer](https://stackoverflow.com/questions/7600910/difference-between-function-with-parentheses-and-without) which said that calling nullary methods without an explicit () is no longer optional in Scala 2.13 or 3.0.  BTW, the change wasn't made to annoy people; it was a necessary side effect of another change to improve type inference by allowing eta expansion of Scala functions.  What this tells me is that you shouldn't bother changing `initialize` from a def to a lazy val.  When we upgrade to 2.13, then scalac will enforce consistency between calls and definitions and we can fix any remaining calls without () at that time.

mbeckerle: This is super helpful. So use () when the method has side effects, but no args. Leave out parens if it is side-effect free. 

Note that a lazy val that only reads things is still considered side-effect free, even though it has this lazy-val once-only behavior. So if you define a def, intending to possibly override with a lazy val, then no parameters is correct, as a lazy val can't have even an empty argument list. 

Also, **_conceptual_** side-effect free applies. Initializing something that happens once as part of setup of some sort is NOT considered a side effect. 

A side-effect is something that has to happen *every* time the method is used. E.g, like assigning something each time, or doing an I/O operation each time.

mbeckerle: Actually, based on this. We really can't use "-Ywarn-nullary-unit", because it forces us to use () for def that want to be overridden as lazy val.

mbeckerle: This code is back to being a def with no args.

build.sbt
```diff
@@ -157,6 +157,25 @@ lazy val commonSettings = Seq(
...
+    "-Ywarn-infer-any",
+    "-Ywarn-inaccessible",
+    "-Ywarn-nullary-unit",
+    "-Ywarn-unused-import"
+  ),
+  // Workaround issue that some options are valid for javac, not javadoc.
+  // These javacOptions are for code compilation only. (Issue sbt/sbt#355)
+  Compile / compile / javacOptions  ++= Seq(
+    "-Werror",
+    "-Xlint:deprecation"
```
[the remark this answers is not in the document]
mbeckerle: fixed

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -397,34 +397,31 @@ final class UStateForSuspension(
...
-
   override def pushDelimiters(node: DelimiterStackUnparseNode) = die
   override def popDelimiters() = die
+  // $COVERAGE-ON$
+
+  override def groupPos = 0 // was die, but this is called when copying state during debugging
+  override def currentInfosetNodeMaybe = Maybe(currentInfosetNode)
+  override def arrayPos = occursIndex
+  override def childPos = 0 // was die, but this is called when copying state during debugging.
+
```
tuxji, line 424: I looked at why groupPos and childPos's lines (420, 423) still aren't covered by tests (their original locations were inside the $COVERAGE-OFF$ block).  My best guess is that we don't have a test forcing the interactive debugger to copy unparser state.  I'll still give you a +1.

tuxji (approved):
+1

PR #539, opened 2021-04-24
Fix TCP/network input stream hangs on isAtEnd

mbeckerle:
Change isAtEnd contract. Does not look ahead to see if next byte/bit causes the stream to hit EOF.
That behavior is incompatible with network TCP streams of input.

Added test rig and tests showing isAtEnd hangs on network
TCP stream when no bytes are available at that time.

Cleanup: InputSource now throws a very specific exception in the case
where an InputStream provided by caller misbehaves.

Cleanup: Exception class derivations improved to
allow cause or message, not only messages.

Cleanup: Coroutines improved to allow the kinds of objects passed to, and
returned (from resume or resumeFinal) to be different for
different coroutines.

Cleanup: Missing tests - improve coverage.
Coroutines unit tests added back (these had been missing).

Cleanup: Layering - fixed bad available() implementation.

DAFFODIL-2502, DAFFODIL-2503, DAFFODIL-2504

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/api/TestParseIndividualMessages.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/api/TestParseIndividualMessages.scala
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.api
+
+import org.apache.daffodil.Implicits.intercept
+import org.apache.daffodil.io.SocketPairTestRig
+import org.apache.daffodil.util.SchemaUtils
+import org.apache.daffodil.util.TestUtils
+import org.junit.Assert._
+import org.junit.Test
+
+import java.io.InputStream
+import java.io.OutputStream
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.DurationInt
+import scala.xml.Node
+
+
+/**
+ * Shows that we can parse exactly 1 message from a TCP network socket
+ * without blocking for bytes past the end of the messsage.
+ *
+ * This only works for DFDL schemas of formats that are specified length.
+ */
+class TestParseIndividualMessages {
+
+  //
+  // DFDL schema for element e1 which occupies exactly 4 bytes.
+  //
+  val exactly4ByteSch = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+      <dfdl:format representation="binary" byteOrder="bigEndian" binaryNumberRep="binary" ref="tns:GeneralFormat"/>,
+      <xs:element name="e1" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="4"/>)
+
+  /**
+   * Test shows that at least for simple fixed-length data, Daffodil parse returns
+   * without requiring more bytes to be read than the exact length required.
+   */
+  @Test def testDaffodilParseFromNetwork1(): Unit = {
+    val sptr = new SocketPairTestRig {
+      override def test(pos: OutputStream, cis: InputStream): Unit = {
+
+        val dp = TestUtils.compileSchema(exactly4ByteSch)
+
+        //
+        // Write exactly 4 bytes to producer network stream
+        //
+        pos.write("1234".getBytes)
+        pos.flush()
+
+        //
+        // Daffodil parse element e1 from consumer input stream
+        //
+        // If we need more than 4 bytes to successfully parse (we shouldn't for this schema)
+        // then this will hang, because only 4 bytes are in fact available.
+        //
+        // Caution: if debugging, this will timeout if you stop inside here!
+        //
+        val (pr: DFDL.ParseResult, xml: Node) =
+        SocketPairTestRig.withTimeout("Daffodil parse") {
+          TestUtils.runDataProcessorOnInputStream(dp, cis, areTracing = false)
+        }
+
+        assertFalse(pr.isError)
+        assertEquals("1234", xml.text)
+        assertEquals(33, pr.resultState.currentLocation.bitPos1b)
+      }
+    }
+    sptr.run()
+  }
+
+  //
+  // DFDL schema for delimited element.
+  //
+  private def delimitedSchema(term: String) = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+      <dfdl:format representation="text" ref="tns:GeneralFormat"/>,
+      <xs:element name="e1" type="xs:string" dfdl:lengthKind="delimited"
+                  dfdl:terminator={ term } />)
+
+  /**
+   * Helper so we can test various delimiter-oriented scenarios.
+   *
+   * @param stringData          Data to write. Should be small enough that the parse will block.
+   * @param terminator          String to use as terminating delimiter of element. Can be more than one delimiter.
+   * @param followingDataString Data to write which should unblock the parse.
+   */
+  private def testHelperDaffodilParseDelimitedFromNetwork(
+    data: String,
+    terminator: String,
+    followingDataString: String) = {
+    val sptr = new SocketPairTestRig {
+      override def test(pos: OutputStream, cis: InputStream): Unit = {
+
+        val dp = TestUtils.compileSchema(delimitedSchema(terminator))
+
+        // Write the data. Should be too short to satisfy the parse.
+        //
+        pos.write(data.getBytes)
+        pos.flush()
+
+        val fut = Future {
+          TestUtils.runDataProcessorOnInputStream(dp, cis, areTracing = false)
+        }
+
+        Thread.sleep(100)
+        assertFalse(fut.isCompleted)
+        //
+        // Writing additional character(s) should unblock the parse.
+        //
+        pos.write(followingDataString.getBytes)
+        pos.flush()
+        val (pr, xml) = Await.result(fut, 100.milliseconds)
+        if (!pr.isError) {
+          assertEquals("1234", xml.text)
+        } else {
+          //parse failed.
+          val diagString = pr.getDiagnostics.map {
+            _.getMessage()
+          }.mkString("\n")
+          fail("Parse failed, but did not time-out.\n" + diagString)
+        }
+      }
+    }
+    sptr.run()
+  }
+
+  /**
+   * This test fails (and so is commented out) but *should* pass.
+   *
+   * Test (when it works) shows that for delimited data, we block seeking more data,
+   * but once the need for a terminator with longest match is satisfied
+   * the parse is unblocked.
+   */
+  // @Test // DAFFODIL-2504 - this fails looking for more data than is needed for terminator.
+  def testDaffodilParseFromNetworkDelimited1(): Unit = {
+    intercept[TimeoutException] {
+      testHelperDaffodilParseDelimitedFromNetwork("1234", "$", "$")
+    }
+    fail("if we get here, then we intercepted a TimeoutException, which means Daffodil was hung.")
+  }
+
+  /**
+   * This test works, and it should work.
+   *
+   * It characterizes that currently we need to be able to read
+   * not only the terminator, but 7 more characters, in order for the reads associated with the delimiter
+   * scanning to be satisfied.
+   *
+   * We're not supposed to need 7 characters though. Only 1 character should do it.
+   */
+  @Test // DAFFODIL-2504
+  def testDaffodilParseFromNetworkDelimited1b(): Unit = {
+    testHelperDaffodilParseDelimitedFromNetwork("1234", "$",
+      "$1234567")
+  }
+
+  /**
+   * This test fails (and so is commented out) but *should* pass.
+   *
+   * Test (when it works) shows that for delimited data, we block seeking more data,
+   * but once the need for a terminator with longest match is satisfied
+   * the parse is unblocked.
+   *
+   * This variant has 2 options for terminator, $ or $$, and one is a prefix of the other
+   * So this test insures that getting a shorter delimiter match doesn't unblock
+   * when getting more data might match a longer delimiter.
+   *
+   * The test then subsequently provides the character for that longer delimiter
+   * which should unblock daffodil.
+   */
+  // @Test // DAFFODIL-2504 - this fails looking for more data than is needed for terminator.
+  def testDaffodilParseFromNetworkDelimited2(): Unit = {
+    intercept[TimeoutException] {
+      testHelperDaffodilParseDelimitedFromNetwork("1234$", "$ $$", "$")
+    }
+    fail("If we get here, we intercepted a TimeoutException, which means Daffodil is hung.")
+  }
+
+  /**
+   * This test works, and it should work.
+   *
+   * It characterizes that currently we need to be able to read
+   * not only the longest possible terminator, but 7 more characters,
+   * in order for the reads associated with the delimiter
+   * scanning to be satisfied.
+   *
+   * We're not supposed to need 7 characters though. Only 1 character should do it.
+   */
+  @Test // DAFFODIL-2504 - this shouldn't require 7 more characters.
+  def testDaffodilParseFromNetworkDelimited2b(): Unit = {
+    testHelperDaffodilParseDelimitedFromNetwork("1234$", "$ $$", "$1234567")
+  }
+
+}
--- a/daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
@@ -22,15 +22,13 @@ import java.io.FileNotFoundException
 import java.nio.channels.Channels
 import java.nio.channels.ReadableByteChannel
 import java.nio.channels.WritableByteChannel
-
 import scala.util.Try
 import scala.xml._
-
 import org.apache.commons.io.output.NullOutputStream
-
 import org.junit.Assert.assertEquals
+import org.apache.daffodil.Implicits._
 
-import org.apache.daffodil.Implicits._; object INoWarnU2 { ImplicitsSuppressUnusedImportWarning() }
+import java.io.InputStream
 import org.apache.daffodil.api.DFDL
 import org.apache.daffodil.api._
 import org.apache.daffodil.compiler.Compiler
@@ -47,6 +45,8 @@ import org.apache.daffodil.processors.DataProcessor
 import org.apache.daffodil.processors.VariableMap
 import org.apache.daffodil.xml.XMLUtils
 import org.apache.daffodil.xml._
+
+object INoWarnU2 { ImplicitsSuppressUnusedImportWarning() }
 
 /*
  * This is not a file of tests.
@@ -82,7 +82,7 @@ object TestUtils {
   }
 
   def testString(testSchema: Node, data: String, areTracing: Boolean = false) = {
-    runSchemaOnData(testSchema, Misc.stringToReadableByteChannel(data), areTracing)
+    runSchemaOnRBC(testSchema, Misc.stringToReadableByteChannel(data), areTracing)
   }
 
   def testBinary(testSchema: Node, hexData: String, areTracing: Boolean = false): (DFDL.ParseResult, Node) = {
@@ -92,11 +92,11 @@ object TestUtils {
 
   def testBinary(testSchema: Node, data: Array[Byte], areTracing: Boolean): (DFDL.ParseResult, Node) = {
     val rbc = Misc.byteArrayToReadableByteChannel(data)
-    runSchemaOnData(testSchema, rbc, areTracing)
+    runSchemaOnRBC(testSchema, rbc, areTracing)
   }
 
   def testFile(testSchema: Node, fileName: String) = {
-    runSchemaOnData(testSchema, Misc.fileToReadableByteChannel(new java.io.File(fileName)))
+    runSchemaOnRBC(testSchema, Misc.fileToReadableByteChannel(new java.io.File(fileName)))
   }
 
   val useSerializedProcessor = true
@@ -182,7 +182,7 @@ object TestUtils {
     } else p
   }
 
-  def runSchemaOnData(testSchema: Node, data: ReadableByteChannel, areTracing: Boolean = false) = {
+  def compileSchema(testSchema: Node) = {
     val compiler = Compiler()
     val pf = compiler.compileNode(testSchema)
     val isError = pf.isError
@@ -191,19 +191,34 @@ object TestUtils {
     if (isError) {
       throw new Exception(msgs)
     }
-    var p = saveAndReload(pf.onPath("/").asInstanceOf[DataProcessor])
+    val p = saveAndReload(pf.onPath("/").asInstanceOf[DataProcessor])
     val pIsError = p.isError
     if (pIsError) {
       val msgs = pf.getDiagnostics.map(_.getMessage()).mkString("\n")
       throw new Exception(msgs)
     }
-    p = if (areTracing) {
-      p.withDebugger(builtInTracer).withDebugging(true)
-    } else p
-    p = p.withValidationMode(ValidationMode.Limited)
+    p
+  }
+
+  def runSchemaOnRBC(testSchema: Node, data: ReadableByteChannel, areTracing: Boolean = false): (DFDL.ParseResult, Node) = {
+    runSchemaOnInputStream(testSchema, Channels.newInputStream(data), areTracing)
+  }
+
+  def runSchemaOnInputStream(testSchema: Node, is: InputStream, areTracing: Boolean = false): (DFDL.ParseResult, Node) = {
+    val p = compileSchema(testSchema)
+    runDataProcessorOnInputStream(p, is, areTracing)
+  }
+
+  def runDataProcessorOnInputStream(dp: DataProcessor, is: InputStream, areTracing: Boolean = false): (DFDL.ParseResult, Node) = {
+    val p1 =
+      if (areTracing) {
+        dp.withDebugger(builtInTracer).withDebugging(true)
+      } else dp
+
+    val p = p1.withValidationMode(ValidationMode.Limited)
 
     val outputter = new ScalaXMLInfosetOutputter()
-    val input = InputSourceDataInputStream(Channels.newInputStream(data))
+    val input = InputSourceDataInputStream(is)
     val actual = p.parse(input, outputter)
     if (actual.isProcessingError) {
       val diags = actual.getDiagnostics
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
@@ -18,11 +18,12 @@
 package org.apache.daffodil.io
 
 import java.nio.ByteBuffer
-
 import scala.collection.mutable.ArrayBuffer
-
 import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.exceptions.ThinException
+import org.apache.daffodil.exceptions.UnsuppressableException
+
+import java.io.InputStream
 
 /**
  * There is a finite limit to the distance one can backtrack which is given by
@@ -54,6 +55,27 @@ case class BacktrackingException(position: Long, maxBacktrackLength: Int)
   extends ThinException("Attempted to backtrack to byte %d, which exceeds maximum backtrack length of %d", position, maxBacktrackLength)
 
 /**
+ * Thrown in the specific case where a java.io.InputStream is not properly
+ * implemented and is returning 0 from the read(buf, off, len) call when len > 0.
+ *
+ * This non-blocking behavior is not supported by java.io.InputStream's contract
+ * and Daffodil depends on InputStreams having only blocking behavior.
+ *
+ * To properly place blame for this error on the InputStream (and not Daffodil's
+ * I/O layer built on top of it) we throw this very specific, informative
+ * exception in this case.
+ * @param inputStream The stream that is misbehaving.
+ */
+class InputStreamReadZeroError(inputStream: InputStream)
+extends UnsuppressableException(
+  s"""InputStream ${ inputStream.toString } illegally
+  | returned 0 from a call to read(buf, off, len).
+  | This is illegal behavior from a java.io.InputStream instance, as InputStream is a blocking API.
+  | This is not a Daffodil bug, but a problem with the InputStream supplied
+  | to Daffodil as a data source.""".stripMargin){
+}
+
+/**
  * The InputSource class is really just a mechanism to provide bytes an
  * InputSourceDataInputStream, which does the heavily lift about converter
  * bits/bytes to numbers and characters. This class does not need to know
@@ -66,6 +88,15 @@ case class BacktrackingException(position: Long, maxBacktrackLength: Int)
  * marks with random access.
  */
 abstract class InputSource {
+
+  /**
+   * Determine if the InputSource has encountered the end-of-data.
+   *
+   * This does NOT perform a read operation (which would be blocking), but just
+   * answers the question of whether prior read operations in fact encountered
+   * the -1 indicating end-of-data
+   */
+  def hasReachedEndOfData: Boolean
 
   /**
    * Determine whether the underlying data has the specified number of bytes
@@ -245,6 +276,12 @@ class BucketingInputSource(
    */
   private var bytesFilledInLastBucket: Int = 0
 
+  private var hasMoreData = true
+
+  override def hasReachedEndOfData: Boolean =
+    !hasMoreData
+
+
   /**
    * Adds new buckets to the buckets array until either we run out of data or
    * we fill up to byteIndex bytes in the bucketIndex. This modifies
@@ -255,7 +292,7 @@ class BucketingInputSource(
    */
   private def fillBucketsToIndex(goalBucketIndex: Long, bytesNeededInBucket: Long): Boolean = {
     var lastBucketIndex = buckets.length - 1
-    var hasMoreData = true
+
     var needsMoreData = goalBucketIndex > lastBucketIndex || (goalBucketIndex == lastBucketIndex && bytesNeededInBucket > bytesFilledInLastBucket)
 
     while (needsMoreData && hasMoreData) {
@@ -263,10 +300,21 @@ class BucketingInputSource(
       // actually needed
       val emptyBytesInLastBucket = bucketSize - bytesFilledInLastBucket
 
+      // we never call read passing len of 0.
+      Assert.invariant(emptyBytesInLastBucket > 0)
+
       // Try to read enough bytes to fill the rest of this bucket. Note that
       // the .read() function could hit EOF (returns -1) or could return
-      // anywhere from zero to emptyBytesInLastBucket bytes
-      val bytesRead = inputStream.read(buckets(lastBucketIndex).bytes, bytesFilledInLastBucket, emptyBytesInLastBucket)
+      // anywhere from 1 to emptyBytesInLastBucket bytes
+      val bytesRead =
+        inputStream.read(
+          buckets(lastBucketIndex).bytes,
+          bytesFilledInLastBucket,
+          emptyBytesInLastBucket)
+
+      // check for bad inputStream behavior. It's not our fault!
+      if (bytesRead == 0)
+        throw new InputStreamReadZeroError(inputStream)
 
       if (bytesRead == -1) {
         // Needed more data but hit EOF, break out with error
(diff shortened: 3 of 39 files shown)


daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
```diff
@@ -191,19 +189,33 @@ object TestUtils {
...
+  }
+
+  def runDataProcessorOnInputStream(dp: DataProcessor, is: InputStream, areTracing: Boolean = false): (DFDL.ParseResult, Node) = {
+    var p = dp
     p = if (areTracing) {
-      p.withDebugger(builtInTracer).withDebugging(true)
-    } else p
+        p.withDebugger(builtInTracer).withDebugging(true)
+      } else p
     p = p.withValidationMode(ValidationMode.Limited)
```
mbeckerle, line 215: indentation off?

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -216,7 +216,9 @@ class JavaIOInputStream(s: InputSourceDataInputStream, finfo: FormatInfo)
     }
   }
 
-  override def available(): Int = 1
+  override def available(): Int = {
+    Assert.usageError("Not to be called on JavaIOInputStream.")
```
mbeckerle, line 220: This is causing failures of unit tests.
I don't understand the prior definition of 1. That means there is always expectation that asking for 1 more byte will not hang, doesn't seem to meet the API definition of available(). 
Not sure we want to be doing a blocking read here so we can return 0 if there is no more data, or not.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -288,7 +285,8 @@ final class PState private (
   }
 
   def currentLocation: DataLocation = {
-    val isAtEnd = !dataInputStream.isDefinedForLength(1)
+    // val isAtEnd = !dataInputStream.isDefinedForLength(1)
+    val isAtEnd = dataInputStream.isAtEnd()
```
mbeckerle, line 289: Arguably, isAtEnd should not be captured into this object at this point, and frozen. 

Because it could be false, yet a subsequent operation on the InputSourceDataInputStream to look and see if there is more data could get back End-of-Data, which would change the answer, that yes, in fact, the parse did end at end of data.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -837,7 +837,9 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
     roundTrip: RoundTrip,
     implString: Option[String]) = {
 
-    val nBits = optLengthLimitInBits.get
+    val nBits = optLengthLimitInBits.getOrElse{
+      Assert.invariantFailed("TDML tests should always have a length limit.")
```
mbeckerle, line 841: Change to a plain 
`Assert.usage(optLengthLimitInBits.isDefined)` will make the coverage error go away.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -255,18 +292,27 @@ class BucketingInputSource(
...
       // Try to fill the rest of this bucket, regardless of how many bytes are
       // actually needed
       val emptyBytesInLastBucket = bucketSize - bytesFilledInLastBucket
 
+      Assert.invariant(emptyBytesInLastBucket >= 0)
       // Try to read enough bytes to fill the rest of this bucket. Note that
       // the .read() function could hit EOF (returns -1) or could return
-      // anywhere from zero to emptyBytesInLastBucket bytes
-      val bytesRead = inputStream.read(buckets(lastBucketIndex).bytes, bytesFilledInLastBucket, emptyBytesInLastBucket)
+      // anywhere from 0 to emptyBytesInLastBucket bytes
```
stevedlawrence, line 306: Should we update this comment to say 1 to emptyBytesInLastBucket byte? ``read()`` should never actually return 0.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -255,18 +292,27 @@ class BucketingInputSource(
...
-    var hasMoreData = true
+
     var needsMoreData = goalBucketIndex > lastBucketIndex || (goalBucketIndex == lastBucketIndex && bytesNeededInBucket > bytesFilledInLastBucket)
 
     while (needsMoreData && hasMoreData) {
       // Try to fill the rest of this bucket, regardless of how many bytes are
       // actually needed
       val emptyBytesInLastBucket = bucketSize - bytesFilledInLastBucket
 
+      Assert.invariant(emptyBytesInLastBucket >= 0)
```
stevedlawrence, line 303: Are there any errors if we change this assert to > instead of >=? I don't think we should ever be calling read with a len of zero bytes.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
```diff
@@ -255,18 +292,27 @@ class BucketingInputSource(
...
-      val bytesRead = inputStream.read(buckets(lastBucketIndex).bytes, bytesFilledInLastBucket, emptyBytesInLastBucket)
+      // anywhere from 0 to emptyBytesInLastBucket bytes
+      val bytesRead =
+        inputStream.read(
+          buckets(lastBucketIndex).bytes,
+          bytesFilledInLastBucket,
+          emptyBytesInLastBucket)
+
+      // check for bad inputStream behavior. It's not our fault!
+      if (bytesRead == 0 && emptyBytesInLastBucket > 0)
```
stevedlawrence, line 314: If we can change the above assert to > 0, this can just become ``bytesRead == 0``

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -117,6 +118,57 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
...
+
+  /**
+   * Determine if we're positioned at the end of data without
+   * doing any additional blocking operation such as reading more
+   * data to test if there is any.
+   *
+   * This depends on the underlying inputSource keeping track of
+   * whether it has previously hit the end of data or not.
+   *
+   * @return
```
stevedlawrence, line 140: Missing scaladoc for ``@return``

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -117,6 +118,57 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
...
+   */
+  final def isAtEnd(): Boolean = {
+    if (bitLimit0b.isDefined) {
+      bitPos0b == bitLimit0b.get
```
stevedlawrence, line 144: I'm not sure it's safe to use bitLimit here? I believe we change ``bitLimit0b`` when complex types have specified lengths to ensure children do not parse beyond those specified lengths. If we have reached the end of the complex type specified length, we shouldn't be considered isAtEnd, but I think this will do that?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -117,6 +118,57 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
...
+    } else {
+      // There is no bit limit, so whether we're at the end
+      // requires that we've hit then end-of-data on a read
+      // operation, AND that we've not backtracked since then,
+      // meaning we're positioned at the end of the last byte
+      // read.
+      hasReachedEndOfData && {
```
stevedlawrence, line 151: I think this is a now a non-backwards compatible change now. isAtEnd can only be true if we've gotten an EOD. And I imagine in a lot of binary formats we'll never try that extra read to get the EOD because we don't ever need to scan.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -117,6 +118,57 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
...
+        //
+        // No real input source can end in the middle of a byte.
+        // The only way that can happen is with the bitLimit having been
+        // set, and we've already dealt with that.
+        //
+        // So we just have to know that we're positioned at the
+        // end of the data.
+        //
+        val inputStreamBytePos0b = inputSource.position()
+        val isAtFinalByte = bytePos0b == inputStreamBytePos0b
```
stevedlawrence, line 161: I thought bytePos0b and inputSource.position() were basically always kept in sync? When we change bytePos0b, don't we also change the InputSource byte position, even when we backtrack. Seems like this is always going to be true, even if we've backtracked. Does the InputSource need a isAtEndOfData since only it really knows whre the last byte in the InputSource is?

daffodil-io/src/test/scala/org/apache/daffodil/layers/TestBase64.scala
```diff
@@ -66,7 +66,7 @@ W1vdXMgcXVvdGVzIG9yIHNvbmcgbHlyaWNzIG9yIGFueXRoaW5nIGxpa2UgdGhhdCBpbnRyb2R1Y
     pairs.foreach {
       case (exp, act) => {
         if (exp != act) {
-          println("differ at character %s (0-based). Expected '%s' got '%s'.".format(i, exp, act))
+          // println("differ at character %s (0-based). Expected '%s' got '%s'.".format(i, exp, act))
```
stevedlawrence, line 69: Should we just delete these printlns?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -490,7 +490,11 @@ class DataLocation private[japi] (dl: SDataLocation) {
   override def toString() = dl.toString
 
   /**
-   * Determine if this data location is at the end of the input data
+   * Determine if this data location is known to be at the end of the data stream.
+   *
+   * Note that this can be false, yet all bits consumed. This occurs if no
+   * subsequent input read operation from the data occurred causing the
+   * input stream to detect end-of-data/EOF.
```
stevedlawrence, line 497: This is an API change. I can imagine cases where people relied on this behavior for checking left over data, and I feel like it's very easy for binary formats to never have this set to true. Prevously, unless you were using a streaming thing, we always knew for sure if we hit EOD.

Maybe we need a tunable to switch between the "read one byte to figured out EOD" vs "must have hit EOD before, and we aren't actually sure if we are at EOD yet"? We can default to the old behavior, but people streaming or doing TCP stuff can switch to the new behavior, with the understanding that isAtEnd might not be accurate?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/io/InputSourceDataInputStream.scala
```diff
@@ -43,5 +43,13 @@ class InputSourceDataInputStream private[japi] (private [japi] val dis: SInputSo
...
    */
-  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr)) 
+  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr))
+
+  /**
+   * Used to see if there is more data or not after a parse operation.
+   * Whether more data is a left-over-data error, or just more data to
+   * parse in a subsequent parse call depends on the intention of the
+   * code calling the parse operation.
+   */
```
stevedlawrence, line 53: I think it's worth adding a comment saying that this will block until either at least either n bytes are read or EOD is hit.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -216,7 +216,7 @@ class JavaIOInputStream(s: InputSourceDataInputStream, finfo: FormatInfo)
     }
   }
 
-  override def available(): Int = 1
+  override def available(): Int = 0
```
stevedlawrence, line 219: Does nothing rely on this value? Surprised this didn't cause anything to fail.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
stevedlawrence: Agreed, but need to consider backwards compatibility.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/io/InputSourceDataInputStream.scala
```diff
@@ -43,5 +43,14 @@ class InputSourceDataInputStream private[sapi] (private [sapi] val dis: SInputSo
...
-  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr)) 
+  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr))
+
+
+  /**
+   * Used to see if there is more data or not after a parse operation.
+   * Whether more data is a left-over-data error, or just more data to
+   * parse in a subsequent parse call depends on the intention of the
+   * code calling the parse operation.
+   */
```
stevedlawrence, line 54: Same comment here as in the JAPI, might want to mention the blocking aspect of this.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -298,6 +298,13 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
 
   def doParseWithBothApis(dpInputStream: java.io.InputStream, saxInputStream: java.io.InputStream,
     lengthLimitInBits: Long): TDMLParseResult = {
+    //
+    // TDML Tests MUST have a length limit. Otherwise they cannot determine if
+    // there is left-over-data or not without doing more reading from the input stream
+    // so as to be sure to hit end-of-data.
+    //
+    Assert.usage(lengthLimitInBits >= 0)
```
stevedlawrence, line 306: Can't TDML tests use the isDefinedForLength(1) thing? Then we don't necessarily have to rely on lengthLimit?

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
```diff
@@ -598,7 +598,7 @@
       <tdml:documentPart type="text"><![CDATA[000118Ridgewood Circle    Rochester           NY123]]></tdml:documentPart>
     </tdml:document>
     <tdml:errors>
-      <tdml:error>6467715096</tdml:error>
+      <tdml:error>6467715464</tdml:error>
```
stevedlawrence, line 601: Any idea why this number changed?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
mbeckerle: Ah, yes it appears we have a stack discipline issue here. We narrow the available data by assigning a bit limit whenever we have a specified length. 

I just searched for uses of isAtEnd.

I'm quite convinced we should remove isAtEnd from the API. We only use it in tests. We know it is incorrect for it to be reading more data. You point out here that because it records the -1/End-of-Data in state of the ISDIS, we can't call it on an inner narrowed specified length scope or we'll be getting isAtEnd . 

Algorithmically, the right thing is for applications to call ISDIS.areBytesAvailable *before* a parse, and not expect upon return from the parse for it to be known whether we are at the end of the data stream or not. 
This code thunk we use to document streaming is wrong:
`
InputSourceDataInputStream is = new InputSourceDataInputStream(dataStream);
JDOMInfosetOutputter jdomOutputter = new JDOMInfosetOutputter();
boolean keepParsing = true;
while (keepParsing) {
  jdomOutputter.reset();
  ParseResult pr = dp.parse(is, jdomOutputter);
  ...
  keepParsing = !pr.location().isAtEnd() && !pr.isError();
}`

It should be:

`
InputSourceDataInputStream is = new InputSourceDataInputStream(dataStream);
JDOMInfosetOutputter jdomOutputter = new JDOMInfosetOutputter();
var keepParsing = true
while (keepParsing && is.isDefinedForLength(1)) { // there is at least 1 bit to parse!
  jdomOutputter.reset();
  ParseResult pr = dp.parse(is, jdomOutputter);
  ...
  keepParsing = !pr.isError();
}`
where we test for data availability at the top, before the call to parse, and there is no checking of isAtEnd at all.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
mbeckerle: It works as 1, it works as 0. I changed it to zero because I have no possible rationale why it would always be 1. 
The default behavior for available method, provided by java.io.InputStream is to always return 0.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
mbeckerle: Yes, I'm convinced the parse result has to expose the ISDIS for this purpose, rather than trying to snapshot it's position and whether it is "at end" at the end of the parse.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
mbeckerle: In the latest, I have deprecated isAtEnd entirely. The definition that remains is not compatible backward because it does not probe the data to see if we have more, or are at EOF.  

I'm wondering if we should put back in that blocking behavior, but now that this is deprecated, that would allow people who still have this code to work the same, while prompting them to rewrite it via the deprecation warning.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
mbeckerle: yes, done

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
mbeckerle: This is fixed now. Instead of isAtEnd depending on bitLimit, the TDML runner just does this itself. 
Right now, isAtEnd just tests if the bitPos == bitLimit when a bit limit is defined, false if no bit limit is defined.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/io/InputSourceDataInputStream.scala
mbeckerle: done

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
mbeckerle: Done

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
mbeckerle: Ok. I get your reasoning. 
Not sure the fix for this. (This code is now gone, but the question is still relevant.)
The scenario is we parsed forward, possibly ran into EOD, possibly not, but then we backtracked, and are no longer at the extreme end of where we reached with the parser. 
We need to recognize that case, and DataInputStream needs a way to support this, perhaps by keeping track of the bytePos of the last byte of the last filled bucket?

daffodil-io/src/test/scala/org/apache/daffodil/layers/TestBase64.scala
mbeckerle: Sure. Will do

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: I suggest we deprecate isAtEnd, but give it the bad-old functionality.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/io/InputSourceDataInputStream.scala
mbeckerle: I will update the comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
mbeckerle: Latest commit to this PR puts back isAtEnd (though still deprecated) with the original blocking behavior. 

Daffodil itself doesn't use this. It's purely to retain compatibility.

daffodil-lib/src/test/scala/org/apache/daffodil/util/TestCoroutines.scala
```diff
@@ -0,0 +1,172 @@
+/*
```
mbeckerle, line 1: This is just putting back a file of missing unit tests. Fixes DAFFODIL-2503

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -151,11 +151,11 @@
  * InputSourceDataInputStream is = new InputSourceDataInputStream(dataStream);
  * JDOMInfosetOutputter jdomOutputter = new JDOMInfosetOutputter();
  * boolean keepParsing = true;
- * while (keepParsing) {
+ * while (keepParsing && is.isDefinedForLength(1)) {
```
stevedlawrence, line 154: Thoughts on onsidering a differenent public API function that gives a little less power to the user? This functions let's the user prefetch anymount they want, e.g. `is.isDefinedForLength(500000)``. Which means we'll block until 500000 bits are cached. Any concerns with letting a user do that?

An alternative is maybe something like ``is.hasData()`` which could essentially be a public alias for ``is.isDefinedForLength(1)``, but limits what a user could do. We could always make isDefinedForLength public in the future if users did need it, but it's harder to take it away wihtout going through a deprecation process.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/io/InputSourceDataInputStream.scala
```diff
@@ -43,5 +43,16 @@ class InputSourceDataInputStream private[japi] (private [japi] val dis: SInputSo
...
-  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr)) 
+  def this(arr: Array[Byte]) = this(SInputSourceDataInputStream(arr))
+
+  /**
+   * Used to see if there is more data or not after a parse operation.
+   * Whether more data is a left-over-data error, or just more data to
+   * parse in a subsequent parse call depends on the intention of the
+   * code calling the parse operation.
+   *
+   * This operation will block until either n bytes are read or end-of-data
```
stevedlawrence, line 54: See other comment. Should be bits and maybe read vs available?

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -863,7 +849,6 @@ public void testJavaAPI17() throws IOException, InvalidUsageException, ClassNotF
         assertTrue(res.isError());
         assertFalse(res.isProcessingError());
         assertTrue(res.isValidationError());
-        assertTrue(res.location().isAtEnd());
```
stevedlawrence, line 866: Is it worth adding ``res.location().bitPos0b == lengthOfData`` comparison? We sometimes referene this file for examples of Java API usage, so showing how to test that we consumed all the data might be useful.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataLoc.scala
```diff
@@ -33,9 +33,20 @@ import org.apache.daffodil.util.MaybeULong
...
+class DataLoc(
+  val bitPos1b: Long,
+  bitLimit1b: MaybeULong,
+  eitherStream: Either[DataOutputStream, DataInputStream],
   val maybeERD: Maybe[ElementRuntimeData]) extends DataLocation {
 
+  @deprecated("Use bitPos1b to compare with expected position (possibly bitLimit1b).", "3.1.0")
+  override def isAtEnd =
+    eitherStream match {
+      case Right(isdis: InputSourceDataInputStream) => isdis.isAtEnd
```
stevedlawrence, line 45: I"m not sure this is worth worrying about, but this potentially has different behavior if you do something like this:
```scala
val res1 = dp.parse(isdis, ...)
val loc1 = res.location()
val res2 = dp.parse(isdis, ...)
val loc2 = res.location()
println(loc1.isAtEnd)
println(loc2.isAtEnd)
```
So if our InputSourceDataInputStream hit an EOD at the end of the second parse (i.e. our stream had two messages), then both ``loc1.isAtEnd`` and ``loc2.isAtEnd`` would be true. But only loc2.isAtEnd should be true. So future parse calls can change the isAtEnd state of previous ParseResults. Maybe this isn't a big deal since I doubt anyone will check isAtEnd of a previous parse location, but maybe it's worth considering? We at least probably want a backwards compatability note somewhere about this if we do keep this.

Could maybe do some logic where DataLoc also captures the bitPos of the InputSourceDataInputStream when it's created. If the data location has changed, then isAtEnd returns false, otherwise it returns isids.isAtEnd? Seems complicated for such a small edge case.

Another alternative would maybe be to have a tunable that complete disables use of isAtEnd in the DataLoc. Depending on the value of that tunable, DataLoc either calls isdis.isAtEnd and stores it (and blocking on TCP connections), or it just always returns false. Maybe call it something like "requireAccurateIsAtEnd" or something. People using TCP that aren't going to use isAtEnd can set this to false. And we default it to true for perfect backwards compatibility. Again, might be extra complication for a small edge case, so not sure it's worth it.

Just want to through out some ideas an see if it sparks any thoughts.

Agree that we definitely need to deprecate isAtEnd and its use though.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
mbeckerle: Good suggestion. I will add this.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -151,11 +151,11 @@
  * InputSourceDataInputStream is = new InputSourceDataInputStream(dataStream);
  * JDOMInfosetOutputter jdomOutputter = new JDOMInfosetOutputter();
  * boolean keepParsing = true;
- * while (keepParsing) {
+ * while (keepParsing && is..hasData()) {
```
stevedlawrence, line 154: Extra dot in ``is..hasData``

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -209,12 +209,12 @@
  * SAXHandler contentHandler = new SAXHandler();
  * xmlReader.setContentHandler(contentHandler);
  * Boolean keepParsing = true;
- * while (keepParsing) {
+ * while (keepParsing && is..hasData()) {
```
stevedlawrence, line 212: Same extra dot.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataLoc.scala
mbeckerle: Ultimately, a ParseResult should be an immutable object, excepting the InputSourceDataInputStream within it, which is obviously mutable. 
We can't answer isAtEnd without potentially mutating the ISDIS, so that can't be part of an immutable ParseResult. 
So really the problem is loc.isAtEnd shouldn't be on the loc object, because we want the loc object to be immutable. That's a big reason to deprecate isAtEnd. The bit pos and bit limit information on a DataLoc is immutable. 

In the latest commit, hasData() is on the DataInputStream, and there is no equivalent on DataLocation/DataLoc. 

I think adding a tunable here is overkill. Deprecating isAtEnd on loc, is easy to replace it with getting the data input stream and calling hasData().

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/package.scala
```diff
@@ -134,11 +134,11 @@ package org.apache.daffodil
  * val is = new InputSourceDataInputStream(dataStream)
  * val scalaOutputter = new ScalaXMLInfosetOutputter()
  * val keepParsing = true
- * while (keepParsing) {
+ * while (keepParsing && is..hasData()) {
```
stevedlawrence, line 137: Double dots

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataLoc.scala
stevedlawrence: All fair. Sounds good to me.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/package.scala
```diff
@@ -188,12 +188,12 @@ package org.apache.daffodil
  * val contentHandler = new SAXHandler()
  * xmlReader.setContentHandler(contentHandler)
  * val keepParsing = true
- * while (keepParsing) {
+ * while (keepParsing && is..hasData()) {
```
stevedlawrence, line 191: Double dots

stevedlawrence (approved):
+1 with double dots fixed

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
mbeckerle: Added these checks here and in SAPI.

daffodil-core/src/test/scala/org/apache/daffodil/api/TestParseIndividualMessages.scala
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Shows that we can parse exactly 1 message from a TCP network socket
+ * without blocking for bytes past the end of the messsage.
+ *
+ * This only works for DFDL schemas of formats that are specified length.
+ */
+class TestParseIndividualMessages {
+
+  import SocketPairTestRig._
```
mbeckerle, line 45: Highlight: This socket pair test rig is an important addition to test capability. It spins up actual network connections so that we can simulate reading from a socket that has no data, and block, later add data to cause it to unblock, etc.

daffodil-core/src/test/scala/org/apache/daffodil/api/TestParseIndividualMessages.scala
```diff
@@ -0,0 +1,230 @@
...
+
+        assertFalse(pr.isError)
+        assertEquals("1234", xml.text)
+      }
+    }
+    sptr.run()
+  }
+
+  /**
+   * Test shows that the isAtEnd method does not block looking for additional data.
```
mbeckerle, line 92: Comment now wrong. The underlying code isn't calling isAtEnd. This shows that reading 4 bytes does not block looking for additional data. 

This test is identical to the test above now, save that last assert. Combine these two tests.

daffodil-io/src/main/scala/org/apache/daffodil/io/StringDataInputStreamForUnparse.scala
```diff
@@ -66,6 +64,9 @@ final class StringDataInputStreamForUnparse
   override def getSomeString(nChars: Long,finfo: FormatInfo): Maybe[String] = dis.getSomeString(nChars, finfo)
   override def getString(nChars: Long,finfo: FormatInfo): Maybe[String] = dis.getString(nChars, finfo)
 
+  // $COVERAGE-OFF$ Nothing should be calling these.
```
mbeckerle, line 67: Highlight: This coverage-on/off stuff works. I got the coverage on this change set to pass codecov's criteria by putting these things around code that shouldn't have coverage.

daffodil-io/src/test/scala/org/apache/daffodil/io/FormatInfoForUnitTest.scala
```diff
@@ -89,3 +89,28 @@ class FormatInfoForUnitTest private ()
     }
   }
 }
+
```
mbeckerle, line 92: Just moved this. It was in the wrong place.

daffodil-io/src/test/scala/org/apache/daffodil/io/SocketPairTestRig.scala
```diff
@@ -0,0 +1,246 @@
...
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
+
+
+/**
+ * Test rig sets up TCP socket unidirectional between a producer
+ * output stream, and a consumer input stream.
+ */
+abstract class SocketPairTestRig {
```
mbeckerle, line 44: Highlight: This is the new test rig.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -490,10 +490,15 @@ class DataLocation private[japi] (dl: SDataLocation) {
   override def toString() = dl.toString
 
   /**
-   * Determine if this data location is at the end of the input data
+   * Determine if this data location is known to be at the end of the data stream.
```
mbeckerle, line 493: Revert comment. isAtEnd has the original blocking behavior. It's just deprecated now because that functionality is so problematic.

daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala
```diff
@@ -45,13 +45,17 @@ abstract class ThinException protected (dummy: Int, cause: Throwable, fmt: Strin
   def this(cause: Throwable) = this(1, cause, null)
```
mbeckerle, line 45: Highlight: These changes are just to improve codecov results. No functional change here. 
Partial fix for DAFFODIL-2509

daffodil-lib/src/main/scala/org/apache/daffodil/util/Coroutines.scala
```diff
@@ -73,8 +71,12 @@
     * Call when a co-routine resumes another (to provide a result of some sort)
     * and then terminates. The coroutine calling this must return from the run()
     * method immediately after calling this.
+    *
```
mbeckerle, line 74: Highlight: minor tweak to Coroutines to allow different types of data flowing on each path between two coroutines.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Coroutines.scala
```diff
@@ -98,6 +103,21 @@
    protected def run(): Unit
  }
 
+ /**
```
mbeckerle, line 106: Another minor tweak to coroutines. This pattern comes up everywhere coroutines are used.

daffodil-lib/src/test/scala/org/apache/daffodil/util/TestCoroutines.scala
```diff
@@ -0,0 +1,173 @@
...
```
mbeckerle, line 1: Highlight: Just adds back the missing unit tests for Coroutines library. Somehow these had gotten lost. This fixes DAFFODIL-2503

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala
```diff
@@ -188,11 +196,13 @@ final class ParseResult(outfile: os.Path,
...
-                          override val processorStatus: ProcessorResult,
-                          loc: DataLocation = Runtime2DataLocation())
+                          override val processorStatus: ProcessorResult)
   extends DFDL.UnparseResult
     with DFDL.State
     with WithDiagnosticsImpl {
+
+  val loc: DataLocation = Runtime2DataLocation(finalBitPos0b + 1, (finalBitPos0b + 1) / 8)
+
   /**
```
tuxji, line 206: Thanks for the heads up and fixing those locations.  You're setting them to the end of the input (why add 1, though?).

jadams-tresys (approved):
+1

Looks like a clean, well tested set of changes.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -194,7 +194,7 @@ class DFDLTestSuite private[tdml] (
   // That avoids creating the test suites repeatedly, but also leaks memory unless
   // you have an @AfterClass shutdown method in the object that calls runner.reset() at end.
   //
-  // @deprecated("2016-12-30", "Use Runner(...) instead.")
+  // @deprecated("Use Runner(...) instead.", "2016-12-30")
```
tuxji, line 197: Does the deprecation annotation work inside a comment?

tuxji (approved):
+1

Reviewed, looks fine, some questions.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: No. We want to deprecate this, but we backed off as there were too many changes required. Some day we can put it back in though. There's a JIRA ticket for updating all tests to use Runner.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala
mbeckerle: comment added. Because DataLocation is 1-based numbers.

PR #541, opened 2021-04-27
Fixes to release candidate container

stevedlawrence:
- Update release candidate container to Fedora 33 since it  packages
  provides packages for new runtime2 dependencies
- Install pinentry and set a UTF-8 LANG, needed to get gpg signing to
  work in F33
- Install clang and mxml-devel. Building a release candidate compiles
  runtime2 libraries, so a compiler and mxml lib is required. Change
  /usr/bin/cc to point to clang instead of gcc.
- Update SBT rpm repo location. Bintray has been deprecated, the SBT rpm
  is now self hosted. Also, no longer curl directly to a repo
  config--instead write the expected file contents. Although not likely,
  this mitigates a potential attack vector
- Specify a branch when init-ing the test git repo to hide a warning
  about changes to default git branch names

DAFFODIL-2506

### DIFF
--- a/containers/release-candidate/README.md
+++ b/containers/release-candidate/README.md
@@ -27,7 +27,7 @@ To build the Daffodil release candidate container image:
 
 To use the container image to build a release run the following:
 
-    podman run -it \
+    podman run -it --privileged \
       -v ~/.gitconfig:/root/.gitconfig \
       -v ~/.gnupg/:/root/.gnupg/ \
       -v ~/.ssh/:/root/.ssh/ \
--- a/containers/release-candidate/Dockerfile
+++ b/containers/release-candidate/Dockerfile
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.fedoraproject.org/fedora-minimal:30
+FROM registry.fedoraproject.org/fedora-minimal:34
 
 WORKDIR /root
 
--- a/containers/release-candidate/daffodil-release-candidate
+++ b/containers/release-candidate/daffodil-release-candidate
@@ -77,6 +77,9 @@ fi
 
 export GPG_TTY=$(tty)
 export WIX=/root/wix311/
+export LANG=en_US.UTF-8
+export CC=clang
+export AR=llvm-ar
 
 
 read -p "Pre Release label (e.g. rc1): " PRE_RELEASE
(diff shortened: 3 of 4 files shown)


mbeckerle (approved):
+1

tuxji (approved):
+1

I might've set the CC and AR environment variables instead (is binutils, where ar comes from, already installed?) but the symlink works for me.  BTW, clang doesn't install any llvm-ar binary (on Ubuntu, clang installs llvm-ar-10), so you'd have to install the Fedora package llvm if you want to set AR=llvm-ar too.  (Fedora has eu-ar from elfutils too, but elfutils is GNU-licensed like binutils.)

Also, Fedora 34 is officially out today, so you might want to use it as well.

stevedlawrence:
Yeah, CC/AR is probably a better method than mangling the symlinkgs. Will test and update.

Binutuils is already installed, or at least something is already pulling it in. Will also test and update llvm-ar.

I'll give Fedora 34 a shot as well.

containers/release-candidate/README.md
```diff
@@ -27,7 +27,7 @@ To build the Daffodil release candidate container image:
 
 To use the container image to build a release run the following:
 
-    podman run -it \
+    podman run -it --privileged \
```
stevedlawrence, line 30: Note the addition of the ``--privileged`` option when running the release candidate container. Building the MSI installer with wine failed when I upgraded to Fedora 34. After lots of debugging, I discovered that Fedora 34 wine would use a relatively new system call (``futex_time64``) when building the MSI. Because of its newness, the default container seccomp policy on Fedora 33 (and likely many other not brand new systems) does not include this system call in its whitelist, so this syscall is blocked in the container and wine crashes. Using ``--privileged`` disables seccomp filtering, and other container security mechanisms, and allows wine to build the MSI.

Instead of using ``--privileged``, we could create a custom seccomp policy, but that's likely a pain to maintain. And there are other security mechanisms, like capabilities, AppArmor, SELinux, etc., that might also get in the way at some point in the future that the ``--privileged`` option disables. So this option is most likely to allow things to work with future upgrades. Plus, the contents of this container are pretty trusted to not do bad things. And even if it was compromised, the container is still limited to what the user running the container can do, which hopefully is a rootless container, so any damange would be limited.

stevedlawrence (commented):
Updates based on review comments

tuxji:
+1 from me for the new changes.  Mike?

mbeckerle (approved):
+1 I don't think this is needed from me, but I'll double up!

PR #545, opened 2021-04-29
Fix some runtime2 bugs

tuxji:
Remove dependency on argp and use getopt instead.  Build with clang
instead of gcc.  Lower minimum required Mini-XML version from 3.2 to
3.0.  Add missing LICENSE and NOTICE files to daffodil-runtime2 jar.

main.yml: Build with clang and llvm-ar instead of cc and ar.

BUILD.md: Lower Mini-XML version to 3.0.  Remove argp and gcc
instructions.  Add clang instructions.

README.md: Lower Mini-XML version to 3.0.

LICENSE: Add to daffdil-runtime2 jar.

NOTICE: Add to daffdil-runtime2 jar.

Makefile: Improve comments explaining how to use targets.

daffodil_argp.[ch]: Rename to daffodil_getopt.[ch].

daffodil_getopt.c: Remove all argp structs and handlers.  Simplify to
just a single function calling getopt and returning a pointer to Error
if any error happens.  Note there is no portable way to parse
"daffodil [options] command [more options] arguments" with getopt.  We
will code the simplest way (hope that getopt will move all non-option
arguments to the end) and users will have to put all options before
all arguments if their getopt doesn't support that behavior ("daffodil
[options] [more options] command arguments").

daffodil_getopt.h: Include "errors.h" and make parse_daffodil_cli
return a pointer to Error so we can use continue_or_exit(error) for
all errors/messages.

daffodil_main.c: Remove fflush_continue_or_exit (no really good reason
to flush a stream right before closing it).  Call continue_or_exit
after parse_daffodil_cli to handle any CLi error.  Simplify rest of
code in main function.

errors.c: Move eof_or_error, get_diagnostics, add_diagnostics up so
they come first in file.  Add error_message switch cases to define all
CLI messages.  Add print_maybe_stop switch cases to print all CLI
messages, making sure to call printf & exit instead of error for CLI
help & version messages.

errors.h: Add CLI messages to ErrorCode and "int c" member to
ErrorCode anonymous union.  Rename ERR_INFOSET_READ/WRITE to
CLI_INVALID_INFOSET since it's a CLI message too.  Declare
daffodil_program_version (we were using argp_program_version before
which is no longer available) and move eof_or_error up before
get_diagnostics.

NestedUnion.c, ex_nums.c: Regenerate with code generator.

CodeGenerator.scala: Remove "-largp" since we no longer need it.

CodeGeneratorState.scala: Replace argp_program_version with
daffodil_program_version since we no longer use "argp.h".

DAFFODIL-2500 DAFFODIL-2505 DAFFODIL-2508


mbeckerle:
The rationale for lowering the dependency to 3.0 for mini-XML is what?
So that people won't have to dig around for the newer one if they already have the old one on their system?

That's kind of at cross purposes to all the scala-steward philosphy we use elsewhere now of moving forward to newer library versions as soon as they're available and stable.

BUILD.md
```diff
@@ -31,26 +31,32 @@ the latest [SBT] version following their websites' instructions or
...
 
-You can use the `dnf` package manager to install most of the build
-requirements needed to build Daffodil:
+You can use the `dnf` package manager to install most of the tools
+needed to build Daffodil:
 
-    sudo dnf install gcc git java-11-openjdk-devel make mxml-devel pkgconf
+    sudo dnf install clang git java-11-openjdk-devel llvm make mxml-devel pkgconf
+
+If you want to use clang instead of gcc, you'll have to set your
```
mbeckerle, line 53: You set it up so it is clang by default now? So is this instruction correct still, or does it have these settings now, and you have to flip them to use gcc now?

BUILD.md
```diff
@@ -61,10 +67,18 @@ commands you type will be able to call the C compiler.
...
 
-You can use the `apt` package manager to install most of the build
-requirements needed to build Daffodil:
+You can use the `apt` package manager to install most of the tools
+needed to build Daffodil:
+
+    sudo apt install build-essential clang-10 clang-format-10 default-jdk git libmxml-dev
 
-    sudo apt install build-essential default-jdk git libmxml-dev
+If you want to use clang instead of gcc, you'll have to set your
```
mbeckerle, line 75: ditto here. Ithought this was the default now.

BUILD.md
```diff
@@ -82,10 +96,18 @@ Install [MSYS2] following its website's instructions and open a new
...
 
-Install [gcc] and [libargp][argp] using MSYS2's `pacman` package
-manager:
+You can use the `pacman` package manager to install most of the tools
+needed to build Daffodil:
+
+    pacman -S clang diffutils git make pkgconf
+
+If you want to use clang instead of gcc, you'll have to set your
+environment variables `CC` and `AR` to the clang binaries' names:
```
mbeckerle, line 105: and here

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/daffodil_main.c
```diff
@@ -15,13 +15,13 @@
...
-#include <stdio.h>          // for NULL, perror, FILE, fclose, fflush, fopen, stdin, stdout
-#include <string.h>         // for strcmp
-#include "daffodil_argp.h"  // for daffodil_parse, daffodil_parse_cli, daffodil_unparse, daffodil_unparse_cli, daffodil_cli, parse_daffodil_cli, DAFFODIL_PARSE, DAFFODIL_UNPARSE
-#include "errors.h"         // for continue_or_exit, Error, print_diagnostics, PState, UState, ERR_FILE_CLOSE, ERR_FILE_FLUSH, ERR_FILE_OPEN, ERR_INFOSET_READ, ERR_INFOSET_WRITE
-#include "infoset.h"        // for walkInfoset, InfosetBase, rootElement, ERD, VisitEventHandler
-#include "xml_reader.h"     // for xmlReaderMethods, XMLReader
-#include "xml_writer.h"     // for xmlWriterMethods, XMLWriter
+#include <stdio.h>            // for NULL, FILE, perror, fclose, fopen, stdin, stdout
+#include <string.h>           // for strcmp
+#include "daffodil_getopt.h"  // for daffodil_cli, parse_daffodil_cli, daffodil_parse, daffodil_parse_cli, daffodil_unparse, daffodil_unparse_cli, DAFFODIL_PARSE, DAFFODIL_UNPARSE
```
mbeckerle, line 20: Line too long

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/errors.c
```diff
@@ -17,34 +17,109 @@
...
+            return true;
+        }
+    }
+    return false;
+}
+
+// error_message - get an internationalized error message
 
 static const char *
 error_message(enum ErrorCode code)
```
mbeckerle, line 76: Feels like this should be in a CLI-oriented file, since it's giving CLI-help.

mbeckerle (approved):
+1 Just a few questions. Looks good.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/errors.c
tuxji: Yeah, I originally had these CLI messages in daffodil_getopt.c and thought upon it for a while before deciding it was easier to bundle all the messages in one place for easier internationalization and flow of control (daffodil_parse_cli needs only return statements, not printf and exit calls too).  However, if we don't care about hardcoding 'daffodil' in the help, daffodil_parse_cli could pass the help message in err->s and both CLI_HELP_USAGE and CLI_PROGRAM_VERSION then would use the same format, "%s\n".  Would you prefer that?

tuxji (commented):
I still need to fix the Windows MSYS2 test failures.  I know CC=clang is causing these test failures but I'm not sure how or why yet.  Using the c/Makefile with CC=clang to compile the C files and run a test works fine on Windows MSYS2, so I need to debug the Scala code next.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/daffodil_main.c
stevedlawrence: What about just removing those comments? If the includes are autogenerated, is the information about why they are added useful? Seems it's only useful if you don't trust the autogenerated includes and want to verify why it added lines, which doesn't seem that necessary.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/errors.c
stevedlawrence: To me, feels like anything CLI related should not be in libruntime. The CLI is just a user of libruntime, and just like any other runtime user, they shouldn't need to update the runtime to change user specific error messages. Almost feels like the CLI needs a completely separate error path, or libruntime needs someway for users to inject addition error capabilities.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/examples/NestedUnion.c
```diff
@@ -25,7 +25,7 @@
 
 // Initialize our program's name and version
 
-const char *argp_program_version = "daffodil-runtime2 3.1.0-SNAPSHOT";
+const char *daffodil_program_version = "daffodil-runtime2 3.1.0-SNAPSHOT";
```
stevedlawrence, line 28: Is this hardcoded version going to cause tests to fail when we update the version?

BUILD.md
```diff
@@ -82,10 +96,18 @@ Install [MSYS2] following its website's instructions and open a new
 "MSYS2 MSYS" window.  We'll need its collection of free programs and
 libraries.
 
-Install [gcc] and [libargp][argp] using MSYS2's `pacman` package
-manager:
+You can use the `pacman` package manager to install most of the tools
+needed to build Daffodil:
+
+    pacman -S clang diffutils git make pkgconf
```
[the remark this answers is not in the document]
tuxji: I wanted to show developers how to install nearly everything they might need for runtime2 development on each platform.  The only time you need diffutils is to run the c/Makefile's "make test" rule, which you'd do only by hand.  Actually you need [xmldiff](https://xmldiff.readthedocs.io/en/stable/installation.html) for that Makefile rule too but I decided not to mention xmldiff in BUILD.md because there's no MSYS2 package and the procedure's too complicated for most developers (install Python3 and pip, use pip to install xmldiff).  I also thought it would be too complicated to show developers how to build and run [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use/blob/master/README.md), another tool that I use for updating the runtime2 C files' includes and comments showing what is used from each include.  Using iwyu requires some manual editing, so it's not easy to automate in CI either.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/daffodil_main.c
tuxji: No, I don't want to remove the comments.  The whole point of the comments is for me to show a developer looking at the code exactly what is being used from each include header.  If a developer is curious to learn more about a symbol, he or she can find exactly where to look by searching the comments to find the right include header and then open that .h file (also the .c file if both are in the repository).

The includes are not autogenerated.  The file has to compile successfully before iwyu can process it and iwyu doesn't edit the file.  It only prints some replacement includes / comments on standard output along with some other output and I have to manually edit the file to replace the includes / comments.  In fact, iwyu won't even print the replacement includes / comments if the only changes would be to the comments (it has no way of knowing what the original comments were), so I have to manually add an extra include in order for iwyu to show me the updated comments.  I wouldn't bother using iwyu unless it was to put these comments there as well as get rid of unneeded includes.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/errors.c
tuxji: Sounds like you would prefer that libcli have its own cli_errors.[ch] with similar enum, struct, and functions just to keep all the CLI_* codes and messages outside of libruntime.  Mike, do you want that too?

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/examples/NestedUnion.c
tuxji: No, this whole file is just a very lightly edited copy of an autogenerated file to be shown as an example of what generated code looks like.  We've had some discussion about failing the CI workflow if there's a difference (which would be a hassle) or auto-updating these example files in a sbt build (which would be better).  If we wanted to auto-update these files, we would have to delete the copyrights and change the filenames (unless you decide it's fine for the code generator to include the Apache copyright in its generated_code.[ch] files).

tuxji (commented):
Well, all the checks pass now (yay!).  Please let me know whether to make further changes after my responses to the last round of reviewer comments.  I've already decided to update main.yml (removing diffutils) and BUILD.md (exporting variables on one line).

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/daffodil_main.c
stevedlawrence: I don't have any problem with these comments, that said,

> If a developer is curious to learn more about a symbol, he or she can find exactly where to look by searching the comments to find the right include header and then open that .h file (also the .c file if both are in the repository).

Most IDE's and even text editors used for coding have an easy way to jump directly to the definition of a symbol. ctags+vim is my go-to, but there's lots of options for a more efficient alternative than searching through #include comments.

Your workflow to update these comments also seems pretty laborious. If it's not automated, I would expect that once other developers get involved that aren't used to this workflow it's going to be hard to keep these comments updated and useful. The removal of superfluous #includes does seem like a useful feature of iwyu though (I wasn't aware of this tool). Seems like there might be some benefit it automating some of these capabilities.

FYI, I've spoken to some devs recently that sound interested in this runtime2 and have similar FPGA goals, hoping to get them involved soon.

tuxji:
I've looked at what would happen if we split error codes between libruntime and libcli.  It turns out that most of the existing error codes had been used only in libcli too.  We would end up with this split between libcli and libruntime:

```c
// cli_errors.h - libcli

enum CliCode
{
    CLI_FILE_CLOSE,
    CLI_FILE_OPEN,
    CLI_HELP_USAGE,
    CLI_INVALID_COMMAND,
    CLI_INVALID_INFOSET,
    CLI_INVALID_OPTION,
    CLI_MISSING_COMMAND,
    CLI_MISSING_VALUE,
    CLI_PROGRAM_ERROR,
    CLI_PROGRAM_VERSION,
    CLI_STACK_EMPTY,
    CLI_STACK_OVERFLOW,
    CLI_STACK_UNDERFLOW,
    CLI_STRTOBOOL,
    CLI_STRTOD_ERRNO,
    CLI_STRTOI_ERRNO,
    CLI_STRTONUM_EMPTY,
    CLI_STRTONUM_NOT,
    CLI_STRTONUM_RANGE,
    CLI_UNEXPECTED_ARGUMENT,
    CLI_XML_DECL,
    CLI_XML_ELEMENT,
    CLI_XML_ERD,
    CLI_XML_GONE,
    CLI_XML_INPUT,
    CLI_XML_LEFT,
    CLI_XML_MISMATCH,
    CLI_XML_WRITE,
};

// errors.h - libruntime

enum ErrorCode
{
    ERR_CHOICE_KEY,
    ERR_FIXED_VALUE,
    ERR_PARSE_BOOL,
    ERR_STREAM_EOF,
    ERR_STREAM_ERROR,
};
```

@mbeckerle and @stevedlawrence , do you still want the codes and messages to be split up?  I would end up duplicating enum CliCode and enum Error, struct CliError and struct Error, and function cli_continue_or_exit and continue_or_exit along with a couple of static functions.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/errors.c
mbeckerle: That makes sense to me yes. We don't have to do this for this change set. We can do it subsequently if you prefer. But we definitely want libruntime to be usable from an application which has nothing whatsoever to do with CLIs nor XML,  and has its own error architecture used to create error messages from the error-codes coming back from the libruntime.

mbeckerle:
I like splitting the errors up that way yes.

tuxji:
OK, I've moved everything CLI-related in libruntime out to libcli now.  I made the error lookup mechanism pluggable (libcli has a hook to insert its own error lookup routine).  I also made changes to help automate clang-format, include-what-you-use, and generated_code.[ch] updates later.

PR #553, opened 2021-05-04
Fix a C portability problem

tuxji:
Some older C compilers (e.g., gcc 4.8.2 on CentOS 7) don't allow a
local variable to be declared inside a for loop condition.  It's much
easier to move the declaration outside the for loop condition than it
is to override the compiler with a different driver name or the
correct driver options.

DAFFODIL-2507

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -43,19 +43,47 @@ commands (or set them to `true` to disable C compilation altogether)
 if you don't want `sbt compile` to call your C compiler with `cc` and
 `ar` as the default commands.
 
-## Fedora 34
+## CentOS 7
 
-You can use the `dnf` package manager to install most of the tools
+You can use the `yum` package manager to install most of the tools
 needed to build Daffodil:
 
-    sudo dnf install clang git java-11-openjdk-devel llvm make mxml-devel pkgconf
+    sudo yum install clang gcc git java-11-openjdk-devel llvm make pkgconfig
 
 If you want to use clang instead of gcc, you'll have to set your
 environment variables `CC` and `AR` to the clang binaries' names:
 
     export CC=clang AR=llvm-ar
 
-However, Fedora has no sbt package in its default repositories.
+However, CentOS has no [mxml-devel][Mini-XML] or [sbt][SBT] packages
+in its own repositories.  You'll have to install the latest [SBT]
+version following its website's instructions and you'll have to build
+the [Mini-XML] library from source:
+
+    git clone -b v3.2 https://github.com/michaelrsweet/mxml.git
+    # ./configure fails if you use CC=clang
+    unset CC AR
+    cd mxml
+    ./configure --prefix=/usr --disable-shared --disable-threads
+    make
+    sudo make install
+
+Now you can build Daffodil from source and the sbt and daffodil
+commands you type will be able to call the C compiler.
+
+## Fedora 34
+
+You can use the `dnf` package manager to install most of the tools
+needed to build Daffodil:
+
+    sudo dnf install clang gcc git java-11-openjdk-devel llvm make mxml-devel pkgconf
+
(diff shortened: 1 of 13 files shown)


tuxji:
All checks have passed, ready for review now.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/infoset.c
```diff
@@ -132,7 +132,8 @@ walkInfosetNode(const VisitEventHandler *handler, const InfosetBase *infoNode)
     const ERD **const childrenERDs = infoNode->erd->childrenERDs;
     const size_t *    offsets = infoNode->erd->offsets;
 
-    for (size_t i = 0; i < count && !error; i++)
+    size_t i;
+    for (i = 0; i < count && !error; i++)
     {
         const size_t offset = offsets[i];
```
stevedlawrence, line 138: Is there a setting we can enable so that we can error when we add non-portable code? Even if we only set it for GitHub actions so we can catch it before it's merged if there are concerns about enabling it all time?

BUILD.md
```diff
@@ -43,6 +43,27 @@ commands (or set them to `true` to disable C compilation altogether)
 if you don't want `sbt compile` to call your C compiler with `cc` and
 `ar` as the default commands.
 
+## CentOS 7
+
+You can use the `yum` package manager to install most of the tools
+needed to build Daffodil:
+
+    sudo dnf install gcc git java-11-openjdk-devel llvm make pkgconfig
```
stevedlawrence, line 51: Should this be yum instead of dnf? Does CentOS 7 have dnf? Should this also be gcc instead of clang like we suggest for fedora, or is clang not even available?

stevedlawrence (approved):
+1

tuxji (commented):
To follow up on the dnf comment (outdated now so I had to comment separately here), yes, that was a mistake and I've corrected it in the BUILD.md.  The latest and greatest commit should work portably on both old and modern C compilers now and we should see warnings now if we use anything that C99 doesn't support.

tuxji:
Still need another +1.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/infoset.h
```diff
@@ -26,22 +26,26 @@
...
-typedef void (*ERDParseSelf)(InfosetBase *infoNode, PState *pstate);
-typedef void (*ERDUnparseSelf)(const InfosetBase *infoNode, UState *ustate);
-typedef const Error *(*InitChoiceRD)(const InfosetBase *infoNode, const InfosetBase *rootElement);
-
-typedef const Error *(*VisitStartDocument)(const VisitEventHandler *handler);
-typedef const Error *(*VisitEndDocument)(const VisitEventHandler *handler);
-typedef const Error *(*VisitStartComplex)(const VisitEventHandler *handler, const InfosetBase *base);
-typedef const Error *(*VisitEndComplex)(const VisitEventHandler *handler, const InfosetBase *base);
-typedef const Error *(*VisitNumberElem)(const VisitEventHandler *handler, const ERD *erd, const void *number);
+struct ERD;
```
mbeckerle, line 29: Surprising to me that this typedef isn't portable. Seriously you can't suppress the need to use the struct keyword everywhere you use these types? Ugh. 

Can we upgrade our usage to a newer still-standard revision of the C language that has this?

mbeckerle (approved):
+1
This is all fine. I just thought we were dealing with modern C compilers here. The inability to treat a struct as a typedef bugs the crap out of me. If that isn't really needed I'd prefer the typedefs.

tuxji (commented):
Thoughts?

PR #554, opened 2021-05-05
WIP: Add tests to reproduce DAFFODIL-2183, but they pass

tuxji:
The unparse test was supposed to fail with the error `Unparse Error: Element {}person does not have a value`, but it passes.

Can we get a new test case that proves the bug still exists?

DAFFODIL-2183

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section13/nillable/literal-value-nils-unparse.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section13/nillable/literal-value-nils-unparse.tdml
@@ -740,6 +740,24 @@
         </xs:choice>
       </xs:complexType>
     </xs:element>
+    
+    <xs:element name="doc4">
+      <xs:complexType>
+        <xs:sequence dfdl:separator="|" dfdl:separatorPosition="infix">
+          <xs:element name="person" maxOccurs="unbounded"
+                      dfdl:occursCountKind="implicit"
+                      dfdl:initiator="Person:" nillable="true"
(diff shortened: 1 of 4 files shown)


stevedlawrence:
I think the bug is saying the expected result is a successful unparse, like what you have, but was instead outputting some error.

Seems to be like this issue was fixed and we just didn't realize it? This bug was opened almost 2 years ago, so that wouldn't surprised me. The tests look like they are correct to me, which would confirm this works as expected now.

stevedlawrence (approved):
+1 with newline issue fix. Suggest changing the tests to use a pipe separator or something instead of %NL; to avoid windows vs unix newlines issues

tuxji:
I've changed the tests to use pipe instead of newline as separator, but are the tests worth merging?  They don't seem to differ very much from the other complex_nil tests.

If we still add the tests, I also wonder if we really need to add both a parser test case to literal-value-nils.tdml and an unparser test case to literal-value-nils-unparse.tdml.  Round-tripping will test both the parser and the unparser so I believe it's redundant to add both a parse test case and an unparse test case, right?

mbeckerle:
Tests get split like this when there is a bug in just one newline. (This case) or there is a round trip complexity, which can be due to using newlines (this case too). Folding the tests back together by eliminating these issues would def be cleaner.

tuxji:
May I have another +1 on the new test?

stevedlawrence:
+1 looks good!

stevedlawrence:
Sorry, didn't realize I already gave a +1, still need someone else

mbeckerle (approved):
+1

PR #557, opened 2021-05-11
Update CLI help information

tuxji:
Compare in-program CLI help information with Daffodil website's CLI
documentation.  Make changes to make them match each other better,
such as capitalizing first word of help information, reordering help
information in same order as displayed by program and website,
reordering subcommands in same order as web page, and changing some
words for more consistency.  Also simplify CLI help in response to
reviewer suggestions:

- Remove confusing [{<namespace>}] from -D and -r options
- Sort usage options essential first, then alphabetically
- Hide -p, --path option since Daffodil allows only "/"
- Clarify validation mode is not optional anymore
- Rename names/testname to testnames for consistency
- Add "Defaults to" in more places

Note: IDEA shows many warnings for Main.scala.  It would be a good
idea to clean it up using IDEA's suggested fixes.  Some of these
warnings will need to be fixed when it's time to support Scala 3.

DAFFODIL-817, DAFFODIL-2477

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -174,18 +174,18 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
    * This is used when the flag is optional and so is its
    * argument.
    *
-   * Let's use --validate [mode] as an example.
+   * Let's use --debug [file] as an example.
    *
    * This optionalValueConverter first determines if the
-   * --validate flag was given.  If it wasn't, Right(None)
+   * --debug flag was given.  If it wasn't, Right(None)
    * is returned.
    *
    * If the flag was given we then need to determine
-   * if 'mode' was also given.  If mode was not given, Right(Some(None))
-   * is returned meaning that the --validate flag was there but 'mode'
-   * was not.  If the mode was given, Right(Some(Some(conv(mode)))) is
-   * returned.  This means that the --validate flag was there as well as
-   * a value for 'mode'.
+   * if 'file' was also given.  If file was not given, Right(Some(None))
+   * is returned meaning that the --debug flag was there but 'file'
+   * was not.  If the file was given, Right(Some(Some(conv(file)))) is
+   * returned.  This means that the --debug flag was there as well as
+   * a filename for 'file'.
    *
    * conv(mode) simply performs the necessary conversion
    * of the string to the type [A].
@@ -284,7 +284,8 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
 
   printedName = "Apache Daffodil"
 
-  helpWidth(76)
+  val width = 87
+  helpWidth(width)
 
   def error(msg: String) = errorMessageHandler(msg)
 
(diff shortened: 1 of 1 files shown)


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -319,40 +319,39 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
 
   // Global Options
-  val debug = opt[Option[String]](argName = "file", descr = "enable debugging. Optionally, read initial debugger commands from [file] if provided.")(optionalValueConverter[String](a => a))
-  val trace = opt[Boolean](descr = "run the debugger with verbose trace output")
-  val verbose = tally(descr = "increment verbosity level, one level for each -v")
-  val version = opt[Boolean](descr = "Show version of this program")
+  val debug = opt[Option[String]](argName = "file", descr = "Enable the interactive debugger. Optionally, read initial debugger commands from [file] if provided.")(optionalValueConverter[String](a => a))
+  val trace = opt[Boolean](descr = "Run this program with verbose trace output")
+  val verbose = tally(descr = "Increment verbosity level, one level for each -v")
+  val version = opt[Boolean](descr = "Show Daffodil's version")
```
stevedlawrence, line 325: I don't feel strongly about this, but looking at a handful of common linux tools, it seems the standard is to start option descriptions with a lower case character.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -319,40 +319,39 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
 
   // Parse Subcommand Options
   object parse extends scallop.Subcommand("parse") {
-    banner("""|Usage: daffodil parse (-s <schema> [-r [{namespace}]<root>] [-p <path>] |
+    banner("""|Usage: daffodil parse (-s <schema> [-r [{<namespace>}]<root>] [-p <path>] |
```
stevedlawrence, line 329: The square bracket, curly brace, and angle bracket is kind of hard to visiaully parse. Also, I think the standard is to use square brackets for optional things and angle brackets for required things, so I'm not sure the namespace is right. But the curly braces are also a bit confusing.

Maybe to simplify this and make it easier to visually parse, I wonder if we should just make this `-r <root>`. And then the help for the root option can describe the syntax (e.g. root element name, optionally preceeded by namespace surrounded in curly braces). Rarely do you actually need the namespace so it's probably not worth complicating the overview of options for this one rarely used feature.

If we do make this change, we should also do it any other time we allow a namespace to be specified. I think maybe the only other place is with the `-D` for variable names?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -319,40 +319,39 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
-    val infile = trailArg[String](required = false, descr = "input file to parse. If not specified, or a value of -, reads from stdin.")
+    val config = opt[File](short = 'c', argName = "file", descr = "Path to XML file containing configuration items")
+    val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing. An optional namespace may be provided.")
+    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'.", default = Some(InfosetType.XML))
+    val output = opt[String](argName = "file", descr = "Output file to write infoset to. If not given or is -, infoset is written to stdout.")
+    val parser = opt[File](short = 'P', argName = "file", descr = "Use a previously saved parser")
+    val path = opt[String](argName = "path", descr = "Path from root element to node from which to start parsing")
```
stevedlawrence, line 347: I wonder if we should hide this path option by adding `hidden = true`? Daffodil only supports `/`, so cluttering the help output with this option that isn't really an option probably doesn't add anything and just makes the help output more verbose. And by hiding it, people can still use it so it won't break backwards compatibility.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -319,40 +319,39 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
-    val tunables = props[String]('T', keyName = "tunable", valueName = "value", descr = "daffodil tunable to be used when parsing.")
-    val config = opt[File](short = 'c', argName = "file", descr = "path to file containing configuration items.")
-    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "infoset type to output. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'.", default = Some(InfosetType.XML))
-    val stream = toggle(noshort = true, default = Some(false), descrYes = "when left over data exists, parse again with remaining data, separating infosets by a NUL character", descrNo = "stop after the first parse, even if left over data exists")
-    val infile = trailArg[String](required = false, descr = "input file to parse. If not specified, or a value of -, reads from stdin.")
+    val config = opt[File](short = 'c', argName = "file", descr = "Path to XML file containing configuration items")
+    val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing. An optional namespace may be provided.")
```
stevedlawrence, line 343: Personally, I'd rather sort these vals by the val name rather than the short value. At first it looked like this was out of order.

Also, maybe we should change this to `short = 'D'` to be consistent with  other options with custom short letters?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -370,78 +369,34 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
-    val stream = toggle(noshort = true, default = Some(false), descrYes = "split the input data on the NUL character, and unparse each chuck separately", descrNo = "treat the entire input data as one infoset")
-    val infile = trailArg[String](required = false, descr = "input file to unparse. If not specified, or a value of -, reads from stdin.")
+    val config = opt[File](short = 'c', argName = "file", descr = "Path to XML file containing configuration items")
+    val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "Variables to be used when unparsing. An optional namespace may be provided.")
+    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to unparse. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', or 'sax'.", default = Some(InfosetType.XML))
+    val output = opt[String](argName = "file", descr = "Output file to write unparsed infoset to. If not given or is -, unparsed infoset is written to stdout.")
```
stevedlawrence, line 390: At first I read this it's writing the infoset, and thought something was wrong. I guess the "unparsed infoset" implies that it is writing the the data that is created from unparsing the infoset, and so is technically right, but it stil feels a bit confusing to me. Especially since there is a concept of an "augmented infoset" that is created during unparsing. It's possible someone might interpret this as writing that infoset. I do understand how saying "unparsed data" could also be confusing though. Maybe we just say "Output file to write data to", with the imlication that the data is the data from unparsing the infoset? Maybe I'm overthinking this...

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -458,53 +413,94 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
-    val list = opt[Boolean](descr = "show names and descriptions instead of running test cases.")
-    val regex = opt[Boolean](descr = "treat <names> as regular expressions.")
-    val tdmlfile = trailArg[String](required = true, descr = "test data markup language (TDML) file.")
-    val names = trailArg[List[String]](required = false, descr = "name of test case(s) in tdml file. If not given, all tests in tdmlfile are run.")
-    val info = tally(descr = "increment test result information output level, one level for each -i")
+    val info = tally(descr = "Increment test result information output level, one level for each -i")
+    val list = opt[Boolean](descr = "Show names and descriptions instead of running test cases")
+    val regex = opt[Boolean](descr = "Treat <testname...> as regular expressions")
+    val tdmlfile = trailArg[String](required = true, descr = "Test Data Markup Language (TDML) file")
+    val names = trailArg[List[String]](required = false, descr = "Name(s) of test cases in tdmlfile. If not given, all tests in tdmlfile are run.")
```
stevedlawrence, line 457: The banner testname but the help is going to say "names". We should make that consistent while we are changing things.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -458,53 +413,94 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
+              |
+              |Performance Options:""".stripMargin)
+
+    descr("Run performance test")
+    helpWidth(76)
+
+    val config = opt[File](short = 'c', argName = "file", descr = "Path to XML file containing configuration items")
+    val vars = props[String]('D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing or unparsing. An optional namespace may be provided.")
+    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output or unparse. Must be one of 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'.", default = Some(InfosetType.XML))
+    val number = opt[Int](short = 'N', argName = "number", default = Some(1), descr = "The total number of files to process. Defaults to 1.")
```
stevedlawrence, line 479: I just noticed that sometimes we include a "Defaults to .." message, but sometimes we don't. For example, infoseType defaults to "xml".  Should we be consistent, and any option that has a default value, we should mention what it defaults to in the help string?

stevedlawrence (approved):
+1, just some minor thoughts

Also,  that so little space available in the help output since we have some relatively long option names and we wrap to 76 chars, makes me wonder:

1. Maybe we shouldn't wrap to 76. Maybe we should consider bumping that up a little bit? Probably doesnt have to be much to make things a little more readable. 
2. Considering the limited space, I dont' think it would be unreasonable to use slight less formal language sometimes if takes up less character but gets the point across just as well. Not sure of any specifics, but maybe something to consider.

stevedlawrence:
Also, can you create a bug for this change so we can track it and make sure the changes are verfied?

There might also be an existing bug? I'm pretty sure we have some bugs about updating CLI help output, but I might just be thinking about removing all the redundancy.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -319,40 +319,39 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
...
   object parse extends scallop.Subcommand("parse") {
-    banner("""|Usage: daffodil parse (-s <schema> [-r [{namespace}]<root>] [-p <path>] |
+    banner("""|Usage: daffodil parse (-s <schema> [-r [{<namespace>}]<root>] [-p <path>] |
               |                       -P <parser>)
-              |                      [--validate [mode]]
-              |                      [-D[{namespace}]<variable>=<value>...] [-o <output>]
-              |                      [-I <infoset_type>]
-              |                      [--stream]
-              |                      [-c <file>] [infile]
+              |                      [-D[{<namespace>}]<variable>=<value>...] [-o <output>]
```
mbeckerle, line 331: what does the curly brace mean?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: I second this change.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: Unparsed data makes more sense than Unparsed infoset.

mbeckerle (approved):
+1

PR #559, opened 2021-05-12
Prepare for 3.1.0 release

stevedlawrence:
DAFFODIL-2514


mbeckerle (approved):
+1

tuxji (approved):
+1

FYI, nothing (not even tests) uses those generated_code.c files.  They're only examples of the C code generator's output at a certain point in time.  I'm fine with updating them manually now but I'm sure we'll want to automate updating them in the future so that you'll be able to change only build.sbt next time.

PR #560, opened 2021-05-12
Verify that we get errors if the files contain a DOCTYPE

mbeckerle:
If the checks all pass, then this is for review. Note that I do not plan to merge this until after the 3.1.0 release.

Tests that show this error detection for each of
* DFDL schemas
* included DFDL schemas
* imported DFDL schemas
* TDML files
* XML Infoset files used in TDML tests
* Config files used in TDML tests
* External variable binding files

Eliminating use of DOCTYPE also eliminates any possibility of
XML General or Paremter Entities.

Consolidated loader/validators of XML and XSD.

Everything that loads XML, XSD, TDML, config, external vars,
or ".dfdl.xsd", now uses DaffodilXMLLoader to do it.

Specific validators that were for Config files or
External Variable Binding files are now gone.

DaffodilXMLLoader was simplified and made more uniform.
Single purpose mixin traits and adapters were eliminated.

The validation in DaffodilXMLLoader uses two methods.

First it uses the XercesValidator (used by the validation feature).

Second it does a validating load with Xerces. This seems to
catch/report different validation problems. (Instrumentation to
prove this may be worth it, so that we can get rid of redundant work
if possible)

DFDL schemas are validated by constructing an XML Schema
object from them, as well as by loading them and validating
against the XML Schema for DFDL schemas.

The DaffodilXMLLoader validates (if requested) using a supplied XML
Schema. But then always loads using the DaffodilConstructingParser
which uses the underlying ConstructingParser - this is needed because
in many cases we are dependent on properly handling CDATA regions
(e.g., TDML files, test infoset XML files, etc) which Xerces doesn't
do properly.

I attempted to implement DAFFODIL-288 to validate the infoset XML
(before unparsing) also but was unsuccessful so far, but a TODO
DAFFODIL-288 marks the place where that fix goes.

New validation is more uniform, and thorough. This caught a number of
small issues like missing "tdml:" prefixes on numerous files'
testSuite elements.
There were various adjustments to accomodate the
more strict validation.

Fixes to SAX due to simple types now being series of Text, Atom, and
EntityRef.

Various other small fixes to TDML runner to insure no diagnostic
errors are being hidden.

DAFFODIL-1422, DAFFODIL-1659


mbeckerle:
All windows tests are failing currently. A CRLF issue - loaders are likely now standardizing CR to LF where they were perhaps not before. It's somehow related to the use of the constructing parser instead of xerces when loading.

mbeckerle:
Nah. I have quite a bit to look at here.

There is a race condition due to multithreaded test execution where tests pass if run individually, but fail if run collectively.

So the loader changes I've made are somewhere sharing state.

I have to chase that down.

I am changing this back to a WIP.

mbeckerle:
Changed back to a WIP - not yet ready for review. 

I have discovered there is state sharing causing a multi-thread error. Tests pass when run one-by one in an IDE, or individually under sbt, but when run as a batch under sbt they fail.

mbeckerle:
If the checks all pass, then this is ready for review.

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUtils.scala
```diff
@@ -54,10 +55,12 @@ object TestSAXUtils {
...
 
+  lazy val loader = new DaffodilXMLLoader()
 
-  lazy val qualifiedWithNestedSchemasFile: File = new File(
-    Misc.getRequiredResource("/test/example_nested_namespaces_qualified.dfdl.xsd"))
-  lazy val qualifiedWithNestedSchemasElem: Elem = scala.xml.XML.loadFile(qualifiedWithNestedSchemasFile)
+  lazy val qualifiedWithNestedSchemasFile =
+    Misc.getRequiredResource("/test/example_nested_namespaces_qualified.dfdl.xsd")
+  lazy val qualifiedWithNestedSchemasElem: Elem =
+    loader.load(URISchemaSource(qualifiedWithNestedSchemasFile), None).asInstanceOf[Elem]
```
mbeckerle, line 63: @olabusayoT  Please note this is what I changed, that eliminated the bad behavior. 

It remains a mystery to me why calling scala.xml.XML.loadFile(...) was resulting in bad behavior, but only when sbt was run from the command line. 
I even commented out all but one test in the TestSAXParseAPI.scala file, but still running sbt "testOnly *TestSAXParseAPI" would run just that one test, and fail with the b02:a namespace prefix problem. 

Changing this to use DaffodilXMLLoader.load fixed the issue. I am not sure why.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -538,47 +401,226 @@ trait SchemaAwareLoaderMixin {
  * do any validation either), however, once Daffodil starts processing the
  * DFDL schema nodes, it resolves references using the same one true XML catalog resolver.
  */
-class DaffodilXMLLoader(val errorHandler: org.xml.sax.ErrorHandler) {
+class DaffodilXMLLoader(val errorHandler: org.xml.sax.ErrorHandler)
+  extends NoBindingFactoryAdapter {
```
mbeckerle, line 405: If we can avoid mixing this in, we should do so, because there are many non-usable methods on this class.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -538,47 +401,226 @@ trait SchemaAwareLoaderMixin {
...
     val res = constructingLoader.load() // construct the XML objects for us.
     constructingLoader.input.close()
     res
   }
 
-  def validateSchema(source: DaffodilSchemaSource) = xercesAdapter.validateSchema(source)
+  // We disallow any of these to force us of our own
+  // load(source, optSchemaURI) entry point.
+  // $COVERAGE-OFF$
+  private def noWay = Assert.usageError("Operation is not supported. Use load(uri) or loadFile(file)")
```
mbeckerle, line 614: These are all inherited methods. Not clear we need the base/trait they come from.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -538,47 +401,226 @@ trait SchemaAwareLoaderMixin {
...
-      if (xercesNode == null) return null
-      // Note: we don't call xercesAdapter.validateSchema(source)
-      // here, because this is an XML loader, not necessarily
-      // just a DFDL schema loader. So for example the doValidation flag
-      // above could be telling us to validate a TDML file or not.
+  def load(source: DaffodilSchemaSource,
+    optSchemaURI: Option[URI],
+    addPositionAttributes: Boolean = false,
+    normalizeCRLFtoLF: Boolean = true): scala.xml.Node = {
+    //
```
mbeckerle, line 574: Note @jw3 we're using the XercesValidator that is built in, here so that we have only one instance of Xerces validation now.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilXMLLoader.scala
```diff
@@ -538,47 +401,226 @@ trait SchemaAwareLoaderMixin {
...
+      // because that will actually interpret things like xsi:schemaLocation attributes
+      // of the root element.
+      //
+      // scopeStack.push(TopScope) // not in scala xml v2.0.0
+      val parser = parserFromURI(optSchemaURI)
+      val xrdr = parser.getXMLReader()
+      val saxSource = scala.xml.Source.fromSysId(source.uriForLoading.toString)
+      // parser.parse(source.uriForLoading.toURL.openStream(), this)
+      xrdr.parse(saxSource)
+      // scopeStack.pop() // not in scala xml v2.0.0
```
mbeckerle, line 596: It would be good to document exactly what additional errors this second validation catches that the first validation does not. Without this numerous negative tests failed. 

At least do this: comment out this check, and list what the tests are that fail.

mbeckerle: Tried this. Tests that fail are:
TestTDMLRunner.test_tdmlNamespaces1 - does not get a TDMLException it is expecting which is specified in the TDML file using xsi:schemaLocation. 

This is not being checked at all without this added validation step. This is why the 2nd check is needed.
Comments in the code have been updated to indicate this. 

These tests are expecting a "DocType is disallowed" exception, and without that 2nd check, they get a NPE. 
I've fixed that so that the TDML runner doesn't throw a NPE any more. 

TestDisallowDocType.test_TDMLFileMustNotHaveDocType
TestDisallowDocType.test_dfdlSchemaMustNotHaveDocType
TestDisallowDocType.test_configMustNotHaveDocType

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
```diff
@@ -129,6 +129,19 @@
                 <!-- We want the default to be 'unqualified' in the longer term but 
                      we leave it 'qualified here to prevent lots of tests from breaking. -->
     <attribute name="elementFormDefault" type="tns:elementFormDefaultType" default="qualified"/>
+        <!-- we want the default for useDefaultNamespace to be false longer term but
+             we leave it as true here to prevent lots of tests from breaking. -->
+    <attribute name="useDefaultNamepace" type="xs:boolean" default="true">
```
mbeckerle, line 134: Typo. useDefaultNamespace.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -126,16 +125,21 @@ final class TDMLDFDLProcessorFactory private (
       val diags = p.getDiagnostics
       Left(diags)
     } else {
-      val dp =
+      val dp = {
+        // Always serialize the parser/unparser objects.
+        // This insures nothing is left uncomputed by lazy evaluation.
```
stevedlawrence, line 130: Mentioned this earlier, but this sounds like a bug to me. We shouldn't need to serialize the parser to ensure everything is  evaluated.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -159,14 +163,18 @@ final class TDMLDFDLProcessorFactory private (
   override def getProcessor(
     schemaSource: DaffodilSchemaSource,
     useSerializedProcessor: Boolean,
-    optRootName: Option[String] = None,
-    optRootNamespace: Option[String] = None): TDML.CompileResult = {
+    optRootName: Option[String],
+    optRootNamespace: Option[String],
+    tunables: Map[String, String]): TDML.CompileResult = {
```
stevedlawrence, line 168: Why pass in tunbles now? Doesn't the TDML runner call setTunables?

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunnerConfig.scala
```diff
@@ -203,7 +203,7 @@ class TestTDMLRunnerConfig {
       ts.runOneTest("test1")
     }
     val msg = e.getMessage()
-    println(msg)
+    // println(msg)
```
stevedlawrence, line 206: Remove comment, same with below tests

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLUnparseCases.scala
```diff
@@ -49,17 +47,18 @@ class TestTDMLUnparseCases {
...
+//    tc.document
+//    val is = tc.inputInfoset
+//    // println(is)
+//    val bar = is.dfdlInfoset.contents
+//    val scala.xml.Elem(pre, label, _, _, _) = bar
+//    assertEquals("ex", pre)
+//    assertEquals("bar", label)
     // ts.trace
     ts.runOneTest("test1")
+    println(ts.loadingDiagnosticMessages)
```
stevedlawrence, line 61: Remove println, maybe add an assert tht checks diagnostics?

daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/charClassEntities.tdml
```diff
@@ -62,11 +62,12 @@
...
-    roundTrip="twoPass">
+    model="charClassEntities.dfdl.xsd" description="CSV-style tests and use of %NL; - DFDL-6-045R">
 
     <document>
-      <documentPart type="text"><![CDATA[0,1,2,3,4
5,6,7,8,9
10,11,12,13,14]]></documentPart>
+      <documentPart type="text"><![CDATA[0,1,2,3,4
+5,6,7,8,9
+10,11,12,13,14]]></documentPart>
```
stevedlawrence, line 70: git autocrlf might change this file.  Is this testing CR in TDML files? I don't think that really going to work without doing some special stuff to ensure git doesn't muck with things, or that future formatters don't change to a standard like LF.

I would say if we need to test loading files with CR/CRLF, we should create strings of data and load that. We really can't easily rely on files having the expected line endings.

project/Dependencies.scala
```diff
@@ -48,7 +48,8 @@ object Dependencies {
   lazy val test = Seq(
     "junit" % "junit" % "4.13.2" % "it,test",
     "com.novocode" % "junit-interface" % "0.11" % "it,test",
-    "org.scalacheck" %% "scalacheck" % "1.15.4" % "it,test"
+    "org.scalacheck" %% "scalacheck" % "1.15.4" % "it,test",
+    "xmlunit" % "xmlunit" % "1.6" % "test"
```
stevedlawrence, line 52: Did you verify the license of XML unit and any dependecies? I also don't see it used any where? Can this be removed?

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUtils.scala
```diff
@@ -217,8 +224,10 @@ object TestSAXUtils {
       val msgs = pf.getDiagnostics.map { _.getMessage() }.mkString("\n")
       fail("pf compile errors: " + msgs)
     }
-    pf.sset.root.erd.preSerialization // force evaluation of all compile-time constructs
+    // pf.sset.root.erd.preSerialization // force evaluation of all compile-time constructs
     val dp = pf.onPath("/").asInstanceOf[DataProcessor]
+    val baos = new ByteArrayOutputStream()
+    dp.save(Channels.newChannel(baos)) // insures everything serialized.
```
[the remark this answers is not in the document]
mbeckerle: Created https://issues.apache.org/jira/browse/DAFFODIL-2521

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
mbeckerle: Some of the tunables affect compilation, hence, unless they are passed in, the cache can return a false match. The unqualifiedPathStepPolicy is the primary one where I was actually getting a test failure (can't recall which test), but there can (perhaps are) others, so I decided to just pass them all in. 
We could make a subset data structure with just the compler-affecting tunables in it, but there are a lot of them, many are used both at compile time and runtime. 

These are the tunables accessed by schema compilation:
unqualifiedPathStepPolicy
requireTextStandardBaseProperty
requireEmptyElementParsePolicyProperty
allowExpressionResultCoercion
requireChoiceDispatchKeyKindProperty
requireFloatingProperty
generatedNamespacePrefixStem
requireTextBidiProperty
requireEncodingErrorPolicyProperty
maxBinaryDecimalVirtualPoint
minBinaryDecimalVirtualPoint
parseUnparsePolicy
maxSkipLengthInBytes
initialRegexMatchLimitInCharacters
defaultEmptyElementParsePolicy
... and any used by expressions which are evaluated at compile time for constant folding purposes.
e.g., maxValidYear, minValidYear, 

Any of these changing *could* result in an SDE vs. successful compilation.

stevedlawrence: That seems too difficult and hard to maintain. I guess it's not a big deal.

A thought though, I wonder if maybe the caching shouldn't happen in the DaffodilTDMLDFDLProcessor? And it should be up to the TDMLRunner to cache things? That way we can cache any processor, regardless of the implementation and implmentations don't have to worry about it?

Though, might not be worth the effort since I think there are some thoughts about removing caching entirely?

mbeckerle: The caching needs to get hoisted out of DaffodilTDMLDFDLProcessor into TDMLRunner in order for it to not be a global cache, but a per TestSuite object cache. We do need some caching to avoid recompilation for every test in a test suite.

stevedlawrence: This is beyond the scope of this change, but will we even need the cache at all? Could our test suites become something like this:
```scala
object TestFoo {
  var runner: Runner = _
  @BeforeClass def setup: Unit = { runner = Runner("foo.tdml") }
  @AfterClass def tearDown: Unit = { runner = null }
}
```
This way we compile the TMDL files when JUnit is about to start the tests, keep it around for the life of the suite, and then set it to null once all tests are complete and let the garbage collector cleanup the runner?

mbeckerle: That will work if the cache is part of the testsuite/runner object, not a global singleton.

project/Dependencies.scala
mbeckerle: Will remove or verify license. I like that there is an official XML such tool rather than rolling our own. That said, turns out I didnt need this.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2402,8 +2402,9 @@ case class DFDLInfoset(di: Node, parent: Infoset) {
     // The expected infoset is loaded using the normalizeCRLFtoLF mode (which
     // is the default), so MS-DOS/Windows CRLFs in expected data XML files will
     // have been converted into LF at this point.
-    if (elem ne null)
+    if ((elem ne null) && elem.toString.contains("\r")) {
       Assert.invariant(!elem.toString.contains("\r"))
```
stevedlawrence, line 2406: Thoughts on if we want to keep this assertion? This is going to convert every epected infoset to a string just to make sure there's no \r. Feels like that might be somewhat expensive?

mbeckerle: Once I get this debugged I will remove this one. Yes it is way too costly.

tuxji:
Should I review now or wait?  I wanted to wait until all the checks pass (which they do now), but your comments about some tests needing unexplainable changes makes me wonder if we should wait for more investigation.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/SchemaCache.scala
```diff
@@ -33,6 +33,21 @@ object ENoWarnTDML { EqualitySuppressUnusedImportWarning() }
...
+  case class Key (
+    uri: URISchemaSource,      // The DFDL schema's URI
+    useSerializedProcessor: Boolean,
+    compileAllTopLevels: Boolean,
+    optRootName: Option[String],
+    optRootNamespace: Option[String],
+    tunables: Map[String, String])  // tunables map - required since tunables effect compilation
+  // (ex: parseUnparsePolicy, unqualifiedPathStepPolicy, perhaps others)
+
+}
```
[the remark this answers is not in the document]
mbeckerle: This is part of fixing a bug where we needed the tunables in the SchemaCache key or they would return a cached result incorrectly when tunables had changed the compilation parameters around namespaces. That got aggravated as part of debugging SAX tests with the various namespace calls. 

It's an interesting point though. If as part of debugging a problem you fix things and improve them that turn out to have been unrelated to your orignal problem, do you back out those changes? Often it's not easy to identify what goes with what, and one could spend a lot of time backing things out and breaking stuff to isolate the changes. 

I cannot tell you if, in conjunction with the other changes, this change to the schema cache would still cause a failure. But it was not correct before because the tunables weren't part of the key. So this is more correct, albeit possibly unrelated. I really just do not know any more if this fix would have mattered.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetInputter.scala
```diff
@@ -68,22 +67,20 @@ class TDMLInfosetInputter(val scalaInputter: ScalaXMLInfosetInputter, others: Se
...
     val res = scalaInputter.getSimpleText(primType)
     val resIsEmpty = res == null || res == ""
-    val othersmatch = others.forall { i =>
-      val st = i.getSimpleText(primType)
+    val otherStrings = others.map { i =>
+      val firstVersion = i.getSimpleText(primType)
+      val finalVersion = i match {
+        case _ if (firstVersion eq null) => ""
+        case jsonii: JsonInfosetInputter => firstVersion.replaceAll("(\r\n|\r)", "\n").
+          replaceAll("\r", "\n")
```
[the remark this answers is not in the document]
mbeckerle: fixed

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetInputter.scala
```diff
@@ -68,22 +67,20 @@ class TDMLInfosetInputter(val scalaInputter: ScalaXMLInfosetInputter, others: Se
...
+        case _ => firstVersion
+      }
+      finalVersion
+    }
+    val othersmatch = otherStrings.forall { case st: String =>
       val stIsEmpty = st == null || res == ""
```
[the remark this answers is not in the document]
mbeckerle: Good catch. I'll fix this and see if anything breaks.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetInputter.scala
```diff
@@ -68,22 +67,20 @@ class TDMLInfosetInputter(val scalaInputter: ScalaXMLInfosetInputter, others: Se
...
       val areSame = res == st || (resIsEmpty && stIsEmpty)
-      if (areSame) {
-        true
-      } else {
-        if (i.isInstanceOf[JsonInfosetInputter]) {
-          // the json infoset inputter maintains CRLF/CR, but XML converts CRLF/CR to
-          // LF. So if this is Json, then compare with the CRLF/CR converted to LF
```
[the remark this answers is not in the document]
mbeckerle: added.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
mbeckerle: I've removed this from here. It now only saves and reloads if useSerializedProcessor. 
This is the compiler's problem to insure that everything is fully evaluated, not code out here using the data processor.

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLUnparseCases.scala
mbeckerle: yup

project/Dependencies.scala
mbeckerle: removed.

mbeckerle (commented):
Responded to all the review comments. Quite a bit of cleanup here.

There are a couple of CLI tests where I left in LOUD comments to review them, because I'm not sure why I had to change them at all.

mbeckerle:
I just pushed fixes to all SL's comments.

Barring major failure again on the CI tests, yes now would be good for a review.

mbeckerle:
This change set also fixes DAFFODIL-1816

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -347,7 +346,11 @@ class DataProcessor private (
 
   override def getDiagnostics = ssrd.diagnostics
 
-  override def newXMLReaderInstance: DFDL.DaffodilParseXMLReader = new DaffodilParseXMLReader(this)
+  override def newXMLReaderInstance: DFDL.DaffodilParseXMLReader = {
+    val xrdr = new DaffodilParseXMLReader(this)
+    XMLUtils.setSecureDefaults(xrdr)
```
stevedlawrence, line 351: Thoughts on just removing this call? We aren't getting a random XMLReader here based on some factory that happens to be on the classpath. We know it's a Daffodil specific XML reader, and we know it's secure from these XML issues. Removing this would potentially mean we could remove the new logic added in set/getFeature to accept features that we don't really support.

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -1265,20 +1265,29 @@ class TestCLIdebugger {
...
       shell.sendLine("step")
-      shell.sendLine("step")
-      shell.sendLine("step")
+      //
+      // NOTE FOR REVIEWER
+      //
+      // This test had to be modified to work with this change set.
+      // I have no idea why this change set would affect behavior
+      // of this kind.
+      //
```
stevedlawrence, line 1274: Maybe related to the serilazation stuff? Maybe forcing serialization is cause different code paths to be triggers? Or maybe thing are getting evaluated in different orders, which somehow changes thigns? Doesn't seem likely, but the only thing I can think of. Does this behavior switch back if we remove your serialization changes?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -72,37 +74,61 @@ trait SchemaSetRuntime1Mixin { self : SchemaSet =>
...
+    } else {
+      log(LogLevel.Compile, "Parser = %s.", ssrd.parser.toString)
+      log(LogLevel.Compile, "Unparser = %s.", ssrd.unparser.toString)
+      log(LogLevel.Compile, "Compilation (DataProcesor) completed with no errors.")
+    }
+    dataProc
   }
 
-}
+}
```
stevedlawrence, line 134: This file is missing a new line. Not sure why that would be removed. Did you IDE settings change?

stevedlawrence:
There was a comment I found in an email but can't currently find it in github, whree @mbeckerle said:

> Gone. However, I would say, that this pluggable XML readers stuff feels like stuff of a bygone era when everyone thought there would be competition around XML libraries, and pluggable XML technologies from vendors, etc. This seems silly in retrospect. But maybe it is needed for backward bug-for-bug compatibility in some cases.
>
>For us, I don't think pluggable XMLReader makes any sense. We're far too sensitive to exactly their characteristics.
>
> think we should explicitly check for the exact XMLReader classes we support, and error if it isn't that.
>
>Thoughts on this check?

I'm wondering if we should just be explicitly creating the XML Reader that we are expecting? As you said, we depend on very specific behavior from a specif XMLReader, so using something like ``SAXParserFactory.newInstance().newSAXParser().getXMLReader`` to get some XMLReader seems like a mistake. I'm wondering if there's a way to directly create a Xerces XMLReader. Maybe we should use the overloaded newInstance function that takes a class name? Or maybe we can directly create the right class ourselves, and avoid this pluggable logic? Then we really can assume that we have a certain XMLReader.

Would it then even be possible to have a daffodil-lib function that gets the things XMLReader stuff we need and set the secure defaults, so there's only one place we need to call setSecureDefaults, with the assumptino that everything gets the XMLReader from our new daffodil-lib function?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -72,37 +74,61 @@ trait SchemaSetRuntime1Mixin { self : SchemaSet =>
...
+    // Note that calling preSerialization is not sufficient, since that's only mixed into
+    // objects with lazy evaluation. A SSRD is just a tuple-like object, does not mixin
+    // preSerialization, and shouldn't need to. We need to
+    // serialize all its substructure to insure all preSerializations, that force
+    // all lazy evaluations, are done.
+    //
+    // Overhead-wise, this is costly, if the caller is about to save the processor themselves
+    // But as there have been cases of Runtime1 processors which end up doing lazy evaluation
+    // that ends up happening late, this eliminates a source of bugs, albeit, by masking them
+    // so they are not detectable.
```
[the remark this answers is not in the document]
mbeckerle: https://issues.apache.org/jira/browse/DAFFODIL-2526

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
mbeckerle: Another way to finesse this is to make DaffodilParseXMLReader use a factory and have a private constructor. That way how it is initialized after construction need not be the caller's responsibility.  I will change this.

stevedlawrence: I'd argue that calling setSecureDefaults is no one's responsiblity, including ours. So a factory just adds extra API complexity. The DaffodilXMLReader doesn't actually read XML (only implements the XMLReader API)--there are no concerns about reading XML securely, so no one, including us, ever needs to set it in a secure mode.

mbeckerle: Good point. I think we need scaladoc to point out that this is by definition secure if we remove the ability to set the secure features, otherwise someone's going to try it, and end up not liking that they can't set it secure.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
mbeckerle: No settings change. I've seen a few of these around here and there.

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
mbeckerle: I commented out the new serialization stuff, and there is no change to the behavior of these tests. 

I remain mystified. 

It appears that some alignment optimizations are working now, so that alignment unparsers aren't being created any longer in one of these tests anyway. That or they're being resolved and aren't in the data structure any more when it is being displayed.

mbeckerle:
re: constructing the specific SAXParser class we need. I've introduced a new DaffodilSAXParserFactory, which insures that the SAXParser one gets is exactly the Xerces one. Also insures that its XMLReader is secure.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -257,7 +257,7 @@
  *  WritableByteChannel wbc = java.nio.channels.Channels.newChannel(os);
  *  DaffodilUnparseContentHandler unparseContentHandler = dp.newContentHandlerInstance(wbc);
  *  try {
- *   XMLReader xmlReader = SAXParserFactory.newInstance().newSAXParser().getXMLReader();
+ *   XMLReader xmlReader = SAXParserFactory().newSAXParser().getXMLReader;
```
stevedlawrence, line 260: Don't you need the SAXParserFactory.newInstance() in Java? I don't think you can just do SAXParserFactory()?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilSAXParserFactory.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.xerces.jaxp.SAXParserFactoryImpl
+
+
+class DaffodilSAXParserFactory private ()
+  extends SAXParserFactoryImpl() {
+
+  override def newSAXParser() = {
+    val res = super.newSAXParser()
+    XMLUtils.setSecureDefaults(res.getXMLReader)
```
stevedlawrence, line 28: Does this make it so all future calls to getXMLReader will have secure defaults set? I assumed that getXMLReader would allocate a new instance everytime it's call, and so would need setSecureDefaults set on it?

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -350,7 +350,7 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor) extends
     val saxOutputChannel = java.nio.channels.Channels.newChannel(saxOutputStream)
     val unparseContentHandler = dp.newContentHandlerInstance(saxOutputChannel)
     unparseContentHandler.enableInputterResolutionOfRelativeInfosetBlobURIs()
-    val xmlReader = SAXParserFactory.newInstance.newSAXParser.getXMLReader
+    val xmlReader = DaffodilSAXParserFactory().newSAXParser.getXMLReader
```
stevedlawrence, line 353: This is still calling setSecureDfeaults below. Is that still needed?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilSAXParserFactory.scala
stevedlawrence: Looking at the Xerces code, I'm wrong. A SAXParser does have only a single xmlReader, and getXMLReader always returns that same reader. So what you have is correct.

mbeckerle: Yeah, it's a getter, not an allocator. But I can also override so that getXMLReader is our override, and then it's 100% clear.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
mbeckerle: No, I missed taking that one out.

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
mbeckerle: Ah, yes. That's a scala-ism that made it into the comments of the JAPI.

project/Dependencies.scala
```diff
@@ -48,7 +48,7 @@ object Dependencies {
   lazy val test = Seq(
     "junit" % "junit" % "4.13.2" % "it,test",
     "com.novocode" % "junit-interface" % "0.11" % "it,test",
-    "org.scalacheck" %% "scalacheck" % "1.15.4" % "it,test"
+    "org.scalacheck" %% "scalacheck" % "1.15.4" % "it,test",
   )
```
tuxji, line 52: I don't mind adding a trailing comma to the last item in a Seq (Scala explicitly [allows](https://docs.scala-lang.org/sips/trailing-commas.html) it for good reasons), but the previous Seq's don't use a trailing comma so we've introduced an inconsistency.  Maybe fix all the other Seqs too?

tuxji (approved):
+1

We're getting close to the end.  I think just a few minor cleanups are left.

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -48,9 +49,31 @@
...
+    public static void setSecureDefaults(XMLReader xmlReader)
+            throws SAXNotSupportedException, SAXNotRecognizedException {
+        xmlReader.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        // since we're not really sure what they mean by secure processing
+        // we make doubly sure by setting these ourselves also.
+        xmlReader.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+        xmlReader.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+        xmlReader.setFeature("http://xml.org/sax/features/external-general-entities", false);
+    }
+
```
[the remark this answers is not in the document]
stevedlawrence: I'm not sure if this is officially, but, I only think of things in japi/sapi as being part of the public API. Other public classes/functions people can use, but they aren't neesssarily stable and we won't make an attempt at backwards compatibility.

Since XMLUtils isn't part of sapi/japi, I would prefer that we not use it in japi/sapi tests since we often point people towards these tests for how to use API. That way we aren't suggesting the use of non public API.

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -1008,6 +1031,7 @@ public void testJavaAPI20() throws IOException, ClassNotFoundException {
         try {
             org.xml.sax.XMLReader unparseXMLReader = javax.xml.parsers.SAXParserFactory.newInstance()
                     .newSAXParser().getXMLReader();
+            setSecureDefaults(unparseXMLReader);
```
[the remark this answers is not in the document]
stevedlawrence: I'd argue they should do neither. Both are internal use only functions. Any one creating a XMLReader outside of Daffodil (like for use with SAX unparsing) should be seureing their own XMLReader. We may want to consider making a public XMLUtils library at some point, but right now we don't have that.

stevedlawrence:
Had one more though, this mostly changes how we parse things like TDML files, DFDL Schemas, etc. But I *think* it leaves out some of the InfosetInputters? I assume the Scala, JDOM, and W3CDOM don't have these security issues since the XML was parsed into these data structures and so are safe to use. But the XMLTextInfosetInputter and JSON use the woodstox and jackson libraries, respectively. Neither of these use an XMLReader of parse. The XML one is StAX which uses a different API. And the JSON one is custom, but similar to StAX.

It looks like the Woodstox StAX properties are defined in three blog posts:
https://cowtowncoder.medium.com/configuring-woodstox-xml-parser-basic-stax-properties-39bdf88c18ec
https://cowtowncoder.medium.com/configuring-woodstox-xml-parser-stax2-properties-c80ef5a32ef1
https://cowtowncoder.medium.com/configuring-woodstox-xml-parser-woodstox-specific-properties-1ce5030a5173

I think these are the javadocs for what an XMLInputFactory supports:
https://docs.oracle.com/javase/6/docs/api/javax/xml/stream/XMLInputFactory.html
https://fasterxml.github.io/woodstox/javadoc/6.2/com/ctc/wstx/api/WstxInputProperties.html

Do we need to set something similar for these? Could probably be done in a separate bug/Pr if you want. This one is getting kind of big.

stevedlawrence:
+1 with @tuxji comments addressed

daffodil-japi/src/main/java/org/apache/daffodil/japi/package-info.java
```diff
@@ -257,7 +257,7 @@
  *  WritableByteChannel wbc = java.nio.channels.Channels.newChannel(os);
  *  DaffodilUnparseContentHandler unparseContentHandler = dp.newContentHandlerInstance(wbc);
  *  try {
- *   XMLReader xmlReader = SAXParserFactory.newInstance().newSAXParser().getXMLReader();
+ *   XMLReader xmlReader = SAXParserFactory.newInstance().newSAXParser().getXMLReader;
```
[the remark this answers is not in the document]
mbeckerle: Will fix. I have never been a fan of  having code snippets in javadoc/scaladoc. It is almost guaranteed to not work. I prefer to put it into examples or tests, and put a URL in the javadoc/scaladoc referring people to that code.

mbeckerle (commented):
I believe the just pushed changes (2 more commits) address all the comments.

I did add tests and changes to XMLTextInfosetInputter and JDOMInfosetInputter to reject DOCTYPE/DTDs in those also.

mbeckerle (commented):
I believe the just pushed changes (2 more commits) address all the comments.

I did add tests and changes to XMLTextInfosetInputter and JDOMInfosetInputter to reject DOCTYPE/DTDs in those also.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetInputter.scala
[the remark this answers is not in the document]
tuxji: What was the resolution of the comments on this line of code?

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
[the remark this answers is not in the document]
tuxji: I have a slightly different suggestion.  If we ever make secure XML loading part of the published API, let's publish DaffodilSAXParserFactory instead of XMLUtils.  I know there's also JDOM parsing and so on, but we don't have to offer every way to load XML securely in the published API, only the way which we recommend as the best way.  When people use DaffodilSAXParserFactory, they insure that the SAXParser they get is exactly the Xerces one and its XMLReader is secure just by changing their first line of code.  That's the easiest way to load XML securely - just a one-line change :).

tuxji:
+1

I reviewed the three new commits.  First and third commits indeed address all my comments and the changes look great.  Second commit has a slight problem with test coverage.  Your @Test def testXMLTextInfosetInputterDisallowsDocType() tried several ways to load XML securely with the Woodstox XML library and found the best way was to both turn off features and add a doctype check / throw exception in XMLTextInfosetInputter.scala.  I believe you then got inspired to add a doctype check / throw exception to a similar place in JDOMInfosetInputter.scala but that line doesn't get called by any test either because we either need a new test to check JDOM secure loading or JDOM doesn't let programs see the doctype at all when you enable secure XML loading.  If the latter is the case, It's probably a good idea to keep the check in case of malicious input and just add a comment saying so along with comments to disable codecov.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/TDMLInfosetInputter.scala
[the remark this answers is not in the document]
mbeckerle: Fixed. The change seemed to cause no changes of behavior detected by any test.

stevedlawrence (approved):
+1

PR #561, opened 2021-05-13
[DAFFODIL-1437] : DFA Rules - eliminate anonymous classes and functions

sndp2693:


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
```diff
@@ -123,6 +127,37 @@ abstract class State(states: => ArrayBuffer[State]) extends Serializable {
...
+
+    override def act(r: Registers): Unit = r.nextState = DFA.EndOfData
+  }
+
+  object LookAheadEndOfDataCharRule extends Rule {
+    override def test(r: Registers): Boolean = r.data1 == DFA.EndOfDataChar
+
+    override def act(r: Registers): Unit = {
+      r.appendToField(r.data0)
+      r.nextState = DFA.EndOfData
```
mbeckerle, line 148: I really like this code cleanup.  

I will create a separate JIRA issue about the codecov issues, because that's really not the responsibility of this change. It is of some concern to me if this specific code is not well covered by our tests.

mbeckerle (approved):
+1
Of the automated checks. One failed. I believe that is spurious and does not reflect an actual problem.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
mbeckerle: https://issues.apache.org/jira/browse/DAFFODIL-2515

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/Rules.scala
```diff
@@ -104,7 +106,9 @@ abstract class State(states: => ArrayBuffer[State]) extends Serializable {
     val res =
       if (pTerm0.isInstanceOf[WSPStarState]) {
         val wspStar = pTerm0.asInstanceOf[WSPStarState]
-        if (wspStar.checkMatch(charIn)) { true } // Was a space
+        if (wspStar.checkMatch(charIn)) {
+          true
+        } // Was a space
```
tuxji, line 111: Comment should be moved to previous line and push previous line below it, e.g.,

```scala
          // Was a space.
          //
          true
```

This makes this branch look similar to the following branches and therefore makes the entire expression easier to read.

tuxji (approved):
+1

Difficult to determine whether refactoring changes were made correctly and all redundant rules replaced from looking at diffs alone.  Indeed concerning that tests didn't cover all rules so a test might break if a replacement rule didn't match the original rule, but all existing tests do pass.  Steve, please take a look too.

stevedlawrence:
+1

Agreed with Mike and John, it's a bit concerning that there's some missing code covereage, but not your fault, and doesn't need to be fixed as part of this PR. Looks really nice otherwise.

@sndp2693, can you fix the comment issue mentioned by @tuxji, squash and force push, and then this should be good for us to merge! Thanks!

sndp2693:
Reopening PR after squashing commits into one.

sndp2693:
I am sorry i deleted remote branch while squashing commits. I have opened a new PR https://github.com/apache/daffodil/pull/568 which addresses review comments.

PR #564, opened 2021-05-14
Skip TDML Runner left over data check if negative bit position

stevedlawrence:
The IBM DFDL implementation has no way to report the final bit position
of a parse or unparse, and so just returns -1 for this value to the TDML
runner. The recent deprecation of isAtEnd in favor of bitPos caused this
-1 value to be used by the TDML runner, and think that there was always
left over data This changes the TDML runner so that if the bit position
is negative, then we skip the left over data check in TDML test.

DAFFODIL-2517


mbeckerle (approved):
+1

jadams-tresys (approved):
+1

PR #565, opened 2021-05-14
Improve release candidate script handling of errors

stevedlawrence:
- Because of the use of `set -e`, any failure that occurs when running
  the daffodil-release-candidate script will immediately halt the
  script. This is the right behavior, but there in some cases there
  isn't an obvious error as the script just stops. Furthermore, it's not
  obvious what steps to take. This traps ERR and calls a function that
  will make it very clear an error occurred and directs the user what to
  do.
- When committing changes to the website, it's possible there are no
  actual changes. This is very likely in the case of an rc2 where the
  docs were already updated for rc1. If there are not changes, git
  commit exits with a non-zero exit code, which causes processing to
  stop. To avoid this, check to see if there are any changes to commit,
  and only then call git commit.
- Fix typo in "Committing site changes"
- Replace "Delete" with "Drop" when referencing how to undo staged jars.
  This is the correct terminology that show up at repository.apache.org

  DAFODIL-2518


tuxji (approved):
+1

Line 303 prints an instruction for the release manager:

  echo "- Close the staged files at https://repository.apache.org"

I read the [Release Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Release+Workflow) and it also said the same thing in step 7: b. Close the staged file at https://repository.apache.org.

The problem is that I don't know what "staged file" means in this context (a "git add" creates a staged file, is that what the instruction means?) or how you "close" a staged file.  Could you please make the Release Workflow clearer?  I would like to see the exact command you actually type at that point.

stevedlawrence:
"Closing the staged files" isn't actually a command. You actually have to log in to https://repository.apache.org in a browser with your ASF credentials, find the daffodil staged release, check the box, and press the "Close" button (same idea for "Drop" if somethings goes wrong). I'll look to see if there is maybe a POST or something we can do to autmoate this, and if not I'll update the workflow to make it more clear.

stevedlawrence:
I think there is an API, but you need to log in to the website to figure out what the right id is (It's always something like orgapachedaffodil-XXXX, where XXXX increments for each staged release). Probably not worth the effort to automate things when you already need to login to get that id.

Also, earlier before the "close" step, when talking about how to verify the files at repository.apache.org, it says;

> To do so, visit that url, login in the top right using id.apache.org credentials, select "Staging Repositories" on the left, and find the orgapachedaffodil-XXXX repository. Inspect the "Content" tab to make sure the appropriate jars are uploaded and appear valid.

So before talking about closing/droping, we do talk about what this repository thing is, and how to interact with it.  But I've also updated the bullets for close/drop/release to be more specific about what to do for those individual steps.

jadams-tresys (approved):
+1

PR #568, opened 2021-05-18
[DAFFODIL-1437] : DFA Rules - eliminate anonymous classes and functions

sndp2693:


sndp2693:
Links to https://github.com/apache/daffodil/pull/561

stevedlawrence:
Thanks! I've confirmed that code in this PR is exactly the same as that in #561, so all +1's carry over. We can merge this once 3.1.0 is released.

@sndp2693, Also I just noticed this, and this isn't a big deal and you don't have to change this, but for future commits, the convention we have been using for commit messages is to put the bug number at the end of the commit message instead of the subject. So for example, the commit message should look like this instead:
```
DFA Rules - eliminate anonymous classes and functions

DAFFODIL-1437
```

PR #570, opened 2021-05-19
Add restrictions to delimiters for escapeCharacters

jadams-tresys:
This commit disallows separators/terminators that start with the same
character as the escapeCharacter. This was something that is missing
from the spec, but will likely be added in the near future. Having a
separator or terminator that starts with the escapeCharacter will cause
a Schema Definition Error to occur, which is how IBM's DFDL
implementation works. This change results in many tests being
redundant as it is no longer necessary to test these sort of edge
cases.

This also fixes a small issue that was preventing some escape scheme
tests from round tripping.

DAFFODIL-1556

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
@@ -48,6 +48,7 @@ import org.apache.daffodil.schema.annotation.props.gen.LengthKind
 import org.apache.daffodil.schema.annotation.props.gen.ChoiceAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.OccursAGMixin
 import org.apache.daffodil.schema.annotation.props.gen.Representation
+import org.apache.daffodil.schema.annotation.props.gen.EscapeKind
 import org.apache.daffodil.schema.annotation.props.gen.Sequence_AnnotationMixin
 import org.apache.daffodil.processors.LengthInBitsEv
 import org.apache.daffodil.schema.annotation.props.gen.LengthUnits
@@ -649,6 +650,36 @@ trait SequenceRuntimeValuedPropertiesMixin
           SDE(
             "The dfdl:terminator and dfdl:separator properties must be distinct. Both contain: %s.",
             inBoth.map { s => "'" + s + "'" }.mkString(", "))
+        }
+      }
+    }
+  }
+
+  def checkDelimiterEscapeConflict(childTerm: Term): Unit = {
+    if (childTerm.optionEscapeScheme.isDefined &&
+        (childTerm.optionEscapeScheme.get.escapeKind == EscapeKind.EscapeCharacter) &&
+        (hasTerminator || hasSeparator)) {
+      val ecEv = childTerm.optionEscapeScheme.get.escapeCharacterEv
+      if (ecEv.isConstant) {
+        val ec = ecEv.constValue
+        var delims = Set[String]()
+
+        val termEv = this.terminatorParseEv
+        if (termEv.isConstant)
+          delims = delims ++ termEv.constValue.map { _.lookingFor }
+
+        val sepEv = this.separatorParseEv
+        if (sepEv.isConstant)
+          delims = delims ++ sepEv.constValue.map { _.lookingFor }
+
+        if (delims.exists { _.startsWith(ec) } )
+          SDE("The dfdl:terminator and dfdl:separator may not begin with the dfdl:escapeCharacter: '%s'.", ec)
+
+        val maybeEecEv = childTerm.optionEscapeScheme.get.optionEscapeEscapeCharacterEv
+        if (maybeEecEv.isDefined && maybeEecEv.get.isConstant) {
+          val eec = maybeEecEv.get.constValue
+          if (delims.exists { _.startsWith(eec) } )
+            SDE("The dfdl:terminator and dfdl:separator may not begin with the dfdl:escapeEscapeCharacter: '%s'.", eec)
         }
       }
     }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextDelimitedUnparser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextDelimitedUnparser.scala
@@ -248,6 +248,8 @@ class TextDelimitedUnparser(override val context: TermRuntimeData)
     escapeEscapeChar: MaybeChar, state: UState): (String, Boolean) = {
     Assert.invariant(delims != null)
     Assert.invariant(field != null)
+    if (hasEscCharAsDelimiter)
+      UnparseError(One(context.schemaFileLocation), One(state.currentLocation), "The dfdl:terminator and dfdl:separator may not begin with the dfdl:escapeCharacter: '%s'.", escapeChar)
 
     val successes: ArrayBuffer[(DFADelimiter, Registers)] = ArrayBuffer.empty
     val fieldReg: Registers = state.dfaRegistersPool.getFromPool("escapeCharacter1")
@@ -305,6 +307,7 @@ class TextDelimitedUnparser(override val context: TermRuntimeData)
               case _ => {
                 // this delim did not match, ignore it and discard its register
                 state.dfaRegistersPool.returnToPool(delimReg)
+                input.resetPos(beforeDelimiter)
               }
             }
           }
@@ -320,8 +323,7 @@ class TextDelimitedUnparser(override val context: TermRuntimeData)
             // matched a delimiter, need to handle escaping it
             val (matchedDelim, matchedReg) = longestMatch(successes).get
             if (matchedDelim.lookingFor.length() == 1 && matchedDelim.lookingFor(0) =#= escapeChar) {
-              if (hasEscCharAsDelimiter) { fieldReg.appendToField(escapeChar) }
-              else if (escapeEscapeChar.isDefined)
+              if (escapeEscapeChar.isDefined)
                 fieldReg.appendToField(escapeEscapeChar.get)
               else
                 UnparseError(One(context.schemaFileLocation), One(state.currentLocation), "escapeEscapeCharacter was not defined but the escapeCharacter (%s) was present in the data.", escapeChar)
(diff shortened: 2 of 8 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextDelimitedUnparser.scala
```diff
@@ -248,6 +248,8 @@ class TextDelimitedUnparser(override val context: TermRuntimeData)
     escapeEscapeChar: MaybeChar, state: UState): (String, Boolean) = {
     Assert.invariant(delims != null)
     Assert.invariant(field != null)
+    if (hasEscCharAsDelimiter)
+      UnparseError(One(context.schemaFileLocation), One(state.currentLocation), "Delimiters may not begin with the dfdl:escapeCharacter (%s)", escapeChar)
```
mbeckerle, line 252: Surprising this is uncovered. I guess you need a specifically unparserTestCase, because a negative parserTestCase will never get to the unparse.

tuxji, line 252: If every schema and every TDML test fails the checkDelimiterEscapeConflict check before they can hit this line, it may be difficult to write a test that covers this line.  We'd have to write a unit test that calls this specific method to hit this line or tell Codecov to ignore this line.

tuxji (approved):
+1  See note below

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -654,6 +655,36 @@ trait SequenceRuntimeValuedPropertiesMixin
...
+          delims = delims ++ sepEv.constValue.map { _.lookingFor }
+
+        if (delims.exists { _.startsWith(ec) } )
+          SDE("The dfdl:terminator and dfdl:separator may not begin with the dfdl:escapeCharacter: '%s'.", ec)
+
+        /*val maybeEecEv = childTerm.optionEscapeScheme.get.optionEscapeEscapeCharacterEv
+        if (maybeEecEv.isDefined && maybeEecEv.get.isConstant) {
+          val eec = maybeEecEv.get.constValue
+          if (delims.exists { _.startsWith(eec) } )
+            SDE("The dfdl:terminator and dfdl:separator may not begin with the dfdl:escapeEscapeCharacter: '%s'.", eec)
```
stevedlawrence, line 682: Why don't we also check EEC? Does that not have the same restriction?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
```diff
@@ -654,6 +655,36 @@ trait SequenceRuntimeValuedPropertiesMixin
     }
   }
 
+  def checkDelimiterEscapeConflict(childTerm: Term): Unit = {
+    if (childTerm.optionEscapeScheme.isDefined &&
+        (childTerm.optionEscapeScheme.get.escapeKind == EscapeKind.EscapeCharacter) &&
+        (hasTerminator || hasSeparator)) {
+      val ecEv = childTerm.optionEscapeScheme.get.escapeCharacterEv
+      if (ecEv.isConstant) {
```
stevedlawrence, line 663: This only checks for constant escape characters and constant terminators/separators. And it only checks those defined on the same element. Do we also need to check all in-scope terminating markup? Which really can only easilyl be checked at runtime? Not sure how much overhead that might have though?

daffodil-test/src/test/scala/org/apache/daffodil/section07/escapeScheme/TestEscapeScheme.scala
```diff
@@ -104,43 +104,10 @@ class TestEscapeScheme {
...
-  @Test def test_scenario4_8(): Unit = { runner2.runOneTest("scenario4_8") }
-  @Test def test_scenario4_8_postfix(): Unit = { runner2.runOneTest("scenario4_8_postfix") }
-  @Test def test_scenario4_9(): Unit = { runner2.runOneTest("scenario4_9") }
   @Test def test_scenario4_9_req_term(): Unit = { runner2.runOneTest("scenario4_9_req_term") }
-  @Test def test_scenario4_10(): Unit = { runner2.runOneTest("scenario4_10") }
   @Test def test_scenario4_10_req_term(): Unit = { runner2.runOneTest("scenario4_10_req_term") }
-  @Test def test_scenario4_11(): Unit = { runner2.runOneTest("scenario4_11") }
-  @Test def test_scenario4_11_postfix(): Unit = { runner2.runOneTest("scenario4_11_postfix") }
-  @Test def test_scenario4_12(): Unit = { runner2.runOneTest("scenario4_12") }
   @Test def test_scenario4_12_req_term(): Unit = { runner2.runOneTest("scenario4_12_req_term") }
```
stevedlawrence, line 112: Have you tested against the a regression suite to make sure this doesn't break anything? I doubt any schemas run into this issue, but we supported it before so I guess it's possible?

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
jadams-tresys: I wasn't sure, which is why I had left this commented out.  I'm not sure how IBM behaves for this case.  I meant to mention that when I pushed it up, but forgot.

daffodil-test/src/test/scala/org/apache/daffodil/section07/escapeScheme/TestEscapeScheme.scala
jadams-tresys: Just tested against the full dfdl-schemas repo and everything passes, with one exception of a single vCard test, which also fails on master and doesn't appear to be related to my changes.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
jadams-tresys: I did some digging into adding runtime checks, which on the unparsing side was easy.  Parsing, however, did not seem straightforward.  It wouldn't be too bad to add checks agains initiators and terminators, but there is kind of a gap with how the escape scheme is part of the individual elements but the separator is part of the parent sequence.  I might be missing something obvious, but I didn't see a straight forward solution to this that wouldn't involve significant changes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextDelimitedUnparser.scala
jadams-tresys: I was able to write a test that covers this without needing to do a unit test.  Just needed a test where the separator is determined at runtime.

jadams-tresys: Test has been added that covers this now.  It needed a separator that was determined at runtime during unparsing.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
jadams-tresys: Does anyone else have any thoughts on if this code should be added or not?  Perhaps I send a message to the dev list.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
stevedlawrence: Maybe we should try to match IBM DFDL behavior? Could we write some tests and see how they handle conflicts with runtime escape schemes/delimiters? Can also determine how they handle EEC as well.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/RuntimePropertyMixins.scala
mbeckerle: Two points:

(1) allowing the EEC to be the same as the start of a delimiter enables what? What obscure format is possible that isn't otherwise - The only example I can contrive is that a format that a format that uses "/" separators could be used as an encapsulation around a format that uses "/" as the EEC. So it's a composition property. I have never seen anything like that, but it may exist somewhere, or a need for it could arise. 

(2) allowing these sorts of overlaps usually just makes bugs or oversights hard for users to spot. We're not doing anybody a favor by allowing this if there is in fact no actual use case. 

The usual decision on these sorts of things is to disallow, knowing we could relax the restriction later. This is conservative in that it preserves freedom to change in the future. 

The only issue in this case is we're adding this restriction late. 

I suggest we add the restriction with the understanding that we might have to back this out, should compatibility issues arise. If those do, helping users reformulate their schemas so as to side-step the problem would be good to try first. 
 
We do have a task, I started on it, but set it aside long ago, to run the Daffodil test suite against IBM DFDL so as to narrow down on test cases that demonstrate differences in behavior.  Probably we should pick and choose areas of it to cross test instead of trying to do all of it. Just delimiter and escape related tests is one such area. That ticket is DAFFODIL-2028. The last commit on the currently parked branch for that says 140 tests fail from daffodil-test. The issues blocking are mostly date/time simple type-related stuff, so it's possible the tests for separators and escapes will cross test now.

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScenarios.tdml
```diff
@@ -551,74 +409,6 @@
...
-    <document>foo$$$//$;</document>
-    <infoset>
-      <dfdlInfoset>
-        <tns:e_infix>
-          <x>foo$$/$;</x>
-  </tns:e_infix>
-      </dfdlInfoset>
-    </infoset>
-  </parserTestCase>
-
```
mbeckerle, line 625: The comment line just below should get a bug report. A bad diagnostic like that is a bug.

mbeckerle (approved):
+1 I am good with putting in the EEC checking. This is a dev branch after all. We can see how it shakes out.

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScenarios.tdml
jadams-tresys: Just double checked this one and the only message that gets displayed now is "Parse Error: Terminator '$;' not found" so something must have changed since it was originally written.  I'll remove the comment.

PR #571, opened 2021-05-19
Fix ULong. ULong.MaxValue is wrong.

mbeckerle:
Add more tests to cover this.

DAFFODIL-2522

### DIFF
--- a/daffodil-lib/src/main/scala/passera/unsigned/ULong.scala
+++ b/daffodil-lib/src/main/scala/passera/unsigned/ULong.scala
@@ -43,7 +43,10 @@ case class ULong(override val longValue: Long) extends AnyVal with Unsigned[ULon
   override def toLong: Long = rep
   override def toFloat: Float = (rep & 0x7fffffffffffffffL).toFloat
   override def toDouble: Double = (rep >>> 1).toDouble * 2.0 + (rep & 1L)
-  def toBigInt: JBigInt = if (rep >= 0) JBigInt.valueOf(rep) else ULong.MaxValueAsBigInt.subtract(JBigInt.valueOf(rep.abs)).add(JBigInt.ONE)
+
+  def toBigInt: JBigInt =
+    if (rep >= 0) JBigInt.valueOf(rep)
+    else JBigInt.valueOf(rep).and(ULong.MaxValueAsBigInt)
 
   // override def intValue = rep
   override def byteValue = toByte
@@ -156,10 +159,8 @@ case class ULong(override val longValue: Long) extends AnyVal with Unsigned[ULon
   override def toString =
     if (rep >= 0L)
       rep.toString
-    else if (rep == Long.MinValue)
-      ULong.maxULongString
     else
-      this.toBigInt.toString // (~(rep - 1)).toString
+      this.toBigInt.toString
 
   def toHexString = rep.toHexString
   def toOctalString = rep.toOctalString
@@ -240,6 +241,17 @@ object ULong {
   val Zero = MinValue
   val MaxValue = ULong(~0L)
 
-  val MaxValueAsBigInt = JBigInt.valueOf(Long.MinValue).abs
-  private val maxULongString = MaxValueAsBigInt.toString()
+  val MaxValueAsBigInt = new JBigInt("FFFFFFFFFFFFFFFF", 16)
+
+  def apply(bi: JBigInt): ULong = {
+    val posBigInt = bi.and(ULong.MaxValueAsBigInt)
+    ULong(posBigInt.longValue)
+  }
+
+  def apply(digits: String, radix: Int): ULong = {
+    val bi = new JBigInt(digits, radix)
+    ULong(bi)
+  }
+
+  def fromHexString(hexDigits: String) = ULong(hexDigits, 16)
 }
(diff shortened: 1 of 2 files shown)


daffodil-lib/src/main/scala/passera/unsigned/ULong.scala
```diff
@@ -240,6 +241,19 @@ object ULong {
...
+  val MaxValueAsBigInt = new JBigInt("FFFFFFFFFFFFFFFF", 16)
+
+  def apply(bi: JBigInt): ULong = {
+    val posBigInt = bi.and(ULong.MaxValueAsBigInt)
+    ULong(posBigInt.longValue)
+  }
+
+  def apply(digits: String, radix: Int): ULong = {
+    if (digits.contains("-") || digits.contains(","))
+      throw new IllegalArgumentException("digits must contain only 0-9A-Fa-F.")
```
stevedlawrence, line 253: Can we rely on JBigInt to verify the correctness of the digits string + radix? And if the minus sign is a problem, we can error if bi is negative, but maybe there's even a reason to allow that?

daffodil-lib/src/test/scala/passera/test/TestULong.scala
```diff
@@ -64,5 +63,45 @@ class TestULong {
...
+    assertEquals(-2, ULong(minusTwo).toInt)
+    assertEquals("7FFFFFFFFFFFFFFF", (ULong(Long.MinValue) - ULong(1)).toHexString.toUpperCase)
+  }
+
+  @Test def testULongShift: Unit = {
+    // NO sign extension since a ULong has no sign bit.
+    assertEquals("7FFFFFFFFFFFFFFF", (ULong.MaxValue >> 1).toHexString.toUpperCase)
+    assertEquals("7FFFFFFFFFFFFFFF", (ULong.MaxValue >>> 1).toHexString.toUpperCase)
+    assertEquals(1, (ULong.MaxValue >> 63).toInt)
+    assertEquals(ULong.MaxValue, ULong.MaxValue >> 64)
```
stevedlawrence, line 104: Not sure I understnd this. If we have 0xFFFFFFFFFFFFFFFF, so all bits are 1, and we shift 64-bits right, shouldn't we lose all bits, i.e. zero? Instead of getting back all bits set? I would think shifting width bits or greater would always be zero.

daffodil-lib/src/test/scala/passera/test/TestULong.scala
```diff
@@ -64,5 +63,45 @@ class TestULong {
...
+    assertEquals("FFFFFFFFFFFFFFFF", v.toHexString.toUpperCase)
+    assertEquals(ULong(0), v + ULong(1))
+    val v1 = ULong.MaxValue.toBigInt.add(JBigInt.TWO) // 0x8000000000000001
+    assertEquals(ULong(1), ULong(v1)) // preserves only 64 bits
+  }
+
+  @Test def testULongFromBigInt: Unit = {
+    val zero = JBigInt.ZERO
+    val one = JBigInt.ONE
+    val two = JBigInt.TWO
```
stevedlawrence, line 88: Looks like JBigInt.TWO wasn't added until Java 9. I guess we'll need ``JBigInt.valueOf(2)`` or something.

daffodil-lib/src/main/scala/passera/unsigned/ULong.scala
mbeckerle: Good point. We should not check things here, and just let JBigInt do it. 
This library has that "style" where it tries to be overly accepting of arguments and "make them work somehow" rather than my preferred style, which is to be as narrow as possible.

daffodil-lib/src/test/scala/passera/test/TestULong.scala
mbeckerle: Ah, ok. Pretty easy to just do new JBigInt(2).

daffodil-lib/src/test/scala/passera/test/TestULong.scala
mbeckerle: Yeah, they are just taking the shift count modulo width. I.e., (count & 0x3F) in this case, 1F for Int, F for short, 7 for byte. 

I am inclined to follow their style, so as to not fully fork this library, but be able to propagate our changes here back to the original authors.

stevedlawrence: Oh yeah, forgot about how Java modulo's the shift value. In which case, this is consistent with signed types and is perfectly reasonable.

mbeckerle: I tried the >>, <<, and >>> operators in scala. They have this same behavior of shift by the width of the number is a no-op. So this library is just following that precedent.  

Makes no sense to me, I think these things are still following an ancient precedent set in the B programming language (as in the language before C), back when doing arg checks was considered too expensive, so making any combination of arguments somehow legal was considered somehow beneficial.  And now we're stuck with it!

Of course our DPath expression language is entirely different, so we need not follow this precedent in our shifting functions, but for this passera unsigned library it probably makes sense to stick with the way it works now.

stevedlawrence (approved):
+1

mbeckerle:
The checks system keeps hanging. I'm merging this anyway, as I have tested it locally, and it is a small focused change.

stevedlawrence:
Yeah, GitHub actions is having issues: https://www.githubstatus.com/

PR #579, opened 2021-05-28
Fix unordered sequences with discriminated content

stevedlawrence:
We currently implement unordered sequences as an repetition of a choice
of the sequence branches. The problem with this approach is that the
ChoiceParser handles discriminated content/resolved points of
uncertainty, and by the time the ChoiceParser returns back to the
SequenceParser we no longer know if the ChoiceParser failed because all
alternatives failed or because a branch was discriminated and it failed.
Although they are both failures, the SequenceParser needs to know the
difference so that diagnostics can be returned correctly.

This changes how unordered sequences are handled to not use the
ChoiceParser at all. Instead, we add special casing to the
SequenceParser so that unordered sequence branches are just treated as
required scalar elements. We parse these alternatives within a point of
uncertainty created by the sequence parser, so that it can know whether
a sequence branch was discriminated or not. This allows proper handling
of failures, diagnostics, and backtracking.

DAFFODIL-2512

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
@@ -36,7 +36,7 @@ import org.apache.daffodil.util.Maybe.One
 abstract class SequenceParserBase(
   srd: SequenceRuntimeData,
   childParsers: Vector[Parser],
-  isOrdered: Boolean = true)
+  isOrdered: Boolean)
   extends CombinatorParser(srd) {
   override def nom = "Sequence"
 
@@ -116,11 +116,11 @@ abstract class SequenceParserBase(
       child = children(scpIndex).asInstanceOf[SequenceChildParser]
 
       child match {
-        case parser: RepeatingChildParser => {
+        case parser: RepeatingChildParser if isOrdered => {
           //
-          // The sequence child is an array/repeating element (or ooptional
-          // element as the runtime doesn't distinguish them.)
-          //
+          // The sequence child is an array/repeating element (or optional
+          // element as the runtime doesn't distinguish them) of an ordered
+          // sequence. Unordred sequences are treated as scalars below.
           //
           val min = parser.minRepeats(pstate)
           val max = parser.maxRepeats(pstate)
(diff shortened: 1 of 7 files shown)


tuxji (approved):
+1

Seems fine.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -268,24 +283,36 @@ abstract class SequenceParserBase(
               pstate.diagnostics = diagnosticsBeforeAttempt
             }
 
-            // We successfully parsed a discriminator, but failed to parse the discriminated content.
-            // Do not continue trying to parse other memebers of the unordered sequence.
+            // We successfully parsed a discriminator in an unordered sequence,
```
mbeckerle, line 286: I think you mean We successfully parsed a discriminator within a child of an unordered sequence, yes? 

So the resultOfTry value UnorderedSeqDiscriminatedFailure means that we failed a child, it was a child of an unordered sequence, and that failure occurred after the discriminator (for this child alternative) had evaluated to true.  Do I have that right?

mbeckerle (approved):
+1 Found a couple typos. My questions may suggest tweaks to the comments they are about. In general this is a good cleanup.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -321,20 +344,19 @@ abstract class SequenceParserBase(
...
 
       // We finished parsing one part of a sequence, which could either be an
-      // array, simple or complex. If the last child is different from when we
-      // started that means we must have added a new element and we can set it
-      // final and walk.
+      // array, simple or complex. We aren't sure if we actully added new
+      // element or not, but in case we did, mark the last node as final and
+      // try walk in case we could output the infoset.
+      //
+      // Note that in the case of unordered sequences we are blocked, and so
```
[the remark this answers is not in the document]
stevedlawrence: Yeah, this comment needs to be clarified.

We have an InfosetWalker that periodically walks the internal infoset data structure and outputs infoset events. But it outputs events only once it knows that an infoset element can't be backtracked and undone, since we can't "undo" events once they are created. There are a couple places where we set flags on infoset elements that say they are "final" and that the infoset can walk/output their event.

The problem is that with unordered sequences the individual infoset elements might think they are final and change that flag, but because those elements are in an unordered sequence they might actually get reordered once the the unordered sequence completes. So even though the elements might claim they are final, they actually aren't.

To deal with this, we put a "block" on the infoset walker right before we start parsing an unordered sequence so none of it's children can be walked. Once the unordered sequence is complete and any reordering is done, we "unblock" the walker.

The terminolgy used for this blocking is "infosetWalkerBlockCount" which is essentially just a refcounter that each infoset item has that prevents, or "blocks", the walker from making moving past it. At the beginning of an unordered sequence parse, we increment this block count and at the end we decrement that block count.

This comment is really just saying that we are about to tell the infoset walker to walk, but it's not actually going to do anything because the unordered sequence has been "blocked". I think we can wrap this walk in a `if (!isOrdered)` conditional to make it even more clear. But ill also update the comment to make it more clear what's going on.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
stevedlawrence: Yep, i'll update the comment.

PR #581, opened 2021-05-31
Daffodil-2473 [WIP} XOR Operator

Samarth08:
Tried the XOR Operator.
Both operands should be unsigned or signed.
Automatically it is getting casted to larger data type.


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -98,6 +98,53 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
   }
 }
 
+case class DFDLXXor(recipes: List[CompiledDPath], argType1: NodeInfo.Kind,argType2:NodeInfo.Kind) extends FNTwoArgs(recipes) {
+  override def computeValue(arg1: DataValuePrimitive, arg2: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+
+    argType1 match {
```
stevedlawrence, line 104: With the changes mentioned in previous comments you no longer need to pass in two types since they will always be the same. And so I think this logic all just becomes

```scala
argType match {
  case NodeInfo.Long => arg1.getLong ^ arg2.getLong
  case NodeInfo.Int => arg1.getInt ^ arg2.getInt
  case NodeInfo.Short => (arg1.getShort ^ arg2.getShort).toShort
  ...
  case NodeInfo.UnsignedLong => arg1.getBigInt.xor(arg2.getBigInt)
  case NodeInfo.UnsignedInt => arg1.getLong ^ arg2.getLong
}
```
You don't have to do any nesting of match/cases because DPath will implicitly cast everything to the correct type.

Some things worth pointing out:
1. I don't think you need to do anything fancy for unsigned types like we had to do for the shifting functions. This is because the high order bits are all zeros for unsinged types and so we will always stay zero for the xor. So I think this can just use normal xor'slike you have
1. For types that are reprsented by java primitives smaller than an Int (i.e. short, byte, unsignedByte), we need to make sure to cast the result to that type (e.g. short needs a toShort call). This is because in Java, the result of an xor always raises the type to an integer, but the result of the function needs to be the same as the original type.

stevedlawrence (commented):
Looking good. Just a couple comments to clarify how DPath works, which I think will make some of the logic a bit simplier

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2177,7 +2180,42 @@ case class DFDLXShiftExpr(nameAsParsed: String, fnQName: RefQName,
...
+  override def targetTypeForSubexpression(subexp: Expression): NodeInfo.Kind = {
+    if (subexp == args(0))
+      inherentType
+    else if (subexp == args(1))
+      inherentType
+    else
+    //$COVERAGE-OFF$
+      Assert.invariantFailed("subexpression %s is not an argument.".format(subexp))
+    //$COVERAGE-ON$
+  }
```
mbeckerle, line 2206: I think you can just have
```
override def targetTypeForSubexpression(subexp: Expression): NodeInfo.Kind = inherentType
```
because all the logic to reconcile the two arg types is over in the inherentType lazy val. 

I think the check for the subexp somehow not being one of the arguments - you don't need to check for that. It's an invariant check anyway, and really shouldn't be the responsibility of every operator like this one to check for that. (If that invariant is worth checking, the caller of targetTypeForSubexpression should do that check.)

mbeckerle (changes requested):

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -98,6 +98,26 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+
+    argType match {
+          case NodeInfo.Long => arg1.getLong ^ arg2.getLong
+          case NodeInfo.Int => arg1.getInt ^ arg2.getInt
+          case NodeInfo.Short => (arg1.getShort ^ arg2.getShort).toShort
+          case NodeInfo.Byte => (arg1.getByte ^ arg2.getByte).toByte
+          case NodeInfo.UnsignedLong => arg1.getBigInt.xor(arg2.getBigInt)
+          case NodeInfo.UnsignedInt => arg1.getLong ^ arg2.getLong
+          case NodeInfo.UnsignedByte => arg1.getInt ^ arg2.getInt
+          case NodeInfo.UnsignedShort => (arg1.getByte ^ arg2.getByte).toByte
```
stevedlawrence, line 112: An xs:unsignedByte is represented by a Short and an xs:unsignedShort is represented by an Int. So these two lines should be:

```scala
          case NodeInfo.UnsignedShort => arg1.getInt ^ arg2.getInt
          case NodeInfo.UnsignedByte => (arg1.getShort ^ arg2.getShort).toShort
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -98,6 +98,26 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+      //$COVERAGE-OFF$
+      case _ => Assert.invariantFailed(s"dfdlx:XOR not supported")
+      // $COVERAGE-ON$
+    }
```
stevedlawrence, line 116: Indentation is off here.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2177,7 +2180,42 @@ case class DFDLXShiftExpr(nameAsParsed: String, fnQName: RefQName,
...
+                        args: List[Expression], constructor: (List[CompiledDPath], NodeInfo.Kind,NodeInfo.Kind) => RecipeOp)
+  extends FunctionCallBase(nameAsParsed, fnQName, args) {
+  override lazy val inherentType = {
+    val argInherentType = args(0).inherentType
+    val arg2=args(1).inherentType
+    schemaDefinitionUnless((
+      argInherentType.isSubtypeOf(NodeInfo.PrimType.UnsignedLong) && arg2.isSubtypeOf(NodeInfo.PrimType.UnsignedLong))||  (argInherentType.isSubtypeOf(NodeInfo.PrimType.Long) && arg2.isSubtypeOf(NodeInfo.PrimType.Long)),
+      "First argument and Second argument both should  be either xs:unsignedLong  or a subtype of those or xs:long or a subtype of those",
+      nameAsParsed,argInherentType.globalQName)
+    argInherentType
```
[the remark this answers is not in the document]
stevedlawrence: Looks good, but we need to bring back the check that makes usre that both ``arg0Type`` and `arg1Type` are xs:unsignedLong, xs:long or a subtype, and that the arguments do not mix signed/unsigned types.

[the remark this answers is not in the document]
stevedlawrence: Looks good, but we need to bring back the check that makes sure that both ``arg0Type`` and `arg1Type` are xs:unsignedLong, xs:long or a subtype, and that the arguments do not mix signed/unsigned types.

stevedlawrence (commented):
Couple small changes and I think this should be good to start writing some tests.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2177,7 +2180,27 @@ case class DFDLXShiftExpr(nameAsParsed: String, fnQName: RefQName,
...
+  override lazy val inherentType = {
+    val arg0Type = args(0).inherentType
+    val arg1Type = args(1).inherentType
+    val argInherentType = if (arg1Type.isSubtypeOf(arg0Type)) arg0Type else arg1Type
+    argInherentType
+  }
+    override def targetTypeForSubexpression(subexp: Expression): NodeInfo.Kind = inherentType
 
+  override def compiledDPath: CompiledDPath = {
+    checkArgCount(2)
```
mbeckerle, line 2195: This and the next 2 codecov warnings make it clear you have no test (as of when codecov ran) that exercises this function, because this code must be exercised by any such test.

Samarth08:
Let me know if I can start with the test cases.

Samarth08:
I am passing the below mentioned type to XOR and getting an error-

Caused by: Schema Definition Error: Unsupported function: dfdlx:DFDLXXor.  

<xs:complexType name="IntXor">
    <xs:sequence dfdl:separator=",">
      <xs:element name="num" type="xs:int" dfdl:lengthKind="delimited" dfdl:textNumberPattern="#0"/>
      <xs:element name="count" type="xs:int" dfdl:lengthKind="delimited"/>
      <xs:element name="value" type="xs:int" dfdl:inputValueCalc="{dfdlx:DFDLXXor(../num,../count)}"/>
    </xs:sequence>
  </xs:complexType>

stevedlawrence:
`DFDLXXor` is the name of the internal class that implements the functionality, which isn't necessarily the same as the name that goes in the schema.

You added the following in Expression.scala which defines the function name to use in a DFDL schema:

```scala
      case (RefQName(_, "xor", DFDLX), args) =>
        DFDLXBitExpr(functionQNameString, functionQName, args,
          DFDLXXor(_,_))
```
So the name of this function is `xor` in the `DFDLX` namespace. So your test schema should just have

```scala
dfdlx:xor(../num,../count)
```

mbeckerle (approved):
+1 Looks good.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/BitFunctionsXor.tdml
```diff
@@ -0,0 +1,331 @@
...
+          </xs:complexType>
+
+            <xs:element name="testByteXor">
+              <xs:complexType>
+                <xs:sequence dfdl:separator="|">
+                  <xs:element name="ByteXor1" type="ex:ByteXor"/>
+                  <xs:element name="ByteXor2" type="ex:ByteXor"/>
+                </xs:sequence>
+              </xs:complexType>
+            </xs:element>
```
stevedlawrence, line 99: Indentation is off is this test file. Some things are indented too far.

stevedlawrence (approved):
+1

Please fix the indentation in the test file, and then squash and this should be good to merge.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/BitFunctions.tdml
```diff
@@ -353,26 +353,26 @@
         <tdml:infoset>
```
mbeckerle, line 353: Our indentation standard is 2-spaces, not 4 as you have here. 

Could you re-indent the XML files as 2-space indentation (generally an IDE setting change and then resave the files)?

With that, you can then squash all these commits together into one commit, and force push that. We will then merge it. 

Thanks.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
mbeckerle: This is indenting 5 spaces. Our convention is 2 spaces.

mbeckerle (commented):
The only thing left is a couple indentation fixes. 

Thanks.

Samarth08:
@mbeckerle Fixed the indentation.
Thank You.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -98,6 +98,26 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
   }
 }
 
+case class DFDLXXor(recipes: List[CompiledDPath], argType: NodeInfo.Kind) extends FNTwoArgs(recipes) {
+  override def computeValue(arg1: DataValuePrimitive, arg2: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+
+    argType match {
+          case NodeInfo.Long => arg1.getLong ^ arg2.getLong
```
mbeckerle, line 105: I think you missed reindenting this scala code.

Samarth08:
I have fixed the indentation, and have squashed the commits.
I can start with 'AND' and 'OR' operator together after this.

PR #583, opened 2021-06-02
Fixes bug introduced by earlier XML validation changes.

mbeckerle:
This bug broke the ibm-dfdl-crosstester.

DAFFODIL-2530

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -30,12 +30,10 @@ import java.nio.charset.CoderResult
 import java.nio.charset.{Charset => JavaCharset}
 import scala.language.postfixOps
 import scala.util.Try
-import scala.xml.Elem
 import scala.xml.Node
 import scala.xml.NodeSeq
 import scala.xml.NodeSeq.seqToNodeSeq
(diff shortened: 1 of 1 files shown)


stevedlawrence (approved):
+1, looks good to me.

But I'm wondering if there's maybe a reason to do things slightly differently? I'm imagining a future where we can trigger multiple implementations to be run at the same time. In that case, this `optEmbededSchema` is going to either have line numbers or not depending on which implementation is triggered first since it's a lazy val. It looks like all other uses of `isCrossTest` don't have this issue since they are all called in a def at when a test is run. So, I'm wondering if maybe `optEmbededSchema` should just always return `defSchema.schemaSource` with the line/col numbers, and if the IBM DFDL cross tester (or any other implementation) doesn't like those additional attributes, then the cross tester should be responsible for stripping them off? Maybe that's too much effort for something that isn't even an issue right now though, and just something to consider if we ever get to something like that.

mbeckerle:
This stripping of line numbers/file attributes .. moving that over into the processor for IBM DFDL makes sense, because it is its inability to tolerate non-native attributes that is at issue. That would be an improvement. 

Since I am changing both Daffodil, and the cross-tester to match and work together right now, it's a good time to make that change.

tuxji (approved):
+1

In response to Steve's thoughts, you could add a comment before the changed line explaining that you needed to pass a schema without line numbers here.

PR #584, opened 2021-06-03
Adds diagnostic capability to TDMLTestNotCompatibleException.

mbeckerle:
DAFFODIL-2531


tuxji (approved):
+1

Adding the cause will aid troubleshooting.

stevedlawrence (approved):
+1

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -748,7 +748,23 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite)
...
+      // The CLI catches this, but won't display the messaging that
+      // elaborates on the cause, because the CLI doesn't utilize junit,
+      // so it expects this to fail.
+      //
+      // But if this isn't caught, and propagates to top level,
+      // then we'll get the diagnostic messaging, including about the
+      // cause.
+      //
+      new TDMLTestNotCompatibleException(testName, implementationName,
+        Option(junitExceptionInstance.failed.get))
```
stevedlawrence, line 766: Is it worth adding a CLI test to cover this line? It's not possible to hit using sbt tests since they have junit on the classpath.

mbeckerle: So a CLI test has to attempt to run a test that has implementation="ibm" let's say, in which case it will fail and will throw this, and the CLI will catch it. I suppose so.

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/testNonCompatibleImplementation.tdml
```diff
@@ -16,10 +16,13 @@
   limitations under the License.
 -->
 
-<tdml:testSuite suiteName="Info" description="CLI"
```
mbeckerle, line 19: The file infoTest.tdml was unused. I copied and edited it, and I guess it was small so we got a diff here instead of a new file and a deleted file.

PR #586, opened 2021-06-04
Improve setting of external variables

stevedlawrence:
- Fix an exception when a schema contains at least one variable are
  marked as external="false" and you try to set any variable (external
  or not) without providing a namespace for the variable. This is very
  common with the CLI or the withExternalVariables API that accepts a
  Map.
- Modify setExternalVariables so that it allows setting the same
  variable multiple times externally. This allows changing the same
  variable in between different calls to parse/unparse.
- When you try to set an external variable and either that variable is
  not externally settable or the variable is not defined, then we
  currently throw an SDE. But SDE's really shouldn't be thrown. We don't
  want to add this SDE to the DataProcessors diagnostics because then
  isError would be true and the DataProcessor could no longer be used.
  Instead, this adds a new ExternalVariableException that is thrown. The
  Java and Scala API's are modified to use this exception. Also modifies
  the CLI to check for this new exception and error.
- Remove VariableUtils.convert function. This is only used in places
  where a conversion isn't needed because the variable is already
  converted from DPath, or we need special logic if the conversion fails
  to create the rite exception.
- Remove VariableMapFactory.setExternalVariables--this function is not
  used anywhere.
- Modify the TDML Runner to no longer catch external variable errors and
  make them part of diagnostics. Instead the TDML runner just tosses the
  error as a TDMLException. New API/CLI tests are added to test the
  various ways that setting external variables can fail and make sure we
  get the correct API exception.

DAFFODIL-2480

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
@@ -17,6 +17,9 @@
 
 package org.apache.daffodil.processors
 
+import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.Map
+
 import org.apache.daffodil.api.ThinDiagnostic
 import org.apache.daffodil.dpath.InvalidPrimitiveDataException
 import org.apache.daffodil.exceptions.Assert
@@ -26,14 +29,15 @@ import org.apache.daffodil.infoset.DataValue
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitiveNullable
 import org.apache.daffodil.infoset.RetryableException
-import org.apache.daffodil.util.Maybe
-import org.apache.daffodil.util.Maybe.Nope
-import org.apache.daffodil.xml.{GlobalQName, UnspecifiedNamespace, NamedQName, RefQName}
 import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.schema.annotation.props.gen.VariableDirection
-
-import scala.collection.mutable.{Map, ArrayBuffer}
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.Maybe.Nope
+import org.apache.daffodil.xml.GlobalQName
+import org.apache.daffodil.xml.NamedQName
+import org.apache.daffodil.xml.RefQName
+import org.apache.daffodil.xml.UnspecifiedNamespace
 
 sealed abstract class VariableState extends Serializable
 
@@ -160,15 +164,6 @@ object VariableUtils {
 
   def setExternalVariables(currentVMap: VariableMap, bindings: Seq[Binding], referringContext: ThrowsSDE) = {
     bindings.foreach { b => currentVMap.setExtVariable(b.varQName, b.varValue, referringContext) }
-  }
-
-  def convert(v: String, rd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive = {
-    try {
-      rd.primType.fromXMLString(v)
-    } catch {
-      case e: InvalidPrimitiveDataException =>
-        referringContext.SDE("Error processing variable %s: %s", rd.globalQName, e.getMessage)
-    }
   }
 
 }
@@ -412,7 +407,7 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
            * variable read too early */
           pstate.SDW("Cannot set variable %s after reading the default value. State was: %s. Existing value: %s",
           variable.rd.globalQName, VariableSet, variable.value)
-          variable.setValue(VariableUtils.convert(newValue.getAnyRef.toString, variable.rd, referringContext))
+          variable.setValue(newValue)
           variable.setState(VariableSet)
         }
 
@@ -432,7 +427,7 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
             }
             case _ => // Do nothing
           }
-          variable.setValue(VariableUtils.convert(newValue.getAnyRef.toString, variable.rd, referringContext))
+          variable.setValue(newValue)
           variable.setState(VariableSet)
         }
       }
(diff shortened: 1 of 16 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -459,66 +454,75 @@ class VariableMap private(vTable: Map[GlobalQName, ArrayBuffer[VariableInstance]
...
+            val value =
+              try {
+                variable.rd.primType.fromXMLString(newValue)
+              } catch {
+                case e: InvalidPrimitiveDataException => {
+                  val msg = "Value for variable %s is not a valid %s: %s".format(
+                    variable.rd.globalQName,
+                    variable.rd.primType.globalQName,
+                    newValue)
+                  throw new ExternalVariableException(msg)
```
mbeckerle, line 513: Converting one exception to another like this always suggests a style to me which is to just chain the exceptions. I.e., make a constructor for ExternalVariableException which takes a cause which is another exception. 

There's nothing wrong with this code as you have it. I would have made ExternalVariableException a base class and created 3 or 4 more specific subclasses for each of the cases where you are throwing them. Then the string assembly all ends up in the getMessage of the specific exception classes. The constructors of the specific exception classes would just be passed the relevant objects needed to render a good message. 

This is a style that lends itself to internationalization better than if the main code assembles strings of messages.

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
```diff
@@ -1174,4 +1178,100 @@ class TestScalaAPI {
...
+    var dp = pf.onPath("/")
+    dp = reserializeDataProcessor(dp)
+
+    // set var without a namespace, ambiguity error because schema contains
+    // two variables with same name but different namespace
+    try {
+      dp = dp.withExternalVariables(Map("var" -> "10"))
+    } catch {
+      case e: ExternalVariableException => {
+        assertTrue(e.getMessage().contains("var"))
```
mbeckerle, line 1197: Suggest insert `val msg = e.getMessage()`  just to make code cleaner if this code is copied as a pattern.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1014,36 +1009,30 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
-      }
-    if (optExtVarDiag.isDefined) {
-      toss(optExtVarDiag.get, implString)
-    } else {
-      val actual = processor.parse(new ByteArrayInputStream(testData), testDataLength)
-      val diagObjs = actual.getDiagnostics
+    try {
+      processor = processor.withExternalDFDLVariables(externalVarBindings)
+    } catch {
+      case e: Exception => toss(e, implString)
```
mbeckerle, line 1015: Is this toss just for a centralized breakpoint/print location?
If that's the case, why not catch Throwable?
If not, then this should be narrower set of exception classes.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1014,36 +1009,30 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
-        val hasErrorDiags = diagObjs.exists { diag =>
-          diag.isError && !diag.isValidation
-        }
-        Assert.invariant(!hasErrorDiags)
+    val actual = processor.parse(new ByteArrayInputStream(testData), testDataLength)
+    val diagObjs = actual.getDiagnostics
+
+    if (actual.isProcessingError) {
+      // Means there was an error, not just warnings.
+      if (diagObjs.length == 1) throw TDMLException(diagObjs.head, implString)
```
mbeckerle, line 1023: why toss above and throw here?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1489,18 +1478,15 @@ case class UnparserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
-        processor = processor.withExternalDFDLVariables(externalVarBindings)
-        None
-      } catch {
-        case diag: Diagnostic =>
-          Some(diag)
-      }
+    try {
+      processor = processor.withExternalDFDLVariables(externalVarBindings)
+    } catch {
+      case e: Exception => toss(e, implString)
```
mbeckerle, line 1484: same question about toss.

mbeckerle (approved):
+1 Just a few questions and other not-required suggestions

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
stevedlawrence: I started with the approach of having multiple exceptions, but felt it was a bit tedious because the way are japi/sapi would also need to have all those exceptions and have converters from the "internal" exceptions to the "public" API exceptions. Which is also why I just went with not chaining the exception here--I didn't want to have to chain it in the public API. Maybe it's worth it though?

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
stevedlawrence: Will do.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: `withExternalDFDLVairables` should only throw an ExternalVariableException, so the toss just converts that exception into a TDMLRunnerException to make things a little nicer looking. This should probably catch `ExternalVariableException` instead of `Exception`

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: This was existing code. I think I copied the toss from somewhere else. Not sure if there's a reason to use toss vs explicitly creating a TDMLException and throwing?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
mbeckerle: Yeah, the redundant layer thing.

We have to rethink the JAPI/SAPI layering, and maybe be just using the API objects internally rather than building this redundant layer around them all. For any exceptions that have to show up on SAPI, why wouldn't we just use those in the runtime1 instead of having runtime1 define them, and having the API have to have encapsulators?

One possibililty is that Daffodil uses the set of 3 or 4 derived exceptions, and that hides the string hacking of messages, and then JAPI/SAPI just have a single ExternalVariableException that encapsulates the internal one as a delegate and calls its getMessage().

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: The usual reason for toss is just for a breakpoint. So you can get the breakpoint before the context changes to the catcher. It's a "coding for debug" style.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
stevedlawrence: Yeah, would be nice to solve this JAPI/SAPI separation. It's pretty terrible.

One reason not to return the internal things is so that we can change them if we want without affecting backwards compatibility. I'm not sure that's a great reason.

Another reason for the separation is that we don't publish Java/Scala docs for internal API. So there would be no documentation about these new exceptions.

I do like having a concept of a stable public API being everything in org.apache.daffodil.japi.*, which makes it easier to figure out what's public and what's internal, and thus stable vs potential not stable.

> One possibililty is that Daffodil uses the set of 3 or 4 derived exceptions, and that hides the string hacking of messages, and then JAPI/SAPI just have a single ExternalVariableException that encapsulates the internal one as a delegate and calls its getMessage().

Feels like this might just add extra maintenance burden. I think the only thing that ever needs to catch these exceptions is the API shim layer, so if it's going to just convert 4 different exceptions to one, do we gain much from having 4 separate exceptions. I guess it's more self documenting, and the error messages are in the exception with a few parameters passed in, rather than passing in a fully formed string?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: Okay, I'll switch it to a TDMLException then.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: Actually, the ExternalVariableException is defined in daffodil-runtime1. Which the TDMLRUnner doesn't have access to. In fact, it looks like all of the TDMLRunner only catches Exception or Throwable, probably for this reason. So I think Exception here is probably the right thing, with the understanding that the only Exception this should throw is the variable ones. And even if it's not, we're still going to throw a TDMLException and the test will fail, so doesn't really change the behavior.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
stevedlawrence: I've pushed an update that I think addresses all comments except this one? Any thoughts one what route you prefer? I don't feel strongly one way or the other.

mbeckerle: I suggest merge this as is, but subsequently put some thought into the JAPI/SAPI rethink and whether this could be modularized in some better way. If we didn't do this "the right way" due to JAPI/SAPI maintenance overheads, then we really do need to rethink those. 

I am expecting we need to rethink the Daffodil APIs anyway given the whole lack-of modularization of back-ends.

tuxji (approved):
+1

PR #587, opened 2021-06-11
Fix unparsing large xs:hexBinary

stevedlawrence:
In some places, we use an Int for variables referring to the length in
bits of xs:hexBinary data. This means we can only unparse MaxInt bits of
hexBinary data, which is only 256MB. This modifies hex binary unparsing
to use Long's in these cases to ensure we can support large hex binary
blobs. The maxLenghHexBinaryInBits tunable is also checked during
unparsing, which ensures we won't try to unparse more than 2GB of
hexBinary data, which is a limitation of Java byte arrays.

DAFFODIL-2532


tuxji (approved):
+1

I see the checks are passing and codecov is 100%, so looks good.

mbeckerle (approved):
+1

PR #592, opened 2021-06-17
Added different exit codes to Daffodil command line

dasmallwood:


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -685,8 +685,8 @@ object Main extends Logging {
   }
 
   def createProcessorFromSchema(schema: URI, rootNS: Option[RefQName], path: Option[String],
-    tunables: Map[String, String],
-    mode: ValidationMode.Type): Option[DFDL.DataProcessor] = {
+                                tunables: Map[String, String],
```
mbeckerle, line 688: Your IDE made a bunch of whitespace changes I think. 
They make your change set look much more complex than it should be.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1584,35 +1584,35 @@ object Main extends Logging {
...
+        3
       }
       case e: BindingException => {
-        log(LogLevel.Error, "%s", e.getMessage())
-        1
+        log(LogLevel.Error, "Binding Error", e.getMessage())
+        4
       }
       case e: NotYetImplementedException => {
         nyiFound(e)
```
mbeckerle, line 1603: nyiFound case also needs to return a unique return code.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1584,35 +1584,35 @@ object Main extends Logging {
...
       }
       case e: UserDefinedFunctionFatalErrorException => {
-        log(LogLevel.Error, "%s", e.getMessage())
+        log(LogLevel.Error, "User Error", e.getMessage())
         e.printStackTrace()
-        1
+        6
       }
       case e: Exception => {
         bugFound(e)
```
mbeckerle, line 1618: bugFound case also needs to return a unique return code.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1584,35 +1584,35 @@ object Main extends Logging {
     } catch {
       case s: scala.util.control.ControlThrowable => throw s
       case e: java.io.FileNotFoundException => {
-        log(LogLevel.Error, "%s", e.getMessage())
+        log(LogLevel.Error, "Missing File Error", e.getMessage())
```
mbeckerle, line 1587: You need tests that elicit these messages. I think they would show you that you need "Missing File Error: %s" otherwise there is no place for the e.getMessage() to get substituted into.

But I also think you should compute the message string from the exception object class name. 

So like this:
```
log(LogLevel.Error, "%s %s", Misc.getNameFromClass(e), e.getMessage())
```
The Misc.getNameFromClass(e) is a utility we created for exactly this sort of purpose. It gives you the class name without the package part.

mbeckerle (changes requested):
Some suggested improvements.

Whitespace spurious changes need to be reverted.

Needs tests. I imagine test coverage analysis (codecov) which runs automatically on a pull request, is going to show that most of your changes are not covered by tests.

I am not entirely sure how our CLI testing framework, based on the 'net.sf.expectit' library, can look at the return code of the process. There might be some way built in to that library, or maybe you just have to do a platform-specific thing. E.g., on unix when running a shell, the magic shell variable "$?" contains the return code of the previous command. But that's not portable to microsoft windows. (There's already another JIRA ticket about the state of the CLI tests, which are full of "if windows" conditionalizations that should be hidden, not repeated test after test. )

mbeckerle:
I also see that your branch is dasmallwood:master. Our standard practice is that you create a branch named after the JIRA ticket such as daf-2335-rc. 

This is useful practice because given review cycles, etc. most of us end up working on more than one ticket simultaneously. It also allows other people to pull down your changes without stepping on the master branch of their local sandbox git repository, which is something we do frequently as part of review, or just to help out.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: I imagine in this case and the ones below, the message from the exception is what we want to display. This new strings dont' provide enough information. For example, the message I get when we hit this FileNotFoundException is 

> [error] notafile.txt (No such file or directory)

Which is maybe a bit terse, but I think provides all the necessary information. We probably want CLI tests to make sure we are getting reasonable error messages.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: Note that the nyiFound function returns an int, which is the return code. So either that function needs to return a different code, or it shouldn't return a code at all, and this should return a number.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1584,35 +1584,35 @@ object Main extends Logging {
...
       }
       case e: OutOfMemoryError => {
         oomError(e)
       }
       case e: UserDefinedFunctionFatalErrorException => {
-        log(LogLevel.Error, "%s", e.getMessage())
+        log(LogLevel.Error, "User Error", e.getMessage())
         e.printStackTrace()
-        1
+        6
```
stevedlawrence, line 1615: Instead of having all these magic numbers spread out, I would suggest we create some vals so that these are self documenting. e.g. val USER_DEFINED_FUNCTION_ERROR_CODE. That way all the codes can be define in one spot and used throughout.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: A similar change is needed as nyiFound.

stevedlawrence (changes requested):

stevedlawrence:
Another thing to consider is that we return error codes in non-Exception cases as well. For example, we have these lines in the parser section of Main.scala:
```scala
          case Some(proc) if (!proc.isError) => {
            ...
            if (error) 1 else 0
          }
          case Some(processor) => 1
          case None => 1
```
There are three different places where we return a 1, which potentially means three different error codes. I think the other two cases represent different failures to compile a schema, though I'm not sure specifically what differentiates them.  I"m also not even sure if they both can be hit. I think if there is a compilation issue, then the CLI outputs diagnostics and then returns None. So I think this wants to be something like:

```scala
          case Some(proc) => {
            Assert.invariant(!proc.isError)
            ...
            if (error) 1 else 0
          }
          case None => 1
```
But the point is, these 1's probably want to have separate error codes as well. Something like COMPILE_SCHEMA_ERROR_CODE, and one like PARSE_ERROR_CODE.

Anytime there is a ``1`` in the run function, that's probably an error code that needs to be changed.

stevedlawrence:
In regards to Mike's comment about checking the return code, since integration tests are just run in a shell with expect, you need to echo the return code and then expect the right code. For example, in some tests we have something like this:

```scala
      val exitCodeCmd = if (Util.isWindows) "echo %errorlevel%" else "echo $?"
      shell.sendLine(exitCodeCmd)

      if (Util.isWindows) shell.expect(contains(exitCodeCmd + "\n"))
      else shell.expect(contains("\n"))

      val sExitCode = shell.expect(contains("\n")).getBefore()
      val exitCode = Integer.parseInt(sExitCode.trim())
      if (exitCode != 0)
        fail("Tests failed. Exit code: " + exitCode)
```
Might make sense to change this to some utility function so we don't have to copy/paste this every time we want to check the return code.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1584,35 +1584,35 @@ object Main extends Logging {
...
       case e: java.io.FileNotFoundException => {
-        log(LogLevel.Error, "%s", e.getMessage())
+        log(LogLevel.Error, "Missing File Error", e.getMessage())
         1
       }
       case e: InvalidParserException => {
-        log(LogLevel.Error, "%s", e.getMessage())
-        1
+        log(LogLevel.Error, "Invalid Parser Error", e.getMessage())
+        2
```
mbeckerle, line 1592: I would also like to see all the error codes defined as symbolic constants in one place. Eg., 

def INVALID_PARSER_ERROR = 2

with all of them in a block together so you can see them all.

 (btw. def vs. val we use def for constants.)

mbeckerle (changes requested):
Suggest use standard scala Enumeration pattern, and also create utility function to encapsulate test complexity associated with portable exit-code checking on unix/windows.

dasmallwood:
@stevedlawrence superseded by
https://github.com/apache/daffodil/pull/601

PR #593, opened 2021-06-22
Detect and error when duplicate tdml:defineConfig's exist

stevedlawrence:
Also refactors similar detection of duplicate test case names and
embedded schemas to use the same function.

This found a single file, tunables.tdml, which had duplicate
configurations. This revealed that the tests for these particular
tunables did not cover all the intended cases, finding elements that
weren't even used in tests. Additional tests are added to cover all
different values of unqualifiedPathStepPolicy and ensure the tests
included in the file have the expected behavior.

DAFFODIL-2533


tuxji (approved):
+1

Looks fine to me.

mbeckerle (approved):
+1

PR #594, opened 2021-06-22
Add dependabot.yml to call Dependabot

tuxji:
Use Dependabot to keep our Dockerfile and Action workflows up to date.

DAFFODIL-2528

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,30 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "docker" # See documentation for possible values
+    directory: "/containers/release-candidate/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
stevedlawrence, line 1: This needs an ASF header to pass the RAT check

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "docker" # See documentation for possible values
+    directory: "/containers/release-candidate/" # Location of package manifests
```
stevedlawrence, line 9: What will this actually check in the release candidate container? Does it build/run it, or just look in the Dockerfile for and container versions that can be updated? For example, right now we use Fedora 34, will it just add a PR when Fedora 35 containers become available?

stevedlawrence (approved):
+1

.github/dependabot.yml
tuxji: Dependabot will just look in the Dockerfile for whatever it can update, e.g., it will submit a PR when a Fedora 35 container supersedes the Fedora 34 container.

tuxji:
One check (Java 11, windows 2019) had a test failure unrelated to this PR:

```log
[info] Test org.apache.daffodil.example.TestScalaAPI.testScalaAPI2 started
testScalaAPI2: [SAPI LOG] [compile] Unparser = <Element name='e1'><DelimiterStack initiator='Nope' separator='Nope' terminator='One((dafint:TerminatorUnparseEv@9749aaa, constant: [Lorg.apache.daffodil.processors.dfa.DFADelimiter;@6fb3f672))'><seq><DelimiterStack initiator='Nope' separator='One((dafint:SeparatorUnparseEv@52b2f7ce, constant: [Lorg.apache.daffodil.processors.dfa.DFADelimiter;@2cc6e99f))' terminator='Nope'><Sequence><ScalarOrderedSeparatedSequenceChildUnparser><Element name='elementGroup'><DelimiterStack initiator='Nope' separator='Nope' terminator='Nope'><DelimiterStack initiator='Nope' separator='One((dafint:SeparatorUnparseEv@3e4eed54, constant: [Lorg.apache.daffodil.processors.dfa.DFADelimiter;@6f878529))' terminator='Nope'><Sequence><ScalarOrderedSeparatedSequenceChildUnparser><Element name='e2'><DelimiterStack initiator='Nope' separator='Nope' terminator='Nope'><seq><CaptureStartOfContentLengthUnparser/><CaptureStartOfValueLengthUnparser/><ConvertTextCombinatorUnparser><ConvertTextNumberUnparser/><SpecifiedLengthExplicitImplicitUnparser><StringNoTruncateUnparser/></SpecifiedLengthExplicitImplicitUnparser></ConvertTextCombinatorUnparser><CaptureEndOfValueLengthUnparser/><RightFillUnparser/><CaptureEndOfContentLengthUnparser/></seq></DelimiterStack></Element><SeparatorUnparser /></ScalarOrderedSeparatedSequenceChildUnparser><ScalarOrderedSeparatedSequenceChildUnparser><Element name='e3'><DelimiterStack initiator='Nope' separator='Nope' terminator='Nope'><seq><CaptureStartOfContentLengthUnparser/><CaptureStartOfValueLengthUnparser/><SpecifiedLengthExplicitImplicitUnparser><StringNoTruncateUnparser/></SpecifiedLengthExplicitImplicitUnparser><CaptureEndOfValueLengthUnparser/><RightFillUnparser/><CaptureEndOfContentLengthUnparser/></seq></DelimiterStack></Element><SeparatorUnparser /></ScalarOrderedSeparatedSequenceChildUnparser><SeparatorUnparser /></Sequence></DelimiterStack></DelimiterStack></Element><SeparatorUnparser /></ScalarOrderedSeparatedSequenceChildUnparser><SeparatorUnparser /></Sequence></DelimiterStack><TerminatorUnparser /></seq></DelimiterStack></Element>.[SAPI LOG] [compile] 
[error] Test org.apache.daffodil.example.TestScalaAPI.testScalaAPI2 failed: expected:<0> but was:<1>, took 0.274 sec
[error] Failed: Total 31, Failed 1, Errors 0, Passed 30
[error] Failed tests:
[error] 	org.apache.daffodil.example.TestScalaAPI
```

That test was supposed to result in zero errors, warnings, and other log messages but it got a spurious schema compiler log message for some reason.  In any case, that test's failure is unrelated to this PR.

.github/dependabot.yml
mbeckerle: If this just changes the numbers, but this is only run when we do a release, then I would suggest we just let these PRs pile up, and accept them before a release, when they will then get tested as part of creating an RC1.

stevedlawrence: I'd actually rather we get to this kind of change well before we are ready to do an RC to avoid any issues.

For example, when we updated to Fedora 34 recently there were a couple of issues that took a bit of work to track down and fix (e.g. dependency changes, missing container privileges). If we wait until we're ready for an RC, there's a good chance new issues will crop up and unnecessarily delay the release. Getting to it earlier gives us more time to get those issues fixed.

mbeckerle: I'm fine with that but accepting a PR of this kind of change implies that the person integrating it has tested that it in fact works. With the regular scalabot automated tests, our CI automatically will catch things that it breaks. So if the licenses are all ok still, once can pretty much accept them without fear. For this, nothing is going to automatically check anything. So you have to be willing to spend the time on creating a mini-release which creates the build container etc. 

That is unless we want to do automated CI for that too.

stevedlawrence: Agreed, these changes definitely shouldn't be accepted without manual testing.

It might be interesting to add the release candidate script to CI. GitHub Actions does have a way to build Docker images to use in actions, though from a quick skim it seems like you need to publish the docker image, and then it can be used for an action. Need to look more into if it's possible to build a docker image and use it in the same CI run--we don't really want to publish this image.

Some poential issues with CI:
* Currently the script tries to publish everything which we don't want to do for a PR. But we can just use the --dry-run option to disable that, so not a big deal.
* The script expects user input. Either need a way to pass everything in as options, or maybe can we just write to stdin
* The script builds from apache/master. It can be configured to build from a different repo/branch, so we'd probably need to configure it to build from the PR branch. I think that information is available from github actions so shouldn't be hard.
* GPG signing requires keys--I guess we could generate keys for testing
* The script clones repositories with git+ssh, which requires keys that github knows about. Need a way to switch to http

Probably all doable, but requires a bit of work. For something that doesn't update very often, it might be best to just stick to manual testing when it's needed.

tuxji: Mike, I can add a `reviewers: [ mbeckerle, stevedlawrence, ... ]` to the docker package manager configuration, which should warn a person not to accept a PR for the release dockerfile unless it has passed manual testing by one of the people who actually have released Daffodil in the past.  Will that change get a +1 from you?

mbeckerle: I will approve this. It's effectively still an experiment. Ultimately my concerns are downstream someday, about how this affects our release workflow, etc.

mbeckerle (approved):
+1

PR #600, opened 2021-06-30
Remove by-name and transient params for runtime data objects

mbeckerle:
All by-name passing of args to runtime structures is gone.

A few TransientParam are left, and they have been converted to use the
Delay type instead of using by-name passing. They are all ones that
actually involve circular relationships, so they cannot be removed
without going to non-functional graph construction.

This is achieved by a Delay object type which is enhanced to avoid
retaining any closures or other objects the Scala compiler creates to
implement by-name/lazy evaluation.

DAFFODIL-2326, DAFFODIL-1879


daffodil-test/src/test/scala/org/apache/daffodil/extensions/TestInputTypeValueCalc.scala
```diff
@@ -72,31 +72,33 @@ class TestInputTypeValueCalc {
   @Test def test_inputTypeCalcString_01(): Unit = { fnRunner.runOneTest("inputTypeCalcString_01") }
   @Test def test_outputTypeCalcInt_01(): Unit = { fnRunner.runOneTest("outputTypeCalcInt_01") }
   @Test def test_outputTypeCalcString_01(): Unit = { fnRunner.runOneTest("outputTypeCalcString_01") }
-  @Test def test_repTypeValueInt_01(): Unit = { fnRunner.runOneTest("repTypeValueInt_01") }
-  @Test def test_logicalTypeValueInt_01(): Unit = { fnRunner.runOneTest("logicalTypeValueInt_01") }
-  @Test def test_repTypeValueString_01(): Unit = { fnRunner.runOneTest("repTypeValueString_01") }
-  @Test def test_logicalTypeValueString_01(): Unit = { fnRunner.runOneTest("logicalTypeValueString_01") }
-  @Test def test_outputTypeCalcNextSiblingInt_01(): Unit = { fnRunner.runOneTest("outputTypeCalcNextSiblingInt_01") }
-  @Test def test_outputTypeCalcNextSiblingString_01(): Unit = { fnRunner.runOneTest("outputTypeCalcNextSiblingString_01") }
+//  @Test def test_repTypeValueInt_01(): Unit = { fnRunner.runOneTest("repTypeValueInt_01") }
```
mbeckerle, line 75: remove

mbeckerle:
@stevedlawrence or anyone really. 
The tickets like DAFFODIL-1879 and DAFFODIL-2376 are hard to verify that they are fixed. Suggestions on how?

stevedlawrence:
This is my process for things like this in JProfiler. Other profilers might be different but the process is probably similar.

What I first to is add an object with some functions that I can insert to act as my trigger point. For example:

```scala
object Trigger() {
  def start(): Unit = println("Start trigger")
  def stop(): Unit = println("Start trigger")
}
```

Then right after a DataProcessor is created and where all compile time references should no longer be accessable, I would add a call to this function, e.g.

```scala
Trigger.start()
```

Then I would run this in a JProfiler configured with a "Method Invocation" Trigger on the `Trigger.start()`. The Action for the trigger would be "trigger heap dump", so we get a heap dump when this Trigger.start method is called. In JProfiler, this action can be configured to perform a full garbage collection. When that checked, in theory compile time objects should have all been garbage collected and won't appear in the heap dump. If they do, you should be able to examine the heap dump and find where those references still exist.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -862,88 +786,68 @@ sealed abstract class ModelGroupRuntimeData(
...
-  lazy val groupMembers = groupMembersArg
-
-  override def preSerialization: Unit = {
-    super.preSerialization
-    variableMap
-    encInfo
-    ci
-    groupMembers
-  }
   @throws(classOf[java.io.IOException])
```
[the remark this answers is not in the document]
stevedlawrence: I'm not really sure. I know Java uses reflection to find/call the writeObject method, and I think it explicitly looks a private writeObject method. Since it's private, I guess it can't really be inherited so maybe the reflection can't find it? Or maybe when Java does the reflection is just simply ignored any inherited functions and only looks at those explicitly defined on the class?

Or maybe this is all wrong and could work? I'm not sure.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
```diff
@@ -28,17 +28,45 @@ package org.apache.daffodil.util
...
+  // eliminates holding onto a bunch of objects due
+  // to closures for the pass-by-name arguments.
+  //
+  lazy val value: T = {
+    val v = box.value
+    box = null // throws away box, which allows GC of all closures, etc.
+    v
+  }
+
+  def hasValue = box ne null
```
stevedlawrence, line 48: Surprised nothing actually calls hasValue? It's pretty clearly correct though. Is it worth adding a unit test to ensure this works as expected and doesn't ever have a regression?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -61,16 +63,26 @@ import org.apache.daffodil.xml.XMLUtils
...
+  def this(parentArg: NodeInfo.Kind) = this(Seq(parentArg), Delay(Seq[NodeInfo.Kind](NodeInfo.Nothing)))
 
   // Eliminated a var here. Doing functional graph construction now below.
-  final override lazy val parents = parentsArg
-  final override lazy val children = childrenArg
+  final override lazy val children = childrenArg.value
+
+  override def preSerialization = {
+    super.preSerialization
+    children
```
stevedlawrence, line 81: We still require this preSerialization stuff to force the evaluation of these Delay values so that things can be garbage collected. Which means that if a user doesn't serialize the parser then things can't garbage collected. Or at least, they can't be garbage collected until a parse/unparse is run and these Delay's are used and evaluated. Which raises the issue, what if there are some Delay's that just happen to not be used? A single unevaluated Delay could hold on to a compile time object, which because it's a tree could hold onto basically the entire compile time stuff.

I think we have another bug for forcing all these serializtion things without actually serializing, so that maybe doesn't necessarily need to be done as part of this bug. But it's possible we won't see a difference until that's fixed.

Note if we didn't rely on writeObject to call preSerialization and we called it ourselves, I think the writeObject stuff could all go away, so fixing this would also simplify that I think.

stevedlawrence (approved):
+1

Great change! But might still be another aspect to fix. Could probably be done as another bug/PR.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
mbeckerle: I am adding a counter to Delay objects so we know how many there are, and after serialization an assertion checks that there the invariant that there are zero unevaluated ones.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
mbeckerle: Really used only by the toString method, to tell it whether to access the value or not.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/typeCalcFunctionErrors.tdml
```diff
@@ -53,60 +53,60 @@
...
-        <xs:restriction base="xs:int"/>
-      </xs:simpleType>
-    </xs:element>
-    <xs:element name="logicalTypeValue_bad_context_01" type="xs:string" dfdl:inputValueCalc="{ dfdlx:logicalTypeValue() }"/>
-    <xs:element name="logicalTypeValue_bad_context_02">
-      <xs:simpleType dfdlx:repType="tns:uint8" dfdlx:inputTypeCalc="{ dfdlx:logicalTypeValue() }" dfdlx:outputTypeCalc="{ 0 }">
-        <xs:restriction base="xs:int"/>
-      </xs:simpleType>
-    </xs:element>
+<!--    <xs:element name="typeCalcDispatch_typeError_03" dfdlx:parseUnparsePolicy="unparseOnly">-->
```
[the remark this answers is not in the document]
tuxji: Remove commented code as noted.

daffodil-test/src/test/scala/org/apache/daffodil/extensions/TestInputTypeValueCalc.scala
tuxji: Why can we remove those tests?  Oh, because they were testing experimental features such as output type calc next sibling, etc.  Agreed, let's remove them all.

tuxji (changes requested):
Would like to see some changes made before merging, but looks good so far.

mbeckerle (commented):
The issue of compile-time objects not being released and GC reclaimed remains.

The file TestForHeapDump.scala has two tests in it which can be used with a profiler like visualVM to get two heap dumps which one can compare/contrast.

One test is a very simple element, but includes the DFDLGeneralFormat. 

The other test is 100% self-contained. No includes/imports.

This latter one, there are no DSOM data objects or Grammar data objects in the heap. 

The former one, with the include statement, retains lots of DSOM and Grammar objects. 

So I've isolated the problem to having to do with include imports, but not yet figured out where the retained reference is exactly.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EncodingRuntimeData.scala
```diff
@@ -121,9 +113,9 @@ trait KnownEncodingMixin { self: ThrowsSDE =>
  */
 
 final class EncodingRuntimeData(
-  @TransientParam charsetEvArg: => CharsetEv,
+  val charsetEv: CharsetEv,
   override val schemaFileLocation: SchemaFileLocation,
-  optionUTF16WidthArg: Option[UTF16Width],
+  @TransientParam optionUTF16WidthArg: Option[UTF16Width],
```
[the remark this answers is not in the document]
stevedlawrence: I thought we only needed (and wanted) the TransientParam annotation on the pass by name args that get evaluated in preSerilize. That way, during serialization the pass-by-name args are forced to be evaluated, stored in some other val, and then is null when deserializaed, effectively allowing garbage collection of those pass-by-name things that might contain pointers to compile time stuff.

A TransientParam on a non-Arg thing seems wrong to me. It means after save/reload, that param will be null and could potentially lead to a NPE.

And with the new Delay stuff, I don't think we need any TransietParam things anymore, since the Delay handles the nullification of the pass-by-name stuff once evaluated so it can be garbage collected. If we want, we can still make the the Delays transient if we force their evaluation and store in a val like we currently do, and never access the Delay again. But we could just always ask for the value of the delay if we need it rather store storing the value in another val. Then we can get rid of all this transient nonsense.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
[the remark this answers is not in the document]
stevedlawrence: As an addendum to a previous comment, I think we can complete remove all transient params, as well as all the preserilize stuff.

The only place that really needs special serializtion logic is delay. We just need to add a writeObject for the Delay so that it forces evaluation of value and calls super.writeObject. That will set box to null, so the pass by name stuff isn't serialized. And the value is evaluated, so will be serialized correctly.

So if Delay parameters aren't made transient, and we add a little extra logic to Delay, all of this preserialization stuff goes away. I think...

mbeckerle:
I was able to identify why the compiler objects were being retained.

Turns out when you use Xerces to validate, it caches the errorHandler objects that were passed to it. 
The DFDLSchemaFile component (part of schema compiler) was acting as its own error handler by 
mixing in ErrorHandler and implementing the error, warning, and fatalError methods. 
Xerces cache would then hold onto the DFDLSchemaFile object, and that drags in every compiler data structure from it. 

Fixed by creating separate error handler. 

Created DAFFODIL-2541 for cleanup of XML loader error handling generally, which now that XML Loading is better centralized, can be far cleaner without all this error-handler complexity, all of which can be isolated in the DaffodilXMLLoader class.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -131,7 +131,7 @@ abstract class Expression extends OOLAGHostImpl()
     res
   }
 
-  private def compiledDPath_ = LV('compiledDPath) { compiledDPath }
+  private lazy val compiledDPath_ = LV('compiledDPath) { compiledDPath }
```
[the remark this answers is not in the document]
mbeckerle: Ah. I believe I understand this. The **_perceived_** number of objects retained was much larger when every property was held as a hash table entry, not in an object member. Now they're object members, so fewer bigger objects.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaComponent.scala
```diff
@@ -74,39 +71,15 @@ trait SchemaComponent
...
-      runtimeData)
+      schemaSet.typeCalcMap)
   }
 
   /**
    * Abbreviation. We use this very often.
    */
   final def ci = dpathCompileInfo
 
-  /**
```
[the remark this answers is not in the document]
mbeckerle: Actually, I did some additional movement of runtime1-specific things into mixins in the runtime1 package. 
This is part of making progress on DAFFODIL-2536, but is just a start really.

mbeckerle (commented):
I believe I have achieved the goals of this change. 

Compile-time objects no longer retained due to validator holding onto DFDLSchemaFile objects.

No more TransientParam for the RuntimeData objects. Less use (and better documented) for the few places left that do use TransientParam. Much less use of PreSerialization. 

There is still a need for serialization to force computation of delayed objects - in just a few places (type calculation stuff is one such - because the delays are not part of schema components where a requiredEvaluations thunk can force them automatically)

Improved OOLAG somewhat.  Cleaned up inconsistencies and errors in use of requiredEvaluationsIfActivated and requiredEvaluationsAlways (and added doc/guidance on when to use which.) A possible improvement is providing Global-element specific methods, and non-global element specific methods so that the usage becomes enforced and automatic. 

OOLAG simplified by removing the hash-tables that cached LV values. This is cleaner and simpler for debugging. The space tradeoff does not (at this point) appear important. This change is not essential, however. It is possible to go back to using def and LV that are hash-table entries per object to avoid the slot of a lazy val when the LV is unused.

However, more refactoring to get the runtime1 mixins out of the DSOM objects and separate runtime1 from the rest of the schema compiler will help with object size by the most direct means. 

Cyclic graphs constructed using Delay objects are now clearly doing exactly that. Though there are other uses of Delay besides this. I thought maybe Delay should become a base, and derived classes with names like Cyclic and OnlyIfNeeded would provide more self-documenting code.

mbeckerle:
I verified that a large suite of public and non-public DFDL schemas all continue to work with this.
Furthermore, they can all be run in one shot with 'sbt test' again now, probably due to prior fixes (not this change set) to the TDML Runner schema cache. 

Here's an image of VisualVM with an 8GB virtual machine running the suite of 60+ schemas and their tests:

image

Here's the output:
```
cd dfdl-schemas
utils/set-daffodil-version 3.2.0-SNAPSHOT
sbt test 2>&1 | grep Failed
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 14, Failed 0, Errors 0, Passed 14
[info] Passed: Total 9, Failed 0, Errors 0, Passed 9
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 18, Failed 0, Errors 0, Passed 18
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 37, Failed 0, Errors 0, Passed 37
[info] Passed: Total 45, Failed 0, Errors 0, Passed 45
[info] Passed: Total 65, Failed 0, Errors 0, Passed 65
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 31, Failed 0, Errors 0, Passed 31
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 175, Failed 0, Errors 0, Passed 175
[info] Passed: Total 65, Failed 0, Errors 0, Passed 65
[info] Passed: Total 40, Failed 0, Errors 0, Passed 40
[info] Passed: Total 12, Failed 0, Errors 0, Passed 12
[info] Passed: Total 8, Failed 0, Errors 0, Passed 8
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 5, Failed 0, Errors 0, Passed 5
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 62, Failed 0, Errors 0, Passed 62
[info] Passed: Total 3, Failed 0, Errors 0, Passed 3
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 22, Failed 0, Errors 0, Passed 22
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 4, Failed 0, Errors 0, Passed 4
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 3, Failed 0, Errors 0, Passed 3
[info] Passed: Total 5, Failed 0, Errors 0, Passed 5
[info] Passed: Total 3, Failed 0, Errors 0, Passed 3
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 11, Failed 0, Errors 0, Passed 11
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 7, Failed 0, Errors 0, Passed 7
[info] Passed: Total 7, Failed 0, Errors 0, Passed 7
[info] Passed: Total 6, Failed 0, Errors 0, Passed 6
[info] Passed: Total 10, Failed 0, Errors 0, Passed 10
[info] Passed: Total 8, Failed 0, Errors 0, Passed 8
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[info] Passed: Total 3, Failed 0, Errors 0, Passed 3
[info] Passed: Total 6, Failed 0, Errors 0, Passed 6
[info] Passed: Total 12, Failed 0, Errors 0, Passed 12
[info] Passed: Total 24, Failed 0, Errors 0, Passed 24
[info] Passed: Total 9, Failed 0, Errors 0, Passed 9
[info] Passed: Total 4, Failed 0, Errors 0, Passed 4
[info] Passed: Total 6, Failed 0, Errors 0, Passed 6
[info] Passed: Total 65, Failed 0, Errors 0, Passed 65
[info] Passed: Total 77, Failed 0, Errors 0, Passed 77
[info] Passed: Total 8, Failed 0, Errors 0, Passed 8
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[info] Passed: Total 694, Failed 0, Errors 0, Passed 694
[info] Passed: Total 8, Failed 0, Errors 0, Passed 8
[info] Passed: Total 31, Failed 0, Errors 0, Passed 31
[info] Passed: Total 694, Failed 0, Errors 0, Passed 694
[info] Passed: Total 9, Failed 0, Errors 0, Passed 9
[info] Passed: Total 8, Failed 0, Errors 0, Passed 8
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -99,31 +98,48 @@ object TermRuntimeData {
...
+ * around and ending up with objects being not garbage collectable.
+ *
+ * The relationhip between RuntimeData and PartialNextElementResolver is
+ * indeed cyclic. Simplest to think of it in terms of elements.
+ * An ERD has a partialNextElementResolver which exists to figure out
+ * the ERD of the next element, which could be this same ERD again, or
+ * could be some other ERD which can then be followed by an element with
+ * this same ERD again. So the cycle can be immediate or flow through
+ * many ERDs and partialNextElementResolvers. Turns out other kinds of
+ * terms can effect the next element resolver process (there's a stack
```
tuxji, line 121: Typo: effect -> affect

tuxji (approved):
+1

I didn't have time to review thoroughly, but looks good.  What was the reason for changing so many def's to lazy val's, how did that help achieve your goal?

mbeckerle:
> +1
> 
> I didn't have time to review thoroughly, but looks good. What was the reason for changing so many def's to lazy val's, how did that help achieve your goal?

So the short answer is that the LVs are always lazy. Implenting these in the "def" style creates a ton of smaller data objects stored in a hash table, and that made analyzing the heap dump too hard. Implementing as "lazy val" makes each DSOM object bigger, because there are lots of lazy val members that never get evaluated. But the hash table is not there, nor the objects in it. The heap dump is therefore far simpler, as the relationship of one object to another is often just one is a member of the other, vs one is in a hash table entry where that hash table is on the other object Anyway, I was able to dig through and spot the problem, which was actually in the validator holding on to the objects. 

Using regular lazy vals also helps with regular breakpoint debugging because you can see the values in a standard debugger inspection of the DSOM object, albeit you have to scroll through lots of members of the object most of which are unused. 

A long time back the whole OOLAG LV system used to be all done with lazy val. We switched to def with a hash table to store the LV objects as one attempt to address compilation time and space. 

However, that wasn't the primary problem with compilation time. It was copying (vs. sharing) of shared structures that was the biggest issue. Now that this is rectified (since 2.6.0), I prefer to use lazy val for these things because of the improved debug, and the simplification of the heap-dump analysis pushed me over the threshold where I decided to change them all back to just using lazy val.

daffodil-cli/src/test/scala/org/apache/daffodil/parsing/TestMain.scala
```diff
@@ -0,0 +1,51 @@
...
+//    // without having to run a staged daffodil to exercise
+//    // the CLI.
+//    //
+//    Main.main( Array[String]( "parse", "-s", testSchemaFile, "-r", "ABC"))
+//    //
+//    // We're expecting to see a SAXParseException reported about
+//    // fixed not a valid value for maxOccurs, not a valid value for
+//    // type nonNegativeInteger.
+//    //
+//  }
```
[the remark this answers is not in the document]
mbeckerle: I'll remove this.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SequenceTermRuntime1Mixin.scala
```diff
@@ -57,7 +57,7 @@ trait ChoiceBranchImpliedSequenceRuntime1Mixin { self: ChoiceBranchImpliedSequen
   override lazy val sequenceRuntimeData: SequenceRuntimeData = {
     new SequenceRuntimeData(
       position,
-      partialNextElementResolver,
+      Delay(partialNextElementResolver),
```
stevedlawrence, line 60: Is there anything that forces these Delays here to be evaluated?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  final lazy val variableRuntimeData = {
+    val vrd = new VariableRuntimeData(
+      this.schemaFileLocation,
+      this.diagnosticDebugName,
+      this.path,
+      this.namespaces,
+      this.external,
+      this.direction,
+      Delay(maybeDefaultValueExpr),
```
stevedlawrence, line 44: Simliar question to above, what causes this Delay to be evaluated? In this case, we do require that maybeDefaultalueExpr is evaluated, but this Delay contains the pass-by-name which isn't evaluated. Will there be any issues with this? I wonder if we need something like
```scala
lazy val maybeDefaultValueExprDelay = Delay { ... }

final lazy val variableRuntimeData  = {
  val vrd = new VariableRuntimeData(
    ...,
    maybeDefaultValueExprDelay,
    ...)
}

requiredEvaluationsAlways(maybeDefaultValueExprDelay.value)
```

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
```diff
@@ -247,10 +246,10 @@ class TestRefMap extends Logging {
         </xs:sequence>
       </xs:complexType>
       <xs:group name="g">
-        <xs:choice>
+        <xs:sequence><!-- a choice here would be UPA ambiguous -->
           <xs:element ref="ex:r1"/>
           <xs:element ref="ex:r1"/>
-        </xs:choice>
+        </xs:sequence>
```
stevedlawrence, line 252: Is this changing the test? Or was this just a broken test that we weren't detecting correctly?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
```diff
@@ -27,18 +27,71 @@ package org.apache.daffodil.util
...
+  // The idea is no matter what the Scala implementation is doing in its implementation of
+  // by-name args, that's on the constructor of the box object,
+  // and we're going to explicitly null-out the reference to the box object
+  // which guarantees we're not holding onto things we don't expect
+  // once the Delay object has been evaluated.
+  //
+  lazy val value: T = {
+    val v = box.value
+    box = null // throws away box, which allows GC of all closures, etc.
+    // Delay.evalDelay()
```
stevedlawrence, line 50: Remove comment

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
```diff
@@ -27,18 +27,71 @@ package org.apache.daffodil.util
...
+   * Create a string representation. Does not force the value to be computed.
+   */
   override def toString = {
-    val bodyString = if (hasValue) value.toString else "..."
+    val bodyString = if (box eq null) value.toString else "..."
     "Delay(" + bodyString + ")"
   }
+
+  override def preSerialization = {
+    value // force evaluation
```
stevedlawrence, line 69: I wonder if this really wants be an ``Assert(box != null)``. I imagine if we are serilazing a Delay, it's because it's part of the stuff needed for runtime, and that fact that it wasn't evaluated means that there was a bug in the code and we a missing something that forces that evaluation?

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
```diff
@@ -27,18 +27,71 @@ package org.apache.daffodil.util
...
-  def apply[T](v: => T) = new Delay(v)
+  /**
+   * Create a delayed expression object.
+   *
+   * @param delayedExpression an argument expression which will not be evaluated until required
+   * @tparam T type of the argument. (Usually inferred by Scala.)
+   * @return the Delay object
+   */
+  def apply[T](delayedExpression: => T) = {
+    // allocateDelay()
```
stevedlawrence, line 86: Remove comment

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -99,31 +98,48 @@ object TermRuntimeData {
 
 }
 
+/**
+ * Base runtime data structure for all terms
+ *
+ * These transient Delay-type args are part of how we
```
stevedlawrence, line 104: Is it worth removing "transient" in this comment. There are no longer any TransientParam stuff.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -527,53 +513,50 @@ final class SimpleTypeRuntimeData(
...
- * These objects have become too big. Most processors don't need most of this stuff.
- *
- * The objective should be to take things OUT of this structure and pass directly to the
- * constructor of the parser/unparser.
+/** Primary Runtime data structure for Elements
  *
  * These objects are for things that are generally heavily used everywhere like information for
  * providing diagnostics.
+ *
+ * These transient Delay-type args are part of how we
```
stevedlawrence, line 521: Same here, should we remove "transient"

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -823,160 +740,126 @@ final class NamespaceAmbiguousElementErrorERD(
   }
 }
 
+/**
+ * Base class for model group runtime data
+ *
+ * These transient Delay-type args are part of how we
```
stevedlawrence, line 746: Here too and other comments in this file

stevedlawrence (approved):
+1 to the new changes. Look great!

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
mbeckerle: I will give this a try and test it. I believe there are still some delays that aren't being forced because they are created in places outside of the OOLAG framework where errors can be caught properly.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
mbeckerle: Broken test. Was a choice between two identical element refs. That's a UPA error. Simplest kind actually. Validation detected this now, so had to fix it.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
mbeckerle: Good point. What we want it to allow construction of these cyclic objects, but something subsequently should force the delay on the outside of the runtimeData objects. Right now the required-evaluation forces the argument of the Delay, not the Delay itself which would be evaluated at serialization. No need for that. In fact per changes below, we can try to make that an error.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SequenceTermRuntime1Mixin.scala
mbeckerle: All are now evaluated. They are forced because ModelGroup has a runtime mixin, and that has a requiredEvaluation that demands modelGroupRuntimeData.initialize, and that initialize method is responsible to demand all delayed values. 

This is the pattern it all follows now. If you are doing functional graph creation with cycles, then there will be an initialize method which is responsible for forcing the delays so you have the graph fully formed.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
mbeckerle: Ended up with all objects that take delays (excepting the Include/Import stuff) due to cyclic connections all have an initialize member (lazy val) which is responsible for forcing the delays. So instead of having a requiredEvaluation for each delay, you have one for each object having delayed parameters.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Delay.scala
mbeckerle: This is now an assertion check that the Delay must have value or it fails.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -90,6 +90,8 @@ trait ElementBase
   requiredEvaluationsIfActivated(checkForAlignmentAmbiguity)
   requiredEvaluationsIfActivated(checkFloating)
 
+
```
mbeckerle, line 93: git rid of excess blank lines.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/Root.scala
```diff
@@ -45,6 +45,9 @@ final class Root(defXML: Node, parentArg: SchemaDocument,
     }
   })
 
+  // check for namespace-only ambiguities
+  requiredEvaluationsIfActivated(possibleNextLexicalSiblingStreamingUnparserElements)
+
```
mbeckerle, line 50: Verify this is needed. May not be.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -33,6 +33,24 @@ import org.apache.daffodil.dsom.PrimitiveType
...
+   * This does all type calculators for the simple types of elements.
+   * There are also global simple type defs to initialize, which have
+   * no corresponding element.
+   */
+  requiredEvaluationsIfActivated(optSimpleType.collect{
+    case stb: SimpleTypeDefBase =>
+      if (stb.optTypeCalculator.isDefined)
+        stb.optTypeCalculator.get.initialize()
+  })
+
```
mbeckerle, line 53: Arguably, the above should be on SimpleTypeRuntime1Mixin. A TypeCalculator is really a Runtime1 thing, and it is carried in the SimpleTypeRuntimeData object (I believe).

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -35,6 +35,17 @@ trait SchemaSetRuntime1Mixin { self : SchemaSet =>
...
+
+  /**
+   * This initialization is required for simpleTypeDefs used only
+   * for type calculations where those simpleTypeDefs do not have
+   * a corresponding element of that type.
+   */
+  requiredEvaluationsAlways(typeCalcMap.foreach{
+    case (_, typeCalculator) =>
+      typeCalculator.initialize()
+  })
```
mbeckerle, line 49: Arguably, the typeCalcMap ends up in a RuntimeData data structure such as the SchemaSetRuntimeData, so this initialization goes on that object's initialize method, and here we would just invoke that method.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/TermRuntime1Mixin.scala
```diff
@@ -409,7 +413,7 @@ trait TermRuntime1Mixin { self: Term =>
    * determines the ElementRuntimeData for that infoset event. This can be used to construct the initial
    * infoset from a stream of XML events.
    */
-  final lazy val partialNextElementResolver: PartialNextElementResolver = {
+  lazy val partialNextElementResolver: PartialNextElementResolver = {
```
mbeckerle, line 416: Verify this can't be final.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
```diff
@@ -85,7 +85,7 @@ trait DFDLNewVariableInstanceRuntime1Mixin { self: DFDLNewVariableInstance =>
       this.namespaces,
       defv.external,
       defv.direction,
-      Delay(maybeDefaultValueExpr),
+      maybeDefaultValueExprDelay,
```
mbeckerle, line 88: This should match what is in setVariable. Should explicilty create the Delay.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
```diff
@@ -102,6 +102,8 @@ trait DFDLNewVariableInstanceRuntime1Mixin { self: DFDLNewVariableInstance =>
 
     Maybe.toMaybe(defaultValExpr)
   }
+
+  val maybeDefaultValueExprDelay = Delay(maybeDefaultValueExpr, 'maybeDefaultValueExpr2)
```
mbeckerle, line 106: This line can be dropped.

daffodil-core/src/test/scala/org/apache/daffodil/api/TestAPI1.scala
```diff
@@ -239,6 +239,7 @@ class TestDFDLParser {
       // As of 18:31 UTC (EST+5) Jun 8, 2012
       TestUtils.testString(sch, "7,018,631,476")
     }
+    println(e)
```
mbeckerle, line 242: Remove println

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/NodeInfo.scala
```diff
@@ -386,19 +385,58 @@ object NodeInfo extends Enum {
...
     type Kind = ArrayIndexKind
   }
 
   protected sealed trait AnyDateTimeKind extends AnyAtomicKind
-  case object AnyDateTime extends TypeNode(AnyAtomic, Seq(Date, Time, DateTime)) with AnyDateTimeKind {
+  case object AnyDateTime extends TypeNode('AnyDateTime, AnyAtomic, Seq(Date, Time, DateTime)) with AnyDateTimeKind {
     type Kind = AnyDateTimeKind
   }
 
+  /**
```
mbeckerle, line 401: Could move back up to original location in the file.

mbeckerle (commented):
A few cleanups to do.

mbeckerle:
Alas, I have discovered that a bunch of well known DFDL schemas fail regression testing with this change. 
So there will be another fix on top of this once I figure out what is going on there. 
```
[error] Test com.tresys.mil_std_2045.Test2045Hdr.test_2045msghdr2 failed: java.util.NoSuchElementException: head of empty list, took 5.202 sec
[error]     at scala.collection.immutable.Nil$.head(List.scala:430)
[error]     at scala.collection.immutable.Nil$.head(List.scala:427)
[error]     at org.apache.daffodil.dpath.NamedStep.downwardStep$lzycompute(Expression.scala:1177)
```
another is, or appears to be new SDEs detected, but this is probably be due to a bug:
```
[error] Test com.tresys.gmtif.TestfGMTIF_DFDL.test_ba16Test1 failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) SchemaDefinitionError: Schema Definition Error: Cannot compare double with string for operator 'le'
[error] Schema context: raw Location line 265 column 6 in file:/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-gmtif/target/classes/com/tresys/gmtif/xsd/gmtif.dfdl.xsd
[error] SchemaDefinitionError: Schema Definition Error: No element corresponding to step tns:RevisitIndex found. Possibilities for this step include: tns:raw, tns:value.
```

daffodil-core/src/main/scala/org/apache/daffodil/dsom/IIBase.scala
```diff
@@ -167,11 +169,11 @@ abstract class IIBase( final override val xml: Node, xsdArg: XMLSchemaDocument,
    * we will detect it.
    */
   protected final lazy val seenBeforeThisFile: IIMap = LV('seenBeforeThisFile) {
-    val res = Delay {
+    val res = Delay( {
       val v = if (notSeenThisBefore) seenBefore.value + mapTuple
       else seenBefore.value
       v
-    }
+    }, 'seenBeforeThisFile)
```
stevedlawrence, line 176: Just a minor stylistic thing, thoughts on swapping the order of the Delay parameters so this would look like this instead:
```scala
    val res = Delay('seenBeforeThisFile, {
      val v = if (notSeenThisBefore) seenBefore.value + mapTuple
      else seenBefore.value
      v
    })
```
That feels a bit easier to read to me.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -178,10 +178,13 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
...
   final lazy val modelGroupRuntimeData = choiceRuntimeData
 
+
+
+
   final lazy val choiceRuntimeData = {
     new ChoiceRuntimeData(
       position,
-      Delay(partialNextElementResolver),
+      Delay(partialNextElementResolver, 'ChoicePartialNextElementResolver),
```
stevedlawrence, line 187: Does this not need a `val ...Delay` variable so that it can be forced to be evaluated?

stevedlawrence: Ah, nevermind, I see now. The initialize function on these runtime objects is responsible for forcing delays, and that function is in a `requiredEvaluation...``. Makes sense.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
stevedlawrence, line 106: Is this needed anymore? Or can you go back to using `Delay(maybeDefaultValueExpr)` and have the VRD.initialize method force the delay to be evaluated?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -138,6 +138,14 @@ sealed abstract class TermRuntimeData(
   val maybeCheckBitOrderAndCharsetEv: Maybe[CheckBitOrderAndCharsetEv])
   extends RuntimeData {
 
+  /**
+   * Cyclic structures require initialization
+   */
+  lazy val initialize: Unit = {
+    partialNextElementResolver
```
stevedlawrence, line 145: Should this be `partialNextElementResolver.force()`?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
```diff
@@ -858,6 +866,13 @@ final class VariableRuntimeData(
     namespacesArg,
     unqualifiedPathStepPolicyArg) {
 
+  /**
+   * Cyclic structures require initialization
+   */
+  lazy val initialize: Unit = {
+    maybeDefaultValueExpr // demand this
```
stevedlawrence, line 873: Same question as above, why don't these need a call to .force()?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/VariableRuntime1Mixin.scala
mbeckerle: yup. I'm fixing this so it matches the style of setVariable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
mbeckerle: So the arg, which is a delay is called partialNextElementResolverDelay.
The lazy val which is the value of this delay, forced is partialNextElementResolver.
So the initialize just asks for the value, that forces the delay. I'm trying to follow this style uniformly.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -198,29 +193,37 @@ class KeysetValueTypeCalculatorUnordered(valueMap: HashMap[DataValuePrimitive, D
...
 
-  override def supportsParse = maybeInputTypeCalcArg.isDefined
-  override def supportsUnparse = maybeOutputTypeCalcArg.isDefined
+  /*
+   * objects with Delay arguments for functional programming construction of
+   * cyclic graphs, need a way to force the delays, resulting in an ordinary
+   * (though cyclic) data structure.
+   */
+  final override def initialize(): Unit = {
+    Assert.invariant(maybeInputTypeCalcDelay ne null)
```
mbeckerle, line 208: FIXME: remove these asserts. They were for debugging.

mbeckerle (commented):
A number of FIXME comments represent small changes needed. Nothing functional spotted.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala
mbeckerle: Not fixing for now.

jadams-tresys (approved):
+1

Looks like a solid change.  Hopefully this makes things easier for new people to understand compared to the TransientParam stuff.  I know that threw me for a loop when I first ran into it.

PR #601, opened 2021-06-30
Distinct return codes for CLI

dasmallwood:


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1578,47 +1579,81 @@ object Main extends Logging {
     1
   }
 
+  object ExitCode extends Enumeration {
+
```
mbeckerle, line 1583: close up some excess whitespace lines in this.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -200,4 +203,26 @@ object Util {
     }
     inputFile
   }
+
```
mbeckerle, line 206: excess blank lines

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1578,47 +1579,81 @@ object Main extends Logging {
...
+    val BindingException = Value(4)
+    val NotYetImplemented = Value(5)
+    val TDMLException = Value(6)
+    val OutOfMemory = Value(7)
+    val UserDefinedFunctionFatalError = Value(8)
+    val BugFound = Value(9)
+
+
+    val ParseError = Value(20)
+    val UnparseError = Value(30)
```
mbeckerle, line 1600: They should be in numeric order, so move this one below.

mbeckerle (commented):
I'll approve this but I want others to weigh in on the values used for the various exit code values. 

There may be some conventions for these number assignments that are best-practice.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1578,47 +1579,81 @@ object Main extends Logging {
...
       }
       case e: OutOfMemoryError => {
         oomError(e)
+        ExitCode.OutOfMemory
       }
       case e: UserDefinedFunctionFatalErrorException => {
         log(LogLevel.Error, "%s", e.getMessage())
         e.printStackTrace()
-        1
+        ExitCode.UserDefinedFunctionFatalError
```
stevedlawrence, line 1649: Suggest renaming this to UserDefinedFunctionError. The "Fatal" part is a way for Daffodil to differentiate between different kinds of UDF errors. But by the time we get to the CLI, the only UDF error that a CLI user could ever see is the fatal error. So all UDF errors are implied to be fatal from a CLI perspective.

stevedlawrence (changes requested):

tuxji (changes requested):
Please don't wait for my approval since I'll be away.  Steve's comments are great, and he'll give you a +1 after you make his requested changes.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -880,7 +880,7 @@ object Main extends Logging {
     }
   }
 
-  def run(arguments: Array[String]): Int = {
+  def run(arguments: Array[String]): ExitCode.Value = {
```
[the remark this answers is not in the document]
stevedlawrence: PropertyGenerator probably isn't a big deal. That won't have access to the Main.scala exit code unless we want to move the exit codes to lib. But many of these codes are very specific to the CLI, so I'm not sure that makes sense. The property generator also isn't ever run by users, it's just generates code as part of the buil.

The interactive debugger is also an interesting case. Although the interactive debugger is trigger from the CLI, it doesn't necessarily have to be, so a sys.exit() might even be wrong here. I wonder if instead it should throw some kind of debugger or interrupt exception that the CLI can catch and exit with an appropriate error code. And no CLI uses could handle it differently?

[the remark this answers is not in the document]
stevedlawrence: Also, I would suggest that if the debugger exits via the "quit" command, that probably shouldn't be an error since that's correct behavior. So whatever happens when the debugger runs "quit", for the CLI it should probably result in a Success exit code.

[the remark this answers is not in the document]
mbeckerle: I would not change PropGen. The debugger can have a defined ExitCode or use generic ExitCode.Failure

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -301,7 +285,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments)
       }
 
     log(LogLevel.Error, "%s", msg)
-    sys.exit(1)
+    sys.exit(64)
```
stevedlawrence, line 288: Please add a enum for this value, something like `ExitCode.Usage` probably makes sense to match the name in `sysexits.h`

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1049,17 +1042,15 @@ object Main extends Logging {
                         "at least " + (bytesAvailable * 8)
                       }
                     val leftOverDataWarning = s"Left over data. Consumed ${loc.bitPos0b} bit(s) with ${remainingBits} bit(s) remaining." + firstByteString + dataHex + dataText
```
stevedlawrence, line 1044: This is no longer a warning, so this variable name isn't quite right. Maybe we should rename it `leftOverDataMessage`?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -990,30 +985,28 @@ object Main extends Logging {
...
                         if (loc.bitLimit0b.isDefined) {
                           (loc.bitLimit0b.get - loc.bitPos0b).toString
                         } else {
                           "at least " + (inStream.inputSource.bytesAvailable * 8)
                         }
-                      log(LogLevel.Warning, "Left over data after consuming 0 bits while streaming. Stopped after consuming %s bit(s) with %s bit(s) remaining.", loc.bitPos0b, remainingBits)
+                      log(LogLevel.Error, "Left over data after consuming 0 bits while streaming. Stopped after consuming %s bit(s) with %s bit(s) remaining.", loc.bitPos0b, remainingBits)
                       keepParsing = false
-                      error = true
+                      exitCode = ExitCode.NoForwardProgress
```
stevedlawrence, line 1004: I wonder if this should just be `ExitCode.LeftOverData` and we can remove `NoForwardProgress`.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1243,8 +1245,8 @@ object Main extends Logging {
           case Some(fileName) => new FileInputStream(fileName)
         }
 
-        val rc = processor match {
-          case None => 1
+        val rc: ExitCode.Value = processor match {
+          case None => ExitCode.UnableToCreateProcessor
           case Some(processor) => {
```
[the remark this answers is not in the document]
stevedlawrence: I don't think you addressed this comment.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1243,8 +1233,9 @@ object Main extends Logging {
           case Some(fileName) => new FileInputStream(fileName)
         }
 
-        val rc = processor match {
-          case None => 1
+        val rc: ExitCode.Value = processor match {
```
stevedlawrence, line 1236: I don't think you need the ExitCode.Value type here

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1720,7 +1722,7 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
                         |Immediately abort all processing.""".stripMargin
       override lazy val short = "q"
       def act(args: Seq[String], state: ParseOrUnparseState, processor: Processor): DebugState.Type = {
-        sys.exit(1)
+        throw new DebuggerExitException("")
```
stevedlawrence, line 1725: If the string is always going to be the empty string, I suggest just removing the `str: String` parameter.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1578,47 +1567,74 @@ object Main extends Logging {
...
+    val ParseError = Value(20)
+    val UnparseError = Value(21)
+    val GenerateCodeError = Value(23)
+    val TestError = Value(24)
+    val PerformanceTestError = Value(25)
+
+
+    val NoForwardProgress = Value(30)
+    val LeftOverData = Value(31)
+    val InvalidParserException = Value(32)
```
stevedlawrence, line 1590: This error code is no longer used, we should remove these, especially since there isn't a significant difference in mening from UnableToCreateProcessor.

stevedlawrence (changes requested):

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
stevedlawrence: Also, this exception isn't caught in the CLI, so if this is thrown it's going to show up as a bug. The CLI should catch this and output an appropriate error code. Probably ExitCode.Success is reasonable. since the debugger exiting is probably not considered a failure.

daffodil-cli/src/it/scala/org/apache/daffodil/executing/TestCLIexecuting.scala
```diff
@@ -130,6 +139,8 @@ class TestCLIexecuting {
     try {
       shell.sendLine(String.format("%s test -l %s", Util.binPath, testTdmlFile))
       shell.expect(contains(output16))
+
+      Util.expectExitCode(ExitCode.TestError, shell)
```
stevedlawrence, line 143: Some of these exit codes seem incorrect to me. For example, this test is just listing the tests in a tdml file. That should succeed, so I would expect this to have an exit code of ExitCode.Success. Feels like this test is checking for the wrong exit code and it's going to fail, or the CLI is returning a TestError when it should be returning a Success.

daffodil-cli/src/it/scala/org/apache/daffodil/executing/TestCLIexecuting.scala
```diff
@@ -162,6 +175,8 @@ class TestCLIexecuting {
       val cmd = String.format("%s test %s byte_entities_6_08", Util.binPath, testTdmlFile)
       shell.sendLine(cmd)
       shell.expect(contains("[Pass] byte_entities_6_08"))
+
+      Util.expectExitCode(ExitCode.TestError, shell)
```
stevedlawrence, line 179: Here's another example. We are running a TDML test, it's passing, yet we are checking for TestError. I would expect this to be a Success.

daffodil-cli/src/it/scala/org/apache/daffodil/generating/TestCLIGenerateC.scala
```diff
@@ -78,6 +83,8 @@ class TestCLIGenerateC {
     try {
       shell.sendLine(generateCmd)
       shell.expect(contains("Unknown option 's'"))
+
+      Util.expectExitCode(ExitCode.GenerateCodeError, shell)
```
stevedlawrence, line 87: I think this and many other tests in this file should be an ExitCode.Usage error.

And looking at the CI, many of these tests are failing because they are looking for the wrong exit code.

Make sure you are running the tests to make sure they work before pushing them for review. To run these CLi integration tests, you need to run them with 

```
sbt IntegrationTest/test
```

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -695,10 +695,13 @@ object Main extends Logging {
...
     //
     val schemaSource = URISchemaSource(schema)
-    val pf = Timer.getResult("compiling", {
+    val res = Timer.getResult("compiling", {
       val processorFactory = compiler.compileSource(schemaSource)
       if (!processorFactory.isError) {
         val processor = processorFactory.onPath(path.getOrElse("/")).withValidationMode(mode)
+        if (processor.isError) {
+          None
+        }
```
stevedlawrence, line 704: This doesn't actually return None from the function. It just means the result of this if statement is None, but that isn't used and so is essentially a no-op. This instead wants to be something like this:
```scala
val processor = ...
displayDiagnostics(processor)
if (processor.isError) {
  None
} else {
  Some(processor)
}
```

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1233,10 +1236,11 @@ object Main extends Logging {
           case Some(fileName) => new FileInputStream(fileName)
         }
 
-        val rc: ExitCode.Value = processor match {
+        val rc = processor match {
           case Some(proc) if (proc.isError) => ExitCode.UnableToCreateProcessor
           case None => ExitCode.UnableToCreateProcessor
           case Some(processor) => {
+            Assert.invariant(processor.isError == false)
```
stevedlawrence, line 1243: Make sure to remove the `case Some(proc) if (proc.isError)` line, that line shouldn't be possible if you've made the above change correctly. Also, note that a similar line occurs in multiple places in the file. Make sure to remove all places where we proc.isError is true. The only place that should be possible is in the createProcessorFrom* functions.

stevedlawrence (changes requested):

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -64,8 +64,8 @@ class TestCLIdebugger {
       shell.expect(contains("error: expression evaluation failed: Schema Definition Error:"))
       shell.expect(contains("(debug)"))
 
-      Util.expectExitCode(ExitCode.Failure, shell)
-      shell.send("quit")
+      shell.sendLine("quit")
+      Util.expectExitCode(ExitCode.BugFound, shell)
```
stevedlawrence, line 68: No tests should ever return a BugFound exit. If that's the case, we have a bug that needs to be fixed.

daffodil-cli/src/it/scala/org/apache/daffodil/performance/TestCLIPerformance.scala
```diff
@@ -50,7 +50,7 @@ class TestCLIPerformance {
       if (exitCode != 0)
         fail("Tests failed. Exit code: " + exitCode)
 
-      Util.expectExitCode(ExitCode.PerformanceTestError, shell)
+      Util.expectExitCode(ExitCode.Success, shell)
```
stevedlawrence, line 53: Looks like this test is already checking the exit code. I'd suggest you delete the previous stuff and only use the new expectExitCode stuff. Other tests do something similar-we should replace that with expectExitCode in all the integration tests that do that.

daffodil-cli/src/it/scala/org/apache/daffodil/performance/TestCLIPerformance.scala
```diff
@@ -180,7 +180,7 @@ class TestCLIPerformance {
       if (exitCode == 0)
         fail("Tests were successful when they were expected to fail. Exit code: " + exitCode)
 
-      Util.expectExitCode(ExitCode.PerformanceTestError, shell)
+      Util.expectExitCode(ExitCode.Success, shell)
```
stevedlawrence, line 183: It looks like this tests is expecting things to fail, and so shouldn't be ExitCode.Success. If this test is passing and exiting with Zero I think there is a bug somewhere.

daffodil-cli/src/it/scala/org/apache/daffodil/performance/TestCLIPerformance.scala
```diff
@@ -341,7 +341,7 @@ class TestCLIPerformance {
...
```
stevedlawrence, line 344: Same as above, looks like this tests should exit with a non-zero exit code.

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
```diff
@@ -284,7 +284,7 @@ class TestCLIunparsing {
       shell.sendLine(cmd)
       shell.expect(contains("There should be exactly one of the following options: schema, parser"))
 
-      Util.expectExitCode(ExitCode.UnparseError, shell)
+      Util.expectExitCode(ExitCode.Success, shell)
```
stevedlawrence, line 287: Seems like this should be a usage error.

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
```diff
@@ -400,7 +400,7 @@ class TestCLIunparsing {
       shell.expect(contains("Bad arguments for option 'schema'"))
       shell.expect(contains("Could not find file or resource"))
 
-      Util.expectExitCode(ExitCode.UnparseError, shell)
+      Util.expectExitCode(ExitCode.Success, shell)
```
stevedlawrence, line 403: Should be a UsageError

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1720,7 +1722,7 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
                         |Immediately abort all processing.""".stripMargin
       override lazy val short = "q"
       def act(args: Seq[String], state: ParseOrUnparseState, processor: Processor): DebugState.Type = {
-        sys.exit(1)
+        throw new DebuggerExitException()
```
stevedlawrence, line 1725: Nothing catches this exception. This should be caught in Main.scala so that we can exit with an appropriate exit code. A generic `ExitCode.Failure` seems reasonable.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -916,7 +912,10 @@ object Main extends Logging {
         }.map{ _.withExternalVariables(retrieveExternalVariables(parseOpts.vars, cfgFileNode))}
 
         val rc = processor match {
-          case Some(proc) if (!proc.isError) => {
+          case Some(proc) if (proc.isError) => ExitCode.UnableToCreateProcessor
```
stevedlawrence, line 915: This case is impossible and can be removed. With the above change, we will never get a Some with isError == true.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1084,8 +1078,10 @@ object Main extends Logging {
         }.map{ _.withExternalVariables(retrieveExternalVariables(performanceOpts.vars, cfgFileNode)) }
          .map{ _.withValidationMode(validate) }
 
-        val rc = processor match {
-          case Some(processor: DataProcessor) if (!processor.isError) => {
+        val rc: ExitCode.Value = processor match {
+          case Some(proc) if (proc.isError) => ExitCode.UnableToCreateProcessor
```
stevedlawrence, line 1082: Same as above, this case can be removed. We should an the case Some below to match the above code.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1334,9 +1326,9 @@ object Main extends Logging {
         val rc = processor match {
           case Some(processor) => {
             Timer.getResult("saving", processor.save(output))
```
stevedlawrence, line 1328: Should add an assert here to match the above cases.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -200,4 +203,21 @@ object Util {
...
+  def expectExitCode(expectedExitCode: ExitCode.Value, shell: Expect): Unit = {
+    val expectedCode = expectedExitCode.id
+    val exitCodeCmd = if (Util.isWindows) "echo %errorlevel%" else "echo $?"
+    shell.sendLine(exitCodeCmd)
+
+    val sExitCode = shell.expect(contains("\n")).getBefore()
+    val actualInt = Integer.parseInt(sExitCode.trim())
```
stevedlawrence, line 213: It looks like some tests are failing because sExitCode isn't an integer, it's like an error message or the empty string or something. My guess is this is because Daffodil just hasn't exited by the time we call this function, and so executing `echo ...` to display the error code won't run until after that finishes. But we're going to immediately start looking for a newline, so if Daffodil is still running and outputs a newline (which is very common), we'll think the exit code is just whatever Daffodil happened to output.

I wonder if maybe we need to do something like this:
```scala
val keyWord = "EXITCODE:"
val exitCodeCmd = "echo " + keyWord + (if (Util.isWindows) "%errorlevel%" else "$?")
shell.sendLine(exitCodeCmd)
shell.expect(contains(keyWord))
val sExitCode = shell.expect("\n").getBefore().trim()
val actualInt = Integer.parseInt(sExitCode)
...
```
So when we send the command to echo the exit code, we prepend a unique string. And then we then wait until we actually see this string. That way if Daffodil hasn't finished yet then we we'll wait until it is finished and the echo is actually run. Only once we see that do we then look for the newline and parse the int. This isn't tested, but something along these lines feels like the right approach.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1578,47 +1570,79 @@ object Main extends Logging {
...
       }
       case e: UserDefinedFunctionFatalErrorException => {
         log(LogLevel.Error, "%s", e.getMessage())
         e.printStackTrace()
-        1
+        ExitCode.UserDefinedFunctionError
+      }
+      case e: DebuggerExitException => {
+        log(LogLevel.Error, "%s", e.getMessage())
+        ExitCode.Failure
```
mbeckerle, line 1638: Should be ExitCode.Success. A quit command is a normal exit.

mbeckerle (approved):
One minor thing still: debugger exit should return ExitCode.Success.

With that fix, this looks to be a good improvement to the code base. 

+1

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: I could see an argument for it being both. It is definitely a normal exit, but it also means we exited without generating a complete infoset. I could imagine someone using a script that triggers the CLI debugger, but still wants the script to be able to detect if the infoset was complete or not once the debugger finishes.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -200,4 +203,25 @@ object Util {
...
     inputFile
   }
+
+  def expectExitCode(expectedExitCode: ExitCode.Value, shell: Expect): Unit = {
+    val expectedCode = expectedExitCode.id
+
+    val keyWord = "EXITCODE:"
+    val exitCodeCmd = "echo " + keyWord + (if (Util.isWindows) "%errorlevel%" else "$?")
+    shell.sendLine(exitCodeCmd)
+    shell.expect(contains(keyWord))
```
stevedlawrence, line 213: It looks like a handful of tests are failing to find this exit code, any idea why that is?

dasmallwood: from running the tests manually i've seen the Expect operation fails due to timeout

mbeckerle: Is there a race condition between what we're telling it to echo back, and stuff it was going to output anyway?

stevedlawrence: Looking at a handful of failing tests, it looks like ones that have this issue are ones that start the shell with:
```scala
val shell = Util.start("", true)
```
The `true` here is a flag sets the `expectErr` parameter, and looking at the getShell function, it ends up doing this:
```scala
    val inputStream = if (expectErr) {
      process.getErrorStream()
    } else {
      process.getInputStream()
    }
    val shell = new ExpectBuilder()
      .withInputs(inputStream)
      ...
```
I think what this is saying is that when `expectErr = true`, expect only captures the error stream (i.e. stderr) instead of the input stream (i.e. stdout). So when `Util.start("", true)`, expect will never see EXITCODE because that is always output to stdout, which isn't captured. So it's impossible to use `expectErr = true` and check the exitcode.

Since all integration tests now need access to stdout to check against the exit code, that suggests that we really can't use the `Util.start("", true)` function anymore to only capture stderr. We always need to capture stdout too.

So, I wonder if our integration tests should just always capture BOTH stdout and stderr, and when they want to check something in stderr, they need to be explicit about it?

Note that we already have a function that does this: `Util.sartIncludeErrors`. The difference here is if you want to expect something in stderr, you need to be explicit and do this:
```scala
shell.expectIn(1, ...)
```
The `1` here says to check stderr.

So I'm thinking maybe that suggests we should just get rid of `getShell` and rename `getShellWithErrors` to `getShell`. That way all tests will capture both stdout and stderr. And it means `expectErr` parameter just goes away, and all tests are started with `Util.start("")`. And then we need to update tests that are looking at stderr (i.e. the ones passing `true` to the `start()` method) to use `shell.expectIn(1, ...)` instead of `shell.expect(...)`. Looks like there's 36 of these tests, so a decent number, but not a ton, and the changes are fairly straightforward.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: Also, we should remove this `log()` call. Although this is an exception, it's an expected exception when the user quits the debugger. When they do this, they'll know that things are going to exit, so they don't need a message confirming things.

mbeckerle (approved):
If we can get the CI to run, I am ok with the changes here. 

I don't know why it doesn't run any more on each push of a new commit.

stevedlawrence:
Interesting, it looks like things are still failing on windows. And the failures are all with the error:
```scala
java.lang.NumberFormatException: For input string: "%errorlevel%"
```
So it seems our new expectExitCode capture our echo line, and not the results. Looking at the output for the integration tests, we see things like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<never-fails>2f6481e6-542c-11eb-ae93-0242ac130002</never-fails>
echo EXITCODE:%errorlevel%

D:\a\daffodil\daffodil>echo EXITCODE:%errorlevel%
EXITCODE:0
```
So we can see some output happened (the never-fails stuff), and before we got a prompt we call sendLine with `echo EXITCODE:%errorlevel%`, which we can see (that's output to the console). Then we will call `expect("EXITCODE")`. But even though echo command has already been output, windows will output it again, and we are going to match that. I guess Linux doesn't output that stuff twice. 

So as to a solution? I wonder if rather than expecting `EXITCODE:`, we can instead look for `\nEXITCODE', with a prepended newline? That way we won't match the keyword in the echo because there's a space before that keyword? Which would mean the expect line becomes this:

```scala
shell.expect(contains("\n" + keyWord))
```

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
```diff
@@ -25,15 +25,16 @@ import net.sf.expectit.matcher.Matchers.regexp
...
 import java.io.File
 import java.io.FileOutputStream
 import java.nio.file.Path
 import java.nio.file.Paths
-import scala.util.Failure
-import scala.util.Success
-import scala.util.Try
+//import scala.util.Failure
+//import scala.util.Success
+//import scala.util.Try
```
stevedlawrence, line 37: Remove commented code

daffodil-cli/src/it/scala/org/apache/daffodil/schematron/package.scala
```diff
@@ -74,12 +75,7 @@ package object schematron {
     try {
       shell.sendLine(cmd).expect(expectation)
 
-      val actualEc = shell.sendLine(echoEC).expect(matchEC).getBefore.trim.split(eol).last
```
stevedlawrence, line 77: The echoEC and matchEC variables can be removed below, they are no longer used.

tuxji:
> If we can get the CI to run, I am ok with the changes here.
> 
> I don't know why it doesn't run any more on each push of a new commit.

GitHub changed its CI policy two or three months ago because miners were spamming repos with one-off pull requests.  A new contributor now needs a committer to manually run CI with a button press each time.  Once the new contributor's first PR is merged, GitHub will run CI on the contributor's next PR automatically.  We could suggest that new contributors make a tiny PR as their first PR before making larger PRs.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -173,9 +173,13 @@ object Util {
     val expectedCode = expectedExitCode.id
 
     val keyWord = "EXITCODE:"
-    val exitCodeCmd = "echo " + keyWord + (if (Util.isWindows) "%errorlevel%" else "$?")
+//    val exitCodeCmd = "echo " + keyWord + (if (Util.isWindows) "%errorlevel%" else "$?")
+
+    val exitCodeCmd = "echo " + keyWord + (if (Util.isWindows) "^|%errorlevel%" else "\\!$?")
```
stevedlawrence, line 178: Please remove the commented code. Also, I think it's worth adding a comment explaining why we have this `^|` and `\\!` stuff. The idea being that the echo command needs to look different from what the echo outputs or the expect command might match the actual command. By echo things that are escaped (an escaped pipe in windows and an escaped exclamation in linux), then the echo command and the echo output include different text, so expect can tell the difference.

stevedlawrence:
Getting so close. Only a handful of windows tests are failing now. If we look at the output of a failing test, we see something like this:

```
 echo '<tns:e1 xmlns:tns="http://example.com">Hello</tns:e1>' | .\daffodil-cli\target\universal\stage\bin\daffodil.bat unparse -s .\daffodil-test\src\test\resources\org\apache\daffodil\section00\general\generalSchema.dfdl.xsd --root e1 -
D:\a\daffodil\daffodil>echo '<tns:e1 xmlns:tns="http://example.com">Hello</tns:e1>' | .\daffodil-cli\target\universal\stage\bin\daffodil.bat unparse -s .\daffodil-test\src\test\resources\org\apache\daffodil\section00\general\generalSchema.dfdl.xsd --root e1 -
echo EXITCODE:^|%errorlevel%
The system cannot find the file specified.
Error:  Unparse Error: Illegal content where start or end event expected: Error: Unexpected EOF in prolog
 at [row,col {unknown-source}]: [1,0] on line 1
Schema context: element reference tns:e1 Location line 19 in file:/D:/a/daffodil/daffodil/./daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchema.dfdl.xsd

D:\a\daffodil\daffodil>echo EXITCODE:^|%errorlevel%
EXITCODE:|21
```

I *think* this is the output for test_3529_CLI_Unparsing_SimpleUnparse_stdin3. This test runs the unparse command and then expects the output "Hello". I think this test passed previously for a similar reason that we just had to fix the expect exit code stuff: We are running a command, windows is outputting that command, and then we are matching the output. So when we `expect` "Hello", it's actually finding the `Hello` in the command, and not in the unparsed data. So these tests are just plain broken and we didn't know it.

So there are two issues:
1. Why are these tests failing on windows
2. Is there a way to ensure we aren't matching the command, and are instead matching the actual unparsed data.

I'm guessing that the answer to 1 is because these tests are using single quotes to escape the XML infoset. But single quotes don't do anything on windows--only double quotes escape things. So since the infoset isn't being escaped properly, windows see's the angle brackets as file redirection (hence "The system cannot find the file specified." error). We use single quotes in these tests because the infoset already includes double quotes in the XML. But I think an easy solution here is to just swap single quotes and double quotes for these tests that fail, so the command would look like this:
```scala
val cmd = String.format("""echo "<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>" | %s ... """)
```
So the double quotes escape things correctly, and we use single quotes inside the XML, which is still valid XML.

For issue 2, I think we probably need to make it so what we unparse is different from the content in the infoset. So if we try to unaprse an infoset containing the string "Hello" the result can't also be "Hello" or expect won't know the difference. One way to make sure there is a diffrence is to add an initiator to the schea element. So for example, if our schema looks like this:
```xml
<element name="e1" dfdl:lengthKind="explicit" dfdl:length="5" type="xs:string" />
```
We can change it to this instead:
```xml
<element name="e1" dfdl:initiator="init:" dfdl:lengthKind="explicit" dfdl:length="5" type="xs:string" />
```
This way if our infoset looks like this:
```xml
<tns:e1 xmlns:tns="http://example.com">Hello</tns:e1>
```
It will unparse to `init:Hello` rather than just "Hello". And then we can `expect` the full `init:Hello` and be confident that we're expecting the unparsed data rather than the infoset data. This does mean that test inputs and outputs need to change. Parsed data needs to look like "init:foo" and unparsed data needs to expect "init:foo"--the XML infoset stays the same though. 

An alternative might just to find a way to prevent windows from printing the command to stdout, but I'm not familiar enough with windows to know if that's possible.

stevedlawrence:
@dasmallwood, please rebase these commits onto the latest master and fix the conflicts. It seems GitHub will not run the checks until the conflicts are resolved.

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -203,4 +171,30 @@ object Util {
...
+
+    shell.sendLine(exitCodeCmd)
+    shell.expect(contains(exitCodeExpectation))
+
+    val sExitCode = shell.expect(contains("\n")).getBefore().trim()
+    val actualInt = Integer.parseInt(sExitCode)
+
+    if (actualInt != expectedCode) {
+      val expectedExitCodeName = expectedExitCode.toString
+      val actualExitCodeName = ExitCode(actualInt)
```
stevedlawrence, line 193: If the `actualInt` does not match the expected exit code, then that likely means there's some bug, and it's possible that `actualInt` doesn't even map to an ExitCode. This line would then will result in a NoSuchElement exception.

We could just not output an actual exit code name, but it probably is useful information to have when it's available. So instead we could do something like this to get an ExitCode name if it exists, otherwise just display "Unknown:
```scala
val actualExitCodeName = ExitCode.values.find { _.id == actualInt }.map { _.toString }.getOrElse("Unknown")
```

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -934,6 +958,8 @@ class TestCLIdebugger {
       shell.sendLine("continue")
       shell.expect(contains("<f xmlns=\"\">2</f>"))
       shell.sendLine("quit")
+
```
stevedlawrence, line 961: I think the quit here needs to be removed. I think it was accidentally added back when conflicts were fixed.

daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -974,6 +1000,8 @@ class TestCLIdebugger {
       shell.sendLine("continue")
       shell.expect(contains("<f xmlns=\"\">3</f>"))
       shell.sendLine("quit")
+
```
stevedlawrence, line 1003: Same here, I think this quit needs to be removed.

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
```diff
@@ -244,19 +255,20 @@ class TestCLIunparsing {
...
     val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
 
-    val shell = Util.start("", true)
+    val shell = Util.start("")
 
     try {
 
-      shell.sendLine(String.format("""echo '<tns:e1 xmlns:tns="http://example.com">Hello</tns:e1>' | %s -v unparse -s %s --root e1""", Util.binPath, testSchemaFile))
-      shell.expect(contains("[info]"))
+      shell.sendLine(String.format("""echo "<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>" | %s -v unparse -s %s --root e1""", Util.binPath, testSchemaFile))
```
stevedlawrence, line 262: Hmm, looks like using the double quotes still doesn't work. I guess echo on windows will still echo double quotes? I think maybe using our `Util.echoN` function will fix this, since that echos out data a little differently and I think will strip off the double quotes? So maybe something like this:

```scala
shell.sendLine(String.format(Util.echoN("\"<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>\"") + "| %s -v unparse ....", ...))
```
Note that  we still need the double quote in whatever is echo'ed to escape the angle brackets, so we need to include an escaped quote on either end of the string passed into `Util.echoN`.

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
```diff
@@ -259,13 +259,13 @@ class TestCLIunparsing {
...
-      shell.sendLine(String.format("""echo "<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>" | %s -v unparse -s %s --root e1""", Util.binPath, testSchemaFile))
+      shell.sendLine(String.format(Util.echoN("\"<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>\"") + "| %s -v unparse -s %s --root e1", Util.binPath, testSchemaFile))
       shell.expectIn(1, contains("[info]"))
 
-      shell.sendLine(String.format("""echo "<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>" | %s -vv unparse -s %s --root e1""", Util.binPath, testSchemaFile))
+      shell.sendLine(String.format(Util.echoN("\"<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>\"") + "| %s -vv unparse -s %s --root e1", Util.binPath, testSchemaFile))
       shell.expectIn(1, contains("[compile]"))
 
-      shell.sendLine(String.format("""echo "<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>" | %s -vvv unparse -s %s --root e1""", Util.binPath, testSchemaFile))
+      shell.sendLine(String.format(Util.echoN("\"<tns:e1 xmlns:tns='http://example.com'>Hello</tns:e1>\"") + "| %s -vvv unparse -s %s --root e1", Util.binPath, testSchemaFile))
```
stevedlawrence, line 268: Looks like you've only updated this single test `test_3580_CLI_Unparsing_SimpleUnparse_verboseMode`. This test is passing in the github actions on windows, so it does seem like echoN is the right fix.

But there are other tests that are failing that need the same change:
* test_3528_CLI_Unparsing_SimpleUnparse_stdin2
* test_3529_CLI_Unparsing_SimpleUnparse_stdin3
* test_3578_CLI_Unparsing_SimpleUnparse_defaultRoot
* test_3583_CLI_Unparsing_SimpleUnparse_rootPath

Also, these two tests use echo but happen to work because they are negative tests that fail for reasons other than echo issues. But for correctness these should use echoN too:
* test_3662_CLI_Unparsing_badSchemaPath
* test_3579_CLI_Unparsing_negativeTest

stevedlawrence:
+1

This is good to squash and be merged. Unfortunately, because there's a merge commit (which should be avoided in the future), you can't just use `rebase -i` that's described in the code contributor workflow to squash everything. Instead you need to do this:

```bash
# make sure you are on the branch to squash
git checkout daf-2235-exitcode

# create a patch that is the diff between your current branch and
# the most recent common ancestor with asf/master
git diff $(git merge-base HEAD asf/master) > patch.diff

# throw away all changes on your current branch by resetting to that
# common ancestor
git reset --hard $(git merge-base HEAD asf/master)

# apply the patch with changes staged
git apply --index patch.diff

# commit the already staged changes, making sure to follow the standards
# for a good commit message mentioned in the code contributor workflow
git commit
```

If things don't look right at any point above, or a command failed, you can always reset back to where we started by running running:

```bash
git reset --hard origin/daf-2235-exitcode
```
But otherwise, if everything looks good, you can force push the squashed commit to update this PR and delete the patch file:

```bash
# push your changes to update this PR
git push --force origin daf-2235-exitcode

# delete the patch
rm patch.diff
```

stevedlawrence (approved):

tuxji (approved):
+1

Lots of changes, looks good, hope we caught all the places where the tests exited the debugger already and didn't need to send a quit command before capturing the exit status.

stevedlawrence:
> hope we caught all the places where the tests exited the debugger already and didn't need to send a quit command before capturing the exit status.

I think we did. If we didn't, we would send the `quit` command to bash instead of the debugger. But `quit` isn't a valid command, so bash will set the exit code to 127 (command not found). And since we always check the exit code in these tests now, a test would fail if it got 127 instead of the expected exit code.

PR #602, opened 2021-07-01
Use a hidden sonar properties file

stevedlawrence:
This makes this consistent with all our other config files (e.g.
codecov, asf, github, sbt)

DAFFODIL-2275


.github/workflows/main.yml
```diff
@@ -166,3 +166,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        with:
+          args: >
+            -Dproject.settings=.sonar-project.properties
```
tuxji, line 171: Do we really have to split the args line?  The YAML folded style is nice for multiple arguments since the indention in each line will be ignored and a line break will be inserted at the end, but I'd rather see one line when we have one argument:

```yml
          args: -Dproject.settings=.sonar-project.properties
```

tuxji (approved):
+1

I agree, let's make the sonar properties file a hidden file.

.github/workflows/main.yml
stevedlawrence: Works for me.

mbeckerle (approved):
+1

PR #603, opened 2021-07-02
Daffodil 2473 remaining[WIP]

Samarth08:


daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1634,6 +1634,16 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
       case (RefQName(_, "rightShift", DFDLX), args) =>
         DFDLXShiftExpr(functionQNameString, functionQName, args,
           DFDLXRightShift(_, _))
+      case (RefQName(_, "xor", DFDLX), args) =>
```
mbeckerle, line 1637: xor, and, or should be renamed to xorBits, andBits, and orBits to clearly distinguish them from boolean operations.

XPath already has regular and/or operators. 

Given that these new things are functions, not operators, and they're in the dfdlx namespace that may seem like overkill, but I would propose that these get added to the DFDL expression language officially at some point, at which point the namespace prefix would go away and they might be recast as infix operators. 

Do we have a notBits operator. We need that also.

mbeckerle (changes requested):
Looks good. 

Please rename funtions with "Bits" suffix.
Add notBits function.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
Samarth08: For notBits we will have just one argument?

mbeckerle: Yes. That means another base class slightly different than the one you have for 2-args bits functions.

Samarth08: I have added not.

mbeckerle (changes requested):
Tiny corrections.
Really just need tests for bitAnd, bitOr, bitNot.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -98,6 +98,79 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+  override def computeValue(arg1: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+    argType match {
+      case NodeInfo.Long => ~arg1.getLong
+      case NodeInfo.Int => ~arg1.getInt
+      case NodeInfo.Short => ~arg1.getShort
+      case NodeInfo.Byte => ~arg1.getByte
+      case NodeInfo.UnsignedLong => arg1.getBigInt.not()
+      case NodeInfo.UnsignedInt => ~arg1.getLong
+      case NodeInfo.UnsignedShort => ~arg1.getInt
+      case NodeInfo.UnsignedByte => ~arg1.getShort
```
stevedlawrence, line 164: I *think* we need slightly different logic for each of the Unsigned operations, simiarly to what we needed to do for shifting. For example, say we have an xs:unsignedShort with a value of `1`. That will be represented here by a java Int. And `~1` is `-2`, so we end up with a negative value for something that should be unsigned.

I think we need to get the value, convert it to the Unsigned type, perform the not operation on that, and then convert back to the right type for the infoset. So for example, for UnsignedInt I think we want something like:

```scala
case NodeInfo.UnsignedInt => (~UInt(arg1.getLong.toInt)).toLong
```
I've only done a couple simple tests, but this seems to work for what I've tested, and follows a similar pattern to what we came up with for shiftLeft/right. We'll need something similar for each of the Unsigned types.

We'll definitely want some tests to make sure this is working as expected.

stevedlawrence (changes requested):

Samarth08:
After rebasing it with master facing issues with DaffodilTunablesGen.Scala while running the test cases.

stevedlawrence:
@Samarth08, what errors are you getting?

Samarth08:
overloaded method value load with alternatives:
  (url: java.net.URL)scala.xml.Node <and>
  (source: scala.xml.InputSource)scala.xml.Node <and>
  (sysID: String)scala.xml.Node <and>
  (reader: java.io.Reader)scala.xml.Node <and>
  (is: java.io.InputStream)scala.xml.Node
 cannot be applied to (org.apache.daffodil.api.URISchemaSource)
        val node = loader.load(URISchemaSource(configOpt.get))

stevedlawrence:
Commit 62ac1047a18979bea1f5d6c668eefeaaae39d64b changed the parameters to the `DaffodilXMLLoader.load` method, which is where the problem is on the line
```scala
val node = loader.load(URISchemaSource(configOpt.get))
```
That line should look like this now:
```scala
val node = loader.load(URISchemaSource(configOpt.get), Some(XMLUtils.dafextURI))
```
Which means that your `TunableGenerator.scala` file is old and isn't updated for some reason, but your `DaffodilXMLLoader.scala` file is the new correct one. Maybe you need to refresh files in your IDE or something? If you're still not sure, feel free to commit your changes and push to this PR and I can take a look and see whats going on.

Samarth08:
I have fetched the latest master but I am not able to run the tests.I am not able to test bitNot.Rest all are tested.

stevedlawrence:
Based on what you've pushed, I'm not sure why you would get the error about ``loader.load``. Did that go away?

Double check your ``DFDLXFunctions.scala`` file. There is a duplicate DFDLXBitAnd classe that are causing issues. Remove that duplicate. Also check ``Expression.scala`, that file is reference DFDLXAnd but should be DFDLXBitAnd.

Samarth08:
The loader.load error was resolved after doing sbt compile again.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -97,7 +97,59 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+}
+
+case class DFDLXBitNot(recipes: CompiledDPath, argType: NodeInfo.Kind) extends FNOneArg(recipes,argType) {
+  override def computeValue(arg1: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+    argType match {
+      case NodeInfo.Long => ~arg1.getLong
+      case NodeInfo.Int => ~arg1.getInt
+      case NodeInfo.Short => ~arg1.getShort
+      case NodeInfo.Byte => ~arg1.getByte
+      case NodeInfo.UnsignedLong => arg1.getBigInt.not()
```
Samarth08, line 143: @stevedlawrence Is this correct for unsignedNot?

stevedlawrence: I don't think so. The documentation for [BigInt.not() function](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#not()) says:

> This method returns a negative value if and only if this BigInteger is non-negative.

The value for us will always be non-negative since it's unsigned, which means just calling `getBigInt.not()` will always created a signed value, which is definitely wrong. Instead, we need to use the ULong, with something similar to the other unsigned types. I think something like this should do it:

```scala
case NodeInfo.UnsignedLong => (~ULong(arg1.getBigInt.longValue()).toBigInt
```
So we get the BigInteger, convert it to a long, then convert that to a ULong, `not` the ULong, and then convert that back to a BigInt.

Samarth08: I am not sure whether not is not working properly or is there something with the test cases.

stevedlawrence: Yeah, I hadn't totally understood how the ULong, UInt, etc. classes work. Apparently if you bitwise-not a UShort or a UByte, it returns a UInt. I guess this is similar to how shift's also return a UInt. Which all mimics Java and it's primitive types--in Java, if you bitwise-not a Short or Byte, you get an Int. I'm not sure why Java decided that made sense, but it means we have to do things differently.

I think probably the cleanest way to do it is like this:
```scala
case UnsignedShort => UShort((~arg1.getInt).toShort).toInt
```

So we get the argument from the infoset, immediately invert it, which will get us an Int. We then cast that Int down to a Short since that's what the UShort constructor expects. We create a UShort with that and then we can get the int value, which will have all the correct bits masked off. So we're basically just using UShort to correctly mask of the bits when we convert it back to an int.

Carrying this idea to the other types, I think we want something like this:

```scala
case UnsignedLong => ULong((~arg1.getBigInt.longValue)).toBigInt
case UnsignedInt => UInt((~arg1.getLong).toInt).toLong
case UnsignedShort => UShort((~arg1.getInt).toShort).toInt
case UnsignedByte => UByte((~arg1.getShort).toByte).toShort
```

Just a side note, that if we ever update the infoset to represent these unsigned types as ULong/UInt/etc in the infoset (see [DAFFODIL-2511](https://issues.apache.org/jira/browse/DAFFODIL-2511)), then this would probably look something like this:

```scala
case UnsignedLong => ~arg1.getULong
case UnsignedInt => ~arg1.getUInt
case UnsignedShort => (~arg1.getUShort).toUShort
case UnsignedByte => (~arg1.getUByte).toUByte
```
The UnsignedShort/Byte cases looks a bit weird due to bitwise-not returning a UInt, but its significantly cleaner.

Samarth08: This is also not working.
Should I remove not form the PR and merge rest of them.Not can be picked separately later.

stevedlawrence: Probably easier to get this working than to back out changes. Can you push what you have and we can take a look?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -97,7 +97,59 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+  }
+}
+
+case class DFDLXBitNot(recipes: CompiledDPath, argType: NodeInfo.Kind) extends FNOneArg(recipes,argType) {
+  override def computeValue(arg1: DataValuePrimitive, dstate: DState): DataValuePrimitive = {
+    argType match {
+      case NodeInfo.Long => ~arg1.getLong
+      case NodeInfo.Int => ~arg1.getInt
+      case NodeInfo.Short => ~arg1.getShort
+      case NodeInfo.Byte => ~arg1.getByte
```
stevedlawrence, line 142: Java's bitwise-not returns an Int, even if the type is short or a byte. But we need Short to stay Short, and Byte to stay Byte, so we need to case then. So these two lines should be:

```scala
case NodeInfo.Short => (~arg1.getShort).toShort
case NodeInfo.Byte => (~arg1.getByte).toByte
```
The unsigned cases looked right to me.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -110,13 +162,47 @@ case class DFDLXXor(recipes: List[CompiledDPath], argType: NodeInfo.Kind) extend
...
+      case NodeInfo.UnsignedShort => arg1.getInt | arg2.getInt
+      case NodeInfo.UnsignedByte => (arg1.getShort | arg2.getShort).toShort
+      //$COVERAGE-OFF$
+      case _ => Assert.invariantFailed(s"dfdlx:bitOr not supported")
+      // $COVERAGE-ON$
+    }
+  }
+}
 
+case class DFDLXNot(recipes: CompiledDPath, argType: NodeInfo.Kind) extends FNOneArg(recipes,argType) {
```
stevedlawrence, line 189: I think this is a duplicate class? DFDLXBitNot is the right one. I think this can be removed.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/BitFunctionsNot.tdml
```diff
@@ -0,0 +1,305 @@
...
+              <num>1</num>
+              <value>4294967294</value>
+            </ex:UnsignedIntNot2>
+          </ex:testUnsignedIntNot>
+        </tdml:dfdlInfoset>
+      </tdml:infoset>
+  </tdml:parserTestCase>
+
+
+  <tdml:parserTestCase name="testUnsignedLongNot" root="testUnsignedLongNot" model="BitFunctions">
```
stevedlawrence, line 252: This test isn't right. This is bitNot of unsigned types, so the <value> part of the infoset should be unsigned, but these test expects -3 and -1. Those should be something else that is signed.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/BitFunctionsNot.tdml
```diff
@@ -0,0 +1,305 @@
...
+              <num>1</num>
+              <value>4294967294</value>
+            </ex:UnsignedShortNot2>
+          </ex:testUnsignedShortNot>
+        </tdml:dfdlInfoset>
+      </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="testUnsignedByteNot" root="testUnsignedByteNot" model="BitFunctions">
+    <tdml:document>-1|6</tdml:document>
```
stevedlawrence, line 289: -1 isn't a valid unsignedByte so this will fail to parse. The test also expects the bitwise-not of 6 is -7, which isn't correct for unsigned bytes.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1634,9 +1634,19 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
           DFDLXXor(_,_))
+      case (RefQName(_, "bitAnd", DFDLX), args) =>
+        DFDLXBitExpr(functionQNameString, functionQName, args,
+          DFDLXBitAnd(_,_))
+      case (RefQName(_, "bitOr", DFDLX), args) =>
+        DFDLXBitExpr(functionQNameString, functionQName, args,
+          DFDLXOr(_,_))
+      case (RefQName(_, "bitNot", DFDLX), args) =>
+        DFDLXBitNot(functionQNameString, functionQName, args,
+          DFDLXNot(_,_))
```
stevedlawrence, line 1648: With this renamed, this should become:
```scala
 DFDLXBitNotExpr(functionQNameString, functionQName, args,
    DFDLXBitNot(_,_))
```

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2205,6 +2215,31 @@ case class DFDLXBitExpr(nameAsParsed: String, fnQName: RefQName,
     res
   }
 }
+
+case class DFDLXBitNot(nameAsParsed: String, fnQName: RefQName,
```
stevedlawrence, line 2219: I think this should be renamed DFDLXBitNotExpr. DFDLXBitNot should be the name of the runtime class that actually does the bitwise-not.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -97,7 +97,59 @@ case class DFDLXRightShift(recipes: List[CompiledDPath], argType: NodeInfo.Kind)
...
+      case NodeInfo.UnsignedInt => UInt((~arg1.getLong).toInt).toLong
+      case NodeInfo.UnsignedShort => UShort((~arg1.getInt).toShort).toInt
+      case NodeInfo.UnsignedByte => UByte((~arg1.getShort).toByte).toShort
+      //$COVERAGE-OFF$
+      case _ => Assert.invariantFailed(s"dfdlx:NOT not supported")
+      // $COVERAGE-ON$
+    }
+  }
+}
 case class DFDLXXor(recipes: List[CompiledDPath], argType: NodeInfo.Kind) extends FNTwoArgs(recipes) {
```
stevedlawrence, line 153: This class should be rename DFDLXBitXor

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -110,13 +162,47 @@ case class DFDLXXor(recipes: List[CompiledDPath], argType: NodeInfo.Kind) extend
       case NodeInfo.UnsignedShort => arg1.getInt ^ arg2.getInt
       case NodeInfo.UnsignedByte => (arg1.getShort ^ arg2.getShort).toShort
           //$COVERAGE-OFF$
-      case _ => Assert.invariantFailed(s"dfdlx:XOR not supported")
+      case _ => Assert.invariantFailed(s"dfdlx:bitXOR not supported")
```
stevedlawrence, line 165: This error message should be dfdlx:bitXor to match the function name that we actually expect.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1634,9 +1634,19 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
       case (RefQName(_, "rightShift", DFDLX), args) =>
         DFDLXShiftExpr(functionQNameString, functionQName, args,
           DFDLXRightShift(_, _))
-      case (RefQName(_, "xor", DFDLX), args) =>
+      case (RefQName(_, "bitXor", DFDLX), args) =>
         DFDLXBitExpr(functionQNameString, functionQName, args,
           DFDLXXor(_,_))
```
stevedlawrence, line 1639: This should be DFDLXBitXor

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1634,9 +1634,19 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
...
+      case (RefQName(_, "bitAnd", DFDLX), args) =>
+        DFDLXBitExpr(functionQNameString, functionQName, args,
+          DFDLXBitAnd(_,_))
+      case (RefQName(_, "bitOr", DFDLX), args) =>
+        DFDLXBitExpr(functionQNameString, functionQName, args,
+          DFDLXOr(_,_))
```
stevedlawrence, line 1645: This should be DFDLXBitOr

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1642,10 +1642,10 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           DFDLXBitAnd(_,_))
       case (RefQName(_, "bitOr", DFDLX), args) =>
         DFDLXBitExpr(functionQNameString, functionQName, args,
-          DFDLXOr(_,_))
+          DFDLXBitOr(_,_))
       case (RefQName(_, "bitNot", DFDLX), args) =>
-        DFDLXBitNot(functionQNameString, functionQName, args,
-          DFDLXNot(_,_))
+        DFDLXBitExprNot(functionQNameString, functionQName, args,
```
stevedlawrence, line 1647: Minor thought looking at this one last time. It might make sense to rename `DFDLXBitExpr` to `DFDLXBitBinaryExpr` and `DFDLXBitExprNot` to `DFDLXBitUnaryExpr`. Those two renames make the difference between the two class more clear (i.e. binary vs unary). And it also makes it clear that Unary isn't necessarily specific to NOT, even though that's currently the only thing we use it for. Maybe someday we'll add NAND, NOR or some other binray or unary operations, and it will be more clear which of DFDLXBitBinaryExpr or DFDLXBitUnary is the right thing to use.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
```diff
@@ -138,14 +138,14 @@ case class DFDLXBitNot(recipes: CompiledDPath, argType: NodeInfo.Kind) extends F
...
-      case NodeInfo.Byte => ~arg1.getByte
+      case NodeInfo.Short => (~arg1.getShort).toShort
+      case NodeInfo.Byte => (~arg1.getByte).toByte
       case NodeInfo.UnsignedLong => ULong((~arg1.getBigInt.longValue)).toBigInt
       case NodeInfo.UnsignedInt => UInt((~arg1.getLong).toInt).toLong
       case NodeInfo.UnsignedShort => UShort((~arg1.getInt).toShort).toInt
       case NodeInfo.UnsignedByte => UByte((~arg1.getShort).toByte).toShort
       //$COVERAGE-OFF$
-      case _ => Assert.invariantFailed(s"dfdlx:NOT not supported")
+      case _ => Assert.invariantFailed(s"dfdlx:Not not supported")
```
stevedlawrence, line 148: This should be `dfdlx:bitNot` in the error message.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DFDLXFunctions.scala
stevedlawrence: Please rename this class to be consistent with the to others. Should be DFDLXBitXor.

stevedlawrence (approved):
+1, just needs a couple more renames to make things more clear.

This also needs to be rebased +squashed to resolve conflicts before it can be merged.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1629,22 +1629,22 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
         FNExactlyOneExpr(functionQNameString, functionQName, args)
 
       case (RefQName(_, "leftShift", DFDLX), args) =>
-        DFDLXShiftExpr(functionQNameString, functionQName, args,
+        DFDLXBitBinaryExpr(functionQNameString, functionQName, args,
           DFDLXLeftShift(_, _))
       case (RefQName(_, "rightShift", DFDLX), args) =>
-        DFDLXShiftExpr(functionQNameString, functionQName, args,
+        DFDLXBitBinaryExpr(functionQNameString, functionQName, args,
```
stevedlawrence, line 1635: left shift and right shift should still use DFDLXShiftExpr. Although they take two arguments like BinaryExpr, they have different restrictions. For example, the second argument must always be a unsigned int for shift. So it's sort of like unary, but with an extra argument. I think keeping these two as DFDLXShiftExpr makes sense. No rename or anything needed here.

stevedlawrence:
Latest changes look good to me, but we can't rebase/merge this PR because it contains conflicts due to the merge commit. In the future, avoid running `git merge` or `git pull` (which does a merge behind the scenes), since they lead to these merge commits that cause issues. Lots of places online say to use git pull or git merge, but for Daffodil those commands should be avoided.

Here's the commands to fix your branch up and squash into a single commit:

```bash
# make sure we are on your branch
git checkout DAFFODIL-2473_Remaining

# create a backup branch just in case we mess things up
git branch DAFFODIL-2473-backup

# create a diff between your branch and master
git diff asf/master > daffodil-2473.patch

# move your branch to asf/master
git reset --hard asf/master

# apply the patch and stage the changes
git apply --index daffodil-2473.patch

# delete the patch file
rm daffodil-2473.patch

# commit your changes, add a new commit message
git commit
```

If that all works, you should now have a single commit containing your squashed changes on top of asf/master. You can force push that to your fork to update the PR, and delete the backup branch.
```bash
# update the PR with your new changes (assuming "origin" is your fork)
git push --force origin DAFFODIL-2473_Remaining

# delete backup branch
git branch -D DAFFODIL-2473-backup
```

If at any point a command doesn't work or looks wrong, you can always reset to the backup branch and it will be like none of this happened:
```
git reset --hard DAFFODIL-2474-backup
```
I've also got a clone of your PR, so if something goes really wrong, I'll have the changes and we can restore them.

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -2205,6 +2215,31 @@ case class DFDLXBitExpr(nameAsParsed: String, fnQName: RefQName,
...
+
+case class DFDLXBitUnaryExpr(nameAsParsed: String, fnQName: RefQName,
+                        args: List[Expression], constructor: (CompiledDPath, NodeInfo.Kind) => RecipeOp)
+  extends FunctionCallBase(nameAsParsed, fnQName, args) {
+  override lazy val inherentType = {
+    val arg0Type = args(0).inherentType
+    schemaDefinitionUnless(
+      arg0Type.isSubtypeOf(NodeInfo.PrimType.UnsignedLong) || arg0Type.isSubtypeOf(NodeInfo.PrimType.Long),
+      "Both arguments for %s must be either xs:unsignedLong or xs:long or a subtype of those, but was %s.",
+      nameAsParsed,arg0Type.globalQName)
```
tuxji, line 2227: Steve's comment about `Both arguments ...` above still needs to be addressed.  Also, we prefer spaces between comma-separated arguments on this line (2227) and line 2237 (also compare lines 2236 & 2237 with line 2253 & 2254 - please follow the same formatting as existing code).

tuxji (approved):
Provisional +1 once all requested changes are completed, rebased, and squashed into a single commit.  

I have a slight concern about the tests.  The functions are bit-oriented operations, but the tests test only a single bit position (all test input values are 0 and 1).  It might be a good idea to replace these test input values with 2-bit-wide values (0, 1, 2, 3 corresponding to 0b00, 0b01, 0b10, 0b11) in order to verify that the functions indeed operate upon each bit in parallel.

mbeckerle (approved):
+1

stevedlawrence:
+1 to the latest changes.

Please squash into a single commit following the steps in https://github.com/apache/daffodil/pull/603#issuecomment-883793657 and this can be merged.

stevedlawrence:
Looks good! Thanks!

PR #616, opened 2021-07-26
Check for single commit in pull requests

stevedlawrence:
- Also shorten the workflow name. In some displays, github truncates the
  workflow + job name to the point where you only see a few letters of
  the job name if the workflow is too long, making it hard to tell which
  job is which.

DAFFODIL-2252


mbeckerle (approved):
+1

tuxji (approved):
+1

I see you've already tested in your repo, thanks.

PR #618, opened 2021-07-26
dfdl prefix properties on dfdl annotation elements should generate us…

mphyo21:
…eful diagnostic

-Added TestDefError.scala file and disabled validateTDML setting to allow the parser to go through the whole file and pickup the TDML error.
-Added generalSchemaDefinitionError.tdml test to catch the TDML "SAXParseException" error

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/dfdlFormatPrefixedAttributesSDE.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/dfdlFormatPrefixedAttributesSDE.tdml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<tdml:testSuite
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:ex="http://example.com"
+  xmlns:tns="http://example.com">
+
+  <tdml:defineSchema name="s1">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" dfdl:lengthKind="delimited"/>
+    <xs:element name="root" type="xs:string"/>
+  </tdml:defineSchema>
+
+<!--
+        Test Name: dfdl_format_prefixed_attributes_sde
+        Schema: s1
+        Root: root
+        Purpose:
+        Daffodil should issue a specific diagnostic whenever it finds dfdl: prefixed attributes in
+        the annotation elements dfdl:format, dfdl:element, dfdl:simpleType, dfdl:sequence, dfdl:choice, etc.
+        It should point out specifically that the author probably wanted to remove the prefix dfdl:
+-->
+    <tdml:parserTestCase name="dfdl_format_prefixed_attributes_sde" root="root"  model="s1" description="">
+        <tdml:document />
+        <tdml:errors>
+            <tdml:error>Schema Definition Error</tdml:error>
+            <tdml:error>dfdl:lengthKind</tdml:error>
+        </tdml:errors>
+    </tdml:parserTestCase>
+</tdml:testSuite>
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchedmaDefinitionError.tdml
```diff
@@ -0,0 +1,54 @@
...
+
+  <tdml:parserTestCase name="schema_definition_err" root="root"
+    model="s1"
+    description="">
+    <tdml:document />
+ <tdml:errors>
+<tdml:error>SAXParseException</tdml:error>
+ <tdml:error> Attribute 'dfdl:lengthKind' is not allowed to appear in element 'dfdl:format'.</tdml:error>
+  </tdml:errors>
+</tdml:parserTestCase>
```
stevedlawrence, line 53: The indentation is off in this file, can you fix this so child element are properly indented?

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchedmaDefinitionError.tdml
```diff
@@ -0,0 +1,54 @@
...
+      It should point out specifically that the author probably wanted to remove the prefix dfdl:
+-->
+
+  <tdml:parserTestCase name="schema_definition_err" root="root"
+    model="s1"
+    description="">
+    <tdml:document />
+ <tdml:errors>
+<tdml:error>SAXParseException</tdml:error>
+ <tdml:error> Attribute 'dfdl:lengthKind' is not allowed to appear in element 'dfdl:format'.</tdml:error>
```
stevedlawrence, line 51: We usually try to make expected error message somewhat generic so that if we ever change the error message we don't need to update the tests. Especial since this error message comes from SAX, which we don't have any control over. It's also worth looking for the word "Schema Definition Error" since that's the kind of error that we should expect to see.

So I would suggest this wants to be something like
```xml
<tdml:error>dfdl:lengthKind</tdml:error>
<tdml:error>dfdl:format</tdml:error>
```
Maybe that's too generic, but the idea is that we just look for certain keywords/phrases that must exist in the error, rather than looking for the entire error message. I think it's also worth loo

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchedmaDefinitionError.tdml
```diff
@@ -0,0 +1,54 @@
...
+<!--
+    Test Name: schema_definition_err
+       Schema: s1
+         Root: root
+      Purpose:
+      Daffodil should issue a specific diagnostic whenever it finds dfdl: prefixed attributes in the annotation elements dfdl:format, dfdl:element, dfdl:simpleType, dfdl:sequence, dfdl:choice, etc.
+      It should point out specifically that the author probably wanted to remove the prefix dfdl:
+-->
+
+  <tdml:parserTestCase name="schema_definition_err" root="root"
```
stevedlawrence, line 45: This name is a bit too generic, there's lot's of things that could cause a schema definition error. I would suggest something like "dfdl_format_prefixed_attributes_sde" or something, so just by looking at the name you might be able to infer that this test is about prefixed attributes in a dfdl format element.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchedmaDefinitionError.tdml
```diff
@@ -0,0 +1,54 @@
...
+    <dfdl:format ref="ex:GeneralFormat" dfdl:lengthKind="delimited"/>
+    <xs:element name="root" type="xs:string"/>
+  </tdml:defineSchema>
+
+<!--
+    Test Name: schema_definition_err
+       Schema: s1
+         Root: root
+      Purpose:
+      Daffodil should issue a specific diagnostic whenever it finds dfdl: prefixed attributes in the annotation elements dfdl:format, dfdl:element, dfdl:simpleType, dfdl:sequence, dfdl:choice, etc.
```
stevedlawrence, line 41: Please wrap this comment so it isn't as long. I'm not sure if we have standard width, but somewhere around 80-100 characters seems generally what we do.

stevedlawrence (approved):
+1 with a few minor changes.

Also, once you get another +1 and things need to be squashed to a single commit, make sure to add the bug number to the bottom of the commit message.

.sbtopts
```diff
@@ -12,5 +12,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
--mem 4096
+#-mem 4096
+-J-Xms4096m
+-J-Xmx4096m
+-J-Xss4M
+-J-XX:ReservedCodeCacheSize=512m
```
tuxji, line 19: You don't need to include this changed file since the same change was already merged in another pull request today.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDefError.scala
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.daffodil.section00.general
+
+import org.apache.daffodil.tdml.Runner
+import org.junit.{AfterClass, Test}
+
+object TestDefError {
+
+  val testDir = "/org/apache/daffodil/section00/general/"
+  val runner = Runner(testDir, "generalSchedmaDefinitionError.tdml",validateTDMLFile = false)
```
tuxji, line 26: In addition to changing the test name as Steve suggested, I suggest also changing the scala and tdml names.  I'm fine with a shorter name than "dfdl_format_prefixed_attributes_sde" or a camel case name, just make the name more descriptive/specific.  Please also pay attention to Scala's [style guide](https://docs.scala-lang.org/style/), e.g., each comma in the above line should have a space character after it.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDefError.scala
```diff
@@ -0,0 +1,39 @@
...
+  @AfterClass def tearDown(): Unit = {
+    runner.reset
+  }
+
+}
+
+class TestDefError {
+
+  import TestDefError._
+  @Test def test_schema_definition_err(): Unit = { runner.runOneTest("schema_definition_err") }
```
tuxji, line 37: When you change the test name in the tdml file, you'll need to change it here as well.

tuxji (approved):
+1 after suggested changes are made

Last suggestion, please be aware that git commit messages have a [standard](https://gist.github.com/turbo/efb8d57c145e00dc38907f9526b60f17) too.  Your commit message shouldn't have a long first line that wraps to the second line and it should have "DAFFODIL-1512" as its last line.

mbeckerle (approved):
+1

PR #620, opened 2021-08-17
Improve path expression error handling

stevedlawrence:
- The context of a dfdl:occursCountKind expression evaluation is
  actually not the element that the property lives on, but on the
  elements parent. This is because the element isn't created until the
  number of occurrences are known. This means that such expressions must
  either be absolute, or start with an ".." upward move. We previously
  had an assertion that checked this, but there's nothing actually
  preventing expressions without upward moves. This assertion is instead
  changed to an SDE.
- The conversions for path expressions were applied only on the last
  downward step. But this meant that if an expression path was made up
  of only upward steps, no conversions would occur. Such expressions are
  rarely reasonable, but are legal, and should usually result in an
  error. But because no conversions happened, where certain types of
  error checks occur, it meant it was possible to miss certain
  compile-time convertability checks, allowing for expressions to fail
  at runtime with a NullPointerExceptions. This modifies conversions to
  be applied to the entire path, instead of on the last downward step.
  This doesn't change what conversions are applied, only ensures that
  they are always applied.
- Additional refactorings related to path steps to make it a little more
  clear what's going on.

DAFFODIL-2553

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -716,7 +716,7 @@ case class UnaryExpression(op: String, exp: Expression)
 
 abstract class PathExpression()
   extends Expression {
-  def steps: Seq[StepExpression]
+  def steps: List[StepExpression]
 
   override def text = steps.map { _.text }.mkString("/")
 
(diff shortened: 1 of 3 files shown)


tuxji (approved):
+1

Simpler code, and some new tests.  Looks good to me.

mbeckerle (approved):
+1 Good improvement and nice cleanups

PR #621, opened 2021-08-18
Ensure newlines are output when debugger displays hidden groups

stevedlawrence:
The debugger uses the XMLTextInfosetOutputter to display the infoset,
which uses hasVisibleChildren to determine when to output newlines in
complex elements--empty complex elements should not output a newline.

However, hasVisibleChildren assumes hidden elements won't ever be
output, but that isn't the case in the debugger. This means that when
the debugger outputs an infoset, newlines aren't correctly output within
hidden groups of empty complex elements.

Since the hasVisibleChildren variable is only used in the
XMLTextInfosetOutputter, and since it doesn't work with the debugger,
this patch just removes it. It is replaced with state added to the
infoset outputter to keep track of when to output newlines when
outputting XML.

DAFFODIL-2554


tuxji (approved):
+1

Logic looks correct to me.

mbeckerle (approved):
+1

PR #622, opened 2021-08-19
Fix parser serialization stack overflow regression

stevedlawrence:
Commit 126789891ed5e834037198e5aa59e1f6b34dcb04 added a fix to prevent
stack overflows during serialization by making parent backpointers
transient and manually serializing them, which avoids serialization of a
highly connected/cyclic graph that requires a large stack depth.

Commit eb603e7f4a342e1a63d1e07f59c714cf531724bc removed the need for
most transient parameters, but accidentally removed the transient
annotations for these parents needed to fix the stack overflow issue.

This adds back the @transient annotations needed to prevent stack
overflows, as well as additional comments to hopefully ensure we don't
accidentally remove these transients in the future.

DAFFODIL-2551


tuxji (approved):
+1

Looks good.

mbeckerle (approved):
+1

PR #623, opened 2021-08-19
Ensure we reset groupIndexStack in sequence when an exception is thrown

stevedlawrence:
While parsing sequences of elements in the SequenceBaseParer, it is
possible that an exception could be thrown.

For example, an easy way to cause this is to have an occursCount
expression that results in a parse error. In this particular case, the
Parse Error is thrown during expression evaluation and is expected to be
caught in the parse1 method and turned into a diagnostic.

But regardless of whether the exception is expected or not, when an
exception is thrown, it causes us to skip the groupIndexStack.pop() call
at the end of the function. This causes the state to be incorrect, and
can either cause unexpected parse behavior, or an error thrown at the
end of the parse when validating state is cleaned up correctly.

To fix this, we  wrap the body of the sequence parser in a try block,
and pop the groupIndexStack in the finally. This way, regardless of an
exception being thrown or now, we will pop the groupIndexStack and
ensure correct state.

DAFFODIL-2555

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
@@ -72,322 +72,326 @@ abstract class SequenceParserBase(
   override protected def parse(pstate: PState): Unit = {
     pstate.mpstate.groupIndexStack.push(1L)
 
-    val children = childParsers
+    try {
 
-    var scpIndex = 0
+      val children = childParsers
 
-    val limit = children.length
+      var scpIndex = 0
 
-    var resultOfTry: ParseAttemptStatus = ParseAttemptStatus.Uninitialized
+      val limit = children.length
 
-    val infosetIndexStart = pstate.infoset.asInstanceOf[DIComplex].childNodes.size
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -72,322 +72,326 @@ abstract class SequenceParserBase(
...
+        pstate.walker.walk()
+      }
 
-      // we've unblocked the unordered sequence, try walking to output
-      // everything we've created
-      pstate.walker.walk()
+      if (child ne null) child.finalChecks(pstate, resultOfTry, priorResultOfTry)
+      ()
+    } finally {
+      pstate.mpstate.groupIndexStack.pop()
```
stevedlawrence, line 393: This is mostly just indentation changes. If you click the gear at the top and click "Hide whitespace changes" you can see that the only thing that really changes is this new finally block.

tuxji (approved):
+1

Good to have a test covering this.  What does "ocke" stand for?

stevedlawrence:
`ocke` stands for `dfdl:occursCountKind="expression"`. I don't think we use that acronym in the codebase, but a handful of tests use it.

mbeckerle (approved):
+1

PR #625, opened 2021-08-20
[RFC] Remove the outputValueCalc CompiledExpression from ElementRuntimeData

stevedlawrence:
When checkAllTopLevel is true, it results in creating ElementRuntimeData
objets for every element, even those that aren't reachable from the root
element. And because the outputValcCalc CompiledExpression lives on the
ERD, it means all OVCs are forced to be compiled, and are required to be
valid within their local context.

However, some OVC expression are intentionally not valid within their
local context. They instead are expected to reach out of global
group/element that they are defined in, and access some other element
outside of that group/element where it is references. So
compiling/checking OVC expressions outside of the context of the root
element doesn't really make sense.

This solves this problem by moving the OVC CompiledExpression out of the
ERD, and accessing/compiling it only when an unparser is actually
created for that element. This means it will only be compiled and
checked in the correct context of where it is actually used.

DAFFODIL-2552

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
@@ -1199,7 +1199,7 @@ trait ElementBaseGrammarMixin
         inputValueCalcPrim, dfdlScopeEnd, EmptyGram)
     }
 
-  protected final lazy val ovcCompiledExpression = { // ovcValueCalcObject.expr
+  final lazy val ovcCompiledExpression = {
     val exprProp = outputValueCalcOption.asInstanceOf[Found]
     val exprText = exprProp.value
     val exprNamespaces = exprProp.location.namespaces
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
@@ -114,7 +114,8 @@ class ElementCombinator(
         uSetVars,
         eBeforeUnparser,
         eUnparser,
-        eAfterUnparser)
+        eAfterUnparser,
+        context.ovcCompiledExpression)
     } else if ((context.lengthKind _eq_ LengthKind.Explicit) ||
       (context.isSimpleType &&
         (context.lengthKind _eq_ LengthKind.Implicit) &&
--- a/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
@@ -17,20 +17,20 @@
 
 package org.apache.daffodil.runtime1
 
+import org.apache.daffodil.api.WarnID
+import org.apache.daffodil.dpath.NodeInfo
 import org.apache.daffodil.dsom.ChoiceTermBase
+import org.apache.daffodil.dsom.ElementBase
 import org.apache.daffodil.dsom.ExpressionCompilers
 import org.apache.daffodil.dsom.SequenceTermBase
 import org.apache.daffodil.dsom.Term
-import org.apache.daffodil.infoset.ChoiceBranchStartEvent
-import org.apache.daffodil.processors.ChoiceRuntimeData
-import org.apache.daffodil.infoset.ChoiceBranchEvent
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.infoset.ChoiceBranchEndEvent
-import org.apache.daffodil.api.WarnID
-import org.apache.daffodil.processors.ChoiceDispatchKeyEv
-import org.apache.daffodil.dpath.NodeInfo
-import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.grammar.Gram
+import org.apache.daffodil.infoset.ChoiceBranchEndEvent
+import org.apache.daffodil.infoset.ChoiceBranchEvent
+import org.apache.daffodil.infoset.ChoiceBranchStartEvent
+import org.apache.daffodil.processors.ChoiceDispatchKeyEv
+import org.apache.daffodil.processors.ChoiceRuntimeData
 import org.apache.daffodil.util.Delay
 
 trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
@@ -136,7 +136,7 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
               // because if so, we have a true ambiguity here.
               case sg: SequenceTermBase => {
                 val nonOVCEltChildren = sg.groupMembers.filter {
-                  case erd: ElementRuntimeData => erd.outputValueCalcExpr.isEmpty
+                  case eb: ElementBase => !eb.isOutputValueCalc
                   case _ => false
                 }
                 nonOVCEltChildren.length > 0
--- a/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
@@ -753,27 +753,49 @@ class TestRefMap {
             <xs:element name="f" type="xs:int" dfdl:inputValueCalc='{ ../../rr/e }' />
           </xs:sequence>
         </xs:group>
+        <xs:group name="x">
+          <xs:sequence>
+            <!--
+            likewise, this group is never used, and so should not create an error, even
+            if checkAllTopLevel is true
+            -->
+            <xs:element name="y" type="xs:int" dfdl:outputValueCalc='{ ../../dne }'/>
+            <xs:element name="z" type="xs:int" dfdl:inputValueCalc='{ ../../dne }' />
+          </xs:sequence>
+        </xs:group>
 
   /**
    * The root element doesn't have any substructure that eventually uses the
-   * group h, so we get the error that indicates h can't be a root (relative path upward)
-   * because we're checking all top level.
+   * group h, and so the IVC/OVC expressions within it which reach past the
+   * root are not validated. This is because such expressions potentially only
+   * make sense within the context of the root element. checkAllTopLevel does
+   * not mean that all elements that aren't used must make sense as root
+   * elements--even though "h" could be a root because it isn't referenced
+   * anywhere, that doesn't necessarily mean we should require that expressions
+   * inside it are not up-and-out expressions, as if it were a root.
    */
   @Test def testPathPastRootFromGlobalDef1(): Unit = {
     val compiler = Compiler().withCheckAllTopLevel(true)
-    val pf = compiler.compileNode(testRootPathsSchema1)
+    val pf = compiler.compileNode(testRootPathsSchema1, None, Some("r"))
     val isError = pf.isError
-    val msgs = pf.getDiagnostics.map(_.getMessage()).mkString("\n")
-    // println(msgs)
-    assertTrue(isError)
-    assertTrue(msgs.toLowerCase.contains("Relative path '../../rr/f' past root element".toLowerCase))
+    assertFalse(isError)
   }
 
   @Test def testPathPastRootFromGlobalDef2(): Unit = {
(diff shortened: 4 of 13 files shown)


stevedlawrence:
Note that I marked this as [RFC] because I'm not sure if this is the right fix. A comment I came across implies that the expected behavior is that if a global element is not referenced anywhere, then we should treat it as if it could be a root, and thus if it has any expressions that reach past that potential root then it is an error. If an element is referenced, then we don't place the same restrictions, and we allow things like expressions that go past the local root.

My guess (though not confirmed) is that the json schema where this bug was discovered has a unused global group defintion that contain elements with expression that went outside of that group def. And since the group wasn't used anywhere, we treated it like a potential root element (even though it's inside a group) and didn't allow those expressions. 

Thougg, one could argue that this isn't correct, and we shouldn't assume an element could be a root until it's actually used as a root. This change seems to have that behavior. I still think this is a good change to have, and if we want the mentioned behavior then we need some other change to treat unreferenced elements as roots, as right now I think it's only enforcing OVC correctness.

mbeckerle:
One comment is that much of this issue is about not having to compile schemas repeatedly just because there are a variety of root elements in them. That variety of root elements is usually about testing. E.g., for a messaging data format, it is very common to have a top-level element that is for a single messsage, as well as a top-level element for an unbounded vector of such elements, so that one can test the schema easily against individual messages (e.g., in a TDML file) or against files containing multiple messages. Top level elements for testing the major components of a message in isolation are also common.

It saves schema compilation time if parsers/unparsers are created for ALL such top level elements, particularly, if they share sub-parser/unparsers. This can be substantial time savings for a large DFDL schema during the dev/test/debug cycle. 

This need to have many available root elements for testing purposes is why I do NOT like the solution of putting a property on the DFDL schema declaring the specific unique root element. But I would not be opposed to such a property labeling each of all possible root elements, thereby allowing global element decls that are NOT intended to be root elements to be distinguished from those intended to be root elements. (Another way to do this would be to have a single top-level DFDL annotation that declares the QNames of all allowed root elements. We'd still need to gather these from all schema files however.)

stevedlawrence:
Is it even possible to take advantage of that right now though? For example, if you wanted to get the parser for two root elements in the same schema wouldn't you need to do something like this:
```scala
val pf1 = Compiler.compile("schema.dfdl.xsd", "root1")
val dp1 = pf1.onPath("/")
val pf2 = Compiler.compile("schema.dfdl.xsd", "root2")
val dp2 = pf1.onPath("/")
```
Isn't this going to compile the schema (and all possible root elements) twice, essentially doubling the work? I do get the point about validating everything, so I think this is reasonable behavior, but I'm not sure it saves schema compilation time. If anything wouldn't it make things worse? At least until we support onPath parameters other than root? I guess ideally we would have this, which probably would be pretty efficient:

```scala
val pf1 = Compiler.compile("schema.dfdl.xsd")
val dp1 = pf1.onPath("/root1")
val dp2 = pf1.onPath("/root2")
```

Also, the more I think about this change, I think it's the right change to have (the OVC expression really doesn't belong on and ERD). But I think maybe we have a bug somewhere else if the expected behavior is for these expressions on "possible" roots should cause an error. For example, inputValueCalc expressions on possible roots don't cause an error, and I didn't change anything related to those. I think previously we just got lucky in that we were forcing ERDs to be created, which forced the OVC to be compiled. So really, we only validated OVC expression on possible roots, and no other expressions.

So if we do want these expressions to cause an error, we likely need some other fix. I think it's still maybe worth merging this change, as it allows these schemas to compile? And we can fix the behavior some other way if we want this to cause an error?

tuxji (approved):
+1 (if you want it)

I've reviewed all the changes and they look good except for one questionable place which needs either a `!` operator or a clarifying comment.  I've read the discussion and I find Steve's comments that these changes are worth merging even if we need another fix persuasive.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -136,7 +136,7 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
               // because if so, we have a true ambiguity here.
               case sg: SequenceTermBase => {
                 val nonOVCEltChildren = sg.groupMembers.filter {
-                  case erd: ElementRuntimeData => erd.outputValueCalcExpr.isEmpty
+                  case erd: ElementRuntimeData => erd.dpathElementCompileInfo.isOutputValueCalc
```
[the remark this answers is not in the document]
stevedlawrence: Good catch. Yeah, I think this is just a bug. My guess is we just don't have a test that coveres this particular case. Which isn't too surprising. It's looking for some UPA violation that XML schema wouldn't normally find. I'm not sure off hand what that would be. I'll see if I can add a test that should hit this SDE but isn't and then make sure adding the `!` fixes it.

mbeckerle:
I think we *do not* want errors from global defs/decls with up-and-out paths. Ideally we want to recognize these, recognize that they can't be used as roots and label them as such, and them treat them as the reusable objects they are, verifying that the expressions make sense in every context where they are used.  Attempting to use such a global def/decl as a root should be an SDE since they are unsuitable for use as roots. But their mere existence is not an error. 

I think this change set will achieve this. 

The schema compilation time savings is a separate issue really.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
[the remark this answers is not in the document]
stevedlawrence: So it looks like this is just buggy code. The ``sg.groupMembers`` variable is of type `Seq[Term]`, but `ElementRuntimeData` is not a `Term`. Which means that this case is impossible to hit. Usually the scala compiler complains when a match case isn't possible, but for some reason it doesn't know in this case. Regardless, because this case is never hit, the missing `!` didn't affect any tests.

Additionally, because this case is wrong, we still SDE but not with the correct "UPA Violation" SDE. We do have a test that covers this UPA Violation error, but the expected error message matches both SDE's, so we never realized we had this regression (I'm not sure when this broke).

To fix this, I've changed this case to match on `ElementBase`, which *is* a `Term`, and updated the test to look for the expected "UPA Violation" message.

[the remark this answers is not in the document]
mbeckerle: Type erasure crap. ```Seq[Term]``` is treated like ```Seq[AnyRef]``` I guess. 

I suggest adding a type decl. E.g., 

```
sg.groupMembers.filter{ t: Term => t match {
... cases here ...
}
```
So that the type error is caught.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1199,7 +1199,7 @@ trait ElementBaseGrammarMixin
         inputValueCalcPrim, dfdlScopeEnd, EmptyGram)
     }
 
-  protected final lazy val ovcCompiledExpression = { // ovcValueCalcObject.expr
+  final lazy val ovcCompiledExpression = { // ovcValueCalcObject.expr
```
mbeckerle, line 1202: You can remove the comment at end of line here.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
```diff
@@ -114,7 +114,8 @@ class ElementCombinator(
         uSetVars,
         eBeforeUnparser,
         eUnparser,
-        eAfterUnparser)
+        eAfterUnparser,
+        context.ovcCompiledExpression)
```
mbeckerle, line 118: This is actually a good example of where something doesn't belong on the ERD, but should instead be a parameter to a parser/unparser. Because it is needed in exactly one place. 

This has the downside that elements with OVC won't be checked unless we specifically construct one for the grammar. I suppose we should demand the grammar object for every top level thing to force checking, but then only demand the parser/unparser for actual root elements.

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
```diff
@@ -753,29 +753,51 @@ class TestRefMap {
...
+            <xs:element name="y" type="xs:int" dfdl:outputValueCalc='{ ../../dne }'/>
+            <xs:element name="z" type="xs:int" dfdl:inputValueCalc='{ ../../dne }' />
+          </xs:sequence>
+        </xs:group>
 
   /**
    * The root element doesn't have any substructure that eventually uses the
-   * group h, so we get the error that indicates h can't be a root (relative path upward)
-   * because we're checking all top level.
+   * group h, and so the IVC/OVC expressions within it which reach passed the
```
mbeckerle, line 769: typo "passed" -> "past"

daffodil-core/src/test/scala/org/apache/daffodil/dsom/TestRefMap.scala
```diff
@@ -753,29 +753,51 @@ class TestRefMap {
...
+   * root are not validated. This is because such expressions potentially only
+   * make sense within the context of the root element. checkAllTopLevel does
+   * not mean that all elements that aren't used must be make sense as root
+   * elements--even though "h" could be a root, because it isn't referenced
+   * anywhere, that doesn't necessarily mean we should require that expressions
+   * inside it as if it were a root.
```
mbeckerle, line 775: phrasing of last line. I think this needs to say "that expressions inside it are not up-and-out expressions, as if it were a root."

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
```diff
@@ -136,7 +136,7 @@ trait ChoiceTermRuntime1Mixin { self: ChoiceTermBase =>
               // because if so, we have a true ambiguity here.
               case sg: SequenceTermBase => {
                 val nonOVCEltChildren = sg.groupMembers.filter {
-                  case erd: ElementRuntimeData => erd.outputValueCalcExpr.isEmpty
+                  case eb: ElementBase => !eb.isOutputValueCalc
```
mbeckerle, line 139: Suggest add type to filter test function: 
``` 
sg.groupMembers.filter{ t: Term => t match {
.... cases....
   }
}
```
extra braces, but worth it for the type checking.

mbeckerle (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ChoiceTermRuntime1Mixin.scala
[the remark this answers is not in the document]
stevedlawrence: Just tried this and scala stil doesn't complain about an ElementRuntimeData case.

I wonder if the issue is because Term is a trait, so we could maybe have something like this:

```scala
class Foo extends ElementRuntimeData with Term
```

So a `Foo` could be part of the `groupMembers` sequence, but it could also be an `ElementRuntimeData`? Maybe if `Term` were sealed scala would know there are no other `Term`s, but I'm not sure we can easily do that since things that mixin Term are spread all over the place?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/ElementCombinator.scala
stevedlawrence: > I suppose we should demand the grammar object for every top level thing

Do you want me to add that as part of this pull request? Does that essentially mean doing something like
```scala
globalElementDecls.foreach { _.asRoot }
```

Also, do we want to do it for all imported/included schemas, or only top level elements in the first/main schema?

mbeckerle: We don't have to do that now. If you do that now, then I think we'd still have the problem where some schemas need to be modified to turn off compileAllTopLevel.

stevedlawrence: Ok, I've created [DAFFODIL-2556](https://issues.apache.org/jira/browse/DAFFODIL-2556) to track that potential change.

PR #630, opened 2021-09-02
Resolve Apache Conscious Language Checker findings

stevedlawrence:
- Change references to Daffodil "main" branch
- Fix test quotes/strings and typos to avoid false positives
- Links to 3rd party repos are changed to 'main' where possible, or to a
  commit hash or tag where not, again to minimize false positives

Findings that are not resolved:

- Dependency license text that contains "his or her"--we probably
  shouldn't modify text from other licenses
- False positive found in "Saxon-HE" dependency

DAFFODIL-2557

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
@@ -562,10 +562,10 @@
     <xs:element name="e_lowercase2" type="xs:string" dfdl:inputValueCalc="{ fn:lower-case('TeSt LoWeR') }" />
     <xs:element name="e_lowercase3" type="xs:string" dfdl:inputValueCalc="{ fn:lower-case('') }" />
 
-    <xs:element name="e_contains1" type="xs:boolean" dfdl:inputValueCalc="{ fn:contains('The quick brown fox.', 'quick') }" />
-    <xs:element name="e_contains2" type="xs:boolean" dfdl:inputValueCalc="{ fn:contains('The quick brown fox.', 'slow') }" />
-    <xs:element name="e_contains3" type="xs:boolean" dfdl:inputValueCalc="{ fn:contains('The quick brown fox.', '') }" />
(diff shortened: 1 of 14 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -562,10 +562,10 @@
     <xs:element name="e_lowercase2" type="xs:string" dfdl:inputValueCalc="{ fn:lower-case('TeSt LoWeR') }" />
     <xs:element name="e_lowercase3" type="xs:string" dfdl:inputValueCalc="{ fn:lower-case('') }" />
 
-    <xs:element name="e_contains1" type="xs:boolean" dfdl:inputValueCalc="{ fn:contains('The quick brown fox.', 'quick') }" />
```
mbeckerle, line 565: This one surprised me. I have no objection to the change. I just wonder if I can be conscious enough about language to avoid things as common as this.

mbeckerle (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
stevedlawrence: Yeah, the tool definitely needs some refining to reduce false positives. It wouldn't surprise me if we continue to see them from time to time. But the fact that this codebase represents many years of development and there were only a handful of things that needed to be changed says this likely won't be something that will require many changes from our current development.

Note that in this particular case, the false positive was a a test that took off the `T` from `The `, which resulted in 'he'. Instead of changing just that one test, I decided to just change all tests using `the quick brown fox` so they were consistent. So this was pretty rare case, and could have been a much smaller change.

tuxji (approved):
+1

tuxji:
Should we amend this PR's commit message to include those instructions as well?

The default branch has been renamed! The `master` branch is now named `main`. 

If you have a local clone, you can update it by running:

```sh
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

stevedlawrence:
Seems reasonable. Will update with instructions.

PR #631, opened 2021-09-03
Fix a missed master to main rename

stevedlawrence:
Commit ba44472a3dfe369fd79442505f163d20d05d5ea5 renamed master to main
but missed an instance due to a grep that missed hidden files. This
caused sonarscan to not be run. Fixed the grep and found only these two
instances.

DAFFODIL-2557

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@
 
 name: CI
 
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #636, opened 2021-09-16
Fix broken tests related to property scoping

stevedlawrence:
The unparse_property_scoping_02 test was added to show that property
scoping with nested define formats works as expected. But it was added
before unparsing was fully implemented, and had a small issue that meant
it never actually worked.

The issue is that the "s1-u" schema does not define the textNumberPattern
property so it inherits the value from DFDLGeneralFormat.dfdl.xsd,
which is a pattern that includes commas. This means the value `-1009`
unparses to `-1,009`, which is one character too long for the explicit
length of 5 and leads to a truncation error.

The solution is to just specify a textNumberPattern of "#####" so there
are no commas and it unparses to the expected length. But this makes the
schema exactly the same as the "s1" schema. So this just removes the
"s1-u" model, and changes the test to use the "s1" model. The test
passes, confirming that nested defineFormat references work as expected.

This also fixes the test_localAnnotation_05 commented out test in the
same file. This test is broken because the expected infoset isn't
ordered to match the schema order of an unordered sequence. The fix is
to just arrange the expected infoset to schema order and change it
two-pass.

Also adds missing `: Unit = ` return types on tests in this same file.

DAFFODIL-2103

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section08/property_scoping/PropertyScoping.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section08/property_scoping/PropertyScoping.tdml
@@ -579,16 +579,16 @@
 
   <tdml:parserTestCase name="localAnnotation_05"
     description="DFDL Property Scoping Rules - DFDL-8-010R"
-    model="localAnnotationSchema" root="e5">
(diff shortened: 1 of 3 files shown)


tuxji (approved):
+1

Looks just fine.

mbeckerle (approved):
+1

PR #640, opened 2021-09-21
Fix handling of %#x25; character entity

stevedlawrence:
We sort of have a two-pass system for handling DFDL strings that contain
character entities related to delimiters. The first past converts
character entities to the character values, excluding things like
escaped percents and charater classes like NL, WSP, etc. The second pass
handles those excluded entities.

However, this first pass replaces the hex entity %#x25; (and
corresponding decimal entity) to a percent sign, which can then
interfere with the second pass, which should never see single percent
signs except for the exceptions mentioned above.

This modifies the first pass so that character entities that result in a
single percent are replaced with a double-percent. This will then be
converted to single precent used escaped-percent logic elsewhere.

This also has refactors related existing code to be more robust and
reduce duplication.

DAFFODIL-2479

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
@@ -128,6 +128,25 @@ final class EntityReplacer {
   def hasHexCodePoint(input: String): Boolean = isMatched(input, hexPattern)
   def hasByteCodePoint(input: String): Boolean = isMatched(input, bytePattern)
 
+  private def replaceEntityWithChar(input: String, entity: String, newChar: Char): String = {
+    val replacement =
+      if (newChar == '%') {
+        // Some character entities are not replaced in this EntityReplacer,
+        // such as double percents or character classes (NL, WSP, etc.). If
+        // this character entity results in a percent character (e.g. %#x25;),
+        // we must replace it with an escaped percent to be handled later.
+        "%%"
+      } else {
+        // This character might mean something special to the replaceAll method
+        // we are about to use (e.g. a dollar sign for regex group references).
+        // To be safe, call quoteReplaement which escapes any characters that
+        // mean something special in a replacement string so they are replaced
+        // with the literal value.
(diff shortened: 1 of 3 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -128,6 +128,25 @@ final class EntityReplacer {
...
+        // this character entity results in a percent character (e.g. %#x25;),
+        // we must replace it with an escaped percent to be handled later.
+        "%%"
+      } else {
+        // This character might mean something special to the replaceAll method
+        // we are about to use (e.g. a dollar sign for regex group references).
+        // To be safe, call quoteReplaement which escapes any characters that
+        // mean something special in a replacement string so they are replaced
+        // with the literal value.
+        Matcher.quoteReplacement(newChar.toString)
```
mbeckerle, line 145: I had to look this up. 

So replaceString actually interprets various characters in the replacement text, as having significance regex-matcher-wise. I.e., you can put back-refs to capture groups in the replacement text, and they are supposed to get substituted?

Bizarre. I'm sure it came in handy somewhere, but it's very weird behavior to me. The regex syntax and any significant regex characters should *only* be interpreted in the argument that IS a regex, to me. Not in the input, and not in the replacement. 

Suggests to me I've never used replaceAll properly ever, and all calls to it in Daffodil should be revisited for this same sort of error. It would seem to me I need to always use str.replaceAll(regex, Matcher.quoteReplacement(repl)) every time to get what I expect.

mbeckerle (approved):
+1

stevedlawrence:
I've looked at all all uses for `replaceAll`, and I don't see any uses that could contain a back slash or a dollar sign. All instances use a hardcoded replacement values, or are now properly escaped with this change.

The one exception was for code that used to handle `dfdl:outputNewLine`, which could theoretically contain a dollar sign, but that is dead code. That property is now handled without `replaceAll`. I've just removed that function to remove the possibility of every using it with bad input.

PR #641, opened 2021-09-22
Improve validation and diagnostics when reloading saved parsers

stevedlawrence:
When saving a parser, write a NUL-terminated ASCII string of the form
`DAFFODIL <VERSION>` directly to the output stream prior to DataProcessor
serialization.

A benefit to writing directly to the head of the OutputStream is that
this avoids the gzip compression and Java serialization in the header,
making it simple to read and validate for compatibility. This also
allows a user to run `head -1z` on a saved parser to quickly determine
what version of Daffodil a saved parser is compatible with.

When reloading a saved parser, we check that the file starts with the
string `DAFFODIL `, and the following data up to a NUL-terminator
matches the current version of Daffodil. If they are the same, we assume
we are reloading with the same version of Daffodil used to save the
parser, and so they should be compatible.

However, we can't really control if a user swaps out incompatible
dependencies between save/reload, and we also don't really know which
versions of libraries are compatible. So this also improves detection
and diagnostics when a reload error occurs that most likely due to
dependency incompatibility. So if there are reload errors, it should
make the cause and resolution more clear.

DAFFODIL-2266

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -19,6 +19,7 @@ package org.apache.daffodil.compiler
 
 import java.io.File
 import java.io.FileInputStream
+import java.io.InvalidClassException
 import java.io.ObjectInputStream
 import java.io.StreamCorruptedException
 import java.nio.channels.Channels
@@ -26,6 +27,7 @@ import java.util.zip.GZIPInputStream
 import java.util.zip.ZipException
 
 import scala.collection.immutable.Queue
+import scala.collection.mutable.ArrayBuffer
 import scala.util.Try
 import scala.xml.Node
 
@@ -283,7 +285,34 @@ class Compiler private (var validateDFDLSchemas: Boolean,
 
   def reload(savedParser: java.nio.channels.ReadableByteChannel): DFDL.DataProcessor = {
     try {
-      val objInput = new ObjectInputStream(new GZIPInputStream(Channels.newInputStream(savedParser))) {
+      val is = Channels.newInputStream(savedParser)
+
+      // Read the required prefix and version information for this saved parser
+      // directly from the input stream. This information is not compressed or
+      // serialized by a Java ObjectOutputStream
+
+      val requiredDataPrefix = "DAFFODIL "
+      requiredDataPrefix.foreach { c =>
+        if (is.read() != c.toInt) throw new InvalidParserException("The saved parser is only compatible with an older version of Daffodil")
+      }
+
+      val ab = new ArrayBuffer[Byte]()
+      var byte = -1
+      while ({ byte = is.read(); byte > 0 }) {
+        ab.append(byte.toByte)
+      }
+      if (byte == -1) {
+        throw new InvalidParserException("The saved parser is corrupted")
+      }
+      val curVersion = Misc.getDaffodilVersion
+      val savedVersion = new String(ab.toArray, "utf-8")
+      if (savedVersion != curVersion) {
+        throw new InvalidParserException("The saved parser is only compatible with Daffodil " + savedVersion + ". Current version is " + curVersion)
+      }
+
+      // Decompress and deserilize the rest of the file using java object deserializtion
+
+      val objInput = new ObjectInputStream(new GZIPInputStream(is)) {
 
         ///
         /// This override is here because of a bug in sbt where the wrong class loader is being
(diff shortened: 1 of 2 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -283,7 +284,31 @@ class Compiler private (var validateDFDLSchemas: Boolean,
 
   def reload(savedParser: java.nio.channels.ReadableByteChannel): DFDL.DataProcessor = {
     try {
-      val objInput = new ObjectInputStream(new GZIPInputStream(Channels.newInputStream(savedParser))) {
+      val is = Channels.newInputStream(savedParser)
+
+      // Read the magic number and version information for this saved parser
+      // directly from the input stream. That information is not compressed or
+      // serialized
+      val magicNumber = "DAFFODIL "
```
mbeckerle, line 292: calling this "magic number" is just confusing. It's not a number, and it's not like magic numbers that are just a specific 2 bytes at start of a file. 

Call it requiredDataPrefixString or something clearer.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -283,7 +284,31 @@ class Compiler private (var validateDFDLSchemas: Boolean,
...
+        if (is.read() != c.toInt) throw new InvalidParserException("The saved parser is only compatible with an older version of Daffodil")
+      }
+
+      val sb = new StringBuilder()
+      var byte: Int = -1
+      while ({ byte = is.read(); byte > 0 }) {
+        sb.append(byte.toChar)
+      }
+      if (byte == -1) {
+        throw new InvalidParserException("The saved parser file is not the correct format")
```
mbeckerle, line 303: This is worse than "not the correct format" it means the file is corrupted, ends early because at the point where we should get a NUL we instead get EOF.  At this point it has a proper prefix string, but doesn't have a proper version number and NUL separating it from the rest. This could say exactly that. Saved parser file has proper prefix 'DAFFODIL" but otherwise is not properly formed.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -283,7 +284,31 @@ class Compiler private (var validateDFDLSchemas: Boolean,
...
+      }
+      if (sb.toString != Misc.getDaffodilVersion) {
+        throw new InvalidParserException("The saved parser is only compatible with Daffodil " + sb.toString)
```
mbeckerle, line 306: Add the current daffodil version that you are attempting to reload into here, to make the message more useful.

mbeckerle (approved):
+1 Code appears to be correct. I suggested minor improvements to make the messages a little more specific.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
stevedlawrence: I guess I was avoiding giving too much information that doesn't really help anything but to confuse a user. No one should be manually modifying  these files, so telling someone specifically what is wrong doesn't really help them fix the problem. If anything, they'll wonder what it means to be missing a prefix and try to fix that problem. The problem is really that the file was corrupted or something (we don't really know). Maybe instead of "not correct format' we sould say "looks to be corrupted"? Though, I'm not sure if that gives any extra information aside from guessing what cause the issue.

stevedlawrence: Note, we also used this same "not the correct format" string below if we get a `ZipException`. And we use "not a valid parser" if there is a `StreamCorruptedException`. However we decided to describe these errors we should be consistent.

mbeckerle: Yeah, I think something like "corrupted" is fine.

stevedlawrence: Ok, I've used "corrupted" and cleaned up the other error messages to improve consistency

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -357,7 +357,14 @@ class DataProcessor private (
 
   def save(output: DFDL.Output): Unit = {
 
-    val oos = new ObjectOutputStream(new GZIPOutputStream(Channels.newOutputStream(output)))
+    val os = Channels.newOutputStream(output)
+
+    // write a null-terminated ASCII string as a simple version identifier
+    val headerString = "DAFFODIL " + Misc.getDaffodilVersion + "\u0000"
+    os.write(headerString.getBytes("US-ASCII"))
```
tuxji, line 364: I reviewed how we call `getBytes` in Daffodil in order to check for inconsistencies and best practices.  I noticed two things: 1) we call `getBytes("ascii")` instead every other place where we want bytes from ASCII characters; and 2) we call `getBytes` without a charset name too many times.  Java's platform default charset is specific to the user and OS.  On many modern Linux systems, it's UTF-8. On Macs, it’s MacRoman. In the US on Windows, it's often CP1250, while in Europe it's CP1252 or in China it's often simplified Chinese (Big5 or a GB*).  I'm agnostic whether we use "ascii", "US-ASCII", or import java.nio.charset.StandardCharsets and use StandardCharsets.US_ASCII (I see Daffodil typically uses all-lowercase strings most often to save space and typing), but we probably should create a bug to replace all parameter-less `getBytes` calls with `getBytes("utf-8")`.

tuxji (approved):
+1

Noticed Daffodil calls `getBytes` without a charset name parameter too many times, but that's a separate issue.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
stevedlawrence: Agreed. It looks like most uses (maybe all?) of `getBytes` without an encoding are in tests, so I"m not sure there's a bug in Daffodil proper. But it does mean our tests are possibly fragile to the system default encoding which we definitely don't want. And improving is definitely worth it.

I've created [DAFFODIL-2561](https://issues.apache.org/jira/browse/DAFFODIL-2561) for this. I'll also update this to use "utf-8" encoding for conistency.

PR #642, opened 2021-09-23
Added test case for DAFFODIL-2562

jadams-tresys:
Added a relatively simple test case demonstrating the bug.

DAFFODIL-2562


stevedlawrence (approved):
+1

mbeckerle (approved):
+1

PR #643, opened 2021-09-24
layering extension to support CRC, checksum, parity

mbeckerle:
Layering feature evolved slightly to support CRC, checksum, parity feature.

Test is based on ipv4 with a layer transform.

This PR starts with 2 commits. First commit is just removing deprecated layering properties that were in the dfdl namespace, and accepted with no prefix like DFDL official properties. Now we are enforcing that one must write dfdlx:layerTransform with the proper extension prefix. 

DAFFODIL-2221

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
@@ -22,12 +22,13 @@ import org.apache.daffodil.schema.annotation.props.gen.LayerLengthKind
 import org.apache.daffodil.schema.annotation.props.gen.LayerLengthUnits
 import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.MaybeInt
-import org.apache.daffodil.processors.LayerLengthInBytesEv
+import org.apache.daffodil.processors.LayerLengthEv
 import org.apache.daffodil.processors.LayerBoundaryMarkEv
 import org.apache.daffodil.processors.LayerCharsetEv
 import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.processors.charset.BitsCharsetAISPayloadArmoring
+
 import java.nio._
 import java.nio.charset._
 import java.io._
@@ -44,7 +45,7 @@ import org.apache.daffodil.schema.annotation.props.gen.EncodingErrorPolicy
 import org.apache.daffodil.schema.annotation.props.gen.UTF16Width
 import org.apache.daffodil.processors.charset.BitsCharsetDecoder
 import org.apache.daffodil.processors.charset.BitsCharsetEncoder
-import org.apache.daffodil.dsom.DPathCompileInfo
+import org.apache.daffodil.processors.SequenceRuntimeData
 
 object AISPayloadArmoringTransformer {
   val iso8859 = StandardCharsets.ISO_8859_1
@@ -164,10 +165,10 @@ object AISPayloadArmoringTransformerFactory
   override def newInstance(
     maybeLayerCharsetEv: Maybe[LayerCharsetEv],
     maybeLayerLengthKind: Maybe[LayerLengthKind],
-    maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
+    maybeLayerLengthEv: Maybe[LayerLengthEv],
     maybeLayerLengthUnits: Maybe[LayerLengthUnits],
     maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
-    tci: DPathCompileInfo): LayerTransformer = {
+    srd: SequenceRuntimeData) = {
 
     val xformer = new AISPayloadArmoringTransformer()
     xformer
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
@@ -17,11 +17,13 @@
 
 package org.apache.daffodil.layers
 
+import org.apache.daffodil.api.ThinDiagnostic
 import org.apache.daffodil.schema.annotation.props.gen.LayerLengthKind
 import org.apache.daffodil.schema.annotation.props.gen.LayerLengthUnits
-import org.apache.daffodil.processors.LayerLengthInBytesEv
+import org.apache.daffodil.processors.LayerLengthEv
 import org.apache.daffodil.processors.LayerBoundaryMarkEv
 import org.apache.daffodil.processors.LayerCharsetEv
+
 import java.util.HashMap
 import org.apache.daffodil.processors.ParseOrUnparseState
 import org.apache.daffodil.util.NonAllocatingMap
@@ -39,7 +41,10 @@ import org.apache.daffodil.processors.parsers.PState
 import org.apache.daffodil.processors.unparsers.UState
 import org.apache.daffodil.io.DirectOrBufferedDataOutputStream
 import passera.unsigned.ULong
-import org.apache.daffodil.dsom.DPathCompileInfo
+import org.apache.daffodil.processors.SequenceRuntimeData
+
+import java.io.InputStream
+import java.io.OutputStream
 
 /**
  * Factory for a layer transformer.
@@ -56,10 +61,10 @@ abstract class LayerTransformerFactory(nom: String)
   def newInstance(
     maybeLayerCharsetEv: Maybe[LayerCharsetEv],
     maybeLayerLengthKind: Maybe[LayerLengthKind],
-    maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
+    maybeLayerLengthEv: Maybe[LayerLengthEv],
     maybeLayerLengthUnits: Maybe[LayerLengthUnits],
     maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
-    tci: DPathCompileInfo): LayerTransformer
+    srd: SequenceRuntimeData): LayerTransformer
 }
 
 /**
@@ -101,29 +106,38 @@ object LayerTransformerFactory {
   register(GZIPTransformerFactory)
   register(IMFLineFoldedTransformerFactory)
   register(ICalendarLineFoldedTransformerFactory)
+
+  /**
+   * Arguably the above could be bundled with Daffodil.
+   *
+   * The transformers below really should be plugins defined
+   * outside Daffodil, per DAFFODIL-1927.
+   */
   register(AISPayloadArmoringTransformerFactory)
   register(FourByteSwapTransformerFactory)
 }
 
 /**
  * Shared functionality of all LayerTransformers.
+ *
+ * A layer transformer is created at runtime as part of a single parse/unparse call.
+ * Hence, they can be stateful without causing thread-safety issues.
  */
-abstract class LayerTransformer()
-  extends Serializable {
+abstract class LayerTransformer() {
 
-  protected def wrapLayerDecoder(jis: java.io.InputStream): java.io.InputStream
+  protected def wrapLayerDecoder(jis: InputStream): InputStream
 
-  protected def wrapLimitingStream(jis: java.io.InputStream, state: PState): java.io.InputStream
+  protected def wrapLimitingStream(jis: InputStream, state: PState): InputStream
 
-  def wrapJavaInputStream(s: InputSourceDataInputStream, fInfo: FormatInfo): java.io.InputStream = {
+  def wrapJavaInputStream(s: InputSourceDataInputStream, fInfo: FormatInfo): InputStream = {
     new JavaIOInputStream(s, fInfo)
   }
 
-  protected def wrapLayerEncoder(jos: java.io.OutputStream): java.io.OutputStream
+  protected def wrapLayerEncoder(jos: OutputStream): OutputStream
 
-  protected def wrapLimitingStream(jis: java.io.OutputStream, state: UState): java.io.OutputStream
+  protected def wrapLimitingStream(jis: OutputStream, state: UState): OutputStream
 
-  def wrapJavaOutputStream(s: DataOutputStream, fInfo: FormatInfo): java.io.OutputStream = {
+  def wrapJavaOutputStream(s: DataOutputStream, fInfo: FormatInfo): OutputStream = {
     new JavaIOOutputStream(s, fInfo)
   }
   /**
@@ -148,7 +162,7 @@ abstract class LayerTransformer()
     // nothing for now
   }
 
-  def addLayer(s: DataOutputStream, state: UState): DataOutputStream = {
+  def addLayer(s: DataOutputStream, state: UState): DirectOrBufferedDataOutputStream = {
     val jos = wrapJavaOutputStream(s, state)
     val limitedJOS = wrapLimitingStream(jos, state)
     val encodedOutputStream = wrapLayerEncoder(limitedJOS)
@@ -173,37 +187,33 @@ abstract class LayerTransformer()
     //
   }
 
-  // These were very useful for debugging. Note that they stop with a ???
-  // error. That's because dumping streams changes their state.
-  //
-  // Keeping these around commented out, for now. While this feature is still
-  // new and may need further debugging.
-  //
-  //  def dumpLayer(is: InputStream) {
-  //    val str = Stream.continually(is.read).takeWhile(_ != -1).map(_.toChar).mkString
-  //    println("dump length " + str.length + " = " + str)
-  //    ???
-  //  }
-  //
-  //  def hexDumpLayer(is: InputStream) {
-  //    val str = hexify(is)
-  //    println("hex dump = " + str)
-  //    ???
-  //  }
-  //
-  //  def hexify(is: InputStream): String =
-  //    Stream.continually(is.read).takeWhile(_ != -1).map(x => "%02x".format(x.toInt)).mkString(" ")
-  //
-  //  def hexDumpLayer(is: InputSourceDataInputStream, finfo: FormatInfo) {
-  //    val jis = new JavaIOInputStream(is, finfo)
-  //    val str = hexify(jis)
-  //    println("DIS hex dump = " + str)
-  //    ???
-  //  }
+  /**
+   * Enables the layer transformer to take necessary state-maniuplating actions before
+   * the parsing of a layer starts.
+   *
+   * Examples might be setting variable values.
+   *
+   * At the time this is called, the state contains the layered I/O stream already.
+   * @param s The parser state, which may be modified.
+   */
+  def startLayerForParse(s: PState):Unit = () // nothing by default
+
+
+  /**
+   * Enables the layer transformer to take necessary state-maniuplating actions after
+   * the unparsing of a layer ends.
+   *
+   * Note that if variables without values are read by this method, then this must create
+   * suspended calculations that clone the necessary state, which are evaluated later.
+   *
+   * Examples might be setting variable values.
+   * @param s The unparser state, which may be modified.
+   */
+  def endLayerForUnparse(s: UState): Unit = () // nothing by default
 }
 
 class JavaIOInputStream(s: InputSourceDataInputStream, finfo: FormatInfo)
-  extends java.io.InputStream {
+  extends InputStream {
 
   private lazy val id = Misc.getNameFromClass(this)
 
@@ -239,7 +249,7 @@ class JavaIOInputStream(s: InputSourceDataInputStream, finfo: FormatInfo)
 }
 
 class JavaIOOutputStream(dos: DataOutputStream, finfo: FormatInfo)
-  extends java.io.OutputStream {
+  extends OutputStream {
 
   private var closed = false
 
@@ -259,3 +269,17 @@ class JavaIOOutputStream(dos: DataOutputStream, finfo: FormatInfo)
     }
   }
 }
+
+abstract class LayerException(
+  layerSRD: SequenceRuntimeData,
+  state: PState,
+  maybeCause: Maybe[Throwable],
+  maybeFormatString: Maybe[String],
+  args: Any*)
+extends ThinDiagnostic(One(layerSRD.dpathCompileInfo.schemaFileLocation), One(state.currentLocation), maybeCause, maybeFormatString, args: _*)
+
+case class LayerNotEnoughDataException(layerSRD: SequenceRuntimeData, state: PState, cause: Throwable, nBytesRequired: Int)
+  extends LayerException(layerSRD, state, One(cause), None, nBytesRequired) {
+  override def isError = true
+  override def modeName = "Parse"
+  }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
@@ -42,7 +42,7 @@ final class LayerEncodingEv(override val expr: CompiledExpression[String], tci: 
 final class LayerCharsetEv(layerEncodingEv: LayerEncodingEv, override val ci: DPathCompileInfo)
   extends CharsetEvBase(layerEncodingEv, ci)
 
-final class LayerLengthInBytesEv(override val expr: CompiledExpression[JLong], override val ci: DPathCompileInfo)
+final class LayerLengthEv(override val expr: CompiledExpression[JLong], override val ci: DPathCompileInfo)
   extends EvaluatableExpression[JLong](
     expr,
     ci)
@@ -70,16 +70,16 @@ final class LayerTransformerEv(
   layerTransformEv: LayerTransformEv,
   maybeLayerCharsetEv: Maybe[LayerCharsetEv],
   maybeLayerLengthKind: Maybe[LayerLengthKind],
-  maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
+  maybeLayerLengthEv: Maybe[LayerLengthEv],
   maybeLayerLengthUnits: Maybe[LayerLengthUnits],
   maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
-  ci: DPathCompileInfo)
-  extends Evaluatable[LayerTransformer](ci)
+  srd: SequenceRuntimeData)
+  extends Evaluatable[LayerTransformer](srd.dpathCompileInfo)
   with NoCacheEvaluatable[LayerTransformer] {
 
   override lazy val runtimeDependencies = layerTransformEv +:
     (maybeLayerCharsetEv.toList ++
-      maybeLayerLengthInBytesEv.toList ++
+      maybeLayerLengthEv.toList ++
       maybeLayerBoundaryMarkEv.toList)
 
   /**
@@ -92,10 +92,10 @@ final class LayerTransformerEv(
     val xformer = factory.newInstance(
       maybeLayerCharsetEv,
       maybeLayerLengthKind,
-      maybeLayerLengthInBytesEv,
+      maybeLayerLengthEv,
       maybeLayerLengthUnits,
       maybeLayerBoundaryMarkEv,
-      ci)
+      srd)
     xformer
   }
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
@@ -63,6 +63,7 @@ import org.apache.daffodil.util.Poolable
 import org.apache.daffodil.infoset.DIComplexState
 import org.apache.daffodil.infoset.DISimpleState
 import org.apache.daffodil.exceptions.Abort
+import org.apache.daffodil.exceptions.ThrowsSDE
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.xml.GlobalQName
 
@@ -327,9 +328,14 @@ final class PState private (
     this.infoset = newParent
   }
 
-  def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: VariableRuntimeData): Unit = {
+  def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: ThrowsSDE): Unit = {
     variableMap.setVariable(vrd, newValue, referringContext, this)
     changedVariablesStack.top += vrd.globalQName
+  }
+
+  def getVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive = {
+    val res = variableMap.readVariable(vrd, referringContext, this)
+    res
   }
 
   /**
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
@@ -33,6 +33,7 @@ import org.apache.daffodil.infoset.DIArray
 import org.apache.daffodil.infoset.DIDocument
 import org.apache.daffodil.infoset.DIElement
 import org.apache.daffodil.infoset.DINode
+import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.infoset.InfosetAccessor
 import org.apache.daffodil.infoset.InfosetInputter
 import org.apache.daffodil.io.DirectOrBufferedDataOutputStream
@@ -50,6 +51,7 @@ import org.apache.daffodil.processors.TermRuntimeData
 import org.apache.daffodil.processors.UnparseResult
 import org.apache.daffodil.processors.VariableBox
 import org.apache.daffodil.processors.VariableMap
+import org.apache.daffodil.processors.VariableRuntimeData
 import org.apache.daffodil.processors.charset.BitsCharset
 import org.apache.daffodil.processors.charset.BitsCharsetDecoder
 import org.apache.daffodil.processors.charset.BitsCharsetEncoder
@@ -73,6 +75,21 @@ abstract class UState(
   areDebugging: Boolean)
   extends ParseOrUnparseState(vbox, diagnosticsArg, dataProcArg, tunable)
   with Cursor[InfosetAccessor] with ThrowsSDE with SavesErrorsAndWarnings {
+
+  final def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: ThrowsSDE) =
+    vbox.vmap.setVariable(vrd, newValue, referringContext, this)
+
+  /**
+   * For unparsing, this throws a RetryableException in the case where the variable cannot (yet) be read.
+   *
+   * @param vrd Identifies the variable to read.
+   * @param referringContext Where to place blame if there is an error.
+   * @return The data value of the variable, or throws exceptions if there is no value.
+   */
+  final def getVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive =
+    vbox.vmap.readVariable(vrd, referringContext, this)
+
+
 
   /**
    * Push onto the dynamic TRD context stack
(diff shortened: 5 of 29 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/AISTransformer.scala
```diff
@@ -77,6 +78,11 @@ class AISPayloadArmoringTransformer()
     val newJOS = new LayerBoundaryMarkInsertingJavaOutputStream(jos, layerBoundaryMark, iso8859)
     newJOS
   }
+
+  def startLayerForParse(s: PState): Unit = () // do nothing
```
mbeckerle, line 82: These should all have "override" keyword on them. 

I believe so far we have only needed the start flavor for parse, and the end flavor for unparse, so end for parse, and start for unparse could be removed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -200,6 +214,54 @@ abstract class LayerTransformer()
   //    println("DIS hex dump = " + str)
   //    ???
   //  }
```
mbeckerle, line 216: Remove commented out code if no longer needed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/IPv4Checksum.scala
```diff
@@ -0,0 +1,403 @@
...
+import org.apache.daffodil.util.Maybe.One
+import org.apache.daffodil.xml.NS
+import org.apache.daffodil.xml.RefQName
+import passera.unsigned.UShort
+
+import java.nio.ByteBuffer
+import java.nio.ShortBuffer
+
+/**
+      The first layer transform, aka 'part1' computes the partial
```
mbeckerle, line 48: Investigate if this can be simplified to just one layer, instead of two.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/IPv4Checksum.scala
```diff
@@ -0,0 +1,403 @@
...
+
+    checksumTotalShort.toInt
+  }
+
+  object Part1 { def apply() = new Part1() }
+
+  final class Part1 private ()
+    extends LayerTransformer() {
+
+    private val hdrPart1ByteArr = new Array[Byte](ipHdrPart1LengthInBytes)
```
mbeckerle, line 164: Add scaladoc explaining these members as their roles and the relationship between them is subtle.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/IPv4Checksum.scala
```diff
@@ -0,0 +1,403 @@
...
+      maybeLayerCharsetEv: Maybe[LayerCharsetEv],
+      maybeLayerLengthKind: Maybe[LayerLengthKind],
+      maybeLayerLengthInBytesEv: Maybe[LayerLengthInBytesEv],
+      maybeLayerLengthUnits: Maybe[LayerLengthUnits],
+      maybeLayerBoundaryMarkEv: Maybe[LayerBoundaryMarkEv],
+      srd: SequenceRuntimeData): Unit = {
+      srd.schemaDefinitionUnless(maybeLayerLengthKind.isEmpty || maybeLayerLengthKind.get == LayerLengthKind.Implicit,
+        "Must have dfdlx:layerLengthKind undefined or 'implicit', but was '%s'.", maybeLayerLengthKind.get.toString)
+      //
+      // ToDo: We can't issue SDW because we don't have a context object
```
mbeckerle, line 359: Not fixing as part of this change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
```diff
@@ -89,13 +100,7 @@ final class LayerTransformerEv(
...
     val layerTransform = layerTransformEv.evaluate(state)
     val factory = LayerTransformerFactory.find(layerTransform, state)
-    val xformer = factory.newInstance(
-      maybeLayerCharsetEv,
-      maybeLayerLengthKind,
-      maybeLayerLengthInBytesEv,
-      maybeLayerLengthUnits,
-      maybeLayerBoundaryMarkEv,
-      ci)
+    val xformer = factory.newInstance(maybeLayerCharsetEv, maybeLayerLengthKind, maybeLayerLengthInBytesEv, maybeLayerLengthUnits, maybeLayerBoundaryMarkEv, srd)
```
mbeckerle, line 103: too long line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -74,6 +76,14 @@ abstract class UState(
   extends ParseOrUnparseState(vbox, diagnosticsArg, dataProcArg, tunable)
   with Cursor[InfosetAccessor] with ThrowsSDE with SavesErrorsAndWarnings {
 
+  final def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: ThrowsSDE) =
+    vbox.vmap.setVariable(vrd, newValue, referringContext, this)
+
+  final def getVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive =
```
mbeckerle, line 82: Add scaladoc explaining throw of RetryableException if the variable can't "yet" be read within a suspendableOperation.

daffodil-test/src/test/resources/org/apache/daffodil/layers/ais.tdml
```diff
@@ -38,6 +38,9 @@ required to pad the data payload to a 6-bit boundary. So it is 0 to 5.
 The *5C is a NMEA 0183 data integrity checksum, preceded by "*" - computed on 
 the whole message including the AIVDM, but excluding the "!"
 
+According to the wikipedia article on NMEA 0183, this checksum is just an XOR of the
+specific message bytes.
+
 Seems another role for layers is computing checksums.
```
mbeckerle, line 44: Park a JIRA ticket for updating this AIS layer to implement the checksum. 

Fix comment: remove word "Seems".

daffodil-test/src/test/resources/org/apache/daffodil/layers/ipv4.dfdl.xsd
```diff
@@ -0,0 +1,261 @@
...
+    Modified with proposed checksum computation via layer transforms
+
+    The xmlns:checksum prefix declaration would probably go on the xs:schema element more normally.
+    It is just here to show all the required things in this one location of the file.
+    -->
+    <xs:sequence>
+      <xs:annotation>
+        <xs:appinfo source="http://www.ogf.org/dfdl/">
+          <!--
+          The checksum field in an ipV4 actually lives in the middle of the
```
mbeckerle, line 97: Revisit if this can be done with just 1 layer, not two.

daffodil-test/src/test/resources/org/apache/daffodil/layers/ipv4.dfdl.xsd
```diff
@@ -0,0 +1,261 @@
...
+        will be a runtime SDE.
+
+        Everything from here to the end of this sequence is unmodified from
+        the original IPV4 schema it is based on.
+        -->
+        <xs:element name="Version" type="tns:bit" dfdl:length="4"/>
+        <xs:element name="IHL" type="tns:bit" dfdl:length="4"/>
+        <xs:element name="DSCP" type="tns:bit" dfdl:length="6"/>
+        <xs:element name="ECN" type="tns:bit" dfdl:length="2"/>
+        <xs:element name="Length" type="tns:bit" dfdl:length="16"/>
```
mbeckerle, line 170: Enhance this schema so that length depends on an element later after the checksum layer. 
This will exercise the ability to suspend while populating this part1 layer.

daffodil-test/src/test/resources/org/apache/daffodil/layers/ipv4.dfdl.xsd
```diff
@@ -0,0 +1,261 @@
...
+
+      This checksum calculation knows exactly how big this data must be
+      and if the representation turned out to be a different length than this, it
+      will be a runtime SDE.
+
+      The result is placed in the checksumValue variable.
+      -->
+      <xs:sequence dfdl:ref="chksum:ipv4ChecksumLayerPart2">
+      <xs:sequence>
+        <xs:element name="IPSrc" type="tns:hexByte" dfdl:length="4"/>
```
mbeckerle, line 200: enhance so at least one of these does a forward reference to a field later, so as to illustrate suspending during unparse during this part 2 layer transform.

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
```diff
@@ -0,0 +1,205 @@
...
+    layerStream
+  }
+
+  protected def wrapLayerEncoder(jos: OutputStream) = jos
+
+  protected def wrapLimitingStream(jos: OutputStream, state: UState) = {
+    originalOutputStream = jos
+    val layerLengthInBytes: Int = layerLengthEv.evaluate(state).toInt
+    val ba = new Array[Byte](layerLengthInBytes)
+    val byteBuf = ByteBuffer.wrap(ba) // bigEndian byte order by default
```
mbeckerle, line 95: Fix comment. We're not using byte order in this example.

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * Shared by both parsing and unparsing.
+   *
+   * Ignores any non-digit character in the argument.
+   *
+   * The checkDigit is the total of all digits, viewed as a string, the last digit of that total.
+   */
+  private def computeCheckDigit(s: String): Long = {
```
mbeckerle, line 107: Explain why this is returning a Long. Because a Long is the type expected to be the representation of unsignedInt which is the variable type.

daffodil-test/src/test/resources/org/apache/daffodil/layers/xsd/checkDigitLayer.dfdl.xsd
```diff
@@ -0,0 +1,51 @@
...
+
+  <include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+
+      <dfdl:format ref="cd:GeneralFormat" />
+
+      <dfdl:defineVariable name="checkDigit" type="xs:unsignedInt"/>
+
```
mbeckerle, line 36: move checkDigitType to here. 
Add facets for min 0 max 9.

daffodil-test/src/test/resources/org/apache/daffodil/layers/checkDigit.dfdl.xsd
```diff
@@ -0,0 +1,95 @@
...
+              dfdl:calendarPatternKind="explicit"
+              dfdl:calendarPattern="yyyy-MM-dd">
+    <restriction base="xs:date"/>
+  </simpleType>
+
+  <simpleType name="u5" dfdl:length="5">
+    <restriction base="tns:fixedInt"/>
+  </simpleType>
+
+  <simpleType name="u1" dfdl:length="1">
```
mbeckerle, line 86: change to checkDigitType and move to other xsd file. 
Add facets for min 0 max 9. 
change base to just unsignedInt. Consider changing type to xs:int so that code can use type Int. (not type Long)

daffodil-test/src/test/resources/org/apache/daffodil/layers/checkDigit.dfdl.xsd
```diff
@@ -0,0 +1,95 @@
...
+    <restriction base="tns:fixedInt"/>
+  </simpleType>
+
+  <simpleType name="fixedInt" dfdl:lengthKind="explicit"
```
mbeckerle, line 90: remove unused type

daffodil-test/src/test/resources/org/apache/daffodil/layers/checkDigit.dfdl.xsd
```diff
@@ -0,0 +1,95 @@
...
+
+  <simpleType name="dateType"
+              dfdl:lengthKind="explicit"
+              dfdl:length="10"
+              dfdl:calendarPatternKind="explicit"
+              dfdl:calendarPattern="yyyy-MM-dd">
+    <restriction base="xs:date"/>
+  </simpleType>
+
+  <simpleType name="u5" dfdl:length="5">
```
mbeckerle, line 82: remove unused type

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
```diff
@@ -70,16 +70,27 @@ final class LayerTransformerEv(
...
+  /**
+   * A layer transformer is never a constant. It always must be evaluated to get
+   * a new one.
+   *
+   * This prevents it being cached/stored as part of the state of the data processor
+   * where it would be shared across threads.
+   *
+   * Otherwise layer transformers couldn't be stateful. But they often need to be stateful.
+   */
+  override protected def isNeverConstant = true
```
stevedlawrence, line 89: Could/should this logic be part of `NoCacheEvaluatable`? Or are these different concepts? Seems like if something shouldn't be cached then it also should never be considered constant? It would be nice if mixing in the right things gave the right behavior without having to set other flags.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
```diff
@@ -89,13 +100,7 @@ final class LayerTransformerEv(
...
     val layerTransform = layerTransformEv.evaluate(state)
     val factory = LayerTransformerFactory.find(layerTransform, state)
-    val xformer = factory.newInstance(
-      maybeLayerCharsetEv,
-      maybeLayerLengthKind,
-      maybeLayerLengthInBytesEv,
-      maybeLayerLengthUnits,
-      maybeLayerBoundaryMarkEv,
-      ci)
+    val xformer = factory.newInstance(maybeLayerCharsetEv, maybeLayerLengthKind, maybeLayerLengthEv, maybeLayerLengthUnits, maybeLayerBoundaryMarkEv, srd)
```
stevedlawrence, line 103: Line too long, rewrap to previous change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
```diff
@@ -327,11 +328,17 @@ final class PState private (
...
 
-  def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: VariableRuntimeData): Unit = {
+  def setVariable(vrd: VariableRuntimeData, newValue: DataValuePrimitive, referringContext: ThrowsSDE): Unit = {
     variableMap.setVariable(vrd, newValue, referringContext, this)
     changedVariablesStack.top += vrd.globalQName
   }
 
+  def getVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive = {
+    val res = variableMap.readVariable(vrd, referringContext, this)
+    changedVariablesStack.top += vrd.globalQName
```
stevedlawrence, line 338: This has no coverage, should our variable code be using this getVariable so that the stack is adjusted correctly, similar to setVariable? Seems inconsistent, and maybe a bug, if setVariable and getVariable are access in different ways?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -74,6 +76,21 @@ abstract class UState(
...
+    vbox.vmap.setVariable(vrd, newValue, referringContext, this)
+
+  /**
+   * For unparsing, this throws a RetryableException in the case where the variable cannot (yet) be read.
+   *
+   * @param vrd Identifies the variable to read.
+   * @param referringContext Where to place blame if there is an error.
+   * @return The data value of the variable, or throws exceptions if there is no value.
+   */
+  final def getVariable(vrd: VariableRuntimeData, referringContext: ThrowsSDE): DataValuePrimitive =
```
stevedlawrence, line 89: Same comment as above, should we be using these set/getVariable functions internally? I assume right now we are just directly access this variable map?

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
```diff
@@ -0,0 +1,205 @@
...
+    }
+    // this stream will be used by the parse when it parses the
+    // schema contents of the layer. We have already consumed the bytes
+    // from the original input stream, and saved those in the byte array
+    // to enable the computation of the checkDigit.
+    val layerStream = new ByteArrayInputStream(ba)
+    val str = new String(ba, charset) // always does replacement on decode errors.
+    val checkDigit = computeCheckDigit(str)
+    state.setVariable(checkDigitVRD, checkDigit, srd) // assign to result variable
+    layerStream
```
stevedlawrence, line 86: Is this standard for layers, to read the entire data into memory, do something with it, and then return it as a new stream that parse then reads? Or is that just done for simplicity because this is a test implementation so not a big deal, or we expect the size of these layers to be small and not be a memory issue?

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
```diff
@@ -0,0 +1,205 @@
...
+    //
+    // ToDo: We can't issue SDW because we don't have a context object
+    // for that sort of compilation time warning. We *should* have that.
+    // That requires a layer factory method called at schema compile time with the
+    // Sequence term (not just term runtime data) as an argument.
+    //
+    val xformer = new CheckDigitExplicit(srd, javaCharset, maybeLayerLengthEv.get)
+    xformer
+  }
+}
```
stevedlawrence, line 205: I'm interesting to see how this all changes with plugability. There's a lot of things being used here that I'd consider internal and wouldn't necessarily want a user to directly access. Not an issue for this PR, though.

stevedlawrence (approved):
+1 looks nice, seems there are very few actual changes, just some minor comments

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -101,15 +102,28 @@ object LayerTransformerFactory {
...
+  register(IPv4Checksum.Factories.Part2)
+
 }
 
 /**
  * Shared functionality of all LayerTransformers.
+ *
+ * A layer transformer is serialized as part of the processor definition;
+ * hence, they must be stateless as we allow the processor definition to be
+ * shared across multiple threads.
```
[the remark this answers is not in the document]
mbeckerle: Will fix. This comment reflects a misunderstanding I had. You are correct. The factory is serialized, the transformer itself is created at runtime.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -200,6 +214,54 @@ abstract class LayerTransformer()
   //    println("DIS hex dump = " + str)
   //    ???
   //  }
+
+  /**
+   * Enables the layer transformer to take necessary state-maniuplating actions before
+   * the parsing of a layer starts.
```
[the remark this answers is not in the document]
mbeckerle: A facade makes sense. As I write more of these layers however, I've concluded a much higher-level pattern can be imposed also which may make exposing the P/Ustate objects unnecessary. E.g., we need the QName of a variable to assign, and the algorithm that takes a byte array and returns an int.Those really are the only needed parameters for checksums/CRC/parity.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -173,37 +187,33 @@ abstract class LayerTransformer()
...
+   *
+   * Examples might be setting variable values.
+   *
+   * At the time this is called, the state contains the layered I/O stream already.
+   * @param s The parser state, which may be modified.
+   */
+  def startLayerForParse(s: PState):Unit = () // nothing by default
```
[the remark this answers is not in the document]
mbeckerle: I think it just works that you only need a start function for the parse, to capture the layer data before regular parsing parses it. You don't have to compute the checksum at that point, but I see no reason not to. 

For unparsing, it's the opposite. You need to first unparse and capture the outbound data, and only after that can you recompute the checksum. 

I originally created start/end calls for each of parse/unparse, then discovered that I didn't need end for parse, nor start for unparse.

daffodil-runtime1/src/main/scala/org/apache/daffodil/layers/LayerTransformer.scala
```diff
@@ -200,6 +214,54 @@ abstract class LayerTransformer()
...
+   * @param s The unparser state, which may be modified.
+   */
+  def startLayerForUnparse(s: UState): Unit
+
+  /**
+   * Enables the layer transformer to take necessary state-maniuplating actions after
+   * the unparsing of a layer ends.
+   *
+   * Note that if variables without values are read by this method, then this must create
+   * suspended calculations that clone the necessary state, which are evaluated later.
```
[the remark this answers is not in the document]
mbeckerle: This is a very good point. I think the plugin API really needs to abstract away this stuff so that a framework that creates suspensions is being used, and drives the plug in methods.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvLayering.scala
mbeckerle: Good point. I'll arrange for NoCacheExecutable to have that behavior, update its scaladoc accordingly, and remove this flag.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala
mbeckerle: This was needed in the first version of IPv4 which had two transforms where the first one assigned a variable that had to be read (gotten) by the second. 

When I simplified IPv4 to a single transform, this was no longer needed. And check-digits doesn't need it either.

Neither one reads any variable. Just writes them. 

But there's nothing stopping a layer transform from wanting to read a variable, e.g, to get a parameter to the algorithm, e.g, a mode flag like compression algorithm to use, or variant of line-folding to use, etc. 

So I think we just need an example that exercises this call.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: I will actually look at how expressions that read variables do it. They are obviously not calling this getter, but perhaps could which would solve the test coverage.

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
mbeckerle: If the checksum was over a large thing, like a whole file, then one would want to use a layered stream that does the checksumming as part of the parser reading the data stream. That would avoid the need to buffer stuff. That's what for example the gzip and line-folding layers do. 

In this case this is IP packets, by definition small. So just grabbing it into a buffer seemed simplest.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
```diff
@@ -0,0 +1,199 @@
...
+                 dfdl:lengthKind="explicit"
+                 dfdl:lengthUnits="bits">
+    <xs:restriction base="xs:unsignedInt"/>
+  </xs:simpleType>
+
+  <xs:simpleType name="hexByte" dfdl:lengthKind="explicit" dfdl:lengthUnits="bytes">
+    <xs:restriction base="xs:hexBinary"/>
+  </xs:simpleType>
+
+  <xs:simpleType name="uint16" dfdl:lengthKind="explicit" dfdl:length="16">
```
tuxji, line 187: It's possible that your format still may be defining `dfdl:lengthUnits="bytes"` at this point which means your `uint16` and `uint32` types may be 16/32 bytes long, not 16/32 bits long.  For completeness and safety, you should have `lengthUnits="bits"` in your basicBigEndianBinary definition above.  Or consider removing all the types except `bits` and `hexByte` since those are the only types your schema is using right now anyway.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
```diff
@@ -0,0 +1,199 @@
...
+    <xs:restriction base="xs:unsignedInt"/>
+  </xs:simpleType>
+
+  <xs:simpleType name="uint32" dfdl:lengthKind="explicit" dfdl:length="32">
+    <xs:restriction base="xs:unsignedInt"/>
+  </xs:simpleType>
+
+  <xs:simpleType name="int32" dfdl:lengthKind="explicit" dfdl:length="4" dfdl:lengthUnits="bytes">
```
tuxji, line 195: I would expect to read 32 bits, not 4 bytes, in an `int32` definition, but you've made this definition (almost) explicit enough to be portable to any schema.  Actually, the funny thing is that the base type `xs:int` is also 32 bits long so the only restrictions you really need are `dfdl:lengthKind="explicit"` and `dfdl:representation="binary"`.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.tdml
```diff
@@ -0,0 +1,202 @@
...
+-->
+<tdml:testSuite
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:eth="urn:ethernet"
+  xmlns:chksum="urn:com.owlcyberdefense.IPv4Checksum"
```
tuxji, line 26: I saw `"urn:org.apache.daffodil.layers.IPv4Checksum"` in the previous file.  It would be nice if both files were consistent with each other.  Also note you may want to replace `urn:ethernet` too.

daffodil-test/src/test/resources/org/apache/daffodil/layers/xsd/IPv4ChecksumLayer.dfdl.xsd
```diff
@@ -0,0 +1,51 @@
...
+      <dfdl:defineVariable name="IPv4Checksum" type="xs:unsignedShort"/>
+      <dfdl:defineFormat name="IPv4ChecksumLayer">
+        <dfdl:format dfdlx:layerTransform='{ "IPv4Checksum" }' dfdlx:layerLengthKind="implicit"/>
+      </dfdl:defineFormat>
+    </xs:appinfo>
+  </xs:annotation>
+
+  <xs:simpleType name="IPv4Checksum"
+    dfdl:lengthKind="explicit" dfdl:length="16" dfdl:lengthUnits="bits">
+    <xs:restriction base="xs:unsignedShort"/>
```
tuxji, line 48: For curiosity, what happens if you define or import a simple type restricting `base="xs:unsignedShort"` with `dfdl:length="16"` and `dfdl:lengthUnits="bits"` in a text schema with `dfdl:representation="text"`?  Does Daffodil automatically override the text representation when parsing/unparsing that type?

daffodil-test/src/test/scala/org/apache/daffodil/layers/CheckDigit.scala
tuxji: Oh, that's right, an uint32 won't fit in an Int so it needs to be returned as a Long (or passera.unsigned.UInt).

daffodil-test/src/test/scala/org/apache/daffodil/layers/IPv4Checksum.scala
```diff
@@ -0,0 +1,239 @@
...
+      srd: SequenceRuntimeData) = {
+      srd.schemaDefinitionUnless(maybeLayerLengthKind.isEmpty ||
+        maybeLayerLengthKind.get == LayerLengthKind.Implicit,
+        "Must have dfdlx:layerLengthKind undefined or 'implicit', but was '%s'.",
+        maybeLayerLengthKind.get.toString)
+      //
+      // ToDo: We can't issue SDW because we don't have a context object
+      // for that sort of compilation time warning. We *should* have that.
+      // That requires a layer factory method called at schema compile time with the
+      // Sequence term (not just term runtime data) as an argument.
```
tuxji, line 229: Is this comment out of date?  I see SequenceRuntimeData being passed above, so it looks like you've already made that change.  If so, please delete the above comment.

daffodil-test/src/test/scala/org/apache/daffodil/layers/IPv4Checksum.scala
```diff
@@ -0,0 +1,239 @@
...
+      //
+      // We would like to warn if maybeLayerLengthUnits is defined here and is not bytes
+      // We would like to warn if maybeLayerBoundaryMarkEv is defined (since it is unused by this layer)
+      // We would like to warn if maybeLayerCharsetEv is defined (since it is unused by this layer)
+      // We would like to warn if maybeLayerLengthEv is defined (since it is unused by this layer)
```
tuxji, line 234: May be worth keeping these comments (or acting on some of them).

tuxji (approved):
+1

With some comments below.

daffodil-test/src/test/scala/org/apache/daffodil/layers/IPv4Checksum.scala
mbeckerle: An SRD isn't enough to issue schema def warnings. Warnings are accumulated at compile time, so have to be issued on a compiler object such as the LocalSequence object. An SRD is enough to issue a runtime SDE, it would be good if we got some of these also at schema compile time. 

I'm leaving fixing this off to the other ticket to refactor these layers to they are pluggable.

daffodil-test/src/test/resources/org/apache/daffodil/layers/xsd/IPv4ChecksumLayer.dfdl.xsd
mbeckerle: I think you will get a text integer, and if the charset is say, ascii, you'll get a 2-byte long (2 ascii chars) integer since that's what would fit in 16 bits.  Seems weird, but there are data specs where they tell you how long every field is, in bits, even for the text strings, purely for uniformity reasons in the spec.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
```diff
@@ -0,0 +1,199 @@
...
+      <dfdl:defineFormat name="basicBigEndianBinary">
+        <dfdl:format
+          ref="tns:GeneralFormat"
+          representation="binary"
+          binaryNumberRep="binary"
+          byteOrder="bigEndian"
+          bitOrder="mostSignificantBitFirst"
+          alignmentUnits="bits"
+          alignment="1"
+          occursCountKind="implicit"/>
```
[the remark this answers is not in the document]
mbeckerle: I like specifying the properties the format cares about, and using GeneralFormat to take care of things the format doesn't care about mike making sure separator="" etc. 

So prefixIncludesPrefixLength is only checked if you use lengthKind="prefixed" and since we're not using that, we don't need the property. That's also why GeneralFormat doesn't have it.  The DFDL spec actually says that you're not supposed to need prefixIncludesPrefixLength unless you have lengthKind prefixed.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
mbeckerle: I will specify lengthUnits bits, and remove the types the example isn't using.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
```diff
@@ -0,0 +1,199 @@
...
+           xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+           xmlns:chksum="urn:org.apache.daffodil.layers.IPv4Checksum"
+           xmlns:eth="urn:ethernet"
+           xmlns:tns="urn:ethernet"
+           targetNamespace="urn:ethernet">
+
+  <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+
+  <xs:import namespace="urn:org.apache.daffodil.layers.IPv4Checksum"
+             schemaLocation="org/apache/daffodil/layers/xsd/IPv4ChecksumLayer.dfdl.xsd"/>
```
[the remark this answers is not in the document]
mbeckerle: Originally, I thought the IPv4 checksum schema would be introducing the variable decls, and such needed to use the layer, and so would not be coupled to a specific implementation of IPv4 packets. 

But I'm thinking of these layer transforms as specific to a particular DFDL schema now. So I'll change to use the same namespace.

daffodil-test/src/test/resources/org/apache/daffodil/layers/IPv4.dfdl.xsd
```diff
@@ -0,0 +1,199 @@
...
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+           xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+           xmlns:fn="http://www.w3.org/2005/xpath-functions"
+           xmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"
+           xmlns:chksum="urn:org.apache.daffodil.layers.IPv4Checksum"
+           xmlns:eth="urn:ethernet"
+           xmlns:tns="urn:ethernet"
+           targetNamespace="urn:ethernet">
```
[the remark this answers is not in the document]
mbeckerle: Historic reason only. I ripped this out of the ethernet-ip schema which is on github, which itself got ripped out of the PCAP schema.

I will make the namespaces more sensible, and uniform.

mbeckerle (commented):
Some replies

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: Created DAFFODIL-2565 for this issue. VRef in DPath was not maintaining the changedVariablesStack.

daffodil-test/src/test/scala/org/apache/daffodil/layers/IPv4Checksum.scala
mbeckerle: Actually, I take it back. You can issue schema definition warnings on a runtime data object. But we really want compilation time errors, not runtime errors.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
mbeckerle: Turns out, there was no bug here. VariableMap.readVariable was, in fact, doing a callback to PState to maintain the changedVariablesStack, so this code was broken because it was adding variables twice. 

I changed DAFFODIL-2565 into an improvement cleanup, as it is cleaner if all variable manipulation happens via the PState. Ideally the variableMap would be a private pstate member, but the debugger currently has access to it.

PR #646, opened 2021-09-27
Remove deprecated layering properties in dfdl: namespace.

mbeckerle:
Broke this change out of other PR since this is really a separate issue.

Removed support for non-dfdlx prefixed layerTransform properties. 

Also removed tests of these.

Only dfdlx:layerTransform and other layering properties are recognized now.

DAFFODIL-2564


stevedlawrence:
+1, does it make sense to add a label like "deprecation" to the Jira issue so we can easily search for this kinds of changes?

stevedlawrence (approved):

mbeckerle:
> +1, does it make sense to add a label like "deprecation" or "backwards compatibility" to the Jira issue so we can easily search for this kinds of changes when building the release notes?

Done. Added deprecation and backward-compatibility labels.

jw3 (approved):
:+1:

PR #650, opened 2021-10-01
Extend C code generator to support hexBinary elements

tuxji:
Allow Daffodil to generate C code from DFDL schemas using the type
"hexBinary" for fixed length and prefixed length opaque byte arrays.
The "hex" simply means that the binary 8-bit bytes are represented by
hexadecimal characters in the XML infoset and the C code knows how to
convert between these binary and text representations when parsing
binary bytes or unparsing XML infosets.  Add hexBinary elements to the
ex_nums test schema to exercise the new code.

cli_errors.[ch]: Add new error messages CLI_HEXBINARY_LENGTH,
CLI_HEXBINARY_PARSE, CLI_HEXBINARY_SIZE to report any problems
converting between hexBinary binary and text representations.

xml_reader.c: Add hexToBinary helper function and call it in
xmlNumberElem when an ERD has typeCode PRIMITIVE_HEXBINARY.  Rename
xmlNumberElem to xmlSimpleElem since it handles more than just numbers
now.  Includes get iwyu update because hexToBinary calls free, malloc,
and memset functions not previously called before.

xml_writer.c: Add binaryToHex helper function and call it in
xmlNumberElem when an ERD has typeCode PRIMITIVE_HEXBINARY.  Rename
xmlNumberElem to xmlSimpleElem since it handles more than just numbers
now.  Includes get iwyu update because binaryToHex calls free and
malloc functions not previously called before.

errors.[c,h]: Add new error message ERR_HEXBINARY_ALLOC to report any
problem allocating memory for prefixed length hexBinary elements.

infoset.[c,h]: Add PRIMITIVE_HEXBINARY to TypeCode enumerations and
call handler->visitNumberElem on it too.  Rename [Vv]isitNumberElem to
[Vv]isitSimpleElem since it handles more than just numbers now.  Add
HexBinary struct to encapsulate opaque byte arrays.  Includes get iwyu
update because HexBinary uses bool and uint8_t types not previously
imported before.

parsers.[ch]: Add new functions parse_alloc to allocate memory for
prefixed length hexBinary elements and parse_hexBinary to read
hexBinary elements into memory.  Give read_stream_update_position
macro new parameters to allow parse_hexBinary to use it.  Includes get
iwyu update because parse_alloc uses free, malloc, and
ERR_HEXBINARY_ALLOC not previously imported before.

unparsers.[ch]: Add new function unparse_hexBinary to write binary
8-bit bytes from hexBinary elements.  Give
write_stream_update_position macro new parameters to allow
unparse_hexBinary to use it.

Runtime2CodeGenerator.scala: Make object Runtime2CodeGenerator extend
HexBinaryCodeGenerator and call either
hexBinaryLengthPrefixedGenerateCode or
hexBinarySpecifiedLengthGenerateCode when it sees a
HexBinaryLengthPrefixed or HexBinarySpecifiedLength grammar object.

BinaryIntegerKnownLengthCodeGenerator.scala: Make some C compilation
warnings go away by changing the initial value used to initialize C
code integer elements from 0xCC to 0x77 (0xCC.. will change sign in a
signed integer but 0x77.. will not change from postive to negative).

CodeGeneratorState.scala: Modify getPrimType, addSimpleTypeERD,
addFieldDeclaration, generateCodeHeader, & generateCodeFile to
generate C code for PrimType.HexBinary elements.

HexBinaryCodeGenerator.scala: Add generator functions to generate C
code for initializing, parsing, and unparsing hexBinary elements.

ex_nums.dfdl.xsd: Add new simple types and hexBinary elements to
exercise the new C code.

ex_nums_parse.dat, ex_nums_unparse_errors.xml,
ex_nums_unparse_runtime[12].xml: Add data and infoset elements for new
hexBinary elements.

examples/ex_nums/generated_code.[ch]: Update to show generated code
which initializes, parses, and unparses new hexBinary elements.

DAFFODIL-2566


daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/HexBinaryCodeGenerator.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  def hexBinarySpecifiedLengthGenerateCode(g: HexBinarySpecifiedLength, cgState: CodeGeneratorState): Unit = {
+    // For the time being this is a very limited back end.
+    // So there are some restrictions to enforce.
+    val e = g.e
+    e.schemaDefinitionUnless(e.bitOrder eq BitOrder.MostSignificantBitFirst, "Only dfdl:bitOrder 'mostSignificantBitFirst' is supported.")
+    e.schemaDefinitionUnless(e.elementLengthInBitsEv.isConstant, "Runtime dfdl:length expressions not supported.")
+
+    val fieldName = e.namedQName.local
+    val arraySize = if (e.occursCountKind == OccursCountKind.Fixed) e.maxOccurs else 0
```
stevedlawrence, line 106: You might consider a different check for `arraySize`? It seems like we support either fixed length arrays or scalars. Maybe something like this instead:
```scala
val arraySize = e.occursCountKind match {
  case OccursCountKind.Fixed if e.maxOccurs > 1 => e.maxOccurs
  case OccursCountKind.Fixed if e.maxOccurs == 1 => 0
  case OccursCountKind.Implicit if e.minOccurs == 1 && e.maxOccurs == 1 => 0
  case _ => SDE("unsupported occursCountKind ...")
}
```

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/HexBinaryCodeGenerator.scala
```diff
@@ -0,0 +1,146 @@
...
+             |    if (ustate->error) return;""".stripMargin
+        cgState.addSimpleTypeStatements(init2, parse2, unparse2)
+      }
+    }
+    if (arraySize > 0)
+      for (i <- 0 until arraySize)
+        addStatements(s"[$i]")
+    else
+      addStatements("")
+  }
```
stevedlawrence, line 145: Seems like much of this should be the same as the prefixedLength case, except there's extra generated code for the prefixed length? Can things be refactor to avoid duplciate code?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums_unparse_runtime2.xml
```diff
@@ -62,11 +67,14 @@
     <le_uint64>64</le_uint64>
     <le_uint8>8</le_uint8>
     <le_nonNegativeInteger8>8</le_nonNegativeInteger8>
+    <hexBinary7>1A2B3C4D5E6F70</hexBinary7>
+    <hexBinaryPrefixed>DEADBEEF</hexBinaryPrefixed>
```
stevedlawrence, line 71: Are there any tests with larger hex binary sizes? It seems like there's some special cases for sizes less than 256 that might not get triggered with only small tests?

stevedlawrence (approved):
+1, looks good, just some minro comments/suggestions

mbeckerle (approved):
This change has a lot of implications for how text strings will work in the future.

PR #654, opened 2021-10-07
Use conistent logic between SAXInfosetOutputter and XMLTextInfosetOutputter

stevedlawrence:
The SAXInfosetOutputter and XMLTextInfosetOutputter used slightly
different logic to figure out namespace mappings to add to an element.
In complex cases with lots of includes, shared references, and
overlapping namespace bindings, this could cause inconsistent namespace
mappings, which causes failures in the TDMLRunner, which mandates
consistency between SAX and non-SAX. The SAX logic should be fine, but
our broken minimizedScope (DAFODIL-2282) leads to issues. By mimicking
the XMLTextInosetOututter logic (i.e. value equality check +
shadowRedefined), we avoid those issues.

Also, fix XML diff reporting of different prefixes and namespace
mappings. The changes are:
- Add the label to the path, otherwise we just output the elements
  parent element and it isn't clear where the issue is
- Add @prefix or @xmlns to indicate the problem is related to
  different prefixes or different namespace mappings
- For namespace reporting, only compare/report the mappings on the
  particular element rather than the entire namespace binding. The
  easiest way to do this is to keep track of the parent scope, use
  NamespaceBinding.buildString(...) to create a string, and compare
  that string.

DAFFODIL-2568

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
@@ -931,6 +931,8 @@ Differences were (path, expected, actual):
     computeDiffOne(
       a,
       b,
+      TopScope,
+      TopScope,
       None,
       Nil,
       ignoreProcInstr,
@@ -952,6 +954,8 @@ Differences were (path, expected, actual):
   def computeDiffOne(
     an: Node,
     bn: Node,
+    aParentScope: NamespaceBinding,
+    bParentScope: NamespaceBinding,
     maybeIndex: Option[Int],
     parentPathSteps: Seq[String],
     ignoreProcInstr: Boolean,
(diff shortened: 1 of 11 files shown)


stevedlawrence:
Sorry, I never addressed this comment:

> The new test cases (complexIncludesNamespaces_01 and complexIncludesNamespaces_02) don't have any comments. I don't quite understand what they're supposed to demonstrate. Are they supposed to demonstrate that the SAX outputter no longer adds an unnecessary (that is, redundant) xmlns attribute, e.g., <raw xmlns:tns="urn:redundant-namespace">1</raw>?

Yep, those schemas are heavily minimized/anonyized version of a much more complicated schema with lots of complex includes and namespace mappings. It's kindof a worse case of someone not following what we now know to be best practices when defining xml namespaces. This complexity, and how we output prefix mappings in SAX, led to an edge case where we added redundant namespace mappings, as you mention. Normally this wouldn't matter since they are redundant, but it meant that SAX infosets and non-SAX infosets weren't exactly the same, and the TDML runner currently enforces that, which caused a failure.

I'll add a comment to the tdml file as Mike suggests to clarify things.

PR #673, opened 2021-11-03
Raise minimum C level to ISO C11 with GNU extensions

tuxji:
Fix pedantic warning "ISO C99 doesn't support unnamed structs/unions"
found when compiling nested.dfdl.xsd's NestedUnion example with
current Makefile's CFLAGS.  Installed centos7 vagrant box and verified
that both gcc and clang support "-Wall Wextra -Wpedantic -std=gnu11"
all the way back to gcc-4 and clang-3 on CentOS 7.  Therefore, will
raise runtime2's minimum C language support level to ISO C11 with GNU
extensions and make all the places where Daffodil calls the C compiler
pass -std=gnu11 to the C compiler to compile code with the same
minimum C level.  ISO C11 also allows "for (" loops to define loop
variables in place so we can make that change too.

build.sbt - Change cFlags to use "-Wpendantic -std=gnu11" instead of
"-pedantic -std=gnu99".

Makefile - Change CFLAGS to use "-Wpedantic -std=gnu11" instead of
"-pedantic -std=gnu99".  Slightly improve Makefile comments' step by
step instructions as well.

daffodil_getopt.c, xml_reader.c, xml_writer.c, errors.c, infoset.c -
Define loop variables in place within "for (" loops.

CodeGenerator.scala - Pass "-std=gnu11" to C compiler.  Now all places
that call the C compiler will compile with the same minimum C level.

NestedUnion/generated_code.[ch] - Replace with re-generated code.
These files are manually updated and their only purpose is to display
how the generated code changes as changes are made to runtime2 over
time.

DAFFODIL-2578

### DIFF
--- a/daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/Makefile
+++ b/daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/Makefile
@@ -14,23 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Step 0: You will need to install the Mini-XML library and xmldiff.
-# Here's how to install both packages on Ubuntu 20.04 (first-time
-# setup only):
+# Step 0: Install the Mini-XML library and xmldiff (first time setup
+# only).  Here's the command to install both packages on Ubuntu.
 
 # $ sudo apt install libmxml-dev xmldiff
 
-# Step 1: Copy your test data files here and either rename them to
-# parse.dat and unparse.xml or set PARSE_DAT and UNPARSE_XML.
-
-# $ cp ../ex_nums_parse.dat parse.dat
-# $ cp ../ex_nums_unparse_runtime2.xml unparse.xml
-
-PARSE_DAT = parse.dat
-UNPARSE_XML = unparse.xml
-
-# Step 2: Compile the C source files into an executable program which
-# can run the parse and unparse checks (e.g., .dat <-> .xml).
+# Step 1: Compile the C source files into an executable program which
+# can run the parse and unparse checks (e.g., .dat <-> .xml).  Just
+# run make with no arguments unless you want to override CC or CFLAGS.
 
 # $ make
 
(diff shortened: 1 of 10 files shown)


daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/examples/NestedUnion/generated_code.c
```diff
@@ -365,7 +368,7 @@ data_parseSelf(data *instance, PState *pstate)
         break;
     default:
         // Should never happen because initChoice would return an error first
-        error.d64 = (int64_t)instance->_choice;
+        error.arg.d64 = (int64_t)instance->_choice;
```
mbeckerle, line 371: Hmmm. I'm not seeing a change to a struct such that there is now an "arg" field you have to navigate. Did I just miss it?

mbeckerle (approved):
+1. Looks fine.

This C language/library stuff has a big Y2K bug: 11 > 99 because 11 => 2011, and 99 => 1999. Gaaaaaak. :-)

mbeckerle (commented):
The check I see that failed is due to a race-condition in one of the tests I wrote where I assumed 100msec would be a long enough wait time.... but it seems sometimes it isn't if the test machine is perhaps heavily loaded.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/examples/NestedUnion/generated_code.c
tuxji: I changed that struct in a C portability fix commit that went in last May...

<https://github.com/apache/daffodil/commit/8c95b39f5e365ae15657bcdc4686b44689b1fbb9#diff-6df97fc1bdf91db52173653361b5a40e5e075b80312a69912aa23e141dad57c6>

However, these "example" generated code files aren't updated automatically, so I didn't re-generate this example file until now.

stevedlawrence (approved):
+1

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/Makefile
```diff
@@ -14,56 +14,56 @@
...
-# Step 3: Run the executable on the test data files and check that the
-# new temp data files match the original test data files.
+# Step 2: Copy your test files here and rename them to parse.dat and
+# unparse.xml or else edit PARSE_DAT and UNPARSE_XML below.
+
+# $ cp ../ex_nums_parse.dat parse.dat
+# $ cp ../ex_nums_unparse_runtime2.xml unparse.xml
+
+PARSE_DAT = parse.dat
+UNPARSE_XML = unparse.xml
```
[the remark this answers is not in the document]
tuxji: Thanks for the tip.  Will remember when creating new data/xml files or changing old data/xml files.

PR #675, opened 2021-11-04
Remove duplicate prefix mapping check in SAX infosets

stevedlawrence:
When converting SAX events to text using our built-in ContentHandler, we
intentionally do not output namespace prefix mappings if a prefix has
already been mapped to the same namespace. This differs from the
XMLTextInfosetOutputter behavior, which does not do this duplicate
removal. So for example, the XMLTextInfosetOutputter might output an
infoset like this with duplicate namespace mappings:

    <ns:a xmlns:ns="foo">
      <ns:b xmlns:ns="foo">test</ns:b>
    </ns:a>

But when using SAX we would output this, with the duplicate mapping
removed:

    <ns:a xmlns:ns="foo">
      <ns:b>test</ns:b>
    </ns:a>

Although these two infosets are functionally the same, our TDMLRunner
requires that they be exactly the same, including the namespace prefix
mappings on every element. Otherwise the TDMLRunner throws an exception
and the test fails.

Ideally, these two infoset outputters would always have the same output.
Also, one could argue that our SAX ContentHandler should more closely
match the SAX events, so if we get events with duplicate mappings, the
ContentHandler should output duplicate mappings. So this modifies the
ContentHandler so that it does not remove duplicate mappings, and it
matches the XMLTextInfosetOutputter behavior/output, allowing tests to
pass.

DAFFODIL-2568

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilParseOutputStreamContentHandler.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilParseOutputStreamContentHandler.scala
@@ -214,13 +214,7 @@ class DaffodilParseOutputStreamContentHandler(out: OutputStream, pretty: Boolean
     while (currentElementPrefixMapping != null) {
       val prefix = currentElementPrefixMapping.prefix
       val uri = currentElementPrefixMapping.uri
-
-      // check to see if the prefix is already mapped to the same URI. If it
-      // is, ignore this mapping since it adds nothing new
-      val maybeUri = XMLUtils.maybeURI(activePrefixMapping, prefix)
-      if (maybeUri.isEmpty || maybeUri.get != uri) {
-        activePrefixMapping = NamespaceBinding(prefix, uri, activePrefixMapping)
-      }
+      activePrefixMapping = NamespaceBinding(prefix, uri, activePrefixMapping)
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

For Mike - the test that sometimes fails to finish within 100ms failed again in this PR too except the run was Java 17 on windows instead of Java 17 on ubuntu.  Both stack traces were almost identical except for a different function and line number in a middle frame.

<https://github.com/apache/daffodil/runs/4096774542?check_suite_focus=true>
<https://github.com/apache/daffodil/runs/4107338275?check_suite_focus=true>

```
[error] Test org.apache.daffodil.api.TestParseIndividualMessages.testDaffodilParseFromNetworkDelimited2b failed: java.util.concurrent.TimeoutException: Futures timed out after [100 milliseconds], took 0.476 sec [0.872 sec]
...
[error]     at org.apache.daffodil.api.TestParseIndividualMessages.testDaffodilParseFromNetworkDelimited2b(TestParseIndividualMessages.scala:210) [testDaffodilParseFromNetworkDelimited1b(TestParseIndividualMessages.scala:173)]
...
[info] Test run org.apache.daffodil.api.TestParseIndividualMessages finished: 1 failed, 0 ignored, 3 total, 1.095s [2.694s]
```

mbeckerle:
I made a new JIRA ticket about the 100ms test timeout failure. DAFFODIL-2582

mbeckerle (approved):
+1

PR #676, opened 2021-11-05
Rename runtime2 test files and make runtime2 tests round-trip

tuxji:
Rename all runtime2 test files to have names like test.dat for test
data files and test.dat.xml for test infoset files.  Simplify runtime2
TDML test cases by enabling round-tripping, merging pairs of parser
test cases and unparser test cases, and removing unneeded attributes
(description, root).  Fix problem found in Runtime2TDMLDFDLProcessor
after enabling round-tripping.

Makefile - Replace PARSE_DAT=parse.dat and UNPARSE_XML=unparse.xml
with TEST_DAT=test.dat and TEST_DAT_XML=test.dat.xml.

*_parse.dat - Remove the "_parse" and rename to *.dat.

*_unparse.xml - Remove the "_unparse" and rename to *.dat.xml.

*.tdml - Enable round-tripping for all test cases by setting
defaultRoundTrip="onePass".  Remove all unparserTestCases made
superfluous by round-tripping.  Change names of all test files.
Remove unneeded description and root attributes.

ex_nums.dfdl.xsd - Simplify representation property bindings.

nested.dfdl.xsd - Simplify representation property bindings and simple
types.

Runtime2TDMLDFDLProcessor.scala - Fix bug found after enabling
round-tripping where errors were still being held by
Runtime2TDMLDFDLProcessor betweeen test cases.  Don't store any errors
in Runtime2TDMLDFDLProcessor since Runtime2TDMLParseResult and
Runtime2TDMLUnparseResult already carry any errors back to TDMLRunner.

Test*.scala - Drop "_parse" from names of test cases and remove
"_unparse" test cases.

Rat.scala - Update names of all runtime2 test data files.

DAFFODIL-2583


daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.dat.xml.runtime2.error
```diff
@@ -20,7 +20,7 @@
   <array>
     <be_bool16>false</be_bool16>
     <be_bool16>true</be_bool16>
```
stevedlawrence, line 22: One concern with changing the extension to `.runtime1`, `.runtime2`, `.error`, etc. extensions is that editors are going to get confused about the file type. I think I prefer the original names, or at least a naming convention where the final extension correctly identifies the file type.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
```diff
@@ -37,62 +37,69 @@
...
     model="ex_nums.dfdl.xsd"
+    name="ex_nums_runtime1"
     config="config-runtime1">
     <tdml:document>
-      <tdml:documentPart type="file">ex_nums_parse.dat</tdml:documentPart>
+      <tdml:documentPart type="file">ex_nums.dat</tdml:documentPart>
     </tdml:document>
     <tdml:infoset>
-      <tdml:dfdlInfoset type="file">ex_nums_unparse_runtime1.xml</tdml:dfdlInfoset>
+      <tdml:dfdlInfoset type="file">ex_nums.dat.xml.runtime1</tdml:dfdlInfoset>
```
stevedlawrence, line 47: Do the different runtimes have different infosets? Does this mean there is a bug somewhere, or is this a implementation detail that one of the runtimes just doesn't support something?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
```diff
@@ -37,62 +37,69 @@
...
     model="ex_nums.dfdl.xsd"
+    name="ex_nums_runtime2_error"
     config="config-runtime2">
     <tdml:infoset>
-      <tdml:dfdlInfoset type="file">ex_nums_unparse_errors.xml</tdml:dfdlInfoset>
+      <tdml:dfdlInfoset type="file">ex_nums.dat.xml.runtime2.error</tdml:dfdlInfoset>
     </tdml:infoset>
     <tdml:errors>
       <tdml:error>value</tdml:error>
+      <tdml:error>boolean_false</tdml:error>
```
stevedlawrence, line 72: I guess we missed this in the past, but it seems odd to get both an infoset and errors. I thought we even had a check that only allowed one or the other, but I guess not.

Though, you can get both validation errors and with an infoset, which is what I think thse are? We have a `<tdml:validatorErrors>` tag for that. I assume runtime2 just doesn't differentiate between the two types of errors?

stevedlawrence (approved):
+1

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
tuxji: Yes, runtime1 and runtime2 have different infosets because Java code and C code will format floating point numbers closely but not exactly the same as each other.  We would need to either make Java and C format floats exactly the same (very difficult) or make TDMLRunner compare floats semantically, not syntactically ([DAFFODIL-2402](https://issues.apache.org/jira/browse/DAFFODIL-2402)).

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
tuxji: This is an `unparseTestCase`, not a `parseTestCase`, so that's why it has both an infoset and errors.  However, I've managed to make Runtime2DataProcessor return validation errors and change this test to a parse test case with a document, an infoset, and validation errors.  (I found out the hard way that an `unparseTestCase` isn't allowed to contain validation errors, unlike a `parseTestCase`.)  My next commit will have these changes.

tuxji (commented):
Responses below.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
stevedlawrence: Ah yes, you're right. I didn't realize this was an unparse test. 

I wonder if unparseTestCase not supporting validation errors is a bug? runtime1 doesn't implement validation for unparse so it wouldn't surprise me if we never came across that.

tuxji: Yes, the TDML standard and TDMLRunner not supporting validation errors for unparse feels like a bug to me.  Even after I changed the test to a parseTestCase, I still couldn't allow the test case to do a round trip (parse, then unparse back to the same data file) because there was no way for Runtime2DataProcessor to return validation errors inside its UnparseResult.  Both parse and unparse runtime2 code report the same validation errors so the test case would have round-tripped fine with a document, an infoset, and the same validation errors for both parse and unparse.  However,  I don't know the process for making changes to the TDML standard (I believe the changes have to be accepted by the IBM folks and others, correct?).

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala
```diff
@@ -89,13 +92,25 @@ class Runtime2DataProcessor(executableFile: os.Path) extends DFDL.DataProcessorB
...
-        val parseResult = new ParseResult(outfile, Failure(parseError), infile)
-        parseResult.addDiagnostic(parseError)
+        val validationWarnings = if (result.out.text.isEmpty) {
+          Seq.empty
+        } else {
+          val warning = new ValidationWarning { override def getMessage: String = result.out.text }
+          Seq(warning)
+        }
+        val validationFailures = if (result.err.text.isEmpty) {
+          Seq.empty
```
mbeckerle, line 105: I find it surprising that this isn't covered by tests. Isn't this the "ordinary" path taken when there are no valiation failures?

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.dfdl.xsd
```diff
@@ -23,26 +23,24 @@
...
 
-    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <!-- Representation property bindings -->
 
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
     <xs:annotation>
         <xs:appinfo source="http://www.ogf.org/dfdl/">
             <dfdl:format
-                    alignment="1"
-                    alignmentUnits="bits"
```
mbeckerle, line 32: Surprising to me that you have lengthUnits 'bits' but not alignmentUnits 'bits'. The two tend to go together.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.tdml
mbeckerle: The unparser doesn't provide any validation, other than the validation inherent in unparsing. There's never a call to Xerces to validate, for example.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
```diff
@@ -140,13 +139,11 @@ final class Runtime2TDMLDFDLProcessorFactory private(
  * TDML XML Infosets, feeding to the unparser, creating XML from the result created by
  * the Runtime2DataProcessor. All the "real work" is done by Runtime2DataProcessor.
  */
-class Runtime2TDMLDFDLProcessor(tempDir: os.Path, executable: os.Path,
-                                var diagnostics: Seq[Diagnostic]) extends TDMLDFDLProcessor {
+class Runtime2TDMLDFDLProcessor(tempDir: os.Path, executable: os.Path) extends TDMLDFDLProcessor {
```
mbeckerle, line 142: extends clause should be on the next line.

mbeckerle (approved):
+1 Very minor suggestions.

tuxji (commented):
Comments below.

PR #678, opened 2021-11-08
long to double function - for preserving binary data floating point p…

mphyo21:
…recision

Added the doubleToRawLong and doubleFromRawLong functions

DAFFODIL-2353

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
@@ -584,7 +584,6 @@ case class WholeExpression(
     // object will have been created and given this as its parent pointer.
     //
     Assert.invariant(subExpr == ifor || ifor.isInstanceOf[FunctionCallExpression])
-
     // The result of this function will be used to coerce the result of the
     // expression to the target type. However, we do not always want to allow
     // type coercion even when it might be possible. For example, if the
@@ -596,7 +595,6 @@ case class WholeExpression(
     // then we should still allow that coercion. Below we allow coercion
     // between decimal-like types where precision would not be lost, and all
     // integer-like types (which check for precision loss when evaluated).
-
     val allowCoercion = (inherentType, targetType) match {
       case (_, _) if inherentType == targetType => true
       case (_, _) if (inherentType.isSubtypeOf(NodeInfo.String) && targetType.isSubtypeOf(NodeInfo.String)) => true
@@ -1663,6 +1661,9 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           NodeInfo.Boolean, NodeInfo.String, DFDLContainsDFDLEntities(_, _))
       }
 
+
+
+
       case (RefQName(_, "testBit", DFDL), args) => {
         DFDLTestBitExpr(functionQNameString, functionQName, args)
       }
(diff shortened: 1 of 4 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/extensions/doubleFromRawLong/doubleFromRawLong.tdml
```diff
@@ -0,0 +1,65 @@
...
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"  >
+
+  <tdml:defineSchema name="doubleFromRawLong.dfdl.xsd">
+
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat"
+      lengthUnits="bytes" lengthKind="explicit" length="8"
```
stevedlawrence, line 29: In general, it's a bad idea to have the default value include an explicit length. Even though it works for this test, that rarely makes sense. Instead, I would recommend you remove `lengthKind="explicit"` and `length="8"` from this. The lengthKind value from GeneralFormat will be `implicit`, which is a sane default. And then you'll need to add `dfdl:lengthkind="explicit"` to the `LongValue` element. 

You also shouldn't need `dfdlx:emptyElementParsePolicy="treatAsEmpty"` for this tests. I would suggest removing that.

Same goes for your other tests.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/doubleToRawLong/doubleToRawLong.tdml
```diff
@@ -0,0 +1,65 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+        <root>
+        <DoubleValue>5.886014640420331E143</DoubleValue>
+        <LongValue>6757689022343245343</LongValue>
+        </root>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+</tdml:testSuite>
```
stevedlawrence, line 64: The indentation is inconsisent in this file. Suggest fixing it up or running it through a formatter.

daffodil-test/src/test/scala/org/apache/daffodil/extensions/TestdoubleToRawLong.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  }
+
+}
+
+class TestdoubleToRawLong {
+  import TestdoubleToRawLong._
+  @Test def doubleToRawLong(): Unit = { runner.runOneTest("doubleToRawLong") }
+
+}
```
stevedlawrence, line 42: Since each .tdml/.scala file only has a single test, I would suggest combining them into a single file. There's not much benefit in separating them, and just adds extra test suites/files to manage. In fact, it might even be better to just include them in an existing .tdml/.scala file. `Functions.tdml` and `TestDFDLExpressions.scala` might be a reasonable place to put these tests.

stevedlawrence (approved):
+1, looks good, just suggest some cleanup

tuxji (approved):
+1

New lines added to Functions.tdml are not always indented correctly, but I'll let it pass (eventually we'll run formatter tools on all source files and clean up all the indentation later).

daffodil-core/src/main/scala/org/apache/daffodil/dpath/Expression.scala
```diff
@@ -1687,6 +1688,16 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
 
       //End typeValueCalc related functions
 
+      case (RefQName(_, "DoubleFromRawLong", DFDLX), args) => {
```
stevedlawrence, line 1691: That name of the function should still start with a lowercase letter. It's only the class name that should have a capital D.

stevedlawrence:
All looks good :+1:  please squash into a single commit and this can be merged

PR #680, opened 2021-11-09
Append needed WiX options to default lightOptions rather than overriding

stevedlawrence:
The lightOptions key in the sbt-native-packager plugin defaults to this:

    lightOptions := Seq(
      "-ext", "WixUIExtension",
      "-ext", "WixUtilExtension",
      "-cultures:en-us",
    )

We override the lightOptions setting with these same values plus some
additional ones for things like localization and disabling validation.
But we accidentally didn't include the WixUtilExtesnsion option. At some
point an update required this extension, and without it MSI builds
started to fail.

We could add the missing WixUtilExtension value to our overridden
lightOptions setting, but to ensure we pick up any defaults that the
sbt-native-packager plugin might add and require in the future, this
instead just appends our custom options to the existing setting by using
++=. This gives us back the WixUtilExtenion that we need for builds to
succeed.

Additionally, minor fixs are made to the bat script:
- Enable setlocal so variables used in the bat file do not pollute
  variable scope outside the bat file
- Quote variables inside if-statements. Otherwise, variables with
  parenthesis in them (which is common in Windows (x86) path) cause
  bat-file errors.

DAFFODIL-2584

### DIFF
--- a/daffodil-cli/build.sbt
+++ b/daffodil-cli/build.sbt
@@ -138,11 +138,9 @@ wixProductUpgradeId := "4C966AFF-585E-4E17-8CC2-059FD70FEC77"
 // re-installation of the same version. Despite the presence of
 // specific XML to enable this, the WiX compiler and linker
 // complain about it unless you specifically suppress the warning.
-lightOptions := Seq(
+lightOptions ++= Seq(
   "-sval", // validation does not currently work under Wine, this disables that
   "-sice:ICE61",
-  "-ext", "WixUIExtension",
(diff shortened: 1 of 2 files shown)


tuxji (approved):
+1

Yep, better to append rather than override, and nice fixes to bat script.  I wondered if there was a bat equivalent of shellcheck that would have detected these problems, but a quick google couldn't find any such tool.

mbeckerle (approved):
+1

PR #683, opened 2021-11-15
Ensure Daffodil compiles with Java 8 compatibility

stevedlawrence:
DAFFODIL-2579

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -156,6 +156,7 @@ lazy val commonSettings = Seq(
   scalaVersion := "2.12.15",
   crossScalaVersions := Seq("2.12.15"),
   scalacOptions ++= Seq(
+    "-target:jvm-1.8",
     "-feature",
     "-deprecation",
     "-language:experimental.macros",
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

Will the time ever come when we drop Java 8 compatibility and upgrade our minimum Java LTS level to Java 11?  When the next Java LTS version after Java 17 comes out, or sooner than that?

stevedlawrence:
Good question. I questions that might help figure out when to drop support?

* Is Java 8 supported?

Depending on which distro of Java you're using, LTS support ranges from 2026 to 2030, so it's going be be supported for a while still. Once LTS is over, we certainly don't need to support it, but I imagine we'll drop support before that.

* Is there a maintenance burden to supporting Java 8?

I don't personally think so, since it's all automated with GitHub actions. I can't recall a time where something failed on Java 8 and so we had to change how we did something

* Are there benefits to requiring Java 11+?

I'm not familiar enough with Java 11+ to say one way or another, but being Scala based I imagine we benefit less than Java projects might. I know back when we supported Java 7 there were some Java 8 only classes we needed (e.g. java.util.Base64), which in part led to dropping Java7 support, but I don't know of any Java 11+ classes or features that we need/use. 
 
* Do we exclude anyone by dropping Java 8?

I'm not sure on this either. I do know of systems that are stuck on RHEL7, but I believe Java 11 is available on those systems. I suspect we don't exclude anyone, but I'm not really sure.


I guess to me, I'd rather err on the side of caution and continue to support Java 8, at least until it becomes a hassle to support or there's some really nice feature that we need.

mbeckerle (approved):
+1

PR #686, opened 2021-11-24
Fix NVI so defaultValue expr has access to properties of term.

mbeckerle:
@jadams-tresys You can pull this and add any other changes. 

This is, I believe, part of the fix for DAFFODIL-2587.

Doesn't have a test showing that this fixes lookAhead when called from the default value of a newVariableInstance.

But it does pass the Term object so that the TermRuntimeData is available from the NewVariableInstanceStartParser/Unparser, and is used as the context when the expression is evaluated. So this *should* fix it. 

Note that I did not make the corresponding change for SetVariable - which presumably also needs the Term so it can get access to the TermRuntimeData, so that bitOrder and other physical properties are available for functions like lookAhead. 

DAFFODIL-2587

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
@@ -30,6 +30,7 @@ import org.apache.daffodil.processors.Evaluatable
 import org.apache.daffodil.processors.Failure
 import org.apache.daffodil.processors.RuntimeData
 import org.apache.daffodil.processors.Success
+import org.apache.daffodil.processors.TermRuntimeData
 import org.apache.daffodil.processors.TypeCalculator
 import org.apache.daffodil.processors.VariableRuntimeData
 import org.apache.daffodil.schema.annotation.props.gen.FailureType
@@ -150,8 +151,10 @@ class TypeValueCalcParser(typeCalculator: TypeCalculator, repTypeParser: Parser,
 
 }
 
-class SetVariableParser(expr: CompiledExpression[AnyRef], decl: VariableRuntimeData)
+final class SetVariableParser(expr: CompiledExpression[AnyRef], decl: VariableRuntimeData, trd: TermRuntimeData)
   extends ExpressionEvaluationParser(expr, decl) {
+
+  override val context = trd
 
   def parse(start: PState): Unit = {
     Logger.log.debug(s"This is ${toString}") // important. Don't toString unless we have to log.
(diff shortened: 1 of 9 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ExpressionEvaluatingParsers.scala
```diff
@@ -161,15 +162,18 @@ class SetVariableParser(expr: CompiledExpression[AnyRef], decl: VariableRuntimeD
   }
 }
 
-class NewVariableInstanceStartParser(override val context: VariableRuntimeData)
+final class NewVariableInstanceStartParser(vrd: VariableRuntimeData, trd: TermRuntimeData)
   extends PrimParser {
+
+  override def context = trd
```
stevedlawrence, line 168: Any chance this could affect diagnostics? For example, if `evaluate` is called and it results in a runtime SDE, what gets used as the context for that SDE? If it uses this `context` variable then the SDE message might only include the term the NVI is placed on instead of the actual NVI element, which is maybe confusing? Reading the code, it's not exactly clear to me what a diagnostic would look like. Might be worth adding a test (or maybe we already have a test) to make sure the diagnostic is reasonable?

mbeckerle: Yes, this will change where blame is placed for the error. Fixing that would require us to pass both the VRD and TRD down further into the software so that the blame can be on the VRD, even though things like bitOrder must come from the TRD. 

Interesting that the def context is not covered by tests according to codeCov. I will see if I can contrive a test that will force it to be used.

jadams-tresys: I added some tests that I believe addresses the missing coverage on codeCov

mbeckerle: Hmmm. The VRD actually just identifies what variable, right? I don't think there is a NVIRuntimeData object. Nor a SetVariableRuntimeData that contain the say, line number and file info for that schema component. So existing error messages are going to refer to the variable itself. Using the TermRuntimeData of the term is going to be a better error message than using the VRD (which is what we had been doing.)

There's a NonTermRuntimeData object which could potentially be the representative object for these NVI or SetVar statements. But I wouldn't change this currently.

mbeckerle: Another commit adds tests that divide by zero in a setVariable expression. We do get a diagnostic, but the behavior is poor as the line info is way off. Added DAFFODIL-2594 and a test that characterizes this for now. 

This bug fix does allow dfdl:newVariableInstance and dfdl:setVariable to invoke dfdlx:lookAhead, so is worth having even if we fix the poor diagnostic behavior later.

mbeckerle: And another commit with tests specifically for getting an error from defaultvalue of dfdl:newVariableInstance.

mbeckerle: These errors *should* use the context member and so provide code coverage on those lines of the NVI implementation.

jadams-tresys (approved):
+1
It will be nice to get better diagnostics too, but I'm OK with that being pushed off to a different bug

stevedlawrence (approved):
+1

PR #689, opened 2021-11-29
Add tests verifying correct textNumberRoundingIncrement behavior

stevedlawrence:
Tests added based on comments at https://unicode-org.atlassian.net/browse/ICU-20425

All tests pass with fixes made as part of ICU 70.1, some tests fail with
older versions.

DAFFODIL-2534


tuxji (approved):
If all tests pass in CI, then here's my +1

jadams-tresys (approved):
+1

PR #691, opened 2021-11-30
Move where runtime2 stores the daffodil version

stevedlawrence:
Currently, runtime2 writes the daffodil version to generated_code.c. We
store some example generated_code.c files in the repo as examples, which
means when we bump the Daffodil version number we must also remember to
update these examples. That's not hard and we do have github checks to
ensure they are updated, but it's one more thing to think about during
the release process.

To avoid this extra work, this adds a new libcli/version.h file and
overwrites it with the actual version number when generating code. Since
this file is not stored along with the generated example code, we don't
have to worry about updating generated code stored in the repo during
version bumps.

This also adds a new ForkCaptureLogger to make it easier to capture
stderr/stdout from our forked processes in sbt and handle it
appropriately. This way we can get a list of generated files from std
out and log errors from stderr.

DAFFODIL-2592

### DIFF
--- a/daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/version.h
+++ b/daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/version.h
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/version.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef VERSION_H
+#define VERSION_H
+
+/**
+ * This file will be overwritten with the correct program version informaion
+ * when code is generated. This exists here only to support compiling the
+ * source during Daffodil builds without generated code to ensure there are no
+ * compilation errors.
+ */
```
tuxji, line 26: Let's use // to be consistent with comments in all other C files, and let's tweak what you said a tiny bit :).

```c
// This file will be overwritten with the correct program version information
// when code is generated. This file exists here to support compiling the other
// C files without generated code during Daffodil builds to ensure there are no
// compilation errors in these C files.
```

tuxji (approved):
+1

Looks great, just a slight tweak requested.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/version.h
stevedlawrence: Thanks! Fixed.

jadams-tresys (approved):
+1

PR #692, opened 2021-11-30
Change component count message from info to debug

jadams-tresys:
This message from the compiler is really only intereseting for debugging
purposes so it's logging should reflect that.

DAFFODIL-2577


tuxji (approved):
+1

stevedlawrence (approved):
+1

PR #694, opened 2021-12-02
Check for RepType when getting element length

jadams-tresys:
DAFFODIL-2596

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
@@ -650,13 +650,13 @@ trait ElementBase
     // lengthUnits is bytes/bits, but the fixed length is given in characters
     // of a variable-width charset, so greater than 0, but we don't know
     // exactly.
-    Assert.usage(isFixedLength)
-    if (lengthKind =:= LengthKind.Explicit) lengthEv.optConstant.get
+    Assert.usage(repElement.isFixedLength)
+    if (repElement.lengthKind =:= LengthKind.Explicit) repElement.lengthEv.optConstant.get
     else {
-      Assert.invariant(lengthKind =:= LengthKind.Implicit)
+      Assert.invariant(repElement.lengthKind =:= LengthKind.Implicit)
       // it's a string with implicit length. get from facets
-      schemaDefinitionUnless(this.hasMaxLength, "String with dfdl:lengthKind='implicit' must have an XSD maxLength facet value.")
-      val ml = this.maxLength
+      schemaDefinitionUnless(repElement.hasMaxLength, "String with dfdl:lengthKind='implicit' must have an XSD maxLength facet value.")
+      val ml = repElement.maxLength
       ml.longValue()
     }
   }
@@ -680,7 +680,7 @@ trait ElementBase
   }
 
   final lazy val fixedLength = {
-    if (isFixedLength) lengthEv.optConstant.get.longValue() else -1L
+    if (isFixedLength) repElement.lengthEv.optConstant.get.longValue() else -1L
     // FIXME: shouldn't even be asking for this if not isFixedLength
     // try changing to Assert.usage(isFixedLength)
     // FIXME: needs to return fixed length in lengthUnits.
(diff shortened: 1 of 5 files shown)


stevedlawrence (approved):
+1 seems fine, but it seems we we're going down the path of changing more and more variables to be something like:

```scala
val someVariable =
  if (optRepTypeElement.isDefined) {
    optRepTypeElement.get.someVariable
  } else {
    // actual variable logic
  }
```
There's now 

Feels like it's going to be a game a whack-a-mole to find all the right places where this is logic is needed. Makes me wonder if maybe we need to doing the `optRepTypeElement.isDefined` check a level or two above before these variables are even accessed? Or maybe we need something like this:

```
val repElement = optRepTypeElement.getOrElse(this)
```

And then when something accesses one of those variables that needs this isDefined check, we can use this new variable, for example:

```
repElement.fixedLengthValue
repElement.maybeFixedLengthInBits
```

I'm not sure if that's actually a good idea or not. Maybe these checks are actually better done or belong in this variables that care about repType? Just throwing out an idea that might make it easier to avoid these issues in the future?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -180,6 +182,22 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+  <tdml:parserTestCase name="InputTypeCalc_keysetValue_03"
```
mbeckerle, line 185: change description string. Add comments to the file explaining the test case.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -50,6 +50,8 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="keysetValue_03" type="tns:keysetValue_01" dfdl:lengthKind="implicit"/>
```
mbeckerle, line 53: So is this the workaround? Wherever you use a type where the type carries a dfdlx:repType you need this dfdl:lengthKind 'implicit' ?

How come the other tests in here didn't need this?

Please add doc/comments explaining this and include the DAFFODIL-NNNN ticket number about the issue.

daffodil-test/src/test/scala/org/apache/daffodil/extensions/TestInputTypeValueCalc.scala
```diff
@@ -43,10 +43,12 @@ class TestInputTypeValueCalc {
   @Test def test_InputTypeCalc_keysetValue_00(): Unit = { runner.runOneTest("InputTypeCalc_keysetValue_00") }
   @Test def test_InputTypeCalc_keysetValue_01(): Unit = { runner.runOneTest("InputTypeCalc_keysetValue_01") }
   @Test def test_InputTypeCalc_keysetValue_02(): Unit = { runner.runOneTest("InputTypeCalc_keysetValue_02") }
+  @Test def test_InputTypeCalc_keysetValue_03(): Unit = { runner.runOneTest("InputTypeCalc_keysetValue_03") }
```
mbeckerle, line 46: The name of this test probably should change also, as I don't think of this issue as a "inputTypeCalc" nor "keysetValue" thing. Suggestion: test_repTypeOnTypeRequiresLengthKindImplicit

mbeckerle (changes requested):
Needs more doc, some test renaming.

mbeckerle:
re: comment by @stevedlawrence on proliferating conditionals, the alternative is some object-oriented refactoring, where things that have repType get a different class than things without repType. This would actually require delegation I think to another object since we don't/can't analyze the repType property when the original DSOM object is created. I.e., we create the ElementBase instance as part of core parsing of the DFDL schema. Then, based on combining of element with its type, we eventually find out that it has a repType or not, and we'd have to construct a distinct object for the repType vs. Not case, and delegate the fixedLengthValue and elementLengthInBitsEv methods on that object. This is do-able, and may in fact be the right approach vs. proliferating conditionals of repType or not.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
jadams-tresys: Will do.  Not a workaround, merely setting it to implicit to trigger the bug that the code changes fixed.  The problem had been that it was thinking that the keysetValue_01 type was a string, which when lengthKind="implicit" causing it to complain about a missing maxLength property as you can't have an implicit length string without one.

This was one of the things we were missing when we were trying to debug the P8 schema yesterday.  The tests that were working had a lengthKind="delimited", which was fine given that it thought it was processing a string.  Adding dfdl:lengthKind="implicit" seemed like the easiest way to trigger the bug we were seeing in the P8 schema.

I'll work on adding some comments and a better name for the test now.

mbeckerle (approved):
+1.

stevedlawrence:
What changed since this was last reviewed? Unfortunately GitHub doesn't have a way to show that.

Also there were some discussions about potentially switching to a OOP/delegation system to reduce all the if statements scattered throughout. Do we want to take that approach, or is that going to be too much effort for not much gain?

jadams-tresys:
> What changed since this was last reviewed? Unfortunately GitHub doesn't have a way to show that.
> 
> Also there were some discussions about potentially switching to a OOP/delegation system to reduce all the if statements scattered throughout. Do we want to take that approach, or is that going to be too much effort for not much gain?

The change to OOP/delegation is what changed from the last commit.  Instead of adding more
`if (optRepTypeElement.isDefined)`
statements throughout the code I am know using the repElement suggestion you mentioned earlier in the review, which can either be the optRepTypeElement, or the element itself.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/ElementBase.scala
```diff
@@ -693,23 +693,19 @@ trait ElementBase
...
-        if (bitsMultiplier > 0) {
-          MaybeULong(fixedLengthValue * bitsMultiplier)
-        } else {
-          MaybeULong.Nope
-        }
+    if (isRepresented && repElement.isFixedLength) {
+      val bitsMultiplier = repElement.lengthUnits match {
+        case LengthUnits.Bits => 1
+        case LengthUnits.Bytes => 8
+        case LengthUnits.Characters => if (knownEncodingIsFixedWidth) knownEncodingWidthInBits else -1
```
stevedlawrence, line 700: Should this be `repElement.knownEncoding...`

stevedlawrence (approved):
+1 This is a definite improvement.

It still feels a bit fragile in that there's a bunch of repElements scattered throughout and it's not obvious that we added all the places where it's need. But I guess if we're missing any they'll eventually show up as bugs, and if it continues we can consider some other refactoring.

PR #695, opened 2021-12-06
setup for 3.3.0-SNAPSHOT development

mbeckerle:
DAFFODIL-2604


stevedlawrence (approved):
+1

I've also merged PR #691, so a rebase should fix this if it fails because the example versions don't match

tuxji (approved):
+1

mbeckerle:
The checks are going to fail, as this needs to be rebased on top of an earlier commit.

PR #697, opened 2021-12-06
Use same version for both log4j-api and log4j-core

tuxji:
Daffodil's libs contain log4j-api-2.13.2 and log4j-core-2.14.1.  They
should contain the same version of both log4j-api and log4j-core.

project/Dependencies.scala: Add log4j-api as well.

DAFFODIL-2602


mbeckerle (approved):
+1

stevedlawrence (approved):
+1

PR #698, opened 2021-12-07
Updated constructor that only requires cause exception as an argument…

mphyo21:
… to DaffodilUnhandledSAXException

DAFFODIL-2433

Add constructor that only requires cause exception as argument to DaffodilUnhandledSAXException

DAFFODIL-2433

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
@@ -263,9 +263,11 @@ object DFDL {
   /**
    * Used in SAXInfosetEvent.causeError to identify when an unexpected error occurs
    */
-  class DaffodilUnhandledSAXException(description: String, cause: Exception)
-    extends org.xml.sax.SAXException(description, cause)
 
+  class DaffodilUnhandledSAXException(description: String, cause: Exception)
+    extends org.xml.sax.SAXException(description, cause) {
+      def this(cause: Exception)= this(null, cause)
+}
   class SAXInfosetEvent() {
     var localName: Maybe[String] = Nope
     var simpleText: Maybe[String] = Nope
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
@@ -207,7 +207,8 @@ class SAXInfosetInputter(
       }
     } catch {
       case e: Exception => {
-        batchedInfosetEvents(currentIndex).causeError = One(new DaffodilUnhandledSAXException(e.getMessage, e))
+        batchedInfosetEvents(currentIndex).causeError = One(new DaffodilUnhandledSAXException(e))
+
       }
     } finally {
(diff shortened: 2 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -207,7 +207,8 @@ class SAXInfosetInputter(
       }
     } catch {
       case e: Exception => {
-        batchedInfosetEvents(currentIndex).causeError = One(new DaffodilUnhandledSAXException(e.getMessage, e))
+        batchedInfosetEvents(currentIndex).causeError = One(new DaffodilUnhandledSAXException(e))
+
```
tuxji, line 211: Line 210 isn't covered by any test, which causes the new constructor to be uncovered too.  

Does anyone have any idea how to write a test that would cause an exception to escape from dp.unparse at line 202 without being wrapped in an unparseResult.isError?  

@mphyo21, if you want to practice your IDE navigation skills, please use your IDE's Go To functionality to find out which code implements the dp.unparse call at line 202 and inform us so we can check if there is any way a test could make an exception escape from that code.  If the IDE can't tell you which code implements dp.unparse, you can practice your IDE debugging skills instead - set a breakpoint at line 202, run a test from your IDE, step into the dp.unparse call, and report back to us which code implements that call.

tuxji (approved):
+1

Looks good, although I would like to understand why no test covers line 210.

daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala
```diff
@@ -263,9 +263,11 @@ object DFDL {
   /**
    * Used in SAXInfosetEvent.causeError to identify when an unexpected error occurs
    */
-  class DaffodilUnhandledSAXException(description: String, cause: Exception)
-    extends org.xml.sax.SAXException(description, cause)
 
+  class DaffodilUnhandledSAXException(description: String, cause: Exception)
+    extends org.xml.sax.SAXException(description, cause) {
+      def this(cause: Exception)= this(null, cause)
```
mbeckerle, line 269: A unit test (not a TDML test, just regular unit test) should construct one of these objects so this line is covered.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
```diff
@@ -207,7 +207,8 @@ class SAXInfosetInputter(
       }
     } catch {
       case e: Exception => {
```
mbeckerle, line 209: Catch of Exception requires documentation. As part of figuring out how to get the change line covered by a test, you should add comments to this scala code as to why catching Exception is ok in this case. 

Does anyone (@stevedlawrence @olabusayoT ) know why this is ok? Normally catching Exception or Throwable is almost always an error. Please chime in.  

If this in fact should not be Exception, but something narrower, then let's fix that.

mbeckerle (changes requested):
Need unit test and there's an issue around getting test coverage of an Exception catch... which is problematic and at least needs doc. 

You may need help from other developers more familiar with this area. I tagged 2 of them in the comments.

mbeckerle:
Closed as inactive. The associated JIRA issues link back to this PR in case it is useful in the future.

PR #701, opened 2021-12-09
Rename version.h to daffodil_version.h

tuxji:
Rename version header and update C code generator

StringDataInputStreamForUnparse.scala: Restore // $COVERAGE-ON$
because there's a // $COVERAGE-OFF$ preceding it (misc fix too small
for its own JIRA issue).

.clang-format: Move from resources to c so it will be included
alongside Makefile in C code generator's output directory (Makefile
has a "format" goal which needs our .clang-format to do it right).

daffodil_getopt.c: Include renamed version header and add iwyu
comment.  Rename daffodil_program_version to match version header's
name.

version.h: Rename to daffodil_version.h (looks better in dired
listing).  Also rename include guards and daffodil_program_version to
match version header's name.

CodeGenerator.scala: Change version header's name and reorder lines
better semantically.

CodeGeneratorState.scala: Rename include guards and
daffodil_program_version to match version header's name.

DAFFODIL-2592


daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/daffodil_getopt.c
```diff
@@ -17,10 +17,10 @@
...
 // clang-format off
 #include "daffodil_getopt.h"
-#include <string.h>      // for strcmp, strrchr
-#include <unistd.h>      // for optarg, getopt, optopt, optind
-#include "cli_errors.h"  // for CLI_UNEXPECTED_ARGUMENT, CLI_HELP_USAGE, CLI_INVALID_COMMAND, CLI_INVALID_INFOSET, CLI_INVALID_OPTION, CLI_MISSING_COMMAND, CLI_MISSING_VALUE, CLI_PROGRAM_ERROR, CLI_PROGRAM_VERSION
-#include "version.h"
+#include <string.h>            // for strcmp, strrchr
+#include <unistd.h>            // for optarg, getopt, optopt, optind
+#include "cli_errors.h"        // for CLI_UNEXPECTED_ARGUMENT, CLI_HELP_USAGE, CLI_INVALID_COMMAND, CLI_INVALID_INFOSET, CLI_INVALID_OPTION, CLI_MISSING_COMMAND, CLI_MISSING_VALUE, CLI_PROGRAM_ERROR, CLI_PROGRAM_VERSION
+#include "daffodil_version.h"  // for daffodil_program_version
```
stevedlawrence, line 23: `CodeGenerator.scala` also needs a change to reference `daffodil_version.h` instead of `version.h`, since it overwrites the file with the correct version number.

Though, another option might be to create an sbt resource generator that generates `daffodil_version.h` rather than having a dummy file that the code generator overwrites.  I can see if I can whip that up if you like that better.

tuxji: I'd made/tested my changes in a generated output directory and I'd forgotten about CodeGenerator.scala, sigh.  

Your sbt resource generator sounds good, but I'll go ahead and update CodeGenerator.scala since that'll be quicker.  Total number of lines of code probably will be smaller this way too.

stevedlawrence: Fair enough. Another thought, should the `ifndef`/`define` in the file be changed from `VERSION_H` to `DAFFODIL_VERSION_H`?

stevedlawrence (approved):
+1

PR #707, opened 2021-12-15
Change layering to pass and use state.

mbeckerle:
Removed state from LayerRuntimeInfo

This fixes layer unparsing as the passed state can be the one created for use
by suspensions.

Eliminated LayerSerializedInfo, as it is the same as LayerRuntimeInfo now.

Eliminated VariableHandle (just use VariableRuntimeData)

Tested with PCAP and EthernetIP, both of which now work.

DAFFODIL-2608

### DIFF
--- a/.sonar-project.properties
+++ b/.sonar-project.properties
@@ -16,7 +16,7 @@
 sonar.organization=apache
 sonar.projectKey=apache-daffodil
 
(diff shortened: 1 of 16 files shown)


tuxji (approved):
+1

Looks good to me.  One check failed due to a downloading problem that's probably a fluke.

stevedlawrence (approved):
+1

I think eventually we probably need a handle or some opaque pointer type of thing for the the variable and state stuff to make the API safer. But as long as we're okay with this API being considered experimental and likely to change, we can do that later. We should create a new ticket to make sure we don't lose this need though, assuming DAFFODIL-2608 is closed when this is merged.

PR #708, opened 2021-12-16
Prepare for patch release 3.2.1

mbeckerle:
Change version in build.sbt to 3.2.1

This is for a urgent patch release to fix 2 CVEs (log4J, JDOM)
and to incorporate the fix for unparsing layers with checksums
(DAFFODIL-2608) which
is a key feature of 3.2.0 found to be broken.

DAFFODIL-2609


stevedlawrence (approved):
+1

tuxji (approved):
+1

PR #715, opened 2021-12-27
Add a "daffodil" alias to the RPM

stevedlawrence:
The official name of the Daffodil RPM is "apache-daffodil", which means
this intuitive command to install the RPM with dnf does not work:

    dnf install daffodil

Instead you need to install using the full name:

    dnf install apache-daffodil

To support both commands, adds a "Provides: daffodil" tag to the RPM
build configuration, which essentially adds an alias so that users can
dnf install either "daffodil" or "apache-daffodil".

DAFFODIL-2613


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #719, opened 2021-12-31
Added test to reproduce shared group bug.

mbeckerle:
Commented out currently so it won't break the build.

DAFFODIL-2615


tuxji (approved):
+1

stevedlawrence (approved):
+1

PR #720, opened 2022-01-04
WIP: Feature to provide a dfdlx:schema that wraps an XSD-based DFDL schema

mbeckerle:
This provides better support for writing DFDL schemas from ordinary XML
infrastructure in IDEs/Tools.

Fix incorrect URI

Removed dfdl-config-format.xsd

This was redundantly defining the dafext namespace. Unusable that way in
conjunction with the dafext.xsd file.

Nothing was including/importing this file anyway.

Moved the one definition in it, of dfdlConfig to dafext.xsd.

Provided xsi:schemaLocation in the XML files that contained dfdlConfig
elements.

Correct invalid tdml files that should be valid.

Mostly this is missing definitions for ex:, tns: and dfdlx:

DAFFODIL-1638

Provide dfdlx:schema element which allows for IDE support.

By creating dfdlx:schema in files with extension ".dfdl" (not
".dfdl.xsd") you can edit the schema in an IDE, and get full support
because the IDE's own XSD support doesn't get in the way.

You must setup the IDE so ".dfdl" files are treated as XML just as
you do with ".tdml" files. Then turn off XML Schema Validation, but
leave turned on XML validation.

### DIFF
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
@@ -107,6 +107,29 @@
     <xs:attribute form="qualified" name="layerLengthUnits" type="dfdlx:LayerLengthUnitsEnum" />
     <xs:attribute form="qualified" name="layerBoundaryMark" type="dfdl:ListOfDFDLStringLiteral_Or_DFDLExpression" />
   </xs:attributeGroup>
+  
+  <!-- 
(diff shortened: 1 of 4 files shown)


tuxji:
Looks OK, but tests fail when generating example files so will wait until that's fixed before reviewing.

mbeckerle:
I didn't even want to create PRs for these, but github won't shut up about
various of my "parked" things pushed to my fork, which are really just
pushed there for backup.

mbeckerle:
Closed as inactive.

PR #721, opened 2022-01-05
Support --infoset null for CLI unparsing

stevedlawrence:
A "null" infoset inputter does not normally make sense in the context of
unparsing because there must always be an infoset to unparse--you can't
unparse null data. But in the context of performance, a "null" infoset
inputter could be considered one that has close to zero overhead
(similar to the null infoset outputter for parsing), which can help to
understand the raw speed of Daffodil unparsing operations and to compare
different infoset inputter overheads.

To support this, this allows "null" to be used for the --infoset option
when unparsing. This triggers the use of a new NullInfosetInputter,
which uses a pre-created array of infoset events that are exactly what
the Daffodil unparser expects. The events are created outside of
measured unparser code, so the only overhead that occurs inside measured
code is indexing into the event array and returning fields from the
indexed event, which should give good picture of raw unparse speed.

DAFFODIL-2619

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.infoset
+
+import java.io.InputStream
+
+import scala.collection.mutable.ArrayBuffer
+
+import scala.xml.Elem
+import scala.xml.SAXParser
+import scala.xml.Text
+import scala.xml.XML
+
+import org.apache.daffodil.dpath.NodeInfo
+import org.apache.daffodil.infoset.InfosetInputterEventType._
+import org.apache.daffodil.util.MaybeBoolean
+import org.apache.daffodil.xml.DaffodilSAXParserFactory
+import org.apache.daffodil.xml.XMLUtils
+
+object NullInfosetInputter {
+
+  case class Event(
+    eventType: InfosetInputterEventType,
+    localName: String = null,
+    namespaceURI: String = null,
+    simpleText: String = null,
+    isNilled: MaybeBoolean = MaybeBoolean.Nope,
+  )
+
+  def toEvents(is: InputStream): Array[Event] = {
+    val elem = {
+      val parser: SAXParser = {
+        val f = DaffodilSAXParserFactory()
+        f.setNamespaceAware(false)
+        val p = f.newSAXParser()
+        p
+      }
+      val node = XML.withSAXParser(parser).load(is)
+      val normalized = XMLUtils.normalize(node)
+      normalized
+    }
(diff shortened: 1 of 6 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.daffodil.infoset.InfosetInputterEventType._
+import org.apache.daffodil.util.MaybeBoolean
+import org.apache.daffodil.xml.DaffodilSAXParserFactory
+import org.apache.daffodil.xml.XMLUtils
+
+object NullInfosetInputter {
+
+  case class Event(
+    eventType: InfosetInputterEventType,
+    localName: String = null,
```
mbeckerle, line 40: Can we exclude the overhead of nextElementErd() by timing it separately allowing us to include it, or exclude it from what we're counting? 

The fact that the serialized XML or JSON or whatever doesn't exactly identify the ERD for us is overhead related to the serialized form. (If we invented our own serialized form we would put an unambiguous ERD identifier in the serialized form, so that there would be no nextElementErd() overhead. )

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
```diff
@@ -0,0 +1,123 @@
...
+    namespaceURI: String = null,
+    simpleText: String = null,
```
mbeckerle, line 42: Similarly, the simpleText conversion into the binary simple data type is overhead we would not have if we were using our own serialization form, because the ERD would tell us exactly the simple type, so we could store the binary value. 

I am less worried about this overhead than the nextElementErd() overhead however. But still if we have fine-enough granularity timers, we could time the conversion of the simple value from string to our value object that is stored in the DISimple node, and then that overhead could be excluded (or counted as part of serialization overhead.)

mbeckerle (approved):
+1 What you've got seems good. But it does not enable us to exclude certain serialization overheads per the comments. I'd like to understand if we can exclude/separate those.

tuxji (approved):
+1

I'm fine with Mike's comments being addressed by performance optimizations elsewhere in another pull request.  This NullInfosetOutputter seems to be focused on the right goal (returning events as quickly as possible).  

However, I think this pull request should have an explicit CLI test case added to ensure the new lines in Main.scala are covered by tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
```diff
@@ -0,0 +1,123 @@
...
+          (text, isNilled)
+        } else {
+          (null, MaybeBoolean.Nope)
+        }
+
+        events += Event(StartElement, localName, namespaceURI, simpleText, isNilled)
+        if (!isSimple) elem.child.foreach { nodeToEvents(_, events) }
+        events += Event(EndElement, localName, namespaceURI)
+      }
+      case _ => // ignore
```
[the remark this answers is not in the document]
stevedlawrence: I think really this function should just expect an `Elem` instead of `Node`. Due to the `XMLutils.normalize` call it's impossible for the node parameter to not be a `Elem`. Will update it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/NullInfosetInputter.scala
stevedlawrence: I've looked at this more to see if it's even possible, and I'm not sure it's possible without pretty big changes. An `InfosetInputter` implementation doesn't currently know anything about ERDs. And the `InfosetInputter` base class only learns about the root ERD when `initialize()` is called, which happens at the beginning of unparse. So there is no way for an InfosetInputter to know about any ERDs prior to unparsing, and so there's no way to figure out all the information needed for `nextElementERD`.

If we really wanted a super minimal `InfosetInputter`, we would have to do as you suggest and create a custom `InfosetOutputter` that serializes the infoset along with all ERD information, and then a custom InfosetInputter could use that to recreate events and avoid `nextElementErd` overhead. But the feels like quite a bit of work to just exclude the overhead of `nextElementErd()`. And since `nextElementErd()` isn't specific to any `InfosetInputter` implementation, I'd rather we just add instrumentation or profile it separately and find ways to optimize it, rather than creating a special `InfosetInputter`/`Outputter` that avoids it entirely.

PR #725, opened 2022-01-12
Remove 4 sonarcloud code smells

mbeckerle:
DAFFODIL-2622


stevedlawrence (approved):
+1

mbeckerle:
Interesting. Sonarcloud told me a local var was unused.

But it isn't unused. When I deleted it, daffodil now fails to compile. 

So this is a sonarcloud error.

tuxji (approved):
+1

Good to remove commented-out lines like these.

PR #726, opened 2022-01-13
Fix C compilation errors due to name collisions

tuxji:
Runtime2 was generating C code with compilation errors from schemas
with name collisions in elements/types or even name collisions with
names in C header files.  Make runtime2 build more distinct C struct
names to prevent those name collisions.  Also make runtime2 generate
correct root element name everywhere using same method.

Main.scala: Don't pass rootNS to generateCode method.

DFDLParserUnparser.scala: Remove generateCode's rootNS parameter.

CodeGenerator.scala: Remove generateCode's rootNS parameter.  Pass
root from CodeGenerator constructor to CodeGeneratorState constructor
instead.  Don't pass rootElementName to generateCodeFile method or
rootNS to generateCode method.

CodeGeneratorState.scala: Add root parameter to CodeGeneratorScala
constructor.  Remove firstElementSeen and change elementNotSeenYet,
erdName, buildName, and localName (rename to cStructName) to handle
elements/types with same names by building hopefully unique names for
their C struct names.  Get root element's name from root parameter
everywhere its name is needed.  Use element's local name, not C struct
name, in defineQNameInit (also addComputations) and rename qnameInit
to qNameInit.  Remove generateCodeFile's rootElementName parameter and
get rootName from constructor's root parameter instead.

ElementParseAndUnspecifiedLengthCodeGenerator.scala: Avoid defining
duplicate ERD structures for simple types as well as complex types.

examples/{NestedUnion,ex_nums}/generated_code.[ch]: Regenerate
generated code examples with more distinct C struct names.

TestCodeGenerator.scala: Don't pass rootNS/None to generateCode.

Runtime2TDMLDFDLProcessor.scala: Don't pass rootNS to generateCode.

collisions.{dat,dat.xml,dfdl.xsd,tdml}: Add test to verify name
collisions don't cause C compilation errors anymore.

TestCollisions.scala: Run above test.

Rat.scala: Ignore collisions.dat (binary file).

DAFFODIL-2623


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1440,11 +1440,10 @@ object Main {
               tunables, generateOpts.language)
 
             // Ask the CodeGenerator to generate source code from the DFDL schema
-            val rootNS = generateOpts.rootNS.toOption
             val outputDir = generateOpts.outdir.toOption.getOrElse(".")
             val rc = generator match {
               case Some(generator) => {
-                Timer.getResult("generating", generator.generateCode(rootNS, outputDir))
+                Timer.getResult("generating", generator.generateCode(outputDir))
```
mbeckerle, line 1446: Is this just decluttering, or is there a different way to deal with namespaces here? Eventually we'll want to support them.

mbeckerle (approved):
+1 The changes are fine, though I am concerned that we'll need namespace support eventually and this rips it out entirely. 

That said, the right way to fix that is by making the QName classes (e.g., NamedQName) support creation of useful identifiers directly. Given a schema set, a QName should be able to determine the best prefix mapped to a namespace, and construct names for a QName that use the prefix characters to represent the namespace. E.g., if xmlns:ex="http://example.com" and a given element named "foo" at top level in a schema has that target namespace, then foo_ex should be the name used for those global elements.

tuxji:
> +1 The changes are fine, though I am concerned that we'll need namespace support eventually and this rips it out entirely.

I hope your comment is just a misapprehension.  I don't believe the changes rip out any namespace support.

> That said, the right way to fix that is by making the QName classes (e.g., NamedQName) support creation of useful identifiers directly. Given a schema set, a QName should be able to determine the best prefix mapped to a namespace, and construct names for a QName that use the prefix characters to represent the namespace. E.g., if xmlns:ex="http://example.com" and a given element named "foo" at top level in a schema has that target namespace, then foo_ex should be the name used for those global elements.

Identifier construction (the cStructName method) actually constructs nested struct identifiers' names starting with the outermost structs' names and appending the nested structs' local names, rather than using only one local name (the nested struct's name) like it did before (the original localName method).  I think I tried appending prefix and local name to create foo_ex for a struct name, but it must not have been sufficient to prevent problems (there were several reasons for the C compilation or runtime errors including duplicate definitions, missing definitions, name collisions with functions in C headers, etc.).  I believe we need to build nested elements' struct names using outer structs' names to avoid the problems I found, although I agree we might need to disambiguate two global non-nested elements' struct names by prepending their prefixes to their local names too if we find that case causes a name collision later.

mbeckerle:
I agree you need to use the nest to avoid local name clashes. 
The global names will also clash without the namespaces being somehow represented. 

I think the Daffodil schema compiler should take all the namespace bindings from all files, and produce a set of good unique namespace to identifier-suffix mapping, based on reusing the prefixes selected by schema authors wherever possible.

PR #731, opened 2022-01-21
Add test that reproduces Runtime SDE bug.

mbeckerle:
DAFFODIL-2629

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
```diff
@@ -0,0 +1,77 @@
...
+      <complexType>
+        <sequence>
+          <element name="someElement" type="xs:int"/>
+          <!--
+          DAFFODIL-2629 This setVariable calling dfdl:valueLength on the prior element. Causes Runtime SDE."
+          -->
+          <sequence>
+            <annotation>
+              <appinfo source="http://www.ogf.org/dfdl/">
+                <dfdl:setVariable ref="f:someVar">{ dfdl:valueLength(someElement, "bytes") }
```
stevedlawrence, line 51: If you use the `value` attribute for the expression does it still fail? Might be good to add a test for both syntaxes to make sure however this is resolved works for both.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
```diff
@@ -0,0 +1,77 @@
...
+                </dfdl:setVariable>
+              </appinfo>
+            </annotation>
+          </sequence>
```
stevedlawrence, line 55: Is it worth added an inputValueCalc and/or outputValueCalc to ensure that the variable actually gets the right value? I could imagine a but where the SDE goes away but doesn't actually calculate the correct value and we wouldn't know.

stevedlawrence (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
mbeckerle: Yup. I  added a test to verify this.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/setVarWIthValueLength.tdml
mbeckerle: Added a subsequent assertion that the variable gets value 4.

PR #732, opened 2022-01-21
Add a global TDML CompileResult cache

stevedlawrence:
We previosly only cache CompileResults on individual DFDLTestSuites.
This meant that we could compiled schemas within a single test suite,
but not between tests suites. This could lead to lots of duplicate
schema compilation if multiple .tdml files referenced the same schema

To avoid this, CompileResults are now stored in a global WeakHashMap
instead of the DFDLTestSuite. And DFDLTestSuites are changed to maintain
strong references to the keys of the CompileResult they use. This
ensures that as long as a DFDLTestSuite exists, so will all of its
CompileResults, so tests within a DFDLTestSuite can share them. And
after the DFDLTestSuite is garbage collected, as long as those weak
references aren't also garbage collected, other DFDLTestSuites will be
able to use them, and help to minimize compilation same schemas used in
different TDML files.

DAFFODIL-2627

### DIFF
--- a/daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
+++ b/daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
@@ -113,6 +113,11 @@ final class Runtime2TDMLDFDLProcessorFactory private(
       // Compile the generated code into an executable
       val executable = generator.compileCode(codeDir)
 
+      // There is is no way to clean up TDMLDFDLProcessors, they are just garbage
+      // collected when no longer needed. So recursively mark all generated
+      // files as deleteOnExit so we at least clean them up when the JVM exits
+      os.walk.stream(tempDir).foreach { _.toIO.deleteOnExit() }
+
       // Summarize the result of compiling the schema for the test
       val compileResult = if (generator.isError) {
         Left(generator.getDiagnostics) // C code compilation diagnostics
@@ -193,11 +198,6 @@ class Runtime2TDMLDFDLProcessor(tempDir: os.Path, executable: os.Path)
   def unparse(parseResult: TDMLParseResult, outStream: java.io.OutputStream): TDMLUnparseResult = {
     unparse(parseResult.getResult, outStream)
   }
-
-  /**
-   * Remove the generated executable
-   */
-  override def cleanUp(): Unit = os.remove.all(tempDir)
 }
 
 final class Runtime2TDMLParseResult(pr: ParseResult) extends TDMLParseResult {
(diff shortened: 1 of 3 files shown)


mbeckerle (approved):
+1

This is definitely the way to go. Much less code by relying on the WeakHashTable behavior.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
```diff
@@ -113,6 +113,11 @@ final class Runtime2TDMLDFDLProcessorFactory private(
       // Compile the generated code into an executable
       val executable = generator.compileCode(codeDir)
 
+      // There is is no way to clean up TDMLDFDLProcessors, they are just garbage
+      // collected when no longer needed. So recursively mark all generated
+      // files as deleteOnExit so we at least clean them up when the JVM exits
+      os.walk.stream(tempDir).foreach { _.toIO.deleteOnExit() }
```
tuxji, line 119: Have you tested that this line actually cleans up all temporary files used by runtime2 tests?  If not, I'll still approve but I'll also check for left-over temporary files during my next work on runtime2 just in case.

In general, Daffodil tests aren't very good about removing temporary files.  I always see a lot of Daffodil-created temporary files in /tmp after running "sbt test" and/or "sbt IntegrationTest/test".

tuxji (approved):
+1

Nice improvement.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
stevedlawrence: > Have you tested that this line actually cleans up all temporary files used by runtime2 tests? If not, I'll still approve but I'll also check for left-over temporary files during my next work on runtime2 just in case.

Yep, I did actually test and verify that it works. Note that it's not enough to have delteOnExit() on just the root directory, it's needs to be on the directory and everything inside it.

> In general, Daffodil tests aren't very good about removing temporary files. I always see a lot of Daffodil-created temporary files in /tmp after running "sbt test" and/or "sbt IntegrationTest/test".

Yeah, we're really bad about this. I too have noticed how much we pollute /tmp. We really need to clean everything up, or at least put everything in /tmp/daffodil/ or something so everything is in one place.

stevedlawrence:
@jadams-tresys, can you test this to make sure this fixes the performance issues you are seeing?

jadams-tresys:
Yes, this fixes the performance issues I was seeing.  Haven't run enough tests to make sure that it is back to the same levels before the changes to the TDML cache, but it seems pretty close.

stevedlawrence:
Sounds good. I'll merge, we can reopen and reinvestigate if we find there are still some issues.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala
mbeckerle: We need a JIRA ticket for this. 

Honestly, the exact right idiom for creating and cleaning up temp files is not obvious to me. E.g., the fact that you can't just call deleteOnExit is news. 

We should put a temp-files section in our coding-style-guide on wiki.

PR #733, opened 2022-01-24
Ensure quasi elements for type calc expressions have a parent

stevedlawrence:
These quasi elements aren't really part of the infoset so a parent
element isn't actually needed, except for in some rare cases such as
when walking the infoset in the interactive debugger. There may also be
other edge cases we are unaware of. So ensure that these quasi type calc
elements have a parent, making it the same as its related simple
element.

DAFFODIL-2627

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala
@@ -156,6 +156,13 @@ class TypeValueCalcUnparser(typeCalculator: TypeCalculator, repTypeUnparser: Unp
     if (ustate.withinHiddenNest)
       tmpInfosetElement.setHidden()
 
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

A simple one-liner change :-).

mbeckerle (approved):
+1

tuxji:
I just did a git pull / git log ORIG_HEAD..HEAD and I noticed too late that two commits have the same JIRA number.  This PR's commit should have had the JIRA number [DAFFODIL-2630](https://issues.apache.org/jira/browse/DAFFODIL-2630) instead of [DAFFODIL-2627](https://issues.apache.org/jira/browse/DAFFODIL-2627).  Not sure whether it's feasible or worth it to amend the JIRA number since the commit is now HEAD~1 instead of HEAD, but letting you know.

stevedlawrence:
Yeah, we could rebase it, but I'm not sure it's worth it now that a handful of other commits have been pushed. I'll add a comment to DAFFODIL-2627 that the commit references the wrong bug number, so there's still a trail, it's just not as easy to follow.

PR #735, opened 2022-01-24
Use setVariable value expression to get referenced elements

stevedlawrence:
When trying to calculcate which elements are referenced via
dfdl:setVariable expressions, which is used for things like value and
content length, we incorectly used the default expression associated
with the dfdl:defineDariable instead of the actual value expression of
the dfdl:setVariable. This modifies the logic to use the correct
expression.

DAFFODIL-2629


mbeckerle (approved):
+1

tuxji (approved):
+1

PR #736, opened 2022-01-25
Disable capture value length parsers for all delimited simple types

stevedlawrence:
We already disable capture value length parsers for simple types with
representation="text" because the parsers for those text simple types
capture the value length themselve. This is needed because these parsers
have special logic for calculating the value length.

But simple types that aren't text (e.g. hex binary, packed decimal) but
are delimited use the same text parsers that capture value length. So
also disable capture length for delimited simple types. This ensures we
do not have multiple capture length logic executed for the same element
that may conflict.

DAFFODIL-2635

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
@@ -79,18 +79,23 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
         setElems.contains(this.dpathElementCompileInfo)
       }
 
-    // For simple elements with text representation, valueLength is captured in
-    // individual parsers since they handle removing delimiters and padding.
+    // For simple elements with text representation, value length is captured
+    // directly by the value parsers, since those parsers also handle removing
+    // delimiters and padding (which must be excluded from value length).
+    // Related, simple types that aren't text (e.g. hex binary, packed decimal)
+    // but are delimited also use the same simple text value parsers that
+    // capture value length.
     //
-    // For complex elements with specified length, valueLength is captured in
+    // For complex elements with specified length, value length is captured in
     // the specified length parsers, since they handle skipping unused
     // element regions. For complex elements, this means lengthKind is not
     // implicit or delimited.
     //
-    // So for these cases we do not want to capture value length, since
-    // they are handled by the parsers as necessary
+    // So for these cases we do not want to capture value length with the
+    // Capture{Start,End}OfValueLengthParsers, since those lengths are captured
+    // by the value parsers
     val capturedByParsers =
-      (isSimpleType && impliedRepresentation == Representation.Text) ||
+      (isSimpleType && (impliedRepresentation == Representation.Text || lengthKind == LengthKind.Delimited)) ||
         (isComplexType && (lengthKind != LengthKind.Implicit && lengthKind != LengthKind.Delimited))
 
     !capturedByParsers && isReferenced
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -81,6 +81,9 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
 
     // For simple elements with text representation, valueLength is captured in
     // individual parsers since they handle removing delimiters and padding.
+    // Related, simple types that aren't text (e.g. hex binary, packed decimal)
+    // but are delimited also use the same test delimited parsers that capture
```
mbeckerle, line 85: "the same test delimited parsers" ???

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -81,6 +81,9 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
 
     // For simple elements with text representation, valueLength is captured in
     // individual parsers since they handle removing delimiters and padding.
```
mbeckerle, line 83: This comment is ambiguous to me. 

"Capture in individual parsers" to me means "the separate capturing parsers", but we mean the opposite of that. 

How about "Capture directly by the value parsers".

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -90,7 +93,7 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
     // So for these cases we do not want to capture value length, since
     // they are handled by the parsers as necessary
     val capturedByParsers =
-      (isSimpleType && impliedRepresentation == Representation.Text) ||
+      (isSimpleType && (impliedRepresentation == Representation.Text || lengthKind == LengthKind.Delimited)) ||
```
mbeckerle, line 96: Is there no analogous flaw/situation for unparsers, or are they doing this stuff right?

mbeckerle (approved):
+1 Just some comments can be clarified a bit.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
stevedlawrence: should be "text delimited parsers", will fix

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
stevedlawrence: makes sense, will fix

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/ElementBaseRuntime1Mixin.scala
stevedlawrence: I think the complexity of unparsing actually makes it so we can completely rely on the CaptureValueLength unparsers--value unparsers don't actually have to do anything special.

For example, when parsing text, the string value parser handles both parsing the string AND removing padding, so the parser must also be responsible for calculating the value length that excludes that padding that it removed. So we can't use the special CaptureValueLength parsers in these cases.

But with unparsing, we have separate unparsers that unparse the value and unparse the padding, so we can wrap the CaptureValueLength unparsers around just value unparser. And this complexity of separate parsers is needed for all the suspension stuff, so that complex allows capture value length to be easy.

If we split parsing into separate parsers like we do for unparsing, we could probably get rid of this "captured by parsers" logic, and potentially get rid of a handful of potential bugs. Capturing value length in two different ways is just asking for issues like this. That might be a pretty sizable refactoring though.

tuxji (approved):
+1

Looks good with Mike's suggested comment fixes.

PR #738, opened 2022-01-31
Added test to confirm delimiter behavior.

mbeckerle:
DAFFODIL-2639

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
@@ -23,12 +23,9 @@ import java.nio.channels.Channels
 import java.nio.channels.ReadableByteChannel
 import java.nio.channels.WritableByteChannel
 import scala.util.Try
-import scala.xml._
 import org.apache.commons.io.output.NullOutputStream
 import org.junit.Assert.assertEquals
 import org.apache.daffodil.Implicits._
-
-import java.io.InputStream
 import org.apache.daffodil.api.DFDL
 import org.apache.daffodil.api._
 import org.apache.daffodil.compiler.Compiler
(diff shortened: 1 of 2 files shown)


stevedlawrence (approved):
+1

daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
```diff
@@ -371,3 +376,96 @@ class Fakes private () {
...
+  val pf = c.compileNode(schema)
+  val pfDiags = pf.getDiagnostics
+  if (pf.isError) throw new StreamParser.CompileFailure(pfDiags)
+  dp = pf.onPath("/")
+    .withValidationMode(ValidationMode.Full)
+  // .withDebuggerRunner(new TraceDebuggerRunner()) // DAFFODIL-2624 - cannot trace in streaming SAPI
+  // .withDebugging(true)
+  val dpDiags = dp.getDiagnostics
+  if (dp.isError) throw new StreamParser.CompileFailure(dpDiags)
+  var compilationWarnings = if (!pfDiags.isEmpty) pfDiags else dpDiags // dpDiags might be empty. That's ok.
```
tuxji, line 446: What's the reason for having compilationWarnings be a var instead of a val?  For that matter, what's the reason for splitting up the initialization via both the constructor and setInputStream when the only caller of this class is right above here in doStreamTest at line 416?  If you pass inputStream to the constructor, you can make dis and dp val's instead of var's too.

tuxji (approved):
+1

daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala
mbeckerle: Code was converted from Java, which is lazy and makes most things var. I've cleaned this up.

PR #739, opened 2022-01-31
Replace WeakHashMap TDML cache with a custom cache we have more control over

stevedlawrence:
Commit 4cc0863e71 modified the TDML compile result cache to use a
WeakHashMap. While this worked in some cases, it relied on the garbage
collector not being too aggressive. If the garbage collector
aggressively cleans up our WeakHashMap entries, or something like an IDE
frequently causes the GC to be run in between tests, it becomes
difficult to share compiled results between different DFDL Test Suites.
This leads to excessive schema compilation.

To fix this, this modifies to global TDML cache to be a normal HashMap,
with expiration times stored with each compile result entry. When a DFDL
Test suite finishes, it marks the compile results it uses with an
expiration time. Entries are removed form this cache when the expire
time is reached.

To minimize the size of this global cache, we also only store
non-embedded schemas in it. Embedded schemas cannot be shared between
different TDML files/DFDL Test suites, so putting them in the global
cache is unnecessary. Instead, compile results for embedded schemas are
stored in a cache local to the DFDL Test Suite. They will be garbage
collected when the DFDL Test Suite is garbage collected.

DAFFODIL-2627

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -461,7 +461,20 @@ class DFDLTestSuite private[tdml] (
   }
 
   /**
-   * Get the CompileResult from from the TDMLCompileResultCache.
+   * Get the CompileResult from a TDMLCompileResultCache.
+   *
+   * This uses one of two different caches to look for compile results:
+   *
+   * Embedded schemas cannot be shared among different DFDL Test Suites, so
+   * they are stored in a cache local to this test suite. This avoids having to
+   * run the expiration logic for such compile results, and the compile results
+   * will be garbage collected when this DFDL Test Suite is garbage collected.
+   *
+   * Non-embedded schemas may need to be shared between different test suites
+   * if different Runners/.tdml files reference the same schema with same
+   * compile options. So they are stored in a global cache, with care used to
+   * ensure we mark them as finished so they can expire and be removed when not
+   * used.
    */
   def getCompileResult(
     impl: AbstractTDMLDFDLProcessorFactory,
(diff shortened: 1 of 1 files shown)


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2585,64 +2613,85 @@ object UTF8Encoder {
...
+        key.tunables)
+      cache += (key -> TDMLCompileResultCacheValue(compileResult, None))
+      compileResult
     }
   }
 
 }
+
+object GlobalTDMLCompileResultCache {
+  val expireTime = 30
```
mbeckerle, line 2693: Add units of measure to name. expireTimeSeconds

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2585,64 +2613,85 @@ object UTF8Encoder {
...
-    optRootNamespace: Option[String],
-    tunables: Map[String, String],
-  )
+/**
+ * Stores the compile result as well as the time when this compile result
+ * should be removed from the cache. If optExpireTime is set to None, it means
+ * a DFDLTestSuite exists that is using this compile result, and so it should
+ * not expire. If optExpireTime is set to Some, it means no DFDLTestSuite
+ * currently uses the compile result, and so it can be expired once we reach
+ * the current time reaches the Some value time
```
mbeckerle, line 2631: I didn't see logic for this. I think also a successful cache hit should re-extend the time limit so instead of 30 seconds after the first compilation the schema needs to be recompiled it would instead be 30 seconds after the last usage of the compiled schema.

mbeckerle (approved):
+1 Two suggestions in comments.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: > I didn't see logic for this

`removeExpiredEntries` has the logic for for removing entries that have `Some` expire time and have reached that expire time. It keeps entries that do not have an expire time via the `getOrElse(true)`

`setEntriesFinished` has the logic for setting the expire time for keys that are finished (which is `now` + the expiration duration).

> I think also a successful cache hit should re-extend the time limit

We sort of do that. At line 2676, when we have a cache hit we set `v.optExpireTime = None`. So when we have a cache hit we disable expire time. Once the DFDLTestSuite finishes that had that cache hit, it will call `setEntriesFinished` and add a new expire time. There's a comment there, but maybe it isn't good enough? 

So the duration isn't actually from the last time there's a cache hit. It's from the latest time a DFDLTestSuite that used that compile result "finished". This way compile results will exist X seconds after the DFDLTestSuite ended, which should hopefully be enough for another test suite to start and start using it.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: Fixed

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: This is a good algorithm. I think just add this detail about the re-extending of the timeout to comments.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -461,7 +461,20 @@ class DFDLTestSuite private[tdml] (
...
   /**
-   * Get the CompileResult from from the TDMLCompileResultCache.
+   * Get the CompileResult from a TDMLCompileResultCache.
+   *
+   * This uses one of two different caches to look for compile results:
+   *
+   * Embedded schemas cannot be shared among different DFDL Test Suites, so
+   * they are stored in a cache local to this test suite. This avoids having to
+   * run the expiration logic for such compile results, and the compile results
+   * will be garbage collected when this DFDL Test Suite is garbage collected.
```
tuxji, line 471: Note that embedded schemas also are written to files in /tmp which should be deleted after the tests finish, but currently are not cleaned up.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2585,64 +2613,107 @@ object UTF8Encoder {
...
+/**
+ * Stores the compile result as well as the time when this compile result
+ * should be removed from the cache.
+ *
+ * If optExpireTimeMillis is set to None, it means a DFDLTestSuite exists that
+ * is using this compile result, and so it will not expire.
+ *
+ * If optExpireTimeMillis is set to Some, it means no DFDLTestSuite currently
+ * uses the compile result, and so it can be expired once we reach the current
+ * time reaches the Some value time
```
tuxji, line 2634: Comment has some unnecessary words, would read better as:

```
 * If optExpireTimeMillis is set to Some, it means no DFDLTestSuite currently
 * uses the compile result, and it can be expired once the current time reaches
 * the Some value time
```

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: That's just a bug, we really need to be cleaning up tmp files created in tests. I'll open a ticket.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: Agreed, will fix

PR #740, opened 2022-01-31
Fix Incorrect isError Scaladoc

olabusayoT:
- scaladoc described opposite behavior of isError

DAFFODIL-2641


stevedlawrence (approved):
+1

mbeckerle (approved):
+1

PR #742, opened 2022-02-01
Update daffodil-cli link syntax

brnnnfx:
This is a simple change to 
- make the URL appears as a hyperlink to be more consistent with other links 
- update the link to HTTPS

### DIFF
--- a/daffodil-cli/README.md
+++ b/daffodil-cli/README.md
@@ -81,6 +81,6 @@ users@daffodil.apache.org mailing lists. Bugs can be reported via the
 
 Daffodil is licensed under the [Apache License, v2.0](https://www.apache.org/licenses/LICENSE-2.0)
 
-This product includes The Saxon XSLT and XQuery Processor from
-Saxonica Limited (<http://www.saxonica.com/>), which is licensed under
+This product includes [The Saxon XSLT and XQuery Processor from
+Saxonica Limited](https://www.saxonica.com/), which is licensed under
 the Mozilla Public License version 2.0.


daffodil-cli/README.md
```diff
@@ -81,6 +81,6 @@ users@daffodil.apache.org mailing lists. Bugs can be reported via the
 
 Daffodil is licensed under the [Apache License, v2.0](https://www.apache.org/licenses/LICENSE-2.0)
 
-This product includes The Saxon XSLT and XQuery Processor from
-Saxonica Limited (<http://www.saxonica.com/>), which is licensed under
+This product includes [The Saxon XSLT and XQuery Processor from
+Saxonica Limited](https://www.saxonica.com/), which is licensed under
```
stevedlawrence, line 85: I found where this text came from in case it was taken from a license or something, in which case we might prefer the exact text. Apparently I came up with this wording in this PR https://github.com/apache/daffodil/pull/502 based on the Saxon licensing  page (https://www.saxonica.com/documentation/#!conditions) which says:

> If you produce a product that includes or requires Saxon-HE, please refer to it prominently as "The Saxon XSLT and XQuery Processor from Saxonica Limited", and include the URL of the home page, which is at http://www.saxonica.com/. As a courtesy, please take reasonable steps to ensure that your users know that they are running Saxon.

I think the new text still meets those requirements, and looks a bit better in the REAMDE.

stevedlawrence (approved):
+1 :+1:

PR #749, opened 2022-02-09
Fix fn:exists when unparsing non-blocking expressions

stevedlawrence:
Unparsing non-blocking expressions represent expression that are only
backwards referencing and will not require a suspension. This means that
the associated node must be final, and so we can immediately respond
that the element exists or not. Rethrowing the associated exception not
correct in this case. That is only correct when suspensions will be
involve, in which case that exception triggers the suspension logic.

DAFFODIL-2648

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/FNFunctions.scala
@@ -553,16 +553,20 @@ trait ExistsKind {
     Assert.invariant(dstate.currentNode ne null)
     val res = dstate.mode match {
       case UnparserNonBlocking => {
-        // we are evaluating the expression, and hoping it will evaluate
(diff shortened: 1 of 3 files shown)


mbeckerle:
Could fn:count() have the same issue? Requires the array is finished.

stevedlawrence:
`fn:count()` uses this same `ExistsKind` trait, so I think this fix for `fn:exists()` should also fix `fn:count()`. I'll add a test to make sure.

mbeckerle:
@olabusayoT This fix also addresses fn:count() which the big schema you are working on is using. 
Can you review this PR so we can merge it?

olabusayoT:
+1 This fixed the isFinished error that was occurring in my large schema during unparsing.

olabusayoT (approved):

olabusayoT:
I actually misread my run. I'm still getting the error, but the PR looks good so I'm keeping the approval

tuxji (approved):
+1

Agreed, looks good.

PR #751, opened 2022-02-11
Modify the release candidate container to work for both daffodil and vscode

stevedlawrence:
- When running the release candidate script, prompt the user for either
  "daffodil" or "daffodil-vscode" to determine which repository to
  release. This changes which repo is checked out, how the repo is
  built, and which files are copied to the dist directory. All other
  automated build steps are common and are unchanged.
- Daffodil release files are still installed to dist.apache.org/repos/dist/dev/daffodil/
- VS Code release files are installed to dist.apache.org/repos/dist/dev/daffodil/daffodil-vscode

Also did some cleanup to the container script:

- Move wix files and Docker setup/entrypoint scripts out of /root and
  into /opt and /usr/bin, respectively. With this change, the only files
  in /root are build files, making it easier to verify release artifacts
- Add the --rm option to podman run. This container is short lived and
  does not need to exist once it finishes

DAFFODIL-2655

### DIFF
--- a/containers/release-candidate/daffodil-release-candidate
+++ b/containers/release-candidate/daffodil-release-candidate
@@ -48,8 +48,29 @@ USAGE
 
 DRY_RUN=false
 
+echo "Which project to release?"
+select PROJECT_REPO in daffodil daffodil-vscode
+do
+  case $PROJECT_REPO in
+    "daffodil")
+      PROJECT_DIST_DIR=""
+      PROJECT_NAME="Daffodil"
+      break
+      ;;
+    "daffodil-vscode")
+      PROJECT_DIST_DIR="$PROJECT_REPO"
+      PROJECT_NAME="Daffodil VS Code Extension"
+      break
+      ;;
+    *)
+      echo "unknown project: $REPLY" >&2
+      exit 1
+      ;;
+    esac
+done
+
 DAFFODIL_CODE_USER="apache"
-DAFFODIL_CODE_REPO="daffodil"
+DAFFODIL_CODE_REPO="$PROJECT_REPO"
 DAFFODIL_CODE_BRANCH="main"
 
 DAFFODIL_SITE_REPO="daffodil-site"
@@ -91,7 +112,7 @@ if [ $? -ne 0 ]; then
 fi
 
 export GPG_TTY=$(tty)
-export WIX=/root/wix311/
+export WIX=/opt/wix311/
 export LANG=en_US.UTF-8
 export CC=clang
 export AR=llvm-ar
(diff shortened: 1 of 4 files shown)


containers/release-candidate/daffodil-release-candidate
```diff
@@ -183,52 +212,81 @@ DAFFODIL_TUTORIALS_DIR=$DAFFODIL_SITE_DIR/site/tutorials
...
+    ;;
+
+  "daffodil-vscode")
 
-echo "Embedding RPM Signature..."
-rpmsign --define "_gpg_name $PGP_SIGNING_KEY_ID" --define "_binary_filedigest_algorithm 10" --addsign $DAFFODIL_RELEASE_DIR/bin/*.rpm
+    echo "Building Convenience Binaries..."
+    mkdir -p $DAFFODIL_RELEASE_DIR/bin/
+    yarn build
+    cp daffodil-debugger-*.vsix $DAFFODIL_RELEASE_DIR/bin/
```
tuxji, line 285: Can do the rename to apache-daffodil-vscode-VERSION.vsix (as suggested in [#83](https://github.com/apache/daffodil-vscode/issues/83)) only here, or also modify the yarn build to produce the renamed binary.

tuxji (approved):
+1

containers/release-candidate/daffodil-release-candidate
stevedlawrence: I forgot about the potential rename.

I'd rather not rename it here if it can be done in the vscode repo. I could imagine the original name of 'daffodil-debugger' might show up some where and confuse users if the original file they actually installed was something else (like the suggested apache-daffodil-vsocde name).

But I think I'll change this to `cp *.vsix $DAFFODIL_RELEASE_DIR/bin/`. There should only ever be one .vsix file so no issues with globbing too much, and this allows it to be named whatever the project thinks makes sense.

stevedlawrence:
Once this change is committed, this container will not really be specific to daffodil anymore. I wonder if it makes sense to move the container directory to its own repository (e.g. `apache/daffodil-release-candidate`)? That way, if we ever need to make changes to the container for the vscode repo we don't have to make changes to this somewhat unrelated repo?

tuxji:
I don't think a separate repo makes sense (`apache/daffodil-release-candidate`).  I have 3 reasons: 1) we're talking about moving only 5 files which hardly justifies a separate repo at the Apache organization level; 2) the path `daffodil/containers/release-candidate` is much easier to understand than the repo name `daffodil-release-candidate`; 3) newcomers will feel more confused and waste more time trying to understand 3 daffodil-* repos than only 1 to 2 daffodil repos.

stevedlawrence:
Fair points. I don't feel strong about this and am fine with keeping it in here.

Here's some additional thoughts in the back of my head, but which I don't think are all that strong of arguments since they're very future stuff and probably doesn't really apply right now:

1. This is *almost* generic enough that it could be used for other apache projects. I sort of envision this repo growing to be a possible starting point for other projects to create their own release candidates. I know it was a pain to come up with all this stuff in the past, and it would be a shame for other new ASF projects to have to do something similar. That said, I doubt I'll want to do the work to make this truely generic. But if it were in a separate repo, others might.

2. I could imagine a case where the daffodil repo is on some sort of code freeze but we want to make changes to the release candidate stuff for vscode (or maybe some other future project). By decoupling this release candidate container from any particular repo, it's easier to make those changes. That's not an issue now because we don't really code freeze, except maybe unofficially near releases, which aren't common.

And maybe a potential argument against:

1. It now requires a bit more effort to keep a third repo in sync. When building a release, we must ensure that the release manager uses the latest daffodil-relase-candidate repository or there might be subtle build issues. Theoretically either the build would fail, or it would be detected as part of the VOTE, so maybe not too big of a deal. But is another point of failure. That said, it's maybe not too different from our current method, where you should build from the latest Daffodil commit. For example, people that primarily do daffodil-vscode development have to clone the daffodil repo to do a release. Not a big deal, but might be a repo they don't always keep up to date.

mbeckerle:
I am ok with leaving it in Daffodil's source tree for now. That works because these release scripts/containers aren't actually part of a release. They're not distributed.

stevedlawrence:
This still needs another +1

mbeckerle (approved):
+1 
Proof will be in the pudding of actually using this to create releases. It all looks good.

PR #754, opened 2022-02-14
WIP: Call setFinished on the correct data output stream

jadams-tresys:
When splitting the data output stream during unparse for a bitOrder
change, we had been setting the incorrect DOS to finished. We need to
call dos.setFinished(UState) before we update the UState's
dataOutputStream to the new split DOS.

DAFFODIL-2653

This is WIP as I do not have a test case that replicates the issue, see DAFFODIL-2653 for a way to reproduce the issue.

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
@@ -327,8 +327,6 @@ abstract class UState(
     }
     if (isSplitNeeded) {
       Assert.invariant(dos.isBuffering) // Direct DOS always has absolute position, so has to be buffering.
-      val newDOS = dos.addBuffered
-      dataOutputStream = newDOS
       //
       // Just splitting to start a new bitOrder on a byte boundary in a new
       // buffered DOS
@@ -344,7 +342,24 @@ abstract class UState(
       // Finished means you won't add data to the end of it any more.
       // It does NOT prevent information like the absoluteBitPos to
       // propagate.
-      dos.setFinished(this)
+      //
+      // When setFinished is called the DOS is going to store the state that we
+      // pass into it in finishedFormatInfo. Eventually this DOS will become a
+      // direct DOS that may be delivered to a following buffered DOS. When that
+      // happens this saved finishedFormatInfo will be used. However, this
+      // requires that the UState does not change while we are waiting for this
+      // DOS to become direct. If the state does change, it will become
+      // incorrect and can lead to undefined behavior. To prevent this, we must
+      // clone the UState so it can no longer change, and pass that clone into
+      // setFinished.
+      val finfo = this match {
+        case m: UStateMain => m.cloneForSuspension(dos)
+        case _ => Assert.invariantFailed("State must be a UStateMain when splitting for bit order change")
+      }
+
+      val newDOS = dos.addBuffered
+      dataOutputStream = newDOS
+      dos.setFinished(finfo)
     }
   }
 


jadams-tresys:
I have been unable to figure out how to create a test case that replicates the issue, however test_ep2 in ./daffodil-test/src/test/resources/org/apache/daffodil/unparser/envelopePayload.tdml appears to be very similar, using a choice at the point where there is a bitOrder change, but it still does not replicate the issue described in DAFFODIL-2653 or the infinite loop when not cloning the UState.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -344,7 +342,21 @@ abstract class UState(
       // Finished means you won't add data to the end of it any more.
       // It does NOT prevent information like the absoluteBitPos to
       // propagate.
-      dos.setFinished(this)
+      //
+      // Need this clone or we could possibly end up in an infinite loop inside of
+      // DirectOrBuffereDataOutputStream.cleanUp()
```
stevedlawrence, line 347: I think there's a better explanation for why we need a clone of the UState, maybe something like this:

> When setFinished is called, the DOS is going to store the state that we pass into it in finishedFormatInfo. At some point in the future, this DOS will become a direct DOS that may be delivered to a following buffered DOS. When that happens, this saved finishedFormatInfo will be used. However, this all requires that the UState does not change while we are waiting for this DOS to become direct. If the state does change, it will be incorrect and can lead to undefined behavior. To prevent this, we must clone the UState so it can no longer change, and pass that clone into setFinished.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -344,7 +342,21 @@ abstract class UState(
...
+      val finfo = this match {
+        case m: UStateMain => m.cloneForSuspension(dos)
+        case _ => this
```
stevedlawrence, line 350: I wonder if this case is even possible? If we are adding a new buffered DOS below, I think that means that this must be a `UStateMain`--we can't split a DOS that is associated with an already cloned UState since that means the UState is from a suspension which must have already been split.  I think `this` *must* be a `UStateMain` and that it *must* be cloned. I think we can just change `this` to something like `Assert.invariantFailed("state must be a UStateMain when splitting for bit order change")`

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -344,7 +342,21 @@ abstract class UState(
...
+      }
+
+      val newDOS = dos.addBuffered
+      // NOTE: We must call set finished on the original dos with the
+      // UState/finfo before we set this.dataOutputStream to the newly split
+      // DOS, otherwise the call to this.dataOutputStream = newDOS will update the
+      // formatInfo to point to the new DOS when we try to call setFinished.
+      dos.setFinished(finfo)
+      dataOutputStream = newDOS
```
stevedlawrence, line 359: I think the order that this happens as mentioned in the comment no longer matters. As long as we call `cloneForSuspension` before we do `dataOutputStream = newDOS` and modify this state this should all work as expected. This is because `setFinished` will use our clone so changing the `dataOutputStream` of this `UState` no longer has any affect like it does now (which was part of the issue). I'm not sure if one order is preferred over the other, but I don't think the comment is needed if we add something similar to the comment mentioned above.

stevedlawrence (approved):
+1

stevedlawrence:
> I have been unable to figure out how to create a test case that replicates the issue

There's clearly a lot of complexity to get to this error. Not only do we need a split with some unknown byte order/bit orer stuff, we also need some suspensions involved and that those suspensions run and finish at specific times, and for state to be modified in specific ways after splits occur. So I'm not sure how easy it is going to be to create a small reproducible test. We might just need someone with access to confirm that the private test no longer fails, that nothing else is broken, and by code inspection that it seems correct.

mbeckerle:
The way suspensions are resolved makes this hard to reproduce. It is deterministic, but not predictable from the DFDL schema. 

That is something to think about. 

In past lives I've dealt with this sort of thing and even non-deterministic things, by forcing deterministic but quasi-random schedules. E.g. by using a pseudo-random schedule changer with fixed start seed. 

If enabled this would, in our case, permute the processing of the suspensions by deferring the processing of a suspension based on a coin-flip of this pseudo-random coin.  E.g., one way would be with 50/50 chance that a suspension would be skipped over, and delayed by one (i.e., exchange head and head-1 of the queue with 50/50 probability). 

Then one could run a small test in a loop repeatedly, and it would explore many permutations of the suspension evaluation orderings.

mbeckerle (approved):
+1
If we have more issues like this we will have to get more clever about testing.

PR #757, opened 2022-02-17
Improve diagnostic for no forward progress

jadams-tresys:
Improved diagnostic message now contains reference to the array and
describes how it is stuck in an infinite loop.

DAFFODIL-2576

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
@@ -57,12 +57,13 @@ abstract class SequenceParserBase(
 
   final protected def checkForwardProgress(
     pstate: PState,
+    parser: RepeatingChildParser,
     currentPos: Long,
     priorPos: Long,
     ais: ArrayIndexStatus): ArrayIndexStatus = {
     Assert.invariant(currentPos >= priorPos)
     if (currentPos == priorPos && pstate.groupPos > 1) {
-      PE(pstate, "No forward progress.")
+      parser.PE(pstate, "Array element parsed succesfully, but consumed no data and is stuck in an infinite loop as it is unbounded.")
       Done
     } else {
       ais
@@ -173,7 +174,7 @@ abstract class SequenceParserBase(
                 // result of try could also be absent if we just ended a group
                 // by not finding a separator
                 //
-                ais = checkForwardProgress(pstate, currentPos, priorPos, ais)
+                ais = checkForwardProgress(pstate, parser, currentPos, priorPos, ais)
               }
               //
               // advance array position.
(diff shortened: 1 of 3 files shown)


mbeckerle (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceParserBases.scala
```diff
@@ -62,7 +62,7 @@ abstract class SequenceParserBase(
     ais: ArrayIndexStatus): ArrayIndexStatus = {
     Assert.invariant(currentPos >= priorPos)
     if (currentPos == priorPos && pstate.groupPos > 1) {
-      PE(pstate, "No forward progress.")
+      PE(pstate, "Error parsing array '%s': Array element parsed succesfully, but consumed no data and is stuck in an infinite loop as it is unbounded.".format(srd.groupMembers(srd.position-1)))
```
stevedlawrence, line 65: Is this part correct?
```scala
srd.groupMembers(srd.position-1))
```
Isn't `srd.position` the position of this sequence in its list of siblings, and not the position of the array that we want to print?

I'm not sure this function has the information to know which is the right child of the sequence that failed to make progress. I *think* the caller of `checkForwardProgress` has that information though. Maybe something like `parser.erd`, since `parser` in that context is the thing that actually parsed an array element but failed to make progress?

stevedlawrence: If the above is correct (I'm still not totally sure), maybe what we really want is to pass in `parser`, and then call `parser.PE(...)`. That way the context of the PE would be on that of the array parser and not of the sequence. And with the right context, you might not even need to include the array name in the error message since it should be included in the context?

If that's true, it might be worth adding the array name to the `<tdml:error>` so that we make sure we get the expected context.

stevedlawrence: Interesting, I'm not sure if that's a bug or not...

It looks like the reason still getting the sequence as the context is that `RepeatingChildParser` extends `SequenceChildParser`, which extends `CombinatorParser` and passes in `srd` as the context. So when we do `parser.PE`, the context of that parser is still that of the sequence, even though this particular parser is really all about the the child. Maybe we should change `SequenceChildParser` so that is passes in `trd` to the `CombinatorParser` instead of `srd`? I think that feels more correct to me, since I would expect the context of a `SequenceChildParser` to be the child and not the Sequence? And then you wouldn't need diagnosticDebugName?

jadams-tresys: Passing the trd to CombinatorParser did the trick.  New commit inbound

tuxji (approved):
+1, although Steve hasn't reviewed a second time yet.

stevedlawrence:
Oops, +1 to the changes :+1:

PR #762, opened 2022-02-23
Change output of dfdlx:trace from stderr to the logger

alexanderrevello:
…ted tests


stevedlawrence (approved):
+1

stevedlawrence:
@alexanderrevello, looks good, can you please add the bug number (DAFFODIL-2575) to the end of the commit message, and update the commit message to be more accurate? I think "restore unneeded input file" doesn't really make sense in the context of just this single commit. See Step 10 of the [Code Contributor Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow) for commit message guidance.

Also, I just noticed your git commit name is `U-COLUMBIA\arevello` and email is `arevello@N12157207.columbia.tresys.com` You don't have to change this, but this looks like some default set by git. Might be worth updating this since you'll need to amend the commit to add the bug number anyway.

Once fixed, you can force push and this should be good to merge :+1:

PR #766, opened 2022-03-02
Update Apache Daffodil KEYS

Shanedell:


KEYS
```diff
@@ -289,3 +289,24 @@ HKqTiP7RdoFIle21qaGNI96V2gUCFajt+znArVsF14yWI5lX9ZKdgjeB2NPDxYkM
 DPh7WksEmrAvbFk5HLWlENsXIoig2Tc=
 =qmBp
 -----END PGP PUBLIC KEY BLOCK-----
+
+pub   ed25519 2022-03-02 [SC] [expires: 2024-03-01]
```
tuxji, line 293: What kind of key is this?  It looks smaller than a RSA key with keysize 4096.

stevedlawrence: Looks like it's ED25519. ASF documentation recommends RSA 4096, but doesn't explicitly say not to use ED25519. Security-wise I imagine it's fine, but I wonder if maybe there are potential issues with support for older version of gpg?

stevedlawrence: Also, it looks like the key has an expiration date. ASF doesn't require an expiration--it might make your life easier in the future to not have an expiration date.

Shanedell: Okay this makes sense I will regenerate using RSA 4096 no expiration

tuxji (approved):
+1

Yes, please delete your branch after merging it.

PR #768, opened 2022-03-07
Daffodil 2663 pluggable charsets

alexanderrevello:


daffodil-test/src/test/scala/org/apache/daffodil/charsets/CustomJavaCharset.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package org.apache.daffodil.charsets
+
+import org.apache.daffodil.processors.charset.BitsCharsetJava
+import org.apache.daffodil.processors.charset.BitsCharsetDecoderByteSize
+import org.apache.daffodil.io.InputSourceDataInputStream
+import org.apache.daffodil.io.FormatInfo
+
+final class CustomJavaCharset(
+  name2: String)
```
mbeckerle, line 27: Strange indentation. 

When there are only a small number of args, we tend not to do the thing where args each get a line, as it reduces code density for no adantage. At 3+ args, sure, because it will line wrap anyway. But just 1 or 2 args keep args on one line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/charsets/CharsetCompilerRegistry.scala
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Given name, finds the factory for the transformer. SDE otherwise.
+   *
+   * The state is passed in order to provide diagnostic context if not found.
+   */
+  def find(name: String, context: ThrowsSDE): CharsetCompiler = {
+    val maybeCompiler: Option[CharsetCompiler] = charsetCompilerMap.get(name)
+    if (maybeCompiler.isEmpty) {
+      val choices = charsetCompilerMap.keySet.mkString(", ")
+      context.SDE("The charset '%s' was not found. Available choices are: %s", name, choices + ", " + CharsetUtils.supportedEncodingsString)
```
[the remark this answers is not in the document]
tuxji: Mike also said he would like Daffodil's current charsets re-implemented as pluggable charsets, so you'll probably need to remove `CharsetUtils.supportedEncodingsString` from this line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/charsets/CharsetTransformerFactory.scala
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This is the serialized object which is saved as part of a processor.
+ * It constructs the charset transformer at runtime when newInstance() is called.
+ *
+ * This must be implmented as part of implementation of a daffodil charset.
+ */
+abstract class CharsetTransformerFactory(val name: String)
+  extends Serializable {
+
+  def newInstance(): CharsetTransformer
```
tuxji, line 32: Mike said `CharsetTransformer` should simply be `BitsCharset`, so you probably want to rename `CharsetTransformerFactory` to `BitsCharsetFactory`.

tuxji, line 32: Mike said `CharsetTransformer` should simply be `BitsCharset`, so you probably want to rename `CharsetTransformerFactory` to `BitsCharsetFactory`.

daffodil-test/src/test/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
```diff
@@ -0,0 +1,31 @@
...
+# "real" in the sense of being fully featured. E.g., AIS only does the payload
+# armoring, not everything else about AIS. IPv4 just does the IPv4 header, not
+# TCP, UDP, ICMP, etc. all of which have related checksum algorithms that should
+# share code. A real CheckDigits has to have different layer length kinds supported
+# so that one can compute check digits from elements with different
+# of length kinds.
+#
+# These are partial/subsets of "real" functionality that give us test coverage
+# of the layering features of Daffodil. They can be reused to create "real"
+# layering, but one should expect to have to adapt their code substantially.
```
tuxji, line 28: Those comments were pasted here from another place and don't belong here, or should be updated to talk only about your test charsets.

tuxji, line 28: Those comments were pasted here from another place and don't belong here, or should be updated to talk only about your test charsets.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.dfdl.xsd
```diff
@@ -0,0 +1,69 @@
...
+  xmlns:tns="urn:org.apache.daffodil.charsets.CustomCharset"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+  <include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format ref="tns:GeneralFormat" initiator="" terminator="" leadingSkip="0" trailingSkip="0" textBidi="no" floating="no" encoding="ASCII" byteOrder="bigEndian" alignment="implicit" alignmentUnits="bits" fillByte="f" occursCountKind="implicit" truncateSpecifiedLengthString="no" ignoreCase="no" representation="text" lengthKind="delimited" nilValueDelimiterPolicy="both" emptyValueDelimiterPolicy="none" documentFinalTerminatorCanBeMissing="yes" initiatedContent="no" separatorSuppressionPolicy="anyEmpty" separatorPosition="infix" textTrimKind="none" />
```
tuxji, line 29: Please override only properties that really need to be different than GeneralFormat's properties, and define them on multiple lines for readability.

tuxji, line 29: Please override only properties that really need to be different than GeneralFormat's properties.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,151 @@
...
+<tdml:testSuite xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:ex="http://example.com"
+  xmlns:tns="urn:org.apache.daffodil.charsets.CustomCharset" defaultRoundTrip="none">
+
+  <tdml:parserTestCase name="parse_charsets" root="s1" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd" roundTrip="none">
```
tuxji, line 27: We already have `defaultRoundTrip="none"` above so we don't need `roundTrip="none"` in each test.

tuxji, line 27: We already have `defaultRoundTrip="none"` above so we don't need `roundTrip="none"` in each test.

tuxji (changes requested):
Mostly amplifying Mike's suggestions for improving the pull request, including re-implementing Daffodil's own charsets as pluggable charsets.

daffodil-test/src/test/scala/org/apache/daffodil/charsets/ISO88591Reverse.scala
```diff
@@ -0,0 +1,48 @@
...
+  override val name = "X-DFDL-ISO-88591-8-BIT-PACKED-LSB-FIRST-REVERSE"
+  override val bitWidthOfACodeUnit = 8
+  override val decodeString = (0 to 255).map { _.toChar }.mkString.reverse
+  override val replacementCharCode = 0x0
+  override val requiredBitOrder = BitOrder.MostSignificantBitFirst
+} with BitsCharsetNonByteSize
+
+
+final class BitsCharsetISO88591ReverseCompiler
+  extends CharsetCompiler("X-DFDL-ISO-88591-8-BIT-PACKED-LSB-FIRST-REVERSE") {
```
mbeckerle, line 35: add hyphen 8859-1

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,151 @@
...
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:ex="http://example.com"
+  xmlns:tns="urn:org.apache.daffodil.charsets.CustomCharset" defaultRoundTrip="none">
+
+  <tdml:parserTestCase name="parse_charsets" root="s1" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd">
+
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <s1>
+          <e1>À1234567</e1>
```
mbeckerle, line 32: Add comment that first character is Unicode Code point 0xC0, LATIN CAPITAL LETTER A WITH GRAVE. Just in case someone is lacking a font that displays it.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,151 @@
...
+
+    <tdml:document>
+      <tdml:documentPart type="byte"><![CDATA[C0 31 32 33 34 35 36 37
+                                          CF CE CD CC CB CA C9 C8]]>
+      </tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <tns:s1>
+          <e1>Ą1234567</e1>
```
mbeckerle, line 55: Add comment that first character is Unicode 0x104, LATIN CAPITAL LETTER A WITH OGONEK.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,151 @@
...
+    </tdml:infoset>
+  </tdml:unparserTestCase>
+
+  <tdml:parserTestCase name="parse_charsets2" root="s2" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd">
+
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <s2>
+          <e1>À1234567</e1>
+          <e2>À1234567</e2>
```
mbeckerle, line 68: First char inside <e2> should be Ą once you have a real 8859-13 decoder, which I suggest you fix below in comments on the scala code for this charset.

mbeckerle (changes requested):
Getting very close. I have only a few more changes I'd like to see.

mbeckerle:
> This now includes commits that are not really part of this change set, but they are showing up as commits in this review.
> 
> That and... github UI is having trouble. I can't seem to narrow the review to just the last 3 commits. It doesn't seem to allow me to select that. (I can select 2 commits, but not 3)
> 
> The exact best way to fix this isn't clear. I think the first 2 and last 5 commits should be in this change set, the other commits shouldn't be part of this.

Note to other reviewers - this got fixed - it's now just the right set of commits.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetTransformerFactory.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.daffodil.processors.charset
+
+/**
+ * Factory for a charset transformer.
+ *
+ * This is the serialized object which is saved as part of a processor.
+ * It constructs the charset transformer at runtime when newInstance() is called.
+ *
+ * This must be implmented as part of implementation of a daffodil charset.
```
tuxji, line 27: Suggest rewriting as

```
 * Factory for a Daffodil charset.
 *
 * This is the serialized object which is saved as part of a processor.
 * It constructs a charset at runtime when newInstance() is called.
 *
 * This must be implemented as part of implementing a Daffodil charset.
 ```

File: daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompiler.scala [hunk 12]
```diff
@@ -0,0 +1,35 @@
+
```
tuxji, line 1: Please rename this file from 'CharsetCompiler.scala' to 'BitsCharsetDefinition.scala' and remove the blank first line.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompiler.scala
```diff
@@ -0,0 +1,35 @@
...
+ * These instances are NOT serialized as part of a saved processor. The compileCharset method
+ * is called and the resulting CharsetTransformerFactory is the serialized object.
+ */
+abstract class CharsetCompiler(nom: String) {
+
+  def name() = nom
+
+  def compileCharset(): BitsCharsetFactory
+
+}
```
tuxji, line 35: Suggest writing as follows

```scala
/**
 * Defines a class which must be dynamically loaded in order to add a charset implementation
 * to Daffodil.  All charsets must implement this class.
 *
 * A saved processor does NOT serialize this class. It calls the newFactory method and serializes
 * the resulting BitsCharsetFactory.
 */
abstract class BitsCharsetDefinition(nom: String) {
  def name() = nom

  def newFactory(): BitsCharsetFactory
}
```

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompilerRegistry.scala
```diff
@@ -0,0 +1,57 @@
+/*
```
tuxji, line 1: Suggest renaming this class from `CharsetCompilerRegistry.scala` to `BitsCharsetDefinitionRegistry.scala`.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompilerRegistry.scala
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.daffodil.processors.charset
+
+import org.apache.daffodil.exceptions.ThrowsSDE
+import org.apache.daffodil.util.SimpleNamedServiceLoader
+
+object CharsetCompilerRegistry {
```
tuxji, line 22: Likewise, rename `CharsetCompilerRegistry` to `BitsCharsetDefinitionRegistry`.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompilerRegistry.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  private lazy val charsetCompilerMap: Map[String, CharsetCompiler] =
```
tuxji, line 24: Change to

```scala
  private lazy val definitionMap: Map[String, BitsCharsetDefinition] =
    SimpleNamedServiceLoader.loadClass[BitsCharsetDefinition](classOf[BitsCharsetDefinition])
```

Update rest of file likewise.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetCompilerRegistry.scala
```diff
@@ -0,0 +1,57 @@
...
+      val choices = charsetCompilerMap.keySet.mkString(", ")
+      SDE("The encoding '%s' was not found. Available choices are: %s", name, choices)
+    } else {
+      optCompiler.get
+    }
+  }
+
+      def SDE(id: String, args: Any*): Nothing = {
+      throw new Exception(id.format(args: _*))
+    }
```
tuxji, line 54: Fix indentation.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
```diff
@@ -33,11 +33,10 @@ object CharsetUtils {
    * encodings as well as the standard ones.
    */
   def getCharset(name: String): BitsCharset = {
-    val cs = DaffodilCharsetProvider.charsetForName(name)
-    cs
+    CharsetCompilerRegistry.find(name).compileCharset.newInstance()
```
tuxji, line 36: Will become

```scala
    BitsCharsetDefinitionRegistry.find(name).newFactory().newInstance()
```

Not sure if Scala conventions would keep empty parentheses or not.

daffodil-runtime1/src/main/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
```diff
@@ -0,0 +1,31 @@
...
+# share code. A real CheckDigits has to have different layer length kinds supported
+# so that one can compute check digits from elements with different
+# of length kinds.
+#
+# These are partial/subsets of "real" functionality that give us test coverage
+# of the layering features of Daffodil. They can be reused to create "real"
+# layering, but one should expect to have to adapt their code substantially.
+#
+org.apache.daffodil.processors.charset.CustomNonByteSizeCharsetCompiler
+org.apache.daffodil.processors.charset.CustomJavaCharsetCompiler
```
tuxji, line 31: Rename file from `CharsetCompiler` to `BitsCharsetDefinition`.  Replace `Compiler` with `Definition`.

daffodil-test/src/test/resources/META-INF/services/org.apache.daffodil.processors.charset.CharsetCompiler
```diff
@@ -0,0 +1,18 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+org.apache.daffodil.charsets.BitsCharset_ISO_8859_1_Reverse_Compiler
+org.apache.daffodil.charsets.BitsCharsetTest_ISO_8859_13_Compiler
```
tuxji, line 18: Rename file from CharsetCompiler to BitsCharsetDefinition. Replace Compiler with Definition.

tuxji (changes requested):
I think your pull request looks pretty close to completion.  I have proposed a new naming strategy that would make your pull request even better.  If the new naming strategy makes sense, the changes won't take too much time if you use your IDE's refactoring support to make the changes easier to do.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/AISPayloadArmoring.scala
```diff
@@ -44,3 +44,19 @@ object BitsCharsetAISPayloadArmoring extends {
   override val replacementCharCode = 0x30
   override val requiredBitOrder = BitOrder.MostSignificantBitFirst
 } with BitsCharsetNonByteSize
+
+final class BitsCharsetAISPayloadArmoringCompiler
+  extends CharsetCompiler("X-DAFFODIL-AIS-PAYLOAD-ARMORING") {
+
+  override def compileCharset() = {
+    new BitsCharsetAISPayloadArmoringTransformerFactory(name)
```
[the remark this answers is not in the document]
mbeckerle: Keep in mind if BitsCharsetAISPayloadArmoringFactory can ever fail and issue a diagnostic, the schema author/user knows this as "X-DAFFODIL-AIS-PAYLOAD-ARMORING", and that's the meaningful name to them for use in a diagnostic. That's the only reason I can think of for passing the name.

daffodil-runtime1/src/main/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
mbeckerle: At this point I'm pretty convinced by @tuxji naming suggestions that those really do improve the readability and clarity of this code. Compiler and Transformer are holdovers from the layering stuff. They make little sense for charsets.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,154 @@
...
+  xmlns:ex="http://example.com"
+  xmlns:tns="urn:org.apache.daffodil.charsets.CustomCharset" defaultRoundTrip="none">
+
+  <tdml:parserTestCase name="parse_charsets" root="s1" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd">
+
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <s1>
+          <!-- First char in e1 is unicode char 0xC0 captial A with grave --> 
+          <e1>Ą1234567</e1>
```
mbeckerle, line 33: You have the comment for the wrong char. This one is the 0x104 captial A with Greek Ogonek. (Ą)

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,154 @@
...
+        </tns:s1>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:unparserTestCase>
+
+  <tdml:parserTestCase name="parse_charsets2" root="s2" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd" roundTrip="true">
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <s2>
+          <!-- First char in e1 is unicode char 0x104 captial A with Greek Ogonek -->
```
mbeckerle, line 68: I think you have these two comments backward. This first one is the A with Grave.

daffodil-test/src/test/resources/org/apache/daffodil/charsets/CustomCharset.tdml
```diff
@@ -0,0 +1,154 @@
...
+  </tdml:unparserTestCase>
+
+  <tdml:parserTestCase name="parse_charsets2" root="s2" model="org/apache/daffodil/charsets/CustomCharset.dfdl.xsd" roundTrip="true">
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <s2>
+          <!-- First char in e1 is unicode char 0x104 captial A with Greek Ogonek --> 
+          <e1>À1234567</e1>
+          <!-- First char in e2 is unicode char 0xC0 captial A with grave -->
+          <e2>Ą1234567</e2>
```
mbeckerle, line 71: This second one is the A with greek ogonek.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/IBM037.scala
```diff
@@ -55,3 +55,27 @@ class BitsCharsetDecoderIBM037
     dec
   }
 }
+
```
mbeckerle, line 58: Above. We could define the big table of magic numbers as:
```
  val decodeString = {
    val bytes = ByteBuffer.wrap((0 to 255).map{ _.toByte }.toArray)
    Charset.forName("ebcdic-cp-us").newDecoder().decode(bytes).toString
  }
```
Or we could do this in a separate cleanup pull request also.

mbeckerle (changes requested):
Couple buggy comments in tests (you have your Grave accents and Greek Ogoneks backward :-)

I think the renaming @tusji suggested will be a big improvement as well, so is very worth doing.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/AISPayloadArmoring.scala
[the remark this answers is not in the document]
alexanderrevello: name() is required for the SimpleNamedServiceLoader, however it does not need to be passed, the implementing class can just set it. I will change it

mbeckerle (approved):
+1 This is an excellent change set.

tuxji (changes requested):
Excellent progress!  Still needs some changes to ensure error paths are handled appropriately, though.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDefinition.scala
```diff
@@ -0,0 +1,34 @@
...
+ * to Daffodil. All charsets must implement this class
+ *
+ * A saved processor does NOT serialize this class. It calls the newFactory method and serializes
+ * the resulting BitsCharsetFactory.
+ */
+abstract class BitsCharsetDefinition {
+
+  def name(): String
+
+  def newFactory(): BitsCharsetFactory
```
stevedlawrence, line 32: Is there a use case for the Factory indirection? Based on the comment it seems to be related to serialization, but I don't see anywhere where we store the created factories where they could be serialized.

The only use I see is `find(...).newFactory().newInstance()`. So we create a factory, create an instance from that factory, and throw away the factory. I wonder if we can simplify this and just treat `BitsCharsetDefinition` as if it were the factory--it can have the `newInstance` function that is directly called when a new instance of a charset is needed?

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.processors.charset
+
+/**
+ * Factory for a Daffodil charset.
+ *
+ * This is the serialized object which is saved as part of a processor.
+ * It constructs a charset at runtime when newInstance() is called.
```
stevedlawrence, line 24: Is this true anymore?Mentioned in another comment, but I don't see this factory being stored anywhere to be serialized? I wonder if things changed so we only serialized the name of a charset, and on deserialization we look up the charset by that string and create a new instance?

stevedlawrence (commented):
Nice change! Couple questions on deprecation ability and maybe we can remove the factory indirection?

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
```diff
@@ -33,11 +33,10 @@ object CharsetUtils {
    * encodings as well as the standard ones.
    */
   def getCharset(name: String): BitsCharset = {
-    val cs = DaffodilCharsetProvider.charsetForName(name)
-    cs
+    BitsCharsetDefinitionRegistry.find(name).newFactory.newInstance
```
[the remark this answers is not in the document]
tuxji: OK, let's keep the Registry suffix then.

daffodil-io/src/main/resources/META-INF/services/org.apache.daffodil.processors.charset.BitsCharsetDefinition
```diff
@@ -0,0 +1,54 @@
...
+org.apache.daffodil.processors.charset.BitsCharset3BitDFI746DUI002Definition
+org.apache.daffodil.processors.charset.BitsCharset3BitDFI747DUI001Definition
+org.apache.daffodil.processors.charset.BitsCharset4BitDFI746DUI002Definition
+org.apache.daffodil.processors.charset.BitsCharset5BitDFI1661DUI001Definition
+org.apache.daffodil.processors.charset.BitsCharset5BitDFI769DUI002Definition
+org.apache.daffodil.processors.charset.BitsCharset5BitPackedLSBFDefinition
+org.apache.daffodil.processors.charset.BitsCharset6BitDFI264DUI001Definition
+org.apache.daffodil.processors.charset.BitsCharset6BitDFI311DUI002Definition
+org.apache.daffodil.processors.charset.BitsCharset6BitICAOAircraftIDAliasDefinition
+org.apache.daffodil.processors.charset.BitsCharset6BitICAOAircraftIDDefinition
```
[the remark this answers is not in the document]
tuxji: I'd prefer deprecating charsets to be done as a separate issue/PR.  Whoever is in the best place to know which charsets should be deprecated (Mike or Steve?) should list these charsets in the issue, and then any committer can work on the PR.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDefinition.scala
tuxji: Well, the comments about serialization were copied from the LayerCompiler classes.  It probably is true that Daffodil processors store / serialize layer classes in such a way that this indirection makes sense, but it may not be true that Daffodil processors need this indirection for BitCharset classes at all.  @alexanderrevello, try removing the indirection and see if all the tests (including integration/CLI tests) still pass.  If so, then Steve is right and the indirection is unneeded.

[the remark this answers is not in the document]
alexanderrevello: parentheses for name are required for simplenamedserviceloader consistency

alexanderrevello: Indirection is not needed and has been removed

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
```diff
@@ -33,11 +33,13 @@ object CharsetUtils {
    * encodings as well as the standard ones.
    */
   def getCharset(name: String): BitsCharset = {
-    val cs = DaffodilCharsetProvider.charsetForName(name)
-    cs
+    val cs = BitsCharsetDefinitionRegistry.find(name).getOrElse(null)
+    if (cs == null)
+      SDE("The encoding '%s' was not found. Available choices are: %s", name, BitsCharsetDefinitionRegistry.supportedEncodingsString)
```
stevedlawrence, line 38: This changes the behavior of this function. In the past it returned null if the encoding wasn't found and the caller was expected to handle it correctly. I'd guess we want to keep that behavior, since some callers might just want to know if a charset exists or not, and do something other than error if not. 

Maybe something like this instead?
```scala
val cs = BitsCharsetDefinitionRegistry.find(name).map(_.newInstance).getOrElse(null)
cs
```

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
```diff
@@ -69,6 +71,10 @@ object CharsetUtils {
   }
 
   val unicodeReplacementChar = '\uFFFD'
+
+  private def SDE(id: String, args: Any*): Nothing = {
+    throw new Exception(id.format(args: _*))
+  }
```
stevedlawrence, line 77: We usually avoid throwing generic Exception's. It often makes it hard to know how to best handle the error. There's also no context associated with this SDE, like schema file or line number, which most SDE's have. I think that's because it just doesn't exist here. Which makes me think more that CharsetUtils should not be throwing any exceptions/SDE's but should instead rely on callers to handle that.

daffodil-runtime1/src/main/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
stevedlawrence, line 31: The comments mention that these charsets are defined in src/test/resources since they are only used for tests. But this services file is in `src/main/resources`. I think this service file also needs to be in `src/test/resources` so that these charsets are only loaded during tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvEncoding.scala
```diff
@@ -93,10 +93,11 @@ abstract class CharsetEvBase(encodingEv: EncodingEvBase, tci: DPathCompileInfo)
 
   override def compute(state: ParseOrUnparseState) = {
     val encString = encodingEv.evaluate(state)
-    val cs = CharsetUtils.getCharset(encString)
```
stevedlawrence, line 96: Should this still use `CharsetUtils`, and `CharsetUtils` should be the only thing that uses the registry? This way we can easily change how the registry works if we ever need to since it's hidden behind `CharsetUtils`

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDefinitionRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.daffodil.util.SimpleNamedServiceLoader
+
+/*
+ * Finds all pluggable BitCharsets and makes them available to Daffodil after they have been
+ * setup as described in BitsCharsetDefinition.scala
+ */
+object BitsCharsetDefinitionRegistry {
+
+  private lazy val bitsCharsetDefinitionMap: Map[String, BitsCharsetDefinition] =
+    SimpleNamedServiceLoader.loadClass[BitsCharsetDefinition](classOf[BitsCharsetDefinition])
```
stevedlawrence, line 28: All charsets have a name, but some also have aliases. For example, `US-ASCII` is the official name, but we also support `ASCII`. This SimpleNamedServiceLoader only knows about the name, which I think will break aliases?

Does our `SimpleNamedServiceLoader` need to change so that it can somehow support aliases?

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDefinitionRegistry.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Given name, finds the BitsCharsetDefinition or null if not found
+   */
+  def find(name: String): Option[BitsCharsetDefinition] = bitsCharsetDefinitionMap.get(name)
```
stevedlawrence, line 33: This doesn't do any case insensitive stuff. I thought we supported lower or upper case charset names? But maybe I'm wrong? Should we, or should all encodings match exactly what is defined by the `name` function returned by the definition?

stevedlawrence: I have confirmed that we do currently support case insensitive encodings. I actually don't see in the code where we handle that, but creating schemas with different cases of the same encoding does work as expected. I'm not sure if we need to do anything special with these new changes or if will just work.

stevedlawrence: Ah, I found how we handle case insensitivity:

https://github.com/apache/daffodil/blob/main/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/Cookers.scala#L82=

The encoding property is always capitalized by our encoding cooker. So in order for a charset to be findable, the name() function must return an all capitalized string. Ours all currently do that I think, but we should probably document that in the `BitsCharsetDefinition` class.

alexanderrevello: I will add that as a comment

alexanderrevello:
yeah I will get rid of those extra commits

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDefinition.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * These are the classes which must be dynamically loaded in order to add a charset implementation
+ * to Daffodil. All charsets must implement this class and be added to the 
+ * org.apache.daffodil.processors.charset.BitsCharsetDefinition file in 
+ * daffodil-io/src/main/resources/META-INF/services. Names are not case sensitive, so BitsCharsetDefinition
+ * that have the same name with different capitalization will return the most recent addition
+ */
+abstract class BitsCharsetDefinition {
+
+  def name(): String
```
tuxji, line 28: Actually, I believe what @stevedlawrence meant was that we should add a comment here saying that name() must return an all-upper-case string for the encoding to be findable by Daffodil.

stevedlawrence: Yep, that's correct, sorry for the confusion. Eventually users are going to start implementing their own custom BitsCharsetDefinition and looking at this class for documentation on what should be returned. Making it clear that name() must return all uppercase is important.

Another option might be modify `SimpleNamedServiceLoader` to have some kind of option to specifiy this behavior, so that it just always capitalizes whatever name returns. Then users don't have to really care. Might need changes to support aliases anyways (not sure how best to handle that).

mbeckerle (changes requested):
One more minor change to get rid of Exception throw method.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/CharsetUtils.scala
stevedlawrence: I think this SDE should be removed. This function should either return a charset or null. The callers will determine how to handle null.

daffodil-runtime1/src/main/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
stevedlawrence: I dont' think this comment was resolved. This file should be in `src/test/resources`, right?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvEncoding.scala
stevedlawrence: This comment wasn't addressed.

tuxji (commented):
I agree with Steve and Mike's suggestions.  Once you make the changes they suggest, we can merge your pull request (which looks like a really nice improvement).

daffodil-runtime1/src/main/resources/META-INF/services/org.apache.daffodil.charsets.CharsetCompiler
alexanderrevello: file was removed as it was no longer needed. Correct version is in the test folder

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/EvEncoding.scala
alexanderrevello: reverted

stevedlawrence (approved):
+1 :+1:  looks great!

Please squash into a single commit, update the commit message if needed (include the bug number on the last line) and this should be good to merged.

mbeckerle (approved):
+1

PR #770, opened 2022-03-10
Added dfdlx:alignmentViaSkip property

mbeckerle:
DAFFODIL-2652

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
@@ -27,6 +27,7 @@ import org.apache.daffodil.exceptions.Assert
 import org.apache.daffodil.dsom.QuasiElementDeclBase
 import org.apache.daffodil.dsom.Root
 import org.apache.daffodil.dsom.Term
+import org.apache.daffodil.schema.annotation.props.gen.AlignmentKind
 
 case class AlignmentMultipleOf(nBits: Long) {
   def *(that: AlignmentMultipleOf) = AlignmentMultipleOf(Math.gcd(nBits, that.nBits))
@@ -49,14 +50,21 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
   requiredEvaluationsIfActivated(hasNoSkipRegions)
 
   /**
+   * If "manual" this property disables all automatic alignment. The
+   * schema author must use leadingSkip, trailingSkip, or just ensure
+   * all the elements/terms are aligned based on their length.
+   */
+  lazy val alignmentKindDefaulted: AlignmentKind =
+    optionAlignmentKind.getOrElse(AlignmentKind.Automatic)
+
+  /**
    * true if we can statically determine that the start of this
    * will be properly aligned by where the prior thing left us positioned.
    * Hence we are guaranteed to be properly aligned.
    */
   final lazy val isKnownToBeAligned: Boolean = LV('isKnownToBeAligned) {
-    if (!isRepresented) {
-      true
-    } else {
+    if (!isRepresented || (alignmentKindDefaulted == AlignmentKind.Manual)) true
+    else {
       val pa = priorAlignmentWithLeadingSkipApprox
       val aa = alignmentApprox
       val res = (pa % aa) == 0
@@ -73,7 +81,8 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
    * considers the surrounding context meeting the alignment needs.
    */
   final lazy val isKnownToBeTextAligned: Boolean = LV('isKnownToBeTextAligned) {
-    if (isKnownEncoding) {
+    if (alignmentKindDefaulted == AlignmentKind.Manual) true // manual alignment
+    else if (isKnownEncoding) {
       if (knownEncodingAlignmentInBits == 1)
         true
       else if (priorAlignmentWithLeadingSkipApprox.nBits % knownEncodingAlignmentInBits == 0)
@@ -87,7 +96,8 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
   }.value
 
   final lazy val isDelimiterKnownToBeTextAligned: Boolean = {
-    if (isKnownEncoding) {
+    if (alignmentKindDefaulted == AlignmentKind.Manual) true // manual alignment
+    else if (isKnownEncoding) {
       if (knownEncodingAlignmentInBits == 1)
         true
       else if (endingAlignmentApprox.nBits % knownEncodingAlignmentInBits == 0)
@@ -106,29 +116,22 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
     AlignmentMultipleOf(alignmentValueInBits.toLong)
   }
 
-  lazy val leadingSkipInBits = {
-    alignmentUnits match {
-      case AlignmentUnits.Bits => leadingSkip
-      case AlignmentUnits.Bytes => leadingSkip * 8
-    }
+
+  private def alignmentSkipInBits(skipProp: Int) = alignmentUnits match {
+    case AlignmentUnits.Bits => skipProp
+    case AlignmentUnits.Bytes => skipProp * 8
   }
+
+  lazy val leadingSkipInBits = alignmentSkipInBits(leadingSkip)
+  lazy val trailingSkipInBits = alignmentSkipInBits(trailingSkip)
 
   private lazy val leadingSkipApprox: LengthApprox = {
     LengthExact(leadingSkipInBits)
   }
 
-  lazy val trailingSkipInBits = {
-    alignmentUnits match {
-      case AlignmentUnits.Bits => trailingSkip
-      case AlignmentUnits.Bytes => trailingSkip * 8
-    }
-  }
-
   protected lazy val trailingSkipApprox: LengthApprox = {
     LengthExact(trailingSkipInBits)
   }
-
-  private lazy val unaligned = AlignmentMultipleOf(1)
 
   // FIXME: DAFFODIL-2295
   // Does not take into account that in a sequence, what may be prior may be a separator.
@@ -150,11 +153,11 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
             // come from a previous element of this same array. Since this
             // array element is implicit length, knowing where it ends requires
             // knowing where it starts and the approximate length of the
-            // children. But we can't know whree it starts without knowing
+            // children. But we can't know where it starts without knowing
             // where the previous one array element ends. And we end up in a
             // loop.
             //
-            // So there isn't much we can do regarding alignement. What we can
+            // So there isn't much we can do regarding alignment. What we can
             // do is determine if this array element is byte aligned AND all of
             // its children are byte lengths/byte aligned, if that is the case
             // then we at least know this array and its elements are byte
@@ -197,7 +200,7 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
           ps.endingAlignmentApprox
         }
         eaa
-      }.toSeq
+      }
 
       val parentAlignmentApprox =
         if (priorSibs.isEmpty || isEverInUnorderedSequence) {
@@ -225,7 +228,7 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
   }
 
   protected lazy val contentStartAlignment: AlignmentMultipleOf = {
-    if ((priorAlignmentWithLeadingSkipApprox) % alignmentApprox == 0) {
+    if (priorAlignmentWithLeadingSkipApprox % alignmentApprox == 0) {
       // alignment won't be needed, continue using prior alignment as start alignment
       priorAlignmentWithLeadingSkipApprox
     } else {
@@ -277,7 +280,7 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
             // can't be last, no possibilities to gather.
             None
         val lastApproxes = lastApproxesConsideringChildren ++ optApproxIfNoChildren
-        Assert.invariant(!lastApproxes.isEmpty)
+        Assert.invariant(lastApproxes.nonEmpty)
         val res = lastApproxes.reduce { _ * _ }
         res
       }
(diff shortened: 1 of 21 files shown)


daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharset.scala
```diff
@@ -55,7 +55,7 @@ trait BitsCharset extends Serializable {
   def aliases: Seq[String] = Nil
   def bitWidthOfACodeUnit: Int // in units of bits
   def requiredBitOrder: BitOrder
-  def mandatoryBitAlignment: Int
+  def mandatoryBitAlignment: Int // ignored when dfdlx:alignmentViaSkip is "yes"
```
mbeckerle, line 58: I think we need a test to see what happens if we're using regular 8-bit ascii charset encoding, but it isn't 8-bit aligned because we forced it to not do alignment via dfdlx:alignmentViaSkip. Similarly for utf-16 that isn't byte aligned.

Getting a runtime error here, so long as the diagnostic indicates the character is not aligned properly, is the right behavior we want.

tuxji: I agree, let's add a test to determine whether the diagnostic looks appropriate.

tuxji (approved):
+1

Looks good.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
```diff
@@ -47,6 +48,27 @@
     </xs:restriction>
   </xs:simpleType>
 
+
+  <xs:attribute name="alignmentViaSkip" type="dfdl:YesNoEnum" default="no">
```
stevedlawrence, line 52: Thoughts on changing this to something like "alignmentPolicy"? I've mentioned other tweaks to deal with the alignment optimizations, but that are more complicated to implement. It might be nice to have a single property that essentially lets us choose between different alginment optimizations. And they could vary from a very pessimistic algorithm like we have now, or this optimistic one that completely disables alignment like this, or somewhere in between? We'd only have two values for now, so funtionally the same as a yes/no choice, but we could expand it in the future.

stevedlawrence (approved):
+1, only thought about the name of the property, but I don't feel strongly about it. It's experimental so we can always change the name later if needed

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -614,40 +614,44 @@ final class InputSourceDataInputStream private(val inputSource: InputSource)
...
-        }
+    withLocalCharBuffer { lcb =>
+      val cb = lcb.getBuf(nChars)
+      val numDecoded = finfo.decoder.decode(this, finfo, cb)
+      if (numDecoded == nChars) {
+        Maybe(cb.flip.toString)
+      } else {
+        setBitPos0b(startingBitPos)
+        Maybe.Nope
       }
```
tuxji, line 634: Code coverage says those lines are never called.  I looked for callers of `getString(nChars, finfo)` and the only place I found seems to be one line in [TextParser.scala](https://github.com/apache/daffodil/blob/55f638de74edfa400ce6cba1de19162069c9298a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextParser.scala#L57) which completely ignores the return value and doesn't check if Maybe.Nope is returned.  The new tests added below only call `getSomeString(nChars, finfo)`, not `getString(nChars, finfo)`.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDecoder.scala
```diff
@@ -32,6 +31,13 @@ import org.apache.daffodil.util.MaybeChar
 class BitsCharsetDecoderMalformedException(val malformedBits: Int)
   extends ThinException
 
+class BitsCharsetDecoderUnalignedCharDecodeException(val bitAlignment1b: Long)
+  extends ThinException {
+  override def getMessage(): String = {
+    s"Charset not byte aligned. Bit alignment (1b): ${bitAlignment1b}."
```
tuxji, line 37: Is reporting the desired `bitAlignment1b` more useful for debugging than reporting the current value of the stream's `bitPos1b`?  Maybe the caller on line 141 should pass `dis.bitPos1b` instead of `dis.bitPos1b % 8` and this exception should capture the value as `bitPos1b` instead of `bitAlignment1b`?

daffodil-test/src/test/scala/org/apache/daffodil/section17/calc_value_properties/TestComputedLengthFields.scala
```diff
@@ -39,13 +39,24 @@ class TestComputedLengthFields {
   @Test def test_computedLengthAroundPrefixedLengths1p(): Unit = { runner.runOneTest("computedLengthAroundPrefixedLengths1p") }
 
   // DAFFODIL-2626 - deadlock interaction between computed length and prefixed-length strings.
-  // @Test def test_computedLengthAroundPrefixedLengths1u(): Unit = { runner.runOneTest("computedLengthAroundPrefixedLengths1u") }
+  // @Test
+  def test_computedLengthAroundPrefixedLengths1u(): Unit = { runner.runOneTest("computedLengthAroundPrefixedLengths1u") }
+
+  // This test shows you can work around DAFFODIL-2626 using the dfdlx:alignmentKine='manual' property.
```
tuxji, line 45: typo: `dfdlx:alignmentKind`

tuxji (approved):
+1 once comments are addressed

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
```diff
@@ -48,13 +50,27 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
...
+   * If "manual" this property disables all automatic alignment. The
+   * schema author must use leadingSkip, trailingSkip, or just ensure
+   * all the elements/terms are alignmed based on their length.
+   */
+  lazy val alignmentKind: AlignmentKind = {
+    val prop = findPropertyOption("alignmentKind")
+    prop match {
+      case NotFound(_, _, _) => AlignmentKind.Automatic // default
+      case _ => AlignmentKind.Manual
+    }
```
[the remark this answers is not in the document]
mbeckerle: All that sort of defensive programming and testing for properties is handled centrally by XML validation of the DFDL schemas. Those validators for short-form props are generated off of the XML schemas for properties and extension properties. 

This is a big advantage of XML over other things without robust schema languages like JSON.

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDecoder.scala
mbeckerle: Arguably we should save the bitPos1b and have the message refer to both the bitAlignment1b, the bitPos1b and the bytePos1b since it depends on the data what is useful to the user. 
I will improve this.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
mbeckerle: I'm going to consolidate, and just have getSomeString().

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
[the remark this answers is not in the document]
stevedlawrence: I think @tuxji is right, I think there is still a logic bug. If the property is defined this ignores the value and sets the kind to manual. So even if you do `dfdlx:alignmentKind="automatic"` you still get manual behavior? We should add a test that `dfdlx:alignmentKind="automatic"` works.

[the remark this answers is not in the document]
mbeckerle: Yup. Upon re-reading, haste makes waste. I will fix this and add a test.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
```diff
@@ -48,13 +51,29 @@ trait AlignedMixin extends GrammarMixin { self: Term =>
...
+   * all the elements/terms are alignmed based on their length.
+   */
+  lazy val alignmentKind: AlignmentKind = {
+    val prop = findPropertyOption("alignmentKind")
+    prop match {
+      case NotFound(_, _, _) => AlignmentKind.Automatic // default since this is a new property
+      case Found("automatic",_,_,_) => AlignmentKind.Automatic
+      case Found("manual",_,_,_) => AlignmentKind.Manual
+      case _ => Assert.impossible("Per validation, can only be 'automatic' or 'manual'")
+    }
```
tuxji, line 66: Much more readable now!  Even covers the impossible case so readers will understand what's happening.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -614,40 +614,23 @@ final class InputSourceDataInputStream private(val inputSource: InputSource)
...
-        } else {
-          setBitPos0b(startingBitPos)
-          Maybe.Nope
-        }
-      }
-    }
-  }
-
+  /**
+   * Almost the same as getString. But returns some characters (up to nChars) if they are available.
```
tuxji, line 618: I believe you've removed getString so this comment should just say "Returns some...", yes?

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -614,40 +614,23 @@ final class InputSourceDataInputStream private(val inputSource: InputSource)
...
-        if (numDecoded > 0) {
-          Maybe(cb.flip.toString)
-        } else {
-          setBitPos0b(startingBitPos)
-          Maybe.Nope
-        }
+    withLocalCharBuffer { lcb =>
+      val cb = lcb.getBuf(nChars)
+      val numDecoded = finfo.decoder.decode(this, finfo, cb)
+      if (numDecoded > 0) { // This is the difference with the above getString
```
tuxji, line 629: This comment should be removed too, yes?

daffodil-io/src/main/scala/org/apache/daffodil/processors/charset/BitsCharsetDecoder.scala
```diff
@@ -32,6 +31,16 @@ import org.apache.daffodil.util.MaybeChar
 class BitsCharsetDecoderMalformedException(val malformedBits: Int)
   extends ThinException
 
+class BitsCharsetDecoderUnalignedCharDecodeException(val bitPos1b: Long)
+  extends ThinException {
+  def bitAlignment1b = bitPos1b % 8
+  def bytePos1b = ((bitPos1b - 1) / 8) + 1
+
+  override def getMessage(): String = {
+    s"Charset not byte aligned. bitAlignment1b=${bitAlignment1b}, bitPos1b=${bitPos1b}, bytePos1b=${bytePos1b}."
```
tuxji, line 40: I like the change!

tuxji (approved):
+1 with comments about comments :)

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextParser.scala
```diff
@@ -54,7 +54,7 @@ class TextParser(override val context: TermRuntimeData)
         if (isDelimRequired) Nope
         else {
           val totalNumCharsRead = 0
-          input.getString(totalNumCharsRead, state)
+          input.getSomeString(totalNumCharsRead, state)
```
stevedlawrence, line 57: This behavior is slightly different, this now doesn't care if it doesn't get totalNumCharsRead--before it would have been an error. Maybe that wasn't possible in this code path? It's hard to tell.

Also a bit concerning that this change has no test coverage. Maybe this code isn't even possible to reach?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ParseErrors.scala
```diff
@@ -39,6 +40,14 @@ class ParseError(rd: Maybe[SchemaFileLocation], val loc: Maybe[DataLocation], ca
   override def toParseError = this
 }
 
+final class CharsetNotByteAlignedError(pstate: PState,
+  cause: BitsCharsetDecoderUnalignedCharDecodeException)
+  extends ParseError(
```
stevedlawrence, line 45: Should this be a schema definition error? The only way this *should* be possible if is someone sets alignmentKind to manual but their schema isn't written correctly for manual alignment. So it's really a sign of a bad schema and maybe we don't want the chance to recover via backtracking?

stevedlawrence (approved):
+1 to changes, minor comments

daffodil-core/src/main/scala/org/apache/daffodil/grammar/AlignedMixin.scala
mbeckerle: You know, that may be the case. I may have over complicated this. This is a regular property, except extension namespace, and default value.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/dfa/TextParser.scala
mbeckerle: I do think it is not reachable any more. I will look at it again, and if it's really dead code I'll remove.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ParseErrors.scala
mbeckerle: This is a tricky point. I went with Parse error because the notion is that a schema could be speculating down a branch where the alignment doesn't work out, and this error allows it to backtrack to the proper branch where it does. This is very analogous to speculating down a branch where decoding characters causes failures or delimiters aren't found properly.

PR #771, opened 2022-03-10
Fix saveParser CLI with config having tunables.

mbeckerle:
The CLI's loading of config files was not
trimming the node, hence, getting PCDATA nodes
instead of the strings it expected.

Shares code from DaffodilTunables object now.

DAFFODIL-2650


daffodil-cli/src/test/resources/configExample.cfg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
mbeckerle, line 1: Missing Apache banner. Fails rat check.

daffodil-core/src/test/resources/test/configExample.cfg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
mbeckerle, line 1: fails rat check.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -76,21 +76,25 @@ class TunableGenerator(schemaRootConfig: scala.xml.Node, schemaRootExt: scala.xm
...
-    |        tunablesMap
+    |        tunablesMap(node)
     |      } else {
     |        Map.empty
     |      }
     |
     |    new DaffodilTunables().setTunables(configTunables)
     |  }
+    |
+    |  def tunablesMap(node: scala.xml.Node) = {
```
stevedlawrence, line 87: What's our standard for including the type of the return value in a function definition? I personally prefer it since I can immediately know the return value, but maybe that' s not very Scala-y?

daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala
```diff
@@ -97,4 +102,15 @@ class TestTunables {
...
+    val resource: URI = Misc.getRequiredResource("test/configExample.cfg")
+    val loader = new DaffodilXMLLoader()
+    val node = loader.load(URISchemaSource(resource), Some(XMLUtils.dafextURI))
+    val tunablesMap = DaffodilTunables.tunablesMap(node)
+    val tunables = DaffodilTunables(tunablesMap)
+    val warnIDStrings = tunables.suppressSchemaDefinitionWarnings.map{ _.toString }
+    assertEquals(2, warnIDStrings.length)
+    assertTrue(warnIDStrings.contains("facetExplicitLengthOutOfRange"))
+    assertTrue(warnIDStrings.contains("encodingErrorPolicyError"))
+  }
```
tuxji, line 115: I am not certain that `testConfigFileLoadingTunables` tests anything different than what `testTunables1` tests in daffodil-cli/src/test/scala/org/apache/daffodil/TestTunables.scala.  The purpose of `testTunables1` is to confirm that `Main.retrieveTunables` now does the same thing as `DaffodilTunables.tunablesMap(node)`, even though that's because `Main.retrieveTunables` was changed to call `DaffodilTunables.tunablesMap(node)` itself.  Why do we need `testConfigFileLoadingTunables` when it tests a subset of the same code as `testTunables1`?  Actually, maybe we simply need to move `Main.retrieveTunables` to `DaffodilTunables.retrieveTunables` and make `testConfigFileLoadingTunables` call it; then we can get rid of `testTunables1`.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -609,17 +609,8 @@ object Main {
...
-      case Some(configNode) => {
-        val tunablesOpt = (configNode \ "tunables").headOption
-        tunablesOpt match {
-          case None => Map.empty
-          case Some(tunableNode) => {
-            tunableNode.child.map { n => (n.label, n.text) }.toMap
-          }
-        }
-      }
+      case Some(configNode) => DaffodilTunables.tunablesMap(configNode)
```
stevedlawrence, line 612: I think the TDMLRunner also does some trim stuff with tunables to solve this same problem. Probably want to modify that to use this new code so there's one place that deal with parsing tunables from Scala Noes.

Also, I wonder if it would be helpful to add another overloaded `tunablesMap` function that accepts a `File` as a parameter. Seems like that might reduce duplication that both Main and TDMLRunner need to support, so that neither needs to load the config file into XML first?

I don't know if this is necessary, but I wonder if eventually we want a `DaffodilConfig` class that handles parsing and extracting inforation from config files? Again with the goal of reducing duplication between Main and the TDMLRunner. That way the variables and tunables (and maybe future config stuff) that is stored in a config file can all be handled consistently.

daffodil-cli/src/test/scala/org/apache/daffodil/TestTunables.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+import java.net.URI
+
+class TestTunables {
+
+  // verifies that DAFFODIL-2650 is fixed.
+  @Test def testTunables1(): Unit = {
+    val resource: URI = Misc.getRequiredResource("test/configExample.cfg")
```
stevedlawrence, line 35: I think this needs to be just "configExample.cfg"?

daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala
```diff
@@ -97,4 +102,15 @@ class TestTunables {
     assertEquals(true, w4.contains("escapeSchemeRefUndefined"))
   }
 
+  @Test def testConfigFileLoadingTunables(): Unit = {
+    val resource: URI = Misc.getRequiredResource("test/configExample.cfg")
```
stevedlawrence, line 106: Same here, does this need to just be "configExample.cfg"?

daffodil-cli/src/test/scala/org/apache/daffodil/TestTunables.scala
```diff
@@ -0,0 +1,45 @@
...
+    val loader = new DaffodilXMLLoader()
+    val node = loader.load(URISchemaSource(resource), Some(XMLUtils.dafextURI))
+    val tunablesMap = Main.retrieveTunables(Map.empty, Some(node))
+    val tunables = DaffodilTunables(tunablesMap)
+    val warnIDStrings = tunables.suppressSchemaDefinitionWarnings.map{ _.toString }
+    assertEquals(2, warnIDStrings.length)
+    assertTrue(warnIDStrings.contains("facetExplicitLengthOutOfRange"))
+    assertTrue(warnIDStrings.contains("encodingErrorPolicyError"))
+  }
+}
```
stevedlawrence, line 45: Any particular reason not to create a CLI test using the normal integration test stuff? This is the only test in daffodil-cli that isn't one of those. Suggest we be consistent and make this an Integration test.

stevedlawrence (approved):
+1 with test fixes, some potential suggestions for additional refactorings

tuxji (approved):
+1 once tests pass correctly

I suggest the easier fix would be to move `Main.retrieveTunables` to `DaffodilTunables.retrieveTunables` and get rid of `testTunables1`.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
mbeckerle: It's public method, so I think should have the return type.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: I think I've addressed this now.

daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala
mbeckerle: refactored to address this and SL's modularity concerns.

tuxji (approved):
+1, nice consolidation

PR #772, opened 2022-03-10
Clear diagnostics (warnings) on saveParser.

mbeckerle:
I'd like suggestions on how to test this. 

DAFFODIL-2267


daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -128,11 +128,12 @@ final class TDMLDFDLProcessorFactory private (
         if (useSerializedProcessor) {
           val os = new java.io.ByteArrayOutputStream()
           val output = Channels.newChannel(os)
-          p.save(output)
+          p.save(output, clearDiagnostics = false) // TDML runner wants to preserve all diagnostics
```
stevedlawrence, line 131: As an alternate approach, what are your thoughts on making it so we never clear diagnostics from a DataProcessor, whether it be serialized or not. The benefit is that when you reload a DataProcessor, you still have access to those compile time diagnostics if you want them. So if you want to know if there were any warnings when you compiled some old DataProcessor, you can easily figure that out.

And then to avoid extraneous diagnostics during parse/unparse we change it so DataProcessor diagnostics do not carry over to the ParseResult diagnostics. So there are two completely separate buckets of diagnostics--compile time and runtime. If a user needs access to the compile time diagnostics (like the TDML runner) , it needs to pull them off of the DataProcessor.

In fact, maybe we already do this and the compile-time and run-time diagnosics are already separate? Maybe the fix is to just make it so when we `reload()` we don't output diagnostics from the DataProcessor? In other words, we remove this line:

https://github.com/apache/daffodil/blob/main/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala#L595

That only affects the CLI. I'm not sure if that is where you were seeing too much diagnostic information.

tuxji (commented):
Just waiting to see how Steve's comments are addressed while adding a few comments of my own.

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala
stevedlawrence: > TDML Runner won't do anything with compilation time diagnostics in its usual path anyway. I noticed that behavior when implementing runtime2's support for returning a processor to TDML Runner.

If at test has <tdml:errors> or <tdml:warnings> then I think TDML runner should look in both compile time and runtime diagnostics, I think...

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/SchemaSetRuntimeData.scala
```diff
@@ -23,20 +23,20 @@ import org.apache.daffodil.processors.unparsers.Unparser
...
-  val diagnostics: Seq[Diagnostic],
-  val elementRuntimeData: ElementRuntimeData,
+  diagnostics: Seq[Diagnostic],
+  elementRuntimeData: ElementRuntimeData,
   /*
    * The original variables determined by the schema compiler.
    */
-  variables: VariableMap,
-  val typeCalculators: TypeCalcMap)
+  variables_ : VariableMap,
```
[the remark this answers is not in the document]
mbeckerle: I made this a case class to get the automatic copy method. But that also makes all the args into public values. So the _ suffix is to say "not really for public use". 

I think this is moot however, because I'm going to try the approach SL suggested.

mbeckerle:
How do I do a CLI test that insists that a particular string is NOT present. 

Can the expect utility just grab all output without hanging waiting for anything, and then I can 
search it?

stevedlawrence:
The `expect()` function resturns a [Result](http://javadox.com/net.sf.expectit/expectit-core/0.9.0/net/sf/expectit/Result.html), which has a `getBefore` function to to get the data before the match. Though it just says "part of the data". I'm not sure how back far it goes. Maybe it's some small buffer, maybe it's since the previous expect call? Not sure. If you can `expect` a string that is guaranteed to appear after a warning (if a warning were printed), then that /might/ work.

tuxji (approved):
+1

Looks good!

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -80,4 +82,52 @@ class TestCLITunables2 {
...
+        // in the expected output
+        //
+        shell.expect(contains("""Schema Definition Warning"""))
+        Util.expectExitCode(ExitCode.Success, shell)
+        shell.sendLine("exit")
+        shell.expect(eof)
+      } finally {
+        shell.close()
+      }
+    } while (false) // workaround scala local block bug.
```
stevedlawrence, line 110: What is the local block bug?

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -80,4 +82,52 @@ class TestCLITunables2 {
...
+    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+    val compiledProcFile = Util.newTempFile("savedProc", ".bin")
+    compiledProcFile.deleteOnExit()
+    val compiledProcFilePath = compiledProcFile.getAbsolutePath()
+    //
+    do {
+      val shell = Util.start("")
+      try {
+        // note: 2>&1 is shell-speak for "connect stderr into stdout"
+        val cmd = String.format("""%s save-parser -s %s %s 2>&1""", Util.binPath, testSchemaFile, compiledProcFilePath)
```
stevedlawrence, line 97: Expect keeps track of the stderr and stdout separately, so you don't have to do this redirect. The `expect()` function accepts a an index for which stream to expect from. It defaults to 0 (i.e. stdout) if not specified, but you can speicfy 1 for stderr. So if you remove the redirect, you can change the below expect to be:

```scala
shell.expect(1, contains("""Schema Definition Warning"""))
```

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -80,4 +82,52 @@ class TestCLITunables2 {
...
+      }
+    } while (false) // workaround scala local block bug.
+    do {
+      val shell = Util.start("")
+      try {
+        // note: 2>&1 is shell-speak for "connect stderr into stdout"
+        val cmd = String.format("""echo a,b| %s parse -P %s 2>&1""", Util.binPath, compiledProcFilePath)
+        shell.sendLine(cmd)
+        shell.sendLine("exit")
+        val output = shell.expect(eof).getBefore
```
stevedlawrence, line 118: This is nice! A good technique to know to just get all output to a string. It's kindof the only good way to check that output doesn't exist.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -592,7 +592,9 @@ object Main {
     try {
       val compiler = Compiler()
       val processor = Timer.getResult("reloading", compiler.reload(savedParser))
-      displayDiagnostics(processor)
+      // note that we do not display the diagnostics that were saved as part of the
+      // saved processor. Those are from compile time, are all warnings, and
+      // are just noise in a production system where we're reloading a parser.
```
stevedlawrence, line 597: Just to confirm, we do actually keep compile time and runtime diagnostics completely separate and compile time warnings do not bleed into the ParseResult diagnostics? And I guess the TDML runner combines them both when checking for expected warnings when a test finishes? That's not how I thought it worked, but good to know I was wrong because that seems like a logical way for it to work.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -592,7 +592,9 @@ object Main {
     try {
       val compiler = Compiler()
       val processor = Timer.getResult("reloading", compiler.reload(savedParser))
-      displayDiagnostics(processor)
```
stevedlawrence, line 595: I'm was thinking it might be useful output diagnostics if the -v option is added. But on second thought, probably not--you shouldn't need to enable parse/unparse verbosity just to get information about a saved parser. Probably the right thing is something like a "parser-info" command that can print out metadata stored inside a saved parser. That could include things like daffodil version (which we already have), java version, date/time built, diagnostics, schema information. But that feels like very low priority, especially since most of this information isn't even available at the moment.

stevedlawrence (approved):
+1, minor comments/questions

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: There's one heap of diagnostics held in static state on the SchemaSetRuntimeData object. This is compilation-time warnings only.
Runtime diagnostics are gathered in the PState/UState. They have to be because of backtracking (well for parsing anyway) we can't be mutating the diagnostics in the SSRD at runtime since it is shared cross threads.

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
mbeckerle: You are supposed to be able to just do this:
```
{
    val foo = bar
    ...
}```
To isolate the scope of foo to just the lines to end of block. 
But the scala compiler won't let you! Hence a `do { val foo = bar ; .... } while(false)` hack achieves the same thing.

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
mbeckerle: I tried this.

`shell.expect(1,....)`

Very problematic. There are two overloads which take an integer. The first is deprecated and takes a long. That one the long is a timeout. The other takes an int. That one the int is an output stream number. 

I even tried

`shell.expect(1.toInt,...)`

but they still indicate I'm calling the deprecated one where the first arg is a timeout. 

Until they actually remove the deprecated one, 
I think I will stick with 2>&1.

stevedlawrence: Ah, I had the wrong function name. It's `expectIn`, we use it in a handful of places so it should work.

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
mbeckerle: Interesting. I have modified this source, and it compiles fine now. May have been an IDE-related issue. 
I will put up a separate quick PR to remove this noise.

PR #773, opened 2022-03-14
Cleanup solarcloud warning. Remove do-while(false) local block workar…

mbeckerle:
…ound

The workaround is unnecessary. An IDE issue of some sort.

DAFFODIL-2674

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
@@ -27,6 +27,9 @@ import org.junit.Assert.assertTrue
 
 class TestCLITunables2 {
 
+  val stdout = 0
+  val stderr = 1
+
   /**
    * Suppresses SDW messages.
    */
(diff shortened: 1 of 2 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
```diff
@@ -107,8 +106,8 @@ class TestCLITunables2 {
       } finally {
         shell.close()
       }
-    } while (false) // workaround scala local block bug.
-    do {
```
stevedlawrence, line 111: Since you're making this change, can you also update to use `expectIn` instead of `expect` so we can remove the redirect? My concern is redirecting the input is going to cause stderr and stdout mixed up and might cause random failures when expecting strings depending on how things get mixed up. Might not actually be possible, but it's maybe good practice to avoid the redirect in favor of expectIn.

stevedlawrence (approved):
+1

daffodil-cli/src/it/scala/org/apache/daffodil/tunables/TestCLITunables2.scala
mbeckerle: Ok. 
Interestingly, fixing this do-while thing..... failed to compile on the github automated CI. Compiles on my machine tho.

stevedlawrence: I was playing around with a small example, and for some reason I needed to provide the type on some variables. Might be the case here as well.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -694,15 +694,6 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
...
 
-  // Provide ability to override existing (default) tunables
-  private def retrieveTunablesCombined(existingTunables: Map[String, String], cfg: DefinedConfig) = {
-    // Note, ++ on Maps replaces any key/value pair from the left with that on the
-    // right, so key/value pairs defined in tunables overrule those defiend in
-    // the config file
-    val combined = existingTunables ++ cfg.tunablesMap
-    combined
-  }
-
```
tuxji, line 705: Your PR & commit message doesn't mention this change - is it another cleanup (removing a function that's not used anymore)?

tuxji (approved):
+1

PR #774, opened 2022-03-14
Fix invariant broken masking of unparse error.

mbeckerle:
Add test files focused on enums using repType/repValues
in the simplest possible way.

DAFFODIL-2673

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
@@ -16,12 +16,12 @@
  */
 package org.apache.daffodil.processors.unparsers
 
-import org.apache.daffodil.processors.Success
 import org.apache.daffodil.processors.SequenceRuntimeData
 import org.apache.daffodil.processors.ElementRuntimeData
 import org.apache.daffodil.processors.TermRuntimeData
 import org.apache.daffodil.exceptions.Assert
-import org.apache.daffodil.processors.parsers.{EndArrayChecksMixin, MinMaxRepeatsMixin}
+import org.apache.daffodil.processors.parsers.EndArrayChecksMixin
+import org.apache.daffodil.processors.parsers.MinMaxRepeatsMixin
 import org.apache.daffodil.schema.annotation.props.gen.OccursCountKind
 
 /**
@@ -107,7 +107,7 @@ abstract class RepeatingChildUnparser(
 
     val actualOccurs = state.arrayPos
 
-    Assert.invariant(state.processorStatus eq Success)
+    // State could be Success or Failure here.
 
     endArray(state, actualOccurs)
   }
--- a/daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
@@ -1,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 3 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
```diff
@@ -107,7 +107,7 @@ abstract class RepeatingChildUnparser(
 
     val actualOccurs = state.arrayPos
 
-    Assert.invariant(state.processorStatus eq Success)
+    // State could be Success or Failure here.
```
stevedlawrence, line 110: This function never sets the state to failure, right? Which must mean that some other unparser is setting proessorStatus to Failure? But shouldn't all unparsers immediateyl throw a UE if there is a failure?

I wonder if this means that that some unparser should actually be throwing a UE instead of assumming something else will eventually figure it out and throwing it?

In fact, I found this, which looks suspicious:

https://github.com/apache/daffodil/blob/main/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala#L96-L99

That looks like an unparse function that is setting a ParseError as a failure. Seems that that should just be throwing a UE, and is maybe the underlying issue here?

mbeckerle: I will investigate this. There may be many places where setFailure is called in the unparser just because people assume a symmetry which isn't really there. The parser is supposed to unwind in orderly fashion because failures drive backtracking. For the unparser, setFailure is supposed to just throw.

mbeckerle: So changing this to just throw when UState.setFailed is called breaks many tests which just end with a throw rather than accumulating a failure that the TDML runner can then examine.

So it does appear that the unparser has the same "orderly unwind" assumption.

stevedlawrence: Hmm, might be a bug that we can fix some other day then? There's only 5 unparsers that check processorStatus which makes me think that we at the very least aren't consistent about how we unwind. I would expect more unparsers would need to check that and aren't unwinding properly. We also probably need some documentation about when an unparser should use setFailed or throw a UE, since apparently we do both right now. And, those unparse functions that create ParserErrors that I pointed out definitely feels wrong.

I'm fine with this change for 3.3.0, but it feels like there's something else going on to me.

stevedlawrence (approved):
+1

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/SequenceChildUnparsers.scala
mbeckerle: The throws you found are evaluated when expressions involving the type-calc functions (which we're phasing out eventually) are called. Expressions can throw (e..g, divide by zero) and those seemingly get caught and converted into failure diagnostics which are handled in orderly fashion (as would be necessary when parsing. Expressions for the most part work the same way parsing/unparsing... well, except for the blocking on forward ref stuff.)

The problem this bug fix is about is where the Enum lookups from String -> Int for unparsing are used. There was a mistaken assumption that when unparsing a sequence, child subunparsers would not be failing. Not sure why that assumption would be made, but it likely has to do with the notion that unparsers should "just throw" when they fail, which we know (by experiment) does not work as it breaks the TDML error processing.

stevedlawrence: > The throws you found are evaluated when expressions involving the type-calc function

Are you sure? The `TypeValueCalcUnparser` calls `outputTypeCalcUnparse` here:

https://github.com/apache/daffodil/blob/main/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/ExpressionEvaluatingUnparsers.scala#L150

And that function creates a ParseError. That's has to be incorrect.

I think you're right that `outputTypeCalcRun` is called only from DPath code, but in that case I would expect that it should create an error that isn't parse specific but is some generic DPath exception.

tuxji (approved):
+1

PR #781, opened 2022-03-31
Daffodil 2643 tdml logs

alexanderrevello:
TDML unit tests can now capture and look for logs
Like they do for warnings and errors, logs completed through log4j are
captured and can be required in the tdml test cases. The root logger
level is by default warning and when a tdml test is run, that level is
changed to info in order to capture logs. Tests were added to
demonstrate the feature as well as how it works in conjunction with
warnings and errors. Library dependencies were updated to give tdml-lib
access to log4jcore libraries needed to make an abstractappender

DAFFODIL-2643

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,6 +104,7 @@ lazy val sapi             = Project("daffodil-sapi", file("daffodil-sapi")).conf
 lazy val tdmlLib             = Project("daffodil-tdml-lib", file("daffodil-tdml-lib")).configs(IntegrationTest)
                               .dependsOn(macroLib % "compile-internal", lib, io, io % "test->test")
                               .settings(commonSettings)
+                              .settings(libraryDependencies ++= Dependencies.core)
 
 lazy val tdmlProc         = Project("daffodil-tdml-processor", file("daffodil-tdml-processor")).configs(IntegrationTest)
                               .dependsOn(tdmlLib, runtime2, core)
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
@@ -91,6 +91,8 @@ object Runner {
    */
   def defaultShouldDoWarningComparisonOnCrossTests = false
 
+  def defaultShouldDoLogComparisonCrossTests = false
+
 }
 
 /**
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
@@ -181,7 +181,8 @@ class Runner private (
         defaultValidationDefault,
         defaultImplementationsDefault,
         Runner.defaultShouldDoErrorComparisonOnCrossTests,
-        Runner.defaultShouldDoWarningComparisonOnCrossTests)
+        Runner.defaultShouldDoWarningComparisonOnCrossTests,
+        Runner.defaultShouldDoLogComparisonCrossTests)
     }
     ts
   }
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.tdml
+
+import org.apache.logging.log4j.core.appender.AbstractAppender
+import org.apache.logging.log4j.core.filter.ThresholdFilter
+import org.apache.logging.log4j.core.LogEvent
+import org.apache.logging.log4j.core.Filter.Result
+import org.apache.logging.log4j.Level
+import org.apache.daffodil.api.Diagnostic
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.Maybe._
+import org.apache.logging.log4j.core.config.plugins.{Plugin, PluginFactory}
+
+@Plugin(
+    name="CustomAppender",
+    category="Core",
+    elementType="appender"
+)
+class TDMLAppender 
+ extends AbstractAppender ("org.apache.daffodil.tdml", ThresholdFilter.createFilter(Level.INFO, Result.ACCEPT, Result.DENY), null, false, null) {
+
+    import TDMLAppender._
+
+    override def append(event: LogEvent): Unit = {
+        diagnostics = new LogDiagnostic(event.getMessage.getFormattedMessage) +: diagnostics
+    }
+}
+
+object TDMLAppender{
+    var diagnostics: Seq[Diagnostic] = Nil
+    var diagnostics2: Seq[String] = Nil
+
+    @PluginFactory
+    def createAppender(): TDMLAppender = {
+            new TDMLAppender
+    }
+
+    def getDiagnostics(): Seq[Throwable] = {
+        diagnostics
+    }
+
+    def clearDiagnostics(): Unit = {
+        diagnostics = Nil
+    }
+}
+
+class LogDiagnostic(msg: String)
+  extends Diagnostic(Nope, Nope, Nope, Maybe(msg)) {
+  override def isError = false
+  override def modeName = "Log"
+}
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -177,7 +177,8 @@ class DFDLTestSuite private[tdml] (
   val defaultValidationDefault: String,
   val defaultImplementationsDefault: Seq[String],
   val shouldDoErrorComparisonOnCrossTests: Boolean,
-  val shouldDoWarningComparisonOnCrossTests: Boolean)
+  val shouldDoWarningComparisonOnCrossTests: Boolean,
+  val shouldDoLogComparisonOnCrossTests: Boolean)
   extends HasSetDebugger {
 
   // Uncomment to force conversion of all test suites to use Runner(...) instead.
@@ -599,6 +600,7 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   lazy val optExpectedOrInputInfoset = (testCaseXML \ "infoset").headOption.map { node => new Infoset(node, this) }
   lazy val optExpectedErrors: Option[ExpectedErrors] = (testCaseXML \ "errors").headOption.map { node => ExpectedErrors(node, this) }
   lazy val optExpectedWarnings: Option[ExpectedWarnings] = (testCaseXML \ "warnings").headOption.map { node => ExpectedWarnings(node, this) }
+  lazy val optExpectedLogs: Option[ExpectedLogs] = (testCaseXML \ "loggers").headOption.map { node => ExpectedLogs(node, this) }
   lazy val optExpectedValidationErrors: Option[ExpectedValidationErrors] = (testCaseXML \ "validationErrors").headOption.map { node => ExpectedValidationErrors(node, this) }
 
   val tcName = (testCaseXML \ "@name").text
@@ -689,6 +691,7 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
     nBits: Option[Long],
     errors: Option[ExpectedErrors],
     warnings: Option[ExpectedWarnings],
+    loggers: Option[ExpectedLogs],
     validationErrors: Option[ExpectedValidationErrors],
     validationMode: ValidationMode.Type,
     roundTrip: RoundTrip,
@@ -864,6 +867,7 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
         nBits,
         optExpectedErrors,
         optExpectedWarnings,
+        optExpectedLogs,
         optExpectedValidationErrors,
         validationMode,
         roundTrip,
@@ -876,10 +880,11 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
     diagnostics: Seq[Throwable],
     errors: ExpectedErrors,
     optWarnings: Option[ExpectedWarnings],
+    optLogs: Option[ExpectedLogs],
     implString: Option[String]): Unit = {
     Assert.usage(this.isNegativeTest)
 
-    // check for any test-specified errors or warnings
+    // check for any test-specified errors, warnings, or logs
     if (!isCrossTest(implString.get) ||
       parent.shouldDoErrorComparisonOnCrossTests)
       VerifyTestCase.verifyAllDiagnosticsFound(diagnostics, Some(errors), implString)
@@ -887,6 +892,10 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
     if (!isCrossTest(implString.get) ||
       parent.shouldDoWarningComparisonOnCrossTests)
       VerifyTestCase.verifyAllDiagnosticsFound(diagnostics, optWarnings, implString)
+
+    if (!isCrossTest(implString.get) ||
+      parent.shouldDoLogComparisonOnCrossTests)
+      VerifyTestCase.verifyAllDiagnosticsFound(diagnostics, optLogs, implString)
   }
 }
 
@@ -901,6 +910,7 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
     optLengthLimitInBits: Option[Long],
     optExpectedErrors: Option[ExpectedErrors],
     optExpectedWarnings: Option[ExpectedWarnings],
+    optExpectedLogs: Option[ExpectedLogs],
     optExpectedValidationErrors: Option[ExpectedValidationErrors],
     validationMode: ValidationMode.Type,
     roundTrip: RoundTrip,
@@ -917,7 +927,7 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
         compileResult.right.foreach {
           case (_, proc) => {
             processor = proc
-            runParseExpectSuccess(dataToParse, nBits, optExpectedWarnings, optExpectedValidationErrors, validationMode, roundTrip, implString)
+            runParseExpectSuccess(dataToParse, nBits, optExpectedWarnings, optExpectedLogs, optExpectedValidationErrors, validationMode, roundTrip, implString)
           }
         }
       }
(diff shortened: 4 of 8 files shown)


mbeckerle (approved):
+1 I particularly liked the extra test to insure no false positive where logs were just silently being ignored.

build.sbt
```diff
@@ -104,6 +104,7 @@ lazy val sapi             = Project("daffodil-sapi", file("daffodil-sapi")).conf
 lazy val tdmlLib             = Project("daffodil-tdml-lib", file("daffodil-tdml-lib")).configs(IntegrationTest)
                               .dependsOn(macroLib % "compile-internal", lib, io, io % "test->test")
                               .settings(commonSettings)
+                              .settings(libraryDependencies ++= Dependencies.core)
```
tuxji, line 107: Why is this change even needed?  You have `.settings(commonSettings)` on the previous line and that calls libraryDependencies ++= Dependencies.common, in short, core ++ infoset ++ test.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala
```diff
@@ -91,6 +91,8 @@ object Runner {
    */
   def defaultShouldDoWarningComparisonOnCrossTests = false
 
+  def defaultShouldDoLogComparisonCrossTests = false
```
tuxji, line 94: Please add `On` for consistency, `defaultShouldDoLogComparisonOnCrossTests`.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.logging.log4j.core.filter.ThresholdFilter
+import org.apache.logging.log4j.core.LogEvent
+import org.apache.logging.log4j.core.Filter.Result
+import org.apache.logging.log4j.Level
+import org.apache.daffodil.api.Diagnostic
+import org.apache.daffodil.util.Maybe
+import org.apache.daffodil.util.Maybe._
+import org.apache.logging.log4j.core.config.plugins.{Plugin, PluginFactory}
+
+@Plugin(
```
tuxji, line 30: Please add comments here and below explaining what each class and object is used for.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -0,0 +1,67 @@
...
+    def getDiagnostics(): Seq[Throwable] = {
+        diagnostics
+    }
+
+    def clearDiagnostics(): Unit = {
+        diagnostics = Nil
+    }
+}
+
+class LogDiagnostic(msg: String)
```
tuxji, line 63: I can't find where LogDiagnostic is used?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -76,7 +76,10 @@ import org.apache.daffodil.util.Misc.hex2Bits
 import org.apache.daffodil.util.SchemaUtils
 import org.apache.daffodil.xml.DaffodilXMLLoader
 import org.apache.daffodil.xml.XMLUtils
-
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.{Logger => log4jLogger}
```
tuxji, line 80: We capitalize classes by convention, `Log4jLogger`.

project/Dependencies.scala
```diff
@@ -32,6 +32,7 @@ object Dependencies {
     "com.typesafe" % "config" % "1.4.2",
     "org.apache.logging.log4j" %% "log4j-api-scala" % "12.0",
     "org.apache.logging.log4j" % "log4j-api" % "2.17.2",
+    "org.apache.logging.log4j" % "log4j-core" % "2.17.2",
     "org.apache.logging.log4j" % "log4j-core" % "2.17.2" % "it,test",
```
tuxji, line 36: I believe line 36 is extraneous now that you've added line 35.  If log4j-core is available in compile scope, it'll also be available in test and integration test scopes too.

tuxji (approved):
+1

Only a few minor changes needed.  You can push a commit making these changes, then rebase, squash, and push another commit (both commits still will be available for review in GitHub's PR ui).

tuxji:
After your next commit, wait for the CI tests to finish so we can check whether the single windows error was a fluke or not:

```
[error] Test org.apache.daffodil.tdml.TestTDMLRunnerLogs.test_unparserLogWhenExpectingError failed: org.apache.daffodil.tdml.TDMLExceptionImpl: (Implementation: daffodil) "Diagnostic message(s) were expected but not found."
[error] Expected: dfdlx:trace
[error]  No diagnostic messages were issued., took 0.65 sec
```

It will be good to run the CI tests two more times (after your next commit, and then again after your squashed commit).

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.daffodil.util.Maybe._
+import org.apache.logging.log4j.core.config.plugins.{Plugin, PluginFactory}
+
+@Plugin(
+    name="CustomAppender",
+    category="Core",
+    elementType="appender"
+)
+class TDMLAppender 
+ extends AbstractAppender ("org.apache.daffodil.tdml", ThresholdFilter.createFilter(Level.INFO, Result.ACCEPT, Result.DENY), null, false, null) {
```
stevedlawrence, line 36: Is there any value in making the log level configurable, maybe with something like:
```xml
<tdml:logs level="info">
....
</tdml:logs>
```
I imagine most tests only care about warnings logs and so would be a reasonable default, but this would allow bumping up the level to info/debug/etc for tests that are checking for something very specific?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -0,0 +1,67 @@
...
+    import TDMLAppender._
+
+    override def append(event: LogEvent): Unit = {
+        diagnostics = new LogDiagnostic(event.getMessage.getFormattedMessage) +: diagnostics
+    }
+}
+
+object TDMLAppender{
+    var diagnostics: Seq[Diagnostic] = Nil
+    var diagnostics2: Seq[String] = Nil
```
stevedlawrence, line 47: These var's in a object doesn't feel thread safe? Could there be issues if multiple TDML tests are run in parrallel?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -599,8 +606,17 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
...
+  lazy val optExpectedLogs: Option[ExpectedLogs] = (testCaseXML \ "logs").headOption.map { node => ExpectedLogs(node, this) }
   lazy val optExpectedValidationErrors: Option[ExpectedValidationErrors] = (testCaseXML \ "validationErrors").headOption.map { node => ExpectedValidationErrors(node, this) }
 
+  if(LogManager.getRootLogger().asInstanceOf[log4jLogger].getAppenders().get("org.apache.tdml")==null && optExpectedLogs.isDefined){
+    val tdmlAppender = TDMLAppender.createAppender
+    val loggers = LogManager.getRootLogger().asInstanceOf[log4jLogger]
+    loggers.addAppender(tdmlAppender)
+    tdmlAppender.start()
+    Configurator.setLevel("org.apache.daffodil", Level.INFO)
+  }
```
stevedlawrence, line 618: Similar question as above, how does this deal with tests run in multiple thredas? I'm not sure exaclty what's happening here but it seems like it's changing the global logger, which could affect other tests?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1149,6 +1183,7 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
     dataToParse: InputStream,
     lengthLimitInBits: Long,
     warnings: Option[ExpectedWarnings],
+    loggers: Option[ExpectedLogs],
```
stevedlawrence, line 1186: Should this be `logs` instead of `loggers`?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1451,6 +1491,10 @@ case class UnparserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
       if (!isCrossTest(implString.get) ||
         parent.shouldDoWarningComparisonOnCrossTests)
         VerifyTestCase.verifyAllDiagnosticsFound(actual.getDiagnostics, optWarnings, implString)
+      
+      if (!isCrossTest(implString.get) ||
+        parent.shouldDoLogComparisonOnCrossTests)
```
stevedlawrence, line 1496: Why are there all these missing codecov lines? Seems like we have tests that should hit them?

daffodil-tdml-processor/src/test/resources/test/tdml/testLogs.tdml
```diff
@@ -0,0 +1,195 @@
...
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:warnings>
+      <tdml:warning>Schema definition warning</tdml:warning>
+    </tdml:warnings>
+    <tdml:logs>
+      <tdml:log>dfdlx:trace</tdml:log>
+    </tdml:logs>
+  </tdml:unparserTestCase>
+</tdml:testSuite>
```
stevedlawrence, line 195: Tests look nice. I wonder if there are any CLI tests that can be removed now that this has the ability to check logs? Previously that could only be done via the CLI.

project/Dependencies.scala
stevedlawrence: log4j-core was intentionally not part of core. The assumption was that users that get Daffodil via maven can add whatever log4j implementation as a dependency that they want. This way we don't force a particular implementation.

With the tdml-runner now using code that depends on log4j-core, that may not be possible now. I'm not sure if there's away around this, but maybe it's not a big deal if log4j-core is now a hard dependency for Daffodil?

Note that if we remove line 36, we can also remove line 50 in the cli dependency (our CLI does bundle log4j-core).

stevedlawrence (commented):
I have some concerns about thread safety. Are they warranted or am I just missing something?

Also, not sure if it's a big deal, but I think this also breaks the ability to provide a different log4j implementation.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: Much of the features for cross testing don't get exercised unless you run the IBM cross testing rig. 

To test these without that we'd have to build a mock cross tester that actually just uses daffodil.

project/Dependencies.scala
mbeckerle: Do we need log4j-core for daffodil-core test, or only daffodil-tdml-lib or even maybe only daffodil-tdml-processor?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: Ah right, I missed that this is only cross testing related. Not worth it to create a mock cross tester. Resolving this conversation.

stevedlawrence:
> The TDML runner needs log4j core, so the CLI will because it can run TDML tests. But Daffodil proper, used via API as it would be when embedded into a system, should not need this logger specified. That tradeoff feels ok to me.

Good point. So it seems in that case the new log4j-core API should not be part of the core dependencies, but instead should be added only to the tdmlLib project, similar to how we add it to the CLI? So something like:

*Dependencies.scala*
```scala
class Dependencies {
  ....

  lazy val tdml = Seq(
    "org.apache.logging.log4j" % "log4j-core" % "2.17.2"
  )
}
```

*build.sbt*
```scala
lazy val tdmlLib = Project(...)
                   ...
                   .settings(libraryDependencies ++= Dependencies.tdml)
```

project/Dependencies.scala
stevedlawrence: I think we only need it for the new `TDMLAppender`, which is defined and used in `daffodil-tdml-lib`. So I think only lib needs this new log4j-core dependency.

Though you raise a good point, I think daffodil-tdml-procesor should be the thing aggregating logs, since different DFDL implementations will have different mechanisms for capturing logs. Daffodil uses log4j, but IBM DFDL might use something completely different. Seems maybe the new TDMLAppender wants to be moved to daffodil-tdml-processor, and the `DaffodilTDMLDFDLProcessor` must return the logs as part of the Parse/Unparse result and the TDMLRunner can then compare against the expected results?

stevedlawrence: Note that adding log4j-core as a dependency to the daffodil-tdml-proessor has a good side effect: Our schema projects that depend on daffodil-tdml-processor for running tests will now transitively pull in log4j-core, and it should get rid of that warning about not having a logger available.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -38,25 +38,32 @@ class TDMLAppender
     import TDMLAppender._
 
     override def append(event: LogEvent): Unit = {
-        diagnostics = new LogDiagnostic(event.getMessage.getFormattedMessage) +: diagnostics
+        checkForKey
+        diagnostics = diagnostics.updated(Thread.currentThread.getId, diagnostics(Thread.currentThread.getId) :+ new LogDiagnostic(event.getMessage.getFormattedMessage))
```
mbeckerle, line 42: line too long

mbeckerle (approved):
Fix one too-long line. Otherwise +1

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
```diff
@@ -38,25 +38,32 @@ class TDMLAppender
...
-        diagnostics = new LogDiagnostic(event.getMessage.getFormattedMessage) +: diagnostics
+        checkForKey
+        diagnostics = diagnostics.updated(Thread.currentThread.getId, diagnostics(Thread.currentThread.getId) :+ new LogDiagnostic(event.getMessage.getFormattedMessage))
     }
 }
 
 object TDMLAppender{
-    var diagnostics: Seq[Diagnostic] = Nil
-    var diagnostics2: Seq[String] = Nil
+    var diagnostics: Map[Long, Seq[Diagnostic]] = Map()
```
stevedlawrence, line 47: Does it make sense to use a mutable map and make this a val?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
stevedlawrence, line 42: I don't think this line is thread safe? If `append` is called by two different threads at the same time, is it possible that one of the updates will be lost? Can this function be synchronized like the other functions?

tuxji:
Please resolve everyone's comments, then we can merge your pull request.

build.sbt
stevedlawrence: I don't think this comment was addressed. I agree that this shouldn't be needed.

project/Dependencies.scala
stevedlawrence: I don't think this comment was addressed. log4-core should only be a dependency to the daffodil-tdml-processor project, not as a core dependency. And the TDMLAppender and related code also wants to be moved to the same project. Other DFDL implemenations (e.g. IBM DFDL) may use a different mechanism for their logs.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
alexanderrevello: its done by log4j so it is thread safe but I will add synchronized anyway

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -864,29 +876,42 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
...
         validationMode,
         roundTrip,
-        implString)
+        implString,
+        parent.tdmlLogger)
 
     }
+
+    if (parent.tdmlLogger ne null)
+      parent.tdmlLogger.clearDiagnostics()
```
stevedlawrence, line 889: This `clearDiagnostics` thing is sort of an artifact of our logger and the way it works (i.e. dealing with global state). It concerns me that other logging systems might need very different abstractions that our basic appended might not support. I wonder if there's a good way to make it so that TDMLProcessor implementations do not need to be restricted to a specific logger abstraction?

For example, what if creating/configuring the logging mechanism is handled entirely within the TDMLProcessor implementation? And the TDML Runner has no concept of that logging mechanism? Then TDML Processor can capture logs however it wants and add them to the CompileResult/ParseResult/UnaprseResult. Then the TDMLRunner can then access them with a new function (e.g. `getLogs`) and do the comparisons you do. This way, the only thign the TDMLRunner needs to know is there is a way to get logs from a result, and it's entirely up to the processor how to best do that.

An added benefit is you don't need to pass around the tdmlLogger to a bunch of functions, since it's all contained in the TDML Processor and the Parse/UnparseResults, which we already pass around where needed.

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunnerLogs.scala
```diff
@@ -0,0 +1,114 @@
...
+object TestTDMLRunnerLogs {
+  val runner = Runner("/test/tdml/", "testLogs.tdml")
+  val tdmlAppender = TDMLAppender.createAppender
+
+  if(LogManager.getRootLogger().asInstanceOf[Log4jLogger].getAppenders().get("org.apache.tdml")==null){
+        val loggers = LogManager.getRootLogger().asInstanceOf[Log4jLogger]
+        loggers.addAppender(tdmlAppender)
+        tdmlAppender.start()
+        Configurator.setLevel("org.apache.daffodil", Level.INFO)
+    }
```
stevedlawrence, line 39: If we do as mentioned above, could this be moved into the TDML Processor? Simplifies the logic in the test files a bit.

project/Dependencies.scala
```diff
@@ -58,4 +58,8 @@ object Dependencies {
   lazy val schematron = Seq(
     "net.sf.saxon" % "Saxon-HE" % "11.2",
   )
+
+  lazy val log4jcore = Seq(
```
stevedlawrence, line 62: I'd suggest calling this `tdml_processor`. If we ever add new `tdml_processor dependencies`, they can go in this variable.

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/tdml/TestTDMLRunnerLogs.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  @AfterClass def shutDown(): Unit = {
+    runner.reset
+  }
+}
+
+class TestTDMLRunnerLogs {
+  import TestTDMLRunnerLogs._
+
+  @Test def test_logWhenExpectingSuccess() = { runner.runOneTest("logWhenExpectingSuccess", tdmlAppender) }
```
stevedlawrence, line 49: Also, if we do as mentioned above, then we don't have to explicitly pass in something to capture logs. It's just something that is done by default by the TDML Processor and is always available to any TDML test.

stevedlawrence (approved):
Looks good, thoughts on potentially moving even more logic into the TDML processor?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLAppender.scala
mbeckerle: Sure, but that doesn't change the fact that a singleton object has mutable state in it. I think this appender needs to be an allocated object, not a singleton.

PR #795, opened 2022-05-16
Extend runtime2 to N-bit booleans and integers (1 <= N <= 64)

tuxji:
Allow runtime2 to read and write N-bit booleans and integers where N
is an explicit length from 1 to 64 bits.  Add C-language unit tests
for easier debugging of N-bit-long numbers (requires Criterion test
framework with MIT license, https://github.com/Snaipe/Criterion, but
never used by or deployed with Daffodil itself so no need to change
Daffodil's LICENSE or NOTICE files).

Also avoid overwriting fields when initializing ERDs and fields of
structures containing choices within choices.  Tweak how runtime2
outputs float numbers in XML to reduce textual differences between
runtime1 and runtime2.  See more detailed file-level changelog below.

DAFFODIL-2696

Makefile: Add test goal to build and run new C-language unit tests.
Tidy C-language test source files in clean, format, and iwyu goals.

daffodil_main.c: Call flushUState after unparsing to write out any
unwritten bits still buffered inside ustate.

xml_writer.c: Revise how fixNumberIfNeeded and xmlSimpleElem print
floats to reduce textual differences between runtime1 and runtime2 to
a few cases where different numbers of significant digits get printed.

infoset.c: Add flushUState to flush any buffered bits not written yet.

intoset.h: Remove initSelf from ERD struct since we call initSelf
statically and don't need to look it up from ERD objects.  Add or
modify PState/UState fields to store bits not yet written or consumed
by Daffodil in a fractional byte.  Declare flushUState.

parsers.c: Add read_bits helper function to store bits not yet
consumed by Daffodil in a fractional byte.  Replace helper macros with
helper functions for easier debuggability since the debugger cannot
step through helper macros (they expand to only one source line).
Merge parse_{bool8, bool16, bool32} into just one parse_bool.  Change
parse_fill_bytes to take end_bitPos0b instead of end_bytePos0b.

parsers.h: Pass num_bits to parse functions to allow them to read
numbers using fewer bits.  Merge parse_{bool8, bool16, bool32} into
just one parse_bool.  Change parse_fill_bytes to take end_bitPos0b
instead of end_bytePos0b.

unparsers.c: Add write_bits helper function to store bits not yet
written by Daffodil in a fractional byte.  Replace helper macros with
helper functions for easier debuggability since the debugger cannot
step through helper macros (they expand to only one source line).
Merge unparse_{bool8, bool16, bool32} into just one unparse_bool.
Change unparse_fill_bytes to take end_bitPos0b instead of
end_bytePos0b.

unparsers.h: Pass num_bits to unparse functions to allow them to write
numbers using fewer bits.  Merge unparse_{bool8, bool16, bool32} into
just one unparse_bool.  Change unparse_fill_bytes to take end_bitPos0b
instead of end_bytePos0b.

bits.c: Add C unit tests to check that unparse and parse functions can
write and read numbers using N bits correctly (if not, these tests
help debug these functions more quickly).  These unit tests require a
C test framework library called Criterion which has to be built and
installed manually on Ubuntu 20.04, so we don't want to make build.sbt
or main.yml run these unit tests.  Developers can manually install
Criterion and run the tests with "make test" during development.

extras.c: Define rootElement to avoid an undefined reference when
linking C unit tests.

CodeGenerator.scala: Skip test files when compiling C source files
from within Daffodil itself.

BinaryBooleanCodeGenerator.scala: Pass boolean lengths from 1 to 32
bits to booleanValueAddField and parse/unparse functions as `num_bits`
parameter.  Replace bool8, bool16, bool32 function names with just one
bool function name.

BinaryFloatCodeGenerator.scala: Makes no sense to accept any lengths
other than 32 bits or 64 bits, but modify to pass lengths in bits to
valueAddField and parse/unparse functions anyway.

BinaryIntegerKnownLengthCodeGenerator.scala: Accept all integer
lengths between 1 and 64 bits in DFDL.  Map to quantized lengths (8,
16, 32, 64 bits) for C numbers but pass actual lengths in bits to
valueAddField and parse/unparse functions.

BinaryValueCodeGenerator.scala: Pass actual lengths in bits from
valueAddField to parse/unparse functions.

CodeGeneratorState.scala: Avoid overwriting fields when initializing
ERDs and fields of structures containing choices within choices by
separating initStatements into initERDStatements and
initSelfStatements and making C code call initSelf functions in
rootElement but call initERD functions in initChoice functions when
choices are resolved.  Skip whitespace characters when generating
relative paths from choice dispatch key expressions.  Insert padding
using lengths in bits instead of lengths in bytes.  Relax pattern
matching in getPrimType to accept all boolean and integer lengths
allowed by DFDL.

ElementParseAndUnspecifiedLengthCodeGenerator.scala: Update a couple
of comments.

HexBinaryCodeGenerator.scala: Move some initialization from initSelf
to initERD to avoid overwriting fields when initializing choices
within choices.

{NestedUnion,ex_nums}/generated_code.c: Show changes to generated code
(updated automatically by sbt compile) due to code generator changes
as well as renaming and rearrangement of elements in ex_nums.dfdl.xsd.

ex_nums}/generated_code.h: Show changes to generated code (updated
automatically by sbt compile) due to renaming and rearrangement of
elements in ex_nums.dfdl.xsd.

ex_nums.dat, ex_nums.dat.xml: Reduce number of significant digits in
doubles and floats to help eliminate textual differences between
runtime1 and runtime2 (allowing ex_nums.runtime{1,2}.dat.xml to be
merged into ex_nums.dat.xml).  Change some test elements to test N-bit
booleans and numbers.

ex_nums.dfdl.xsd: Force all elements to be 1-bit aligned to eliminate
differences between runtime1 and runtime2 (runtime2 doesn't pay any
attention to alignment yet).  Change some test elements to test N-bit
booleans and numbers.

ex_nums.error.dat, ex_nums.error.dat.xml: Same changes as ex_nums.dat,
ex_nums.dat.xml (also merging ex_nums.runtime{1,2}.error.dat.xml into
ex_nums.error.dat.xml).

ex_nums.tdml: Add comments to show how to run tests with daffodil
commands.  Since float numbers (for our test cases) look the same now,
need only one ex_nums.dat.xml and one ex_nums.error.dat.xml for both
runtime1 and runtime2 tests.

egress_xdcc_bw.15.dat, egress_xdcc_bw.15.dat.xml: Reduce number of
significant digits in alt, lon doubles to eliminate differences
between runtime1 and runtime2.

egress_xdcc_bw.19.dat, egress_xdcc_bw.19.dat.xml: Reduce number of
significant digits in some x,y,z doubles to eliminate differences
between runtime1 and runtime2.

ingress_xdcc_bw.115.dat, ingress_xdcc_bw.115.dat.xml: Reduce number of
significant digits in some alt,bearing,confidence,lat,lon,speed
doubles to eliminate differences between runtime1 and runtime2.


tuxji:
@stevedlawrence I can't find much documentation on sbt's GlobFilter.  How do I modify runtime2 / Compile / cSources to exclude any *.c files in the tests subdirectory (or at least exclude bits.c) in order to avoid needing to install Criterion during CI?  I found <https://www.scala-sbt.org/1.x/docs/Globs.html> but cSources doesn't seem to be using Globs?

stevedlawrence:
> How do I modify runtime2 / Compile / cSources to exclude any *.c files in the tests subdirectory

I'm not sure, but the easiest way might be to remove the recursive glob (i.e. the `**`) and instead add multiple lines with a normal glob for each directory we care about, e.g.:
```scala
((Compile / resourceDirectory).value / "org" / "apache" / "daffodil" / "runtime2" / "c" / "libcli" * GlobFilter("*.c")).get(),
((Compile / resourceDirectory).value / "org" / "apache" / "daffodil" / "runtime2" / "c" / "libruntime" * GlobFilter("*.c")).get()
```

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/unparsers.c
```diff
@@ -17,135 +17,443 @@
...
+        size_t num_bits_unused = BYTE_WIDTH - num_bits_fill;
+        if (num_bits_unused > num_bits) num_bits_unused = num_bits;
+        if (num_bits_unused)
+        {
+            ustate->bitPos0b += num_bits_unused;
+            ustate->unwritBits <<= num_bits_unused;
+            ustate->unwritBits |= HIGH_BITS(whole_byte, num_bits_unused);
+            ustate->unwritLen += num_bits_unused;
+            num_bits -= num_bits_unused;
+        }
```
stevedlawrence, line 96: I think I have the same questions for uparse as parse related to bitPos0b and when it's updated.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/parsers.c
```diff
@@ -18,140 +18,508 @@
...
+    else
+    {
+        // Convert data endianness to host endianness
+        buffer.integer = le32toh(buffer.integer);
+
+        // Need only num_bits so remove any padding bits
+        size_t msb_ix = (num_bits - 1) / BYTE_WIDTH;
+        size_t shift = (BYTE_WIDTH - num_bits % BYTE_WIDTH) % BYTE_WIDTH;
+        buffer.bytes[msb_ix] >>= shift;
+    }
```
stevedlawrence, line 156: Can/should this call `parse_endian_uint32` and then compare the result to true/false rep? That should minimize duplicate code, right?

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/generators/CodeGeneratorState.scala
```diff
@@ -495,29 +509,27 @@ class CodeGeneratorState(private val root: ElementBase) {
...
+          case n if n <= 8 =>  PrimType.UnsignedByte
+          case n if n <= 16 => PrimType.UnsignedShort
+          case n if n <= 32 => PrimType.UnsignedInt
+          case n if n <= 64 => PrimType.UnsignedLong
+          case _ =>  e.SDE("Unsigned integer lengths longer than 64 bits are not supported.")
         }
       case p => e.SDE("PrimType %s is not supported in C code generator.", p.toString)
     }
-    if (primType != e.optPrimType.get)
-      e.SDW(WarnID.IgnoreDFDLProperty, "Ignoring PrimType %s, using %s", e.optPrimType.get.toString, primType.toString)
```
stevedlawrence, line 520: Why remove this warning? This seems like an mistake in a schema that a user should be notified of?

stevedlawrence (approved):
+1 :+1: 

I think I may not have followed the bitPosition stuff, might just be missing something obvious. Just a few minor comments, otherwise, looks good to me

tuxji (commented):
All great comments, all responses below.  Next commit will resolve these comments.

Ready for second reviewer's comments.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libcli/xml_writer.c
```diff
@@ -22,30 +22,55 @@
...
 #include "cli_errors.h"  // for CLI_XML_DECL, CLI_XML_ELEMENT, CLI_XML_WRITE, LIMIT_XML_NESTING
 #include "errors.h"      // for Error, Error::(anonymous)
 #include "stack.h"       // for stack_is_empty, stack_pop, stack_push, stack_top, stack_init
 // clang-format on
 
-// Fix a real number to conform to xsd:float syntax if needed
+// Fix a real number's syntax to remove as many textual differences
+// between runtime1/runtime2 as possible.  A better approach may be
+// to make TDMLRunner compare floats and doubles in XML numerically
+// with a small epsilon instead of textually character by character.
```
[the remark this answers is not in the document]
stevedlawrence: The TDML Runner uses the XML comparison logic XMLUtils.scala, with the type aware comparison happening in the `textIsSame` function: 

https://github.com/apache/daffodil/blob/main/daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala#L1194-L1214=

Unfortunately, the way it knows how to use special comparison logic is with the `xsi:type` attribute, which must be on either the actual infoset or the expected infoset. We've been putting them on the actual infoset when we need the type aware comparison since the runtimes don't current support it (see [DAFFODIL-2402]](https://issues.apache.org/jira/browse/DAFFODIL-2402)). Though, it might be a bit of a headache to add it for all doubles/floats--we don't need this very often.

If this makes Java and C mostly the same, maybe this is a better approach. I'm also not against change Java to match C if that's easier.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/parsers.c
```diff
@@ -18,140 +18,484 @@
...
+        size_t num_bytes = num_bits / BYTE_WIDTH;
+        if (num_bytes)
+        {
+            size_t count = fread(bytes, 1, num_bytes, pstate->stream);
+            if (count < num_bytes)
+            {
+                pstate->error = eof_or_error(pstate->stream);
+                return;
+            }
+            pstate->bitPos0b += count * BYTE_WIDTH;
```
stevedlawrence, line 57: As I think about this some more, I wonder if `read_bits` doesn't want to update bitPos0b until after it has successfully read the full num_bits, i.e. right before the function returns success?

For example, if we're at bit position X and we need Y bits for some number, if we can't read all Y bits then we probably want a diagnostic that says "Y bits were not available starting at bit position X". But if we increase bitPos0 before we finish reading all bits and then hit an error, the error message would look like "Y bits were not available starting at position Z", where Z is some position after X, and it might not make much sense.

So I'd argue bitPosition wants to be the position the last place where read_bits succeeded (even if we don't actually output those diagnostics like that). Which simplifies the logic, the lastline of the function just becomes `pstate->bitPos0b += num_bits` and it's the only place bitPos0b is changed in the function.

There might even be an argument that the bitPosition shouldn't be updated until the read bits are successfully converted to a number. That way if there was an error you could say something like "failed to convert X bits to a boolean, starting at position Y". Again, I don't know if runtime2 outputs that kind of diagnostic, but it makes me think that that is what bitPos wants to represent: the last position of a successful read.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/parsers.c
```diff
@@ -18,140 +18,484 @@
...
+    // Parse all unsigned integers in this helper function
+    union
+    {
+        uint8_t  bytes[sizeof(uint64_t)];
+        uint64_t integer;
+    } buffer;
+
+    // Read data bits
+    buffer.integer = 0;
+    read_bits(buffer.bytes, num_bits, pstate);
```
stevedlawrence, line 295: Should these calls to `read_bits` check for an error and return? I guess technically it doesn't matter since this will just end up converting 0x0000... to a number, but that's just wasted cycles.

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/parsers.c
```diff
@@ -18,140 +18,484 @@
...
+}
+
+// Parse all binary booleans, real numbers, and integers in helper
+// functions, but wrap calls for type safety and simpler calls
+
+void
+parse_be_bool(bool *number, size_t num_bits, int64_t true_rep, uint32_t false_rep, PState *pstate)
+{
+    uint64_t integer;
+    parse_endian_uint64(BIG_ENDIAN_DATA, &integer, num_bits, pstate);
```
stevedlawrence, line 329: Why uint64, aren't bools limited to 32 bits in runtime2?

stevedlawrence (approved):

daffodil-runtime2/src/main/resources/org/apache/daffodil/runtime2/c/libruntime/parsers.c
tuxji: I've made sure all calls to `read_bits` are followed by `if (pstate->error) return;` now (ditto for `write_bits` in unparsers.c).

tuxji (commented):
Good comments, all addressed in new commit.

Ready for your review, @mbeckerle!

tuxji:
Daffodil committers, we need more of you to review pull requests from time to time.  Please review my pull request so I can have a second +1.  Thanks!

mbeckerle (approved):
+1

PR #799, opened 2022-05-31
DAFFODIL-2369 Any character following "//" will be ignored in ByteDocumentPart/BitDocumentPart

OWL-Varun:


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2361,35 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+  *              If the string runs into "//", any character following it will be skipped (including valid ones).
+  *              Characters on a new line will be considered again, unless they are found after a "//"
+  * Parameters: String containing the acceptable characters (0,1 for bits) (0-9,A-F,a-f for bytes)
+  * Return: ListBuffer[Char] of the accepted characters from the ByteDocumentPart/BitsDocumentPart element
+  */
+  def commentCompatibleFormat(validCharacters: String) : ListBuffer[Char] = {
+    val tempListBuffer = new ListBuffer[Char]()
+    var noCommentBool = true
+
+    for (i <- 0 to partRawContent.length()-1){
```
mbeckerle, line 2376: In general, iterating things with an index is frowned upon by many experienced programmers now, as it's too hard to tell if you've done the iteration just right or not. For someting that needs to be super-fast maybe. But for this usage clarity is way more important than speed. 

A way to do this that is much easier to review is more functional.

E.g., 
first split the string at line-endings to get a list of strings, one per line.
Then split each line at "//" discarding all but the first part.
Then replace any char matching regular expression "[^0-9a-fA-F]" with "". (for bytes case)
Then concatenate and mkstring from all the remaining pieces. 

Each of those operations should be 1 line of code.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
```diff
@@ -345,6 +345,46 @@ class UnitTestTDMLRunner {
     assertEquals(expected, actual)
   }
 
+  @Test def testCommentBit(): Unit = {
```
mbeckerle, line 348: Corner case to add a test for:

```
<documentPart type="bits">
// this doc part contains no bits 
// at all. It is just comments.
// 101010101
</documentPart>
```

Also, I think this should not be an error:
```
<documentPart type="bits">01011010 // just a comment here no line ending </documentPart>
```
We should put a test in for this. 
The reason I like this is that it lets short things stay on a single line.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2214,20 +2215,16 @@ class ByteDocumentPart(part: Node, parent: Document) extends DataDocumentPart(pa
...
 }
 
 class BitsDocumentPart(part: Node, parent: Document) extends DataDocumentPart(part, parent) {
-  lazy val bitDigits = {
-    val res = partRawContent.split("[^01]").mkString
-    res
-  }
+  val validBits = "01"
+
+  lazy val tempListBuffer = commentCompatibleFormat(validBits)
```
mbeckerle, line 2226: naming. How about dataWithoutComments?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2361,35 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
     }
   }.trim.toUpperCase()
 
+  /*
+  * Description: Compare ByteDocumentPart/BitsDocumentPart string to a string of valid characters.
```
mbeckerle, line 2365: Instead of describing your algorithm here, imagine the user doesn't care about the algorithm they just need to know how to call it properly. I don't think you need to say more than "removes comments from '//' to end of line or end of document part element. Any characters other than those explicitly allowed are also considered comments and are removed."

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2361,35 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+  *              If the string runs into non-valid characters, skip over them.
+  *              If the string runs into "//", any character following it will be skipped (including valid ones).
+  *              Characters on a new line will be considered again, unless they are found after a "//"
+  * Parameters: String containing the acceptable characters (0,1 for bits) (0-9,A-F,a-f for bytes)
```
mbeckerle, line 2369: The parameter arg contains not just acceptable chars, but also can be comment chars or comment syntax introduced by //.

mbeckerle (changes requested):

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: I'd also suggest we pass in the string to remove comments from (e.g. rawPartContent), rather than requiring that it be a member of the class that the function accesses. That way this is more of a utility function (and could even go in some Utils object) and not something specific to certain classes that need a certain member variable.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
```diff
@@ -345,6 +345,46 @@ class UnitTestTDMLRunner {
...
+        val xml = <document bitOrder="LSBFirst">
+                <documentPart type="bits">00000010 //this is a label111</documentPart>
```
stevedlawrence, line 350: The syntax highlighting here is a little odd because // is a scala comment. I wonder if that could cause confusion in IDEs? Maybe we should use a comment character that is unique and doesn't necessarily mean anything special in scala as far as syntax highlighting goes?

Also, why did we decide not to use XML comments? Syntax highlighting would just work, and they should be pretty straightforward to remove using normal XML utilities. Is that just too verbose?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: >  Then replace any char matching regular expression "[^0-9a-fA-F]" with "". (for bytes case)

Should we consider that an error rather than replacing invalid characters with empty string? Isn't the main purpose of the comment syntax to avoid cases where we accidentally misinterpret characters? I suggest any characters not in a comment *must* be a valid character for that the document type, or a space to allow for alignment/readability. Any other character should throw a TDMLException about an invalid document part.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
tuxji: Yes, I also prefer functional implementations where you pass in everything, they return everything to you as a result, and they don't modify any mutable state (unless you pass it in as an argument too).

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
tuxji: Yes, I agree with performing more strict validation of the document part contents (any invalid character except whitespace must be in a comment or must throw an exception).

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
tuxji: [DAFFODIL-2369](https://issues.apache.org/jira/browse/DAFFODIL-2369) is worth reading to answer these questions.  // to end of line is only one of two C/C++ comment markers DAFFODIL-2369 suggests - I strongly suggest also implementing the /* */ convention in order to avoid weird syntax highlighting in Scala code like this file.  DAFFODIL-2369 also says breaking up the data into separate documentPart elements in order to use standard XML comments is not effective enough due to the clutter involved.

tuxji (changes requested):
I agree with Mike's and Steve's suggestions, and I have a suggestion of my own too (implement the /* */ convention as well as the // to end of line convention to avoid weird syntax highlighting inside Scala test files).

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: There are places where I know some non-whitespace separators are used for good effect E.g., things like

01|01 0001 10|11 

where the vertical bars are used to show where the byte boundaries or element boundaries are. So I think allowing whitespace and certain specific punctuation characters like brackets and parens in the mix is very useful. (at least '|()[].' are useful)

There is a fair amount of investment in old-style comments where people have done 

<documentPart type="bits"> trackStability 011 <documentPart>

So I think quite a lot of impact if that feature is removed allowing (and ignoring) those letters. 

Yes ultimately I would like to fix these, but many files (hundreds) of DFDL schema tests will have to be changed by users. 

I think we would need an option on the tdml:testSuite level to change the default of this from allowed to deprecated to disallowed.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
stevedlawrence: I'm not sure I'm sold on the the clutter argument--it's not great, but it's not awful.

That said, the bug mentions CDATA, which makes me think of what I think is a strong argument against using XML comments, though isn't mentioned in the bug: XML comments aren't treated like comments in CDATA blocks, it's just normal text.

It's not uncommon for TDML tests to used CDATA blocks in the documentPart to ensure IDE's don't rewrap XML and mess up formatting. My main reason for using XML comments is you could just do something like `(testSuite \ "document" \ "documentPart").text` and all the XML comments would be automatically removed. But that's not the case with XML Comments in CDATA, so we would still need special comment removing code anyways. And if we're doing that, we might as well pick something without any potential confusion that comes with XML comments in CDATA, and is maybe a little prettier.

I agree that adding /* */ feels important. Inline comments seem like they could be used a lot.

mbeckerle (changes requested):
Pretty close. I would like not to get warnings on '|()[]' characters inside bits/bytes. 

Some smaller scala-isms to fix.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2362,32 +2359,32 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+    //Throw exception if /* or */ still found. This means user input was not formatted correctly.
+    if(noCommentsBothFormats.contains("/*") || noCommentsBothFormats.contains("*/")){
+      throw TDMLException("Improper formatting of /* */ style comment", None)
     }
-    tempListBuffer
+
+    //Remove the new lines - concatenate after
+    lazy val noCommentsString = noCommentsBothFormats.split('\n').mkString
+
+    //Iterate over and check value of the characters, if invalid character found create log and skip over it
```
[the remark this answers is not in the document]
OWL-Varun: What are your thoughts on adding " " and "." to the noWarnChars variable, as well? Is that being too lenient or have you seen enough instances where " " and "." are used that we should accept not warning the user?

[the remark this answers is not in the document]
OWL-Varun: What are your thoughts on adding " " and "." to the noWarnChars variable, as well? Is that being too lenient or have you seen enough instances where " " and "." are used that we should accept not warning the user?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2358,35 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
     }
   }.trim.toUpperCase()
 
+  /*
+  * Allow "//" and "/* */" (inline) to act as comments.
+  * Any characters not explicitly allowed are also considered comments and are removed.
+  */
+  def removeComments(validCharactersSet : String, userData : String ) : String = {
+    lazy val doubleForwardPattern = "//.*".r
+    lazy val openClosePattern = "/[*].*[*]/".r
```
stevedlawrence, line 2367: I think we need to make the `.*` non-greedy? Image we had something like this:
```xml
<documentPart type="bits">
01010101 /* something */ 01010101 /* something else */
</documentPart>
```
If it's a greedy match, everything will be considered a comment except for the first set of binary numbers. We should add a test with something like this.

Another related thought, do we want to allow that syntax to work with multiple lines, e.g.:

```xml
<documentPart type="bits">
01010101 /* something that spans
            across multiple lines */
10101010 /* something else that
            spans multiple lines */
</documentPart>
```

That seems pretty reasonable to me, in which case the dot need to match newlines. Note that the dot in the double-slash comment doesn't need to match newlines, since by default it ends at a newline.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2358,35 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+    lazy val noCommentsForward = doubleForwardPattern.replaceAllIn(userData,"")
+    lazy val noCommentsBothFormats = openClosePattern.replaceAllIn(noCommentsForward,"")
+
+    //Throw exception if /* or */ still found. This means user input was not formatted correctly.
+    if(noCommentsBothFormats.contains("/*") || noCommentsBothFormats.contains("*/")){
+      throw TDMLException("Improper formatting of /* */ style comment", None)
+    }
+
+    //Remove the new lines - concatenate after
+    lazy val noCommentsString = noCommentsBothFormats.split('\n').mkString
```
stevedlawrence, line 2379: Is it possible for `\r` to also exist in documentParts, like if running on windows? Or does it all get canonicalized to `\n`?

mbeckerle: So, in theory, XML loading should convert both \r\n and \r to lone \n. 

But, this has proven problematic for people like us, who want to use XML as a data language; hence, there is chatter on the internet (e.g., saw something about this on stack overflow) about options on loaders to shut this off.

I think the right thing is stick an Assert.invariant(...test..) check in the code that scans for \r characters and fails if any are detected. With a comment that says "// XML is not supposed to contain CR characters ever", that would make it clear without adding logic.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
OWL-Varun: Good catch on the regex being greedy.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2358,30 +2358,48 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
   /*
   * Allow "//" and "/* */" (inline) to act as comments.
-  * Any characters not explicitly allowed are also considered comments and are removed.
+  * Any valid XML characters not explicitly allowed are also considered comments and are removed.
   */
-  def removeComments(validCharactersSet : String, userData : String ) : String = {
-    lazy val doubleForwardPattern = "//.*".r
-    lazy val openClosePattern = "/[*].*[*]/".r
+  def canonicalizeData(validCharactersSet : String, userData : String) : String = {
+    val noWarnCharsSet = "|()[]."
```
mbeckerle, line 2379: In test testMIL2045_47001D_Page70_TableB_I, we make extensive use of x and X to illustrate bits of a byte copying exactly the way this is done in a particular spec document.

I think that X and x should be added here. 

I also think you should create a new version of testMIL2045_47001D_Page70_TableB_I which uses the new comment syntax. (The existing test should generate warnings but still work)

Suggest you do two such rewrites. One with /*... */ comments, which can be wrapped around the early field name part of each documentPart line, the other would flip and put the bits part first, and use the "//" comments and move the field name comments to after the bits.

mbeckerle (changes requested):
Please add X and x to the tolerated characters in bits/bytes, and add a couple existing tests revised to show the new way to do this. 

Remove unnecessary java style getters.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2358,30 +2358,48 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
     }
   }.trim.toUpperCase()
 
+  private lazy val doubleForwardPattern = "//.*".r
+  private lazy val openClosePattern = "/[*](.|\n)*?[*]/".r
+
+  //Accessor for doubleForwardPattern regex
```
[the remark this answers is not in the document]
stevedlawrence: Yes, that is a bit similar. I think the difference is that these patterns never really changed. They don't depend on any TDML files at all--they are essentially constant. Although the function uses those patterns, the result of the canonicalizeData function really is just a function on two inputs, the valid charset and the user data. We could even move those patterns into the function itself to make that more clear.

Note, an alternative approach that might make this distinction more clear is if this function was a utility that doesn't know anything about where it's being called. In that case, these regexs never change and would just live along with that utility. But the data to canonicalize depends on the caller and so that should be passed in. For example, if we had something like this:

``` scala
object Misc {
  private lazy val doubleForwardPattern = ...
  private lazy val openClosePattern = ...
  private lazy val noWarnChars = ...

  def canonicalizeTestData(validCharacterSet: String, userData: String): String = {
    ...
  }
}
```

So the patterns and no warn chars are all defined in this object and never change. The canonicalization function will use those to canonicalize, but really it's a just function of the two input parameters.

And this would be called like:

```scala
val bitsDigits = Misc.canonicalizeTestData(validBits, partRawContent)
```

Note that something like this has a benefit that the pattern regexs are only compiled once, rather than every time we create a new DocumentPart, so it's going to be a bit more efficient. Though I'm not sure how much that would really matter.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2364,6 +2358,53 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+    }
+
+    //Split on new line and white spaces - concatenate after
+    val noCommentsString = noCommentsBothFormats.split("\\s").mkString
+
+    //Check value of the characters, if invalid character found create log and skip over it
+    val validData = noCommentsString.map{ch =>
+      if(validCharactersSet.contains(ch)) ch
+      else { if(!noWarnCharsSet.contains(ch)) doWarning=true; ""}
+    }
```
stevedlawrence, line 2399: This doesn't feel very scala-y, especially with the mutable doWarning. There's probably a handful of ways to do this, but one approach is something like this:

```scala
val noWarns = noCommentsString.filter { !noWarnChars.contains(_) } // remove ignored characters
val hasInvalid = noWarns.exists { !validCharacters.contains(_) } // test if invalid characters exist
if (hasInvalid) Logger.log.warn(...)
val validData = noWarns.filter { validCharacters.contains(_) } // remove all invalid characters for backwards compatability
validData
```

This is a bit inefficient since it looks for scans the string multiple times, but I think it's a bit easier to read, and I doubt the performance would be affected here very much.

Also note that you don't need a "return" statement, those are not really used in scala. Whatever is the last line in a function is what is returned, so you would just do `validData` as the last line. Also note that validData is already a string, so you don't need to call `mkString`.

mbeckerle: Varun changed it to this flag at my suggestion. So we're debating this functional-programming or not on this PR thread. 

I actually think it is clearer that as you traverse, if you notice the warning is needed you set the flag, rather than the redundant computations which always require me to understand why one is doing it that way. 

I don't feel strongly about "to FP or not to FP" in this particular case.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2360,6 +2354,45 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+        if (!noWarnCharsSet.contains(ch)) doWarning = true
+        false
+      }
+    }
+
+    if (doWarning) {
+        Logger.log.warn("Data contains invalid character(s). Consider using a comment (// or /* */).")
+    }
+
+    return validData
```
stevedlawrence, line 2392: There shouldn't be a `return` keyword here. Just simply `validData` as this line is the scala standard.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2360,6 +2354,45 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+object CanonData {
+  private lazy val doubleForwardPattern = "//.*".r
+  private lazy val openClosePattern = "(?s)/[*].*?[*]/".r
+  private lazy val noWarnCharsSet = "|()[].Xx \n"
+
+  /*
+  * Allow "//" and "/* */" to act as comments.
+  * Any valid XML characters not explicitly allowed are also considered comments and are removed.
+  */
+  def canonicalizeData(validCharactersSet : String, userData : String) : String = {
```
stevedlawrence, line 2366: Scala style is to not have a space before the colons.

stevedlawrence (commented):
Minor style suggestions, but I think tests showing the invariant can feel means we shouldn't have an invariant on \r existing

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
```diff
@@ -490,4 +490,355 @@ class UnitTestTDMLRunner {
...
+                                        0101 /*Data 2&#13;Explanation*/
+              </documentPart>
+              </document>
+    val exc = intercept[Exception] {
+      val doc = new Document(xml, null)
+      doc.documentParts.collect { case x: BitsDocumentPart => x }
+      val dp = doc.documentParts.collect { case x: BitsDocumentPart => x }
+      val firstPart = dp(0).bitDigits
+    }
+    assertTrue(exc.getMessage().contains("Invariant broken"))
```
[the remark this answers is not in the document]
stevedlawrence: I think we assumed that `\r` could never exist in Scala-XML Nodes. In normal XML, all CR should be converted to LF, but I guess that's not the case with Scala Nodes and `&#13;`, so that assumption was wrong.

This assumption was necessary when we used `\n` in a regex, but we've since switched to DOTALL which matches both `\n` and `\r`, so I don't think it really matters anymore (in that case).

However, we do have `\n` in the `noWarn` character list but not `\r`, so we will create a warning if `\r` is used like in this test, which we probably don't want. So I suggest that we remove that invariant and add `\r` to that noWarn list so we don't log a warning if it exists and it will be removed just like other invalid characters like `\n`.

mbeckerle (approved):
+1 All my comments have been addressed.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/UnitTestTDMLRunner.scala
```diff
@@ -490,4 +490,318 @@ class UnitTestTDMLRunner {
...
+              </document>
+    val doc = new Document(xml, null)
+    val dp = doc.documentParts.collect { case x: BitsDocumentPart => x }
+    val firstPart = dp(0)
+    assertEquals("01010101", firstPart.bitDigits)
+  }
+
+  @Test def testCommentBitCarriageReturn(): Unit = {
+    val xml = <document bitOrder="LSBFirst">
+              <documentPart type="bits">0101&#13;00 /*Data 1*/
```
stevedlawrence, line 615: Didn't this cause the invariant to be thrown before? You still have that invariant so I would expect that to still be thrown? I think I'm missing something?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2360,6 +2354,45 @@ sealed abstract class DocumentPart(part: Node, parent: Document) {
...
+  private lazy val openClosePattern = "(?s)/[*].*?[*]/".r
+  private lazy val noWarnCharsSet = "|()[].Xx \n\r"
+
+  /*
+  * Allow "//" and "/* */" to act as comments.
+  * Any valid XML characters not explicitly allowed are also considered comments and are removed.
+  */
+  def canonicalizeData(validCharactersSet: String, userData: String): String = {
+    var doWarning: Boolean = false
+    Assert.invariant(!userData.contains('\r'))      // \r should not exist in userData
```
stevedlawrence, line 2368: Won't this cause an invariant if the data contains `&#13x;`?

stevedlawrence: Yeah, looks like those tests are failing because of this invariant. This line should be removed.

stevedlawrence (approved):
+1 :+1:

Please squash into a single commit with commit message format described in Step 10 of the [Code Contributor Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow) and this should be good to merge.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: Confirming you can put a CR in XML documents using &#13;. It is CR and CRLF as whitespace characters that get normalized into LF. 

So I believe when an Infoset is created from XML text (as by a TDML test that uses an XML infoset) it can contain CR via `&#13;` The Daffodil InfosetInputter does NOT convert these to LF. 

Interestingly, the daffodil XML infoset outputters used by parsing, those create XML and always they convert CRLF and CR to LF. So you cannot get a CR in an XML infoset from the daffodil parser.

PR #801, opened 2022-06-07
Eliminate build warning on VSCode IDE

mbeckerle:
Smallest possible change set. 1 character !

DAFFODIL-2702

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
@@ -130,7 +130,7 @@ final class EntityReplacer {
 
   private def replaceEntityWithChar(input: String, entity: String, newChar: Char): String = {
     val replacement =
-      if (newChar == '%') {
+      if (newChar =#= '%') {
         // Some character entities are not replaced in this EntityReplacer,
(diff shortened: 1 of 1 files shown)


stevedlawrence (approved):
+1

tuxji (approved):
+1

Will you please explain why given that newChar is a Char and '%' is a literal character, perhaps also an integer, you had to change newchar == '%' to newChar =#= '%'?  I had to grep to find out that =#= is defined in daffodil-lib/src/main/scala/org/apache/daffodil/TypedEquality.scala to compare two things that must be convertible, such as numbers, but I still don't see how it helps from looking at its source code.

`@inline def =#=(right: T) = left == right`

mbeckerle:
Changing from "==" to "=#=" just uses one of the typed equality things in that scala file, which makes the VSCode scala stuff not create an IDE "problem item" associated with an unused import. There must be something else in the imported equality stuff being used that VSCode cannot detect (probably an implicit) because otherwise our conventional sbt build would also complain about this unused import. Our sbt build doesn't complain, but VSCode does, so I added a use that can't be avoided/missed. 

The typed equality "=#=" operator is just "==" under the hood, but the Scala type stuff surrounding it prevents you from using it unless the types are convertible.  So you get a compile time warning about doing it wrong if the types are not convertible, rather than untyped == which just gives you false.

PR #805, opened 2022-06-22
Warn when expression cannot be compiled because of no context

stevedlawrence:
In some cases it is possible for an expression to exist without any
compilation context needed for paths, such as when in a global group or
complex type that is never referenced. Daffodil still attempts to
compile all expressions for correctness, but it cannot do so in these
cases.

When this is detected, we provide a warning that compilation is skipped
for this expression. Even though the expression is never used, the
Daffodil codebase still expects to have a CompiledExpression in these
cases. To support this without major changes, this creates a new
RuntimeAbortOp expression recipe, which causes the expression to be
treated as non-constant during compilation but leads to an error if ever
evaluated at runtime.

DAFFODIL-2654


daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
```diff
@@ -311,3 +311,18 @@ abstract class Converter extends RecipeOp {
 trait ToString extends Converter {
   override def computeValue(a: DataValuePrimitive, dstate: DState): DataValueString = a.getAnyRef.toString
 }
+
+
+/**
+ * In some cases, expressions that are never used lack context and so cannot be compiled.
+ * Because they aren't used, it it safe to ignore them, but parts of Daffodil code still
+ * require a CompiledExpression. In these cases, we can use this RuntimeAbort recipe in
```
mbeckerle, line 319: Can you add description of an example of something that requires a compiled expression even though unused?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
```diff
@@ -311,3 +311,18 @@ abstract class Converter extends RecipeOp {
...
+ * the CompiledExpression. This recipe always throws an IllegalStateException, which
+ * treats the expression as non-constant during compilation but causes an error if
+ * evaluated at runtime.
+ */
+case class RuntimeAbortOp(expr: String) extends RecipeOp {
+  override def run(dstate: DState): Unit = {
+    throw new java.lang.IllegalStateException(s"Expression should not have been evaluated during runtime: $expr")
```
mbeckerle, line 326: I would expect this to be an Assert.invariantFailed(...).  

I guess I don't really understand any reason why this should get evaluated that isn't a Daffodil bug.

stevedlawrence: After compilation we attempt to evaluate expressions to determine if they are constant or not. Throwing an `IllegalStateException` is one way that we determines it's not constant. Maybe we should be more explicit about it and do something like this:

```scala
dstate match {
  case _: DStateForConstantFolding => throw new java.lang.IllegalStateException(...)
  case _ => Abort.invariantFailed(...)
}
```

This way when we're evaluating for constant folding we throw a an excetion saying this is isn't a constant, otherwise we assume we're evaluating during a parse/unparse and that creates a more appropriate invariant?

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
stevedlawrence: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/DPathRuntime.scala
stevedlawrence: I've added this change to make it a bit more clear

tuxji (approved):
+1

PR #806, opened 2022-06-29
Fix output of info parser / info unparser in debug

OWL-Varun:
When the user is in debug mode with the subcommand parse, the ouput for
info unparser will now display correctly.
When the user is in debug mode with the subcommand unparse, the output
for info parser will now display correctly.

When the user is in debug mode while calling the parse or unparse
subcommand, we want to make sure that the info printed to the screen for
parser and unparser are accurate. This is done by comparing the
subcommand and the info argument. The "info parser" command prints the
parser information only when the parser subcommand is utilized. The "info
unparser" command prints the unparser information only when the unparse
subcommand is utilitzed.

DAFFODIL-2488

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
@@ -1440,5 +1440,56 @@ class TestCLIdebugger {
     }
   }
 
+  @Test def test_CLI_Debugger_parse_unparser_not_available(): Unit = {
+    val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/charClassEntities.dfdl.xsd")
+    val inputFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/input3.txt")
+    val (testSchemaFile, testInputFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile), Util.cmdConvert(inputFile)) else (schemaFile, inputFile)
+
+    val shell = if (Util.isWindows) Util.start("", envp = DAFFODIL_JAVA_OPTS) else Util.start("")
+
+    try {
+      val cmd = String.format("%s -d parse -s %s %s", Util.binPath, testSchemaFile, testInputFile)
+      shell.sendLine(cmd)
+      shell.expect(contains("(debug)"))
+
+      shell.sendLine("info parser")
+      shell.expect(contains("parser: <Element name='matrix'><DelimiterStackParser>...</DelimiterStackParser></Element>"))
+
+      shell.sendLine("info unparser")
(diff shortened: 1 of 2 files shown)


daffodil-cli/src/it/scala/org/apache/daffodil/debugger/TestCLIDebugger.scala
```diff
@@ -1440,5 +1440,114 @@ class TestCLIdebugger {
...
+
+      shell.sendLine("complete")
+
+      Util.expectExitCode(ExitCode.Success, shell)
+    } finally {
+      shell.close()
+    }
+  }
+
+  @Test def test_CLI_Debugger_parse_unparser_not_available_no_breaks(): Unit = {
```
stevedlawrence, line 1472: I'm not sure a separate test for no breaks is needed. I don't think this provides any real additional correctness. And integration tests are quite slow so we want to minimize unnecessary tests.

In fact, this "no breaks" test is probably the correct way to write the test, with the `expects` immediately after the associated `sendLine`. I would suggest removing the above test and just keeping this one (with `no_breaks` removed).

Same suggestion below for the parser not available tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/debugger/InteractiveDebugger.scala
```diff
@@ -1595,7 +1595,22 @@ class InteractiveDebugger(runner: InteractiveDebuggerRunner, eCompilers: Express
...
+              }
+            }
+            case ustate: UState => {
+              if (name == "unparser") {
+                debugPrintln("%s: %s".format(name, processor.toBriefXML(2)))
+              } else {
+                debugPrintln("parser: not available")
+              }
+            }
+          }
```
stevedlawrence, line 1613: Just a thought, I wonder if there is value in adding a new `info processor` command, which would print out either the parser or unparser depending on which is available (i.e. the current behavior of the InfoParser and InfoUnparser commands)?

And then the "not available" logic would move to the individual parser/unparse commands. E.g.

```scala
class InfoParser ... {
  def act(...) = {
    state match {
      case _: PState => debugPrintln("%s: %s".format(name, processor.toBriefXML(2)))
      case _ => debugPrintln("%s: not available".format(name))
    }
  }
}
```

InfoUnparser would be the same but would match on `UState` instead of `PState`.

We could also deprecate info parser/unparser and just have the info processor command if we wanted. I'm not sure there's much benefit from having both when they do basically the same thing.

If there's consensus that this seems reasonable, we can either add it to this PR, or just open a new feature ticket to implement some other day. I have no strong preference.

mbeckerle: Suggest we integrate this as is, but that next change should be another ticket, but might as well get it done right away while all fresh in mind. 

Why would I suggest this?... because the world seems full of interrupts of context to me right now.

PR #812, opened 2022-07-11
Use Exificient to support EXI format for infosets

jadams-tresys:
This is primarily an example of how to use EXI with Daffodil, although
an additional class was added to support the output of an EXI infoset.
This class may not be necessary with some refactoring and could instead
be replaced by passing in an EXI aware content handler, similar to how
we use an EXI aware XMLReader to support unparsing from an EXI infoset


daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -149,6 +149,14 @@ class Compiler private[sapi] (private var sCompiler: SCompiler) {
     new ProcessorFactory(pf.asInstanceOf[SProcessorFactory])
   }
 
+  @throws(classOf[java.io.IOException])
+  def compileStreamSource(streamSource: org.apache.daffodil.api.InputStreamSchemaSource,
+    optRootName: Option[String] = None,
```
jadams-tresys, line 154: I must have missed this leftover from the previous commits, will remove.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -669,7 +674,8 @@ object Main {
       case InfosetType.JSON => Left(new JsonInfosetOutputter(os, pretty = true))
       case InfosetType.JDOM => Left(new JDOMInfosetOutputter())
       case InfosetType.W3CDOM => Left(new W3CDOMInfosetOutputter())
-      case InfosetType.SAX => Right(new DaffodilParseOutputStreamContentHandler(os, pretty = true))
+      case InfosetType.SAX => Right(new DaffodilParseOutputStreamXMLContentHandler(os, pretty = true))
+      case InfosetType.EXI => Right(new DaffodilParseOutputStreamEXIContentHandler(os))
```
mbeckerle, line 678: Per comments below, right here (or in a private method of Main called from here) is where the code you have in the DaffodilParseOutputStreamEXIContentHandler constructor would go. That would allow you to remove the dependency on EXI from the runtime library and put it only on the CLI library.

tuxji (commented):
I get the impression from the comments so far that this PR needs further changes (and SAX support refactored a bit) first.  I'll hold off reviewing until these changes go in, but I'll point out that some new lines in DaffodilParseOutputStreamContentHandler.scala aren't covered by tests.

jadams-tresys:
I've made some cleanup changes for removing SAX from the CLI tool, but I didn't address the following from Mike B:

Ex: there should be no code in DataProcessor associated with particular SAX handlers, and no base classes or traits one must derive from in order to use our SAX API. I think the requirements were not clear, and SAX is for some reason fairly confusing. But there are things we need only for unit testing that are in the primary libraries.

Ex: Class DaffodilParseXMLReader and Class DaffodilUnparseContentHandler are defined as if they are part of our API, when they should just be under src/test.

DaffodilParseXMLReader and DaffodilUnparseContentHandler to me seem like they might be necessary as I believe they are responsible for translating the various SAX events into something that Daffodil understands.  I feel like these, or something similar, are necessary in order for someone to use Daffodil with SAX.  I feel like it shouldn't be required of someone using the API to have to define something for Daffodil to be able understand SAX events.

I do think the API could be smoothed out in this regard, where the DaffodilParseXMLReader and DaffodilUnparseContentHandler aren't really exposed by the API.  For example, when parsing we are currently happy to accept a users ContentHandler, but then the user needs to call DFDL.DataProcessor.newXMLReaderInstance to get the current processors XMLReader, and then set that XMLReader's ContentHandler to the one they are trying to pass into Daffodil.  To me this feels pretty clunky and could be hidden, but I haven't looked at this long enough to come up with a cleaner solution.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -662,14 +666,20 @@ object Main {
...
+  : Either[InfosetOutputter, ContentHandler] = {
     val outputter = infosetType match {
       case InfosetType.XML => Left(new XMLTextInfosetOutputter(os, pretty = true))
       case InfosetType.SCALA_XML => Left(new ScalaXMLInfosetOutputter())
       case InfosetType.JSON => Left(new JsonInfosetOutputter(os, pretty = true))
       case InfosetType.JDOM => Left(new JDOMInfosetOutputter())
       case InfosetType.W3CDOM => Left(new W3CDOMInfosetOutputter())
-      case InfosetType.SAX => Right(new DaffodilParseOutputStreamContentHandler(os, pretty = true))
+      case InfosetType.EXI => {
+        val exiFactory = DefaultEXIFactory.newInstance()
```
mbeckerle, line 677: So this is schema-independent EXI for now correct?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -662,14 +666,20 @@ object Main {
...
+        exiFactory.getFidelityOptions.setFidelity(FidelityOptions.FEATURE_PREFIX, true)
+        val exiResult = new EXIResult()
```
mbeckerle, line 679: Is this also where we would turn on other options for the EXI encoder, like telling it to compress the data to minimize its size?

mbeckerle (approved):
+1
Minor questions. No real change suggestions. I see how the DaffodilParseOutputStreamContentHandler doesn't end up being part of the API. 

I think we'll want to create a standalone application that uses Daffodil and EXI that will serve as a test rig to both try other EXI implementations, try schema-aware EXI, and to test performance.

mbeckerle:
Binary data test files need to be added to the rat-check exclusions.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1400,8 +1408,15 @@ object Main {
...
-    val xmlReader = DaffodilSAXParserFactory().newSAXParser.getXMLReader
+    contentHandler: DFDL.DaffodilUnparseContentHandler,
+    infosetType: InfosetType.Type): UnparseResult = {
+    val xmlReader = infosetType match {
+      case InfosetType.EXI => {
+        val exiSource = new EXISource()
+        exiSource.getXMLReader
+      }
+      case _ => DaffodilSAXParserFactory().newSAXParser.getXMLReader
+    }
```
stevedlawrence, line 1419: I imagined that we would have a new `InfosetInputter` and `InfosetOutputter` specific to EXI (i.e. `EXIInfosetInputter`/`EXIInfosetOutputter`). That way anyone would be able to use this by just specifying the new infosetter without having to do all this SAX logic themselves. Is that something we're trying to avoid? That would move all of this logic out of the CLI and would make it easily usable in the API. It could stil use the SAX InfosetOutputter/Inputter under the hood, just makes it easier to use.

Another thought, this is just using our SAXInfosetOutputter and plugging that into this API--I wonder if we would get better performance if we directly use the EXI API? Converting our internal infoset representation to SAX events has some amount of overhead ([DAFFODIL-2400](https://issues.apache.org/jira/browse/DAFFODIL-2400)) that might be avoided if we call the appropriate API calls. Just skimming the code, it looks like EXIBodyEncoder is maybe one of the more important API classes and is pretty straightforward. It should map pretty closely to our InfosetOuputter API:

https://github.com/EXIficient/exificient-core/blob/master/src/main/java/com/siemens/ct/exi/core/EXIBodyEncoder.java

This also seems especially important in the case of unparsing. Daffodil SAX unparsing is extra complicated because Daffodil InfosetInputters are pull-based based, but SAX is push-based, so we have this mess of coroutines to link to two and it creates a lot of overhead. If we could directly use the EXI API, or at least a StAX based API for an EXIInfosetOutputter, I imagine performance would be significantly better.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -873,8 +883,6 @@ object Main {
 
               val parseResult = eitherOutputterOrHandler match {
                 case Right(saxContentHandler) =>
-                  // reset in case we are streaming
-                  saxContentHandler.reset()
```
stevedlawrence, line 877: Why is this reset removed? Seems like this is important?

build.sbt
```diff
@@ -57,6 +57,7 @@ lazy val io               = Project("daffodil-io", file("daffodil-io")).configs(
 lazy val runtime1         = Project("daffodil-runtime1", file("daffodil-runtime1")).configs(IntegrationTest)
                               .dependsOn(io, lib % "test->test", udf, macroLib % "compile-internal, test-internal")
                               .settings(commonSettings, usesMacros)
+                              .settings(libraryDependencies ++= Dependencies.exi)
```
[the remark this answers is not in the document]
stevedlawrence: I wonder if there is value is splitting all of our infoset inputters/outputters into separate subprojects with separate dependencies? That way our runtime doesn't depend on a bunch of different libraries, many of which wouldn't be used.

This means API users would need to add a new dependency depending on which infoset they want to use (e.g. daffodil-runtime1-infoset-xml, daffodil-runtime-infoset-json, daffodil-runtime1-infoset-sax), but it makes runtime and it's dependencies smaller.

We could still have the CLI depend on all or a subset of these for which ever infoset inputters we want to make available.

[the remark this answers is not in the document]
mbeckerle: If we do this, please use new distinct packages for each such component, and those packages should NOT span module boundaries. This is a requirement for Java Modules/OGSI packaging that our existing modules and packages are not adhering to, and we need to fix (separate JIRA ticket), but,... no point in making it worse.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle: Init protocol was converted over putting it on startDocument/endDocument which is where SAX APIs want you to do it, so that you don't have to extend SAX handler classes with additional methods.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: Yes, the options I believe go onto the EXIFactory.

build.sbt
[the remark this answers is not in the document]
jadams-tresys: I'm not sure that I like the idea of moving those 3 infoset inputters/outputter into a separate library.  I think Daffodil should have some subset of infoset handlers built in (the most common, which I would assume to be these 3) and then perhaps the other less common infoset implementations could be set up as separate libraries?  I know it's not exactly the most consistent design, but it would feel strange to me to not include such basic functionality in the main Daffodil library.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: That is correct.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: I'm open to ideas on this one.  @mbeckerle seemed more interested in relying on the SAX handlers.  It's probably worth it for me to do some performance testing to see how performance is with this current SAX/EXI code.  If we aren't seeing any significant performance improvements, perhaps it would be worth it to use EXI directly in a more traditional Infoset inputter/outputter.

build.sbt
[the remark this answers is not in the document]
stevedlawrence: Agreed. We could maybe take advantage of optional dependencies for this. That way runtime1 would depend on all of infoset subprojects, but a subset of the less common ones would be marked as optional--any API users that want to use an optional one would need to explicitly add them. This would make it trivial to change the "core" set of infoset inputters/outputters by just marking some as required vs optional.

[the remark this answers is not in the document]
stevedlawrence: Also, I'm not suggesting this be done for this PR. This is a separate issue that requires a bit more thought. This PR just sparked this idea of a potential improvement to me.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -421,7 +425,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
 
     val config = opt[File](short = 'c', argName = "file", descr = "XML file containing configuration items")
     val vars = props[String](name = 'D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing. Can be prefixed with {namespace}.")
-    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output or unparse. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', or 'null'. Defaults to 'xml'.", default = Some(InfosetType.XML))
+    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output or unparse. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', 'exi', or 'null'. Defaults to 'xml'.", default = Some(InfosetType.XML))
```
stevedlawrence, line 428: I don't think we should remove SAX from the performance command. We know there are issues with SAX performance and this is really the only way we currently have to test that.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1400,8 +1408,15 @@ object Main {
...
-    contentHandler: DFDL.DaffodilUnparseContentHandler): UnparseResult = {
-    val xmlReader = DaffodilSAXParserFactory().newSAXParser.getXMLReader
+    contentHandler: DFDL.DaffodilUnparseContentHandler,
+    infosetType: InfosetType.Type): UnparseResult = {
+    val xmlReader = infosetType match {
+      case InfosetType.EXI => {
+        val exiSource = new EXISource()
+        exiSource.getXMLReader
+      }
+      case _ => DaffodilSAXParserFactory().newSAXParser.getXMLReader
```
stevedlawrence, line 1418: What was the reason for removing SAX support aside from it's a little unintuitive since it's an API, very similar to JDOM, Scala-XML, W3CDOM, etc, that we do still support? It looks like the way you've made the changes SAX should all still work. In fact, right here we still have support for normal SAX.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: In addition to potential performance improvements, one benefit of creating a custom EXIInfosetInputter/Outputter is that is becomes part of the Daffodil API, and anyone using Daffodil can easily using it by just choosing the right class. With this change, EXI support is only added to the CLI--anyone wanting to using EXI must use code similar to what is in the CLI and do it themselves.

And we don't even really need a new release for that (except for maybe the reset() change?). Can we already claim we support EXI via our SAX API?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence, line 877: Should we error or warning if streaming is enabled? The result of streamed EXI seems useless since we throwing NUL bytes as separators. And it's going to be impossible to unparse stream since EXI data likely has NUL bytes in it that we use for a separator In fact, we use a java.util.Scanner for finding the NUL delimiter, which is text based. So even if there were no NUL bytes in the input aside from our delimiter, I'm imagine things still wouldn't work. We'll need a completely different mechanism for supporting streaming EXI.

stevedlawrence (approved):
+1

This seems reasonable for initial support. But it's not clear to me why we are removing SAX support, and it feels like there are some potential benefits to creating custom EXIInfosetInputter/Outputter.

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -1278,18 +1278,18 @@ class TestCLIparsing {
...
+  @Test def test_XXX_CLI_Parsing_SimpleParse_exi(): Unit = {
 
     val schemaFile = Util.daffodilPath("daffodil-test/src/test/resources/org/apache/daffodil/section00/general/generalSchema.dfdl.xsd")
     val (testSchemaFile) = if (Util.isWindows) (Util.cmdConvert(schemaFile)) else (schemaFile)
 
     val shell = Util.start("")
 
     try {
-      val cmd = String.format(Util.echoN("Hello") + "| %s parse -I sax -s %s -r e1", Util.binPath, testSchemaFile)
+      val cmd = String.format(Util.echoN("Hello") + "| %s parse -I exi -s %s -r e1 | md5sum", Util.binPath, testSchemaFile)
```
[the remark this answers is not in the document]
stevedlawrence: I wonder if these will fail CI once we get passed the rat check?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
mbeckerle-yahoo: We should error here until we know what we're doing with trying to use EXI for streaming. I have no idea if EXI files concatenated can be processed in any sensible way, are self-length aware, etc. XML is not sensible this way. It really wants to read until "end of data" before issuing an endDocument() SAX call/event. Some experiments are really needed here to see whether a sensible stream of SAX events can come from concatenated EXI data, or not. We may find that EXI streaming requires trickery like putting a length header on each chunk.

mbeckerle-yahoo: (Sorry that comment from "mbeckerle-yahoo" which is a test account. I will login now with my regular userid. )

mbeckerle: Ok. Back signed in with proper id. 
So my comment is "yeah,.... what that yahoo up there said"

stevedlawrence: Skimming through the EXI spec, it says:

> An EXI body can represent an EXI document with a single root element or an EXI fragment with zero or more root elements

Maybe when when streaming we use the "fragment" mode and each "infoset" is treated as a fragment. Without reading the spec more throughly though, it's not clear to me exactly what a fragment represents is and if it is streamable. But might be another option and something to look into if concatenated EXI's doesn't work.

Note that I don't think we need to add this support in this PR. An error as Mike suggests seams reasonable for now.

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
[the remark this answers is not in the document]
jadams-tresys: Yes, this would fail.  I have simply disabled this test on windows as there isn't a straightforward way to get a hash of the output of a command and I'm hesitant to deal with temporary files and cleaning them up without some sort of more secure API beyond just redirecting output to a file.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: An error message has been added if streaming mode is set with EXI infosets

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -879,6 +882,12 @@ object Main {
             var keepParsing = true
             var exitCode = ExitCode.Success
 
+            if (parseOpts.stream.toOption.getOrElse(false) && (infosetType == InfosetType.EXI)) {
+              exitCode = ExitCode.NotYetImplemented
+              keepParsing = false
+              System.err.println("Streaming mode with EXI infosets is not currently supported")
+            }
```
stevedlawrence, line 889: This only errors when parsing. Streaming can also be enabled for unparsing so we should detect that too.

Also, maybe a cleaner way to check the options is using scallops `validateOpt` function, I think something like this:

```scala
validateOpt(infosetType, stream) {
  case (Some(InfosetType.EXI), Some(true)) => Left("Error message here ... ")
  case _ => Right(Unit)
}
```

We use something similar a few places in the scallop configuration.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
+  Licensed to the Apache Software Foundation (ASF) under one or more
```
stevedlawrence, line 2: Accidentally remove the *

PR #815, opened 2022-07-19
Update yarn build command to be yarn package

shanedell:
This is in correlation with https://github.com/apache/daffodil-vscode/pull/232

### DIFF
--- a/containers/release-candidate/daffodil-release-candidate
+++ b/containers/release-candidate/daffodil-release-candidate
@@ -297,8 +297,7 @@ case $PROJECT_REPO in
 
     echo "Building Convenience Binaries..."
     mkdir -p $DAFFODIL_RELEASE_DIR/bin/
-    yarn install
-    yarn build
+    yarn package
     cp *.vsix $DAFFODIL_RELEASE_DIR/bin/
 
     ;;


containers/release-candidate/daffodil-release-candidate
```diff
@@ -298,7 +298,7 @@ case $PROJECT_REPO in
     echo "Building Convenience Binaries..."
     mkdir -p $DAFFODIL_RELEASE_DIR/bin/
     yarn install
-    yarn build
+    yarn package
```
stevedlawrence, line 301: Looks like we can remove `yarn install`? Doesn't seem to be needed anymore. I think that was only needed for the os thing which has since been removed/replaced.

stevedlawrence (approved):
+1

containers/release-candidate/daffodil-release-candidate
Shanedell: Oh yeah that is true because the packaging script does run it

Shanedell:
@stevedlawrence Actually this will still need the yarn install for it to be able to compile the code and run the webpack command I will add it back in

stevedlawrence:
> @stevedlawrence Actually this will still need the yarn install for it to be able to compile the code and run the webpack command I will add it back in

The current main branch of the vscode repo works fine for me:
```
git clean -xdf && yarn package
```
I don't get any webpack errors, and it looks like the vsix file was built successfully.

Shanedell:
You are right I actually run in prepackage so it can be removed will do that one oversight on my part

arosien (approved):
👍

PR #827, opened 2022-08-16
Set starting absolute bit positions of DOSs more often, avoid deadlocks

stevedlawrence:
Currently we are very pessimistic about specifying a known length of a
simple type suspension by always returning Nope (i.e. there is no known
length). This meant that it was impossible to determine the absolute
starting position of split buffers from those suspensions. And because
absolute bit positions are so important for allowing suspensions to
evaluate (e.g. alignment, length calculations) it increased the chance
that a suspension would deadlock.

To resolve this issue, this modifies the simple type suspension to use
the actual length where possible. This isn't always possible, like when
padding or right fill might be needed, but that often isn't the case.

Even with that change, the absolute starting positions could not be
calculated for a new split in cases where the current DOS did not have
an absolute position and only had a relative position. This meant
deadlocks were still fairly likely. So in these cases, we calculate and
store the length of the DOS (relative bit position + suspension length).
Once the absolute starting position of the DOS it determined, we can use
that length to set the starting absolute position of the following DOS.

Also add an assert to ensure that these new length calculations are
correct by checking that the ending bit position of one DOS matches the
starting bit position of the next DOS when merging. This discovered unit
tests with an off by one error.

These changes increase the chance of an absolute starting bit position
being known and allows more suspensions to be evaluated which should
lead to less deadlocks.

Unrelated, remove println from a previous commit that was used for
debugging and accidentally made it into the commit.

DAFFODIL-2717

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
@@ -149,7 +149,7 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
    * of the former starting bit pos so as to be able to
    * convert relative positions to absolute positions correctly.
    */
-  protected final def maybeAbsStartingBitPos0b = {
+  final def maybeAbsStartingBitPos0b: MaybeULong = {
     if (this.maybeAbsolutizedRelativeStartingBitPosInBits_.isDefined)
       maybeAbsolutizedRelativeStartingBitPosInBits_
     else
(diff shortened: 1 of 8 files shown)


mbeckerle (approved):
+1 

So the only thing missing is a test that reproduces the original issue, but that can be part of the Daffodil code base. Did you look into creating such? 

The deadlock is reproducible using a public DFDL schema, so that's at least available, but it would be great if we had a test in the daffodil test suite that deadlocks without this fix, and doesn't deadlock with it.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -149,7 +149,7 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
    * of the former starting bit pos so as to be able to
    * convert relative positions to absolute positions correctly.
    */
-  protected final def maybeAbsStartingBitPos0b = {
+  final def maybeAbsStartingBitPos0b = {
```
tuxji, line 152: Now that this method is public, it should have an explicit return type annotation per Scala's style [guide](https://docs.scala-lang.org/style/types.html).

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -185,6 +185,20 @@ trait DataOutputStreamImplMixin extends DataStreamCommonState
       Assert.usageError("You cannot set the abs starting bit pos again.")
     }
     checkInvariants()
+
+    // We now know the absolute starting position of this DOS. If we also know its
+    // length, then we can set the absolute starting position of the next DOS as well
+    // (if one exists and if it doesn't already have an absolution starting
```
tuxji, line 191: absolution  -> absolute

tuxji (approved):
+1

stevedlawrence:
> So the only thing missing is a test that reproduces the original issue, but that can be part of the Daffodil code base. Did you look into creating such?

I wasn't able to create a minimal example that reproduced the issue. Suspensions are tricky. But I did confirm that mil-std-2045 schema issue is fixed and that nothing broke in my regression suite.

PR #828, opened 2022-08-17
Allow DataProcessor's copy function to be public.

mbeckerle:
I need this function for various test rigs I am creating for large schemas where copying the DP saves 10 seconds a test for reloading another one just because I can't copy the object.

DAFFODIL-2723


tuxji (approved):
+1

Agreed.

stevedlawrence (approved):
+1

Seems reasonable to me, but the `withDebugger()`/`withDebugging()` functions already do a copy. The `setDebugger()`/`setDebugging()` functions do mutate the `DataProcessor`, but they are deprecated and probably shouldn't be used.

mbeckerle:
There's some wierdness. E.g., withDebugger's argument is dbg: AnyRef which it then downcasts to Debugger(?) No comment as to rationale (probably I wrote this.... grrrr). 

Making copy public is just a quick painless work around.

stevedlawrence:
Is there a bug with the AnyRef used in `withDebugger`? Should we open a ticket?

mbeckerle:
Created https://issues.apache.org/jira/browse/DAFFODIL-2724

PR #829, opened 2022-08-17
All InfosetOutputter functions now return Unit

stevedlawrence:
Commit 3b213ce30b deprecated InfosetOutputter functions that returned
false, and instead expected them to throw an Exception. If an
implementation returned false from one of these functions, then Daffodil
threw a usageError and stopped processing, with the assumption that no
one actually used this.

Unfortunately, the SAXInfosetOutputter actually does return false to
signify errors, which could lead to usageErrors when outputting SAX (or
EXI which uses SAX behind the scenes).

This changes the InfosetOutputter API to remove the Boolean return
value, and changes all concrete implementations to use this new API and
throw an Exception when an error is encountered. This breaks backwards
compatibility, but ensures consistence and a cleaner API.

Note that this does not completely address DAFFODIL-2721, but is a quick
change to fix the SAX and EXI regression. Additional work is needed to
fully resolve the issue.

DAFFODIL-2721

### DIFF
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
@@ -123,18 +123,18 @@ abstract class InfosetOutputter extends SInfosetOutputter {
   /**
    * Called by Daffodil internals to signify the beginning of the infoset.
    *
-   * @return true on sucess, false if there was an error and Daffodil should stop all
-   *         future calls to the InfosetOutputter
+   * @throws Exception if there was an error and Daffodil should stop parsing
    */
-  def startDocument(): Boolean
+  @throws[Exception]
+  def startDocument(): Unit
 
   /**
    * Called by Daffodil internals to signify the end of the infoset.
    *
-   * @return true on sucess, false if there was an error and Daffodil should stop all
-   *         future calls to the InfosetOutputter
+   * @throws Exception if there was an error and Daffodil should stop parsing
    */
-  def endDocument(): Boolean
+  @throws[Exception]
+  def endDocument(): Unit
 
   /**
    * Called by Daffodil internals to signify the beginning of a simple element.
@@ -142,11 +142,11 @@ abstract class InfosetOutputter extends SInfosetOutputter {
    * @param diSimple the simple element that is started. Various fields of
    *                 DISimple can be accessed to determine things like the
    *                 value, nil, name, namespace, etc.
-   * @return true on sucess, false if there was an error and Daffodil should stop all
-   *         future calls to the InfosetOutputter
+   * @throws Exception if there was an error and Daffodil should stop parsing
    */
 
-  def startSimple(diSimple: DISimple): Boolean
+  @throws[Exception]
+  def startSimple(diSimple: DISimple): Unit
 
   /**
    * Called by Daffodil internals to signify the end of a simple element.
@@ -154,10 +154,10 @@ abstract class InfosetOutputter extends SInfosetOutputter {
    * @param diSimple the simple element that is ended. Various fields of
    *                 DISimple can be accessed to determine things like the
    *                 value, nil, name, namespace, etc.
-   * @return true on sucess, false if there was an error and Daffodil should stop all
-   *         future calls to the InfosetOutputter
+   * @throws Exception if there was an error and Daffodil should stop parsing
    */
-  def endSimple(diSimple: DISimple): Boolean
+  @throws[Exception]
+  def endSimple(diSimple: DISimple): Unit
 
   /**
    * Called by Daffodil internals to signify the beginning of a complex element.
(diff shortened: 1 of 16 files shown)


daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
```diff
@@ -123,85 +123,85 @@ abstract class InfosetOutputter extends SInfosetOutputter {
   /**
    * Called by Daffodil internals to signify the beginning of the infoset.
    *
-   * @return true on sucess, false if there was an error and Daffodil should stop all
-   *         future calls to the InfosetOutputter
+   * @throws Exception if there was an error and Daffodil should stop parsing
    */
-  def startDocument(): Boolean
+  @throws[Exception]
```
tuxji, line 128: I'm not sure why these javadoc errors are happening (java.lang.Exception should be built-in), but DAFFODIL-2721 says we may want to create specific exception(s) that both the inputter and outputter should be expected to throw when errors happen.  I think if you create such exception(s) (their names could be InfosetException or InfosetInputterException/InfosetOutputterException, take your choice), the javadoc errors would go away.  I also wonder if the specific exception should be a checked exception or inherit from RuntimeException (less code would be needed in the latter case).

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/TeeInfosetOutputter.scala
```diff
@@ -19,11 +19,9 @@ package org.apache.daffodil.infoset
 
 
 /**
- * Receive infoset events and forward them to one or more InfosetOutputters.
- * For infoset events that return a boolean, this returns true only if all
- * outputters return true, otherwise false is returned. Additionally, all
- * events are called on all the outputters regardless of the return of any
- * previous outputters.
+ * Receive infoset events and forward them to one or more InfosetOutputters. A
+ * thrown exception from one outputters is not caught and bubbles up, and no
```
tuxji, line 23: "one outputters" might read better as "any outputter"

tuxji (changes requested):
I think Exception might be too general and it might be better to throw a specific exception (that also might make the CI pass).

daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
stevedlawrence: Yeah, I can't figure it out. Seems to be a known issue with scaladoc--I've found lot's of other people with similar issues and no solution. Though it's not clear to me why it works for some exceptions types but not others. For now, I've just removed the `@throws` in the comments.

This will probably go away once we add a custom exception. My initial goal here is to just fix the sax/exi regressions as quickly as possible since I know of active development/testing around that.

tuxji (approved):
+1 now that CI checks pass

daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
mbeckerle, line 128: Very broad throw class. You are basically telling people "don't catch this". 

Why not InfosetInputterException or InfosetOutputterException which encapsulate any mystery throw.

mbeckerle (changes requested):
throws Exception as signature feels problematic.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala
stevedlawrence: > Very broad throw class. You are basically telling people "don't catch this".

I feel like it's really saying "if you have any error at all, just throw an Exception, we don't really care what it is. We'll turn it into an SDE". Maybe that's not a good API decision though.

Switching to custom exceptions does add extra complexity. InfosetOutputter implementations are going to do something like this:

```
def startSimple(...): Unit = {
  try {
    // do someting
  } catch {
    catch e: SomeExpectedException => throw new InfosetOutputterException(e)
  }
}
```

And then our `InfosetWalker` code is going to look like this:

```scala
def doOutputter(outputterFunc: => Unit, ...) {
  try {
    outputterFunc
  } catch {
    case e: InfosetOutputterException => context.SDE("infoset outputter failed: " + e.getCause.toString)
    case e: Exception => Assert.usage("Should only throw InfosetOutputterException")
  }
}
```

So we catch the expected exception, unwrap it, and turn it into an SDE. And any other exception becomes an usage error.

This makes for a cleaner API and detects uncaught exceptions, but I imagine most implementations are just going to wrap all their logic in giant try/catch blocks (needed for each of the `InfosetOutputter` functions), wrap that exception in our custom exception, and then we'll immediately unwrap it and SDE.

This is probably the right approach, but it does adds a lot of extra boilerplate to InfosetOutputter impelmentations.

It's made even more difficult in Scala since it doesn't warn about uncaught exceptions. So if our we're missing any try/catches in our built-in InfosetOutputters, we aren't going to know until a usage error pops up. Maybe that's what we want though?

mbeckerle: I think Scala 3 adds throws declarations to Scala. It was considered a bug, not a feature, that they were missing. 

I'm ok with just documenting why it is ok for any kind of exception to be thrown.

mbeckerle (approved):
+1. Just include comments about the rationale for why any Exception being thrown is ok.

PR #830, opened 2022-08-17
Add dependency scanning for the build.sbt 

jhutchings1:
:wave: Hello! I'm a PM at GitHub working on our supply chain security features. We've been testing out a new capability called the [dependency submission API](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api) that lets repositories that use build systems like Gradle or SBT upload information about the libraries they depend on so that @dependabot can send alerts when there are known vulnerabilities. Are you interested in trying this out?

### DIFF
--- a/.github/workflows/dependency-scan.yml
+++ b/.github/workflows/dependency-scan.yml
@@ -1,0 +1,42 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Dependency scan
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  push:
+    branches: [ "main" ]
+    
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3.0.2
+
+      - name: Sbt Dependency Submission
+        # You may pin to the exact commit or the version.
+        # uses: scalacenter/sbt-dependency-submission@ab086b50c947c9774b70f39fc7f6e20ca2706c91
+        uses: scalacenter/sbt-dependency-submission@v2.0.1


.github/workflows/dependency-scan.yml
```diff
@@ -0,0 +1,29 @@
+name: Dependency scan
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
stevedlawrence, line 9: Do we want this to run on pull requests? Seems like we would only want it to run when things are actually pushed to main? Someone adding a new dependency to a pull request shouldn't change the GitHub dependency graph.

stevedlawrence (commented):
Seems interesting, though it seems like the only benefit we would really get is having an accurate [dependency graph](https://github.com/apache/daffodil/network/dependencies).

Unfortunately it seems like Dependabot still doesn't support sbt dependencies, but I guess this would allow it to create alerts? I'm not sure where those alerts would go or how helpful they are, especially since we have Scala Steward to create dependency PRs.

Might be worth adding and if the the alerts are annoying (we already get a lot of dependency notifications) we can always disable it?

jhutchings1:
> Unfortunately it seems like Dependabot still doesn't support sbt dependencies, but I guess this would allow it to create alerts? I'm not sure where those alerts would go or how helpful they are, especially since we have Scala Steward to create dependency PRs.

Correct. Since Scala projects usually use dependencies from Maven Central, we've got a database of [advisories](https://github.com/advisories) that can drive alerts. If you'd like to see a preview of the dependency graph that this calculates, you can check out the one in my [fork](https://github.com/jhutchings1/daffodil/network/dependencies).

mbeckerle:
A thought is that the dependency graph does not distinguish between main dependencies and test dependencies. Or I didn't see that at first glance. 

That distinction is quite important. 

Main dependencies are passed through to consumers who use daffodil. 

Test dependencies are encountered only by developers.

jhutchings1:
> A thought is that the dependency graph does not distinguish between main dependencies and test dependencies. Or I didn't see that at first glance.

The dependency graph supports indicating whether a dependency is runtime or development, and that makes its way through to alerts. I can't say for sure whether the SBT action differentiates. 
https://github.blog/changelog/2022-06-23-dependabot-alerts-filter-alerts-by-the-scope-of-the-dependency-runtime-and-development/

mbeckerle:
> > A thought is that the dependency graph does not distinguish between main dependencies and test dependencies. Or I didn't see that at first glance.
> 
> The dependency graph supports indicating whether a dependency is runtime or development, and that makes its way through to alerts. I can't say for sure whether the SBT action differentiates. https://github.blog/changelog/2022-06-23-dependabot-alerts-filter-alerts-by-the-scope-of-the-dependency-runtime-and-development/

Super. You are way ahead of me.

stevedlawrence (approved):
+1 seems cool to me

@jhutchings1, looking at the dependency graph of your fork I noticed two minor issues, I guess they are just bugs in the plugin?

1. Daffodil creates many jars/subprojects out of this repo, they are all listed separately in the dependency graph (e.g. org.apache.daffodil:daffodil-core_2.12). Because they are part of this repo, I would expect that they don't show up in this dependency graph. Is there a setting or something we need to change? Or maybe this is a bug in the sbt-dependency-submission that it should ignore subprojects.
2. Some dependencies don't have links to their repositories, including the Daffodil jars. Is seems mostly scala projects, so maybe this is also a bug in the sbt-dependency-submission thing? Or maybe there is maybe some configuration that Scala projects tend to not set that we should be setting?

tuxji (approved):
+1 once suggested changes are made

After making the suggested changes, please squash your commits to one commit (our policy is each pull request should have only one commit to be merged).  Thanks for your welcome addition!

tuxji:
Oops - got this from CI:

```
scalacenter/sbt-dependency-submission@v2.0.1 is not allowed to be used in apache/daffodil. Actions in this workflow must be: within a repository owned by apache, created by GitHub, verified in the GitHub Marketplace, or matching the following: */*@[a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9][a-f0-9]+, AdoptOpenJDK/install-jdk@*, JamesIves/github-pages-deploy-action@5dc1d5a192aeb5ab5b7d5a77b7d36aea4a7f5c92, TobKed/label-when-approved-action@*, actions-cool/issues-helper@*, actions-rs/*, al-cheb/configure-pagefile-action@*, amannn/action-semantic-pull-request@*, apache/*, burrunan/gradle-cache-action@*, bytedeco/javacpp-presets/.github/actions/*, chromaui/action@*, codecov/codecov-action@*, conda-incubator/setup-miniconda@*, container-tools/kind-action@*, container-tools/microshift-action@*, dawidd6/action-download-artifact@*, delaguardo/setup-graalvm@*, docker://jekyll/jekyll:*, docker://pandoc/core:2.9, eps1lon/actions-label-merge-conflict@*, gaurav-nelson/github-action-markdown...
```

I filed an Infra issue [INFRA-23593](https://issues.apache.org/jira/browse/INFRA-23593) to have this action whitelisted.

stevedlawrence:
Looks like a `revert-830-main` branch was created to revert this? I don't *think* we need to revert the commit. Once INFRA allows this we should be able to rerun the workflow for that commit and it should pass. So the new branch should be safe to delete, I think?

tuxji:
Yep, I just deleted the branch.  I had clicked a Revert button to see what would happen and had changed my mind when I saw it would take another commit to revert the previous commit.  Thanks for alerting me that new branch was still around.

PR #832, opened 2022-08-23
Daffodil 2720 wip create new char class entities

OWL-Varun:


tuxji (changes requested):
Before we can review your pull request, please add tests for each new DFDL entity.  The coverage report tells me that your new lines of code need to be covered by automated tests.  You may have run manual tests but such testing is not reproducible or maintainable.

mbeckerle (changes requested):
So far so good. We need tests now to exercise each of these new char/char-class entities, which differentiate them from existing WSP family of char-class entities. 

The quintessential example of this is CSV, whiere the separators are not only "," but are
```LSP*;,%LSP+```
So that any within-a line space before the comma is part of the delimiteer, and any (but 1) space after is also part of the delimiter. This will exclude the line-ending chars, so will behave better than ```%WSP*;,%WSP+;``` when lines are too short. 

Then test cases without enough commas in a row should NOT be confused by ```%NL;``` line endings in the middle of entries, as those won't be allowed as part of the separator.

File: daffodil-core/src/test/scala/org/apache/daffodil/general/TestPrimitives.scala [hunk 1]
```diff
@@ -137,6 +137,131 @@ class TestPrimitives {
...
+    val (_, actual) = TestUtils.testString(sch, "SingleSpace, MultSpace,   SingleTab,\u0009MultTab,\u0009\u0009Mix,\u0009\u0009 \u0009     \u0009End")
+    val expected: Node = <e1><s1>SingleSpace</s1><s2>MultSpace</s2><s3>SingleTab</s3><s4>MultTab</s4><s5>Mix</s5><s6>End</s6></e1>
+    TestUtils.assertEqualsXMLElements(expected, actual)
+  }
+
+  @Test def testLSPStarDelimiter(): Unit = {
+    val sch = SchemaUtils.dfdlTestSchema(
+      <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>,
+
+      <dfdl:format ref="tns:GeneralFormat" representation="text" lengthUnits="bytes" encoding="US-ASCII" initiator="" separator="" terminator="" ignoreCase="no"/>,
```
mbeckerle, line 186: Cannot review as this line is too long. This is true throughout this file that these dfdl:format elements are too long. 
The github review tool really gives me no way to scroll to the right here.

daffodil-core/src/test/scala/org/apache/daffodil/general/TestPrimitives.scala
```diff
@@ -137,6 +137,131 @@ class TestPrimitives {
...
+
+      <xs:element name="e1" dfdl:lengthKind="delimited">
+        <xs:complexType>
+          <xs:sequence dfdl:separator=",%SP;" dfdl:separatorPosition="infix">
+            <xs:element name="s1" type="xs:string" dfdl:lengthKind="delimited"/>
+            <xs:element name="s2" type="xs:string" dfdl:lengthKind="delimited"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+    val (_, actual) = TestUtils.testString(sch, "SingleSpace, End")
```
mbeckerle, line 220: You need a test showing that if there are two spaces after the comma, that the s2 element contains the 2nd space. `<s2> End</s2>`

daffodil-core/src/test/scala/org/apache/daffodil/general/TestPrimitives.scala
```diff
@@ -137,6 +137,131 @@ class TestPrimitives {
...
+            <xs:element name="s1" type="xs:string" dfdl:lengthKind="delimited"/>
+            <xs:element name="s2" type="xs:string" dfdl:lengthKind="delimited"/>
+            <xs:element name="s3" type="xs:string" dfdl:lengthKind="delimited"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+    val (_, actual) = TestUtils.testString(sch, "SingleSpace, MultSpace,     End")
+    val expected: Node = <e1><s1>SingleSpace</s1><s2>MultSpace</s2><s3>End</s3></e1>
+    TestUtils.assertEqualsXMLElements(expected, actual)
+  }
```
mbeckerle, line 243: You need a negative test showing that SP+ does *not* match no spaces.

daffodil-core/src/test/scala/org/apache/daffodil/general/TestPrimitives.scala
```diff
@@ -137,6 +137,131 @@ class TestPrimitives {
...
+      <xs:element name="e1" dfdl:lengthKind="delimited">
+        <xs:complexType>
+          <xs:sequence dfdl:separator=",%SP*;" dfdl:separatorPosition="infix">
+            <xs:element name="s1" type="xs:string" dfdl:lengthKind="delimited"/>
+            <xs:element name="s2" type="xs:string" dfdl:lengthKind="delimited"/>
+            <xs:element name="s3" type="xs:string" dfdl:lengthKind="delimited"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>)
+    val (_, actual) = TestUtils.testString(sch, "None, SingleSpace,   MultSpace")
```
mbeckerle, line 260: Expected this data to be "None,SingleSpace, ..." to show that ,%SP*; would match zero spaces after the comma.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DFDLDelimiter.scala
```diff
@@ -86,112 +92,183 @@ class Delimiter {
...
-    // 3	0		1		0		WSP+
-    // 4	0		1		1		WSP+
-    // 5	1		0		0		WSP
-    // 6	1		0		1		WSP+
-    // 7	1		1		0		WSP+
-    // 8	1		1		1		WSP+
-    if (numWSP_Plus != 0) {
+  // 				TRUTH TABLE
+  //		delim		delim+	delim*	RESULT
+  // 1	0		0		0		NONE
```
mbeckerle, line 227: Still looks wierd. Headings not over columns, etcc.

mbeckerle (changes requested):
Need a few more tests, and it would be good to invent tests to cover more of the identified codecov issues. 

Minor code formatting issues. 

No logic issues I could spot. Looking pretty close.

daffodil-lib/src/main/scala/org/apache/daffodil/cookers/EntityReplacer.scala
```diff
@@ -44,7 +44,7 @@ class EntitySyntaxException(msg: String) extends Exception(msg)
 final class EntityReplacer {
 
   val dfdlEntityName = "NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC[1-4]|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|NBSP|NEL|LS"
-  val dfdlCharClassEntityName = "NL|WSP|WSP\\*|WSP\\+|ES"
+  val dfdlCharClassEntityName = "NL|WSP|WSP\\*|WSP\\+|ES|LSP|LSP\\*|LSP\\+|SP\\*|SP\\+"
```
stevedlawrence, line 47: Mentioned elsewhere, but I wonder if we want to remove SP from `dfdlEntityName` and `entityCharacterUnicode` and add it to `dfdlCharClassEntityName`. It's technically not a char class, but I think maybe all the logic just works if we treat it as a character class of just a single character. That way SP is treated just like WSP and LS as far as all the reduce delim buf logic and delimiter logic goes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DFDLDelimiter.scala
```diff
@@ -86,112 +92,183 @@ class Delimiter {
...
-            // WSP exists and an equivalent representation was found
-            x.index = idx // Set the delimiter's index
-            q += x
+      if (delimType == "wsp") {
+        delim match {
+          case wsp: WSPDelim => numDelim_Standard += 1
+          case wsp: WSPPlusDelim => numDelim_Plus += 1
+          case wsp: WSPStarDelim => numDelim_Star += 1
+          case _ => {
+            val (newQ, newIdx) = updateQueue(numDelim_Standard, numDelim_Plus, numDelim_Star, q, idx, delimType)
```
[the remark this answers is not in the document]
OWL-Varun: You're right. With my previous code, the logic in updateQueue() needed to be run a couple times so I thought it best to create a method so that the method could be called. With your suggested changes, the code can be left the same as you said.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/AnnotatedSchemaComponent.scala
```diff
@@ -186,6 +188,25 @@ trait ResolvesScopedProperties
...
+     """LSP+ entity cannot act as separator without setting dfdlx:inlineWhiteSpace property to "allow".""")
+    schemaDefinitionUnless(!delimSep.contains("%LSP*;"),
+    """LSP* entity cannot act as separator without setting dfdlx:inlineWhiteSpace property to "allow".""")
+    schemaDefinitionUnless(!delimSep.contains("%SP;"),
+     """SP entity cannot act as separator without setting dfdlx:inlineWhiteSpace property to "allow".""")
+    schemaDefinitionUnless(!delimSep.contains("%SP+;"),
+     """SP+ entity cannot act as separator without setting dfdlx:inlineWhiteSpace property to "allow".""")
+    schemaDefinitionUnless(!delimSep.contains("%SP*;"),
+     """SP* entity cannot act as separator without setting dfdlx:inlineWhiteSpace property to "allow".""")
+  }
```
stevedlawrence, line 208: I'm not sure if this is the best way to handle this. There are a lot of different DFDL properties aside from the separator that can contain these new character classes (i.e. most properties). So the check probably doesn't want to check for the existence of these special values in specific properties.

Instead, what we probably want to do is pass a boolean into the functions that deal with processing entities that says whether to allow or error when the new entities are found. That way we don't have to check every single property that could have one of these, but instead just need to check where we parse strings.

daffodil-japi/src/test/resources/test/japi/mySchema5.dfdl.xsd
```diff
@@ -19,11 +19,12 @@
...
   targetNamespace="http://example.com" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
   xmlns:xsd="http://www.w3.org/2001/XMLSchema"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tns="http://example.com">
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tns="http://example.com"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions" >
 
   <annotation>
     <appinfo source="http://www.ogf.org/dfdl/">
-      <dfdl:format ref="tns:GeneralFormat" />
+      <dfdl:format ref="tns:GeneralFormat" dfdlx:inlineWhiteSpace="allow"/>
```
stevedlawrence, line 27: I would suggest no changing the behavior of existing tests. Create new tests specifically to check for these new entities.

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
```diff
@@ -76,6 +77,23 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:attribute name="inlineWhiteSpace" type="dfdlx:InlineWhiteSpaceEnum" default="disallow">
```
stevedlawrence, line 80: I'm not sure I understand the name `inlineWhiteSpace`. This property doesn't really control whether or not inline white space is allowed, it really controls whether or not you can use these new character classes.

As I review this, I even wonder if a property to control this is needed? Doesn't the fact that a schema is using `%LSP*;` mean they want/need the capability enabled? And it doesn't help with compatibility with other DFDL implementations because they aren't going to know about the new dfdlx:inlineWhiteSpace extension since that's specific to Daffodil. Are we over complicating things by adding this property?

stevedlawrence (commented):
Looking good, just a handful of minor suggestions/questions.

PR #833, opened 2022-08-24
WIP: Draft - Allow UStateForSuspension to be created without infoset DINode

mbeckerle:
This doesn't fix the bug, but changes the symptom.

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
@@ -118,7 +118,10 @@ abstract class UState(
 
   var dataOutputStream: DirectOrBufferedDataOutputStream
 
-  def currentInfosetNode: DINode
+  final def currentInfosetNode: DINode =
+    if (currentInfosetNodeMaybe.isEmpty) null
+    else currentInfosetNodeMaybe.get
+
   def currentInfosetNodeMaybe: Maybe[DINode]
   def escapeSchemeEVCache: MStackOfMaybe[EscapeSchemeUnparserHelper]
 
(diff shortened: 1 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala
```diff
@@ -394,8 +397,10 @@ final class UStateForSuspension(
   extends UState(vbox, mainUState.diagnostics, mainUState.dataProc, tunable, areDebugging) {
 
   dState.setMode(UnparserBlocking)
-  dState.setCurrentNode(thisElement.asInstanceOf[DINode])
-  dState.setContextNode(thisElement.asInstanceOf[DINode])
+  if (currentInfosetNodeMaybe.isDefined) {
+    dState.setCurrentNode(thisElement.asInstanceOf[DINode])
+    dState.setContextNode(thisElement.asInstanceOf[DINode])
+  }
```
stevedlawrence, line 403: After talking this morning, and thinking about this some more, I believe this is incorrect. In order to not have a DINode in when a UStateForSuspension is created, it must mean we are creating a suspension after unparsing has completed. The the only way that could happen is if we are creating a suspension from within another suspension that is evaluting at the end of unparse, which should never be allowed. So to have a UStateForSuspension without a currentNode implies there was some other invariant that isn't holding that we just aren't checking for.

mbeckerle: I concur, these changes are likely the wrong thing and will get backed out. 

The path I will proceed on is to first determine why this splitDOS operation is being missed and delayed to this point. Also I will be seeking a way to establish and check the invariant that this sort of thing is overwith at the end of parsing, so that we'll get a loud assertion failure if something does this, where it has to split sfter unparse is complete.

PR #835, opened 2022-08-30
Update to Scala XML 2.1.0

stevedlawrence:
The main difference in this new Scala XML version is changing the
XML.load() function to not ignore comments and processing instructions
and not coalesce cdata. These changes caused  some tests to fail which
required minor modifications. There are no changes to transitive
dependencies or license that need to be incorporated.

Additionally, the DaffodilConstructionLoader is modified to drop
processing instructions and comments. This gives a similar behavior to
the old XML.load function, and fixes bugs related to unexpected
comments.

DAFFODIL-2685, DAFFODIL-2237

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
@@ -292,17 +292,20 @@ class DaffodilConstructingLoader private[xml] (uri: URI,
   }
 
   /**
-   * Same CRLF/CR => LF processing as text gets.
+   * Drops comments
    */
   override def comment(pos: Int, s: String): Comment = {
-    Comment(text(pos, s).text)
+    // returning null drops comments
+    null
   }
 
   /**
-   * Same CRLF/CR => LF processing as text gets.
+   * Drops processing instructions
    */
-  override def procInstr(pos: Int, target: String, txt: String) =
-    ProcInstr(target, text(pos, txt).text)
+  override def procInstr(pos: Int, target: String, txt: String) = {
+    // returning null drops processing instructions
+    null
+  }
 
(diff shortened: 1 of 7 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
```diff
@@ -295,14 +295,17 @@ class DaffodilConstructingLoader private[xml] (uri: URI,
    * Same CRLF/CR => LF processing as text gets.
```
tuxji, line 295: Please replace this comment with "Drops comments."

daffodil-lib/src/main/scala/org/apache/daffodil/xml/DaffodilConstructingLoader.scala
```diff
@@ -295,14 +295,17 @@ class DaffodilConstructingLoader private[xml] (uri: URI,
...
    */
   override def comment(pos: Int, s: String): Comment = {
-    Comment(text(pos, s).text)
+    // returning null drops comments
+    null
   }
 
   /**
    * Same CRLF/CR => LF processing as text gets.
```
tuxji, line 303: Please replace this comment with "Drops processing instructions."

daffodil-lib/src/test/scala/org/apache/daffodil/xml/test/unit/TestXMLLoader.scala
```diff
@@ -115,15 +117,13 @@ b&"<>]]></x>"""
     // IF this test fails, it means that the scala xml loader have been FIXED (hooray!)
     // and our need for the ConstructingParser may have gone away.
```
tuxji, line 118: Please remove the comments at lines 115-118 since they're no longer needed.

tuxji (approved):
+1 with requested changes to some comments

mbeckerle (approved):
+1
Glad to get this upgrade done. The variations of different loaders over CDATA and other subtleties ahve been a pain in the neck.

PR #836, opened 2022-09-02
Exi fixes

jadams-tresys:
This PR addresses the issues around XML entities and nil elements, the resolution to those 2 things are as follows:

1. We should not be escaping anything before handing data off to the SAX ContentHandler.  It appears that the ContentHandler itself is responsible for escaping things as necessary.

2. It is not well documented in Exificient, but the SAX_NAMESPACE_PREFIXES_FEATURE is required to be enabled in order for it to function properly.  Not setting it was resulting in the issues with nil elements.

This PR also provides schema aware parsing, however in the couple tests that I have tried i have not noticed any difference.  The EXI files being created are identical to the schema unaware parsing.


daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -1300,28 +1300,6 @@ class TestCLIparsing {
...
-
-      Util.expectExitCode(ExitCode.Success, shell)
-
-      shell.send("exit\n")
-      shell.expect(eof)
-    } finally {
-      shell.close()
-    }
-  }
-
```
stevedlawrence, line 1324: Why remove his EXI test? Isn't this the only test that verifies EXI?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -695,9 +696,13 @@ object Main {
       case InfosetType.W3CDOM => Left(new W3CDOMInfosetOutputter())
       case InfosetType.SAX => Right(new DaffodilParseOutputStreamContentHandler(os, pretty=true))
       case InfosetType.EXI => {
-        val exiFactory = DefaultEXIFactory.newInstance()
-        exiFactory.getFidelityOptions.setFidelity(FidelityOptions.FEATURE_PREFIX, true)
-        val exiResult = new EXIResult()
+        val exiFactory = DefaultEXIFactory.newInstance
+        if (schemaOpt.isDefined) {
+          val grammarFactory = GrammarFactory.newInstance
+          val grammar = grammarFactory.createGrammars(schemaOpt.get.toURL.openStream(), DFDLCatalogResolver.get)
```
stevedlawrence, line 702: The `getInfosetOutputter` function is called for every iteration of a performance run. I imagine building this grammar is pretty expensive, so should probably be down outside the main loop and passed into this function when EXI is needed. In fact, maybe we should create the entire EXIFactory outside of the main loop and pass that into this function, so that this function only creates a new EXIResult/handler?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
```diff
@@ -63,8 +63,7 @@ class SAXInfosetOutputter(xmlReader: DFDL.DaffodilParseXMLReader,
       if (diSimple.hasValue) {
         val text =
           if (diSimple.erd.optPrimType.get.isInstanceOf[NodeInfo.String.Kind]) {
-            val s = remapped(diSimple.dataValueAsString)
-            scala.xml.Utility.escape(s)
+            remapped(diSimple.dataValueAsString)
```
stevedlawrence, line 66: Do we need to remove this same escaping for other XML infoset outputters that create objects? I think the Scala, W3C, and JDOM infoset outputters do the same thing. The InfosetInputters probably need to be changed to not unescape things as well?

That doesn't necessarily need to be done in this commit, but we should create a bug to investigate it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilParseOutputStreamContentHandler.scala
```diff
@@ -290,7 +290,9 @@ class DaffodilParseOutputStreamContentHandler(out: OutputStream, pretty: Boolean
   }
 
   override def characters(ch: Array[Char], start: Int, length: Int): Unit = {
-    writer.write(ch, start, length)
+    val str = new String(ch, start, length)
+    val escaped = scala.xml.Utility.escape(str)
+    writer.write(escaped.toArray, 0, escaped.length)
```
stevedlawrence, line 295: I believe `writer.write` has an overload that accepts a `String`. If we use that it might avoid `toArray`, which overhead with that needing to create a copy, I think?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -894,7 +899,7 @@ object Main {
               case Some(file) => new FileOutputStream(file)
             }
             val infosetType = parseOpts.infosetType.toOption.get
-            val eitherOutputterOrHandler = getInfosetOutputter(infosetType, output)
+            val eitherOutputterOrHandler = getInfosetOutputter(infosetType, output, parseOpts.schema.toOption)
```
stevedlawrence, line 902: Is there any value to have a unique infoset kind value for exi schema aware, for example `-I exi` vs `-I exi-sa`? This would allow us to create both and test for things like size and speed.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -1458,9 +1463,11 @@ object Main {
     data: InputSourceDataInputStream,
     saxContentHandler: ContentHandler): ParseResult = {
     val saxXmlRdr = processor.newXMLReaderInstance
-    saxXmlRdr.setContentHandler(saxContentHandler)
+    // SAX_NAMESPACE_PREFIXES_FEATURE is needed to preserve nil attributes with EXI
+    saxXmlRdr.setFeature(XMLUtils.SAX_NAMESPACE_PREFIXES_FEATURE, true)
     saxXmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBDIRECTORY, blobDir)
     saxXmlRdr.setProperty(XMLUtils.DAFFODIL_SAX_URN_BLOBSUFFIX, blobSuffix)
+    saxXmlRdr.setContentHandler(saxContentHandler)
```
stevedlawrence, line 1470: Are no changes needed to the `unparseWIthSax` function for schema unaware exi unparsing? Doesn't the `EXISource` `XMLReader` need to know the grammar to be able to deserialize the EXI file?

 Or does it just serialize the entire grammar into the EXI file? Seems like that would defeat the purpose of schema awareness so I imagine that's not the case?

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
tuxji: There appear to be 4 CLI tests which call -I exi but I'd like to know why the test was removed too.

```
daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
1311:      val cmd = String.format(Util.echoN("Hello") + "| %s parse -I exi -s %s -r e1 | md5sum", Util.binPath, testSchemaFile)

daffodil-cli/src/it/scala/org/apache/daffodil/performance/TestCLIPerformance.scala
79:      val cmd = String.format("%s performance -I exi -N 2 -t 2 -s %s -r matrix %s", Util.binPath, testSchemaFile, testInputFile)
207:      val cmd = String.format("%s performance --unparse -I exi -N 2 -t 2 -s %s -r e3 %s", Util.binPath, testSchemaFile, testInputFile)

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
618:      val cmd = String.format("%s unparse -I exi -s %s --root e1 %s", Util.binPath, testSchemaFile, testInputFile)
```

tuxji (commented):
0 until I see Steve's questions answered.  Please also add one or more JIRA issues to your PR message.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: I didn't realize that getInfosetOutputter gets called on every run  Having the factory set up once will also help with clarity for unparsing as well as there will only be one EXIFactory created that can be used for both parsing and unparsing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetOutputter.scala
jadams-tresys: I'm not sure.  The only reason I really needed to change it here was that Exificient's library was essentially double escaping the XML entity characters, so clearly they were performing some level of escaping already.  It makes sense to me that in SAX the ContentHandler should receive the raw string values and can then perform whatever escaping it needs to do.

For these other formats I don't know them well enough to say if there are any conventions like this with them.  Worth looking into, but probably outside the scope of this PR.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: That wouldn't be difficult to implement, so I don't see why not.  As you said it would also make for easier testing.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: This will be addressed in my next commit.  Going to modify things to only create one EXIFactory instance that can be used for both parsing and unparsing.

jadams-tresys:
Last commit addresses Steve's comments.  Will update the commit message with a bug ticket number (DAFFODIL-1954) when I squash the commits before merging.

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/input14.exisa
```diff
@@ -0,0 +1 @@
+����4::8���0��62�1���f �
```
stevedlawrence, line 1: I think these exisa files are causing the rat check to fail.

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/input18.exisa
```diff
@@ -0,0 +1 @@
+�,��4::8���0��62�1����[�
```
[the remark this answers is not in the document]
jadams-tresys: Actually have some fun results for this.  Tested with some data from a fairly significant DFDL schema project (~18k lines of schema) and got the following results:

Original File: 174 byte binary file
XML Infoset: 1493 bytes
EXI Infoset: 699 bytes
Schema aware EXI infoset: 160 bytes

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -378,6 +390,21 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
...
+
+    validateOpt(infosetType, stream) {
+      case (Some(InfosetType.EXI), Some(true)) => Left("Streaming mode is not currently supported with EXI infosets.")
+      case _ => Right(Unit)
+    }
+
+    validateOpt(infosetType, stream) {
+      case (Some(InfosetType.EXISA), Some(true)) => Left("Streaming mode is not currently supported with EXI infosets.")
+      case _ => Right(Unit)
+    }
```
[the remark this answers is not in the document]
jadams-tresys: Definitely makes it cleaner.

jadams-tresys:
There appears to be an issue either with Exificient and/or its interaction with our DFDLCatalogResolver that is causing the NITF schema to stack overflow during a parse when using schema aware EXI.  I've tested schema aware EXI with other relatively large schemas and there weren't any issues, but it definitely needs to be looked into.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -338,18 +339,9 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
...
-    validateOpt(infosetType, stream) {
-      case (Some(InfosetType.EXISA), Some(true)) => Left("Streaming mode is not currently supported with EXI infosets.")
-      case _ => Right(Unit)
-    }
-
-    validateOpt(infosetType, schema) {
-      case (Some(InfosetType.EXISA), None) => Left("A schema must be specified to use schema-aware compression with EXI")
+    validateOpt(infosetType, stream, schema) {
+      case (Some(InfosetType.EXI), Some(true), _) => Left("Streaming mode is not currently supported with EXI infosets.")
+      case (Some(InfosetType.EXISA), _, None) => Left("A schema must be specified to use schema-aware compression with EXI")
```
stevedlawrence, line 344: This is missing the case where the infosetType is InfosetType.EXSI and stream is true. Same with below.

stevedlawrence:
I played around with NITF and was able to reproduce the issue without any imports, so I doubt it's a catalog resolver issues.

Doing a little googling of the stack trace, I came across this bug report: https://sourceforge.net/p/exificient/bugs/31/

The maintainer discovered the issue is with very large maxOccurs, such as 99999:

> ... the EXI grammars do actually need to check whether the last state 99999 is reached. EXIficient therefore creates 99999 states and sometimes even more given that one may have nested elements etc.

Not surprisingly, the NITF schema uses `maxOccurs="999"` in a number of places, and `maxOccurs="99999"` in one place. I changed those to `unbounded` and schema aware exi worked, confirming that's the issue for NITF.

So I think this is just a limitation of the exificient library (and maybe EXI in general?), and that schemas with these very large maxOccurs should be avoided, and instead be replaced with unbounded. I don't think there's anything we can do in Daffodil to fix this issue.

stevedlawrence (approved):
+1

tuxji (approved):
+1

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/input18.exisa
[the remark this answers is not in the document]
jadams-tresys: The binary file that got sent through has some sections that are a lot of 0's I think.  The gzipped version of the original binary file is 145 bytes for comparison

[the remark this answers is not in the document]
mbeckerle: > The binary file that got sent through has some sections that are a lot of 0's I think. The gzipped version of the original binary file is 145 bytes for comparison

Is the EXI Compression flag enabled for this schema-aware version?

mbeckerle:
> I played around with NITF and was able to reproduce the issue without any imports, so I doubt it's a catalog resolver issues.
> 
> Doing a little googling of the stack trace, I came across this bug report: https://sourceforge.net/p/exificient/bugs/31/
> 
> The maintainer discovered the issue is with very large maxOccurs, such as 99999:
> 
> > ... the EXI grammars do actually need to check whether the last state 99999 is reached. EXIficient therefore creates 99999 states and sometimes even more given that one may have nested elements etc.
> 
> Not surprisingly, the NITF schema uses `maxOccurs="999"` in a number of places, and `maxOccurs="99999"` in one place. I changed those to `unbounded` and schema aware exi worked, confirming that's the issue for NITF.
> 
> So I think this is just a limitation of the exificient library (and maybe EXI in general?), and that schemas with these very large maxOccurs should be avoided, and instead be replaced with unbounded. I don't think there's anything we can do in Daffodil to fix this issue.

Use of maxOccurs="unbounded" is going to fail XSAT testing of schemas (quality check which looks for problematic XSD stuff). Unbounded things like regex's containing ".*" (any number of characters), or occurs unbounded get flagged and in some cases must be defended if not fixed. Saying that these must be unbounded because of Exefiicient bugs is not a very good reason really. 

I think this bug in Exeficient isn't something we can work around by just changing all schemas to use unbounded vs. sensible realistic upper bounds. I'd be curious if maxOccurs 99, 999, 9999 can work with enlarged stack space. 


E.g., did you really mean 8 quadrillion is allowed?

daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/input18.exisa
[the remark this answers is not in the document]
jadams-tresys: The EXIFactory CodingMode.COMPRESSION flag was not used for this.  However, I tested it out using the flag and it actually increased the size to 162 bytes for schema aware.  It does decrease the size of schema-unaware EXI to 541 bytes though.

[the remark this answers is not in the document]
stevedlawrence: One potential explanation is EXI's integer encoding is more compact that the original data format:

https://www.w3.org/TR/exi/#encodingUnsignedInteger

I believe it is the same as `varints` in google protocol buffers if anyone is familiar with that.

So for example, the original data format might represent an integer as a fixed 4 bytes. But a parsed value of `0` would only take up one byte in the EXI forma, saving 3 bytes. If there are enough of these more compact numbers it seems reasonable that the EXI file could be smaller than the original, even with the added overhead of EXI. I'm sure the EXI format uses other techniques to keep things compact.

PR #838, opened 2022-09-08
Require EPEL when building Daffodil on CentOS/RHEL

stevedlawrence:
The mxml-dependency is available in the EPEL repos for EL 7, 8, and 9. Rather than requiring that users build mxml from scratch on CentOS/RHEL, modify the build instructions to just require the EPEL repository, and provide a link to the instructions.

This also combines the CentOS and Fedora instructions to avoid any potential confusion or errors caused by users following Fedora instructions to build on RHEL.

DAFFODIL-2710

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -43,35 +43,11 @@ commands (or set them to `true` to disable C compilation altogether)
 if you don't want `sbt compile` to call your C compiler with `cc` and
 `ar` as the default commands.
 
-## CentOS 7
+## Fedora/CentOS/RHEL
 
-You can use the `yum` package manager to install most of the tools
-needed to build Daffodil:
-
-    sudo yum install clang gcc git java-11-openjdk-devel llvm make pkgconfig
-
-If you want to use clang instead of gcc, you'll have to set your
-environment variables `CC` and `AR` to the clang binaries' names:
-
-    export CC=clang AR=llvm-ar
-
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

Very concise :-).

mbeckerle (approved):
+1

mbeckerle:
Do we need to test this on RHEL 9 too?

stevedlawrence:
> Do we need to test this on RHEL 9 too?

Probably a good idea, I know plenty of systems moving to RHEL 9, and it should be an easy test. I'll give it a shot.

stevedlawrence:
Turns out there are no mxml rpms available for EL9 yet. The maintainer of the package is still active, so I imagine it will get added once he gets around to it. But that means these instructions don't work for for RHEL/CentOS 9 yet.

I did test manually downloading and installing the EL8 mxml rpms in CentOS 9 and everything built correctly and tests ran, so there should be no issue once mxml is added to EL9 EPEL.

I guess for RHEL9, our options are:
1. add back instructions for building mxml from source
2. give instructions to download/install the EL8 mxml-rpms
3. just know that building on RHEL9 isn't supported yet (with the above options as workarounds if it's really needed) and see if there's anything we can do to get mxml added to EPEL 9 in the meantime.

Considering EL9 is relatively new, I'd prefer the last option.

PR #840, opened 2022-09-09
Update mbeckerle signing key to sha-512

mbeckerle:
DAFFODIL-2727


stevedlawrence (approved):
+1

Looks good. I confirmed this imports correct and has the correct digest algorithm.

olabusayoT (approved):
+1

PR #846, opened 2022-09-22
Add 'exisa' to help message for infoset types.

jadams-tresys:
Missed this during the initial implementation.

DAFFODIL-1959

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -313,7 +313,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
 
     val config = opt[File](short = 'c', argName = "file", descr = "XML file containing configuration items")
     val vars = props[String](name = 'D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing. Can be prefixed with {namespace}.")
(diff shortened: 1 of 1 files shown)


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -313,7 +313,7 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
 
     val config = opt[File](short = 'c', argName = "file", descr = "XML file containing configuration items")
     val vars = props[String](name = 'D', keyName = "variable", valueName = "value", descr = "Variables to be used when parsing. Can be prefixed with {namespace}.")
-    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', 'exi', or 'null'. Defaults to 'xml'.", default = Some(InfosetType.XML))
+    val infosetType = opt[InfosetType.Type](short = 'I', argName = "infoset_type", descr = "Infoset type to output. Use 'xml', 'scala-xml', 'json', 'jdom', 'w3cdom', 'sax', 'exi', 'exisa', or 'null'. Defaults to 'xml'.", default = Some(InfosetType.XML))
```
stevedlawrence, line 316: Thoughts on changing this to something like:

```scala
descr = "Infoset type to output. Type can be: " + InfosetType.values.mkString(", ") + ". Defaults to xml".
```

This way we don't need to remember to update the help everytime a new infoset type is added?

stevedlawrence (approved):
+1

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
jadams-tresys: Yup, I like that change.  Will do.

tuxji (approved):
+1

Good suggestion, Steve.

PR #849, opened 2022-09-28
Handle validation of float/double with INF/NaN values

stevedlawrence:
Currently, min/max/Inclusive/Exclusive restriction facets are converted to BigDecimal during compilation, and parsed values are converted to BigDecimal to compare against those values when validation is enabled. However, float and double types can parse to infinity or not-a-number, which cannot be converted to a BigDecimal and causes an abort.

To fix this, when checking min/max/Inclusive/Exclusive restrictions we first check if the data is a float or a double, and if so we convert the facet BigDecimal value to a float or double, and use the Float/Double compareTo function which handles Inf/NaN appropriately.

DAFFODIL-2720


tuxji (approved):
+1

I like how clear and concise the new lines of code are.

PR #859, opened 2022-10-19
Pass schemas to Exificient via String, not stream

jadams-tresys:
Passing the schemas in by stream resulted in not having correct path information and made it impossible for the DFDLCatalogResolver to resolve the paths to included schemas. Instead we want to pass the path of the schema in to Exificient and maintain that correct path information.

DAFFODIL-1959


stevedlawrence (approved):
+1 :+1:

tuxji (approved):
+1

PR #862, opened 2022-10-21
Expose the DFDLCatalogResolver to the SAPI/JAPI

jadams-tresys:
DAFFODIL-2739

### DIFF
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
@@ -61,6 +61,7 @@ import org.apache.daffodil.util.Maybe._
 import org.apache.daffodil.util.MaybeULong
 import org.apache.daffodil.xml.NS
 import org.apache.daffodil.xml.XMLUtils
+import org.apache.daffodil.xml.DFDLCatalogResolver
 
 /**
  * API Suitable for Java programmers to use.
@@ -1154,4 +1155,29 @@ class DaffodilUnparseContentHandler private[japi] (sContentHandler: SDaffodilUnp
     contentHandler.processingInstruction(target, data)
   override def skippedEntity(name: String): Unit =
     contentHandler.skippedEntity(name)
+}
+
+/**
+ * Returns the EntityResolver used by Daffodil to resolve import/include
+ * schemaLocations.
+ *
+ * The entity resolver attempts to resolve namespaces and systemId's in the
+ * following order:
+ *
+ * 1. Use an org.apache.xml.resolver.Catalog/CatalogManager. By default the
+ *    Catalog only includes the daffodil-built-in-catalog.xml, but additional
+ *    catalogs can be added by putting CatalogManager.properties on the
+ *    classpath when daffodil is run.
+ *
+ * 2. If not resolved in step 1, schemaLocations are resolved relative to the
+ *    importing schema URI, which could either be a file on the filesystem or in
+ *    a jar on the classpath.
+ *
+ * The EntityResolver isn't thread safe, but it also is expensive and stateful,
+ * so we use ThreadLocal to only create one instance per thread.
+ */
+object DaffodilXMLEntityResolver {
+  def getEntityResolver: org.xml.sax.EntityResolver = DFDLCatalogResolver.get
+  def getXMLEntityResolver: org.apache.xerces.xni.parser.XMLEntityResolver = DFDLCatalogResolver.get
+  def getLSResourceResolver: org.w3c.dom.ls.LSResourceResolver = DFDLCatalogResolver.get
 }
--- a/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
+++ b/daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
@@ -58,6 +58,7 @@ import org.apache.daffodil.util.Maybe._
 import org.apache.daffodil.util.MaybeULong
 import org.apache.daffodil.xml.NS
 import org.apache.daffodil.xml.XMLUtils
+import org.apache.daffodil.xml.DFDLCatalogResolver
 
 private class Daffodil private {
   // Having this empty but private companion class removes the constructor from
@@ -1088,4 +1089,29 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
     contentHandler.processingInstruction(target, data)
   override def skippedEntity(name: String): Unit =
     contentHandler.skippedEntity(name)
+}
+
+/**
+ * Returns the EntityResolver used by Daffodil to resolve import/include
+ * schemaLocations.
+ *
+ * The entity resolver attempts to resolve namespaces and systemId's in the
+ * following order:
+ *
+ * 1. Use an org.apache.xml.resolver.Catalog/CatalogManager. By default the
+ *    Catalog only includes the daffodil-built-in-catalog.xml, but additional
+ *    catalogs can be added by putting CatalogManager.properties on the
+ *    classpath when daffodil is run.
+ *
+ * 2. If not resolved in step 1, schemaLocations are resolved relative to the
+ *    importing schema URI, which could either be a file on the filesystem or in
+ *    a jar on the classpath.
+ *
+ * The DFDLCatalogResolver isn't thread safe, but it also is expensive and stateful,
+ * so we use ThreadLocal to only create one instance per thread.
+ */
+object DaffodilXMLEntityResolver {
+  def getEntityResolver: org.xml.sax.EntityResolver = DFDLCatalogResolver.get
+  def getXMLEntityResolver: org.apache.xerces.xni.parser.XMLEntityResolver = DFDLCatalogResolver.get
+  def getLSResourceResolver: org.w3c.dom.ls.LSResourceResolver = DFDLCatalogResolver.get
 }
--- a/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
+++ b/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
@@ -1205,4 +1205,12 @@ public class TestJavaAPI {
         assertEquals(expectedData, bos.toString());
     }
 
+    @Test
+    public void testJavaAPI26() throws IOException, ClassNotFoundException, ExternalVariableException {
+        // Demonstrates the use of the various EntityResolver methods from Java
+
+        assertTrue(DaffodilXMLEntityResolver.getEntityResolver() != null);
+        assertTrue(DaffodilXMLEntityResolver.getXMLEntityResolver() != null);
+        assertTrue(DaffodilXMLEntityResolver.getLSResourceResolver() != null);
+    }
 }


daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
   override def skippedEntity(name: String): Unit =
     contentHandler.skippedEntity(name)
 }
+
+/**
+ * Exposes access to the DFDLCatalogResolver to the Java API.
```
stevedlawrence, line 1095: DFDLCatalogResolver is really an internal thing that most users aren't going to know about, so describing this as a way to get access to that might not be all that helpful. I think we can just remove this line?

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
...
+ *
+ * Resolves URI/URL/URNs to loadable files/streams.
```
stevedlawrence, line 1097: Should way make it clear that this thing is what Daffodil uses to resolve entities? E.g.

> Returns the EntityResolver used by Daffodil to resolve import/include schemaLocations

The idea being that if you are using DFDL schemas in outside of Daffodil and you need Daffodil's same behavior, then this is what you should use.

Also, should we be specific how this thing we return actually resolves entities? For example, it looks at the catalog manager, it looks at class paths, it looks at relative paths, etc. I'm not sure exactly what the logic is, but this feels like the right place for that documentation, and would be useful information for people using this class and know how Daffodil actually resolves entities. It would also resolve [DAFFODIL-2616](https://issues.apache.org/jira/browse/DAFFODIL-2616)

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
...
+ *
+ * The DFDLCatalogResolver implements the following interfaces:
+ *  - org.apache.xerces.xni.parser.XMLEntityResolver
+ *  - org.w3c.dom.ls.LSResourceResolver
+ *  - org.apache.sax.EntityResolver
+ *  - org.apache.sax.ext.EntityResolver2
```
stevedlawrence, line 1103: I'm not sure listing the interfaces is particularly helpful, especially since most people probably don't care too much about them aside from EntityResolver? 

Maybe instead we just change the get function to look something like this:
```scala
def get: EntityResolver = DFDLCatalogResolver.get
```
That way it is self documenting that this returns an `EntityResolver`.

Though, would a user ever need one of the other resolver interfaces? I would guess no? But if so, maybe we have additional functions that return each one that a user might need? E.g.

```
def getEntityResolver: EntityResolver = DFDLCatalogResolver.get
def getXMLEntityResolver: XMLEntityResolver = DFDLCatalogResolver.get
...
```

Seems a bit verbose, and I'm not sure if that makes sense or not?

Maybe another option is just have a class, e.g.

```scala
class DaffodilXMLEntityResolver extends DFDLCatalogResolver
```
And then maybe the javadoc just shows the interfaces that are implemented? I'm not sure if it does or not?

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
...
+ */
```
stevedlawrence, line 1104: I think we need a blurb about thread safety. Something about the returned EntityResolver not being thread safe, but if this get function is called from within different threads then it is since it returns a unique `EntityResolver` per thread instance. I'm not sure the best way to word that, but something like that feels important.

And thinking about threads, I wonder if we should always return a new instance, and we just say it's not thread safe? If the user want to make optimizations like what Daffodil does and store instances in a ThreadLocal then they can do that themselves. I'm not sure we should be doing that for them and forcing them into a single instance per thread?

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
...
+object DaffodilXMLEntityResolver {
+  def get = DFDLCatalogResolver.get
+}
```
stevedlawrence, line 1107: Have you checked that the java/scala docs show up reasonably well? I'm not sure how well comments on objects get converted to java docs.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
```diff
@@ -1089,3 +1090,18 @@ class DaffodilUnparseContentHandler private[sapi] (sContentHandler: SDaffodilUnp
...
+ * Resolves URI/URL/URNs to loadable files/streams.
+ *
+ * The DFDLCatalogResolver implements the following interfaces:
+ *  - org.apache.xerces.xni.parser.XMLEntityResolver
+ *  - org.w3c.dom.ls.LSResourceResolver
+ *  - org.apache.sax.EntityResolver
+ *  - org.apache.sax.ext.EntityResolver2
+ */
+object DaffodilXMLEntityResolver {
+  def get = DFDLCatalogResolver.get
```
stevedlawrence, line 1106: Would be good to add some tests to call this function as an example to make sure the Java API usage is reasonable and isn't messed up by being scala. Doesn't necessarily need to do something complex with it aside from may ensure it's not null or something.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
jadams-tresys: I think I like the idea of the multipl getXXXResovler functions.  I believe I needed to use the XMLEntityResolver with one of the EXI implementations, so I definitely needed more than just the SAXEntityResolver interface.

daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala
jadams-tresys: Actually looks just like a normal Java class documentation, so I think this is good to go

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -1155,3 +1156,23 @@ class DaffodilUnparseContentHandler private[japi] (sContentHandler: SDaffodilUnp
...
 }
+
+/**
+ * Returns the EntityResolver used by Daffodil to resolve import/include
+ * schemaLocations.
+ *
+ * DFDLCatalogResolver uses an org.apache.xml.resolver.Catalog/CatalogManager to
+ * attempt to resolve namespaces and systemId's. If we are unable to resolve a
+ * file with the Catalog, we attempt to resolve the file using the systemId path
+ * from the schemaLocation attribute. If this fails we throw a
```
stevedlawrence, line 1167: > we attempt to resolve the file using the systemId path from the schemaLocation attribute.

Does this require a little extra clarification? Don't we look at both the classpath and the filesystem, and if so, which do we do first? Is it worth mentioning relative paths in schemaLocations are relative to the file that includes (it I think that's right)?

This might also work better as a numeric list, e.g.

```
The returned resolver looks in the following places to resolve entities in this order:

1. org.apache....CataogManager  ...
2. Java classpath ...
3. Filesystem ...
```

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -1155,3 +1156,23 @@ class DaffodilUnparseContentHandler private[japi] (sContentHandler: SDaffodilUnp
   override def skippedEntity(name: String): Unit =
     contentHandler.skippedEntity(name)
 }
+
+/**
+ * Returns the EntityResolver used by Daffodil to resolve import/include
+ * schemaLocations.
+ *
+ * DFDLCatalogResolver uses an org.apache.xml.resolver.Catalog/CatalogManager to
```
stevedlawrence, line 1164: Don't mention `DFDLCatalogResolver`, that class isn't public so mentioning it in public API is just going to be confusing. The user will just get an `EntityResolver`, `XmlEntityResolver`, etc.--that's all they need to know about.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -1155,3 +1156,23 @@ class DaffodilUnparseContentHandler private[japi] (sContentHandler: SDaffodilUnp
...
+ * attempt to resolve namespaces and systemId's. If we are unable to resolve a
+ * file with the Catalog, we attempt to resolve the file using the systemId path
+ * from the schemaLocation attribute. If this fails we throw a
+ * SAXParseException.
+ *
+ * The DFDLCatalogResolver isn't thread safe, but it also is expensive and stateful,
+ * so we use ThreadLocal to only create one instance per thread.
+ */
+object DaffodilXMLEntityResolver {
+  def get = DFDLCatalogResolver.get
```
stevedlawrence, line 1174: I don't think we should return this. No one needs access to a `DFDLCatalogResolver` and it's capabilities, they just need one of the other kind of revolvers. If we do find someone needs access to this, we can consider making it public. But until then let's keep that implementation private. It's much easier to make something public that's private than to make something private that's public.

stevedlawrence (approved):
+1, which some minor suggestions

tuxji (approved):
+1

I agree with all of Steve's comments.  I'm good with the PR after you resolve them.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
jadams-tresys: Actually, looking through our code I don't think we technically reach out to the filesystem at all.  We only search for things on the classpath or relative to some context URI t hat gets passed in

stevedlawrence: Ah, I guess that makes sense, and it just depends on if the root schema was found in on the classpath or the filesystem. If the root schema was found on the classpath all imports will look on the classpath (since they are relative to things found on the classpath), and likewise with the filesystem.

So maybe instead of mentioning the classpath we say something like, catalogmanager is first, otherwise schemaLocations are resolves relative to the importing schema URI, which could be a file on the filesystem or in a jar on the classpath. Probably can be worded better, but something like that to make it clear we support classpaths or file depending on the original root schema.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -1155,3 +1156,28 @@ class DaffodilUnparseContentHandler private[japi] (sContentHandler: SDaffodilUnp
...
+ * 1. Use an org.apache.xml.resolver.Catalog/CatalogManager. By default the
+ *    Catalog only includes the daffodil-built-in-catalog.xml, but additional
+ *    catalogs can be added by putting CatalogManager.properties on the
+ *    classpath when daffodil is run.
+ *
+ * 2. If not resolved in step 1, schemaLocations are resolved relative to the
+ *    importing schema URI, which could either be a file on the filesystem or in
+ *    a jar on the classpath.
+ *
+ * The DFDLCatalogResolver isn't thread safe, but it also is expensive and stateful,
```
stevedlawrence, line 1176: One more instance of `DFDLCatalogResolver` to remove. New changes look good to me.

PR #863, opened 2022-10-24
Change recoverable errors to validation errors instead of schema definition warnings

stevedlawrence:
A dfdl:assert with failureType="recoverableError" is currently treated as a schema definition warning. However, SDW's are usually for things that can be safely ignored, and these asserts are generally used for validation checking beyond the capabilities of XSD that should not be ignored, or should at least be treated differently so that a user can easily differentiate between the two.

This modifiers these asserts so that the errors are turned validation errors accessible just like those created when validation is turned on. This also refactors the handling of errors to avoid duplication between assert expressions and assert patterns, as well as changes to make error messages consistent.

Note that this is backwards incompatible change, but is arguably the more correct behavior since these asserts really are predominantly used for extended validation checks.

DAFFODIL-2357


olabusayoT (approved):
Looks great! +1

tuxji (approved):
+1

Agreed.

PR #865, opened 2022-10-28
Refactor integration tests for clarity and speed

stevedlawrence:
The current integration tests have too much boilerplate that can be
refactored with some new APIs. Additionally, the integration tests are
incredibly slow due to all the forking and creation of new VMs. This
fixes both of these issues.

The new integration test API creates a new `runCLI` function, accepting
an array of string arguments (with a new "args" string interpolator to
make specifying this array easier), various execution parameters, a
function that contains the normal `Expect` logic all our of integration
tests already use, and finally the expected exit code.

A new `path` function is created to help deal with paths in a OS
agnostic manner.

New `withTempFile` and `withTempDirectory` functions are added to help
with creating temporary files in a special temporary daffodil directory,
and to ensure they are cleaned up when the test is complete.

A new `withSysProp` function is used to temporarily change and restore
system properties, helping to avoid changes that might interfere with
other tests.

With these new functions, integration tests now look something like
this:

```scala
    val schema = path("path/to/schema.dfdl.xsd")
    runCLI(args"parse -s $schema") { cli =>
      cli.send("data to parse", inputDone = true)
      cli.expect("<data>string to expect</data>")
    } (ExitCode.Success)
```

The runCLI function supports executing Daffodil in either a new process
or thread, defaulting to thread unless a test requires a separator
process (e.g. classpath changes). Streams are setup for handling
stdin/out/err. When threaded, the thread calls the
org.apache.daffodil.Main run function, and configures Main and log4j to
output to these streams. When using a new processing, the daffodil
sh/bat script is directly executed.

A new `Expect` instance is configured so that it can read/write to these
streams and validate the output, with a new CLITester class used to
provide helper wrappers around `Expect` to make logic a bit less
verbose.

Multiple changes were also made to Main and the Debugger to ensure they
always use the specified streams instead of always using stdin/out/err.
This is really only needed for this integration testing, but could
potentially be useful in other use cases where stdin/out/err need need
to be mimicked without forking a new process.

Note that the thread approach isn't thread safe, but integration tests
are already not run in parallel. By using threads instead of spawning
new processes where possible, we drastically decrease the overhead, with
test suites now taking a number of seconds instead of a number of
minutes, averaging and order of magnitude speed up.

This also tweaks various CLI tests to not expect data from files,
instead hardcoding what to expect in each test. This is a bit less
exact, but it make the logic simpler and more consistent, and allows us
to remove many files from the repo. Also found and removed many debugger
command files that are never actually used.

This also enables the blob CLI tests but adds an assumption so that they
are skipped if the large input files have not been generated. This gives
us a reminder that those tests exist and we can periodically ensure they
work by generating the files.

This also ensures coverage is enabled in the github workflow every time
we execute SBT. Otherwise, changes of coverage enablement cause a
recompile. By doing this we avoid recompiling the same files multiple
times which saves a couple of minutes overall. Note that we originally
disabled coverage on some steps because it caused failures, but whatever
caused that seems to have been fixed.

DAFFODIL-2381


tuxji (commented):
I think the new API looks fine and I would like to see the rest of the integration tests converted to use it too.  The converted tests are much smaller as well as run much faster, which is a big quality of code improvement.  As I noted in a couple of places, we might even be able to remove the old API entirely (which I would prefer).  

We may need to pay attention to other places in Daffodil which call some variation of "println" and determine if these places need to be updated or not.  We also may need to increase code coverage (this pull request currently hits only 65% code coverage).

daffodil-cli/src/it/scala/org/apache/daffodil/parsing/TestCLIParsing.scala
```diff
@@ -265,1124 +157,547 @@ class TestCLIparsing {
...
   @Test def test_CLI_Parsing_XCatalog_Resolution_Failure(): Unit = {
-    val schemaFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_import_failure.dfdl.xsd")
-    val testSchemaFile = if (Util.isWindows) Util.cmdConvert(schemaFile) else schemaFile
+    val schema = path("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_import_failure.dfdl.xsd")
 
-    val xcatalogFile = Util.daffodilPath("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_invalid.xml")
-    val testXcatalogFile = if (Util.isWindows) Util.cmdConvert(xcatalogFile) else xcatalogFile
+    val xcatalogFile = path("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/xcatalog_invalid.xml")
 
     val DAFFODIL_JAVA_OPTS = Map("DAFFODIL_JAVA_OPTS" -> ("-Dxml.catalog.files=" + testXcatalogFile + " -Xms256m -Xmx2048m -Dfile.encoding=UTF-8"))
```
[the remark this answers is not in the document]
stevedlawrence: I've converted pretty much everything to this new API, you are right that tests that set DAFFODIL_JAVA_OPTS don't really need to do that, and in the case where it is needed there are workarounds using setProperty.

However, most of the UserDefinedFunction integration tests set the DAFFODIL_CLASSPATH variable. Essentially what these tests do is put all `.class` files defined in daffodil-udf on the classpath, and then individual tests also add specific `META-INF/services` directories on the classpath to tests loading specific UDF. I'm not sure there's a way to modify the classpath like this without forking a new process? Any ideas oh if that's possible?

Another consideration is that the `UserDefinedFunctionService` object loads all the available UDF's on initialization and stores their information in a mutable HashMap. So if we were able to have classpath specific for these tests, we would also need to ensure this object is reset/reloaded appropriately. It's possible sbt magic already handles resetting objects? I feel like they don't persist inbetween tests, or maybe test suites?

Any thoughts?

Note that even if we do fork for only the UDF integration tests, it still feels like a pretty big win. With the current changes, all integration tests take about 2 minutes on my machine (half of that is UDFs), which is a big improvement over the usual 20+ minutes.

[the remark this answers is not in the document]
tuxji: Yes, it's possible but far from easy to search a different classpath at runtime without forking a new process.  Java 9 and later versions have removed the hackish (it was never part of the API) ability to modify the system classloader at runtime.  The only portable solution is to create a custom classloader and modify your application to use the custom classloader instead of the system classloader.  There appears to be no solution that will allow the UserDefinitionFuction integration tests to avoid forking unless we modify not only these tests, but also the daffodil-udf code (if it's possible to restrict the scope of the changes to only that library), or else the Main class itself.  I think we will have to be satisfied with the speedup we get from converting most of the integration tests and accept the slowness of forking for the UDF integration tests.  Running all integration tests in 2 minutes instead of 20+ minutes is still wonderful.

Here are the two best StackOverflow questions about changing the Java classpath at runtime.  I also googled for Scala as well as Java answers and didn't find a better approach although I did find a Scala library which lets you search any directory or jars for classes (the drawback is you can't load these classes unless the classes are also loadable by your system classloader or a custom classloader).

https://stackoverflow.com/questions/252893/how-do-you-change-the-classpath-within-java
https://stackoverflow.com/questions/60764/how-to-load-jar-files-dynamically-at-runtime
http://software.clapper.org/classutil/

mbeckerle (approved):
+1

This is a big improvement. The speed will be very nice, but also elimintes a bunch of portability concerns with the tests also where MS-Windows-isms were not adeuqately abstracted.

That is, this fixes DAFFODIL-2381

I belive this will also make debugging these tests more straightforward, as one can just set breakpoints in Daffodil per usual and debug the test, rather than various temp/hack ways Ih've had to modify the tests (have them call main, instead of spawning a process) for debugging in IDE.

stevedlawrence:
I've completed the refactor using a new API and address all comments, I think this is ready for a full review.

stevedlawrence:
Looks like the PR checks finished in 17-25, with most taking about 20 mins. Compared to the most recent merged commit which took between 30-60 averaging on the higher end, this is a nice speed up.

mbeckerle:
This will be post 3.4.0, i.e., on the 3.5.0 branch yes?

stevedlawrence:
> This will be post 3.4.0, i.e., on the 3.5.0 branch yes?

Correct, once this gets the +1's, I'll hold off merging until 3.4.0 is released.

mbeckerle (approved):
+1

daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
```diff
@@ -17,253 +17,143 @@
...
+import org.apache.daffodil.CLI.Util._
 import org.apache.daffodil.Main.ExitCode
 
 class TestCLISaveParser {
 
-  val output1 = Util.getExpectedString("output1.txt")
-  val output4 = Util.getExpectedString("output4.txt")
-  val output6 = Util.getExpectedString("output6.txt")
-  val output12 = Util.getExpectedString("output12.txt")
-  val savedParserFile = new File("savedParser.xsd.bin")
```
tuxji, line 36: This PR inlines these files' expected strings into the tests below and uses withTempFile { parser instead of savedParserFile.  Nice!

daffodil-cli/src/it/scala/org/apache/daffodil/saving/TestCLISaveParser.scala
```diff
@@ -272,173 +162,108 @@ class TestCLISaveParser {
...
+      runCLI(args"-d save-parser -s $schema -r matrix $parser") { cli =>
+        cli.expectErr("Some error about -d not being valid with save-parser")
+      } (ExitCode.Usage)
+    }
+  }
 
-  // See DFDL-1141
-  /*@Test def test_3036_CLI_Saving_SaveParser_debug() {
+  // DAFFODIL-1141
+  /*@Test*/ def test_3037_CLI_Saving_SaveParser_trace(): Unit = {
```
tuxji, line 217: This PR uncomments the test's body and keeps only the `@Test` commented out.  Nice, this will keep the test compiling even after changes to functions.

daffodil-cli/src/it/scala/org/apache/daffodil/unparsing/TestCLIUnparsing.scala
```diff
@@ -17,679 +17,319 @@
...
-      cmd = String.format("""echo '<ex:validation_check xmlns:ex="http://example.com">test</ex:validation_check>' | %s unparse -s %s -r validation_check --validate off """, Util.binPath, testSchemaFile)
-      shell.sendLine(cmd)
-
-      shell.sendLine("exit")
-      shell.expect(eof)
-    } finally {
-      shell.close()
-    }
+  // DAFFODIL-1346
+  /*@Test*/ def test_3576_CLI_Unparsing_validate(): Unit = {
```
tuxji, line 171: Again, nice job keeping a commented out test compiling.

project/Rat.scala
```diff
@@ -39,22 +39,6 @@ object Rat {
...
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1333"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1334"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1337"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1338"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1339"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1340"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1382"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1591"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1602"),
-    file("daffodil-cli/src/it/resources/org/apache/daffodil/CLI/debugger/1863"),
```
tuxji, line 57: Does the PR delete these files because their contents had already been inlined as input strings into the corresponding CLI tests in an earlier PR?  Nothing seems to read or use these files even before this PR was created, so it's time to remove them.

tuxji (approved):
+1

daffodil-cli/src/it/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -17,183 +17,501 @@
...
+   */
+  private class CLIThread(args: Array[String], in: InputStream, out: OutputStream, err: OutputStream) extends Thread {
+    var exitCode = ExitCode.Failure
+
+    override def run(): Unit = {
+      val psOut = new PrintStream(out)
+      val psErr = new PrintStream(err)
+
+      // configure the CLI and log4j to use our custom streams, nothing should
+      // not actually use stdin/stdout/stderr
```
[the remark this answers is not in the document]
stevedlawrence: I used this command to find non-test related println/printf:

    grep -R --exclude-dir=test -e '\(println\|printf\)' * 

This found the following:

* Main/CLIDebuggerRunner/TraceDebuggerRunner - these all use new logic to output to a specified stream
* DirectOrBufferedDataOutputStream.scala - this println occurs in the `dumpChain()` function, which is used for debugging suspensions
* PropertyGenerator.scala - run by sbt to generate code--needs to always output to stdout for sbt
* CodeGenerator.scala - run by sbt to update example files--needs to be output to stdout for sbt
* Tak.scala - used for timing TDML tests/performance related

All other instances are in commented out code. So I think all instances that related to the CLI have been fixed, or are intended to output to stdout. We might want to go through all the tests and commented out code and inspect those println's to see if they are really needed, but I think that's out of scope for this ticket.

daffodil-cli/src/it/scala/org/apache/daffodil/executing/TestCLIexecuting.scala
```diff
@@ -1,245 +0,0 @@
...
-import net.sf.expectit.ExpectIOException
-import org.apache.daffodil.Main.ExitCode
-
-class TestCLIexecuting {
-
-  val output3 = Util.getExpectedString("output3.txt")
-  val output13 = Util.getExpectedString("output13.txt", true)
-  val output14 = Util.getExpectedString("output14.txt", true)
-  val output15 = Util.getExpectedString("output15.txt", true)
-  val output16 = Util.getExpectedString("output16.txt", true)
```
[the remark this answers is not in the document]
stevedlawrence: The `TestCLIexecuting` and `TestCLIlisting` files were both really about running the CLI "test" subcommand to run TDML tests, with one focused on executing tdml tests and the other focused on listing tdml tests. And in fact, the "executing" file even had multiple tests that were just listing tests, and I found some tests that were duplicated in the two files, so the intended distinction wasn't even clear. I also think "executing" also very clear since that could be about executing anything, like  parse, unparse, code generator, etc.

So to make it more obvious that this is the file for the "test" command, I combined these two files into a "TestCLItdml" file (I thought "TestCLItest" might be confusing?) and removed duplicates when I found them.

project/Rat.scala
stevedlawrence: Looking at the commit history, when debugger tests were originally added (nearly 10 years) ago they used the `-d` option and passed in these files. A short time later (04e50f9aad93eae15a02d7eff4b9440e84bc9e2d), we switched to ExpectIt and changed these tests to inline the debugger commands, and I guess we just forgot to remove the files.

stevedlawrence:
> We also may need to increase code coverage (this pull request currently hits only 65% code coverage).

I didn't really add any new code except in couple minor places. I think lack of coverage in this PR is because I've made changes to printlns that were just already missing coverage.

In the  past I know I've hesitated to add new CLI tests to get coverage because of the high overhead. But now that integration tests are much quicker, we may want to consider adding new tests to improve our CLI coverage. I think that's out of scope for this PR though, since this is more about just improving the CLI test infrastructure.

mbeckerle:
When this says 64% coverage, is that the unit tests plus the CLI tests? Or just the CLI tests alone?

stevedlawrence:
I believe it's both unit tests + CLI tests. But most of the lines missing coverage are println's in Main.scala that can only be tested with he CLI/Integration tests.

mbeckerle:
Something doesn't add up.

We were at roughly 80% coverage on a 110K-lines code base.

We dropped to 64% coverage.  That's a 16% drop, meaning 17K-lines are no longer covered that were.

That's can't be explained by println statements in the Main/CLI code. 

So I'm assuming my arithmetic here doesn't make sense. But I clearly don't understand what we mean by 80+% dropping to 64% coverage.

stevedlawrence:
I think that 64% is just the the coverage of this PR, not the coverage of the entire codebase. So of the lines this PR changed to `src/main`, 64% of those lines have a test that hit them. The overall coverage is still about 80%.

Here's the codecov page for this PR that makes it a bit more clear:

https://app.codecov.io/gh/apache/daffodil/pull/865

mbeckerle:
Ok. Then I suggest adding a ticket to enhance code coverage here. But I wouldn't hold up committing this much improvement for that.

stevedlawrence:
[DAFFODIL-2746](https://issues.apache.org/jira/browse/DAFFODIL-2746) added to improve CLI code coverage.

PR #875, opened 2022-11-14
Remove all API deprecated methods/classes/objects

Shanedell:
Remove all API deprecated methods/classes/objects

- daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala: Use withDebugger and withDebugging instead of setDebugger and setDebugging. Rename function setupDebugOrTrace to withDebugOrTrace, also update it to return the processor to be used.
- daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala: Remove deprecated functions.
- daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala: Remove deprecated functions. Also add withDebugger and withDebugging functions.
- daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala: Remove isAtEnd function.
- daffodil-io/src/test/scala/org/apache/daffodil/io/TestInputSourceDataInputStream8.scala: Remove deprecated functions.
- daffodil-japi/src/main/java/org/apache/daffodil/japi/logger/LogLevel.java: Delete file as nothing in it is used.
- daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala: Remove deprecated functions and unused imports.
- daffodil-japi/src/main/scala/org/apache/daffodil/japi/infoset/Infoset.scala: Remove deprecated functions.
- daffodil-japi/src/main/scala/org/apache/daffodil/japi/logger/Logger.scala: Delete file as nothing in it is used.
- daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java: Rename test to be with instead of set.
- daffodil-lib/src/main/scala/org/apache/daffodil/api/Diagnostic.scala: Remove deprecated function isAtEnd.
- daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala: Rename setTunable and setTunables to be withTunable and withTunables.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala: Remove deprecated functions. Also add withDebugger and withDebugging functions.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/dsom/CompiledExpression1.scala: Remove deprecated constant.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/JsonInfosetInputter.scala: Remove deprecated this constructor. Also remove Either from default constructor as only InputStream used, remove private so default constructor is callable.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/JsonInfosetOutputter.scala: Remove dummy value and deprecated this constructor.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetInputter.scala: Remove deprecated this constructor. Also remove Either from default constructor as only InputStream used, remove private so default constructor is callable..
- daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala: Remove dummy value and deprecated this constructor.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataLoc.scala: Remove deprecated isAtEnd function.
- daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala: Remove deprecated functions and update var to val for some variables.
- daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala: Remove deprecated functions, coverage on and off comments. Also add withDebugger and withDebugging functions.
- daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/Daffodil.scala: Remove deprecated functions.
- daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/infoset/Infoset.scala: Remove deprecated functions.
- daffodil-sapi/src/main/scala/org/apache/daffodil/sapi/logger/Logger.scala: Delete file as nothing in it is used.
- daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/processor/TDMLDFDLProcessor.scala: Remove deprecated functions.
- daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala: Remove deprecated functions.
- daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala: Remove deprecated functions.
- daffodil-lib/src/main/scala/org/apache/daffodil/processors/Runtime.scala: Delete file as HasSetDebugger no longer used.
- daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala: Remove references to HasSetDebugger as it was removed.

[DAFFODIL-2743](https://issues.apache.org/jira/projects/DAFFODIL/issues/DAFFODIL-2743?filter=allissues)


daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -825,50 +524,15 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
...
 
 /**
- * Result of calling [[DataProcessor#parse(java.nio.channels.ReadableByteChannel, InfosetOutputter, long)]], containing
+ * Result of calling [[DataProcessor#parse(InputSourceDataInputStream, InfosetOutputter)]], containing
  * the diagnostic information, and the final data location
  */
 class ParseResult private[japi] (pr: SParseResult, deprecatedOutput: Maybe[JDOMInfosetOutputter])
   extends WithDiagnostics(pr) {
 
-  /**
```
stevedlawrence, line 850: This `deprecatedOutput` variable can be removed. This was originally to maintain backwards compatibility with some of the deprecated `parse()` and below `result()` functions, but with those removed this can now be removed. Same in the Scala API.

Shanedell: Should this go for any variable that has `deprecated` in its name?

stevedlawrence:
> I also want to know why we didn't see compilation warnings alerting us that these deprecated methods still were being called

I don't think we called any deprecated methods, even in unit tests. Probably not a great idea from a code coverage perspective, but they were deprecated, so maybe it wasn't super important. We current set the `-deprecation` flag in scalacOptions, so I think  deprecation warnings are treated as errors, which is why we don't have any tests calling those functions, or else they would error.

tuxji:
> We current set the `-deprecation` flag in scalacOptions, so I think deprecation warnings are treated as errors, which is why we don't have any tests calling those functions, or else they would error.

OK, I see, the compiler just didn't tell us that some javadocs or scaladocs still referred to deprecated functions.  Treating deprecation warnings as errors is fine with me.  Now the integration tests are failing and will need to be fixed as well.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
Shanedell: Is `separatorPolicy` also deprecated?

Shanedell: Should all the items in `DeprecatedProperty` be removed as well

stevedlawrence: `separatorPolicy` is a legacy property that is deprecated. Seems reasonable to remove support for all these deprecated properties as well, unless we want to do that separately. Maybe we only want to remove deprecated APIs for now since their use is likely pretty rare, and remove support for deprecated schema properties at some other point in the future since many schemas may still use these.

Shanedell: That would make sense to me because a lot of the stuff I removed already wasn't being used by anything.

Shanedell: @stevedlawrence Do you think anything other than `deprecatedOutput` and changing the default value of `allowExpressionResultCoercion` to false or do you think it is mostly schema stuff?

stevedlawrence: I think it depends what our goal here is. If the goal is to ensure we don't break existing schemas and we're only removing deprecated APIs (which could require code changes but no schema changes), then we shouldn't change this tunable--changing it may require schema changes if schemas rely on this coercion.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -145,7 +137,6 @@ class DataProcessor private (
   var validationMode: ValidationMode.Type,
   private var externalVars: Queue[Binding])
   extends DFDL.DataProcessor
-  with HasSetDebugger
```
stevedlawrence, line 148: There's a comment just above that says:
```scala
  //
  // Once the deprecated API goes away, these var can become val.
  //
```
We should probably make that change now.

Shanedell:
> Your pull request description needs a body as well as the title and the DAFFODIL-2743 line. The body should say which Scala classes had to be changed because they still had references to deprecated methods. I also want to know why we didn't see compilation warnings alerting us that these deprecated methods still were being called. I think it would be in scope for this PR to (temporarily or permanently) add a compilation option to enable deprecation warnings so we can see what other problems are there.

@tuxji Do you want the specific class names and which methods were updated, just the class names that need updated/removed or just the file name?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
Shanedell: That makes sense to me. I think it would be best to do the schema changes in another issue/PR as I feel this one is quite large enough that splitting out the other changes would make sense.

tuxji:
> Do you want the specific class names and which methods were updated, just the class names that need updated/removed or just the file name?

Shane, the best level of detail to put in a pull request is somewhat subjective.  I personally review my own pull requests one file at a time and write a brief message about each file because doing so forces me to think about why the PR changes each file and sometimes reminds me of something I may have overlooked.  Here is an example I see in your own pull request.  The first changed file is Main.scala so I would write something like...

`Main.scala: Call DataProcessor methods instead of deprecated HasSetDebugger methods.`

I see two issues immediately when I look at Main.scala.  The first issue is that when I search all the files changed for other mentions of HasSetDebugger, I don't see HasSetDebugger's definition, only that it is no longer one of DataProcessor's traits.  I know immediately that the PR has not removed HasSetDebugger's definition yet (FYI, you can find HasSetDebugger's definition in the file daffodil-lib/src/main/scala/org/apache/daffodil/processors/Runtime.scala).  The second issue I see is that you can't simply replace `setDebugg*` calls with `withDebugg*` calls.  The `setDebugg*` calls are mutative - they change the called object's state.  The `withDebugg*` calls are immutative - they return a copy of the called object with the new value instead of the old value, but the called object's state hasn't changed in any way.  This could be why your integration tests are failing.

This example shows how the very act of writing down why your PR changes each file can help you notice problems yourself.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Or at least a separate change set, so they're more easily spotted and the change backed out if need be.

Shanedell:
@tuxji Your comment makes a lot of sense to me thank you. Also with the Integration tests, is there a way I can run a single test or suite tests at a time instead of all them at once? This would just be to be to test fixes for them timing out.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
Shanedell: This will be added in the next commit

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
```diff
@@ -35,11 +35,6 @@ import org.apache.daffodil.util.Indentable
 class XMLTextInfosetOutputter private (writer: java.io.Writer, pretty: Boolean, dummy: Int)
```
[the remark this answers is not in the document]
Shanedell: Should this go for each file it is in? I am seeing this in these files

```txt
daffodil-lib/src/main/scala/org/apache/daffodil/exceptions/Assert.scala --> line 32
daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/JsonInfosetOutputter.scala --> line 28
daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala --> line 35 (item commented on)
```

mbeckerle:
I put the instructions for running just one test on a wiki page, since I had always forgotten it by the time I needed it again.

https://cwiki.apache.org/confluence/display/DAFFODIL/sbt+Tips+and+Tricks

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetOutputter.scala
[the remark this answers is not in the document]
Shanedell: I removed this from both the XML and JSON Infoset outputter's as it seemed to not cause issues but did have other effects inside of `Assert.scala`

Shanedell:
@stevedlawrence @mbeckerle @tuxji This should be good for another good look over as I fixed the issue that was causing issues with integration tests.

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/TunableGenerator.scala
```diff
@@ -91,11 +91,11 @@ class TunableGenerator(schemaRootConfig: scala.xml.Node, schemaRootExt: scala.xm
...
     |  extends Serializable {
     |
-    |  def setTunables(tunables: Map[String, String]): DaffodilTunables = {
-    |    tunables.foldLeft(this) { case (dafTuns, (tunable, value)) => dafTuns.setTunable(tunable, value) }
+    |  def withTunables(tunables: Map[String, String]): DaffodilTunables = {
+    |    tunables.foldLeft(this) { case (dafTuns, (tunable, value)) => dafTuns.withTunable(tunable, value) }
     |  }
     |
-    |  def setTunable(tunable: String, value: String): DaffodilTunables = {
+    |  def withTunable(tunable: String, value: String): DaffodilTunables = {
```
stevedlawrence, line 98: This functions aren't technically deprecated, but this rename is semi-related to other deprecated functions and is a good change--the generated `setTunable` functions are immutable and so should use the same `with*` naming convention :+1:

daffodil-runtime1/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesLoader.scala
```diff
@@ -46,7 +46,7 @@ object ExternalVariablesLoader {
 
   def loadVariables(bindings: Seq[Binding], referringContext: ThrowsSDE, vmap: VariableMap): VariableMap = {
     Assert.usage(referringContext != null, "loadVariables expects 'referringContext' to not be null!")
-    VariableUtils.setExternalVariables(vmap, bindings, referringContext)
+    VariableUtils.withExternalVariables(vmap, bindings, referringContext)
     vmap
```
stevedlawrence, line 50: Does `withExternalVariable` mutate the vmap, or does it create a copy? I assume it creates a copy with the new values, so I would expect this to be something like:
```scala
val newVMap = VariableUtils.withExernalVariables(...)
newVMap
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetInputter.scala
```diff
@@ -191,9 +191,6 @@ object XMLTextInfoset {
 class XMLTextInfosetInputter private (input: Either[java.io.Reader, java.io.InputStream])
```
stevedlawrence, line 191: Can this remove the `Either`? Seems like we only need the `InputStream` now?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
```diff
@@ -140,7 +140,7 @@ class VariableInstance private (val rd: VariableRuntimeData)
 
 object VariableUtils {
 
-  def setExternalVariables(currentVMap: VariableMap, bindings: Seq[Binding], referringContext: ThrowsSDE) = {
+  def withExternalVariables(currentVMap: VariableMap, bindings: Seq[Binding], referringContext: ThrowsSDE) = {
     bindings.foreach { b => currentVMap.setExtVariable(b.varQName, b.varValue, referringContext) }
```
stevedlawrence, line 144: This calls `setExternalVariable` which looks like it mutates the variable map. Because of that this should probably stay as `setExternalVariables` to make it clear this mutates state.

This also related to another comment I made about what calls this function, that comment probably no longer applies.

Note I wonder if there is a bug...do our API `withExternalVariables` functions in the DataProcessor end up calling this `setExternalVariables` function and mutate state instead of copying? We may need to investigate that, but that can be done separately.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -117,6 +116,7 @@ import org.apache.daffodil.xml.QName
 import org.apache.daffodil.xml.RefQName
 import org.apache.daffodil.xml.XMLUtils
 import org.apache.daffodil.xml.DFDLCatalogResolver
+// import org.apache.daffodil.processors.HasSetDebugger
```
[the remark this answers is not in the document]
tuxji: I would go further and remove the source file defining HasSetDebugger:

daffodil-lib/src/main/scala/org/apache/daffodil/processors/Runtime.scala

Once nothing uses a piece of code anymore, it becomes dead code and should be removed.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -630,23 +630,25 @@ object Main {
...
-        if (conf.trace()) {
-          new TraceDebuggerRunner(STDOUT)
-        } else {
-          if (System.console == null) {
-            Logger.log.warn(s"Using --debug on a non-interactive console may result in display issues")
-          }
-          conf.debug() match {
-            case Some(f) => new CLIDebuggerRunner(new File(f), STDIN, STDOUT)
-            case None => new CLIDebuggerRunner(STDIN, STDOUT)
+  def setupDebugOrTrace(proc: DataProcessor, conf: CLIConf): Option[DataProcessor] = {
```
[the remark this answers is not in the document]
tuxji: I would have suggested the same change too, thanks, Steve.

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -971,7 +971,7 @@ public void testJavaAPI21() throws IOException, ClassNotFoundException {
     }
 
     @Test
-    public void testJavaAPI22_setExternalVariablesUsingAbstractMap() throws IOException, ClassNotFoundException, ExternalVariableException {
+    public void testJavaAPI22_withExternalVariablesUsingAbstractMap() throws IOException, ClassNotFoundException, ExternalVariableException {
```
tuxji, line 974: Good catch.

daffodil-runtime1/src/main/scala/org/apache/daffodil/externalvars/ExternalVariablesLoader.scala
tuxji: As Steve noticed later on, `VariableUtils.setExternalVariables` was renamed to `VariableUtils.withExternalVariables` but it still mutates the variable map so this renaming change should be reverted.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/VariableMap1.scala
tuxji: I checked daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala lines 297, 308, and 332 in the main branch.  They all have similar bodies:

```scala
     val newBindings = loadExternalVariables(extVars)
     copy(externalVars = newBindings)
```

`loadExternalVariables` is defined in the same file and it returns a new data structure, so the `withExternalVariables` methods do not mutate state instead of copying.  There is no need for concern.

tuxji (commented):
More changes needed, but we are getting closer to the end.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetInputter.scala
Shanedell: Removing the `Either` seems to cause issue with the `this` constructor. I can try to get that worked out for both JSON and XML infoset inputters if desired.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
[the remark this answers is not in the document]
Shanedell: @tuxji I did not delete `HasSetDebugger` as it still is used by `daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala`. The `setDebugger` and `setDebugging` methods were set as deprecated here that is why they weren't delete but this can be done if it is something we want to do.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetInputter.scala
Shanedell: @stevedlawrence To get this to work the `private` keyword need deleted for example for `JsonInfosetInputter` I needed to make the class def be this
```scala
class JsonInfosetInputter(input: java.io.InputStream)
```
instead of having
```scala
class JsonInfosetInputter private (input: java.io.InputStream)
```
would this be how we wanted to do it?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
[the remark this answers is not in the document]
tuxji: I looked at TDMLRunner.scala and I don't think it has any good reason why DFDLTestSuite needs to extend HasSetDebugger.  If you remove the import, the extends HasSetDebugger, and the override keyword from the `setDebugger` and `setDebugging` methods and you still can build the code and run the tests without errors, we'll know for sure.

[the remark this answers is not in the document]
Shanedell: Makes sense, doing this worked fine for the tests.

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/XMLTextInfosetInputter.scala
stevedlawrence: Makes sense, we now only need the one constructor so it can be made public.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -929,7 +930,7 @@ object Main {
 
             processor = processor.withValidationMode(validate)
 
-            setupDebugOrTrace(processor.asInstanceOf[DataProcessor], conf)
+            val debugProcessor = setupDebugOrTrace(processor.asInstanceOf[DataProcessor], conf)
```
tuxji, line 933: In this scope, processor is a mutable var so you may not need `debugProcessor`.  Perhaps line 933 can replace processor exactly like line 931 already replaces processor:

`processor = setupDebugOrTrace(processor.asInstanceOf[DataProcessor], conf)`

That change should work unless the compiler complains about types, and it also makes me think the method name might look better if it was renamed to `withDebugOrTrace`.

There's a region further down where processor is a val, so debugProcessor is really needed there.  Mike, Steve, would you prefer that both regions use a mutable var processor or an immutable val processor and a val debugProcessor?  I think it's best practice to use val and careful variable naming to make it clear a value is changing, but var is probably fine in a small lexical scope especially when processor is replaced 2 times, not just 1 time.

tuxji (approved):
+1

Changes look good now, although I noticed one region of Main.scala uses a mutable var processor.  One could take advantage of that mutability or leave it alone; I'll let others weigh in on what should be best practice.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
stevedlawrence: Yeah, the var approach seems reasonable to me in this case. Especially since `debugProcessor` might not even have debugging enabled depending on the provide conf options, so it could potentially be confusing. It's more like `maybeDebugProcessor` which is ugly and still confusing.

An alternative, though maybe a bit more invasive, might be to do what we do with the performance command, which looks like this:

```scala
        val processor = {
          if (performanceOpts.parser.isDefined) {
            createProcessorFromParser(performanceOpts.parser(), performanceOpts.path.toOption, validate)
          } else {
            val tunables = DaffodilTunables.configPlusMoreTunablesMap(performanceOpts.tunables, optDafConfig)
            createProcessorFromSchema(performanceOpts.schema(), performanceOpts.rootNS.toOption, performanceOpts.path.toOption, tunables, validate)
          }
        }.map{ _.withExternalVariables(combineExternalVariables(performanceOpts.vars, optDafConfig)) }
         .map{ _.withValidationMode(validate) }
```
We could just add a `.map { withDebugOrTrace(_, conf) }` to the end of that.

This way all processor setup happens in one spot for each subcommand, the processor variable can always be a val, and is consistent among the different subcommands.

Shanedell: I will add the `.map { withDebugOrTrace(_, conf) }`. But with that should I also remove the line that does `var processor = proc` and just use `proc` instead?

stevedlawrence: Makes sense, that would use the same patern as the unparse and performance subcommands

Shanedell: So `processor` would now need to be `Option[DataProcessor]` instead of `Option[DFDL.DataProcessor]` right?

Shanedell: So I was able to get it work like this:

```scala
val processor: Option[DataProcessor] = {
  if (parseOpts.parser.isDefined) {
    createProcessorFromParser(parseOpts.parser(), parseOpts.path.toOption, validate)
  } else {
     val tunables = DaffodilTunables.configPlusMoreTunablesMap(parseOpts.tunables, optDafConfig)
      createProcessorFromSchema(parseOpts.schema(), parseOpts.rootNS.toOption, parseOpts.path.toOption, tunables, validate)
  }
}.map{ _.withExternalVariables(combineExternalVariables(parseOpts.vars, optDafConfig)) }
 .map{ _.withValidationMode(validate) }
 .map{ proc => withDebugOrTrace(proc.asInstanceOf[DataProcessor], conf) }
```

Doing `.map{ withDebugOrTrace(_.asInstanceOf[DataProcessor], conf) }` or `.map{ withDebugOrTrace(_, conf) }` caused errors. Is this what would be desired or would you want me to add `withDebugger` defintions to `DataProcessorBase` trait that you just mentioned?

stevedlawrence: I have no strong preference. It might be cleaner if the withDebugger defs were added to the trait though. I'm not sure if there was a reason for those to not be added originally. Might have just been an oversight.

Shanedell: So when I add them in I still need to do the `.asInstanceOf[DataProcessor]`, is there something I was missing because up until `withDebugOrTrace` I believe that processor is treated as a `DFDL.DataProcessor` right? So that would always be needed unless converting other methods to switch the type of `DataProcessor` being set right?

stevedlawrence: You might need to change `withDebugOrTrace` to accept a `DFDL.DataProcessor` instead of a `DataProcessor`.

Shanedell: Oh okay I think I got will push that up once I get it fully tested locally to make sure it works fully.

Shanedell: @tuxji @stevedlawrence The rename of `setupDebugOrTrace` to `withDebugOrTrace`, adding `withDebugger` method to`DataProcessorBase` and updating the way the processor is set is up if you want to look it over.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -128,24 +120,20 @@ object DataProcessor {
...
   // because this object itself isn't serialized. A SerializableDataProcessor is.
   // The values these will have (since this is a base class) are the correct default values that we want
   // back when the object is re-initialized.
   //
-  protected var areDebugging : Boolean,
-  protected var optDebugger : Option[Debugger],
+  protected val areDebugging : Boolean,
+  protected val optDebugger : Option[Debugger],
   var validationMode: ValidationMode.Type,
   private var externalVars: Queue[Binding])
```
stevedlawrence, line 135: Can `validationMode` and `externalVars` become  a val too like above? We don't have the set* functions anymore so nothing should mutate this I think?

stevedlawrence (approved):
+1 :+1: 

Just one comment about some more parameters that I *think* can become vals.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
stevedlawrence: Right above this we do `val dpToSave = dp.copy()`. We could change that to something like his:
```scala
val dpToSave = this.copy(
  externalVars = Queue.empty[Binding],
  validationMode = ValidationMode.Off
)
```
That's a fairly small change and I think makes our DataProcessor fully immutable.

Shanedell: That makes sense I will throw that in

Shanedell: @stevedlawrence That look good? So since I have two approvals, I should wait till the CI passed, then squash my commits, push up the squashed commit and then merge once CI passes?

tuxji: Yes, all the latest changes look good (thanks, Shane and Steve, for adding them).  Yes, CI has passed so you can squash your commits, push a single squashed commit, and merge (if you have write permission).

tuxji: Yes, the commit message can reuse your nicely written PR description (updated for the latest changes).  That would be very nice for git history.

PR #877, opened 2022-11-18
Remove leftover deprecated variables and methods

Shanedell:
Remove leftover deprecated variables and methods.

- daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala: Remove deprecated variable compilerExternalVarSettings and method withExternalDFDLVariablesImpl. Update some var's to val's
- daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSet.scala: Remove deprecated variable compilerExternalVarSettings.
- daffodil-core/src/main/scala/org/apache/daffodil/runtime1/SchemaSetRuntime1Mixin.scala: Remove ExternalVariablesLoader and references to compilerExternalVarSettings.
- daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/processor/TDMLDFDLProcessor.scala: Remove deprecated method withExternalDFDLVariables.
- daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/DaffodilTDMLDFDLProcessor.scala: Remove deprecated method withExternalDFDLVariables.
- daffodil-tdml-processor/src/main/scala/org/apache/daffodil/tdml/processor/Runtime2TDMLDFDLProcessor.scala: Remove deprecated method withExternalDFDLVariables.

[DAFFODIL-2743](https://issues.apache.org/jira/projects/DAFFODIL/issues/DAFFODIL-2743?filter=allissues)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
@@ -26,7 +26,6 @@ import java.nio.channels.Channels
 import java.util.zip.GZIPInputStream
 import java.util.zip.ZipException
 
-import scala.collection.immutable.Queue
 import scala.collection.mutable.ArrayBuffer
 import scala.util.Try
 import scala.xml.Node
@@ -63,17 +62,7 @@ object ForUnparser extends ParserOrUnparser
 object BothParserAndUnparser extends ParserOrUnparser
 
 final class ProcessorFactory private(
-  private var optRootSpec: Option[RootSpec],
-  /*
-   * compilerExternalVarSettings supports the deprecated API
-   * where external variable settings can be supplied to the compiler
-   * instance.
-   *
-   * The non-deprecated API is to deal with external variable settings
-   * on the data processor instance. This is passed here so that
-   * the PF can propagate it to the DP.
-   */
-  var compilerExternalVarSettings: Queue[Binding],
+  private val optRootSpec: Option[RootSpec],
   schemaSource: DaffodilSchemaSource,
   val validateDFDLSchemas: Boolean,
   checkAllTopLevel: Boolean,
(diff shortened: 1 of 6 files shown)


Shanedell:
@tuxji @stevedlawrence These changes should removing the deprecated items that were inside of `Compiler.scala` and then traversed over another couple of files.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -144,14 +130,6 @@ class InvalidParserException(msg: String, cause: Throwable = null) extends Excep
 
 class Compiler private (var validateDFDLSchemas: Boolean,
   var tunables : DaffodilTunables,
```
stevedlawrence, line 132: Can we remove the `var`s here too?

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -64,16 +63,6 @@ object BothParserAndUnparser extends ParserOrUnparser
 
 final class ProcessorFactory private(
   private var optRootSpec: Option[RootSpec],
```
stevedlawrence, line 65: Can we remove this `var`, similar to what we did with the `DataProcessor`?

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
```diff
@@ -144,14 +130,6 @@ class InvalidParserException(msg: String, cause: Throwable = null) extends Excep
...
 class Compiler private (var validateDFDLSchemas: Boolean,
   var tunables : DaffodilTunables,
-  /*
-   * Supports deprecated feature of establishing external vars on the compiler object.
-   * These are just saved and passed to the processor factory which incorporates them into
-   * the variable map of the data processor.
-   *
-   * This argument can be removed once this deprecated feature is removed.
-   */
-  private var externalDFDLVariables: Queue[Binding],
```
stevedlawrence, line 154: Right below this are three private `var`'s that I think can also become `val`s?

stevedlawrence (approved):
+1 :+1:  might be a couple var's that can also become vals like we did in the previous PR

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
Shanedell: Remove it completely or change it to `val`?

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
Shanedell: When you say remove do you want them removed completely or see if they can be `val`?

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
stevedlawrence: Yeah, change to a val. My mistake, that wasn't very clear.

daffodil-core/src/main/scala/org/apache/daffodil/compiler/Compiler.scala
stevedlawrence: In these cases, we can just remove the `var` keyword and scala treats them as private vals. We still need to keep them as constructor parameters.

Shanedell: Okay that makes sense, so for these ones you had asked about should I remove the `private` keyword or keep it for ones that have it?

stevedlawrence: I believe `private var` can be removed and it is implied to be private val.

Shanedell: Makes sense, do the changes just pushed look good?

stevedlawrence: Looks reasonable to me :+1:

tuxji (approved):
+1
👍

PR #879, opened 2022-11-18
Improve SAX parse/unparse performance

stevedlawrence:
Four changes are made to improve SAX unparse performance. For this PR, they commits are separated for easier review. The changes are:

1. Use a cached thread pool for coroutine threads. Instead of creating a new thread everytime a new coroutine is created (i.e. everytime we unparse something with SAX), coroutines now get threads from a cached thread pool. This allows reuse of previously created threads, avoiding the overhead of creating threads.

2. Initialize SAX prefixMapping with TopScope instead of null. Avoids some null checks and throw/catch of NullPointerExceptions.

3. Store the prefix string in ElementRuntimeData instead of a def. Looking up the prefix shows up in profiling. There's no need to relook it up everytime, so just store it as a lazy val.

4. Replace match case with reference equality. Type matching uses instanceof, which has some amount of overhead
compared to straight reference equality. Also reorder the anyURI condition so the resolve uri condition is first, since it is almost always going to be false short circuit the condition.

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/util/Coroutines.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/util/Coroutines.scala
@@ -20,9 +20,38 @@
  import org.apache.daffodil.exceptions.UnsuppressableException
 
  import java.util.concurrent.ArrayBlockingQueue
+ import java.util.concurrent.Executors
+
+ import scala.concurrent.ExecutionContext
+ import scala.concurrent.Future
  import scala.util.Failure
  import scala.util.Success
  import scala.util.Try
+
+ object Coroutine {
+   /**
+    * This execution context should be used when creating Coroutine threads to
+    * improve performance. Creating threads has high overhead, but getting
+    * threads from a thread pool we reduce some of that overhead.
+    *
+    * This thread pool is a cached thread pool, which means unused threads are
+    * closed after 60 seconds of unuse. This also places no limits on the
+    * number of threads that will be created, so if many coroutines are created
+    * at the same time we could potentially starve the system of
+    * threads/processing. However, we only create one thread per coroutine, and
+    * we currently only create one coroutine per SAX unparse, and it is
+    * hopefully unlikely that a user will create enough parallel SAX unparse
+    * calls to cause issues.
+    */
+   val executionContext = new ExecutionContext {
+     private val threadPool = Executors.newCachedThreadPool()
+     def execute(runnable: Runnable): Unit = threadPool.submit(runnable)
+
+     //$COVERAGE-OFF$
+     def reportFailure(t: Throwable): Unit = {} //do nothing
+     //$COVERAGE-ON$
+   }
+ }
 
  /**
   * General purpose Co-routines.
@@ -55,15 +84,14 @@
     */
    protected def isMain: Boolean = false
 
-   private var thread_ : Option[Thread] = None
+   private var thread_ : Option[Future[Unit]] = None
 
    private final def init(): Unit = {
      if (!isMain && thread_.isEmpty) {
-       val thr = new Thread {
-         override def run(): Unit = self.run()
-       }
+       val thr = Future {
+         self.run()
+       } (Coroutine.executionContext)
        thread_ = Some(thr)
-       thr.start()
      }
    }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
@@ -18,6 +18,7 @@
 package org.apache.daffodil.processors
 
 import scala.xml.NamespaceBinding
+import scala.xml.TopScope
 
 import javax.xml.XMLConstants
 import org.apache.daffodil.api.DFDL
@@ -32,6 +33,7 @@ import org.apache.daffodil.infoset.InfosetInputterEventType.StartDocument
 import org.apache.daffodil.infoset.InfosetInputterEventType.StartElement
 import org.apache.daffodil.infoset.SAXInfosetInputter
 import org.apache.daffodil.util.MStackOf
+import org.apache.daffodil.util.Maybe
 import org.apache.daffodil.util.Maybe.Nope
 import org.apache.daffodil.util.Maybe.One
 import org.apache.daffodil.util.Misc
@@ -75,7 +77,7 @@ class DaffodilUnparseContentHandler(
   private lazy val inputter = new SAXInfosetInputter(this, dp, output)
   private var unparseResult: DFDL.UnparseResult = _
   private lazy val characterData = new StringBuilder
-  private var prefixMapping: NamespaceBinding = _
+  private var prefixMapping: NamespaceBinding = TopScope
   private lazy val prefixMappingTrackingStack = new MStackOf[NamespaceBinding]
 
   private lazy val tunablesBatchSize = dp.tunables.saxUnparseEventBatchSize
@@ -158,7 +160,7 @@ class DaffodilUnparseContentHandler(
    * is already done using the prefixMappings
    */
   override def endPrefixMapping(prefix: String): Unit = {
-    prefixMapping = if (prefixMapping == null) prefixMapping else prefixMapping.parent
+    prefixMapping = prefixMapping.parent
   }
 
   /**
@@ -324,14 +326,8 @@ class DaffodilUnparseContentHandler(
     val maybeNamespaceURI =
       if (uri.nonEmpty) {
         One(uri)
-      } else if (prefixMapping != null) {
-        try {
-          One(prefixMapping.getURI(qNamePrefix))
-        } catch {
-          case _: NullPointerException => Nope
-        }
       } else {
-        Nope
+        Maybe(prefixMapping.getURI(qNamePrefix))
       }
 
     batchedInfosetEvents(currentIndex).localName = maybelocalName
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/SAXInfosetInputter.scala
@@ -96,15 +96,14 @@ class SAXInfosetInputter(
     } else {
       throw new NonTextFoundInSimpleContentException(getLocalName())
     }
-    primType match {
-      case _: NodeInfo.String.Kind =>
-        val remapped = XMLUtils.remapPUAToXMLIllegalCharacters(res)
-        remapped
-      case _: NodeInfo.AnyURI.Kind if resolveRelativeInfosetBlobURIs && res.nonEmpty =>
-        val absUri = resolveRelativeBlobURIs(res)
-        absUri
-      case _ =>
-        res
+    if (primType eq NodeInfo.String) {
+      val remapped = XMLUtils.remapPUAToXMLIllegalCharacters(res)
+      remapped
+    } else if (resolveRelativeInfosetBlobURIs && (primType eq NodeInfo.AnyURI) && res.nonEmpty) {
+      val absUri = resolveRelativeBlobURIs(res)
+      absUri
+    } else {
+      res
     }
   }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/RuntimeData.scala
@@ -632,9 +632,9 @@ sealed class ElementRuntimeData(
 
   def isComplexType = !isSimpleType
 
-  def prefix = this.minimizedScope.getPrefix(namedQName.namespace)
+  lazy val prefix = this.minimizedScope.getPrefix(namedQName.namespace)
 
-  def prefixedName = {
+  lazy val prefixedName = {
     if (prefix != null) {
       prefix + ":" + name
     } else {


stevedlawrence:
Note that PR #873 needs to be merged before we can really get accurate numbers about how much this improves things.

I think there is still alot of overhead related to the threads need to connect the SAX push-style API with our Daffodil unparse pull-style API, but this seems to have reduced it so unparse is a bit closer to other infoset inputters.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+    //$COVERAGE-OFF$
+    def reportFailure(t: Throwable): Unit = {} //do nothing
+    //$COVERAGE-ON$
+  }
+}
+
+/**
+ * DaffodilUnparseContentHandler produces SAXInfosetInputterEvent objects for
+ * the SAXInfosetInputter to consume and convert to events that the
+ * Dataprocessor unparse() can use. The SAXInfosetInputterEvents are built base
```
tuxji, line 72: base -> based

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+ * on information that is passed to this ContentHandler from an XMLReader.
+ *
+ * This class runs the DataProcessor unparse() method in a separate thread,
+ * with a shared ArrayBlockigQueue to provide SAXInfosetInputerEvents. This
```
tuxji, line 76: ArrayBlockigQueue -> ArrayBlockingQueue, SAXInfosetInputerEvents -> SAXInfosetInputterEvents

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+ * queue has a maximum size defined by the saxUnaprseEventBatchSize tunble
```
tuxji, line 77: saxUnaprseEventBatchSize -> saxUnparseEventBatchSize

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
-   * user provides 1 as the batchSize, under the hood we'll have [event1] batched.
-   *
-   * DataProcessor.unparse will call hasNext and getEventType for the initialization call, and that
-   * hasNext will check if cnextIndex (i.e currentIndex + 1) is non-empty. As currentIndex is 0, and
-   * it is the maximum index, there is no index 1. It will context switch to get a new batched event,
-   * which, would overwrite event1 before we get to process it.
+   * To avoid allocating many SAXInfosetInputterEvents, we instead pre-allocate
+   * events and clear/mutate each event as needed. Note that we need 2 more
+   * pre-allocated events than the queue size. This is because while
+   * processing, there could be a most eventQueueSize events already on the
```
tuxji, line 136: a -> at

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
 
   /**
-   * This is a flag that is set to true when startPrefixMapping is called. When true, we make
-   * the assumption that we don't need to use the Attributes parameter from startElement to get the
-   * namespaceURI information and will solely rely on start/endPrefixMapping. If false, we will use
-   * Attributes to get the namespaceURI info.
+   * Index into the pre-allocated event array that the content handler is
+   * currently mutating and preparing to put() on the eventQueue(). When all
+   * information for this event is gathered, we put() it on the eventQueue and
+   * increment the index into the pre-allocated events array and start muating
```
tuxji, line 149: muating -> mutating

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+   * Add the current event to the eventQueue. If the eventQueue is full, this
+   * blocks until the SAXInfosetInputter take()'s an event and frees up a slot
+   * in the queue. Once the event is added, we update state to point to the
+   * next SAXInfosetInputterEvent that we should mutate as we receive more SAX
+   * events.
+   */
+  private def addCurrentEventToQueue(): Unit = {
+    // before we put() an event, check to make sure there hasn't been an error
+    // from the SAXInfosetInputter. If there is an error, checkUnparseResult()
+    // throws and exception which bubbles up to the XMLReader to handle. Note
```
tuxji, line 203: and -> an

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+        //$COVERAGE-ON$
+      }
+
+      // We are finished unparsing. At this point, it is possible that the
+      // ContentHandler thread is blocked trying to put() an event on the
+      // eventQueue, but the SAXInfosetInputter thread could have failed in a
+      // way where it won't ever take() an item from the queue, leaving the
+      // ContentHandler deadlocked. Now that unparseResult is defined, we can
+      // clear eventQueue which unblocks the ContentHandler. And since
+      // unparseResult is set, the ContentHandler should not attempt anymore
```
tuxji, line 273: anymore -> any more

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala
```diff
@@ -17,327 +17,486 @@
...
+   * This function should be called whenever mixed content (i.e non-whitespace
+   * characters) should be checked for and, since it isn't allowed in a DFDL
+   * infoset, state mutated to indicate the error.
+   *
+   * Note that if mixed content is detected we cannot simply throw a
+   * SAXException. This is because if this ContentHandler thread finishes, the
+   * unparseTask is going to eventually consume all events from the eventQueue
+   * and block trying to take() an event that will never come, leaving a
+   * hanging thread. Instead, we set the mixedContent value of the current
+   * event to the characterData we have accumulated, and the SAXInfosetInputer
```
tuxji, line 376: SAXInfosetInputer -> SAXInfosetInputter

tuxji (approved):
+1, looks good

mbeckerle (approved):
+1 If this change improves performance over the co-routine stuff then I'm good with it.

However, single-thread performance tests should be single-thread, and we should be striving for a way to force it to be single-thread only with no overlap of the producer and consumer of the unparser event queue, as it is very hard to measure pure overhead without that.

stevedlawrence:
> Can this be set small enough as to enforce sequential behavior, i.e., no parallelism between the caller of the content handler and the unparser?

With this approach, no. Even with the batch size tunable is set to 1, the ContentHandler and SAXInfosetInputter can (and likely will) both do work a the same time. The ContentHandler will be preparing the next event while the SAXInfosetInputter is unparsing using the current event.

> To reduce overhead, we need to enqueue many events before context switching and allowing the unparser to run. Arguably, we should just queue up events to some max count, or until we get endDocument. For small messages we would then get exactly one context switch per message.

Some of the changes here weren't specific to the ArrayBlockingQueue approach (e.g. thread pool reuse, split() removal). I can apply them to the current coroutine approach and see how it compares. These changes definitely had a big speed up, but I'm not sure which individual changes had the biggest effect.

> I continue to be of the opinion that overlap parallelism here is not an advantage. It just muddies the waters about timing and overhead of unparsing.

I think one potential advantage of this parallel approach is if the incoming sax events are sporadic or relatively slow (e.g. serialized over a network/diode). With the coroutine approach, we won't attempt to do any unparsing until we get a full batch of events (or we reach the endDoc event). And if the batch size is set to something large to avoid context switching, it might mean a lot of waiting doing nothing until we get those events. With this parallel approach, we can at least start unparsing immediately and do work in the time waiting for those SAX events. Though, I'm not sure how likely that is with SAX so maybe it's not worth really considering. And maybe if that is the case, tuning the batch size to a small number is the right approach.

mbeckerle:
> I think one potential advantage of this parallel approach is if the incoming sax events are sporadic or relatively slow (e.g. serialized over a network/diode). With the coroutine approach, we won't attempt to do any unparsing until we get a full batch of events (or we reach the endDoc event). And if the batch size is set to something large to avoid context switching, it might mean a lot of waiting doing nothing until we get those events. 

While some applications (large data streaming) may be able to benefit from this, I know of at least one Daffodil cyberian implementation where the XML will be presented to Daffodil all at once, as a single buffer of XML data, to be unparsed.  In that case, any context switching is effectively just added overhead unless there are idle CPUs and idle memory bandwidth on the system (which I consider unlikely).

tuxji:
This PR has received 2 approvals and passed all CI checks but now has a conflicting file.  @stevedlawrence, please resolve the conflict and merge the PR or rework/close the PR if you don't want to merge it in its current state.

stevedlawrence:
> please resolve the conflict and merge the PR or rework/close the PR if you don't want to merge it in its current state.

Based on @mbeckerle's comment, I've done a bit more testing and profiling of each individual change in this PR and I think the removal of the coroutines are a bit slower. I'll update this PR once i get a better grasp of which changes really do help with performance and which don't. I'll convert this PR to a draft and update once it's ready.

stevedlawrence:
I've updated this PR with just the changes that I've found makes a noticeable improvement. The biggest improvement is using a thread pool so we reuse our coroutine threads where possible, but the other three commits did make a slight improvement on average. I've also update the PR description.

tuxji (approved):
+1

Additional comments may be needed to maintain safeguards against denial-of-service attacks.

daffodil-lib/src/main/scala/org/apache/daffodil/util/Coroutines.scala
```diff
@@ -20,10 +20,25 @@
...
+ import scala.concurrent.ExecutionContext
+ import scala.concurrent.Future
  import scala.util.Failure
  import scala.util.Success
  import scala.util.Try
 
+ object Coroutine {
+   val executionContext = new ExecutionContext {
+     private val threadPool = Executors.newCachedThreadPool()
+     def execute(runnable: Runnable): Unit = threadPool.submit(runnable)
```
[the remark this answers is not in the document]
stevedlawrence: > How is the thread pool eventually closed, or does this executionContext stay around as long as the program keeps running?

Yep, the JavaDoc mentions the same 60 second inactivity before throwing away a thread. So in low activity environments the threads are automatically discarded. We lose the performance gains, but if it's low activity I would think lower performance would be fine.

>  How does Daffodil guarantee that only a small number of coroutine threads are used at any time?

Is this a guarantee we want to make?

There is only ever going to be one thread per active call to SAX unparse. So the only way to hit the INT_MAX limit is to invoke INT_MAX SAX unparses at the same time in parallel. I imagine trying to do that is going to cripple the system regardless of whether we're using coroutines/SAX or not.

I think we just rely on the user to limit how many parallel unparse calls they think is reasonable, and we'll spawn one thread for each of those for our thread pool. If those users don't limit it any way and allow INT_MAX parallel unparses, then that's on them and we'll do our best to keep up.

If we wanted, we could add allow callers to pass in an execution context to the `newContentHandlerInstance` function, which would give them control over how threads are spawned. I'm not against this, but I'm not sure if people would ever really use it...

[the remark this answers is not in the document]
tuxji: I didn't know we add only 1 extra thread per SAX unparse call anyway, which is only going to double the total number of threads, not increase the number of threads exponentially.  No, we don't need to allow callers to pass in an execution context.  I agree it's really on the users to limit their calls if they use Daffodil as a server and make as many SAX unparse calls as there are connections from client machines.

mbeckerle:
I don't have access to my computer today, but I wanted to look this over.

I don't understand at all why a thread pool should make any difference and
if using a thread pool does matter it suggests we are doing something
terribly wrong. We should create the coroutine thread exactly once.

stevedlawrence:
> We should create the coroutine thread exactly once.

Every new instance of the DaffodilUnparseContentHandler is a new coroutine and so will create a new thread. This is needed since a user could different SAX unparse calls in parallel and so they would need mutiple DaffodilUnparseContentHandlers, each with their own thread.

And in the CLI performance command, each call to unparse creates a new DaffodilUnparseContentHandler, and thus a each uparse will create a new thread.

stevedlawrence:
> I thought that when an event was queued for the peer co-routine, the main thread then would read from the reply queue, hence blocking itself until the peer replies back to it.
>
> That was the notion for how the two coroutines never overlap in doing work.
>
> Is it not functioning that way now?

You are correct, this is how our coroutine implementation works.

But because java doesn't support native coroutines, the non-main coroutine executes in a separate thread. So even though there is no real parallelism, we do still have to create threads. This PR attempts to reduce the overhead of creating those threads by reusing them from a thread pool.

mbeckerle (approved):
+1 Re-approving

PR #883, opened 2022-11-29
Add macOS runner

pkatlic:
Update code to support building on macOS Monterey. Rename strtonum function to avoid conflict with built-in standard function definition on macOS. Add macOS GitHub runner and concurrency control to cancel in-progress GitHub jobs. This was tested on macOS 12.6.1.

DAFFODIL-2747

main.yml: Add macOS runner and dependency installation step. Add concurrency control to cancel in-progress GitHub jobs.

xml_reader.c: Rename strtonum function to strtoinum in order to avoid naming conflict with macOS functions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
 
   # Build Daffodil and run some checks.
@@ -38,6 +42,13 @@ jobs:
         scala_version: [ 2.12.17 ]
         os: [ ubuntu-20.04, windows-2019 ]
         include:
+          - os: macos-12
+            shell: bash
+            distribution: temurin
+            java_version: 17
+            scala_version: 2.12.17
+            env_cc: cc
+            env_ar: ar
           - os: ubuntu-20.04
             shell: bash
             env_cc: clang-10
@@ -69,6 +80,10 @@ jobs:
       ############################################################
       # Setup
       ############################################################
+
+      - name: Install Dependencies (macOS)
+        if: runner.os == 'macOS'
+        run: brew install libmxml
 
       - name: Install Dependencies (Linux)
         if: runner.os == 'Linux'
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -24,6 +24,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
```
stevedlawrence, line 29: What does this do? It's not totally clear to me from the GitHub documentation.

stevedlawrence (approved):
+1 thanks! :+1:

.github/workflows/main.yml
```diff
@@ -38,6 +42,13 @@ jobs:
         scala_version: [ 2.12.17 ]
         os: [ ubuntu-20.04, windows-2019 ]
         include:
+          - os: macos-12
+            shell: bash
+            distribution: temurin
+            java_version: 17
+            scala_version: 2.12.17
+            env_cc: cc
+            env_ar: ar
```
tuxji, line 51: FYI, the reason why we used include was because we couldn't add macos-12 to `os: [ ubuntu-20.04, windows-2019 ]` without also getting `java_version: [ 8, 11, 17 ]` combinations as well, and we found out the rest of the variables would be empty if we didn't define them in include as well.

However, I've just noticed that we can use [exclude](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude) which could be more concise (we'd need only 4 exclude lines instead of 7 include lines, although we'd have to define env_cc, env_ar, shell in the matrix itself too).  I guess it's a matter of taste, but here goes anyway:

```yaml
      matrix:
        os: [ macos-12, ubuntu-20.04, windows-2019 ]
        distribution: [ temurin ]
        java_version: [ 8, 11, 17 ]
        env_cc: [ clang ]
        env_ar: [ ar ]
        scala_version: [ 2.12.17 ]
        shell: [ bash ]
        exclude:
          - os: macos-12
            java_version: 8
          - os: macos-12
            java_version: 11
        include:
          - os: ubuntu-20.04
            env_ar: llvm-ar-10
          - os: windows-2019
            env_ar: llvm-ar
            shell: msys2 {0}
```

We can go with the above if it looks better; it does seem to be more concise.

tuxji (approved):
+1

I thought of some improvements to the workflow, see below.  We can make these improvements if they look better, thoughts?

.github/workflows/main.yml
pkatlic: Thanks for looking into this, will make the suggested change.

PR #885, opened 2022-12-02
Add infosetWalkerSkipMin/Max tunables

stevedlawrence:
Allows configuration of how often the InfosetWalker skips walk() calls. Setting to a value of zero disables skipping walk() calls. A value of zero is useful during testing to ensure the InfosetWalker isn't incorrectly removing infoset elements, which can lead to seemingly random null pointer exceptions.

Add multiple tests to show that the InfosetWalker currently removes infoset elements that are still needed by separator logic, that this does not arise in a particular case with the default skip min/max values, and that disabling infoset removal works around the issue.

DAFFODIL-2755

Note, this does not fix 2755, just makes it easier to create a reproducible test to trigger the issue.

### DIFF
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
@@ -207,6 +207,40 @@
             </xs:restriction>
           </xs:simpleType>
         </xs:element>
+        <xs:element name="infosetWalkerSkipMin" default="32" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              Daffodil periodically walks the internal infoset to send events to the configured
+              InfosetOutputter, skipping at least this number of walk attempts. Larger values
+              mean delayed InfosetOutputter events and more memory usage; Smaller values mean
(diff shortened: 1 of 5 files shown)


daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -207,6 +207,38 @@
             </xs:restriction>
           </xs:simpleType>
         </xs:element>
+        <xs:element name="infosetWalkerSkipMin" default="32" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              Daffodil periodically walks the internal infoset to send events to the configured
+              InfosetOutputter, skipping at least this number of walk attempts. Larger values
+              mean delayed InfosetOutputter events and more memory usage; Smaller values mean
+              more CPU usage. Set this value to zero to never skip any walk attempts.
```
mbeckerle, line 216: Suggest adding phrasing that this is specifically for an advanced testing behavior, and should not need to be adjusted by users ever.

mbeckerle (approved):
+1

Suggest just adding doc that these tunables are really for an advanced test mode, not for user tuning.

tuxji (approved):
+1

PR #886, opened 2022-12-06
Update scaladocs mentioning deprecated API

tuxji:
Ensure nothing mentions the deprecated API removed in previous commits
for DAFFODIL-2743.  Also fix some IDEA nits and improve DataProcessor
constructor.

SchemaSet.scala: Simplify scaladoc to stop mentioning deprecated API.
Fix some IDEA nits in the same code block with IDEA quick fixes.

SchemaSetRuntime1Mixin.scala: Make onPath call first DataProcessor
constructor instead of now-deleted second constructor.

DataProcessor.scala: Make SerializedDataProcessor extend first
DataProcessor constructor instead of now-deleted second
constructor. Change first DataProcessor constructor from private to
public, reorder its last 4 parameters, and make its last 2 parameters
optional.  Delete second DataProcessor constructor and remove its
scaladoc which had been saying it was using deprecated
compilerExternalVars.  Make copy call first constructor with reordered
parameters.  Remove another scaladoc which had been mentioning tunables
not even passed to method anymore.

DAFFODIL-2743

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
@@ -26,7 +26,6 @@ import java.nio.channels.Channels
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
 import java.util.zip.GZIPOutputStream
-import scala.collection.immutable.Queue
 import org.apache.daffodil.Implicits._
 object INoWarn4 {
   ImplicitsSuppressUnusedImportWarning() }
@@ -84,27 +83,26 @@ class InvalidUsageException(msg: String, cause: Throwable = null) extends Except
 object DataProcessor {
 
   /**
-   * This is the DataProcessor constructed from a saved processor.
+   * This is the SerializableDataProcessor constructed from a saved processor.
    *
    * It enables us to implement restrictions on what you can/cannot do with a reloaded
    * processor versus an original one.
    *
-   * When we create one of these, it will have default values for everything
-   * settable like debuggers, debug mode.
+   * When we reload a processor, we want it to have default values for everything settable
+   * like validation mode, debug mode, and debugger.
    *
-   * Note that this does preserve the externalVars and validationMode. That is because
-   * those may be needed by serializations other than our own save/reload (e.g., Apache Spark which
-   * serializes to move things for remote execution).
+   * Note that this class does preserve variableMap and validationMode. That is because
+   * serializations other than our own save/reload may need such settings (e.g., Apache Spark
+   * which serializes to move objects for remote execution).
    *
-   * Hence, we're depending on the save method to explicitly reset validationMode and
-   * externalVars to initial values.
+   * Hence, we're depending on the save method to explicitly reset them to default values.
    */
   private class SerializableDataProcessor(
-    val data: SchemaSetRuntimeData,
-    tunable: DaffodilTunables,
-    externalVars: Queue[Binding], // must be explicitly set to empty by save method
-    validationModeArg: ValidationMode.Type) // must be explicitly set from Full to Limited by save method.
-    extends DataProcessor(data, tunable, externalVars, validationModeArg) {
+    ssrd: SchemaSetRuntimeData,
+    tunables: DaffodilTunables,
+    variableMap: VariableMap,            // must be explicitly reset by save method
+    validationMode: ValidationMode.Type, // must be explicitly turned off by save method
+  ) extends DataProcessor(ssrd, tunables, variableMap, validationMode) {
 
     override def withValidationMode(mode: ValidationMode.Type): DataProcessor = {
       if (mode == ValidationMode.Full) {
@@ -119,7 +117,7 @@ object DataProcessor {
  * The very last aspects of compilation, and the start of the
  * back-end runtime.
  */
-class DataProcessor private (
+class DataProcessor(
   val ssrd: SchemaSetRuntimeData,
   val tunables: DaffodilTunables, // Compiler-set tunables
   val variableMap: VariableMap,
@@ -129,11 +127,10 @@ class DataProcessor private (
   // The values these will have (since this is a base class) are the correct default values that we want
   // back when the object is re-initialized.
   //
-  protected val areDebugging : Boolean,
-  protected val optDebugger : Option[Debugger],
-  val validationMode: ValidationMode.Type,
-  private val externalVars: Queue[Binding])
-  extends DFDL.DataProcessor
+  val validationMode: ValidationMode.Type = ValidationMode.Off,
+  protected val areDebugging : Boolean = false,
+  protected val optDebugger : Option[Debugger] = None,
+) extends DFDL.DataProcessor
   with Serializable
   with MultipleEventHandler {
 
@@ -141,50 +138,33 @@ class DataProcessor private (
 
   /**
    * In order to make this serializable, without serializing the unwanted current state of
-   * debugger, external var settings, etc. we replace, at serialization time, this object
-   * with a [[SerializableDataProcessor]] which is a private derived class that
-   * sets all these troublesome var slots back to the default values.
+   * validation mode, debugging mode, debugger, etc. we replace, at serialization time, this
+   * object with a [[SerializableDataProcessor]] which is a private derived class that
+   * sets all these troublesome slots back to the default values.
    *
    * But note: there is serialization for us to save/reload, and there is serialization
    * in other contexts like Apache Spark, which may serialize objects without notifying us.
    *
-   * So we preserve everything that something like Spark might need preserved (validation modes, external vars)
+   * So we preserve everything that something like Spark might need preserved (validation mode)
    * and reinitialize things that are *always* reinitialized e.g., debugger, areDebugging.
    *
-   * That means when we save for reloading, we must explicitly clobber validationMode and externalVars to
-   * initialized values.
+   * That means when we save for reloading, we must explicitly clobber validationMode in save().
    *
-   * @throws java.io.ObjectStreamException
+   * @throws java.io.ObjectStreamException Must be part of writeReplace's API
    * @return the serializable object
    */
   @throws(classOf[java.io.ObjectStreamException])
   private def writeReplace() : Object =
-    new SerializableDataProcessor(ssrd, tunables, externalVars, validationMode)
-
-  /**
-   * The compilerExternalVars argument supports the deprecated feature to assign external var bindings
-   * on the compiler object.
-   *
-   * These are just incorporated into the initial variable map of the data processor.
-   */
-
-  def this(
-    ssrd: SchemaSetRuntimeData,
-    tunables:DaffodilTunables,
-    compilerExternalVars: Queue[Binding] = Queue.empty,
-    validationMode: ValidationMode.Type = ValidationMode.Off) =
-    this(ssrd, tunables, ExternalVariablesLoader.loadVariables(compilerExternalVars, ssrd, ssrd.originalVariables),
-      false, None, validationMode, compilerExternalVars)
+    new SerializableDataProcessor(ssrd, tunables, variableMap.copy(), validationMode)
 
   def copy(
     ssrd: SchemaSetRuntimeData = ssrd,
     tunables: DaffodilTunables = tunables,
-    areDebugging : Boolean = areDebugging,
-    optDebugger : Option[Debugger] = optDebugger,
+    variableMap: VariableMap = variableMap.copy(),
     validationMode: ValidationMode.Type = validationMode,
-    variableMap : VariableMap = variableMap.copy,
-    externalVars: Queue[Binding] = externalVars) =
-    new DataProcessor(ssrd, tunables, variableMap, areDebugging, optDebugger, validationMode,  externalVars)
+    areDebugging: Boolean = areDebugging,
+    optDebugger: Option[Debugger] = optDebugger,
+  ) = new DataProcessor(ssrd, tunables, variableMap, validationMode, areDebugging, optDebugger)
 
   // This thread local state is used by the PState when it needs buffers for
   // regex matching. This cannot be in PState because a PState does not last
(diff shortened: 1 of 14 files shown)


stevedlawrence (approved):
+1, one minor question if we even need this other DataProcessor constructor?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -154,27 +154,23 @@ class DataProcessor private (
...
   def this(
     ssrd: SchemaSetRuntimeData,
     tunables:DaffodilTunables,
-    compilerExternalVars: Queue[Binding] = Queue.empty,
+    externalVars: Queue[Binding] = Queue.empty,
     validationMode: ValidationMode.Type = ValidationMode.Off) =
-    this(ssrd, tunables, ExternalVariablesLoader.loadVariables(compilerExternalVars, ssrd, ssrd.originalVariables),
-      false, None, validationMode, compilerExternalVars)
+    this(ssrd, tunables, ExternalVariablesLoader.loadVariables(externalVars, ssrd, ssrd.originalVariables),
+      false, None, validationMode, externalVars)
```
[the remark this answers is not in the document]
tuxji: I had initially thought SchemaSetRuntime1Mixin.scala was the only caller of this constructor too and removed the externalVars parameter from this constructor, but I got a compilation error which showed that there was a second call of this constructor from SerializableDataProcessor's constructor in DataProcessor.scala at line 107 that I didn't know about.  SerializableDataProcessor extends DataProcessor(data, tunable, externalVars, validationModeArg) in order to preserve the externalVars and validationMode in case they may be needed by other serializations besides Daffodil save/reload such as Apache Spark which serializes in order to move objects for remote execution.  

We can't remove the externalVars field from DataProcessor because we inform DataProcessor about external variable bindings in that field.  DataProcessor's primary constructor is private (probably because of the mix of parameters some of which should be serialized all the time, some which should be serialized only sometimes, and some which should never be serialized).  DataProcessor has both a class and an object, but the object DataProcessor has no apply method, only a private class SerializableDataProcessor.  Therefore, this particular (public) constructor is the only way which callers outside DataProcessor.scala can create a DataProcessor.  Our options are to keep this public constructor or add an apply method to the object DataProcessor which calls the private constructor.

At the very least, I need to fix this constructor's scaladoc.  I had changed the scaladoc to say it was needed by SerializableDataProcessor, but I misstated the reason.  I can make the scaladoc state that this constructor exists to construct a SerializableDataProcessor, not the other way around.  Or I can remove this constructor and add an apply method to object DataProcessor, which would put that apply method next to the class SerializableDataProcessor and make the relationship between them more obvious.  I would also add a second apply method with only 2 parameters and its own scaladoc to let external callers like SchemaSetRuntime1Mixin.scala create a DataProcessor.

Thoughts?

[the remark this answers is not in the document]
stevedlawrence: Hmm, good points.

Digging in some more based on what you've found, the only differences between the public and private DataProcessor constructors are the ability to pass in `areDebugging` and `optDebugger`. I'm not sure how important it is to restrict setting debugging when creating a DataProcessor (especially since this DataProcessor is internal only), so we could potentially just make the current private constructor public and remove the public one. And then the `SerializableDataProcessor` could just extend the current private one? And the current private constructor could set default values of the debugging parameters if we want the to be optional?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -129,10 +131,10 @@ class DataProcessor private (
   // The values these will have (since this is a base class) are the correct default values that we want
   // back when the object is re-initialized.
   //
-  protected val areDebugging : Boolean,
-  protected val optDebugger : Option[Debugger],
+  private val externalVars: Queue[Binding],
```
stevedlawrence, line 134: I think maybe `externalVars` can go away completely? The only way to set external variables should be with `withExternalVariables`, which I think should already create a copy of the variable map and mutate it with the new bindings?

I *think* externalVars was only needed because the compiler/processor factory could have variables set on them, so they needed to be carried down this way to the DataProcessor?

So maybe the `SchemaSetRuntimeData` just passes in the `originalVariables` Variable map when it creates the DataProcosser, and that is copied/mutated as needed?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -154,37 +156,21 @@ class DataProcessor private (
...
-    variableMap : VariableMap = variableMap.copy,
-    externalVars: Queue[Binding] = externalVars) =
-    new DataProcessor(ssrd, tunables, variableMap, areDebugging, optDebugger, validationMode,  externalVars)
+  def copy(ssrd: SchemaSetRuntimeData = ssrd,
+           tunables: DaffodilTunables = tunables,
+           variableMap: VariableMap = variableMap.copy(),
+           externalVars: Queue[Binding] = externalVars,
+           validationMode: ValidationMode.Type = validationMode,
+           areDebugging: Boolean = areDebugging,
+           optDebugger: Option[Debugger] = optDebugger) =
```
stevedlawrence, line 172: I *think* this indentation isn't the scala standard? I think the previous way is right.

tuxji:
I've removed all mentions/uses of `externalVars` in DataProcessor and restored `copy`'s original formatting.  I'm letting CI run the tests for me; if they pass, then this pull request is ready for a second reviewer to review it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -84,27 +83,25 @@ class InvalidUsageException(msg: String, cause: Throwable = null) extends Except
...
+   * Hence, we're depending on the save method to explicitly turn off validationMode.
    */
   private class SerializableDataProcessor(
     val data: SchemaSetRuntimeData,
     tunable: DaffodilTunables,
-    externalVars: Queue[Binding], // must be explicitly set to empty by save method
-    validationModeArg: ValidationMode.Type) // must be explicitly set from Full to Limited by save method.
-    extends DataProcessor(data, tunable, externalVars, validationModeArg) {
+    validationModeArg: ValidationMode.Type) // must be explicitly turned off by save method
+    extends DataProcessor(data, tunable, data.originalVariables, validationModeArg) {
```
stevedlawrence, line 104: Should this be the same (or a copy of) the VariableMap from the DataProcessor? That way if something like Spark wants to serialize the existing VariableMap then it can? We just don't do it thorugh bindings anymore but directly serialize the actual VariableMap?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -224,14 +203,14 @@ class DataProcessor private (
   }
 
   // TODO Deprecate and replace usages with just tunables.
```
stevedlawrence, line 205: Should we do what this comment suggests and replace `getTunable` with `tunbles`?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
```diff
@@ -311,18 +271,13 @@ class DataProcessor private (
...
+    // different than its original state.  Note that the serialization
+    // system *does* preserve validationMode since it may be required by
+    // other software like Apache Spark. But for our save/reload purposes,
+    // we don't want validationMode preserved.
     //
-    // But for our save/reload purposes, we don't want them preserved.
-    //
-
     val dpToSave = this.copy(
-      externalVars = Queue.empty[Binding], // explicitly set these to empty so restored processor won't have them.
```
stevedlawrence, line 325: I think we might still want this line if we add back the VariableMap to the SerializableDataProcessor. Or i guess really we want this to be `externalVars = ssrd.originalValues`?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
tuxji: I'd taken a look and backed away when I saw getTunable was used in DFDL.DataProcessor, but I just looked again and I think it's feasible and wouldn't require deprecation since getTunable doesn't seem to be part of the Java or Scala API.  The only places using getTunable seem to be some tests and runtime1 itself.  I'll replace getTunable with tunables.

```grep
./daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala:66:    val t1 = dp1.getTunables()
./daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala:67:    val t2 = dp2.getTunables()
./daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala:72:    val t3 = dp2.getTunables() // modified tunables at 'run-time'
./daffodil-core/src/test/scala/org/apache/daffodil/general/TestTunables.scala:73:    val t4 = dp1.getTunables() // obtain first data processor to see if anything changed
./daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetCursor.scala:94:    infosetInputter.initialize(rootERD, u.getTunables())
./daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetCursor1.scala:52:    ic.initialize(rootERD, u.getTunables())
./daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetCursorFromReader.scala:58:    inputter.initialize(rootERD, u.getTunables())
./daffodil-core/src/test/scala/org/apache/daffodil/infoset/TestInfosetCursorFromReader2.scala:66:    inputter.initialize(rootERD, u.getTunables())
./daffodil-core/src/test/scala/org/apache/daffodil/util/TestUtils.scala:348:    override def getTunables(): DaffodilTunables = { tunables }
./daffodil-runtime1/src/main/scala/org/apache/daffodil/api/DFDLParserUnparser.scala:195:    def getTunables(): DaffodilTunables
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DaffodilUnparseContentHandler.scala:81:  private lazy val tunablesBatchSize = dp.getTunables().saxUnparseEventBatchSize
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala:206:  def getTunables(): DaffodilTunables = tunables
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala:476:    inputter.initialize(ssrd.elementRuntimeData, getTunables())
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala:661:    val tunables = dataProc.getTunables()
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/PState.scala:691:    val tunables = dataProc.getTunables()
./daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/unparsers/UState.scala:692:        dataProc.getTunables(),
./daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala:66:  override def getTunables(): DaffodilTunables = ???
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
tuxji: I think you meant `variableMap = ssrd.originalVariables` since we've gotten rid of all uses/mentions of `externalVars`.  Anyway, should a saved processor preserve its own `variableMap` or replace it with `ssrd.originalVariables`?  That is the real question we need to ask.  Mike @mbeckerle , do you want to weigh in?

tuxji (commented):
Additional changes seem reasonable but want a discussion whether a saved processor should preserve its own variableMap or reset variableMap.

mbeckerle:
When you restore a saved compiled schema, you should get the initial values of any DFDL external variables, regardless of whether they had any values set before the schema was pre-compiled. 

Saving the parser does NOT lock-down/freeze or even utilize any variable values.  Saving the parser is supposed to be transparent, i.e., only faster, than using the uncompiled schema. 

For example, expressions that refer to variables, even if the variable has been externally set, cannot be constant-folded by the schema compiler to remove the variable reference. Rather, any attempt to reference a variable at schema compile time causes the expression folding to stop. 

There are, however, some tunables you can set which only affect schema compilation. Those are only used by the schema compiler so clearly one cannot subsequently modify them after restoring a saved processor. (You can, but it won't do anything because the compilation is already over. )  Example of this is a tunable to suppress a particular compile-time warning, or to specify a particular interpretation  (UnqualifiedPathStepPolicy is an example.)

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
tuxji: I ended up omitting variableMap because variableMap already has a default value (variableMap.copy()) which is copied if not passed explicitly.  

Mike @mbeckerle , will you please review all the changes so far?  I want to get this PR done and out of the way.

daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/Runtime2DataProcessor.scala
```diff
@@ -46,28 +46,19 @@ import org.apache.daffodil.util.Maybe.Nope
...
-  
   override def withDebugging(flag: Boolean): DFDL.DataProcessor = ???
 
-  override def validationMode: ValidationMode.Type = ???
-
-  override def getTunables(): DaffodilTunables = ???
-
   override def save(output: DFDL.Output): Unit = ???
 
+  override def tunables: DaffodilTunables = ???
```
mbeckerle, line 59: Suggest put coverage off/on around these stubs.

mbeckerle (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/DataProcessor.scala
stevedlawrence: Doesn't this change the behavior? I believe previously if you `save()` and `reload()` you would get the original variables values defined in the schema? With this, doesn't a `reload()` get whatever variables were set on the `DataProcessor` at the time of `save()`?

stevedlawrence: From Mikes other comment:

> Saving the parser does NOT lock-down/freeze or even utilize any variable values. Saving the parser is supposed to be transparent, i.e., only faster, than using the uncompiled schema.

So it seems like we do want `variableMap = ssrd.originalVariables` so that calling `save()` essentially erases any calls to `withVariableMap`

mbeckerle: Ah. Yes, steve is right.  We do want the originalVariables.

PR #889, opened 2022-12-08
Store the last added child in the PState for use in separator logic

stevedlawrence:
In some cases, separator logic needs knowledge about the most recently added child of an element. This is currently handled by examining the infoset and finding its last child element using the maybeMostRecentlyAddedChild function.

The problem with this approach is that the InfosetWalker is allowed to "release" elements from the infoset that it thinks are no longer needed. There isn't a good way to tell the infoset that these last children are potentially still needed for separator logic, and so it could actually release them prior to the separator logic trying to access them, which leads to a null pointer exception.

To fix this, this modifies the PState to add a slot for the last modified child of the current infoset element, and modifies the ElementParser's to set this state appropriately. This way, the InfosetWalker is free to remove elements as it normally does, but the last child is still available when needed.

This also removes the maybeMostRecentlyAddedChild functions since this kind of access to the infoset can lead to null pointers.

Also modifies the SAXInfosetOutputter to check isNilled correctly, which supports checking if complex elements are nilled without being final yet. For similar reasons, modifies the InfosetWalker so that it does not walk into Complex elements if there is a chance that it could be nilled and we might not be sure.

DAFFODIL-2755

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/InfosetImpl.scala
@@ -1111,15 +1111,6 @@ final class DIArray(
 
   final def length: Long = _contents.length
 
-  final def maybeMostRecentlyAddedChild(): Maybe[DIElement] = {
-    val len = contents.length
-    if (len == 0) Maybe.Nope
-    else {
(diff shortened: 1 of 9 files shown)


mbeckerle (approved):
+1 Very clean and small change.

tuxji (approved):
+1

Looks good to me 👍

PR #890, opened 2022-12-12
Deep copy Attributes with SAX performance unparse command

stevedlawrence:
When running the CLI performance unparse command with the `-I sax` option, we deep copy all SAX events and convert them to array of objects so we can replay those events in the actual performance loop. This avoids the overhead of parsing the XML infoset when testing SAX.

However, we don't currently deep copy the Attributes parameter in the startElement function. So if a SAX XMLReader implementation reuses/mutates the same Attributes object for multiple calls to startEvent (which Xerces does), then all of our StartElement events will share the same Attributes object with mutated values, leading to unexpected failures.

To fix this, this creates an AttributesImpl object from the Attributes object passed into the startElement function, which takes a persistent snapshot that is safe to store in our array of events.

DAFFODIL-2400


tuxji (approved):
+1

Good example of why functional computing with immutative state produces fewer bugs than procedural computing with mutative state.

mbeckerle (approved):
+1

PR #891, opened 2022-12-12
Update CI workflows and build tools

tuxji:
Tell CI not to cancel runs from 2 or more PRs made from the same head ref.  Bump older tools or names to newer tools or names.

dependency-scan.yml: Bump ubuntu from 20.04 to 22.04.

main.yml: Add github.actor to concurrency.group to avoid cancelling runs when 2 or more contributors make PRs from the same head ref (saw fix elsewhere, copy it just in case).  Bump ubuntu and its env_cc, env_ar from 20.04 to 22.04.  Also bump windows-2019 to windows-2022.

BUILD.md: Modernize build instructions by using more correct names Java/sbt instead of JDK/SBT, removing unneeded version numbers, and changing Java URL to Adoptium instead of AdoptOpenJDK.

DEVELOP.md: Use more correct name sbt instead of SBT.

README.md: Use more correct names Java/sbt instead of JDK/SBT.

Dockerfile: Use more correct name sbt instead of SBT.

daffodil-cli/build.sbt: Use more correct name sbt instead of SBT.

Product_en-us.wxl: Use more correct name sbt instead of SBT.

c/Makefile: Remove Ubuntu version to avoid updating Makefile again.

TestScalaAPI.scala: Use more correct name sbt instead of SBT (https://www.scala-sbt.org/1.x/docs/Faq.html).

DAFFODIL-2756

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -19,16 +19,16 @@
 
 Daffodil's build requirements include:
 
-* JDK 8 or higher
-* SBT 0.13.8 or higher
+* Java 8 or higher
+* sbt 0.13.8 or higher
 * C compiler C99 or higher
 * Mini-XML Version 3.0 or higher
 
-You will need the Java Software Development Kit ([JDK]) and the Scala
-Build Tool ([SBT]) to build Daffodil, run all tests, create packages,
-and more.  You can install the latest [Java 11 LTS][JDK] version and
-the latest [SBT] version following their websites' instructions or
-install them using your operating system's package manager.
+You will need the Java Software Development Kit ([Java]) and the Scala
+Build Tool ([sbt]) to build Daffodil, run all tests, create packages,
+and more.  You can install the latest [Java LTS][Java] version and the
+latest [sbt] version following their websites' instructions or install
+them using your operating system's package manager.
 
 Since Daffodil now has a C backend as well as a Scala backend, you
 will need a C compiler supporting the [C99] standard or later and the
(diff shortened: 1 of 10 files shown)


stevedlawrence (approved):
+1 :+1:

BUILD.md
```diff
@@ -19,15 +19,15 @@
 
 Daffodil's build requirements include:
 
-* JDK 8 or higher
-* SBT 0.13.8 or higher
+* Java 8 or higher
+* sbt 0.13.8 or higher
```
mbeckerle, line 23: Do we really support sbt this old? I don't want to even bother checking it.

BUILD.md
```diff
@@ -19,15 +19,15 @@
...
 * C compiler C99 or higher
 * Mini-XML Version 3.0 or higher
 
-You will need the Java Software Development Kit ([JDK]) and the Scala
-Build Tool ([SBT]) to build Daffodil, run all tests, create packages,
-and more.  You can install the latest [Java 11 LTS][JDK] version and
-the latest [SBT] version following their websites' instructions or
+You will need the Java Software Development Kit ([Java]) and the Scala
+Build Tool ([sbt]) to build Daffodil, run all tests, create packages,
+and more.  You can install the latest [Java 11 LTS][Java] version and
```
mbeckerle, line 29: Why stipulate Java 11? 
We should either say Java 8, or Java 8 (Java 11 and Java 17 are also supported)

BUILD.md
```diff
@@ -59,35 +59,35 @@ environment variables `CC` and `AR` to the clang binaries' names:
...
 
 If you want to use clang instead of gcc, you'll have to set your
 environment variables `CC` and `AR` to the clang binaries' names:
 
-    export CC=clang-10 AR=llvm-ar-10
+    export CC=clang AR=llvm-ar-14 # or llvm-ar-10 or whatever you have
 
-However, Ubuntu has no [sbt][SBT] package in its own repositories.
-You'll have to install the latest [SBT] version following its
+However, Ubuntu has no [sbt] package in its own repositories.
```
mbeckerle, line 81: This may not be true any longer. I got prompted to update sbt by the ubuntu software updater recently.

BUILD.md
```diff
@@ -59,35 +59,35 @@ environment variables `CC` and `AR` to the clang binaries' names:
...
+You'll have to install the latest [sbt] version following its
 website's instructions.
 
 Now you can build Daffodil from source and the sbt and daffodil
 commands you type will be able to call the C compiler.
 
 ## Windows 10
 
-Install the latest [Java 11 LTS][JDK] version and the latest [SBT]
+Install the latest [Java 11 LTS][Java] version and the latest [sbt]
```
mbeckerle, line 90: Java version issue again.

mbeckerle (approved):
A few suggestions, but +1 regardless of how those come out.

BUILD.md
stevedlawrence: I believe that version of sbt (or thereabouts) added the ability to use the version of sbt defined in project/build.properties. So as long as a user has that version or newer, it should work and download the correct version we actually build and test with. That said, 0.13 is really old, so if we want to bump it to something more modern that doesn't seem unreasonable.

BUILD.md
tuxji: If you check your apt sources, you'll probably find the sbt project's apt repository.  When I updated this file, I checked again and there was still no sbt package in Debian or Ubuntu apt repositories.  BTW, I found out the reason for that from sbt's Wikipedia page, which says:

> The sbt project is "bootstrapped" — it uses sbt to build itself, as many compilers also do, and it considers that dogfooding is a positive feature. To the Debian project, however, that is considered a circular dependency, that they try to minimize. As a result, sbt is not yet in Debian.

BUILD.md
tuxji: Removed 11 here too.

tuxji (commented):
Comments below, thanks for the review.

PR #894, opened 2022-12-15
Change message to use propName instead of propNameArg

mike-mcgann:
The diagnostic message uses the value of `propNameArg` which could be set to null. The `propName` variable contains an auto-generated value when one is not provided in the constructor.

[DAFFODIL-2647](https://issues.apache.org/jira/browse/DAFFODIL-2647)


tuxji (approved):
+1

For anyone curious to understand why the change works, expand the diff upwards to include the earlier 20 lines and you will see the definition of the inherited AutoPropNameBase's propName value.  I also checked and did not find any other places where propNameArg needs to be replaced with propName.

mbeckerle (approved):
+1

mbeckerle:
I'm going to disregard the macos spurious check failure. That's not due to this fix certainly.

PR #895, opened 2022-12-16
Require the first schema to include the DFDL namespace.

mike-mcgann:
If a schema document is passed to Daffodil that does not contain a DFDL namespace, it is currently ignored and a warning is emitted. If this is the only document then no operation actually occurs and this condition can be masked when warnings are disabled. Enforce the first schema to include the DFDL namespace but emit warnings on subsequent imports or inclusions.

[DAFFODIL-2441](https://issues.apache.org/jira/browse/DAFFODIL-2441)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
@@ -57,13 +57,20 @@ trait SchemaSetIncludesAndImportsMixin { self: SchemaSet =>
   lazy val allSchemaFiles = {
     val fd = fakeXMLSchemaDocument //bootstrap
     val sa = fd.seenAfter
+    val first = sa.value.head._2
     val sfl = sa.value.flatMap {
       case (_, ii) => {
         val sf = ii.iiSchemaFileMaybe // maybe not if we've already seen this file for the same namespace.
+        // Require the first schema file to have a DFDL namespace. Other included or imported schemas can be
+        // standard XSD schemas but emit a warning that the schema is being ignored.
         sf.filter { f =>
-          if (f.isDFDLSchemaFile)
+          if (f.isDFDLSchemaFile) {
             true
-          else {
+          } else if (f eq first) {
+            f.SDE("Non-DFDL Schema file. Does not have DFDL namespace definition on schema root element.\n" +
+              "Add xmlns:dfdl='%s' to the root element.", XMLUtils.DFDL_NAMESPACE)
+            false
+          } else {
             f.SDW(WarnID.IgnoreImport, "Non-DFDL Schema file ignored. Does not have DFDL namespace definition on schema root element.\n" +
               "Add xmlns:dfdl='%s' to the root element if this file must be part of the DFDL schema.", XMLUtils.DFDL_NAMESPACE)
             false
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestSchemaWithoutDFDLNamespace.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestSchemaWithoutDFDLNamespace.scala
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.section00.general
+
+import org.apache.daffodil.tdml.Runner
+import org.junit.AfterClass
+import org.junit.Test
+
+object TestSchemaWithoutDFDLNamespace {
+  val testDir = "/org/apache/daffodil/section00/general/"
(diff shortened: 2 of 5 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
```diff
@@ -57,13 +57,20 @@ trait SchemaSetIncludesAndImportsMixin { self: SchemaSet =>
   lazy val allSchemaFiles = {
     val fd = fakeXMLSchemaDocument //bootstrap
     val sa = fd.seenAfter
+    val (_, first) = if (sa.value.isEmpty) (null, null) else sa.value.head
```
stevedlawrence, line 60: Is it actually possible for `sa.value.isEmpty` to be true? Seems like we must always have at least one value since we must be compiling something. If so, could this be simplified to this:
```scala
val first = sa.value.head._2
```
Then we don't this `null` stuff, which we tend to avoid in scala.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
```diff
@@ -57,13 +57,20 @@ trait SchemaSetIncludesAndImportsMixin { self: SchemaSet =>
...
         val sf = ii.iiSchemaFileMaybe // maybe not if we've already seen this file for the same namespace.
+        // Require the first schema file to have a DFDL namespace. Other included or imported schemas can be
+        // standard XSD schemas but emit a warning that the schema is being ignored.
         sf.filter { f =>
-          if (f.isDFDLSchemaFile)
+          if (f.isDFDLSchemaFile) {
             true
-          else {
+          } else if (f eq first) {
+            f.SDEButContinue("Non-DFDL Schema file. Does not have DFDL namespace definition on schema root element.\n" +
```
stevedlawrence, line 70: Why `SDEButContinue` instead of `SDE`? Seems like we would want to immediately stop if the main schema isn't a DFDL schema.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
mike-mcgann: Updated with the recommended change.

daffodil-core/src/main/scala/org/apache/daffodil/dsom/SchemaSetIncludesAndImportsMixins.scala
mike-mcgann: Updated with the recommended change.

stevedlawrence (approved):
+1 :+1:

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestSchemaWithoutDFDLNamespace.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.daffodil.tdml.Runner
+import org.junit.{AfterClass, Test}
+
+object TestSchemaWithoutDFDLNamespace {
+  val testDir = "/org/apache/daffodil/section00/general/"
+  val runner = Runner(testDir, "testSchemaWithoutDFDLNamespace.tdml")
+
+  @AfterClass def shutDown(): Unit = {
+    runner.reset
```
tuxji, line 28: For future reference, we don't want new code to have easily fixable IDE warnings (val runner: Runner = ..., and runner.reset()).  IDEA even tries to prevent you from committing code with such warnings so other devs changing the same file will have to fix the warnings or answer yes, commit anyway.  Scala 3 migration will need these fixes anyway if my memory is correct.

tuxji (approved):
+1

P.S. For future reference, a PR author waits for all reviewers to approve before rebasing one's branch, instead of rebasing immediately after each change.  GitHub still keeps all previous comments and their context on the Conversation tab which is good (we do lose the context and comments on the Files changed tab), but it's harder for other reviewers to see what the PR author changed between force pushes and it's also easier for the PR author to do the rebase/squash/force push only once at the end than to keep doing it each time.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestSchemaWithoutDFDLNamespace.scala
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.section00.general
+
+import org.apache.daffodil.tdml.Runner
+import org.junit.{AfterClass, Test}
```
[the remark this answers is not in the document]
mike-mcgann: Imports have been updated.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestSchemaWithoutDFDLNamespace.scala
mike-mcgann: I'm not seeing any warnings over here. Can you let me know what the warnings are and I'll see if I can get those turned on. Thanks.

tuxji: I'm not sure why you wouldn't see these warnings since IDEA should be enabling some kinds of warnings out of the box already.  You can check your IDEA settings: Open File -> Settings, type scala into the search box, hit the Enter key, and click on Editor -> Inspections to see all the enabled Scala warnings (Build -> Compiler -> Scala Compiler also has an Enable warnings setting which should be enabled too).  

Another possibility is that you're not seeing the warnings because IDEA is very subtle in showing them (a yellow bar in the right gutter, a count of warnings in the upper right corner, and you have to click on the count to see a list of warnings in the Problems tab).

mike-mcgann: I'm seeing them now. They were there but very subtle as you indicated. I'm brand new to Scala and IDEA so it take me a bit to learn these. Thanks for the info. 

I've updated the tests to include the recommended changes from the warnings.

PR #896, opened 2022-12-18
Add custom error handler for CLI EXI encode/decode

jadams-tresys:
Adding custom error handler that implements SAX ErrorHandler and Transform ErrorListener helps contain some of the error messages that were getting printed to stderr and were polluting the output of our integration tests.

There is still an issue with one of the tests causing a stack trace from an IllegalArgumentException to be printed that occurs when decoding a schema aware encoded EXI file with a non-schema aware decoder. This appears to be getting printed directly to stderr from within Exificient.

DAFFODIL-2750


stevedlawrence:
> There is still an issue with one of the tests causing a stack trace from an IllegalArgumentException to be printed

I did a little digging, and I think I found the cause. For reference, here's the stack trace:

```
java.lang.IllegalArgumentException
  | => dat java.lang.AbstractStringBuilder.appendCodePoint(AbstractStringBuilder.java:800)
        at java.lang.StringBuilder.appendCodePoint(StringBuilder.java:240)
	at com.siemens.ct.exi.core.io.channel.AbstractDecoderChannel.decodeStringOnlySupplementaryCodePoints(AbstractDecoderChannel.java:111)
	at com.siemens.ct.exi.core.io.channel.AbstractDecoderChannel.decodeStringOnly(AbstractDecoderChannel.java:89)
	at com.siemens.ct.exi.core.coder.AbstractEXIBodyDecoder.decodeLocalName(AbstractEXIBodyDecoder.java:170)
	at com.siemens.ct.exi.core.coder.AbstractEXIBodyDecoder.decodeQName(AbstractEXIBodyDecoder.java:133)
	at com.siemens.ct.exi.core.coder.AbstractEXIBodyDecoder.decodeStartElementGenericStructure(AbstractEXIBodyDecoder.java:484)
	at com.siemens.ct.exi.core.coder.EXIBodyDecoderInOrder.decodeStartElement(EXIBodyDecoderInOrder.java:126)
	at com.siemens.ct.exi.main.api.sax.SAXDecoder.parseEXIEvents(SAXDecoder.java:313)
	at com.siemens.ct.exi.main.api.sax.SAXDecoder.parse(SAXDecoder.java:250)
	at net.sf.saxon.resource.ActiveSAXSource.deliver(ActiveSAXSource.java:190)
	at net.sf.saxon.event.Sender.send(Sender.java:105)
	at net.sf.saxon.jaxp.IdentityTransformer.transform(IdentityTransformer.java:364)
	at org.apache.daffodil.Main$.runIgnoreExceptions(Main.scala:1307)
	at org.apache.daffodil.Main$.run(Main.scala:1433)
	at org.apache.daffodil.CLI.Util$CLIThread.run(Util.scala:281)
```

The issue is that `appendCodePoint` can throw an IllegalArgumentException if the code point isn't valid unicode (this isn't really documented except for the fact that it is "exactly like Characters.toChars" which does document throwing this exception). Exificient doesn't catch this, so if this happens it bubbles all the way up to the `deliver` function in Saxon and is finally caught (and printed) here:

https://saxonica.plan.io/projects/saxonmirrorhe/repository/he/entry/src/main/java/net/sf/saxon/resource/ActiveSAXSource.java?utf8=%E2%9C%93&rev=he_mirror_saxon_11_4#L233

Looking at the latest dev branch of Saxon, this call to printStrackTrace has been removed, but still exists in the 11.4 release we use. So this will likely be fixed in the next version of Saxon. However, I *think* this is a bug in Exificient since it shouldn't throw an IllegalArgumentException, but should probably validate the codepoint and throw a SAX exception instead?

Does this sound right? Maybe we should open a ticket with efficient?

stevedlawrence (approved):
+1, looks like the other stack trace is a bug in exificient/saxon

stevedlawrence:
In fact, now that I think about it, that's probably also the reason why we need this code:

https://github.com/apache/daffodil/blob/main/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala#L1306-L1312

So what we thought was a generic exception was more likely an exificient bug.

Once Exifcient fixes this issue, we can change that from catching an Exception to whatever SAX exception they end up throwing.

mbeckerle:
Suggest: If you have a ticket URL for Exificient you could cite that in the comment just before the catch of e:Exception.

mbeckerle (approved):
+1

tuxji (approved):
+1

Agree that EXI ticket URL should be cited in a comment before the catch of Exception at https://github.com/apache/daffodil/blob/main/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala#L1306-L1312.

PR #897, opened 2022-12-20
Update release-candidate to install VS Code Extension docs.

Shanedell:
Update release-candidate to install VS Code Extension docs.

- containers/release-candidate/Dockerfile updated to install pandoc, needed to build the VS Code extension docs.
- containers/release-candidate/src/daffodil-release-candidate updated to clone the VS Code extension repos wiki, build the html and add them into the daffodil-site.

This comes from https://github.com/apache/daffodil-vscode/pull/381


containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,23 @@ case $PROJECT_REPO in
...
 
+    echo "Building Extension Docs..."
+    git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
+    for markdown_file in $(echo "\
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code \
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code:-Development \
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code:-Roadmap \
+      Home"
+    ); do
+      pandoc -t html -f markdown -o "${markdown_file}.html" ".wiki/${markdown_file}.md"
```
stevedlawrence, line 312: I thought the original goal was to have a single PDF easily downloadable for offline documentation? But this just creates multiple html pages, which I think are really just snapshot of the wiki. And the fomatting isn't as nice. For example, the github pages have a navigation menu on he right of each page, and the images are scaled to fit.

So I guess my question is how important is this snapshotting of wiki documentation? Seems to me most of the time people will always be using the latest version of the extension and go to the github wiki. It feels different from the Daffodil JAPI/SAPI where often times upgrading to a new version really does incur a cost so people do need API docs of older versions. And Java/Scala docs also aren't easily available since they need to be generated, unlike the github wiki.

containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,23 @@ case $PROJECT_REPO in
     yarn package
     cp *.vsix $DAFFODIL_RELEASE_DIR/bin/
 
+    echo "Building Extension Docs..."
+    git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
+    for markdown_file in $(echo "\
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code \
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code:-Development \
+      Apache-Daffodil™-Extension-for-Visual-Studio-Code:-Roadmap \
+      Home"
```
stevedlawrence, line 310: Can we automate getting the list of files? Seems it would be difficult to remember to update this script every time a new wiki page is created

containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,23 @@ case $PROJECT_REPO in
...
+    ); do
+      pandoc -t html -f markdown -o "${markdown_file}.html" ".wiki/${markdown_file}.md"
+    done
+    rm -rf .wiki
+
+    echo "Installing Extension Docs..."
+    rm -rf $DAFFODIL_VSCODE_DOCS_DIR
+    mkdir -p $DAFFODIL_VSCODE_DOCS_DIR
+    cp -R *.html $DAFFODIL_VSCODE_DOCS_DIR
```
stevedlawrence, line 319: Would it be possible to generate the *.html files in a different directory, or maybe in the .wiki directocy? I could imagine someday we might add a an .html file to the root of the daffodil-vscode repo and it shouldn't be copied to the docs directory.

containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,23 @@ case $PROJECT_REPO in
...
+
```
stevedlawrence, line 320: Below under "Committing Site Changes", we have a line that looks like this:
```
git diff-index --quiet HEAD || git commit -m "Stage release v$VERSION-$PRE_RELEASE"
```
Since this now could commit something for daffodil or daffodil vscode, I suggest we change it to this:
```
git diff-index --quiet HEAD || git commit -m "Stage $PROJECT_NAME release v$VERSION-$PRE_RELEASE"
```

stevedlawrence (commented):
Approach looks right, just a couple minor suggestions.

containers/release-candidate/src/daffodil-release-candidate
Shanedell: Oh if that is what we want it makes sense to me. I am not sure if I can combine into one single page. From the code I made it from it seemed like it was doing page by page so that is what I did. But I can try testing what you mention here. @scholarsmate Are you able to better explain what was desired for saving off the documentation from the wiki?

containers/release-candidate/src/daffodil-release-candidate
tuxji: Yes, and you can get the list of files quite easily using globbing like this,

```shell
cd .wiki
for markdown_file in *.md; do
  pandoc -t html -f markdown -o "../${markdown_file}.html" "${markdown_file}"
done
```

containers/release-candidate/src/daffodil-release-candidate
Shanedell: Yeah I don't know though about a single PDF is that possible with pandoc?

tuxji: If snapshotting the wiki still serves a useful purpose, there are other tools you can look at such as [github-wikito-converter](https://github.com/yakivmospan/github-wikito-converter) which can convert the entire wiki to a single HTML or PDF file in one go:

```shell
sudo npm install -g github-wikito-converter
git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
gwtc --file-name <filename> --format <html|pdf> --title "<wiki title>" .wiki
```

tuxji (commented):
Suggesting how to use globbing and other tools worth trying.

Shanedell:
@stevedlawrence I like @tuxji 's suggestion to use [github-wikito-converter](https://github.com/yakivmospan/github-wikito-converter). It works pretty well and makes one page. But when trying it with pdf it seems to not work right but it works for html properly, do we want to use that instead of what do you think?

stevedlawrence:
I have no strong preference. I didn't come up with this requirement so I don't really know what the goal here is. Personally, if I wanted offline documentation I'd probably prefer a PDF, but I don't think I'll really use this (github wiki is fine for me) so don't let me sway the decision.

containers/release-candidate/src/daffodil-release-candidate
Shanedell: So I had some issues with getting pdf to work completely with that tool. However, I was able to produce a single docx file for all Wiki markdown files.

containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,17 @@ case $PROJECT_REPO in
     yarn package
     cp *.vsix $DAFFODIL_RELEASE_DIR/bin/
 
+    echo "Building Extension Docs..."
+    doc_name="Apache-Daffodil-Extension-for-Visual-Studio-Code-$VERSION.docx"
+    git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
+    pandoc -t docx -f markdown -o .wiki/$doc_name .wiki/*.md
```
tuxji, line 307: Does pandoc allow you to create a single HTML file from multiple markdown files?  That is, does the following work as well?

```shell
    echo "Building Extension Docs..."
    doc_name="Apache-Daffodil-Extension-for-Visual-Studio-Code-$VERSION.html"
    git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
    pandoc -t html -f markdown -o .wiki/$doc_name .wiki/*.md
```

Everyone has a web browser and can read a HTML file, but I'm not sure everyone has a word processor which can read a .docx file.  Are there any compelling reasons for picking docx as the format for reading documentation?

Shanedell: Yeah it can do a single HTML file as well. The snippet you sent would be correct, I was thinking we wanted something to download but maybe it is better to keep it as HTML instead of trying to use pdf or markdown that would be fine by me. I will say though on the html instance some of the images are huge that would be the main thing that looks off. @scholarsmate It would be nice to here your thoughts as I believe you made the portable document issue (https://github.com/apache/daffodil-vscode/issues/380) for the vscode repo.

stevedlawrence: Another thing to consider with doc/pdf is there are animated gifs in our wiki. I'm not sure how well that translates to document formats. It might be worth considering if those gifs provide value and if the documentation could be made to work with static images, but that's another issue.

tuxji (approved):
+1 unless someone doesn't want a single HTML file with some large images.

mbeckerle (approved):
+1

Shanedell:
@tuxji @mbeckerle @stevedlawrence Do you think it would be best to create a HTML and DOCX file? Then on the site have options under `Documentation` for `HTML | DOCX`? This way the user can view the single HTML page in their browser or download the DOCX file to view locally if they so desire?

tuxji:
> Do you think it would be best to create a HTML and DOCX file? 

That is fine with me, especially if the docx file has better formatting in case someone wants to print the guide on paper.

containers/release-candidate/src/daffodil-release-candidate
scholarsmate: @Shanedell, the portable document requirement stems from a need to have access to the documentation on an air-gapped SharePoint site.

Shanedell:
@tuxji @mbeckerle Can you guys give this another look over before merging? I generating both HTML and DOCX, but I generate the HTML doc with `github-wikito-converter` as it creates a better html page than pandoc does.

containers/release-candidate/src/daffodil-release-candidate
```diff
@@ -300,6 +301,18 @@ case $PROJECT_REPO in
     yarn package
     cp *.vsix $DAFFODIL_RELEASE_DIR/bin/
 
+    echo "Building Extension Docs..."
+    doc_name="Apache-Daffodil-Extension-for-Visual-Studio-Code-$VERSION"
```
stevedlawrence, line 305: The convention we've been using in this script is ALL_CAP variable names.

Shanedell: I will fix that now, my apologies

containers/release-candidate/Dockerfile
```diff
@@ -42,6 +43,10 @@ RUN \
     winetricks && \
   dnf clean all
 
+# Install node dependencies:
+#   github-wikito-converter - creates html doc from github wiki repo
+RUN npm install -g github-wikito-converter
```
tuxji, line 48: Reviewing again as you requested. I don't know how much data npm install downloads under the covers, but if we go to the trouble of calling dnf clean all to clear everything that was downloaded and cached by dnf install, is there any equivalent command we want to append to this line to clear any npm cache?  Other than that, I believe you're good to squash and merge.

Shanedell: There could be. However, I noticed using this `github-wikito-converter` it fails to convert all pages to HTML it just links to README files which we don't want. So I am going to remove this line from the Dockerfile and go back to using pandoc for both docx and html.

PR #898, opened 2022-12-20
Add emptyElementParsePolicy to main DFDL namespace.

mike-mcgann:
The emptyElementParsePolicy is currently implemented as a property that is an extension to DFDL. As of DFDL 1.0, this property is now available in the main schema. This property has now been added and a deprecation message is emitted when the old property is used. The property value of `treatAsMissing` has been renamed to `treatAsAbsent` in DFDL 1.0.

[DAFFODIL-2496](https://issues.apache.org/jira/projects/DAFFODIL/issues/DAFFODIL-2496)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
@@ -438,11 +438,11 @@ class ScalarOrderedSequenceChild(sq: SequenceTermBase, term: Term, groupIndex: I
 
   /**
    * Must deal with nils, emptyness and string/hexBinary exceptional behavior
-   * including the behavior for dfdlx:emptyElementParsePolicy 'treatAsMissing' which special cases
+   * including the behavior for dfdl:emptyElementParsePolicy 'treatAsAbsent' which special cases
    * Required elements like scalars, iff they are emptyRep, emptyValueDelimiterPolicy,
    * nilValueDelimiterPolicy, complex elements that nillable, or fully defaultable.
    *
-   * So we have ((simpleStringHexBinary x (treatAsMissing, treatAsEmpty), simpleOther, complex) x (nillable, not) x
+   * So we have ((simpleStringHexBinary x (treatAsAbsent, treatAsEmpty), simpleOther, complex) x (nillable, not) x
    * 4 behaviors. That's 32 combinations. Let's start with fewer cases and more runtime
    * decisions, and specialize if we think it will help clarity or performance.
    */
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
@@ -103,7 +103,8 @@
       </dfdl:defineFormat>
       
       <dfdl:defineFormat name="GeneralFormat">
-        <dfdl:format ref="GeneralFormatOriginal" />
+        <dfdl:format ref="GeneralFormatOriginal"
+            emptyElementParsePolicy="treatAsEmpty"/>
       </dfdl:defineFormat>
         
       <dfdl:defineFormat name="GeneralFormatPortable">
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
@@ -430,9 +430,11 @@ trait TextStandardExponentRepMixin extends PropertyMixin {
  */
 sealed trait EmptyElementParsePolicy extends EmptyElementParsePolicy.Value
 object EmptyElementParsePolicy extends Enum[EmptyElementParsePolicy] {
-  case object TreatAsMissing extends EmptyElementParsePolicy
+  case object TreatAsMissing extends EmptyElementParsePolicy // deprecated
   case object TreatAsEmpty extends EmptyElementParsePolicy
-  override lazy val values = Array(TreatAsMissing, TreatAsEmpty)
+  case object TreatAsAbsent extends EmptyElementParsePolicy
+
+  override lazy val values = Array(TreatAsMissing, TreatAsEmpty, TreatAsAbsent) // deprecated: TreatAsMissing
 
   def apply(name: String, context: ThrowsSDE): EmptyElementParsePolicy = stringToEnum("emptyElementParsePolicy", name, context)
 }
@@ -454,7 +456,7 @@ trait EmptyElementParsePolicyMixin extends PropertyMixin {
 
   /**
    * Property determines whether Daffodil implements empty elements in a manner consistent with
-   * IBM DFDL (as of 2019-05-02), which has policy treatAsMissing, or implements what is
+   * IBM DFDL (as of 2019-05-02), which has policy treatAsAbsent, or implements what is
    * described in the DFDL spec., which is treatAsEmpty - if the syntax of
    * empty (or nullness) is matched, create an empty (or null) item, even if optional, unless
    * the element is entirely absent.
@@ -468,15 +470,10 @@ trait EmptyElementParsePolicyMixin extends PropertyMixin {
       // prop is not required AND not defined so use tunable value
       // but issue warning (which can be suppressed)
       val defaultEmptyElementParsePolicy = this.tunable.defaultEmptyElementParsePolicy
-      // This property is an extension, so we don't want to require users to
-      // add this property just to silence this warning, especially since the
-      // property might change in the future. So silently use a default without
-      // outputting a warning. We may want to turn this warning on when the
-      // property makes its way into the official DFDL spec.
-      //SDW(
-      //  WarnID.EmptyElementParsePolicyError,
-      //  "Property 'dfdlx:emptyElementParsePolicy' is required but not defined, using tunable '%s' by default.",
-      //  defaultEmptyElementParsePolicy)
+      SDW(
+        WarnID.EmptyElementParsePolicyError,
+        "Property 'dfdl:emptyElementParsePolicy' is required but not defined, using tunable '%s' by default.",
+        defaultEmptyElementParsePolicy)
       defaultEmptyElementParsePolicy
     }
   }
--- a/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
+++ b/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
@@ -36,7 +36,7 @@
     <xs:restriction base="xs:string">
       <xs:enumeration value="dfdlx:alignmentKind" />
       <xs:enumeration value="dfdlx:choiceBranchKeyRanges" />
-      <xs:enumeration value="dfdlx:emptyElementParsePolicy"/>
+      <xs:enumeration value="dfdlx:emptyElementParsePolicy"/> <!-- deprecated -->
       <xs:enumeration value="dfdlx:inputTypeCalc"/>
       <xs:enumeration value="dfdlx:objectKind"/>
       <xs:enumeration value="dfdlx:outputTypeCalc"/>
@@ -94,11 +94,11 @@
     </xs:restriction>
   </xs:simpleType>
 
-  <xs:attribute name="emptyElementParsePolicy" type="dfdlx:EmptyElementParsePolicyEnum"/>
+  <xs:attribute name="emptyElementParsePolicy" type="dfdlx:EmptyElementParsePolicyEnum"/> <!-- deprecated -->
   <xs:simpleType name="EmptyElementParsePolicyEnum">
     <xs:restriction base="xs:string">
       <xs:enumeration value="treatAsEmpty" />
-      <xs:enumeration value="treatAsMissing" />
+      <xs:enumeration value="treatAsMissing" /> <!-- deprecated -->
     </xs:restriction>
   </xs:simpleType>
   
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
@@ -74,11 +74,11 @@ object ParseAttemptStatus {
    *
    * The EmptyRep for simpleTypes enables default values to be substituted at parse time.
    *
-   * For simple types xs:string and xs:hexBinary, the property dfdlx:emptyElementParsePolicy controls
+   * For simple types xs:string and xs:hexBinary, the property dfdl:emptyElementParsePolicy controls
    * whether the EmptyRep is allowed for strings and hexBinary. In required positions, when
-   * dfdlx:emptyElementParsePolicy is 'treatAsMissing', a required string/hexBinary that has EmptyRep
+   * dfdl:emptyElementParsePolicy is 'treatAsAbsent', a required string/hexBinary that has EmptyRep
    * causes a Parse Error, and an optional EmptyRep causes nothing to be added to the infoset (the empty string
-   * or hexBinary value is suppressed). When dfdlx:emptyElementParsePolicy is 'treatAsEmpty', a required
+   * or hexBinary value is suppressed). When dfdl:emptyElementParsePolicy is 'treatAsEmpty', a required
    * string/hexBinary with EmptyRep creates an empty string or zero-length byte array in the infoset.
    * An optional EmptyRep behaves differently depending on whether the EmptyRep is truly zero-length, or
    * dfdl:emptyValueDelimiterPolicy is such that EmptyRep is non-zero-length. When truly zero-length, no
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/MultipartBody.dfdl.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/MultipartBody.dfdl.xsd
@@ -53,8 +53,8 @@
 				dfdl:terminator="--%CR;%LF;">
 				<xsd:element name="BodyPart" dfdl:lengthKind="delimited"
 					minOccurs="1" maxOccurs="unbounded" type="xsd:string"
-					dfdl:occursCountKind="implicit" 
-          dfdlx:emptyElementParsePolicy="treatAsEmpty" />
+					dfdl:occursCountKind="implicit"
+				 	dfdl:emptyElementParsePolicy="treatAsEmpty" />
 			</xsd:sequence>
 		</xsd:complexType>
 	</xsd:element>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
@@ -193,12 +193,7 @@
       ref="ex:GeneralFormatPortable"
       representation="text"
       lengthKind="delimited"
-      separatorPosition="infix"
-      dfdlx:emptyElementParsePolicy="treatAsEmpty"/><!-- remove extension proerty for IBM cross tests -->
-    <!--
-    Note: dfdlx:emptyElementParsePolicy should become regular DFDL emptyElementParsePolicy
-    once implemented in DAFFODIL-2496. The enum 'treatAsMissing' is renamed to 'treatAsAbsent'
-    -->
+      separatorPosition="infix"/>
 
     <xs:element name="file1">
       <xs:complexType>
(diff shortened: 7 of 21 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
```diff
@@ -438,11 +438,11 @@ class ScalarOrderedSequenceChild(sq: SequenceTermBase, term: Term, groupIndex: I
 
   /**
    * Must deal with nils, emptyness and string/hexBinary exceptional behavior
-   * including the behavior for dfdlx:emptyElementParsePolicy 'treatAsMissing' which special cases
+   * including the behavior for dfdlx:emptyElementParsePolicy 'treatAsAbsent' which special cases
```
stevedlawrence, line 441: Should we remove the `x` in the `dfdlx` prefix here since the extension is deprecated?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
```diff
@@ -99,6 +99,7 @@
     <xs:restriction base="xs:string">
       <xs:enumeration value="treatAsEmpty" />
       <xs:enumeration value="treatAsMissing" />
+      <xs:enumeration value="treatAsAbsent" />
```
stevedlawrence, line 102: I don't think `treatAsAbsent` should be added here? This simple type restriction is used only for the dfdx extension property, which uses `treatAsMissing` instead of `treatAsAbsent`. If someone wants to use `treatAsAbsent`, let's force them to use the non-extension property.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
```diff
@@ -74,9 +74,9 @@ object ParseAttemptStatus {
    *
    * The EmptyRep for simpleTypes enables default values to be substituted at parse time.
    *
-   * For simple types xs:string and xs:hexBinary, the property dfdlx:emptyElementParsePolicy controls
+   * For simple types xs:string and xs:hexBinary, the property dfdl:emptyElementParsePolicy controls
    * whether the EmptyRep is allowed for strings and hexBinary. In required positions, when
-   * dfdlx:emptyElementParsePolicy is 'treatAsMissing', a required string/hexBinary that has EmptyRep
+   * dfdl:emptyElementParsePolicy is 'treatAsAbsent', a required string/hexBinary that has EmptyRep
    * causes a Parse Error, and an optional EmptyRep causes nothing to be added to the infoset (the empty string
    * or hexBinary value is suppressed). When dfdlx:emptyElementParsePolicy is 'treatAsEmpty', a required
```
stevedlawrence, line 81: `dfdlx` should be `dfdl` here.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
```diff
@@ -432,7 +432,9 @@ sealed trait EmptyElementParsePolicy extends EmptyElementParsePolicy.Value
 object EmptyElementParsePolicy extends Enum[EmptyElementParsePolicy] {
   case object TreatAsMissing extends EmptyElementParsePolicy
   case object TreatAsEmpty extends EmptyElementParsePolicy
-  override lazy val values = Array(TreatAsMissing, TreatAsEmpty)
+  case object TreatAsAbsent extends EmptyElementParsePolicy
+
+  override lazy val values = Array(TreatAsMissing, TreatAsEmpty, TreatAsAbsent)
 
   def apply(name: String, context: ThrowsSDE): EmptyElementParsePolicy = stringToEnum("emptyElementParsePolicy", name, context)
```
stevedlawrence, line 439: Around line 470 in this file, we have this comment:
```scala
// This property is an extension, so we don't want to require users to
// add this property just to silence this warning, especially since the
// property might change in the future. So silently use a default without
// outputting a warning. We may want to turn this warning on when the
// property makes its way into the official DFDL spec.
//SDW(
//  WarnID.EmptyElementParsePolicyError,
//  "Property 'dfdlx:emptyElementParsePolicy' is required but not defined, using tunable '%s' by default.",
//  defaultEmptyElementParsePolicy)
```
Do we want to re-enable this warning on now that this property is in the spec?

Do we also want to remove the defaulting behavior? I imagine no, since it might break a number of schemas? I guess after having this warning on for some number of releases we can flip the switch and require this property.

Do we want to add a value to DFDLGeneralFormat.dfdl.xsd? Is there an obvious choice or is it really dependent on the file format?

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -194,11 +194,7 @@
       representation="text"
       lengthKind="delimited"
       separatorPosition="infix"
-      dfdlx:emptyElementParsePolicy="treatAsEmpty"/><!-- remove extension proerty for IBM cross tests -->
-    <!--
-    Note: dfdlx:emptyElementParsePolicy should become regular DFDL emptyElementParsePolicy
-    once implemented in DAFFODIL-2496. The enum 'treatAsMissing' is renamed to 'treatAsAbsent'
-    -->
+      emptyElementParsePolicy="treatAsEmpty"/><!-- remove extension property for IBM cross tests -->
```
stevedlawrence, line 197: Does IBM support this property now? Do we no longer need to remove this property for cross tests to work?

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -282,14 +278,10 @@
       representation="text"
       lengthKind="delimited"
       separatorPosition="infix"
-      dfdlx:emptyElementParsePolicy="treatAsMissing"/> <!-- remove extension proerty for IBM cross tests -->
-    <!--
-    Note: dfdlx:emptyElementParsePolicy should become regular DFDL emptyElementParsePolicy
-    once implemented in DAFFODIL-2496. The enum 'treatAsMissing' is renamed to 'treatAsAbsent'
-    -->
+      emptyElementParsePolicy="treatAsAbsent"/> <!-- remove extension property for IBM cross tests -->
```
stevedlawrence, line 281: Same question here, does this still need to be removed for IBM?

stevedlawrence (commented):
Looks good! Just a couple minor comments and questions

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
mike-mcgann: This is a bit of a hack and I couldn't find a way to easily make this cleaner. Let me know if you can think of a better way. 

The issue is that we want to support the new and old attribute, for now, at the same time. An element could have a dfdl:emptyElementParsePolicy or a dfdlx:emptyElementParsePolicy but the property generator fabricates variables using the unprefixed name. Therefore we cannot easily have both of these properties with different types since they map to the same symbol. The schema validator correctly uses the empty/absent enumeration when evaluating the attribute with the dfdl prefix but then the parser uses the dfdlx enumeration definition (which is manually constructed and ignored by the property generator). The absent value was added to the dfdlx enumeration to support both new and old since there can be only one type. I guess the cleanest way would be to change the property generator to use fully qualified attribute names, or at least use them when necessary, but I'm not sure what the wider consequences of that change would be.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
stevedlawrence: @mbeckerle thoughts on the above questions?

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: I refactored this a bit.

Schema 4 now uses the old dfdlx:emptyElementParsePolicy and the IBM test for that is the same as before. I added a new test for daffodil only that checks that a parse error and a deprecation warning is issued. 

Schema 5 uses the new dfdl:emptyElementParsePolicy and that test only runs on daffodil. I'm not sure what happens on IBM for that.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/SequenceChild.scala
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
mike-mcgann: Updated

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
stevedlawrence: I think I'm still confused.

I understand why ByHandMixin.scala needs all three enumerations since both dfdl and dfdlx properties use the same EmptyElementParsePolicy object to convert a string to the type.

> but the property generator fabricates variables using the unprefixed name

The property generator excludes the EmptyElementParsePolicy type, so doesn't it completely ignore the definitions in this dfdlx.xsd file? So aren't the enumerations in this file just about doing DFDL schema validation and nothing else? And since treatAsAbsent isn't valid in he dfdlx property shouldn't it be removed here and only be defined in the DFDL_partX xsd file?

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mbeckerle: The property dfdl:emptyElementParsePolicy value 'treatAsEmpty' is an optional feature of DFDL.
IBM does not support 'treatAsEmpty' behavior, only 'treatAsAbsent'.

As for whether they added this property, let's test it. We have the IBM DFDL Cross tester.  https://github.com/OpenDFDL/ibmDFDLCrossTester

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
mike-mcgann: I think I was the one that was confused. I didn't realize that with the exclusion that the xsd file was completely ignored. I've removed that value from the schema and all the tests run fine. Thanks for that clarification.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
mbeckerle: We should re-enable this warning, yes. Making the property required right now would break people's schemas. 

We do need to also add this property to DFDLGeneralFormat,  with value treatAsEmpty. But this may need to be in Daffodil-only part of that if IBM DFDL doesn't tolerate the property.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
tuxji: Unfortunately, https://www.ibm.com/docs/en/integration-bus/10.0?topic=dfdl-unsupported-features does not say anything about emptyElementParsePolicy.

tuxji (changes requested):
I think this PR needs to re-enable a warning, and I hope it won't break any tests, but we need to check first.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/MultipartBody.dfdl.xsd
```diff
@@ -54,7 +54,7 @@
 				<xsd:element name="BodyPart" dfdl:lengthKind="delimited"
 					minOccurs="1" maxOccurs="unbounded" type="xsd:string"
 					dfdl:occursCountKind="implicit" 
-          dfdlx:emptyElementParsePolicy="treatAsEmpty" />
+                    dfdl:emptyElementParsePolicy="treatAsEmpty" />
```
mbeckerle, line 57: Indentation in this file is totally goofed up. Rather than fix it, and have whitespace fixing be part of this PR, let's accumulate whitespace fixups on a ticket. 

I already added this file to this ticket: https://issues.apache.org/jira/browse/DAFFODIL-2759

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
stevedlawrence: > we want to add the property to DFDLGeneralFormat.dfdl.xsd and give it the same 'treatAsEmpty' default value. 

I just thought of one potential issue when adding to DFDLGeneralFormat. If IBM DFDL doesn't support this new property then any schemas that used DFDLGeneralFormat.dfdl.xsd might now error when cross run with IBM? To ensure backwards compat, maybe we don't want to add the property to the file, especially since it's optional (as Mike says) and we already have built in defaulting behavior? Though as you point out, that would mean lots of schemas would need to manually add this property to suppress the warning.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -278,15 +274,11 @@
...
-      dfdlx:emptyElementParsePolicy="treatAsMissing"/> <!-- remove extension proerty for IBM cross tests -->
-    <!--
-    Note: dfdlx:emptyElementParsePolicy should become regular DFDL emptyElementParsePolicy
-    once implemented in DAFFODIL-2496. The enum 'treatAsMissing' is renamed to 'treatAsAbsent'
-    -->
+            ref="ex:GeneralFormatPortable"
+            representation="text"
+            lengthKind="delimited"
+            separatorPosition="infix"
+            dfdlx:emptyElementParsePolicy="treatAsMissing"/> <!-- remove extension property for IBM cross tests -->
```
mbeckerle, line 281: Should this be just no-prefix emptyElementParsePolicy='treatAsAbsent' now?

I probably wrote this test originally, but I don't understand the comment "remove extension property for IBM cross tests" since this is used by an IBM cross test below. It must be that IBM DFDL is able to ignore the dfdlx extensions or this wouldn't have worked in test test_sep_ssp_never_4_ibm.

So I think the comment line "remove extension property for IBM cross tests" is spurious now.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -278,15 +274,11 @@
...
 
     <!--
     treatAsMissing should have no effect here, because everything is optional.
```
mbeckerle, line 284: change comment to 'treatAsAbsent'.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -304,7 +296,6 @@
 
     <!-- Same, but has minOccurs=maxOccurs for the arrays.
      That makes all array elements "required"
-
      treatAsMissing causes this to fail. Required empty non-defaultable is an error in that case.
```
mbeckerle, line 299: change comment to treatAsAbsent, and remove parenthetic note about fixing 2496.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -334,7 +325,7 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
-  <tdml:parserTestCase name="test_sep_ssp_never_4" root="file2" model="s4"
+  <tdml:parserTestCase name="test_sep_ssp_never_4_ibm" root="file2" model="s4"
                        implementations="ibm">
     <tdml:document>madonna,,,,,,,,,</tdml:document>
     <!--
```
mbeckerle, line 331: I hate it when github won't let me put a comment on the line I want to. They have, in their infinite wisdom, descided that if you didn't change a line or the line isn't close enough to a change, no comment is allowed. 

Below a tdml:error is expecting 'treatAsMissing'. I believe this should change, per the comment, now.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
```diff
@@ -349,5 +340,73 @@
...
+                       implementations="daffodil">
+    <tdml:document>madonna,,,,,,,,,</tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>Empty element not allowed</tdml:error>
+      <tdml:error>required</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
 
 </tdml:testSuite>
```
mbeckerle, line 412: While you are at it, add a final new-line.

mbeckerle (changes requested):
Small clean ups I think, and we need to find out whether IBM DFDL accepts the real property name or not yet.

daffodil-lib/src/main/scala/org/apache/daffodil/schema/annotation/props/ByHandMixins.scala
tuxji: IBM should have its own DFDLGeneralFormat.dfdl.xsd so that schemas can import DFDLGeneralFormat.dfdl.xsd without worrying about problems like that, yes?  If the IBM cross tester is using Daffodil's DFDLGeneralFormat.dfdl.xsd with IBM's DFDL processor, that seems like a behavior which needs to be changed, don't you think?

stevedlawrence: DFDLGeneralFormat.dfdl.xsd is a thing Daffodil invented, I don't think IBM has any concept of it. DFDL schemas use the full path of `org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd `, which is daffodil specific and what the cross tester will use since that is what schemas import.

I guess we do have two different defineFormats: `DFDLGeneralFormat` and `DFDLGeneralFormatPortable`. We could put the new property in the DFDLGeneralFormat define, but many schemas (including most of our tests) don't use the "portable" variant and will probably fail if cross tested. Maybe that's okay?

mbeckerle: Yes, tests that are portable have to use GeneralFormatPortable because there are some other subtle incompatibilities between IBM DFDL and Daffodil - things about time zones (like Z vs. -00:00 suffixes equivalent values, but not identical, so the same test doesn't compare correctly), lack of support for encodingErrorPolicy 'replace', etc.

mbeckerle: To clarify, both GeneralFormat and GeneralFormatPortable are defined in DFDLGeneralFormat.dfdl.xsd. They share a base format with all the fully portable properties.

mike-mcgann: I've added the property to `DFDLGeneralFormat.dfdl.xsd` and re-enabled the warning. This required an update to a test where it was relying on the property to be missing and defaulting. I've updated that test to use a different schema instead of using the tuneable.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/MultipartBody.dfdl.xsd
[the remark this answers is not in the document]
mike-mcgann: Updated.

mike-mcgann: I went ahead and updated this from an earlier request.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: Added

tuxji (approved):
+1

Have you run the IBM cross tester successfully?  If so, changes look good now, but wait for Steve or Mike to review too.

stevedlawrence (approved):
+1 :+1: 

> Have you run the IBM cross tester successfully?

Yeah, it would be good to confirm that a schema using GeneralFormatPortable still works when used with IBM DFDL.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: @mbeckerle helped me out this morning getting the IBM test runner working and we found out that this property is not supported at this time. We updated the tests to reflect this.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: Updated during our meeting.

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: Updated during our meeting

daffodil-test/src/test/resources/org/apache/daffodil/usertests/SepTests.tdml
mike-mcgann: This is either updated or got revised out

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
```diff
@@ -103,7 +103,8 @@
       </dfdl:defineFormat>
       
       <dfdl:defineFormat name="GeneralFormat">
-        <dfdl:format ref="GeneralFormatOriginal" />
+        <dfdl:format ref="GeneralFormatOriginal"
+            emptyElementParsePolicy="treatAsEmpty"/>
```
stevedlawrence, line 107: I see you moved the property from GeneralFormatOriginal to GeneralFormat in this update. I guess this is so that the property is not defined in GeneralFormatPortable, so as long as schemas use the portabale format, IBM DFDL works as expected?

mike-mcgann: Yes.

mbeckerle (approved):
+1

tuxji:
I just noticed that this PR needs to be merged by a Daffodil committer.  I confirm that all PR comments have been resolved, @mike-mcgann has squashed the PR to one commit, and all tests have passed successfully.

PR #899, opened 2022-12-22
Fixed issue where text alignment was expected for enum

jadams-tresys:
This small change checks against the repElement (the dfdx:repType) instead of the represented element in the enumeration (typically a string).

In the past we would typically expect the alignment to line up with whatever encoding was used for strings, but when using 1-bit alignment, daffodil wouuld complain about the strings being unaligned even though they are internally represented by something other than a string that is 1-bit aligned.

Now we check to see what kind of representation (text or binary) the repElement is and check alignment based on that.

DAFFODIL-2596


jadams-tresys:
There may be other areas where repElement should be used in this code, but this definitely fixes the alignment issues that Mike B was seeing.

stevedlawrence (approved):
+1, but we really need to come up with a design where this kind of bug isn't possible. It wouldn't surprise me if there are other places where this kind of change is needed. And even if we track them all down, it's going to be too easy for new changes to not do this right.

mbeckerle (approved):
+1

PR #902, opened 2023-01-03
Added test to better demonstrate DAFFODIL-2596

jadams-tresys:
Previous test wasn't actually hitting this error condition that was being hit prior to commit f3ef3b9f5a5cddaf74c04d8d3026f8b99d72cbff. This test will trigger the issue. The faulty test has been removed.

DAFFODIL-2596

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
@@ -50,9 +50,6 @@
       </xs:complexType>
     </xs:element>
 
-    <!-- lengthKind="implicit" triggers DAFFODIL-2596 -->
-    <xs:element name="keysetValue_03" type="tns:keysetValue_01" dfdl:lengthKind="implicit"/>
-
     <xs:simpleType name="uint8" dfdl:lengthKind="explicit" dfdl:length="1">
       <xs:restriction base="xs:unsignedInt"/>
     </xs:simpleType>
@@ -183,23 +180,6 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
-  <tdml:parserTestCase name="RepType_lengthKind_inherited_01"
-    root="keysetValue_03" model="inputTypeCalc-Embedded.dfdl.xsd"
-    description="Verify that the lengthKind property is inherited from the repType, not the element using the repType">
-
-    <tdml:document>
-    <tdml:documentPart type="byte">
-    01
-    </tdml:documentPart>
-    </tdml:document>
-    <tdml:infoset>
-      <tdml:dfdlInfoset xmlns:xs="http://www.w3.org/2001/XMLSchema"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-        <keysetValue_03>one</keysetValue_03>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
-  </tdml:parserTestCase>
-
   <tdml:unparserTestCase name="InputTypeCalc_unparse_keysetValue_00"
     root="keysetValue_00" model="inputTypeCalc-Embedded.dfdl.xsd" description="Extensions - inputTypeCalc keysetValue transform">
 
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -256,22 +236,6 @@
     </tdml:infoset>
   </tdml:unparserTestCase>
 
-  <tdml:unparserTestCase name="InputTypeCalc_unparse_keysetValue_03"
-    root="keysetValue_03" model="inputTypeCalc-Embedded.dfdl.xsd" description="Extensions - inputTypeCalc keysetValue transform">
```
tuxji, line 260: I have a question for @mbeckerle and @stevedlawrence.  Are there any other reasons for adding an unparserTestCase besides expecting an error on the unparse (the error case) or checking only the unparse (the roundTrip="none" case)?  An unparserTestCase is not allowed to check the twoPass / threePass cases, only the none / onePass cases, and I can't see any need for having a pair of identical parserTestCase and unparserTestCase tests both checking successful onePass cases.

In this TDML file, we should need only the parserTestCase tests because the tdml:testSuite has no defaultRoundTrip attribute and each test case has no roundTrip attribute.  The default behavior if nothing is specified in either the tdml:testSuite or the test case itself is onePass.   So, the parserTestCase tests parse the given data, compare to expected infoset, expect a match, and also unparse the new infoset, compare to given data, and expect a match.  The unparserTestCase tests just do the unparse, compare first followed by the parse, compare in the opposite order, and the order should be unimportant if you expect a successful onePass roundtrip.  Is my understanding correct?  

If so, most of the Daffodil TDML test cases should be parserTestCase tests (I count 3958 of them in Daffodil) and the need for unparserTestCase tests should be rare (I count 538 of them in Daffodil).

tuxji (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
stevedlawrence: > The default behavior if nothing is specified in either the tdml:testSuite or the test case itself is onePass. 

I think the default is actually "none":

https://github.com/apache/daffodil/blob/main/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/RunnerFactory.scala#L92

So since the `Runner` and TDML files don't specify value, all the parserTestCases do not round trip and so we do need the unparserTestCases. We could potentially change the default for these files and remove those test unparseTestCase, but that would require some testing and confirmation that they are actually the same.

But your reasoning is all correct for onePass tests. In general, we should only need an unparserTestCase if roundTrip=onePass fails because the unparse doesn't exactly match the original data.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -183,23 +180,6 @@
...
-    </tdml:documentPart>
-    </tdml:document>
-    <tdml:infoset>
-      <tdml:dfdlInfoset xmlns:xs="http://www.w3.org/2001/XMLSchema"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-        <keysetValue_03>one</keysetValue_03>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
-  </tdml:parserTestCase>
-
```
stevedlawrence, line 202: Are these tests faulty or do they just not trigger the particular issue? The schema seems correct to me and might be worth keeping if it's testing something different than other tests.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
tuxji: Oh, I had read the statement "The default behavior if nothing is specified in either the tdml:testSuite or the test case itself is onePass" in the last section, Round-Trip Testing, of the Daffodil website's [Test Data Markup Language (TDML)](https://daffodil.apache.org/tdml/) documentation.  Since that statement and the source code disagree with each other, we need to fix one of them.  I assume changing defaultRoundTripDefaultDefault in the source code will break lots of TDML tests so I will open a JIRA issue on the documentation page.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
stevedlawrence: Is there value in keeping the tests? Do they test anything different? I assume at point they were broken and have since been fixed (maybe not with your recent change though?) so maybe worth keeping?

jadams-tresys: In the end I don't think it was testing anything different than the previously existing keyset value tests.  So the test I removed is unnecessary.

stevedlawrence (approved):
+1

PR #905, opened 2023-01-06
Fix mixed content error in SAX unparse

stevedlawrence:
Currently, when mixed content is found when unparsing with the SAX API, the DaffodilUnparseContentHandler throws an IllegalContentWhereEventExpected exception. This has two issues:

1. A SAX XMLReader API only expects SAXExceptions to be thrown from a ContentHandler
2. Throwing the exception here ends execution of the ContentHandler subroutine without signaling the unparse() subroutine/thread. This means the unparse thread is deadlocked waiting for more events that will never come.

To solve these issues, instead of throwing the exception, this just adds the mixed content string to the current SAXInfosetEvent. When the SAXInfosetInputter reads that event and sees the mixed content, it will throw the IllegalContentWhereEventExpected exception. This is now thrown in he unparse thread, which causes the unparse thread to stop unparsing and create an UnparseResult, send it to the ContentHandler thread, and cleanly exit. The ContentHandler thread can then receive the UnparseResult and convert it to a SAXException.

Also update a number of comments to make certain sections more clear.

DAFFODIL-2767

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
@@ -26,6 +26,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.xml.sax.InputSource
+import org.xml.sax.SAXException
 import org.xml.sax.SAXParseException
 import org.xml.sax.XMLReader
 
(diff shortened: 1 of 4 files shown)


stevedlawrence:
As I re-reviewed this, I got thinking that this may not completely solve the hanging thread problem (though I think this fix is still correct and needed).

For example, say we have an XMLReader parsing some XML and sending events to our DaffodilUnparseContentHandler, and assume it's sent a bunch of events and the coroutine stuff is doing it's thing with a separate thread. Then the XMLReader hits somthing that isn't valid and the *XMLReader* (not our ContentHandler) throws a SAXException. When this happens, it doesn't notify the ContentHandler in any way that it's not going to get anymore events. In our specific case, that means our coroutine thread is going to hang since it still things more threads are on the way.

In the generic case it means there's no way for a ContentHandler to know when to clean up state in an error case. It can't differentiate between "more events sure are taking a long time to arrive" vs "no more events are ever coming". Seems like a flaw in the SAX API...

I guess we could have some function that users must call if the XMLReader throws an exception that would allow us to shutdown the coroutine thread. But I don't think there's any way to force a user to call that function, so it's not a great API...

Any other thoughts?

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
```diff
@@ -100,6 +101,26 @@ class TestSAXUnparseAPI {
...
+    val unparseContentHandler = dp.newContentHandlerInstance(wbc)
+    xmlReader.setContentHandler(unparseContentHandler)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACES_FEATURE, false)
+    xmlReader.setFeature(XMLUtils.SAX_NAMESPACE_PREFIXES_FEATURE, true)
+    val infoset = <p:list xmlns:p="http://example.com" ignored="attr"><p:w>9</p:w><p:w>1</p:w><p:w>0</p:w></p:list>
+    val bai = new ByteArrayInputStream(infoset.toString.getBytes)
+    xmlReader.parse(new InputSource(bai))
+    val ur = unparseContentHandler.getUnparseResult
+    assertTrue(!ur.isError)
+    assertEquals(testData, bao.toString)
```
tuxji, line 121: I searched this file but I can't find testData's definition.  How is testData defined?  Also, please make the comment explain why this test's case matters compared to the 2 new tests you've already added below:

1. testUnparseContentHandler_unparse_mixed_01 to test the case of unparsing with mixed content before a start element.
2. testUnparseContentHandler_unparse_mixed_02 to test the case of unparsing with mixed content before an end element.

tuxji (approved):
+1 on what's here so far, even though it may be an incomplete fix to the handing thread problem

I'm not sure how to solve the situation where an XMLReader hits something that isn't valid, throws an exception itself, doesn't notify the ContentHandler that no more events are coming, and the coroutine thread hangs since it thinks more events are coming.  However, things become easier to figure out when you have a test which reproduces the hanging thread problem.  Then you can set breakpoints, run the test in the debugger, and experiment with how to fix the flow of control.  I would suggest you add a new failing test in this pull request if it's easy to write the test and defer fixing the test to a future PR if it's hard to fix the test.

mbeckerle:
I guess the assumption in the SAX API is that the handlers are stateless, or don't require a shutdown/reset call. One would like to not have to define APIs to accomodate statefulness and resources needing to be released, but which just discard objects. It's a garbage collected language after all. But alas, it seems this is not adequate, as GC cleanup does not release resources like threads. 

My guess is we need to add a reset/cleanup method, and point out that if it isn't called, that the thread will leak.  In our own test rig/CLI we need to find a place to call this reset/shutdown. 

I suspect many applications are not going to care. I suspect they will not create many of these, and their lifetime will be the lifetime of the entire application/JVM. 

When we do message-streaming it is important for unparse to not construct new handlers (therefore threads) on every call to parse/unparse one message.

mbeckerle (approved):
+1

daffodil-core/src/test/scala/org/apache/daffodil/processor/TestSAXUnparseAPI.scala
stevedlawrence: `testData` comes from an `import TestSAXUtils._` at the top of the file. There's a number of TestSAX* files that do this to test the SAX API. It's definitely not very clear.

This test case is actually to improve missing coverage that I found when was doing SAX performance stuff. The CI shows it adds about +.04% coverage.

I actually have one more SAX related change in the works to refactor this SAX stuff a bit. It made sense we originally added it but I found it confusing coming back to this code years later to investigate performance. So while it was fresh in my mind I wanted to make the code more clear with more comments. I'll move this test to that PR, since it really has nothing to do with mixed content.

stevedlawrence:
> I would suggest you add a new failing test in this pull request if it's easy to write the test and defer fixing the test to a future PR if it's hard to fix the test.

It's not really specific to mixed content, and I'm not sure how easy it will be to find a reproducable test. I'll create a new ticket for this issue (I think something like reset is probably the right approach as Mike suggests) and see about creating a reproducable test and a fix soon.

PR #907, opened 2023-01-09
Properly handle dfdlx:repValues=''

jadams-tresys:
This should have been failing schema validation, but instead would result in an error when we attempted to pasre a number from an empty string.

DAFFODIL-2632

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enumInvalid.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enumInvalid.tdml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


mbeckerle (changes requested):
Not passing all tests.

jadams-tresys:
Forgot to add the new tdml file I created when I did a git commit.  Should all pass now.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enumInvalid.tdml
```diff
@@ -0,0 +1,69 @@
...
+  </tdml:defineSchema>
+
+  <parserTestCase name="emptyRepValues" model="s3" root="r2" validation="off">
+    <document>
+      <documentPart type="bits">01010101</documentPart>
+    </document>
+    <errors>
+      <error>Schema Definition Error</error>
+      <error>'' is not a valid value</error>
+      <error>NonEmptyListofDFDLStringLiteralOrNonEmptyListOfInteger</error>
```
stevedlawrence, line 65: Should we check that `%ES;` is in the error message? The whole point of this special type is so we can give a better message, so we should verify that the unique part is included and it actually is more helpful.

mbeckerle (approved):
+1 Once all tests pass I approve.

I have nothing to add beyond the one comment by SL about tightening up the error message in the TDML a bit.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enumInvalid.tdml
jadams-tresys: The %ES; check had only been for nilValue.  Does it make sense to have a dfdlx:repValues="%ES;"?  If it does and we want to suggest that, this will need more changes and we would have to go back to ignoring the minLength restriction and add a check for empty string before attempting to parse as an number.

stevedlawrence: Ah, you're right, this is a test for repValues. I don't think %ES; is valid for repValues so what you have is probably fine. Presumably we already have a test that checks for the ES error message when nilValue="".

stevedlawrence (approved):
+1

PR #909, opened 2023-01-11
Implement 'P' textNumberPattern feature.

mbeckerle:
Also a Cobol feature.

Also this change set incorporates additional review feedback on the 'V' feature change set.

Note that the 'P' and 'V' feature does not support some obscure textNumberPattern as yet such as:

1. a textNumberPattern with a trailing 'V' (no 0 digits after)

2. combining 'V' with an exponent specification.

Those usages are allowed technically by the DFDL v1.0 specification (the grammar in section 13.6.1.1 labeled Figure 4), but that usage is so obscure that I am NOT going to create a bug report about implementing these behaviors until someone complains.

DAFFODIL-2763


mbeckerle:
Abandoning. Forgot to rebase on latest first.

PR #911, opened 2023-01-12
Implement 'P' textNumberPattern feature.

mbeckerle:
The 'P' is also a COBOL feature from their picture clauses.

Also this change set incorporates additional review feedback on the 'V' feature change set.

Note that the 'P' and 'V' feature does not support some obscure textNumberPattern as yet such as:

1. a textNumberPattern with a trailing 'V' (no 0 digits after)

2. combining 'V' with an exponent specification.

Those usages are allowed technically by the DFDL v1.0 specification (the grammar in section 13.6.1.1 labeled Figure 4), but that usage is so obscure that I am NOT going to create a bug report about implementing these behaviors until someone complains.

Similarly it is technically possible to use say, CR (carriage return) or ASCII NUL as your plus sign
character. I claim that is a mistake in the DFDL spec to allow that, so use of line-ending characters is not allowed.

DAFFODIL-2763

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
@@ -18,6 +18,7 @@
 package org.apache.daffodil.grammar.primitives
 
 import com.ibm.icu.text.DecimalFormat
+import org.apache.daffodil.api.WarnID
 import org.apache.daffodil.cookers.EntityReplacer
 import org.apache.daffodil.dpath.NodeInfo.PrimType
 import org.apache.daffodil.dsom._
@@ -45,27 +46,56 @@ case class ConvertTextCombinator(e: ElementBase, value: Gram, converter: Gram)
   override lazy val unparser = new ConvertTextCombinatorUnparser(e.termRuntimeData, value.unparser, converter.unparser)
 }
 
-// This is a separate object for unit testing purposes.
+/**
+ * This is a separate object so that we can easily unit test these subtle regular
+ * expressions.
+ */
 private[primitives]
 object TextNumberPatternUtils {
+
+  //  DFDL v1.0 Spec says
+  //  It is a Schema Definition Error if any symbols other than "0", "1" through "9" or #
+  //  are used in the vpinteger region of the pattern.
+  //
+  // The prefix and suffix chars can surround the vpinteger region, and there can be
+  // a positive and a negative pattern.
+
+  /**
+   * The prefix and suffix are quoted any single character
+   * or a single non-special character (excepting + and - which are allowed
+   * though they appear in the table of special characters for textNumberFormat).
+   *
+   * By default these regex will not allow line-ending chars either.
+   */
+  val prefixChars = """(?:'.'|[^0-9#PVE\.\,\;\*\@\'])?""" // same for suffix.
+
+  /**
+   * sharps and at least 1 digit before the V.
+   */
+  val beforeVChars = """#*[0-9]+"""
+
+  /**
+   * one or more digits after the V. (No trailing sharps)
+   */
+  val afterVChars = """[0-9]+"""
+
+  /**
+   * The negative pattern must be a valid pattern, but
+   * does NOT have to match the positive one, as it is
+   * always ignored.
+   *
+   * The shortest possible match is just a # or just a digit.
+   * If a V appears it must have a digit on either side.
+   */
+  val negVNumberChars = """#|#+|#*[0-9]+(?:V[0-9]+)?"""
+
 
   /**
    * A regex which matches textNumberPatterns that legally use the V
    * (implied decimal point) character.
    */
-   private[primitives] lazy val vregexStandard = {
-    //  DFDL v1.0 Spec says
-    //  It is a Schema Definition Error if any symbols other than "0", "1" through "9" or #
-    //  are used in the vpinteger region of the pattern.
-    //
-    // The prefix and suffix chars can surround the vpinteger region, and there can be
-    // a positive and a negative pattern.
-    //
-    // The prefix and suffix cannot be digits, # or P or V. No quoted chars in prefix either.
-    //
-     val prefixChars = """[^0-9#PV']?""" // same for suffix.
-     val beforeVChars = """#*[0-9]+"""
-     val afterVChars = """[0-9]+"""
+  lazy val vRegexStandard = {
+
      //
      // Each of the capture groups is defined here in order
      //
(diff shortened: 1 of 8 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
```diff
@@ -89,21 +92,55 @@ trait TextDecimalVirtualPointMixin {
   }
   // $COVERAGE-ON$
 
+
+  /**
+   * Always creates an integer from a JFloat, JDouble, or JBigDecimal
+   * by scaling the argument by the virtual decimal point scaling factor.
+   *
+   * Aborts if the result of scaling is not an integer. That is, the virtual decimal
```
mbeckerle, line 100: Fix scaladoc. Doesn't abort. See discussion at line 119.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -186,41 +305,57 @@ trait ConvertTextNumberMixin {
...
+   *
+   * Value is the number of digits to the right of the 'V' for V patterns.
+   *
+   * For P patterns, the value is the number of P characters when
+   * the P chars are on the right of the digits, or it is negated if the
+   * P chars are on the left of the digits.
+   *
+   * Examples:
+   *
+   * "000.00" returns 2, so text "12345" yields 123.45
```
stevedlawrence, line 337: Same question here, should this be `000V00`?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -186,41 +305,57 @@ trait ConvertTextNumberMixin {
...
+    if (hasV) {
       //
       // check for things incompatible with "V"
       //
-      val virtualPoint = textDecimalVirtualPointFromPattern
       Assert.invariant(virtualPoint >= 1) // if this fails the regex is broken.
       virtualPoint
+    } else if (hasP) {
+      //
+      // check for things incompatible with "V"
```
stevedlawrence, line 353: Should be `incompatiable with "P"` instead of "V"?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -275,17 +410,67 @@ case class ConvertTextStandardNumberPrim(e: ElementBase)
...
+      r.findFirstMatchIn(patternNoQuoted) match {
+        case Some(r(_, _, afterV, _, _, _, _, _)) => afterV.length
+        case None =>
+          e.SDE(
+            s"""The dfdl:textNumberPattern '%s' contains 'V' (virtual decimal point).
+               | Other than the sign indicators, it can contain only
+               | '#', then digits 0-9 then 'V' then digits 0-9.
+               | The positive part of the dfdl:textNumberPattern is mandatory.""".stripMargin('|'),
+            pattern)
+      }
```
stevedlawrence, line 426: Just trying to think about edge cases, assume a pattern is invalid and it has both `V` and `P`, e.g. `P00V00`. In this case both `hasV` and `hasP` will be true, and we'll fall into this `hasV` block, the regex will fail and we'll say it's not a valid V pattern? Seems like a reasonable way to error, just double checking that I'm understanding this all right.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -275,17 +410,67 @@ case class ConvertTextStandardNumberPrim(e: ElementBase)
...
+   *
+   * @param ps        match string for positive pattern P characters
+   * @param digits    match string for positive pattern digits (and sharps)
+   * @param negPre    negative prefix match (null if there is no negative pattern at all)
+   * @param negPs     match string for negative pattern P character (or null)
+   * @param negDigits match string for negative pattern P digits (and sharps)
+   */
+  private def checkPPPMatchSyntax(ps: String, digits: String, negPre: String, negPs: String, negDigits: String): Unit = {
+    Assert.invariant(ps.length >= 1)
+    digits.length >= 1
```
stevedlawrence, line 464: Was this intended to be an `Assert(digits.length >= 1)`. I'm surprises Scala didn't give a compiler warning about this line not doing anything. Maybe it doesn't know if `.length` has side effects or not?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -275,17 +410,67 @@ case class ConvertTextStandardNumberPrim(e: ElementBase)
...
+    if (negPre ne null) {
+      e.schemaDefinitionUnless(ps == negPs,
+        """In the dfdl:textNumberPattern '%s' the P Symbols do not match
+          | between positive and negative patterns""".stripMargin,
+        pattern)
+      e.schemaDefinitionUnless(digits == negDigits,
+        """In the dfdl:textNumberPattern '%s' the '#' and digit symbols do not match
+          | between positive and negative patterns""".stripMargin,
+        pattern)
     }
```
stevedlawrence, line 474: Looking back at the spec I saw this line:

> If there is an explicit negative subpattern, it serves only to specify the negative prefix and suffix; the number of digits, minimal digits, and other characteristics are ignored in the negative subpattern

Which gives me two questions:
1. Should we allow the negative and positive P's to be different? Seems like it's probably a user doing something they didn't intend if that's the case so maybe it's is worth an SDE as you have. Or at the very least this should be a warning.
2. Similarly, should the V and P regex patterns be much more lax about what they allow in the negative part since it's ignored? Again, if they don't match it probably an accident and an SDE seems very reasonable, but maybe there are schemas that will just want to keep the textNumberPattern as simple as possible and do something like `+000V0000;-0` so that they can specific the virtual decimal point in the postive pattern, and then just the prefix/suffix in the negative pattern and not worry about them having to match.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -114,21 +191,66 @@ object TextNumberPatternUtils {
...
-   * This is broken out separately for unit testing purposes.
+   * This method is here in this object for unit testing purposes.
    *
    * @param patternStripped the dfdl:textNumberPattern pattern - with all quoting removed.
    * @return None if the pattern is illegal syntax for use with V (virtual decimal point)
    *         otherwise Some(N) where N is the number of digits to the right of the V character.
    */
-  private[primitives] def textDecimalVirtualPointForZoned(patternStripped: String): Option[Int] = {
-    val r = TextNumberPatternUtils.vregexZoned
+  def textNumber_V_DecimalVirtualPointForZoned(patternStripped: String): Option[Int] = {
```
stevedlawrence, line 200: Since this is no longer `private[primitives]`, does it make sense to move the regex match logic to the `textDecimalVirtualPointFromPattern` val in `PrimivitesZoned.scala`? That would make the different `textDecimalVirtualPointFromPattern` implementations look more similar instead of one calling this function and the other having the logic inline.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
```diff
@@ -89,21 +92,52 @@ trait TextDecimalVirtualPointMixin {
   }
   // $COVERAGE-ON$
 
+
+  /**
+   * Always creates an integer from a JFloat, JDouble, or JBigDecimal
+   * by scaling the argument by the virtual decimal point scaling factor.
+   *
+   * If the argument is some other JNumber, i.e., not a JFloat, JDouble, or JBigDecimal
+   * it can only be an integer type. This method this returns the argument value
```
stevedlawrence, line 101: Should be "This method returns"

stevedlawrence (approved):
+1 :+1: just some minor comments/questions

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -80,14 +99,72 @@ object TextNumberPatternUtils {
      // don't forget the ^ and $ (start of data, end of data) because we want
      // the match to consume all the characters, starting at the beginning.
     val vPattern=
-     s"""^${posPrefix}${posBeforeV}V${posAfterV}${posSuffix}(?:;${negPrefix}${negBeforeV}V${negAfterV}${negSuffix})?$$"""
+     s"""^${posPrefix}${posBeforeV}V${posAfterV}${posSuffix}""" ++
+       s"""(?:;${negPrefix}${negBeforeV}(?:V${negAfterV}${negSuffix})?)?$$"""
```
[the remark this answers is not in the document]
tuxji: I also noticed that this long-line change uses ++ (the sequence append method) to concatenate the two strings together.  Usually we use + (the string concat method) to concatenate two strings in Scala, same as in Java.  Using ++ works, but it looks surprising to programmers like me and the Scala compiler may generate less optimal code.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -186,41 +305,57 @@ trait ConvertTextNumberMixin {
    * but cannot be used as an operational pattern given that
    * potentially lots of things have been removed from it.
    */
-  final protected lazy val patternStripped = {
+  final protected lazy val patternNoQuoted = {
```
tuxji, line 308: Did you mean `patternNoQuotes`, which would tell the programmer more clearly that all quotes will be removed, not just one pair of quotes?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -275,17 +410,67 @@ case class ConvertTextStandardNumberPrim(e: ElementBase)
...
+      val rr = TextNumberPatternUtils.pOnRightRegexStandard
+      val rl = TextNumberPatternUtils.pOnLeftRegexStandard
+      val rightMatch = rr.findFirstMatchIn(patternNoQuoted)
+      val leftMatch = rl.findFirstMatchIn(patternNoQuoted)
+      (leftMatch, rightMatch) match {
+        case (None, None) => e.SDE(
+        """The dfdl:textNumberPattern '%s' contains 'P' (virtual decimal point positioners).
+            |However, it did not match the allowed syntax which allows the sign indicator
+            |plus digits on only one side of the P symbols.""".stripMargin,
           pattern)
```
tuxji, line 437: |'s don't line up with """ (could also start the first line with | and align the rest of the |'s to the first |).

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
```diff
@@ -34,7 +32,8 @@ import org.apache.daffodil.processors.Success
 import org.apache.daffodil.processors.TermRuntimeData
 import org.apache.daffodil.processors.TextNumberFormatEv
 
-import java.lang.{Long => JLong, Float => JFloat, Double => JDouble, Number => JNumber}
+import java.lang.{Float => JFloat, Number => JNumber, Long => JLong, Double => JDouble}
```
tuxji, line 35: Should these be on separate lines?

tuxji (approved):
+1

mbeckerle (commented):
All current issues should be fixed now.

PR #917, opened 2023-01-18
Improved diagnostic message for enumeration invalid key/value

jadams-tresys:
Before attempting to pass an invalid value in an enumeration would result in an error referencing a keyset-value mapping with no mention of the enumeration itself. This should more obviously point to the fact that the given value is not present in the enumeration.

DAFFODIL-2607

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
@@ -153,7 +153,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
       })
       ans1 match {
         case None => {
-          (DataValue.NoValue, One(s"Key ${x} not found in keyset-value mapping"))
+          (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues"))
         }
         case Some((_, _, v)) => (v, Maybe.Nope)
       }
@@ -163,7 +163,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration"))
     }
   }
 
@@ -173,20 +173,16 @@ class KeysetValueTypeCalculatorUnordered(valueMap: HashMap[DataValuePrimitive, D
   extends TypeCalculator(srcType, dstType) {
 
   override def inputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
-    if (valueMap.contains(x)) {
-      valueMap.get(x) match {
-        case Some(a) => (a, Maybe.Nope)
-        case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
-      }
-    } else {
-      (DataValue.NoValue, One(s"Key ${x} not found in keyset-value mapping"))
-
+    valueMap.get(x) match {
+      case Some(a) => (a, Maybe.Nope)
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues"))
     }
   }
+
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration"))
     }
   }
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
@@ -295,6 +295,15 @@
       </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="enumWithRange" dfdlx:repType="tns:enumVal">
+      <xs:restriction base="xs:string">
+        <xs:enumeration value="VALUE=0" dfdlx:repValues="0"/>
+        <xs:enumeration value="VALUE=1" dfdlx:repValues="1"/>
+        <xs:enumeration value="VALUE=2" dfdlx:repValueRanges="4 5"/>
+      </xs:restriction>
+    </xs:simpleType>
+
+
     <xs:element name="root">
       <xs:complexType>
         <xs:sequence>
@@ -302,6 +311,15 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
+
+    <xs:element name="r2">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int2" type="tns:enumWithRange"/>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
   </tdml:defineSchema>
 
   <tdml:parserTestCase name="inherited_LengthKind"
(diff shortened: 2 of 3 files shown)


jadams-tresys:
Small change and really surprised we didn't have any existing tests that checked for this error situation where we attempt to use an invalid value for an enumeration.  I checked against the dfdl-regression test suite and it seems that pretty much everything passes.  I believe there may be a couple tests in P8 that need updating to not check for "keyset-value" anymore.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -176,17 +176,17 @@ class KeysetValueTypeCalculatorUnordered(valueMap: HashMap[DataValuePrimitive, D
...
-      (DataValue.NoValue, One(s"Key ${x} not found in keyset-value mapping"))
+      (DataValue.NoValue, One(s"Key ${x} not found in enumeration key-value mapping"))
 
     }
   }
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration key-value mapping"))
```
stevedlawrence, line 189: I wonder if we can be a bit more precise and thus more helpful? My concern is "key-value mapping" isn't a term a schema developer might know since none of the properties use those terms.

I think when parsing we are looking for a value in `dfdlx:repValues`, and when unparsing we are looking for a string in the enumeration `value`? Is it worth differentiating and saying something like this when parsing:

> Value ${x} not found in enumeration dfdlx:repValues properties

And this for unparsing:

> Value ${x} not found in enumeration value properties

I'm not sure I like that wording a ton better, but I think mentioning value/repValues might help to point the user at the right attribute that we are looking for a match but didn't find?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -153,7 +153,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
       })
       ans1 match {
         case None => {
-          (DataValue.NoValue, One(s"Key ${x} not found in keyset-value mapping"))
+          (DataValue.NoValue, One(s"Key ${x} not found in enumeration key-value mapping"))
```
stevedlawrence, line 156: I guess we were missing this coverage in the past, but it might be nice to add additional tests to improve coverage of all these different error conditions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
mbeckerle: You can stick "(keyset-value mapping)" at the end of the messages to maintain backward compatibility with past message-based tests maybe. So long as this isn't needed to understand the primary diagnostic message. I agree with SL that this should mention enumeration, mention dfdlx:repValues (I would not search for the dfdlx prefix in tests though since that's likely to go away some day), and mention being unable to determine the repValue for the unrecognized enumeration value for unparsing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
mbeckerle, line 156: I would not even use the term "key" here. 

Parsing we're looking for the repValue of the enumeration.  So like "The value '993' was not found in any dfdlx:repValues for the enumeration type."

Unparsing we're looking for the enumeration value (to determine the corresponding repValue).  So like "The enumeration value 'xyzzy' was not defined."

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -321,4 +321,21 @@
...
+    <tdml:document>
+    <tdml:documentPart type="byte">
+    00 00 00 03
+    </tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Value</tdml:error>
+      <tdml:error>3</tdml:error>
+      <tdml:error>not found</tdml:error>
+      <tdml:error>enumeration</tdml:error>
```
mbeckerle, line 337: This should be mentioning "repValues" in the errors also.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml
```diff
@@ -321,4 +321,21 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+  <tdml:parserTestCase name="valueNotFound"
```
mbeckerle, line 324: unparser test case also needed.

mbeckerle (changes requested):

jadams-tresys:
Comments should be addressed now.  I actually had an unparser test case before, but it was triggering the same error message so I didn't think it was adding any additional coverage.  I was expecting to get the "key not found" error, but I think the only way that happens is when we are expecting a different type (ie expecting an int and get a string).

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -153,7 +153,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
       })
       ans1 match {
         case None => {
-          (DataValue.NoValue, One(s"Key ${x} not found in enumeration key-value mapping"))
+          (DataValue.NoValue, One(s"Key ${x} not found in enumeration dfdlx:repValues keyset-value mapping"))
```
mbeckerle, line 156: I suggest drop the words Key and Keyset entirely, if you want them in the message for compatiblity reasons add literally "\n(formerly keyset-value mapping)" at the end of the message, but don't otherwise use this terminology.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -163,7 +163,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues keyset-value mapping"))
```
mbeckerle, line 166: These coverage errors should definitely go away now right?

mbeckerle (approved):
+1 Suggest one more tweek to wording. I think coverage should be fixed now too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -176,17 +176,17 @@ class KeysetValueTypeCalculatorUnordered(valueMap: HashMap[DataValuePrimitive, D
...
-      (DataValue.NoValue, One(s"Key ${x} not found in keyset-value mapping"))
+      (DataValue.NoValue, One(s"Key ${x} not found in enumeration dfdlx:repValues keyset-value mapping"))
 
     }
   }
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues keyset-value mapping"))
```
stevedlawrence, line 189: These all mention repValues, but don't some of these look for the actual enumeration `value` instead for unparsing?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
stevedlawrence: I suggest we just change tests to remove `keyset` or `mapping` if any look for those words instead of including the "formerly..." wording in the error message. We don't guarantee compatibility with error messages for users, and that part of the error message is confusing and probably isn't worth keeping for compatibility.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
jadams-tresys: As far as I can tell, the only place referencing the keyset-mapping error message is like 2 tests in P8, which I was about to update anyway with some small changes from SBSA, so I'll just update the tests there.  None of the other schemas in the regression test repo reference that error message.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
jadams-tresys: Fixed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
jadams-tresys: With my most recent commit which adds 2 more tests coverage is good for all of these error cases except for line 179.  I might be missing something, but I'm not sure it's possible to hit that line.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -176,17 +176,17 @@ class KeysetValueTypeCalculatorUnordered(valueMap: HashMap[DataValuePrimitive, D
     if (valueMap.contains(x)) {
       valueMap.get(x) match {
         case Some(a) => (a, Maybe.Nope)
-        case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+        case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues"))
```
mbeckerle, line 179: This is unreachable. If valueMap.contains(x), then valueMap.get(x) will not return None. 
You  don't need both an if and a match case here. You valueMap.get(x) returns an Option, so you can just do the match-case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
```diff
@@ -163,7 +163,7 @@ class KeysetValueTypeCalculatorOrdered(valueMap: HashMap[DataValuePrimitive, Dat
   override def outputTypeCalc(x: DataValuePrimitive, xType: NodeInfo.Kind): (DataValuePrimitiveNullable, Maybe[Error]) = {
     unparseMap.get(x) match {
       case Some(v) => (v, Maybe.Nope)
-      case None => (DataValue.NoValue, One(s"Value ${x} not found in keyset-value mapping"))
+      case None => (DataValue.NoValue, One(s"Value ${x} not found in enumeration dfdlx:repValues"))
```
tuxji, line 166: Is there a reason why the message here at line 166 is not identical to the message at line 156?  That is, why does the other message mention only "enumeration" while this message mentions both "enumeration" and "dfdlx:repValues"?

BTW, I think the order should be reversed: "not found in dfdlx:repValues enumerations".

tuxji (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/TypeCalculator.scala
jadams-tresys: You are right that they should be reversed.  Had the order mixed up in my head.  They aren't identical as on the inputTypeCalc side we are looking for a repValue and the outputTypeCalc side we are looking at the enumeration value.

PR #918, opened 2023-01-18
Emit warning when lengthUnits="bits" is used with bigints.

mike-mcgann:
The DFDL spec prohibits the use of lengthUnits="bits" for types xs:integer and xs:nonNegativeInteger. This update will now emit a DeprecatedTypeUse warning in this case and will be changed to an error in a future release.

Deprecation/Compatibility:

Previous Daffodil releases let schemas define every type's length using "bits" as the length unit even though the specification allows bit length units only for a specific set of types' binary representations and does not allow bit length units for any other type's binary representation or any type's text representation. A deprecation warning is now issued
when bit length units are incorrectly used. 

[DAFFODIL-1172](https://issues.apache.org/jira/browse/DAFFODIL-1172)


daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
```diff
@@ -903,4 +903,37 @@
...
+        0f
+      </tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>f
+        <r1>15</r1>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:warnings>
+      <tdml:warning>lengthUnits='bits' will only be supported for integer types of 64 bits or less</tdml:warning>
```
mbeckerle, line 935: I think the message should not say "integer types of 64 bits or less" but say lengthUnits bits is not supported for xs:integer nor xs:nonNegativeInteger. 

Why.. because this is perfectly legal: <element name="x" type="xs:integer" dfdl:length="3" dfdl:lengthUnits="bytes"/> which is a 24-bit integer. So is that an integer of type 64 bits or less?

It's the types that cause the issue, not the 64. The restriction was to make DFDL easier to implement as a byte-centric big-num library would suffice for xs:integer and xs:nonNegativeInteger.

daffodil-test/src/test/resources/org/apache/daffodil/section12/length_properties/LengthProperties.tdml
```diff
@@ -911,7 +911,7 @@
     
     <xs:element name="bit26" type="xs:short" dfdl:lengthKind="explicit" dfdl:length="5" dfdl:byteOrder="bigEndian" />
     <xs:element name="bit27" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="9" dfdl:byteOrder="bigEndian" />
-    <xs:element name="bit28" type="xs:integer" dfdl:lengthKind="explicit" dfdl:length="13" dfdl:byteOrder="bigEndian" />
+    <xs:element name="bit28" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="13" dfdl:byteOrder="bigEndian" />
```
mbeckerle, line 914: Lots of schemas from 3rd parties may run in to this issue. I have seen schemas where the author tends to use the more general type xs:integer thinking it is equivalent to all the smaller types. 

I think we need to look beyond just Daffodil's tests to see the impact of this change.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -546,7 +554,13 @@ trait ElementBaseGrammarMixin
   private def explicitBinaryLengthInBits() = {
     val lengthFromProp: JLong = repElement.lengthEv.optConstant.get
     val nbits = repElement.lengthUnits match {
-      case LengthUnits.Bits => lengthFromProp.longValue()
+      case LengthUnits.Bits => repElement.primType match {
+         case PrimType.Integer | PrimType.NonNegativeInteger => {
+           SDW(WarnID.DeprecatedTypeUse, "In a future release, lengthUnits='bits' will only be supported for integer types of 64 bits or less")
```
mbeckerle, line 559: Say lengthUnits 'bits' will not be supported for types xs:integer and xs:nonNegativeInteger.

mbeckerle (approved):
+1 

But I think the message should just mention the xs:integer and xs:nonNegativeInteger types, and not use the terminology of "integer types less than 64 bits" which is less clear.

daffodil-runtime2/src/test/resources/org/apache/daffodil/runtime2/ex_nums.dfdl.xsd
```diff
@@ -150,7 +150,7 @@
                                         dfdl:byteOrder="littleEndian"/>
                             <xs:element name="le_uint8" type="xs:unsignedByte"
                                         dfdl:byteOrder="littleEndian"/>
-                            <xs:element name="le_nonNegativeInteger10" type="xs:nonNegativeInteger"
+                            <xs:element name="le_uint10" type="xs:unsignedInt"
                                         dfdl:byteOrder="littleEndian"
                                         dfdl:length="10"
```
tuxji, line 155: I also still want ex_nums to test the integer and nonNegativeInteger types, but we'll need new elements for them.  You don't have to do it in this PR - I can add the new elements later.

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypesUnparse.tdml
```diff
@@ -118,7 +118,7 @@
 
     <xs:element name="float_01" type="xs:float" />
     <xs:element name="double_01" type="xs:double" />
-    <xs:element name="integer_01" type="xs:integer"
+    <xs:element name="integer_01" type="xs:int"
       dfdl:lengthKind="explicit" dfdl:lengthUnits="bits" dfdl:length="8" />
```
tuxji, line 122: This is an example of a test schema which wouldn't need any changes under the interpretation that length must be a multiple of 8 bits.

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
tuxji: Also, best practice in `<tdml:warning>` is to look only for two or three important words, so that the tests can be more robust to future error or warning message changes.

tuxji (changes requested):
I don't think the spec makes it clear that lengthUnit="bits" cannot be used with xs:integer and xs:nonNegativeInteger.  I see the following table:

```
12.3.7.2.1   Length of Base-2 Binary Number Elements

Non-floating point numbers with binary representation and dfdl:binaryNumberRep 'binary' are represented as a bit string which contains a base-2 representation.

The value of the specified length is constrained per the table below. The lengths are expressed in bits and are inclusive.

Type                   | Minimum value of length | Maximum value of length
xs:nonNegativeInteger  | 1                       | Implementation-dependent (but not less than 64)
xs:integer             | 2                       | Implementation-dependent (but not less than 64)
xs:decimal             | 8[39]                   | Implementation-dependent (but not less than 64)

Table 23: Allowable Specified Lengths in Bits for Base-2 Binary Number Elements

[39] Type decimal must be a minimum of 8 bits because lengthUnits 'bits' is not allowed for xs:decimal.
```

The footnote [39] is the only place I see the phrase "lengthUnits 'bits' is not allowed for" and it applies only to decimal.  I think if the length really cannot be a minimum of 1 or 2 bits, then another valid interpretation is that if lengthUnit="bits" is used with integer/nonNegativeInteger, then the length itself must be a multiple of 8 bits or a SDW/SDE will be thrown.

Do we have any code which throws a SDE if lengthUnit="bits" is used with decimal?  Please add a test to verify that a SDE is thrown for decimal so we can be certain of Daffodil's behavior.

stevedlawrence:
> I don't think the spec makes it clear that lengthUnit="bits" cannot be used with xs:integer and xs:nonNegativeInteger. I see the following table:

Section 12.3 of the says this in the table describing lengthUnits="bits":

> 'bits' may only be used for xs:boolean, xs:byte, xs:short, xs:int, xs:long, xs:unsignedByte, xs:unsignedShort, xs:unsignedInt, and xs:unsignedLong simple types with binary representation, and for calendar (date and time) simple types with binary packed representation.

I don't understand the reasoning, bit it is pretty unambiguous.

Note that this also adds that it's not allowed for `representation="binary"`. So we *can* have `lengthUnits="bits"` for these types when `representation="text"`, we should ensure we still allow that.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -103,6 +103,14 @@ trait ElementBaseGrammarMixin
 
   lazy val prefixedLengthElementDecl: PrefixLengthQuasiElementDecl = LV('prefixedLengthElementDecl){
     Assert.invariant(lengthKind == LengthKind.Prefixed)
+    optPrimType match {
+      case Some(primType) => primType match {
+        case PrimType.Integer | PrimType.NonNegativeInteger if lengthUnits == LengthUnits.Bits =>
```
stevedlawrence, line 108: Is this check necessary here? I would expect a prefixed length quasi element would eventually end up hitting the other change below at some point?

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/ExplicitTests.tdml
mike-mcgann: Updated message and warning check

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mike-mcgann: Updated message

mbeckerle:
The DFDL spec is clear that (section 12.3 definition of lengthKind property:

'bits' may only be used for xs:boolean, xs:byte, xs:short, xs:int, xs:long, xs:unsignedByte, xs:unsignedShort, xs:unsignedInt, and xs:unsignedLong simple types with binary representation, and for calendar (date and time) simple types with binary packed representation.

Integer types other than these are xs:integer and xs:nonNegativeInteger.

Numeric types other than those are float, double, decimal. 

There has been agreement in the DFDL workgroup that lengthUnits bits will be allowed for hexBinary, and these can be arbitrarily long.

mike-mcgann:
> If we are currently allowing bits, i.e., the code is present and tested, for integer and nonNegativeInteger types, then we should not disable this functionality, but rather should just add a tunable, defaulting to allowing the current behavior, which can be flipped to turn on stricter rules.

The current behavior is still allowed except that it now emits a warning. I updated the tests to reflect the desired usage. Do we want to:

- Remove the warning, convert the SDW to a tunable SDE?
- Update tests to try both the updated and deprecated way?
- Revert tests back to the deprecated way and update to the desired usage later?

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mike-mcgann: I think so. I wrote the test first to see if prefixed would work as expected with the other change and it didn't. My understanding is that there are two types here, the type of the element itself and the type of the prefix. This case checks the type of the element itself and the later check is for the prefix type.

mbeckerle:
Let me opine about this for a sec and I think the right thing to do will emerge.

DFDL didn't want to allow 'bits' for some things to both make implementations not have to agonize of features nobody would need, and to avoid users causing themselves grief. E.g., if we only want to support IEEE floats on byte boundaries then restricting xs:float to not have length units 'bits' makes some sense.

What we have learned is that data format spec documents often pick one way of expressing length, e.g., bits, and stick with it. When creating a DFDL schema it is helpful to use the units that are in the spec document, even if the usage is restricted to multiples of 8 bits long.  So it is convenient in some cases to say dfdl:lengthUnits='bits' even for types where bits makes no sense like xs:float. 

So for that reason, I think Daffodil never implemented checks to enforce what can have length units bits.

But, to the letter of the DFDL spec, length units 'bits' isn't allowed, so not portble.

So I think we want a tunable that says to warn on this usage (which can be suppresed), and a tunable to allow/disallow length units bits - i.e., cause it to be an SDE. 

This would apply to all the types where we allow bits, but the DFDL spec does not. 
Including hexBinary, float, double, decimal, integer, nonNegativeInteger (assuming I have the right list).

mbeckerle:
To be clear: we want a warning (which can be suppressed via existing means), and then a tunable which says whether to SDW, or SDE. Would default to SDW now.

mbeckerle:
> > Note that this also adds that it's not allowed for `representation="binary"`. So we _can_ have `lengthUnits="bits"` for these types when `representation="text"`, we should ensure we still allow that.
> 
> I have asked for a clarification at the DFDL workgroup about this. I expect this is just poor wording, and there is no intention to allow 'bits' lengthUnits for these types ever.

Per discusssion on DFDL Workgroup 2023-01-19, the wording in the spec does not seem ambiguous. bits is allowed only for the specific set of types with binary representation, and this is to be interpreted as meaning that bits length units are allowed for that specific set of types only when they have binary representation, and bits is not allowed otherwise. Hence, bits is not allowed when they have text representation.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -103,6 +103,14 @@ trait ElementBaseGrammarMixin
...
+          SDW(WarnID.DeprecatedTypeUse, "In a future release, lengthUnits='bits' will only be supported for integer types of 64 bits or less")
```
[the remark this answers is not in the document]
mike-mcgann: Updated to check against a set of allowed types

mike-mcgann:
> To be clear: we want a warning (which can be suppressed via existing means), and then a tunable which says whether to SDW, or SDE. Would default to SDW now.

Updated to emit a warning by default and to raise an error when the `allowBigIntegerBits` tunable is set to true.

mike-mcgann:
> Hence, bits is not allowed when they have text representation.

Updated to add a check for binary representation.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1368,4 +1371,32 @@ trait ElementBaseGrammarMixin
...
+  private def assertLengthUnits(elem: ElementBase = context): Unit = {
+    elem.lengthUnits match {
+      case LengthUnits.Bits if elem.representation == Representation.Binary =>
+        elem.optPrimType match {
+          case Some(primType) =>
+            if (!allowedBitTypes.contains(primType))
+              if (tunable.allowBigIntegerBits)
+                SDW(WarnID.DeprecatedBigIntegerBits, s"In a future release, lengthUnits='bits' will only be supported for the following types: $allowedBitTypesText")
+              else
+                SDE("lengthUnits='bits' is only supported for the following types: $allowedBitTypesText")
```
tuxji, line 1396: Since `elem` can be `context` or `repElement`, I wonder if we need to call SDE/SDW on `elem` instead of `this` since the SDE/SDW methods include the calling object's source location in the printed message?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -101,6 +101,17 @@
...
     <xs:complexType>
       <xs:all>
+        <xs:element name="allowBigIntegerBits" type="xs:boolean" default="true" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>
+              Daffodil allows the use of lengthUnits="bits" when the type is xs:integer and
+              xs:nonNegativeInteger even though this is prohibited by the specification. When
+              this tunable is true, a deprecation warning is emitted when this condition is
+              encountered. When this tunable is false, this condition will instead result in a
+              schema definition error.
```
tuxji, line 111: We've changed the check from disallowing 2 indefinite size integer types to allowing only 9 definite size integer types which means that users will get warnings or errors for many more types (float, double, etc.).  I think it's important to reword this documentation so that users will know how much this tunable will impact their schemas.  Remember Daffodil used to let many schema writers pick one length unit (bits or bytes) and write every type's length using that unit without any schema diagnostics, so the new SDE/SDW diagnostic will be a big change in behavior for users.  Consider saying something like:

```
Previous Daffodil releases let schemas define every type's length using "bits" as the length unit
even though the specification allows bit length units only for a specific set of types' binary 
representations and does not allow bit length units for any other type's binary representation 
or any type's text representation.  When this tunable is true, a deprecation warning is issued
when bit length units are incorrectly used.  When this tunable is false, a schema definition
error will be issued instead.
```

tuxji (approved):
+1 with some suggestions

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
mbeckerle: There is a separate JIRA ticket for making SDW warnings include instructions on how to suppress them (mention the specific symbol needed to put into the config file). That will help with this issue. Also.

This ticket or any that increases warnings users get, will need a release note. Probably generic one about how this release of Daffodil issues more warnings about non-portable aspects of DFDL usage, .... and how to suppress them.

stevedlawrence: Speaking of which, we need a place to keep track of deprecation descriptions so that we remember to actually put them in the release notes.

One approach is if a commit causes deprecation/compatibility issues then there must be a clearly labeled section at the end of the commit message about what was deprecated with a specific keyword. For example, such a commit message might look like:

```
Make some non-backwards compatible/deprecation change

Description of what this commit does and why. This is more focused
on the actual implementation details of the change, like a normal
commit message.

Deprecation/Compatibility:

Message that can be copy/pasted into the release notes in the
deprecation/compatibility section. This is focused on what was
deprecated or causes backwards compatibility  issues and the
migration details for users to resolve those issues.

DAFFODIL-XYZ
```

This way when the release manager is generating the release notes they can just look at the git log for commits with the "Deprecation/Compatibility" keyword and copy the message without really having to think too much.

Alternatively we could add something like that to a comment in the ticket, but it I find searching with JIRA can be a little hard.

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -103,6 +103,7 @@ trait ElementBaseGrammarMixin
 
   lazy val prefixedLengthElementDecl: PrefixLengthQuasiElementDecl = LV('prefixedLengthElementDecl){
     Assert.invariant(lengthKind == LengthKind.Prefixed)
+    assertLengthUnits()
```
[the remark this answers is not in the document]
mike-mcgann: Updated the name to `checkLengthUnits`

daffodil-core/src/main/scala/org/apache/daffodil/grammar/ElementBaseGrammarMixin.scala
mike-mcgann: Updated to call `elem.{SDE,SDW}`

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
mike-mcgann: Documentation has been updated

stevedlawrence:
Can you rebase to the lastest master and resolve the conflicts?

Also, can you add the Deprecated/Compatiability stuff to your commit message? Unfortunately updating the PR description doesn't change the commit to match and must be done manually.

stevedlawrence (approved):
+1 :+1:

PR #922, opened 2023-01-19
Move CLI src/it tests to src/test except UDF testing

mbeckerle:
sbt test and sbt it:test still work. Only the CLI UDF tests are in sbt it:test now.

You can also run all the CLI src/test/scala tests from IntelliJ IDEA by right-click.

Note that sbt daffodil-cli/test fails due to, I think, relative paths in the tests that assume the daffodil root directory.

DAFFODIL-2772


mbeckerle:
Abort. Too many things subtly broken.

project/Rat.scala
```diff
@@ -39,34 +39,34 @@ object Rat {
...
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input16.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input18.json"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input18.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input18.exi"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input18.exisa"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/input19.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/inputBig1M.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/prefix.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/test_DFDL-714.txt"),
+    file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/uuid.txt"),
```
tuxji, line 69: I know you aborted this pull request, but I want to note that you can ignore every file in the `daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input` directory with a single line, `file("daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input")`.  Most of the files are .txt files and you can move the 3 .py/.xml files (the only files which have licenses) to another directory if you feel like keeping them apart.

stevedlawrence: I wonder if we could/should adopt a file naming convention for files that are ALv2 but cannot have a license header in them? Then we can just ignore all files that match that pattern and remove most of this list, and we don't have to worry about accidentally ignoring entire directories that might have a mix of licenses.

PR #923, opened 2023-01-19
Move src/it tests to src/test except UDF testing

mbeckerle:
Posting this for assistance. I have no clue what is wrong here.

Under my IDE, I can now right click on daffodil-cli/src/test/scala and choose run all tests, and all of them run as unit tests. All of them run, save the 18 UDF tests which still live under src/it. 

Running 'sbt it:test' successfully runs the 18 UDF tests of daffodil-cli.

But 'sbt test' yields 24 errors (out of 237 in daffodil-cli.). 

```
[error] Failed: Total 237, Failed 24, Errors 0, Passed 209, Skipped 4
[error] Failed tests:
[error] 	org.apache.daffodil.schematron.TestSvrlOutput
[error] 	org.apache.daffodil.parsing.TestCLIparsing
[error] 	org.apache.daffodil.debugger.TestCLIdebugger
[error] 	org.apache.daffodil.tunables.TestCLITunables
[error] 	org.apache.daffodil.schematron.TestValidating
[error] 	org.apache.daffodil.generating.TestCLIGenerateC
[error] 	org.apache.daffodil.performance.TestCLIPerformance
[error] 	org.apache.daffodil.tdml.TestCLItdml
[error] 	org.apache.daffodil.unparsing.TestCLIunparsing

```
Alas, the messages from these are decidedly sparse. Just says expect timed out, and I cannot reproduce this under the debugger. The failure does not happen. 

I looked at a few of these 24 failing tests and cannot spot anything similar or problematic about them. 

Note that the intelliJ IDEA IDE, gets its configuration from the build.sbt. 

Does anyone have an idea what might be wrong. 

DAFFODIL-2772


stevedlawrence:
I *think* I might know what the problem is. Each CLI test changes stdout/err/in to a temporary buffer during the duration of a test, which is needed to actually capture the output for expect. But this means CLI tests aren't thread safe and cannot be parallelized.

So in build.sbt we have this line to disable parallel CLI tests:

```scala
IntegrationTest / parallelExecution := false,
```
But that means all of our CLI tests in the `Test` configuration will be run in parallel and randomly break. My guess is your IDE doesn't do this, or does some other magic when it runs tests.

To fix this, I think we can remove that line above from `build.sbt` and then add these lines to `daffodil-cli/build.sbt`
```scala
Test / parallelExecution := false
IntegrationTest / parallelExecution := false
```

This ensures that all daffodil-cli tests, whether forking or not, are run sequentially and avoids these thread unsafe issues. I tested it locally and it seems to work...

Except for one failure, apparently the test `org.apache.daffodil.parsing.TestCLIparsing.test_XXX_CLI_Parsing_SimpleParse_null` sets fork=true so that it can capture eof and make sure there is no output. But checking for eof and using getBefore is hack that isn't guaranteed to always work. We should just change that test so it outputs to a temp file and then checks that the file is empty.

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/test_DFDL-714.txt
```diff
@@ -1 +1 @@
-Hello World 
+Hello World
```
mbeckerle, line 1: I did not change this file. ONly moved it.

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/trace_input.dfdl.xsd
```diff
@@ -1,56 +1,56 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
mbeckerle, line 1: I made no changes to this. May just be that it was converted CRLF -> LF by my IDE? Or tabs to spaces?

daffodil-cli/src/test/scala/org/apache/daffodil/exi/TestEXIEncodeDecode.scala
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.daffodil.saving
+package org.apache.daffodil.exi
```
mbeckerle, line 18: I guess my IDE has decided that directory names and packages must conform. 
I think the right thing here is for the dirctory name to be "saving", and the package to match.

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/test_DFDL-714.txt
stevedlawrence: Looking at the before and after files in a hex editor, it looks like a CRLF was replaced with an LF. Probably your IDE doing it, but it's a fine change. We can't rely on line endings anyways due to git auto-crlf, and I'd prefer things be LF anyways.

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/trace_input.dfdl.xsd
stevedlawrence: Yep, I think CRLF -> LF, same as the other file. Change is fine

daffodil-cli/build.sbt
```diff
@@ -27,6 +27,7 @@ enablePlugins(WindowsPlugin)
 (IntegrationTest / test) := (IntegrationTest / test).dependsOn(Compile / stage).value
 (IntegrationTest / testOnly) := (IntegrationTest / testOnly).dependsOn(Compile / stage).evaluated
 (IntegrationTest / testQuick) := (IntegrationTest / testQuick).dependsOn(Compile / stage).evaluated
+(IntegrationTest / dependencyClasspath) ++= (Test / exportedProducts).value
```
stevedlawrence, line 30: Might be worth adding a comment why we need this, maybe something like:
> Add the classes in src/test/ to the IntegrationTest classpath so we can share CLI utliities in Util.scala

stevedlawrence (approved):
+1 once tests pass, suggestions for fix in a comment

daffodil-cli/src/test/scala/org/apache/daffodil/exi/TestEXIEncodeDecode.scala
stevedlawrence: I think the `saving` package is intended for the `daffodil save-parser` command. I think the exi package name is probably fine. 

Alternatively, we could just get rid of all these separate packages. I don't think we really gain anything by having each test suite it a unique package. We could just move all of our scala files in a "cli" directory and then all have the package names be `org.apache.daffodil.cli` or something more generic.

tuxji: I second Steve's suggestion to remove the extra packages below org.apache.daffodil.cli.

mbeckerle:
Pushed changes. Tests fixed. Comment added.  Packages revised to mostly flatten/remove them.

build.sbt
```diff
@@ -178,7 +178,7 @@ lazy val commonSettings = Seq(
   sourceManaged := baseDirectory.value / "src_managed",
   resourceManaged := baseDirectory.value / "resource_managed",
   libraryDependencies ++= Dependencies.common,
-  IntegrationTest / parallelExecution := false,
+  // IntegrationTest / parallelExecution := false, // allow test in parallel
```
stevedlawrence, line 181: Thoughts on removing this commented line? There's nothing inherent about integration tests that require non-parallelism, it was really just our CLI tests which happened to be all integration tests, but that's no longer the case anymore.

mbeckerle: I would rather be explicit about this. Particularly since the CLI testing can't do this.

mbeckerle:
There may be still some sort of timing/race-condition thing, as one of the CI test runs (Java 8 on Windows) a single CLI test failed. 
This is of course spurious, but indicates a separate issue with the CLI test rig or expect framework we use for those.

mbeckerle:
@tuxji your re-review?

tuxji (approved):
+1 with requested changes

I do not like cliTest as a package / folder name but Daffodil tests never get packaged and deployed so I won't object.  However, please consider renaming CLI as a package / folder name since good Scala style uses all lowercase:

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/* -> cli/* (or cliTest for consistency)
daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala -> cliTest/Util.scala (like the rest)

daffodil-cli/build.sbt
```diff
@@ -23,11 +23,21 @@ enablePlugins(JavaAppPackaging)
 enablePlugins(RpmPlugin)
 enablePlugins(WindowsPlugin)
 
+// CLI tests are not thread safe
+// due to use of temporary buffer used to capture stdout/stderr.
+// So we cannot test in parallel
+Test / parallelExecution := false
+IntegrationTest / parallelExecution := false
```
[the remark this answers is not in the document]
stevedlawrence: The core of the problem is that the the CLI runs from `object Main`, and we mutate the variables STDIN, STDOUT, STDERR in that object everytime we run a CLI test so it uses our `expect` buffers instead of the actual stdin/out/err. Because these are mutated in a single object it's not thread safe.

I guess one thing we might be able to do is change our Main object so it's an actually a class that can be instantiated and is thread safe. Then the object Main just becomes something like this:

```scala
object Main {
  override def main(args: Seq[String]): Unit = {
    val main = new Main()
    main.run(args)
  }
}
```
This would then allow our CLI tests to avoid using the Main object and instead create a new Main instance for each test which can be thread safe, e.g.

```scala
val in, out, err = // get expect buffers
val main = new Main(in, out, err)
main.run(args)
```

I think the one other potential problem area is capturing log4j logs. Like with Main, we also configure log4j to write to these `expect` buffers. But that's also global state and so it's not thread safe either. I don't think there is a way to make log4j have thread local state. Maybe what we can do is have a custom log4j handler that maintains a ThreadLocal of the buffers it should write too.  When we start a thread, we mutate the ThreadLocal with the new buffers to use. And when log4j has a message to write, we get the buffers from the ThreadLocal. I'm not sure if that would actually work, but seems plausible...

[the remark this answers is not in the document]
stevedlawrence: Note that for the integration tests that fork it is still important to not run in parallel, otherwise we get too much forking and quickly run out of memory.

daffodil-cli/src/test/scala/org/apache/daffodil/cliTest/TestBlob.scala
```diff
@@ -43,15 +40,15 @@ class TestBlob {
    * will need to generate the test file(s) using the gen_blob.py script. It can
    * be found in and should be run from the directory:
    *
-   * daffodil-cli/src/it/resources/org/apache/daffodil/CLI/input/
+   * daffodil-cli/src/test/resources/org/apache/daffodil/CLI/input/
```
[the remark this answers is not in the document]
mbeckerle: Nothing under src/test/resources changed directories. Only scala code packages. 
The resources weren't all chopped up into little packages.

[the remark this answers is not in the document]
tuxji: Did you see the comment I wrote under my +1?  I suggested you rename CLI to cli or cliTest, although I wouldn't worry about it since the PR's merged already:

+1 with requested changes

I do not like cliTest as a package / folder name but Daffodil tests never get packaged and deployed so I won't object. However, please consider renaming CLI as a package / folder name since good Scala style uses all lowercase:

daffodil-cli/src/test/resources/org/apache/daffodil/CLI/* -> cli/* (or cliTest for consistency)
daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala -> cliTest/Util.scala (like the rest)

PR #924, opened 2023-01-19
Rewrite XML to/from PUA remappers

mbeckerle:
This is just part 1 of several changes needed for fixing DAFFODIL-1559 which provides a general capability to preserve CRLFs and isolated CR in all data.

The only significant algorithmic change here is getting rid of a whole pass to determine if a string needs remapping.
Now we do the scan and if/when we detect a need to remap we turn on the remapping machinery and accumulate into a string builder.  

DAFFODIL-1559

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/util/CharacterSetRemapper.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/util/CharacterSetRemapper.scala
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.daffodil.util
+
+/**
+ * A abstract base for Remappers which convert strings.
+ *
+ * The public interface is just `def remap(s: String): String`.
+ *
+ * There are protected methods that implementations must provide.
+ *
+ * Contains shared implementation methods also.
+ *
+ * NOTE: This is inner loop stuff. Keep it and derived classes lean and fast.
+ * Use a java-like coding style. While loops, not map/flatmap/etc. avoid tuples.
+ */
+trait CharacterSetRemapper {
+
+  /**
+   * Remaps characters. Provides the previous and following characters since some remappings
+   * require this context for Surrogate Pairs, CRLF->LF etc.
+   *
+   * Plays a trick with negating the return value in order to avoid having to
+   * return more than one value, which is potentially less efficient.
+   *
+   * @param prev The character prior to the one being considered. (Needed for surrogates)
+   * @param curr The character under consideration for remapping.
+   * @param next The next character afterwards. (Needed for surrogates and CRLF pairs)
+   * @return The remapped character (as an Int) or that same remapped character Int
+   *         value negated, which signals that curr+next was remapped to a single character.
+   *         Such as is needed if CRLF is remapped to just LF.
+   */
+  protected def remap (prev: Char, curr: Char, next: Char): Int
+
+  /**
+   * Remaps the string. Returns the original string object if no remapping is required.
+   *
+   * Because of surrogate pairs, and the difference between 16-bit string codepoints
+   * and real character codes, lots of things that traverse strings need
+   * to consider either the codepoint after (if current is a leading surrogate)
+   * or codepoint before (if current is a trailing surrogate).
+   *
+   * This is not the only kind of character set remapping. In particular this is
+   * restricted to replace 1 character with 0 or 1 character in the remapped string.
+   * Other character set remappers can convert 1 character into 0 to many characters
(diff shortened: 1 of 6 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/util/CharacterSetRemapper.scala
```diff
@@ -0,0 +1,131 @@
...
+   * needed for surrogates or CRLFs.
+   *
+   * See XMLUtils.walkUnicodeCharset for a more general kind of remapping that can
+   * replace 1 character with N as well as being context sensitive about adjacent
+   * characters before and after.
+   *
+   * This algorithm uses a StringBuilder which is not synchronized
+   * so it is noticably faster than StringBuffer, and since the StringBuilder
+   * is local to the function, we don't have to worry about any threading issues.
+   * This makes for a noticeable speed increase.
```
mbeckerle, line 71: Add to comment about inner-loop hence, this very Java coding style as performance matters for this.

mbeckerle:
There are numerous more corner cases covered by testing now as well.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
```diff
@@ -0,0 +1,117 @@
...
+          // well formed surrogate pairs are preserved
+          curr
+        } else {
+          // curr is an isolated surrogate, so to preserve we must remap to PUA
+          curr + 0x1000
+        }
+      }
+      case _ if (curr >= 0xE000 && curr <= 0xF8FF) => { // Unicode PUA is E000 to F8FF.
+        if (checkForExistingPUA)
+          Assert.usageError("Pre-existing Private Use Area (PUA) character found in data: '%s'".format(curr))
```
stevedlawrence, line 83: Should this throw an actual exception that can be handled and something done with? If a caller has enabled the `checkForExistingPUA` flag might they want to convert that to an SDE or something?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
```diff
@@ -0,0 +1,117 @@
...
+        else curr
+      }
+      case _ if (curr < 0xFFFE) => curr
+      case 0xFFFE => 0xF0FE
+      case 0xFFFF => 0xF0FF
+      case _ =>
+        Assert.usageError("Character code beyond U+FFFF. Codepoint: %s".format(curr))
```
stevedlawrence, line 90: Can you disable code coverage for this line?

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
```diff
@@ -0,0 +1,117 @@
...
+  extends CharacterSetRemapper {
+
+  /**
+   * This direction of remapping is simpler. No context characters are needed, and
+   * it never returns a negated character code.
+   */
+  override protected def remap(prevIgnored: Int, c: Int, nextIgnored: Int): Int = {
+    val res = c match {
+      case _ if (c >= 0xE000 && c <= 0xE01F) => c - 0xE000 // Ascii c0 controls
+      case _ if (c >= 0xE800 && c <= 0xEFFF) => c - 0x1000 // surrogate codepoints
```
stevedlawrence, line 110: This is isoluated surrogates right? Not well-formed surrogates? Might be worth the distinction in the comment just to make it more clear.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/XMLUtils.scala
```diff
@@ -1299,11 +1126,11 @@ Differences were (path, expected, actual):
    * We have to use our own PUA remapping trick if we want to be sure to preserve
    * CR in XML.
    */
-  def escape(str: String, sb: StringBuilder = new StringBuilder()): StringBuilder = {
+  def escape(s: String, sb: StringBuilder = new StringBuilder()): StringBuilder = {
```
stevedlawrence, line 1129: We do have a number of places where we call `scala.xml.Utility..escape`? I wonder if those are bugs?

I also wonder if maybe we want to do the same copy-on-write kind of thing for this function? Or maybe that's pre-optimization? Can be done in a separate PR if thats something we think we need.

stevedlawrence (approved):
+1 nice change!

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
tuxji: Unicode allows code points up to 0x10FFFF (21 bits), but the size of Java's Char type is 16 bits.  Since the arguments are coming from or going back to Char variables, they'll be 16 bits long unless a programming error happens.

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
```diff
@@ -0,0 +1,117 @@
...
+      case 0xF0FE => 0xFFFE
+      case 0xF0FF => 0xFFFF
```
tuxji, line 112: Be nice to add comments explaining these remappings too.

tuxji (approved):
+1 with requested changes to comments

daffodil-lib/src/main/scala/org/apache/daffodil/xml/PUARemappers.scala
```diff
@@ -0,0 +1,135 @@
...
+
+object Unicode {
+  def isLeadingSurrogate(c: Char) = {
+    c >= 0xD800 && c <= 0xDBFF
+  }
+
+  def isTrailingSurrogate(c: Char) = {
+    c >= 0xDC00 && c <= 0xDFFF
+  }
+}
```
stevedlawrence, line 30: I just noticed that `java.lang.Character` has `isHighSurrogate` and `isLowSurrogate` functions that I think are the same as these functions.

mbeckerle:
Thank you for the thorough review. I've addressed all feedback. 
Once the checks pass I will merge this.

PR #925, opened 2023-01-20
Update tests for nested separators

jadams-tresys:
The nest1 test was not defined correctly and has been removed. The nest2 test needed a small modification to look for the appropriate '!' terminator, so that has been added and the test has been uncommented. With this small change, the nest2 test works as expected.

DAFFODIL-2101

### DIFF
--- a/daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/nested-separator-delimited.tdml
+++ b/daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/nested-separator-delimited.tdml
@@ -202,8 +202,6 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
-
-
   <tdml:defineSchema name="nest1">
     <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
     <dfdl:format ref="ex:GeneralFormat" occursCountKind="parsed" />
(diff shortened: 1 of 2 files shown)


mbeckerle (approved):
+1 So presumably the nest1 test added no real value.

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/nested-separator-delimited.tdml
```diff
@@ -202,61 +202,14 @@
...
-          <a>4</a>
-          <a>5</a>
-          <a>6</a>
-          <a>7</a>
-          <a>8</a>
-          <a>9</a>
-        </nest1>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
-  </tdml:parserTestCase>
```
stevedlawrence, line 249: Is there any value in keeping this test but changing it to a negative test that show's that this kind of nesting of separators doesn't work, and that finding the outer separator will cause the array to end?

stevedlawrence (approved):
+1

daffodil-test-ibm1/src/test/resources/test-suite/tresys-contributed/nested-separator-delimited.tdml
jadams-tresys: I will update it to expect the error.

PR #926, opened 2023-01-23
Allow string constructor for dfdl:hexBinary

mike-mcgann:
The constructor for dfdl:hexBinary raises a schema definition error when passed a string as input. This change updates it to behave in the same way as xs:hexBinary by using string input as hex characters.

[DAFFODIL-1454](https://issues.apache.org/jira/browse/DAFFODIL-1454)

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSHexBinary.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSHexBinary.scala
@@ -20,35 +20,9 @@ package org.apache.daffodil.dpath
 import java.math.{ BigInteger => JBigInt, BigDecimal => JBigDecimal }
 import org.apache.daffodil.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.infoset.DataValue.DataValueByteArray
+import org.apache.daffodil.util.Misc
 
 trait HexBinaryKind {
-
-  /**
-   * http://travisdazell.blogspot.com/2012/11/converting-hex-string-to-byte-array-in.html
-   */
-  protected def hexStringToByteArray(str: String): Array[Byte] = {
-    val len = str.length
-
-    if ((len % 2) != 0)
-      throw new NumberFormatException("Failed to evaluate expression: A hexBinary value must contain an even number of characters.")
-
-    val arr = new Array[Byte](len / 2)
-    var i = 0
-    while (i < len) {
-      val upper = Character.digit(str.charAt(i), 16)
-      val lower = Character.digit(str.charAt(i + 1), 16)
-
-      if (upper == -1)
-        throw new NumberFormatException("Failed to evaluate expression: Invalid hexadecimal digit '%c' at index %d of '%s'".format(str.charAt(i), i, str))
-      if (lower == -1)
-        throw new NumberFormatException("Failed to evaluate expression: Invalid hexadecimal digit '%c' at index %d of '%s'".format(str.charAt(i + 1), i + 1, str))
-
-      val byte = (upper << 4) + (lower)
-      arr(i / 2) = byte.asInstanceOf[Byte]
-      i += 2
-    }
-    return arr
-  }
 
   protected def reduce(numeric: Any): Array[Byte] = {
     val res: Array[Byte] = numeric match {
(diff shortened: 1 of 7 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypesUnparse.scala
```diff
@@ -50,7 +50,9 @@ class TestSimpleTypesUnparse {
   @Test def test_hexBinary_unparse_10(): Unit = { runner.runOneTest("hexBinary_unparse_10") }
   @Test def test_hexBinary_unparse_11(): Unit = { runner.runOneTest("hexBinary_unparse_11") }
   @Test def test_hexBinary_unparse_12(): Unit = { runner.runOneTest("hexBinary_unparse_12") }
+  @Test def test_hexBinary_unparse_13(): Unit = { runner.runOneTest("hexBinary_unparse_13") }
```
stevedlawrence, line 53: At the top of this we have this test commented out. That can be removed now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSHexBinary.scala
```diff
@@ -67,7 +67,7 @@ trait HexBinaryKind {
       case bi: JBigInt if (bi.bitLength <= 63) => reduce(bi.longValue())
       case bi: JBigInt => bi.toByteArray()
       case bd: JBigDecimal if (try { bd.toBigIntegerExact(); true } catch { case e: ArithmeticException => false }) => reduce(bd.toBigIntegerExact())
-      case str: String => reduce(new JBigInt(str))
+      case str: String => reduce(new JBigInt(str, 16))
```
stevedlawrence, line 70: Since the spec says `dfdl:hexBinary behaves identically to the XPath 2.0 constructor function of the same name...` is there any value in making it so the string case is exactly the same as the `XSHexBinary` constructor? We can't really call its run function since we need an instance, but it just calls `StringToHexBinary.computeValue(...)`. Should we do the same? Note that that just calls the above `hexStringToByteArray` function. Maybe we should call that instead?

And to make things more confusing, we also have a `Misc.hex2Bytes` function, which seems like it has basically the same logic but throws an IllegalArgumentException's instead of NumberFormatException's. Maybe we should drop the hexStringToByteArray function in favor of Misc.hex2Byte, and change it to throw NumberFormatExceptions?

stevedlawrence (approved):
+1, maybe a chance for removing code dulication

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -11808,7 +11808,7 @@
     </tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
-        <ex:dfdlHexBinary02>D0</ex:dfdlHexBinary02>
+        <ex:dfdlHexBinary02>0208</ex:dfdlHexBinary02>
```
tuxji, line 11811: Why is this change occurring?

tuxji (approved):
+1 with a question and I like Steve's suggestions too

daffodil-test/src/test/scala/org/apache/daffodil/section05/simple_types/TestSimpleTypesUnparse.scala
mike-mcgann: Thanks. I missed seeing that.

daffodil-runtime1/src/main/scala/org/apache/daffodil/dpath/XSHexBinary.scala
mike-mcgann: Updated to use `Misc.hex2Bytes` and this is actually more correct since it enforces the string to have an even number of digits.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
mike-mcgann: The original test assumed the behavior that a string of `"208"` gets converted to an integer of `208` and then converted to `0xd0` but the `208` should be parsed as hex digits and result in `0x0208`. A leading zero was added as the hexBinary constructor enforces an even number of digits.

PR #927, opened 2023-01-23
Add test showing both name and ref defined causes an SDE

stevedlawrence:
Add a test to verify that an SDE is thrown if a schema uses both name
and ref to define an element.

[DAFFODIL-768](https://issues.apache.org/jira/browse/DAFFODIL-768)

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/general.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/general.tdml
@@ -217,4 +217,32 @@
         <tdml:error>in no namespace</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
+
+  <tdml:defineSchema name="nameAndRef">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+    <dfdl:format ref="ex:GeneralFormat" />
+    <xs:element name="root1">
+      <xs:complexType>
+        <xs:sequence>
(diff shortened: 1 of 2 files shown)


tuxji (approved):
+1 

I could not understand the pull request's purpose just from looking at its first comment since the only thing the comment said was the issue number.  I had to look at [DAFFODIL-768](https://issues.apache.org/jira/browse/DAFFODIL-768) and then look at the commit message to realize that the commit's message had not included both a summary and a description as well as the issue number.  Next time, please write at least 3 non-blank lines (summary, description, issue number) in a PR commit message.  We probably need to mention this best practice in DEVELOP.md too.

```
Add test showing both name and ref defined causes an SDE

Add a test to verify that a SDE is thrown if a schema uses both name and ref
to define an element.

DAFFODIL-768
```

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/general.tdml
```diff
@@ -217,4 +217,32 @@
...
     </tdml:errors>
   </tdml:parserTestCase>
+
+  <tdml:defineSchema name="nameAndRef">
+    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+    <dfdl:format ref="ex:GeneralFormat" />
+    <xs:element name="root1">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="nameAndRef" type="xs:string" ref="ex:root1" dfdl:lengthKind="delimited" />
```
tuxji, line 227: Just noticed this belatedly.  The `ref="ex:root1"` should be `ref="ex:root2"` in order to avoid having two errors mixed into the same line.

PR #928, opened 2023-01-23
Changed NYI to SDE

mbeckerle:
Activated tests, made them expect an error.

DAFFODIL-847

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
@@ -1520,49 +1520,31 @@
     </tdml:infoset>
 
   </tdml:parserTestCase>
-  
-  <!--
-     Test Name: textStandardDecimalSeparator10
-        Schema: textNumberPattern
-          Root: tnp26
-       Purpose: This test demonstrates that textStandardDecimalSeparator can be a whitespace-separated list of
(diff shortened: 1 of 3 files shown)


stevedlawrence (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
```diff
@@ -1520,49 +1520,31 @@
...
 
   </tdml:parserTestCase>
-  
-  <!--
-     Test Name: textStandardDecimalSeparator10
-        Schema: textNumberPattern
-          Root: tnp26
-       Purpose: This test demonstrates that textStandardDecimalSeparator can be a whitespace-separated list of
-                single characters that appear (individually) in the data as the decimal separator
-  -->
```
tuxji, line 1530: Even though this comment and the next one were outdated, why were they deleted instead of updated?

tuxji (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
mbeckerle: I believe these block comments, which just restate what is obvious from the test itself, serve no purpose. 

I will change the test names however, to better reflect the specific thing they are testing.

PR #929, opened 2023-01-24
Remove ThreadLocal wrapper of ICU4J DecimalFormat

stevedlawrence:
The DecimalFormat used for parsing and unparsing text numbers has been thread-safe since ICU 59. We have upgraded well past that, and so we no longer need to use ThreadLocal to ensure thread-saftey. Remove them from the TextNumberFormatEv and return a DecimalFormat that can be shared among threads, simplifying the code a bit and avoiding potential overhead related to ThreadLocal.

[DAFFODIL-2774](https://issues.apache.org/jira/browse/DAFFODIL-2774)


mbeckerle (approved):
+1 

So this is a thread-safe object now, so long as we only share one that is fully initialized. 

I suggest comments explaining that where the object is defined in EVTextNumberFormat, but other than that all good.

tuxji (approved):
+1, looks good

PR #931, opened 2023-01-25
Fix issue with groups containing choices: WIP

jadams-tresys:
There was an issue with groups containing choices were unable to successfully find their next element when unparsing. Essentially what was happening was that the element following the choice is required, but it's possible that due to the group being shared there are different possible next elements as the separate group references have different contexts. So, instead of returning an ErrorERD for a choice with a required next element, we return an Maybe.Nope which allows the trdStack to continue iterating.

This feels a little hacky in it's current implementation, but it does work and doesn't seem to break anything.  There may be a more suitable fix that gets the trdStack to iterate in this group/choice edge case, but wanted to open this up for discussion.

DAFFODIL-2615, DAFFODIL-2562

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
@@ -21,6 +21,7 @@ import org.apache.daffodil.core.dsom._
 import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.schema.annotation.props.gen.NilKind
+import org.apache.daffodil.lib.util.Logger
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.xml.QNameBase
 import org.apache.daffodil.runtime1.dsom._
@@ -194,6 +195,7 @@ trait TermRuntime1Mixin { self: Term =>
     }
     thisItself
   }
+
   /*
    * Returns a Closed or Open list of posslble elements that could follow this Term, within its
    * lexically enclosing model group.
@@ -229,10 +231,16 @@ trait TermRuntime1Mixin { self: Term =>
       // term itself has not provided a match to the incoming event, and we need to
       // see if things following the term provide a match.
       //
-      case stb: SequenceTermBase =>
+      case stb: SequenceTermBase => {
         followingLexicalSiblingStreamingUnparserElements
-      case _ =>
-        possibleSelfPlusNextLexicalSiblingStreamingUnparserElements
+      }
+      case _: ElementBase | _: ChoiceTermBase => {
+        // If this element is closed, we only want it and not it's siblings
+        if (possibleThisTermNextStreamingUnparserElements.isClosed)
+          possibleThisTermNextStreamingUnparserElements
+        else
+          possibleSelfPlusNextLexicalSiblingStreamingUnparserElements
+      }
     }
     //
     // Check for ambiguity except for namespaces
@@ -332,24 +340,28 @@ trait TermRuntime1Mixin { self: Term =>
           res.getOrElse(Open(Nil))
         }
       }
-      val res = thisItself match {
-        //
-        // check this case separately first to avoid evaluating
-        // followingLexicalSiblingStreamingUnparserElements unless
-        // we have to.
-        //
-        case Closed(pnes1) => thisItself // most common case - a required element.
-        case _ => {
-          val res: PossibleNextElements =
-            (thisItself, followingLexicalSiblingStreamingUnparserElements) match {
-              case (Open(pnes1), Closed(Nil)) =>
-                thisItself // case of Open(...) followed by end of complex element.
-              case (poss1, Closed(pnes2)) => Closed((poss1.pnes ++ pnes2).distinct)
-              case (poss1, poss2) => Open((poss1.pnes ++ poss2.pnes).distinct)
-            }
-          res
+      Logger.log.debug(s"""
+        NextElementResolver -> thisItself: $thisItself\n
+        NextElementResolver -> following: $followingLexicalSiblingStreamingUnparserElements""")
+      val res: PossibleNextElements =
+        (thisItself, followingLexicalSiblingStreamingUnparserElements) match {
+          case (_, Closed(Nil)) => {
+            thisItself // case of Open(...) followed by end of complex element.
+          }
+          case (Closed(pnes1), Open(List())) => {
+            thisItself // case of Closed(...) followed by empty list, which is default for nothing
+          }
+          case (Closed(pnes1), Closed(pnes2)) => {
+            thisItself // When everything is closed, we only want thisItself and not the entire list of Closed possibilities
+          }
+          case (Open(pnes1), Closed(pnes2)) => {
+            // If there are optional elements before a required element, the whole sequence is considered closed
+            Closed((pnes1 ++ pnes2).distinct)
+          }
+          case (poss1, poss2) => {
+            Open((poss1.pnes ++ poss2.pnes).distinct)
+          }
         }
-      }
       res
     }.value
 
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
@@ -161,7 +161,16 @@ class TestChoice {
   @Test def test_direct_dispatch_17(): Unit = { runnerCH.runOneTest("direct_dispatch_17") }
 
   // DAFFODIL-2562
-  // @Test def test_dispatch_group_choice(): Unit = { runnerCH.runOneTest("dispatch_group_choice") }
+  @Test def test_dispatch_group_choice(): Unit = {
+    runnerCH.runOneTest("dispatch_group_choice")
+  }
+
+  @Test def test_dispatch_group_choice_2(): Unit = {
+    runnerCH.runOneTest("dispatch_group_choice_2")
+  }
+  @Test def test_dispatch_group_choice_3(): Unit = {
+    runnerCH.runOneTest("dispatch_group_choice_3")
+  }
 
   @Test def test_explicit_01(): Unit = { runnerCE.runOneTest("explicit_01") }
   @Test def test_explicit_02(): Unit = { runnerCE.runOneTest("explicit_02") }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/PartialNextElementResolver.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/PartialNextElementResolver.scala
@@ -18,6 +18,7 @@
 package org.apache.daffodil.runtime1.infoset
 
 import org.apache.daffodil.lib.exceptions.Assert
+import org.apache.daffodil.lib.util.Logger
 import org.apache.daffodil.lib.util.MStackOf
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.Maybe._
@@ -45,6 +46,7 @@ trait NextElementResolver { self: InfosetInputter =>
     nameSpace: String,
     hasNamespace: Boolean,
   ): ElementRuntimeData = {
+    Logger.log.debug(s"NextERDResovler -> trdStack: $trdStack")
     val iter = trdStack.iter
     iter.reset()
     var firstOne: Boolean = true
@@ -75,6 +77,10 @@ trait NextElementResolver { self: InfosetInputter =>
         // or it will issue an UnparseError if there are required elements in the
         // set of possible elements.
         //
+        Logger.log.debug(s"""
+          NextERDResovler -> current trd: $trd\n
+          NextERDResovler -> resolver: $resolver\n
+          NextERDResovler -> name: $name\n""")
         maybeERD = resolver.maybeNextElement(name, nameSpace, hasNamespace)
         // The cases where ERD is in a hidden context are addressed where the elements
         // are assigned their value i.e ElementUnparser
@@ -297,6 +303,10 @@ class SeveralPossibilitiesForNextElement(
     namespace: String,
     hasNamespace: Boolean,
   ): Maybe[ElementRuntimeData] = {
+    Logger.log.debug(s"""
+      NextERDResovler -> trd: $trd\n
+      NextERDResovler -> looking for: $local\n
+      NextERDResovler -> nextERDMap: $nextERDMap\n""")
     val optnextERD =
       if (hasNamespace) {
         val sqn = StepQName(
@@ -314,7 +324,7 @@ class SeveralPossibilitiesForNextElement(
           // It was statically determined at compile time that this
           // NextElementResolver does not have any possible NextERDs with the
           // same local name. Because of this, just find the first thing that
-          // matches and return it it is was found.
+          // matches and return it if it was found.
           nextERDMap.find(_._1.local == local).map(_._2)
         } else {
           // It was statically determined at compile time that some nextERDs
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section15/choice_groups/choice.tdml
@@ -1805,6 +1805,26 @@ it sure is/
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="dd10">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="prev" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="1" />
+          <xs:element name="key" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="1" />
+          <xs:choice dfdl:choiceDispatchKey="{ ./ex:key }">
+            <xs:sequence dfdl:choiceBranchKey="a">
+              <xs:element name="elt_a" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="3" />
+            </xs:sequence>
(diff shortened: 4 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
```diff
@@ -302,7 +302,7 @@ class SeveralPossibilitiesForNextElement(
       }
 
     val res = optnextERD
-    if (isRequiredStreamingUnparserEvent && res.isEmpty) {
+    if (isRequiredStreamingUnparserEvent && res.isEmpty && !trd.isInstanceOf[ChoiceRuntimeData]) {
```
stevedlawrence, line 305: I don't know enough about this code to know if this is the right fix, but here are some initial thoughts. Hoping @mbeckerle has a better idea.

I'm wondering if this issue isn't specific to choices? For example, say we had a shared group like which acts a little bit similar to the choice in this test:

```xml
  <group name="check">
    <sequence>
      <element name="not_valid" type="ex:str0" minOccurs="0"/>
    </choice>
  </group>
```

I guess this is different because in this case `isRequiredStreamingUnparseEvent` might be false because `not_valid` is optional which allows it to go up the stack?

Maybe that means in the case of this choice that has an empty branch `isRequiredStreamingUnparseEvent` should be false? I guess that isn't right though because the nackInfo/ackInfo that it's looking for next is required, it's just required outside the scope of this choice?

Another thought, should all choices skip this logic like you have now, or is this specific to choices that could have a branch with no element (e.g. empty sequence, all optional elements)?

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestSharedGroups.scala
```diff
@@ -35,6 +35,6 @@ class TestSharedGroups {
 
   @Test def testShowsSharedGroupOkFirstUse() = { runner.runOneTest("Ack") }
   // DAFFODIL-2615 - shared groups problem
-  //@Test def testShowsSharedGroupFailsSecondUse() = { runner.runOneTest("Nack") }
+  @Test def testShowsSharedGroupFailsSecondUse() = { runner.runOneTest("Nack") }
```
stevedlawrence, line 38: Is there any value in adding a negative test with a bad infoset like this:

```
          <messageID>2</messageID>
          <nack>X</nack>
          <ackInfo>A</ackInfo>
```

So we mix nack and ackinfo, so we ensure that the correct map is used and not some combined map or something where both nackInfo and ackInfo are accepted?

daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala
jadams-tresys: The only reason I think it might be choice specific is that yesterday while we were discussing this issue one of the things we tried was changing the choice inside the group to just a sequence and everything worked fine, so clearly wasn't hitting the same edge case.

As for the rest of your comment, I'll wait for Mike so weigh in as I'm really not sure what the best approach is.  Just saw a quick and easy way to fix it and saw that it worked for both shared groups (DAFFODIL-2615) and when the last branch of a choice in a group is optional (DAFFODIL-2562).

mbeckerle: It's not at all unlikely that your added predicate fixes an oversight in the original code. 
However, we do need a rationale for this, and we need to put a comment in that contains that rationale. 

This section of the code is checking whether we found a unique ERD corresponding to the incoming infoset element event, or not. 

The predicate you have enhanced is testing for 

1. the element is required to appear in the data stream (not optional element within the current group)
2. no ERD is found for it. (in this ERD resolver from this TRD on the trd stack)

You've added that it is also only an error if the current term is not a choice group.  If it is not an error, then the search will continue with the next resolver down the stack.  I think Steve is perhaps right that this is not about choices in general, but about choices with empty or all-optional branches. 

I'm inclined to think what is wrong is more likely to be that isRequiredStreamingUnparserEvent is true when it should be false.  That's my theory anyway. 

As for how to test this, ... Re: steve's thought: _Another thought, should all choices skip this logic like you have now, or is this specific to choices that could have a branch with no element (e.g. empty sequence, all optional elements)?_

We should create tests that try choices where there are branches which are not the first branch, and which contain: 
- nothing (empty sequence)
- syntax only (like a terminator, but on an otherwise empty sequence)
- optional content only (two cases: a nested choice with an empty sequence branch, and minOccurs=0)

We definitely also need a test where the shared group is a sequence with only optional content. 

Once we have those tests, I think we'll be in a better position to write a rationale and understand if we're missing a concept for choices, or if the problem is with isRequiredStreamingUnparserEvent, or ... something else entirely.

mbeckerle: Following up to my own comment. I think the issue is that the choice whose TRD is on top of stack is either Open or Closed based on analysis in the schema compiler. The isRequiredStreamingUnparserEvent should be false if the choice is Open. A choice is closed if every branch is closed. A sequence is closed if it contains a required element. I suspect there is some flaw in the Open/Closed logic leading to incorrect isRequiredStreamingUnparser event.

jadams-tresys: I've spent some time digging into the Open/Closed choice branches and it seems like we are doing the right thing with the choice branches where the Choice doesn't close until all the branches are closed.  However, at the bottom of possibleSelfPlusNextLexicalSiblingStreamingUnparserElements we have the following match statement:

`      val res = thisItself match {                                                                                                                                              
        //                                                                                                                                              
        // check this case separately first to avoid evaluating                                                                                                                                              
        // followingLexicalSiblingStreamingUnparserElements unless                                                                                                                                              
        // we have to.                                                                                                                                              
        //                                                                                                                                              
        case Closed(pnes1) => thisItself // most common case - a required element.                                                                                                                                              
        case _ => {                                                                                                                                              
          val res: PossibleNextElements =                                                                                                                                              
            (thisItself, followingLexicalSiblingStreamingUnparserElements) match {                                                                                                                                              
              case (Open(pnes1), Closed(Nil)) => {                                                                                                                                              
                System.err.println("Open + Closed")                                                                                                                                              
                thisItself // case of Open(...) followed by end of complex element.                                                                                                                                              
              }                                                                                                                                              
              case (poss1, Closed(pnes2)) => {                                                                                                                                              
                System.err.println("Closed")                                                                                                                                              
                Closed((poss1.pnes ++ pnes2).distinct)                                                                                                                                              
              }                                                                                                                                              
              case (poss1, poss2) => {                                                                                                                                              
                System.err.println("Open")                                                                                                                                              
                Open((poss1.pnes ++ poss2.pnes).distinct)                                                                                                                                              
              }                                                                                                                                              
            }                                                                                                                                              
          res                                                                                                                                              
        }
`

You can ignore the print statements that I added.  For these choice we are hitting the second case and therefore returning a Closed list of possible elements, which results in isRequiredStreamingUnparserEvent being true and triggers the ErrorERD being created and we get a failure.  I have tried changing the 2nd case to the following:

`case (poss1, Closed(pnes2)) if (!this.isInstanceOf[ChoiceTermBase]) => { `

Which results in our use case hitting the 3rd case and gets past the error we are seeing.  Note, this does not fix the issue in DAFFODIL-2562 so if this is an acceptable solution, more work will need to be done to support that issue.

jadams-tresys: @mbeckerle Could not get the code formatted properly in that comment, but does that seem like a reasonable fix?

stevedlawrence: Easiest way to include code is to wrap it in three backticks, you can also put `scala` after the first set to get the right highlighting:

````
```scala
// scala code here
val foo = ....
```
````

Comment (mbeckerle) [reply daffodil-runtime1/src/main/scala/org/apache/daffodil/infoset/PartialNextElementResolver.scala:305]: We just need a lot more information to know if this is sensible or not. 

How about print out the term and the value of possibleThisTermNextStreamingUnparserElementsDef each time it is computed. 

Then at runtime, let's display every pushTRD and popTRD, and each time we try to resolve a incoming event, let's print out the result of trying each partial next element resolver.

File: daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala [hunk 3]
```diff
@@ -38,103 +38,103 @@ class TestChoice {
...
-  @Test def test_direct_dispatch_11(): Unit = { runnerCH.runOneTest("direct_dispatch_11") }
-  @Test def test_direct_dispatch_12(): Unit = { runnerCH.runOneTest("direct_dispatch_12") }
-  @Test def test_direct_dispatch_13(): Unit = { runnerCH.runOneTest("direct_dispatch_13") }
-  @Test def test_direct_dispatch_14(): Unit = { runnerCH.runOneTest("direct_dispatch_14") }
-  @Test def test_direct_dispatch_15(): Unit = { runnerCH.runOneTest("direct_dispatch_15") }
-  @Test def test_direct_dispatch_16(): Unit = { runnerCH.runOneTest("direct_dispatch_16") }
-  @Test def test_direct_dispatch_17(): Unit = { runnerCH.runOneTest("direct_dispatch_17") }
-
-  // DAFFODIL-2562
+  //@Test def test_choice_noBranch (): Unit = { runnerCH.runOneTest("choice_noBranch") }
```
mbeckerle, line 41: Pretty sure you didn't want to check this in.

jadams-tresys: More like I forgot that pushing up a commit would automatically update this pull request.  My machine had been acting up so I wanted to push things up to my branch for safe keeping, just forgot it would update this pull request.  I think I might actually delete this now that I have a better understanding of what the problem is.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
```diff
@@ -229,10 +230,16 @@ trait TermRuntime1Mixin { self: Term =>
       // term itself has not provided a match to the incoming event, and we need to
       // see if things following the term provide a match.
       //
-      case stb: SequenceTermBase =>
+      case stb: SequenceTermBase => {
         followingLexicalSiblingStreamingUnparserElements
-      case _ =>
-        possibleSelfPlusNextLexicalSiblingStreamingUnparserElements
+      }
+      case _ => {
```
mbeckerle, line 236: ```case _ ``` leaves something to be desired here. 

I think the possibilities are ElementBase or ChoiceTermBase.

Can we do 

```case _: ElementBase | _: ChoiceTermBase => ... ```

File: daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala [hunk 5]
```diff
@@ -332,24 +339,27 @@ trait TermRuntime1Mixin { self: Term =>
...
-        case _ => {
-          val res: PossibleNextElements =
-            (thisItself, followingLexicalSiblingStreamingUnparserElements) match {
-              case (Open(pnes1), Closed(Nil)) =>
-                thisItself // case of Open(...) followed by end of complex element.
-              case (poss1, Closed(pnes2)) => Closed((poss1.pnes ++ pnes2).distinct)
-              case (poss1, poss2) => Open((poss1.pnes ++ poss2.pnes).distinct)
-            }
-          res
+      // System.err.println("thisItself: " + thisItself)
```
mbeckerle, line 342: Either delete these println, or insert a line ```// keep these println for debug``` at least.

But it would be better to convert these to logging, and make the messages a bit less terce.

I'd like to think that turning on the logging on some reasonably sized test case, one could reason from the output to convince oneself it is working right.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
```diff
@@ -332,24 +339,27 @@ trait TermRuntime1Mixin { self: Term =>
...
+      // System.err.println("following: " + followingLexicalSiblingStreamingUnparserElements)
+      val res: PossibleNextElements =
+        (thisItself, followingLexicalSiblingStreamingUnparserElements) match {
+          case (poss1, Closed(Nil)) => {
```
mbeckerle, line 346: poss1 unused. Can be _

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
```diff
@@ -332,24 +339,27 @@ trait TermRuntime1Mixin { self: Term =>
...
+            thisItself // case of Closed(...) followed by empty list, which is default for nothing
+          }
+          case (Closed(pnes1), Closed(pnes2)) => {
+            thisItself // When everything is closed, we only want thisItself and not the entire list of Closed possibilities
+          }
+          case (Open(pnes1), Closed(pnes2)) => {
+            // If there are optional elements before a required element, the whole sequence is considered closed
+            Closed((pnes1 ++ pnes2).distinct)
+          }
+          case (poss1, poss2) => {
```
mbeckerle, line 359: This case is ```case (Open(pnes1), Open(pnes2)) => ```
It would be better to write it that way. Scala *should* detect that the match case is exhaustive.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
```diff
@@ -161,7 +161,9 @@ class TestChoice {
   @Test def test_direct_dispatch_17(): Unit = { runnerCH.runOneTest("direct_dispatch_17") }
 
   // DAFFODIL-2562
-  // @Test def test_dispatch_group_choice(): Unit = { runnerCH.runOneTest("dispatch_group_choice") }
+  @Test def test_dispatch_group_choice(): Unit = {
```
mbeckerle, line 164: Do we really just have this one test? Is this functionality well covered by other tests?
Does this one test really prove/convince that this is all working properly?

mbeckerle (changes requested):
A few things. 

I'm surprised you didn't add more tests.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
jadams-tresys: Definitely need to add more.  I had just gotten things to a point where it seemed I had a reasonable fix and all tests were passing.  Wanted to get this pushed up for review to make sure things looked good.  Going to add more tests now.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
jadams-tresys: Actually, leaving it like this also absorbs the DoNotUse cases, which matches the behavior we had previously.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
jadams-tresys: I added a couple more tests just to verify that having the same choice outside of the group doesn't change anything.  I also verified that all the cases of Open/Closed match statement in TermRuntimeMixin1.scala are being hit, as that was the bulk of the changes for this commit.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
jadams-tresys: They have now been changed to log debug statements.

mbeckerle (approved):
+1 This is an important fix. 

I only have one more suggestion. Add a string to your log.debug statements identifying that this topic is what they are about. Like "NextERDResolver: " or something so they are labeled.

daffodil-test/src/test/scala/org/apache/daffodil/section15/choice_groups/TestChoice.scala
```diff
@@ -161,7 +161,16 @@ class TestChoice {
...
+    runnerCH.runOneTest("dispatch_group_choice")
+  }
+
+  @Test def test_dispatch_group_choice_2(): Unit = {
+    runnerCH.runOneTest("dispatch_group_choice_2")
+  }
+  @Test def test_dispatch_group_choice_3(): Unit = {
+    runnerCH.runOneTest("dispatch_group_choice_3")
+  }
```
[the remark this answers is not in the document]
jadams-tresys: I have, but I'm seeing a lot of failures to compile that seem to be related to something else with DAFFODIL-3.5.0.  For what it's worth I don't think I'm getting any additional failures over the main branch.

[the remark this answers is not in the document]
mbeckerle: Please send the output from the 3.5.0 base regression to dev list. The fact that some of those tests are for Owl-specific/FOUO formats non-Owl people don't have access to doesn't matter. We all need visibility to 3.5.0 regression status.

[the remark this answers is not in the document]
jadams-tresys: @mbeckerle Should I go ahead and merge this PR or did we want to wait until we resolved the regression test stuff?  @mike-mcgann confirmed that he was seeing the same sort of issues as I was with the 3.5.0 main branch

PR #934, opened 2023-01-25
Fix handling of NaN/Infinifty when parsing text integers

stevedlawrence:
When parsing text integers, ICU still attempts to parse NaN and infinty representations and returns Double.NaN, Double.POSITIVE_INFINITY, or Double.NEGATIVE_INFINITY if matched. There is no way to disable this behavior in ICU.

So this modifies the ConvertTextStandardNumberParser to throw a PE when the prim type is an integer and ICU returns a Double, which means it must have parsed an NaN/Inf which is not considered valid by DFDL.

This also adds a new isInteger method to PrimNumeric to more easily determine if a numeric primitive type is an integer without having to do a match-case on types, and uses this new method in a few places.

Also modifies PrimNumericInteger so that doubles and floats that are NaN or infinite are not considered valid for integer types. This ensures we error instead of converting to zero, such as when used in expressions or reading values from an infoset.

Also adds an ICU specific test to detect if ICU ever changes this behavior.

Note that the original bug expects an SDE, but this is not correct. When and integer fails to parse because NaN or INF is the value, it should be a PE. It is also a PE if a length expression tries to convet a double NaN/INF to a long. It is only an SDE if the length value returned is negative. Tests are aded to verify this behavior.

[DAFFODIL-908](https://issues.apache.org/jira/browse/DAFFODIL-908)

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
@@ -173,27 +173,32 @@ case class ConvertTextStandardNumberParser(
         val df = textNumberFormatEv.evaluate(start).get
         val strCheckPolicy = if (df.isParseStrict) str else str.trim
         val pos = new ParsePosition(0)
-        val icuNum: Number = df.parse(strCheckPolicy, pos)
-
-
-        if (icuNum == null) {
-          PE(start, "Unable to parse %s from text: %s",
-            context.optPrimType.get.globalQName, str)
-          return
-        }
-
-        // sometimes ICU will return their own custom BigDecimal, even if the
-        // value could be represented as a BigInteger. We only want Java types,
-        // so detect this and convert it to the appropriate type
-        val num1 = icuNum match {
+        val icuNum: Number = df.parse(strCheckPolicy, pos) match {
+          case null => {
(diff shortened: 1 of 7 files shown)


mbeckerle (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
```diff
@@ -173,27 +173,32 @@ case class ConvertTextStandardNumberParser(
...
+          }
           case bd: ICUBigDecimal => {
+            // sometimes ICU will return their own custom BigDecimal, even if the
+            // value could be represented as a BigInteger. We only want Java types,
+            // so detect this and convert it to the appropriate type
             if (bd.scale == 0) bd.unscaledValue
             else bd.toBigDecimal
           }
-          case _ => icuNum
+          case num => num
```
tuxji, line 199: Can we use `case num: Number => num` to be clearer or would that cause Scala to complain that a default case is still missing?

tuxji (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/ConvertTextStandardNumberParser.scala
stevedlawrence: Scala has no problem with it and I guess is able to detect all cases are covered. Will add.

PR #935, opened 2023-01-25
Fix generated C code for offset between 2 array slots

tuxji:
The C code generator interpolates an ElementBase instead of a String into the computation for the offset between 2 array slots, which can result in a namespace prefix being printed as well as the local name, causing the C compiler to issue an error about the semi-colon.

CodeGeneratorState.scala: Rename addArrayImplementation's parameter from "e" to "elem".  Define a local constant String value "e" with the elem's local name.  Keep interpolating $e into the computation.

DAFFODIL-2777


stevedlawrence (approved):
+1

PR #936, opened 2023-01-26
Only add positionally required separators for arrays with occursCountKind="implicit"

stevedlawrence:
If a schema has a separated array/optional with occursCountKind as fixed or expression, then we currently always unparse maxOccurs separators even if there were less than maxOccurs instances in the infoset. This is incorrect--we should only output a separator for each instance in the infoset and no more, regardless of the value of min/maxOccurs. Only with occursCountKind set to implicit should min/maxOccurs be taken into account when unparsing positionally required separators.

The unparsePositionallyRequiredSeps function handles the separator logic for implicit arrays, but should be skipped otherwise. This modifies the condition so it is only considered for the implicit case.

Also makes the "ock" variable in the MinMaxRepatsMixin public since it is useful in a number of places, including this new check.

[DAFFODIL-2231](https://issues.apache.org/jira/browse/DAFFODIL-2231)

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/UnseparatedSequenceUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/UnseparatedSequenceUnparsers.scala
@@ -43,8 +43,6 @@ class RepOrderedUnseparatedSequenceChildUnparser(
   extends RepeatingChildUnparser(childUnparser, srd, erd)
   with Unseparated {
 
-  private val ock = erd.maybeOccursCountKind.get
-
   override def checkArrayPosAgainstMaxOccurs(state: UState): Boolean = {
     if (ock eq OccursCountKind.Implicit)
       state.arrayPos <= maxRepeats(state)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
@@ -460,7 +460,7 @@ trait MinMaxRepeatsMixin {
 
   def erd: ElementRuntimeData
 
-  private val ock = erd.maybeOccursCountKind.get
+  final val ock = erd.maybeOccursCountKind.get
 
   private val minRepeats_ = {
     val mr =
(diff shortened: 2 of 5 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/UnseparatedSequenceUnparsers.scala
```diff
@@ -43,8 +43,6 @@ class RepOrderedUnseparatedSequenceChildUnparser(
   extends RepeatingChildUnparser(childUnparser, srd, erd)
   with Unseparated {
 
-  private val ock = erd.maybeOccursCountKind.get
-
```
tuxji, line 47: I don't see an override here.  Was this a redundant definition of ock that Scala didn't warn about or what?

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
```diff
@@ -460,7 +460,7 @@ trait MinMaxRepeatsMixin {
 
   def erd: ElementRuntimeData
 
-  private val ock = erd.maybeOccursCountKind.get
+  final val ock = erd.maybeOccursCountKind.get
```
tuxji, line 463: The `final` was to help find the redundant definitions, yes?

tuxji (approved):
+1

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/processors/unparsers/UnseparatedSequenceUnparsers.scala
stevedlawrence: The one from MinMaxRepeatsMixin was private, so this couldn't use the one from the mixin and so had to redefine it. If the one in MinMaxRepeatsMixin was protected/public, I imagine scala would have created a warning. But since they were the same, we can just use the definition from MinMaxRepeatsMixin.

daffodil-runtime1/src/main/scala/org/apache/daffodil/processors/parsers/SequenceChildBases.scala
stevedlawrence: Yeah, just to make sure it's defined in only a single place so we don't accidentally override it with something that's not right.

mbeckerle (approved):
+1

PR #938, opened 2023-01-27
Add test for separatorSuppressionPolicy='never' (WIP)

mike-mcgann:
The bug in the referenced ticket appears to have been fixed but shows that there is no test coverage for when the separatorSupressionPolicy is set to never. This update adds a test for this case.

[DAFFODIL-2219](https://issues.apache.org/jira/browse/DAFFODIL-2219)


daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequenceGroupDelimiters.tdml
```diff
@@ -614,4 +614,37 @@
...
+    <tdml:document>
+      <tdml:documentPart type="text"
+                         replaceDFDLEntities="true"><![CDATA[Hello%CR;%LF;world%CR;%LF;]]></tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <tns:policyNever>
+          <tns:A>Hello</tns:A>
+          <tns:B>world</tns:B>
+          <tns:C></tns:C>
```
mbeckerle, line 644: The data looks correct. 

The only issue is whether this C element should be here, or be removed because it's empty. That depends on property emptyElementParsePolicy. If that has value "treatAsEmpty" then this is the correct behavior. If "treatAsAbsent" then I think this element should not appear in the infoset.

mbeckerle (approved):
+1 Just verify that property emptyElementParsePolicy is 'treatAsEmpty'.

daffodil-test/src/test/resources/org/apache/daffodil/section14/sequence_groups/SequenceGroupDelimiters.tdml
mike-mcgann: Confirmed that emptyElementParsePolicy is 'treatAsEmpty'. In this test case it simply uses the value provided in the general format.

daffodil-test/src/test/scala/org/apache/daffodil/usertests/RC-Test.scala
```diff
@@ -17,30 +17,23 @@
...
 }
 
-class RCTest5 {
+class RCTest {
 
-  import RCTest5._
-
-  @Test def test5p(): Unit = { runner.runOneTest("parse-test-5") }
-
-  // DAFFODIL-2217
```
stevedlawrence, line 42: Does this also resolve part of DAFFODIL-2217, or were these issues actually related to 2219 are were labeled incorrectly?

daffodil-test/src/test/scala/org/apache/daffodil/usertests/RC-Test.scala
```diff
@@ -17,30 +17,23 @@
...
-  // @Test def test5u() { runner.runOneTest("unparse-test-5") }
+  //import RCTest._
 
   // DAFFODIL-2219
```
stevedlawrence, line 38: Is this test fixed as part of this PR, or is 2219 not full resolved?

mike-mcgann: It looks like the ticket references were swapped in the source code. [DAFFODIL-2219](https://issues.apache.org/jira/browse/DAFFODIL-2219) is resolved and that was for unparse-test-5. The issue referenced in parse-test-6 is ticket [DAFFODIL-2217](https://issues.apache.org/jira/browse/DAFFODIL-2217) and is still open.

daffodil-test/src/test/scala/org/apache/daffodil/usertests/RC-Test.scala
mike-mcgann: See below.

stevedlawrence (approved):
+1 :+1:  please squash and this is good to merge

PR #939, opened 2023-01-27
Avoid CI false negatives

stevedlawrence:
Two separate commits for easier review:

####  Adjust CLI test rig to avoid timeouts

Sometimes it can take a while for the CLI thread to start due to
resource constraints. This means the ExpectIt tests could fail just
waiting for the CLI start up, giving us false negatives.

To avoid this, we now configure the ExpectIt tests to never timeout and
run them in a separate thread. The main thread then waits for the CLI to
complete (with an infinite duration, it's better for a bad test to hang
than for a good test to time out due to resource constraints). After the
CLI completes, only then does the main thread wait the specified timeout
to allow the expect tests to finish. If those tests don't finish at this
point, then we interrupt the test thread and throw whatever exception is
created.

With this approach, we can ensure the expensive CLI logic is complete,
before we start considering the timeout, which should give he much less
expensive ExpectIt logic enough time to finish, especially since it does
not need to wait for CLI logic.

[DAFFODIL-2751](https://issues.apache.org/jira/browse/DAFFODIL-2751)

#### Adjust socket tests to avoid time outs

- Create socket producer/consumer pairs in the main thread without a
  timeout. The tests require these sockets to be created so just wait
  however long it takes. Sometimes it may take a while to create the
  sockets due to resource constraints which would lead to false
  negatives if we timed out.
- If a particular code hunk should never time out (unless the test
  breaks), then remove the withTimeout call. This means if the test ever
  fails we might hang forever, but that is unlikely, and it's better
  than getting random false negatives and trying to determine if they
  are spurious or not
- For tests that actually do test for a hang and require a timeout,
  adjust timeouts to be at least 1 second. This gives us more confidence
  that the test actually timed out and not that it just took too long to
  run the test

[DAFFODIL-2751](https://issues.apache.org/jira/browse/DAFFODIL-2751)

### DIFF
--- a/daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
+++ b/daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
@@ -170,9 +170,8 @@ object Util {
    *   runs the CLI in a new process instead of a thread and will likely decrease
    *   performance
    * @param fork if true, forces the the CLI to run in a new process
-   * @param timeout how long to wait, in seconds, for the CLI to exit after the
-   *   testFunc has returned. Also how long to wait for individual expect
-   *   operations in the CLITester
+   * @param timeout how long to wait, in seconds, for the testFunc to finish after
+   *   the CLI has completed. Test testFunc is interrupted if this timeout is reached
    * @param debug if true, prints arguments and classpath information to
    *   stdout. Also echos all CLITester input and output to stdout.
    * @param testFunc function to call to send input to the CLI and validate
@@ -187,7 +186,7 @@ object Util {
     (testFunc: (CLITester) => Unit)
     (expectedExitCode: ExitCode.Value): Unit = {
 
-    val (toIn, fromOut, fromErr, threadOrProc: Either[CLIThread, Process]) =
+    val (toIn, fromOut, fromErr, cliThreadOrProc: Either[CLIThread, Process]) =
       if (classpaths.nonEmpty || fork) {
         // spawn a new process to run Daffodil, needed if a custom classpath is
         // defined or if the caller explicitly wants to fork
@@ -228,11 +227,20 @@ object Util {
         (toIn, fromOut, fromErr, Left(thread))
       }
 
+    // Create an ExpectIt object that reads/writes the streams created above.
     val eb = new ExpectBuilder()
     eb.withOutput(toIn)
     eb.withInputs(fromOut, fromErr)
     eb.withInputFilters(replaceInString("\r\n", "\n"))
-    eb.withTimeout(timeout, TimeUnit.SECONDS)
+    // Disable timeouts on expect calls. We do this because often times the CLI can
+    // take a while to start up (e.g. spawning a thread, schema compilation) which
+    // can lead to timeouts and false negatives. Instead, we will spawn the expect
+    // body in a separate TestThread. Only once the CLI finishes do we start a
+    // timeout countdown and require the test body logic to complete in a specified
+    // number of seconds. If if doesn't finish, then we interrupt the test thread and
+    // capture any errors. So although expect will never timeout itself, we may time
+    // it out after the CLI is complete.
+    eb.withInfiniteTimeout()
     eb.withExceptionOnFailure()
     if (debug) {
       eb.withEchoOutput(System.out)
@@ -241,24 +249,116 @@ object Util {
     val expect = eb.build()
     val tester = new CLITester(expect, toIn)
 
-    try {
-      testFunc(tester)
-    } finally {
-      threadOrProc match {
-        case Left(thread) => thread.join(timeout * 1000)
-        case Right(process) => process.waitFor(timeout, TimeUnit.SECONDS)
+    // run the test function in a new thread. We do this so that we can configure
+    // ExpectIt to have an infinite timeout. This way we can trigger a timeout only
+    // after the CLI has exited, at which point we give the test thread timeout
+    // seconds to finish its tests and interrupt it if it still hasn't finished. This
+    // way if the CLI is slow to start up, ExpectIt won't create a timeout exception
+    // and fail the test. We at least let the CLI finish before doing that
+    val testThread = new TestThread(testFunc, tester, cliThreadOrProc)
+    testThread.start()
+
+    // wait for the CLI process/thread to finish with no timeout--we really don't
+    // know how long the CLI is going to take, but we assume the CLI will eventually
+    // finish. If it doesn't, we'll see the hang and investigate. Note that if there
+    // is a hang here, it is most likely a bad test. The most common cause is the CLI
+    // is waiting for something on stdin (e.g. EOF, debugger command), but the test
+    // body is not providing it. For CLI tests taking significantly longer than
+    // expected (6x the timeout, which is 1 minute for the default timeout), we can
+    // assume it is hung and output a message with helpful reference to possible
+    // causes. We still don't kill it though, since we might just be very resources
+    // starved and it might finish eventually.
+    val cliWarningTimeoutMS = 6 * timeout * 1000
+    var isAlive = true
+    while (isAlive) {
+      isAlive = cliThreadOrProc match {
+        case Left(cliThread) => {
+          cliThread.join(cliWarningTimeoutMS)
+          cliThread.isAlive
+        }
+        case Right(cliProcess) => {
+          cliProcess.waitFor(cliWarningTimeoutMS, TimeUnit.MILLISECONDS)
+          cliProcess.isAlive
+        }
       }
-      expect.close()
-      toIn.close()
-      fromOut.close()
-      fromErr.close()
+      if (isAlive) {
+        System.err.println("CLI test may be hanging, see DAFFODIL-2751 for possible causes")
+      }
     }
 
-    val actualExitCode = threadOrProc match {
-      case Left(thread) => thread.exitCode
-      case Right(process) => ExitCode(process.exitValue)
+    // now that the CLI has finished, we give the test thread timeout seconds to
+    // finish verifying the output. Hopefully it was doing all the checks in parallel
+    // with the CLI thread so it should be mostly complete and finish in time.
+    testThread.join(timeout * 1000)
+
+    // If the test thread still isn't done, assume it failed and is stuck expecting
+    // something that isn't ever coming from the CLI. Interrupt the thread to end it
+    // and join to wait for it to cleanup and die. Although this join is technically
+    // infinite, as long as the expect library or test body doesn't explicitly
+    // capture and ignore the interrupt and then block on something, this should not
+    // hang
+    testThread.interrupt()
+    testThread.join()
+
+    // the test thread is finally done, clean up our in/out/err buffers
+    expect.close()
+    toIn.close()
+    fromOut.close()
+    fromErr.close()
+
+    // if the test thread didn't end cleanly then it must have thrown an exception
+    // (e.g. assertion failed, interrupted exception). Just rethrow that exception
+    // and cause the test to fail
+    testThread.optException.map { e => throw e }
+
+    // if the test thread didn't throw an exception then that means all of its tests
+    // passed. We just need to verify the CLI exit code
+    val actualExitCode = cliThreadOrProc match {
+      case Left(cliThread) => cliThread.exitCode
+      case Right(cliProcess) => ExitCode(cliProcess.exitValue)
     }
     assertEquals("Incorrect exit code,", expectedExitCode, actualExitCode)
+  }
+
+  /**
+   * Run CLITester function body in a new thread
+   *
+   * If an exception was thrown during evaluation it is stored in the optException
+   * variable. This thread will not exit until the cli thread/process has completed
+   */
+  private class TestThread(
+    testFunc: (CLITester) => Unit,
+    tester: CLITester,
+    cli: Either[Thread, Process])
+    extends Thread {
+
+    var optException: Option[Throwable] = None
+
+    private val exceptionHandler = new Thread.UncaughtExceptionHandler {
+      def uncaughtException(t: Thread, e: Throwable): Unit = {
+        optException = Some(e)
+      }
+    }
+
+    override def run(): Unit = {
+      // if this thread throws an exception (e.g. an expect function fails, this
+      // thread is interrupted), we just handle it and store it in optException above
+      this.setUncaughtExceptionHandler(exceptionHandler)
+
+      // now run the test
+      testFunc(tester)
+
+      // it is possible we finished the test before the CLI actually finished.
+      // Unfortunately, we can't let this thread die until the CLI is complete
+      // because ExpectIt has some magic that will close the pipes we set up and
+      // cause the CLI to get a pipe broken exception instead of finishing. So we
+      // just wait for the CLI to finish. Once the CLI is done this thread can
+      // cleanly die and the main thread can figure out what happened.
+      cli match {
+        case Left(thread) => thread.join()
+        case Right(process) => process.waitFor()
+      }
+    }
   }
 
   /**
@@ -267,7 +367,7 @@ object Util {
    * stdin/out/err.
    */
   private class CLIThread(args: Array[String], in: InputStream, out: OutputStream, err: OutputStream) extends Thread {
-    var exitCode = ExitCode.Failure
+    var exitCode: ExitCode.Value = _
 
     override def run(): Unit = {
       val psOut = new PrintStream(out)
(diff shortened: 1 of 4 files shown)


daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -241,33 +241,101 @@ object Util {
...
+    testThread.join(timeout * 1000)
+
+    // If the test thread still isn't done, assume it failed and is stuck expecting
+    // something that isn't coming. Interrupt the thread to end it. If it is already
+    // finished, this does nothing.
+    testThread.interrupt()
+
+    // at this point we may need to wait for the test thread to die from the
+    // interrupt. Expect might be doing some cleanup as it handles the interruption
+    while (testThread.isAlive) Thread.sleep(100)
```
tuxji, line 274: In your pull request, you say:

> If a particular code hunk should never time out [...]. This means if the test ever fails we might hang forever, but that is unlikely, and it's better than getting random false negatives and trying to determine if they are spurious or not

Is this line the place you were talking about where we might hang forever?  If so, please include the second quoted sentence in the comment preceding this line so that readers will know we already considered the possibility of the while loop never terminating and are okay with it.

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -241,33 +241,101 @@ object Util {
...
+
+    // the test thread is finally done, clean up our in/out/err buffers
+    expect.close()
+    toIn.close()
+    fromOut.close()
+    fromErr.close()
+
+    // if the test thread didn't end cleanly then it must have throw an exception
```
tuxji, line 282: *throw -> thrown

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -241,33 +241,101 @@ object Util {
...
       case Right(process) => ExitCode(process.exitValue)
     }
     assertEquals("Incorrect exit code,", expectedExitCode, actualExitCode)
   }
 
+  /**
+   * Run CLITester functionb body in a new thread
+   *
+   * If an exception was thrown during evaluation it is store in he optException
+   * variable. This thread will not exit until the cli thread/process has completed
```
tuxji, line 300: *functionb -> function

*store in he -> stored in the

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -241,33 +241,101 @@ object Util {
...
+   */
+  private class TestThread(testFunc: (CLITester) => Unit, tester: CLITester, cli: Either[Thread, Process]) extends Thread {
```
tuxji, line 302: Long line, better formatted like:

```scala
class Class(
  arg1,
  arg2,
) extends Base {
```

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -278,7 +346,17 @@ object Util {
       Main.setInputOutput(in, psOut, psErr)
       configureLog4j(psErr)
 
-      exitCode = Main.run(args)
+      try {
+        exitCode = Main.run(args)
+      } catch {
+        case t: Throwable => {
+          // Main.run should never throw an exception so if it did it means the CLI
+          // hit a bug. So just print it the exception and set the exit code the
```
tuxji, line 354: line 354 should say: `So just print the exception and set the exit code to`

tuxji (approved):
+1

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
mbeckerle: I don't like these unbounded loops. 

I think if this spins 10 times (1 full second) or maybe 20 times, then it should start outputting something to somewhere like stderr indicating that it seems to be stuck.

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
```diff
@@ -228,11 +227,12 @@ object Util {
         (toIn, fromOut, fromErr, Left(thread))
       }
 
+    // Create an ExpectIt object that reads/writes the streams created above
     val eb = new ExpectBuilder()
     eb.withOutput(toIn)
     eb.withInputs(fromOut, fromErr)
     eb.withInputFilters(replaceInString("\r\n", "\n"))
-    eb.withTimeout(timeout, TimeUnit.SECONDS)
+    eb.withInfiniteTimeout()
```
mbeckerle, line 235: I'd prefer if we wrote our own thing that waits forever, but  that polls every 10ms or so, and after 2 seconds starts outputting messages to stderr or something. It can output 1 message every 5 seconds or something so as not to generate massive output if things do hang. 

The other wait forever loop and this one should identify themselves in the messages they display so we can tell them apart.

mbeckerle (approved):
+1 However, I suggest avoiding these silent waitforever situations. There are two I think, and rather than wait forever, we should wait some sensible amount of time and then start creating a low-volume of messages to inform about what is going on. It's still wait forever, just not fully silent.

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
stevedlawrence: That's kind of what happens now, but I think it's actually more useful than a "thread taking too long" kind of message.

With this change, we tell expect to never timeout (i.e. infinite timeout), but then run all the expect logic in a separate thread (the new TestThread class). Once the CLI is done, we give `timeout` seconds for the test thread to finish up all its checks and interrupt the thread if it reaches that timeout. That interrupt causes the test thread to die, we capture the exception caused by the interrupt, and rethrow it in the main thread for sbt to output.

So essentially we tell expect to not do timeout logic and we do it ourselves. The idea being that we don't want expect to timeout until after the CLI is done. Otherwise long CLI startup times can lead to expect failures.

Note that if it does hit that timeout (e.g. if I add a long `Thread.sleep` in a test body), we get something like this:

```
Test org.apache.daffodil.cliTest.TestCLIparsing.test_3677_CLI_Parsing_elementFormDefault_qualified failed: java.lang.InterruptedException: sleep interrupted, took 11.973 sec
[error]     at java.lang.Thread.sleep(Native Method)
[error]     at org.apache.daffodil.cliTest.TestCLIparsing.$anonfun$test_3677_CLI_Parsing_elementFormDefault_qualified$1(TestCLIParsing.scala:34)
```

Or if a test body expects something that doesn't exist, we get this:
```
[error] Test org.apache.daffodil.cliTest.TestCLIparsing.test_3677_CLI_Parsing_elementFormDefault_qualified failed: net.sf.expectit.ExpectIOException: Expect operation fails (timeout: -1 ms) for matcher: contains('<tns:e1>does not exist</tns:e1>'), took 4.863 sec
[error]     at net.sf.expectit.ExpectImpl.expectIn(ExpectImpl.java:106)
[error]     at net.sf.expectit.AbstractExpectImpl.expectIn(AbstractExpectImpl.java:57)
[error]     at net.sf.expectit.AbstractExpectImpl.expect(AbstractExpectImpl.java:61)
[error]     at org.apache.daffodil.CLI.Util$CLITester.expect(Util.scala:464)
[error]     at org.apache.daffodil.cliTest.TestCLIparsing.$anonfun$test_3677_CLI_Parsing_elementFormDefault_qualified$1(TestCLIParsing.scala:34)
```

In both of these examples, we get a line number that says exactly which line in the testBody expect was at when it was interrupted.

I'll add some comments to make it more clear and why this infinite timeout isn't really infinite.

daffodil-cli/src/test/scala/org/apache/daffodil/CLI/Util.scala
stevedlawrence: > Is this line the place you were talking about where we might hang forever?

This is one potential place. Really the only way for this loop to hang is if the expect library ignores the above interrupt (e.g. catches an `InterruptedException`, ignores it, and continues doing expect stuff). In my development of this stuff, it seems the expect library handles interrupts as expected. And I do see logic in the expect code to make sure it fails expect()s if it has been interrupted.

And now that I think about it, I would actually prefer to change this line to `testThread.join()` to wait until the test thread dies from the interrupt. I think `join` is just implemented as a very similar `isAlive` loop, so this would be cleaner.

Another place things could potentially hang with these changes is in this code above:
```scala
   threadOrProc match {
      case Left(thread) => thread.join()
      case Right(process) => process.waitFor()
    }
```
So if the CLI gets stuck in an infinite loop or is blocked waiting for input then this will hang forever. There really isn't much looping logic in the CLI itself, so if this did hang it probably means there is a bug Daffodil core causing an infinite loop, which would probably also cause TDML tests to hang. Or more likely it means the test body is broken and the CLI is expecting input on stdin and the test body isn't providing it.

The other places that could now hang are the tests related to socket reading where there are potentially blocking reads. But those are very unlikely to hang since the tests are written correctly so the reads shouldn't block, and Java isn't likely to change the blocking behavior of sockets, so our tests making sure reads don't block will likely never fail and never need to change.

I do agree that it's worth adding comments to these places in the CLI where a hang is theoretically possible and what could cause it.

> I think if this spins 10 times (1 full second) or maybe 20 times, then it should start outputting something to somewhere like stderr indicating that it seems to be stuck.

Just playing devils advocate, we don't do anything like this for any other tests, either unit or TDML. And IMO, all our other tests have just as much a likelihood to to enter an infinite loop as these tests (which I argue are both very unlikely). And if they happened, then it would be consistent and easily repeatable.

I do admit that these CLI tests are definitely more complicated with the different thread interactions, but expect ignoring interrupts feels very unlikely. So practically the only place this runCLI will hang is with a infinite loop in the CLI (also feels unlikely, and would likely show up in TDML tests that would hang) or if the test body is broken and leaves the CLI hanging (e.g. the CLI is at a debugger breakpoint waiting for debugger commands, the CLI is reading from stdin and doesn't provide the right input/never closes stdin).

In fact, we had one test that actually did leave the CLI hanging waiting for debug input and these changes made it very obvious and repeatable. The test hung at the cliThread join() call, it needed a Ctrl-C to kill the test, and sbt output the exception that looked like this

```
[error] Test org.apache.daffodil.cliTest.TestCLIdebugger.test_3585_CLI_Debugger_simpleDebugger_unparse failed: java.lang.InterruptedException: null, took 12.068 sec
[error]     at java.lang.Object.wait(Native Method)
[error]     at java.lang.Thread.join(Thread.java:1257)
[error]     at java.lang.Thread.join(Thread.java:1331)
[error]     at org.apache.daffodil.CLI.Util$.runCLI(Util.scala:257)
[error]     at org.apache.daffodil.cliTest.TestCLIdebugger.test_3585_CLI_Debugger_simpleDebugger_unparse(TestCLIDebugger.scala:873)
```

So I knew exactly which test it was and which join it was hanging on. And it turned out the test was just broken and missing a debugger `"quit"` command, fixed in this PR.

We could definitely put a time limit on the CLI joins and start outputting messages (like "CLI loop is taking longer than expected") like you suggest, but those are just going to be false negatives more often than not. Tests taking longer than expected due to resource constrains is what led to randomly failing PRs. Random error messages with tests still passing may just confuse things more. And with a Ctrl-C we get precise information about which test and which join was hung.

stevedlawrence: All that to say, this would be a reasonable approach and I don't think is too messy if we want a warning for the most likely case of a bad test causing the CLI to hang:

```scala
    val cliWarningTimeoutMS = 4 * timeout * 1000
    var isAlive = true
    while (isAlive) {
      isAlive = cliThreadOrProc match {
        case Left(cliThread) => {
          cliThread.join(cliWarningTimeoutMS)
          cliThread.isAlive
        }
        case Right(cliProcess) => {
          cliProcess.waitFor(cliWarningTimeoutMS, TimeUnit.MILLISECONDS)
          cliProcess.isAlive
        }
      }
      if (isAlive) {
        System.err.println("CLI test taking significantly longer than expeced--may be hung waiting on debugger input or closed stdin")
      }
    }
```

So every 4x timeout we output a message if the CLI hasn't completed yet. I'm not sure if 4x the timeout it too long or not long enough, but I really want to avoid false negatives.

I'm still not sure we need this extra complication for the `threadTest.join()` though--the likelihood of that hanging after an interrupt feels very low to me. A bad test causing the CLI to hang does feel much more possible though.

tuxji: Are you saying you could replace line 240,

`while (testThread.isAlive) Thread.sleep(100)`

with this much longer piece of code?  That is, replace Thread.sleep (putting the current thread to sleep until the test thread dies) with a cliThread.join or cliProcess.join (waiting for the CLI thread/process to die)?  These are different threads, so I think you meant another place for your new piece of code?

stevedlawrence: Sorry, that comment got a longer than I expected and I may have made things more confusing.

I suggest changing line 240
```scala
while (testThread.isAlive) Thread.sleep(100)
``` 
With this:
```scala
testThread.join()
```
This is functionally the same thing but a bit cleaner. Neither approach should ever hang in practice because the expect library handles interrupts correctly and will die with the `interrupt()` call right above this line.

For the CLI join/waitFor, which is more likely to hang but mostly requires a broken test, if we want a message I suggest we replace this:
```scala
threadOrProc match {
  case Left(thread) => thread.join()
  case Right(process) => process.waitFor()
}
```

with this:

```scala
val cliWarningTimeoutMS = 4 * timeout * 1000
var isAlive = true
while (isAlive)
  isAlive = cliThreadOrProc match {
    case Left(cliThread) => {
      cliThread.join(cliWarningTimeoutMS)
      cliThread.isAlive
    }
    case Right(cliProcess) => {
      cliProcess.waitFor(cliWarningTimeoutMS, TimeUnit.MILLISECONDS)
      cliProcess.isAlive
    }
  }
  if (isAlive) {
    System.err.println("CLI test taking significantly longer than expected--" +
      "may be hung waiting on debugger input or closed stdin")
  }
}
```

This will still hang like before, but would output a message to stderr every (4 * timeout) seconds that the CLI doesn't end end.

If that's too messy we could probably make some helper functions that do something similar.

tuxji: OK, I'll defer to Mike since that was his request.  I'll say this, however, the phrase "---may be hung waiting on debugger input or closed stdin" is still confusing (at least to me) since it assumes some context the reader should know.  I'm not sure how a bad test can cause "waiting on debugger input or closed stdin"; should we just say "---check whether CLI test is broken" or something like that?

stevedlawrence: The likely case isn't that the CLI is broken, but that a CLI test is broken. For example, we could have a CLI test that does this:


```scala
    runCLI(args"--debug parse ...") { cli =>
      cli.expect("(debug)")
      cli.sendLine("break foo")
      cli.sendLine("continue")
      cli.expect("(debug")
    } (ExitCode.Success)
```

So we start the CLI debugger, set a breakpoint, and continue until that breakpoint is hit. At this point the CLI is paused waiting for input, but the test doesn't provide anymore input. The CLI is now hung because no more input it coming from the test body, but the CLI doesn't know that--it can't differentiate between "no more input is coming" and "we are limited in resources and the input is just coming really really slowly".

So this isn't a bug in the CLI, it's a bug in the test not giving the CLI enough information to finish.

The other example I can think of is if daffodil schema needs to hit EOF to know when to finish parsing/unparsing, but a CLI test doesn't close stdin. So the CLI thinks more data is coming because there's no EOF, but the test body never sends anymore data or closes stdin.

I think these are the most common cases of a broken tests, so I tried to describe them (very tersely) in the error message. But I'm fine with a more generic "broken message" if the message is more confusing than helpful.

tuxji: OK, the reader needs to read that much detailed information to understand the message better.  Since we print that message every 4  * timeout seconds, the message needs to be terse, however.  The message will have to tell the reader the CLI test may be hanging and a place where to read the more detailed explanation you just wrote.  Something like:

```
CLI test may be hanging--see DAFFODIL-2751
```

Then you can add the detailed explanation you wrote to the issue's description so that anyone can understand how a bad CLI test might hang the CI workflow.  The description should say in conclusion, please debug the test to rule out these possible causes rather than "we are limited in resources and the input is just coming really slowly".  If it would be better, you can create a new issue just to serve as the place to read the explanation.

PR #946, opened 2023-02-02
Reduce the maximum length of xs:hexBinary to Int.MaxValue / 2

stevedlawrence:
xs:hexBinary is represented as an Array[Byte], which has a Java implementation maximum length of Int.MaxValue and is what we currently limit hexBinary data to. However, when we output hexBinary bytes to an infoset we convert it to a String, with each byte converted to two characters, effectively doubling the length. Since Java Strings have the same Int.MaxValue length limit, the maximum size of hexBinary that we can actually support is only half of Int.MaxValue. This changes the default and maximum value of the maxHexBinaryLengthInBytes tunable, and thus the length of xs:hexBinary elements, to 1073741823.

Backwards/Compatibility:

The default and maximum value of the maxHexBinaryLengthInBytes tunable, and thus the maximum length of xs:hexBinary elements, is reduced in half from 2147483647 to 1073741823 bytes. Schemas requiring larger values should switch to the Binary Large Object extension or convert the single large hexBinary element to an array of multiple smaller hexBinary elements.

DAFFODIL-2782

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/util/Misc.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/util/Misc.scala
@@ -37,6 +37,7 @@ import java.nio.file.Paths
 import scala.collection.JavaConverters._
 
 import org.apache.daffodil.equality._
+import org.apache.daffodil.exceptions.Assert
 
 /**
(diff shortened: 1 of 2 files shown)


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #948, opened 2023-02-03
Refactor how Main calls "generate c"

tuxji:
The original way Main calls "generate c" requires refactoring in order to generalize to calling "generate \<lang\>" for additional languages.

Main.scala: Refactor object c into a class LanguageConf which takes a language parameter (actually, by name in order to avoid a ClassCastException at runtime).  Make object c extend LanguageConf instead of scallop.Subcommand.  When conf.generate gets called, get conf.generate.\<language\> options from a match expression on conf.generate.subcommand.

DAFFODIL-2786

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
@@ -454,30 +454,40 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
     footer("""|
               |Run 'daffodil generate <language> --help' for subcommand specific options""".stripMargin)
 
-    object c extends scallop.Subcommand("c") {
-      banner("""|Usage: daffodil generate c -s <schema> [-r <root>]
-                |                           [-c <file>] [-T<tunable>=<value>...]
-                |                           [outdir]
-                |
-                |Generate C code from a DFDL schema to parse or unparse data
-                |
-                |Generate Options:""".stripMargin)
+    // Takes language by name so we can pass it to scallop.Subcommand and interpolate it into
+    // strings without getting a runtime java.lang.ClassCastException on Scala 2.12 (class
+    // scala.collection.mutable.WrappedArray$ofRef cannot be cast to class java.lang.String)
+    // @param languageArg Typed on command line & passed to Compiler.forLanguage
+    // @param languageName Formal descriptive name of language for help/usage messages
+    class LanguageConf(languageArg: => String, val languageName: String)
+      extends scallop.Subcommand(languageArg) {
 
-      descr("Generate C code from a DFDL schema")
+      val language = languageArg
+      banner(s"""|Usage: daffodil generate $language -s <schema> [-r <root>]
+                 |                              [-c <file>] [-T<tunable>=<value>...]
+                 |                              [outdir]
+                 |
+                 |Generate $languageName code from a DFDL schema to parse or unparse data
+                 |
+                 |Generate Options:""".stripMargin)
+      descr(s"Generate $languageName code from a DFDL schema")
       helpWidth(width)
 
-      val language = "c"
       val config = opt[File](short = 'c', argName = "file", descr = "XML file containing configuration items")
       val rootNS = opt[RefQName]("root", argName = "node", descr = "Root element to use. Can be prefixed with {namespace}. Must be a top-level element. Defaults to first top-level element of DFDL schema.")
       val schema = opt[URI]("schema", required = true, argName = "file", descr = "DFDL schema to use to create parser")(fileResourceURIConverter)
       val tunables = props[String](name = 'T', keyName = "tunable", valueName = "value", descr = "Tunable configuration options to change Daffodil's behavior")
 
-      val outdir = trailArg[String](required = false, descr = "Output directory in which to create 'c' subdirectory. If not specified, uses current directory.")
+      val outdir = trailArg[String](required = false, descr = s"Output directory in which to create '$language' subdirectory. If not specified, uses current directory.")
 
       requireOne(schema) // --schema must be provided
       validateFileIsFile(config) // --config must be a file that exists
     }
+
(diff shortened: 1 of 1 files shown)


daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -454,30 +454,37 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
...
+    // into strings without getting a puzzling java.lang.ClassCastException at runtime (class
+    // scala.collection.mutable.WrappedArray$ofRef cannot be cast to class java.lang.String)
+    class LanguageConf(languageArg: => String) extends scallop.Subcommand(languageArg) {
+      val language = languageArg
+
+      banner(s"""|Usage: daffodil generate $language -s <schema> [-r <root>]
+                 |                              [-c <file>] [-T<tunable>=<value>...]
+                 |                              [outdir]
+                 |
+                 |Generate ${language.toUpperCase} code from a DFDL schema to parse or unparse data
```
stevedlawrence, line 467: Thoughts on making `val language = languageArg.toUpperCase` and then just using `language` in the banner/options?

That way if we ever support languages where the language name isn't as simple as the upper cased command then they can override it, e.g.:

```scala
object rust extends LanguageConf("rust") {
  override val language = "Rust"
}

object rust extends LanguageConf("cpp") {
  override val language = "C++"
} 

object rust extends LanguageConf("csharp") {
  override val language = "C#"
}
```

stevedlawrence (approved):
+1, minor suggestion for future extensibility.

Does this mean a new support language is nearing?

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
tuxji: I like your suggestion for better extensibility!  I still need to show the original lowercase language name in "daffodil generate c" usage messages but I'll use languageArg for that purpose.

I can't put a timeline on when a new supported language might come out, but I hope it will come eventually.

daffodil-cli/src/main/scala/org/apache/daffodil/Main.scala
```diff
@@ -454,30 +454,40 @@ class CLIConf(arguments: Array[String]) extends scallop.ScallopConf(arguments) {
...
-                |
-                |Generate C code from a DFDL schema to parse or unparse data
-                |
-                |Generate Options:""".stripMargin)
-
-      descr("Generate C code from a DFDL schema")
+    // Takes language by name so we can pass it to scallop.Subcommand and interpolate it
+    // into strings without getting a puzzling java.lang.ClassCastException at runtime (class
+    // scala.collection.mutable.WrappedArray$ofRef cannot be cast to class java.lang.String)
+    class LanguageConf(languageArg: => String) extends scallop.Subcommand(languageArg) {
```
stevedlawrence, line 460: I just tested this with

```scala
    object rust extends LanguageConf("rust") {
      override val languageName = "Rust"
    }
```

But `daffodil generate rust --help` outputs this:

> Generate null code from a DFDL schema to parse or unparse data

I wonder if this needs to use an early initializer? Or maybe it needs to be def or lazy val?

tuxji: Great catch, Steve.  I looked at pre-initialized values but no thanks, that syntax will puzzle new Scala programmers and they would need examples in order to use that syntax.  I tried both def and lazy val and I found out that lazy val works better in practice.  You need to use def in both places for def to work.  If you use val in the object rust, the Scala compiler won't warn you, but the help message will print "null" again.  However, the Scala compiler will refuse to compile Main if you don't use lazy val in both places; it will tell you `value languageName must be declared lazy to override a concrete lazy value` if you use val and `method languageName needs to be a stable, immutable value` if you use def.  Lazy val it is.

mbeckerle: Unless scallop.Subcommand class also takes a by-name argument there's no point in taking a by-name argument in your LanguageConf class. 

This may be part of the issue. At the time the superclass is constructed it may not have a value for languageArg yet.

tuxji: > Unless scallop.Subcommand class also takes a by-name argument there's no point in taking a by-name argument in your LanguageConf class. This may be part of the issue. At the time the superclass is constructed it may not have a value for languageArg yet.

Er, the whole point of taking the by-name argument is to avoid a runtime exception which invariably occurs if you pass a lang: String argument and use it in both scallop.Subcommand(lang) and val language = lang.  Note scallop.Subcommand doesn't take a by-name argument, but it takes a String* argument which causes lang to somehow become wrapped in a scala.collection.mutable.WrappedArray$ofRef object at runtime in BOTH places.  Formerly I was passing a literal "c" in both scallop.Subcommand("c") and val language = "c" before, so that's why this exception didn't happen before.  The exception also goes away if LanguageConf takes two redundant lang1: String, lang2: String arguments and passes either lang1 or lang2 to scallop.Subcommand(lang1) and the other one of lang1 or lang2 to val language = lang2.  If I add an Assert.invariant(lang1 == lang2) line to double check that both lang1 and lang2 were passed the same value ("c"), however, that line will trigger the exception right there.  Passing the string argument by-name still allows the string to be passed to scallop.Subcommand(languageArg) but prevents the string from being wrapped when it's used in val language = languageArg.  This code is as good as we can get until we upgrade to a new Scala dialect which may stop wrapping the lang: String argument at the wrong place or the wrong time.

tuxji:
On second thought, I'm now passing languageName as the second constructor parameter.  Then new language options won't need to override languageName at all.

```scala
object rust extends LanguageConf("rust", "Rust")
object cpp extends LanguageConf("cpp", "C++")
object csharp extends LanguageConf("csharp", "C#")
```

Mike, a +1 please.

mbeckerle (approved):
+1

PR #949, opened 2023-02-06
Fix overflow in DataLoc positions

stevedlawrence:
Multiple values in the DataLoc improperly convert from long to int, which could lead to overflows when bit positions grow to larger than Int.MaxValue. This leads to assertions on DataLoc calculations that are expeced to only be positive. This fixes the calculations so we keep bit positions as a long type to avoid these overflows. Also removes some unnecessary math.max calls since we know certain calculations will always be positive.

DAFFODIL-2785

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataLoc.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataLoc.scala
@@ -49,24 +49,30 @@ class DataLoc(
   Assert.usage(bitLimit1b.isEmpty || bitLimit1b.get >= 0)
   Assert.usage(bitPos1b >= 1)
 
-  val bitPos0b = math.max(bitPos1b - 1, 0).toInt
-  val bitLimit0b = if (bitLimit1b.isDefined) MaybeULong(bitLimit1b.get - 1) else MaybeULong.Nope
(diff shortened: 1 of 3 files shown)


tuxji (commented):
The changes look good.

However, there's no unit test to ensure that the error remains fixed despite future changes.  A unit test could construct a DataLoc with a location larger than Int.MaxValue and verify that the DataLoc calculations no longer overflow and throw exceptions.  Might such a test be worth the extra time?

stevedlawrence:
> Might such a test be worth the extra time

Test added. Also rebased on main to fix conflicts.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataLoc.scala
```diff
@@ -49,9 +49,9 @@ class DataLoc(
   Assert.usage(bitLimit1b.isEmpty || bitLimit1b.get >= 0)
   Assert.usage(bitPos1b >= 1)
 
-  val bitPos0b = math.max(bitPos1b - 1, 0).toInt
+  val bitPos0b = bitPos1b - 1
```
mbeckerle, line 52: I think it is worth it to add the type declaration of Long to these members. Especially public ones.

mbeckerle (approved):
+1. I just suggest adding the type declarations of Long to the public members like bitPos0b.

PR #950, opened 2023-02-07
Remove timeout from socket test

stevedlawrence:
This timeout could lead to false negatives on continuous integration tests with limited resources. This test should always finish unless a bug is introduced to Daffodil, so just remove the timeout. This means if this test ever does fail due to a real bug, we might hang forever, but that is unlikely and it's better than getting false negatives and trying to determine if spurious or not.

DAFFODIL-2751


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #953, opened 2023-02-07
Pass through dfdlx:trace argument unchanged

mike-mcgann:
An assertion with the test expression `{dfdlx:trace(42, 'answer')}` was converting the first argument to a boolean and then also attemping to convert it again when used as a return value. This update passes through the argument without converting its type to boolean.

[DAFFODIL-2628](https://issues.apache.org/jira/browse/DAFFODIL-2628)


tuxji (approved):
+1

PR #959, opened 2023-02-10
Manually extract runtime2 files out of the jar

stevedlawrence:
CI tests randomly fail because the jar filesystem used to extract runtime2 files sometimes already exists. We have a mutex that should prevent that, but that does not seem to be sufficient. The reason isn't clear.

To avoid this problem, this replaces the jar filesystem with logic to open the jar, inspect the jar entries, and extract the files that we need. It is a bit more verbose, but does not have any resource synchronization issues.

DAFFODIL-2751

### DIFF
--- a/daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/CodeGenerator.scala
+++ b/daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/CodeGenerator.scala
@@ -18,24 +18,20 @@
 package org.apache.daffodil.runtime2
 
 import java.io.File
-import java.nio.file.FileSystems
+import java.net.JarURLConnection
 import java.nio.file.Files
 import java.nio.file.Paths
-import java.util.Collections
-import org.apache.daffodil.runtime1.api.DFDL
-import org.apache.daffodil.lib.api.Diagnostic
+import scala.collection.JavaConverters._
+
 import org.apache.daffodil.core.compiler.Compiler
 import org.apache.daffodil.core.dsom.Root
+import org.apache.daffodil.lib.api.Diagnostic
+import org.apache.daffodil.lib.util.Misc
+import org.apache.daffodil.runtime1.api.DFDL
 import org.apache.daffodil.runtime1.dsom.SchemaDefinitionError
 import org.apache.daffodil.runtime2.generators.CodeGeneratorState
-import org.apache.daffodil.lib.util.Misc
 
 import scala.util.Properties.isWin
-
-/**
- * We need a mutux object for exclusive access to a code block
- */
-private object mutex {}
 
 /**
  * Generates and compiles C source files from a DFDL schema encapsulated in the parameter.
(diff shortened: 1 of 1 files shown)


daffodil-runtime2/src/main/scala/org/apache/daffodil/runtime2/CodeGenerator.scala
```diff
@@ -66,21 +62,23 @@ class CodeGenerator(root: Root) extends DFDL.CodeGenerator {
...
+      val jarFile = jarConnection.getJarFile()
+      jarFile.entries.asScala
+        .filter { entry => ("/" + entry.getName).startsWith(resources) }
+        .filterNot { entry => entry.isDirectory }
+        .foreach { entry =>
+          val entryPath = "/" + entry.getName
+          val subPath = os.SubPath(entryPath.stripPrefix(resources + "/"))
+          val dstPath = codeDir / subPath
+          val stream = jarFile.getInputStream(entry)
+          os.write(dstPath, stream, createFolders = true)
```
tuxji, line 77: The advantage of keeping using `os.copy` is that it will copy the original file's attributes (date/time and permissions).  I'd like to suggest an alternative fix which might work as well:

```scala
    // Copy all our C resources to the code subdirectory (using synchronized
    // to avoid calling FileSystems.newFileSystem concurrently)
    val resources = "/org/apache/daffodil/runtime2/c"
    val resourceUri = Misc.getRequiredResource(resources)
    if (resourceUri.getScheme == "jar") {
      val env: java.util.Map[String, String] = Collections.emptyMap()
      val fileSystem = Misc.synchronized { FileSystems.newFileSystem(resourceUri, env) }
      try {
        val resourceDir = os.Path(fileSystem.getPath(resources))
        os.copy(resourceDir, codeDir)
      }
      finally fileSystem.close()
    } else {
      val resourceDir = os.Path(Paths.get(resourceUri))
      os.copy(resourceDir, codeDir)
    }
```

What do you think?  I've been using that code in another code generator even before this PR was made for a while and it seems to work.  I've also made that improvement to the C code generator in another PR I've been working on to remove Runtime2DataProcessor and merge its code into Runtime2TDMLDFDLProcessor, BTW.

stevedlawrence: I don't think this solves the problem we are seeing with CI though, where `newFileSystem` throws a `FileSystemAlreadyExistsException`. Here's a recent example:

https://github.com/apache/daffodil/actions/runs/4144366800/jobs/7167330500

Things are synchronized (and everything is currently synchronized vs your suggestion which just synchronizes the `newFileSystem` call), so I wouldn't think this wouldn't be a problem, but this has happened multiple times in GitHub actions. I couldn't reproduce it or figure out the cause though. My best guess is maybe `newFileSystem` is getting called in separate processes where `synchronized` doesn't have an effect, and maybe you can't have the same `newFileSystem` even in separate processes? I don't think our tests run the code generate in separate processes with the recent CLI changes, so that doesn't feel likely, but I'm not sure what else would cause it.

> The advantage of keeping using os.copy is that it will copy the original file's attributes (date/time and permissions).

Can jar files (i.e. zip files) even store permissions? And if it can, do we even want those same permissions? As a user I would expect all generated C files, whether they are copied out of a jar or written by the code generator, to have the same permissions. For example, it would feel odd if I had permissions to delete the generated code files but not the files copied from a jar. Seems files should all be generated with whatever user is calling the generator and with whatever umask is set.

tuxji: I just ran some commands and found out zip files store permissions, but jar files strip permissions.  Both zip and jar files do store dates/times and restore them after unpacking.  I'm not using permissions at all and copying dates/times is only a nice to have, not a requirement.  

I had noticed the problem with CI myself earlier as well.  I had a hypothesis that the previous critical section was too large and shrinking it to just the `newFileSystem` call might work better.  I can't promise that my hypothesis might be true but do you want to try it?  If not, I'll merge your change into my PR branch.

stevedlawrence:
>  I can't promise that my hypothesis might be true but do you want to try it.

Sure thing, I'll make the change and trigger a handful of CI runs and see if it comes back.

stevedlawrence:
I was able to trigger the failure with the new code:

https://github.com/apache/daffodil/actions/runs/4166436854/jobs/7211548107#step:15:233

So your suggested change with a different synchronization block didn't fix it.

Without an idea of what the the cause is, I suggest we switch to the original approach that avoids `newFileSystem`. It's probably not too hard to extract the modification time from each JarEntry and apply it to the file after writing it if that's preferred.

tuxji:
> I was able to trigger the failure with the new code:

OK, let's switch back to your solution.  The date/time is only a nice to have; you don't have to bother implementing it.  That's the kind of thing more easily done by calling a library method, assuming someone has already written a library method to copy a folder from a jar file to a folder on disk.  I looked quickly at Apache FileUtils but didn't see one.

tuxji (approved):
+1 to original commit (drop the second commit)

stevedlawrence:
I've removed the new commit, reverting back to the original PR

stevedlawrence:
Closing, this change has been pulled into #961

PR #960, opened 2023-02-13
Modify GitHub action concurrency to not cancel main branch CI runs

stevedlawrence:
If a CI run isn't a pull request (i.e. github.head_ref isn't defined), then we currently fall back to github.ref_name. But this value isn't unique (e.g. it is just "main" for merges into the main branch), so it means merges into the main branch are in the same concurrency group which can lead to multiple main CI runs to be canceled and appear as failing.

This instead uses the github.run_id as the fallback value, which is unique for non-PR's so they will be canceled. This also removes the github.actor since we don't really need to differentiate between different users making updates to the same PR. We only ever want to run an action for the latest PR commit and cancel all previous ones, regardless of the actor.

DAFFODIL-2789

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,10 +24,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
-# Cancel CI runs in progress when a branch or pull request is updated.
-
+# Cancel CI runs in progress when a pull request is updated.
 concurrency:
-  group: ${{ github.actor }}-${{ github.head_ref || github.ref_name }}-${{ github.workflow }}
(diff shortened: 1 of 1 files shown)


tuxji (approved):
+1

I found out you don't need to quote most string literals in YAML files although there are some [exceptions](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml).  The YAML specification allows single or double quotes, but GitHub itself [allows only single quotes](https://docs.github.com/en/actions/learn-github-actions/expressions#literals) inside `${{ }}` expressions:

> `string` - You don't need to enclose strings in ${{ and }}. However, if you do, you must use single quotes (') around the string. To use a literal single quote, escape the literal single quote using an additional single quote (''). Wrapping with double quotes (") will throw an error.

jadams-tresys (approved):
+1

PR #964, opened 2023-02-14
Add test showing incorrect value of dfdl:occursIndex()

stevedlawrence:
When emptyElementParsePolicy="treatAsMissing", the dfdl:occursIndex() function returns the iteration count of an array instead of the actual count of the array. Add a test showing this incorrect behavior.

DAFFODIL-2791


tuxji (approved):
+1

jadams-tresys (approved):

PR #965, opened 2023-02-15
Format test and sbt files

mike-mcgann:
This change applies scalafmt to test and sbt files, adds additional checks to the CI workflow, and updates DEVELOP.md.

[DAFFODIL-2133](https://issues.apache.org/jira/browse/DAFFODIL-2133)


daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestBlob.scala
```diff
@@ -110,19 +111,23 @@ class TestBlob {
...
-          } (ExitCode.Success)
+          runCLI(args"parse -s $schema -o $infoset $input", timeout = 120) { cli => }(
+            ExitCode.Success,
+          )
 
-          runCLI(args"unparse -s $schema -o $unparse $infoset", timeout = 120) { cli =>
-          } (ExitCode.Success)
+          runCLI(args"unparse -s $schema -o $unparse $infoset", timeout = 120) { cli => }(
+            ExitCode.Success,
+          )
```
stevedlawrence, line 130: This particular changes makes this much less readable. For consistency, we really want all these tests to have the same pattern of

```scala
runCLI(args"...") { cli =>
  // zero or more lines of actual checks
}(ExitCode.Foo)
```

It seems the ones with bad formatting are those without actual CLI expect checks, likely because all the test cares about is the exit code. I don't think there are too many of those--maybe we just add a comment that makes it clear no expect logic is intentional, and it might also prevent scalafmt from changing these lines? For example:

```scala
runCLI(args"...") { cli =>
  // this block intentionally left blank
}(ExitCode.Success)
```

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/schematron/TestEmbedded.scala
```diff
@@ -40,99 +41,111 @@ class TestEmbedded {
...
 
   @Test def testWithNs2(): Unit = {
     val schema = path("daffodil-schematron/src/test/resources/xsd/with-ns-1.dfdl.xsd")
 
-    runCLI(args"""parse --validate schematron="${jsonEscape(schema.toString)}" -s $schema""") { cli =>
-      cli.send("2;1", inputDone = true)
-      cli.expectErr("Validation Error")
-    } (ExitCode.ParseError)
+    runCLI(args"""parse --validate schematron="${jsonEscape(schema.toString)}" -s $schema""") {
+      cli =>
```
stevedlawrence, line 136: I think this is another example about this being very aggressive about line lengths which ultimately hurts readability.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/schematron/TestSvrlOutput.scala
```diff
@@ -17,55 +17,64 @@
...
-              case <svrl:failed-assert>{  _* }</svrl:failed-assert> => true
-              case _ => false
-            }
-            // we should not have found failures
-            assertFalse(res.isDefined)
-          case _ =>
-            fail("schematron pattern didnt match")
+    withTempFile(
+      ".conf",
+      { conf =>
```
stevedlawrence, line 56: I don't even understand why it made this change.

project/Rat.scala
```diff
@@ -101,12 +112,24 @@ object Rat {
...
+    ),
+    file(
+      "daffodil-test/src/test/resources/org/apache/daffodil/runtime2/MPU_green_to_orange_60004.dat",
+    ),
+    file(
+      "daffodil-test/src/test/resources/org/apache/daffodil/runtime2/MPU_orange_to_green_60006.0.dat",
+    ),
+    file(
+      "daffodil-test/src/test/resources/org/apache/daffodil/runtime2/MPU_orange_to_green_60006.1.dat",
+    ),
```
stevedlawrence, line 132: Another example where long lines really should be allowed. This changed 1 line that's over the line length limit, to two short lines and a third line that's still over the limit, just a bit shorter.

stevedlawrence (approved):
+1

Feels like a number of these change hurt readability due to slightly long lines being preferred, but I'm not sure there's much we can do about it.

DEVELOP.md
```diff
@@ -138,11 +138,10 @@ run `git push` from your reading copy.
 
 ### Code Style
 
-Daffodil mandates standard Scala formatting and its code generally
-keeps close to that formatting consistently.  No one has run an
-automatic Scala formatter on the codebase yet (there has not been much
-need since the code is already formatted pretty well) but
-[DAFFODIL-2133][] is waiting for someone willing to do that work.
+Daffodil code is formatted using [Scalafmt](https://scalameta.org/scalafmt/).
+Run `sbt scalafmtCheckAll` to see if changes to the source code reqiure
```
tuxji, line 142: *require

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestBlob.scala
```diff
@@ -110,19 +111,23 @@ class TestBlob {
    *
    ***/
   @Test def test_2GB_blob(): Unit = {
-    val schema = path("daffodil-cli/src/test/resources/org/apache/daffodil/cli/large_blob.dfdl.xsd")
+    val schema = path(
+      "daffodil-cli/src/test/resources/org/apache/daffodil/cli/large_blob.dfdl.xsd",
+    )
```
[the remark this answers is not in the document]
tuxji: We have two options:

1. Increase maxColumns (which means a lot of files get reformatting changes again).
2. Shorten our file paths ("daf/cli/src/test/resources/org/apache/daffodil/cli/large_blob.dfdl.xsd" or go even further and remove the org/apache/daffodil entirely, since test files can use shorter paths).

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestBlob.scala
tuxji: Yes, a well-placed comment like "// checks only exit code" in those places probably will preserve the desired formatting.

daffodil-cli/src/it/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
```diff
@@ -179,105 +225,159 @@ class TestCLIUdfs {
...
       cli.expectErr("[error] Schema Definition Error: Unsupported function: jsudf:replace")
-    } (ExitCode.UnableToCreateProcessor)
+    }(ExitCode.UnableToCreateProcessor)
   }
 
   /**
    * Tests the case when a function class:
    *   throws custom error on evaluate
    */
   @Test def test_UDFClass_CustomExceptionOnEvaluate(): Unit = {
```
tuxji, line 274: This is one of the two IT tests which are failing in CI.  This one fails with `net.sf.expectit.ExpectIOException: Expect operation fails (timeout: -1 ms) for matcher: contains('at org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse.evaluate(StringFunctionsProvider.scala:56)')`.  I've looked at the formatting changes and I don't think they've changed the string arguments, only the whitespace around them.  You'll have to debug the test to see what the text being matched against really contains.  I'm pretty sure the line number changed because the other test file got reformatted (the StringFunctions.Reverse file).  Simply take out the line number from the string argument at line 288 below.

daffodil-cli/src/it/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
```diff
@@ -179,105 +225,159 @@ class TestCLIUdfs {
...
-      cli.expectErr("at org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords.<init>(StringFunctionsProvider.scala:65)")
-    } (ExitCode.UserDefinedFunctionError)
+      cli.expectErr(
+        "[error] User Defined Function could not be initialized: {http://example.com/scala/udf}rev-words.",
+      )
+      cli.expectErr(
+        "Cause: org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords$CustomException: UDF Error!",
+      )
+      cli.expectErr(
+        "at org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords.<init>(StringFunctionsProvider.scala:65)",
```
tuxji, line 332: Likewise, remove the line number (or more) from this string argument to avoid `Expect operation fails (timeout: -1 ms) for matcher: contains('at org.sbadudfs.udfexceptions2.StringFunctions.ReverseWords.<init>(StringFunctionsProvider.scala:65)'`.  Being too verbose makes the test more brittle (what if the path changes due to renaming or shortening changes as well?).

daffodil-cli/src/it/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
```diff
@@ -179,105 +225,159 @@ class TestCLIUdfs {
...
     runCLI(args"parse -s $schema -r user_func2", classpath) { cli =>
       cli.send("strng", inputDone = true)
-      cli.expectErr("[error] User Defined Function 'ssudf:reverse' Error. Cause: org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse$CustomException: UDF Error!")
-      cli.expectErr("at org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse.evaluate(StringFunctionsProvider.scala:56)")
-    } (ExitCode.UserDefinedFunctionError)
+      cli.expectErr(
+        "[error] User Defined Function 'ssudf:reverse' Error. Cause: org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse$CustomException: UDF Error!",
+      )
+      cli.expectErr(
+        "at org.sbadudfs.udfexceptions.evaluating.StringFunctions.Reverse.evaluate(StringFunctionsProvider.scala:56)",
```
stevedlawrence, line 288: Looks like this test is failing due to line numbers change from formatting. We should remove the line number from this tests. Our tests don't need to be that specific.

tuxji:
> Feels like a number of these change hurt readability due to slightly long lines being preferred, but I'm not sure there's much we can do about it.

We are not that helpless; the long lines due to long test file paths are under our control.  I will wait for @mbeckerle's opinion whether we should make further changes like that or live with the formatting changes.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestBlob.scala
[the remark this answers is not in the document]
mike-mcgann: Or maybe define prefixes in a constant and reference that later:

```scala
val cliTestResources = "daffodil-cli/src/test/resources/org/apache/daffodil/cli"

val schema = path(s"{cliTestResources}/large_blob.dfdl.xsd"`)
```

DEVELOP.md
mike-mcgann: Thanks. Updated.

daffodil-cli/src/it/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
mike-mcgann: Line numbers have been removed from the tests

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestBlob.scala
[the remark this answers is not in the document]
stevedlawrence: One reason I like the full path is it's easy to copy/paste when a test fails and I need to manually run a test. But I guess that happens rarely enough, and it's not much extra effort to manually replace the variable with the common part.

This seems like a reasonable approach. I guess it's also good that it removes more boilerplate, and makes the actual test file (i.e. the important thing) stand out a bit more.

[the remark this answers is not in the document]
tuxji: Using a constant to shorten paths is reasonable as well although I've used copy/paste on test file paths to go look at a test file like Steve too.  Everything we do with test files will become easier to do if we actually remove their parent directories as much as possible, though.  Many tests shouldn't need more than 1 or 2 parent directories, IMO.

Whatever we decide to do, let's do it in another PR.  I want to merge this PR before I work on my next runtime2 PR (daffodil-codegen-c and DaffodilC instead of Runtime2).  How long until we can hear from @mbeckerle too?

tuxji (approved):
+1

Please squash to 1 commit and then I'll go ahead and merge.

PR #966, opened 2023-02-15
Better error message when contentLength/valueLength argument is an array

mike-mcgann:
The `dfdl:contentLength` and `dfdl:valueLength` functions accept a `NodeType.Exists` type as the first argument to require the value to be a node and to prevent the use of a literal. When passed an array, which is prohibited, the error message is misleading because it refers to the parent of the array and not the array itself. This change introduces an explicit check for an array and raises an SDE with a better error message. 

[DAFFODIL-1651](https://issues.apache.org/jira/browse/DAFFODIL-1651)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
@@ -745,6 +745,8 @@ abstract class PathExpression() extends Expression {
    */
   lazy val isPathToOneWholeArray: Boolean = {
     if (steps == Nil) false // root is never an array
+    if (steps == Nil) false // root is never an array
+    else if (isSelf) false // self path is never an array
     else
       steps.last.isArray &&
       !steps.last.pred.isDefined && // last cannot have a [N] pred
@@ -757,6 +759,11 @@ abstract class PathExpression() extends Expression {
           else if (step.isArray) step.pred.isDefined
           else true
       }
+  }
+
+  def isSelf = steps match {
+    case List(Self(_)) => true
+    case _ => false
   }
 }
 
@@ -2829,13 +2836,21 @@ sealed abstract class LengthExprBase(
   ) {
 
   protected final def leafReferencedElements = {
-    val arg = args(0).asInstanceOf[PathExpression]
-    val steps = arg.steps
+    val path = args(0) match {
+      case arg: PathExpression => arg
+      case _ => {
+        // $COVERAGE-OFF$
+        Assert.invariantFailed("NodeInfo.Exists, but arg was not a PathExpression.")
+        // $COVERAGE-ON$
+      }
+    }
+    if (path.isPathToOneWholeArray)
+      SDE(s"First argument to ${fnQName} cannot be a path to an array")
+    val steps = path.steps
     val lst = steps.last
     val res = lst.stepElements.toSet
     res
   }
-
 }
 
 case class ContentLengthExpr(
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
@@ -387,7 +387,7 @@ object NodeInfo extends Enum {
   }
 
   /**
-   * For things like fn:exists, fn:empty, dfdl:contenLength
+   * For things like fn:exists, fn:empty, dfdl:contentLength
    */
   protected sealed trait ExistsKind extends AnyType.Kind
   case object Exists extends TypeNode('Exists, AnyType) with AnyTypeKind {
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
@@ -459,6 +459,37 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="valueLength14">
+      <xs:complexType>
+        <xs:sequence dfdl:separator="%WSP;">
+          <xs:element name="first" type="xs:string" dfdl:lengthKind="delimited"/>
+          <xs:element name="other" type="xs:string" dfdl:lengthKind="delimited" maxOccurs="unbounded"/>
+          <xs:element name="len" type="xs:unsignedInt" dfdl:inputValueCalc='{dfdl:valueLength(../ex:other, "bytes")}'/>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="valueLength15" type="xs:unsignedInt" dfdl:inputValueCalc='{dfdl:valueLength("three", "bytes")}'/>
+
+    <xs:element name="valueLength16">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="e1" maxOccurs="unbounded">
+            <xs:annotation>
+              <xs:appinfo source="http://www.ogf.org/dfdl/">
+                <dfdl:assert>{ dfdl:valueLength(., 'bytes') eq 3 }</dfdl:assert>
+              </xs:appinfo>
+            </xs:annotation>
+            <xs:complexType>
+              <xs:sequence dfdl:separator=",">
+                <xs:element name="v1" type="xs:string"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
     <xs:element name="contentLength">
       <xs:complexType>
         <xs:sequence dfdl:separator=",">
@@ -494,7 +525,38 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
-    
+
+    <xs:element name="contentLength4">
+      <xs:complexType>
+        <xs:sequence dfdl:separator="%WSP;">
+          <xs:element name="first" type="xs:string" dfdl:lengthKind="delimited"/>
+          <xs:element name="other" type="xs:string" dfdl:lengthKind="delimited" maxOccurs="unbounded"/>
+          <xs:element name="len" type="xs:unsignedInt" dfdl:inputValueCalc='{dfdl:contentLength(../ex:other, "bytes")}'/>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="contentLength5" type="xs:unsignedInt" dfdl:inputValueCalc='{dfdl:contentLength("three", "bytes")}'/>
+
+    <xs:element name="contentLength6">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="e1" maxOccurs="unbounded">
+            <xs:annotation>
+              <xs:appinfo source="http://www.ogf.org/dfdl/">
+                <dfdl:assert message="{ dfdl:contentLength(., 'bytes') }">{ dfdl:contentLength(., 'bytes') eq 3 }</dfdl:assert>
+              </xs:appinfo>
+            </xs:annotation>
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="v1" type="xs:string"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
     <xs:element name="e_concat">
       <xs:complexType>
         <xs:sequence dfdl:separator=",">
@@ -959,8 +1021,6 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
-
-
   </tdml:defineSchema>
 
 <!--
@@ -3898,6 +3958,77 @@
   </tdml:parserTestCase>
 
 <!--
+    Test Name: valueLength_14
+       Schema: Functions.dfdl.xsd
+         Root: valueLength14
+      Purpose: This test demonstrates that it is an error to get the
+               valueLength of an array
+-->
+  <tdml:parserTestCase name="valueLength_14" root="valueLength14"
+                       model="Functions.dfdl.xsd" description="Section 23 " roundTrip="false">
+
+    <tdml:document>
+      <tdml:documentPart type="text">one two three</tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>cannot be a path to an array</tdml:error>
+      <tdml:error>dfdl:valueLength</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+<!--
+    Test Name: valueLength_15
+       Schema: Functions.dfdl.xsd
+         Root: valueLength15
+      Purpose: This test demonstrates that it is an error to get the
+               valueLength of a literal
+-->
+  <tdml:parserTestCase name="valueLength_15" root="valueLength15"
+                       model="Functions.dfdl.xsd" description="Section 23 " roundTrip="false">
+
+    <tdml:document>
+      <tdml:documentPart type="text">one two three</tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>expected a path expression</tdml:error>
+      <tdml:error>xs:string</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+<!--
+    Test Name: valueLength_16
+       Schema: Functions.dfdl.xsd
+         Root: valueLength16
+      Purpose: This test demonstrates a valueLength on a complexType using
+               a path to self
+-->
+
+  <tdml:parserTestCase name="valueLength_16" root="valueLength16"
+                       model="Functions.dfdl.xsd" description="Section 23 " roundTrip="false">
+
+    <tdml:document>
+      <tdml:documentPart type="text">ABC,DEF,GHI</tdml:documentPart>
+    </tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <valueLength16>
+          <e1>
+            <v1>ABC</v1>
+          </e1>
+          <e1>
+            <v1>DEF</v1>
+          </e1>
+          <e1>
+            <v1>GHI</v1>
+          </e1>
+        </valueLength16>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+<!--
     Test Name: valueLength_unparse_0
        Schema: Functions.dfdl.xsd
          Root: valueLength5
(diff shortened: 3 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2113,7 +2113,7 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           functionQName,
           args,
           NodeInfo.Long,
-          NodeInfo.Exists,
+          NodeInfo.Scalar,
           NodeInfo.String,
           DFDLContentLength(_),
         )
```
tuxji, line 2119: Please find the comment on the next line (not shown in the diff) and update it so it no longer refers to `Exists` but refers to `Scalar` instead.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2829,15 @@ sealed abstract class LengthExprBase(
   ) {
 
   protected final def leafReferencedElements = {
+    if (!args(0).isInstanceOf[PathExpression]) {
+      SDE(s"expected a path expression but got type ${args(0).inherentType.globalQName}")
+    }
     val arg = args(0).asInstanceOf[PathExpression]
```
tuxji, line 2835: > I added an additional test to ensure there is an error when a string literal is used as the first argument. That causes a ClassCastException. I wasn't sure how to restrict that from happening using the type system and I just hacked in a type check to instead throw an SDE. I need some suggestions on how to implement this fix properly.

Is this the place you mentioned where you want a better suggestion how to avoid the ClassCastException from the asInstanceOf[PathExpression] on line 2835?  Your type check looks fine to me but any thoughts, Steve?  The only suggestion I have is to call schemaDefinitionUnless instead which collapses the if and SDE into a single function call.  This file already has 27 SDE calls and 9 schemaDefinitionUnless calls, so your type check will be in good company.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -382,18 +380,28 @@ object NodeInfo extends Enum {
...
 
+  /**
+   * For things like dfdl:contentLength, dfdl:valueLength
+   */
+  protected sealed trait ScalarKind extends AnyType.Kind
+  case object Scalar
+    extends TypeNode('Scalar, AnyType, Seq(AnySimpleType, Complex))
+    with AnyTypeKind {
+    type Kind = ScalarKind
+  }
```
tuxji, line 403: I'm not familiar with the TypeNode constructor, so Steve please take a look too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -993,6 +1001,7 @@ object NodeInfo extends Enum {
   lazy val allTypes =
     allDFDLTypes ++ List(
       Complex,
+      Scalar,
       Array,
```
tuxji, line 1005: I notice Exists is listed in allAbstractTypes (above at line 969) while Scalar is listed in allTypes here.  I'm sure Scalar is more analogous to Complex than Exists, but I thought I should mention it.

tuxji (approved):
+1

Your changes look good to me after you update a comment, but let's wait for Steve to review too.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2131,7 +2131,7 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           functionQName,
           args,
           NodeInfo.Long,
-          NodeInfo.Exists,
+          NodeInfo.Scalar,
```
stevedlawrence, line 2134: Right above this is a big comment about when "exists" makes sense. That needs to be removed/changed.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2829,15 @@ sealed abstract class LengthExprBase(
   ) {
 
   protected final def leafReferencedElements = {
+    if (!args(0).isInstanceOf[PathExpression]) {
+      SDE(s"expected a path expression but got type ${args(0).inherentType.globalQName}")
```
stevedlawrence, line 2833: Suggest capitalizing "e" in expected..

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
stevedlawrence: I believe this is correct. TypeNode basically just sets up a type hierarchy. AnyType is the parent of Scalar, and it's children are simple types and complex types.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
stevedlawrence: Good point. Scalar's children are Complex (here in `allTypes`) and AnySimple (above in allAbstractTypes), so there doesn't seem to be an obvious place for. I don't think it currently matters because nothing actually uses `allAbstractTypes`, but it feels like Scalar is more like Exists/AnySimpleType than Complex and does want to be in the abstract lis.

stevedlawrence (approved):
+1 :+1:

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
mike-mcgann: Moved to the abstract list

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mike-mcgann: Updated

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mike-mcgann: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
mbeckerle: Scalar is the opposite of Array, so I think this should be like Array, no children, no parent. 

But really we only need Exists.

I'm pretty sure we could get rid of the NodeInfo.Array easily, as it is barely used anywhere. I think it was my mistake to put Array in this taxonomy originally. I don't think it is actually used. 

The things I've seen that care if things are arrays or not aren't using NodeInfo objects or scala types like Array.Kind, they're calling PathExpression.steps.last.isArray, and I think that's what you want also. 

Actually I think you should add
```
def isArray: Boolean = steps.last.isArray
def isScalar: Boolean = !isArray
```
to PathExpression. 

Exists means "I don't care about the type" what it does care about is that the expression is a path expression i.e., where one can evaluate "does this exist". 

You want a minor variation on that where you can also ask if it's array or not. 

Dimensionality in DFDL/XSD is orthogonal to type.  I've always known that, but somehow decided a loooong time ago (like 2011) to put Array into the type hierarchy. That looks like a mistake now. 

Perhaps NodeInfo.Exists should be renamed to NodeInfo.PathExpr.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: Pretty sure you don't need Scalar. Just use Exists.

Exists means "I don't care about the type, I just need it to be a PathExpression"

More about this below in other comments.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: Steve is right, the type system cannot do this.

mbeckerle (changes requested):

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
mike-mcgann: Changed back to Exists. I can go ahead and rename Exists to PathExpr in this PR if you think we should make that change now. 

I found a isPathToOneWholeArray method in PathExpression which looks like it has the exact information needed here.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2829,15 @@ sealed abstract class LengthExprBase(
   ) {
 
   protected final def leafReferencedElements = {
+    if (!args(0).isInstanceOf[PathExpression]) {
```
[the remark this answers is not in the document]
mike-mcgann: Added the match pattern. 

Right now on the main branch there is a ClassCastException is the first argument to a function so there is something else wrong. I've filed ticket DAFFODIL-2795 for that issue and have commented out the relevant tests.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2829,19 @@ sealed abstract class LengthExprBase(
...
 
   protected final def leafReferencedElements = {
-    val arg = args(0).asInstanceOf[PathExpression]
+    val arg = args(0) match {
+      case arg: PathExpression => arg
+      case _ => Assert.invariantFailed("NodeInfo.Exists, but arg was not a PathExpression.")
+    }
+    // test_lengthPatternBinaryPatternLimit in section 12 shows that the length can
+    // be evaluated when it is an array of hex binary values
+    if (arg.inherentType != NodeInfo.HexBinary && arg.isPathToOneWholeArray)
```
mike-mcgann, line 2838: See comment in the source above. Making this permissive for hexbinary values was needed for that test to pass.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2829,19 @@ sealed abstract class LengthExprBase(
   ) {
 
   protected final def leafReferencedElements = {
-    val arg = args(0).asInstanceOf[PathExpression]
+    val arg = args(0) match {
+      case arg: PathExpression => arg
+      case _ => Assert.invariantFailed("NodeInfo.Exists, but arg was not a PathExpression.")
```
mbeckerle, line 2834: It's annoying, but invariantFailed always wants the COVERAGE-ON/COVERAGE-OFF stuff around it. 
I wish we could somehow build that in so it is implied for these assertions, but I don't know how. 

Maybe there's a way to configure coverty so that it has a file that declares these sort of things so that one does not need to surround them?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: This indicates a more significant issue.

The expresion "." when it corresponds to an element declaration that is an array, it does NOT denote the array, it denotes the one infoset node only. 

So if the array element is name 'a', then "." is not equivalent to "../a" which references the whole array. 

Furthermore, suppose you write ".[1]", and '.' corresonds to an element that is  declared to be an array. 

This was a subject of recent DFDL workgroup discussion, and the conclusion is that "." is never an array. So ".[1]" is always equivalent to just "." and refers to a single node only. 

So I think the path expression "." (A PathExpression containing one step: Self) is going to need special treatment.

tuxji:
> So I think the path expression "." (A PathExpression containing one step: Self) is going to need special treatment.

@mbeckerle Are you saying that this pull request needs an additional change?  Can you help @mike-mcgann by being more specific where the additional change may be needed?

stevedlawrence (commented):
Didn't realize I hadn't submitted a review. I had similar comments as Mike, which might help to answer @tuxji's question.

mbeckerle:
> > So I think the path expression "." (A PathExpression containing one step: Self) is going to need special treatment.
> 
> @mbeckerle Are you saying that this pull request needs an additional change? Can you help @mike-mcgann by being more specific where the additional change may be needed?

Yes, another change is needed.  To be more specific. hexBinary is not what needs an exception in this code. We just have a test for hexBinary which calls `dfdl:valueLength(., "bytes")`. 

The exception is not needed for hexBinary type, but for "." as the first argument to valueLength. For an array of *any* type, even a complex type, this expression `dfdl:valueLength(. , "bytes")`  (or contentLength, and for any units: bits, bytes, or characters) needs to treat "." specially, because "." doesn't refer to an array, ever, regardless of how the corresponding element declaration is declared. "." always refers to exactly one infoset DIElement node. 

I think the change is simple though. Add `def isSelf: Boolean` to PathExpression, and it is true if the path is just the Self step alone. Then use that to test for the exception, not the type. 

The hexBinary test should then still work, but we should also add a test with a small simple complex type to show this working for that case also.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mike-mcgann: > It's annoying, but invariantFailed always wants the COVERAGE-ON/COVERAGE-OFF stuff around it. I wish we could somehow build that in so it is implied for these assertions, but I don't know how.

I bracketed the invariant with COVERAGE-ON/COVERAGE-OFF. I checked the docs but there doesn't seem to be a way to configure this at function invocation time.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2829,13 +2834,23 @@ sealed abstract class LengthExprBase(
...
+      case arg: PathExpression => arg
+      case _ => {
+        // $COVERAGE-OFF$
+        Assert.invariantFailed("NodeInfo.Exists, but arg was not a PathExpression.")
+        // $COVERAGE-ON$
+      }
+    }
+    // test_lengthPatternBinaryPatternLimit in section 12 shows that the length can
+    // be evaluated when it is an array of hex binary values
+    if (!path.isSelf && path.isPathToOneWholeArray)
```
stevedlawrence, line 2847: Does the `path.isSelf` condition want to be checked inside `isPathToOneWholeArray`. A self path can never be a path to a whole array, so I imagine we would need this same check anywhere `isPathToOneWholeArray`, or we could just include it in he logic?

stevedlawrence:
> > The hexBinary test should then still work, but we should also add a test with a small simple complex type to show this working for that case also.
> 
> It looks like test [valueLength11](https://github.com/apache/daffodil/blob/main/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml#L418-L435) covers that condition.

That test doesn't use a self expression. vaLen also isn't an array using self. I think Mike was suggesting a similar test as the hexBinary one hat failed, but using a complex type instead of a simple hexbinary type?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mike-mcgann: I moved the `path.isSelf` check to be inside `isPathToOneWholeArray` and that seems to work.

mike-mcgann:
> That test doesn't use a self expression. vaLen also isn't an array using self. I think Mike was suggesting a similar test as the hexBinary one hat failed, but using a complex type instead of a simple hexbinary type?

I understand now. I've added those as contentLength_6 and valueLength_16

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -459,6 +459,37 @@
...
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="valueLength15" type="xs:unsignedInt" dfdl:inputValueCalc='{dfdl:valueLength("three", "bytes")}'/>
+
+    <xs:element name="valueLength16">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="e1">
```
stevedlawrence, line 477: We might want to make e1 an array (e.g. `maxOccurs="unbounded"`) since this issue was all about getting the length of arrays, ande make sure a self expression in a complex really is referencing the right thing. Same with the contentLength6 schema.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -387,7 +387,7 @@ object NodeInfo extends Enum {
   }
 
   /**
-   * For things like fn:exists, fn:empty, dfdl:contenLength
+   * For things like fn:exists, fn:empty
```
mbeckerle, line 390: Not sure this comment should be changed. I think dfdl:contentLength and dfdl:valueLength should still be using NodeInfo.Exists as their required type for the first argument.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: This is quite clean now. I like it.

mbeckerle (approved):
+1
I think this is sufficient. Only open suggestion at this point I think is steve lawrence's suggestion to improve tests a bit.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
mike-mcgann: Yeah, unbounded makes for a better test. I've added that change.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
mike-mcgann: Thanks for catching that one. Updated.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
stevedlawrence: You have made v1 an array, bu the valueLength path references e1 which is still a scalar. Can the maxOccurs unbounded be moved to the e1 element so it's a complex array?

mike-mcgann: It has been moved up to the e1 element.

PR #967, opened 2023-02-21
Consider ChoiceGroupRef's open when determining next elements

jadams-tresys:
When determining possible following elements for a group that is a choice, it is important to consider them Open as the group can be shared and have multiple contexts. The way things had worked before is we would push TermRuntimeData's for each group reference onto the stack when looking for the next possible element. However, with since these groups are actually choices with required elements in every branch they were considered closed and if there was an issue attempting to find a following element with the first group ref's TRD we would create an InvalidErrorERD instead of continuing down the stack until we got to the next group ref.

DAFFODIL-2615


daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
```diff
@@ -354,8 +355,12 @@ trait TermRuntime1Mixin { self: Term =>
           case (Closed(pnes1), Closed(pnes2)) => {
             thisItself // When everything is closed, we only want thisItself and not the entire list of Closed possibilities
           }
-          case (Open(pnes1), Closed(pnes2)) => {
+          case (Open(pnes1), Closed(pnes2)) if !this.isInstanceOf[ChoiceGroupRef] => {
```
jadams-tresys, line 358: I can seen an argument to be made here that this logic should actually be part of SeveralPossibilitiesForNextElement in PartialNextElementResolver for when we are determining if the element is actually required or not.  Figured I'd leave it up for discussion.

stevedlawrence: I guess one benefit here is that this is checked at compile time instead of runtime? Seems like the preferred logic.

stevedlawrence (approved):
+1, seems reasonable, though I'm not too familiar with this code, would like @mbeckerle's input

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
```diff
@@ -354,8 +355,12 @@ trait TermRuntime1Mixin { self: Term =>
...
             // If there are optional elements before a required element, the whole sequence is considered closed
+            //
+            // If this is a ChoiceGroupRef though, it could be one of multiple group references and thus we may
+            // not have the appropriate context to find the expected following element, so we should remain Open
```
mbeckerle, line 362: This comment says "so we should remain open", but the next line of code is a Closed.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/TermRuntime1Mixin.scala
mbeckerle: This seems ok to me here.

mbeckerle (approved):
+1 with fix to inconsistent comment.

PR #972, opened 2023-02-24
Update occursIndex to skip absent elements

mike-mcgann:
The occursIndex function should return the current index of the element in the infoset. Daffodil was using the current array position as this value. When the emptyElementParsePolicy is set to treatAsAbsent, empty elements are not included in the final infoset and therefore the current array position may not match the index in the final infoset.

This change introduces a new variable in the parser and unparser states for tracking the current occusIndex value which is not incremented when an absent element is seen.

[DAFFODIL-2791](https://issues.apache.org/jira/browse/DAFFODIL-2791)

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/debugger/InteractiveDebugger.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/debugger/InteractiveDebugger.scala
@@ -1837,7 +1837,7 @@ class InteractiveDebugger(
         val longDesc = desc
 
         def getSomeValue(state: StateForDebugger): Option[Long] = {
-          if (state.arrayPos != -1) Some(state.arrayPos) else None
+          if (state.occursPos != -1) Some(state.occursPos) else None
         }
       }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/ProcessorStateBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/ProcessorStateBases.scala
@@ -70,7 +70,8 @@ trait StateForDebugger {
   def bitLimit0b: MaybeULong
   def childPos: Long
   def groupPos: Long
-  def arrayPos: Long
+  def arrayIterationPos: Long
+  def occursPos: Long
   def variableMapForDebugger: VariableMap
   def delimitedParseResult: Maybe[dfa.ParseResult]
   def withinHiddenNest: Boolean
@@ -83,7 +84,8 @@ case class TupleForDebugger(
   val bitLimit0b: MaybeULong,
   val childPos: Long,
   val groupPos: Long,
-  val arrayPos: Long,
+  val arrayIterationPos: Long,
+  val occursPos: Long,
   val variableMapForDebugger: VariableMap,
   val delimitedParseResult: Maybe[dfa.ParseResult],
   val withinHiddenNest: Boolean,
@@ -473,7 +475,8 @@ abstract class ParseOrUnparseState protected (
       bitLimit0b,
       childPos,
       groupPos,
-      arrayPos,
+      arrayIterationPos,
+      occursPos,
       variableMap.copy(), // deep copy since variableMap is mutable
       delimitedParseResult,
       withinHiddenNest,
@@ -506,7 +509,8 @@ abstract class ParseOrUnparseState protected (
   final def bytePos = bytePos0b
 
   def groupPos: Long
-  def arrayPos: Long
+  def arrayIterationPos: Long
+  def occursPos: Long
   def childPos: Long
 
   def hasInfoset: Boolean
@@ -604,7 +608,8 @@ final class CompileState(
   tunable: DaffodilTunables,
 ) extends ParseOrUnparseState(tci.variableMap, Nil, maybeDataProc, tunable) {
 
-  def arrayPos: Long = 1L
+  def arrayIterationPos: Long = 1L
+  def occursPos: Long = 1L
   def bitLimit0b: MaybeULong = MaybeULong.Nope
   def bitPos0b: Long = 0L
   def childPos: Long = 0L
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
@@ -78,7 +78,8 @@ object MPState {
 
   final class Mark {
 
-    var arrayIndexStackMark: MStack.Mark = _
+    var arrayIterationIndexStackMark: MStack.Mark = _
+    var occursIndexStackMark: MStack.Mark = _
     var groupIndexStackMark: MStack.Mark = _
     var childIndexStackMark: MStack.Mark = _
     var occursBoundsStackMark: MStack.Mark = _
@@ -86,19 +87,23 @@ object MPState {
     clear()
 
     def clear(): Unit = {
-      arrayIndexStackMark = MStack.nullMark
+      arrayIterationIndexStackMark = MStack.nullMark
+      occursIndexStackMark = MStack.nullMark
       groupIndexStackMark = MStack.nullMark
       childIndexStackMark = MStack.nullMark
       occursBoundsStackMark = MStack.nullMark
     }
 
     def captureFrom(mp: MPState): Unit = {
-      arrayIndexStackMark = mp.arrayIndexStack.mark
+      arrayIterationIndexStackMark = mp.arrayIterationIndexStack.mark
+      occursIndexStackMark = mp.occursIndexStack.mark
       groupIndexStackMark = mp.groupIndexStack.mark
       childIndexStackMark = mp.childIndexStack.mark
     }
+
     def restoreInto(mp: MPState): Unit = {
-      mp.arrayIndexStack.reset(this.arrayIndexStackMark)
+      mp.arrayIterationIndexStack.reset(this.arrayIterationIndexStackMark)
+      mp.occursIndexStack.reset(this.occursIndexStackMark)
       mp.groupIndexStack.reset(this.groupIndexStackMark)
       mp.childIndexStack.reset(this.childIndexStackMark)
     }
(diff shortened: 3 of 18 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestSepTests.scala
```diff
@@ -70,6 +70,8 @@ class TestSepTests {
   @Test def test_sep_ssp_never_5(): Unit = { runner.runOneTest("test_sep_ssp_never_5") }
 
   // DAFFODIL-2791
-  @Test def test_treatAsMissing_occursIndex(): Unit = { runner.runOneTest("test_treatAsMissing_occursIndex") }
```
mbeckerle: This line doesn't seem too long to me. Fits fine on my screen for github code review and my browser is zoomed to 133%.

Should we allow longer lines?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceParserBases.scala
```diff
@@ -194,6 +194,12 @@ abstract class SequenceParserBase(
                 pstate.mpstate.moveOverOneArrayIndexOnly()
               }
 
+              // If the emptyElementParsePolicy is set to treatAsAbsent, don't
+              // increment the occursIndex if the element is absent
+              if (resultOfTry != AbsentRep) {
```
mbeckerle, line 199: Does resultOfTry only come back AbsentRep if treatAsAbsent is set?
I'm not sure that's the case. If so then ok.

I expected this to test resultOfTry as you have, but also to inspect the property (which has to be captured onto the runtime data structures) to see if it is treatAsAbsent.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/unparsers/UState.scala
```diff
@@ -459,7 +464,8 @@ final class UStateForSuspension(
 
   override def groupPos = 0 // was die, but this is called when copying state during debugging
   override def currentInfosetNodeMaybe = Maybe(currentInfosetNode)
-  override def arrayPos = occursIndex
+  override def arrayPos = arrayIndex
+  override def occursIndex = occurs
```
stevedlawrence, line 468: Suggest occursPos for consisence.

mbeckerle (approved):
+1

daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestSepTests.scala
tuxji: I'm using a laptop screen since I'm away from both home and office now, so I see only this much of the line and I would have to use the horizontal scrollbar to see the rest of the line (my browser is zoomed to 100%):

```scala
  @Test def test_treatAsMissing_occursIndex(): Unit = { runner.runOneTest("test_treatAsMissing_occursI
```

BTW, I checked and found out if I increase maxColumn to 100, sbt scalafmtCheckAll fails and scalafmtAll reformats a bunch of files to re-join some files that had been split before.  Any time we change the allowed length, we'll reformat some files.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
```diff
@@ -108,12 +113,16 @@ object MPState {
 class MPState private () {
 
   val arrayIndexStack = MStackOfLong()
```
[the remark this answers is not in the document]
mike-mcgann: It is kind of iterating over one array and creating a new array which may or may not be the same length. How about naming these two to be `arraySourcePos` and `arrayTargetPos`?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceParserBases.scala
mike-mcgann: I was assuming that if treatAsAbsent was not set that it would be EmptyRep instead which I think is the case? I did try to see if I could check the emptyElementParsePolicy at that point but I wasn't able to navigate the object tree to find it.

mbeckerle: Ok. I suspect you are right, and if tests aren't breaking on account of this, then we are almost certainly ok.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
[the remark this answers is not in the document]
stevedlawrence: I'm not sure source and target are obvious to me, especially since it could be confusing depending on parse vs unparse. For example, when parsing the source is data but when unparsing the source is the infoset. But along the same lines, maybe using infoset/data would make the distinction more obivous? So maybe something like arrayDataPos and arrayInfosetPos?

@mbeckerle, I skimmed the DFDL spec but couldn't really find any obvious terms to steal from DFDL. Maybe DFDL already has concepts for this kind of thing?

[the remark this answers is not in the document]
mike-mcgann: I went ahead and put in the first suggestion of `arrayIterationIndexStack` and `arrayIterationPos`. We can easily refactor the name if we come up with a better one.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
```diff
@@ -108,12 +113,16 @@ object MPState {
...
+  val occursIndexStack = MStackOfLong()
 
   def moveOverOneArrayIndexOnly() = arrayIndexStack.push(arrayIndexStack.pop + 1)
   def moveBackOneArrayIndexOnly() = arrayIndexStack.push(arrayIndexStack.pop - 1)
 
   def arrayPos = arrayIndexStack.top
 
+  def incrementOccursIndex() = occursIndexStack.push(occursIndexStack.pop + 1)
```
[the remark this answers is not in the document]
mike-mcgann: Updated to `moverOverOneOccursIndexOnly` and removed `moveBackOneArrayIndex`

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/unparsers/UState.scala
```diff
@@ -413,7 +415,8 @@ final class UStateForSuspension(
   override var dataOutputStream: DirectOrBufferedDataOutputStream,
   vbox: VariableBox,
   override val currentInfosetNode: DINode,
-  occursIndex: Long,
+  arrayIndex: Long,
+  occurs: Long,
```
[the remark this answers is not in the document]
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/ProcessorStateBases.scala
```diff
@@ -507,6 +507,7 @@ abstract class ParseOrUnparseState protected (
 
   def groupPos: Long
   def arrayPos: Long
+  def occursIndex: Long
```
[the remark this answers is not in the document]
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceParserBases.scala
```diff
@@ -194,6 +194,12 @@ abstract class SequenceParserBase(
                 pstate.mpstate.moveOverOneArrayIndexOnly()
               }
 
+              // If the emptyElementParsePolicy is set to treatAsAbsent, don't
+              // increment the occursIndex if the element is absent
+              if (resultOfTry != AbsentRep ) {
+                pstate.mpstate.incrementOccursIndex()
+              }
```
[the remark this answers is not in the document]
mike-mcgann: There was no actual reason for that. I moved the absent check into the `ais eq Done`  block

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/unparsers/UState.scala
mike-mcgann: Updated

mike-mcgann:
> I wonder if it would be worth moving checks that aren't specific to OS and JDK versions into their own "lint" job inside main.yml? It must be discouraging to make a change to a PR and see 3 out of 6 matrix jobs fail across the board. 

Yeah, I think that is a better long-term solution. At that time it would also be nice to see if we can have the tests run in a docker container with all the dependencies and configurations installed and ready-to-go. A bonus to that is having a image you can use locally that is the same one used in the CI build.

stevedlawrence:
> I wonder if it would be worth moving checks that aren't specific to OS and JDK versions into their own "lint" job inside main.yml?

Agreed. It would also make the OS/JDK jobs finish 1-2 minutes quicker. Not a huge difference, but the quicker tests finish the better. The new job could also be configured to always run all checks even if one fails.

> At that time it would also be nice to see if we can have the tests run in a docker container with all the dependencies and configurations installed and ready-to-go.

I wouldn't want to have to maintain multiple docker images for CI tests. We already have 7 different OS/JDK combinations we have. Though I agree that a dev container could be useful. Though, most of the dependency work is handled by sbt which is straightforward to install. I think the only deps it doesn't handle is gcc and MiniXML dep, which I think most distros already provide.

Whatever is decided can all be done as a separate PR though.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
```diff
@@ -107,12 +112,16 @@ object MPState {
...
-  val arrayIndexStack = MStackOfLong()
+  val arrayIterationIndexStack = MStackOfLong()
+  val occursIndexStack = MStackOfLong()
+
+  def moveOverOneArrayIterationIndexOnly() =
+    arrayIterationIndexStack.push(arrayIterationIndexStack.pop + 1)
 
-  def moveOverOneArrayIndexOnly() = arrayIndexStack.push(arrayIndexStack.pop + 1)
-  def moveBackOneArrayIndexOnly() = arrayIndexStack.push(arrayIndexStack.pop - 1)
+  def arrayPos = arrayIterationIndexStack.top
```
stevedlawrence, line 121: Can this be renamed `arrayIterationPos`?

mike-mcgann: Thanks for finding that one. I should have done a grep for arrayPos.

mike-mcgann:
> I wouldn't want to have to maintain multiple docker images for CI tests. We already have 7 different OS/JDK combinations we have.

We only need one for linux and latest and greatest JDK. It could be used for that build and for the lint check. The other combos could be done as it is now. It might be faster for CI to download the image than to install the dependencies and there is a chance that the CI could cache the image.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/debugger/InteractiveDebugger.scala
```diff
@@ -1837,7 +1837,7 @@ class InteractiveDebugger(
         val longDesc = desc
 
         def getSomeValue(state: StateForDebugger): Option[Long] = {
-          if (state.arrayPos != -1) Some(state.arrayPos) else None
+          if (state.arrayIterationPos != -1) Some(state.arrayIterationPos) else None
```
stevedlawrence, line 1840: Should this return `occursIndexPos`? The debugger command is called `occursIndex` so that might be what users expect?

mike-mcgann: This was an auto-refactor here. I went ahead and updated it to use occursPos.

stevedlawrence (approved):
+1, good to squash and merge (which you should able to do now that you're PMC).

PR #973, opened 2023-02-24
Ignore errors from path expressions that are not descendants of the distinguished root

stevedlawrence:
It is not uncommon for schemas to create global element, group, or types declarations that are not descendants of the distinguished root, and which contain path expressions that got up and out of the global declaration. This is useful in cases where a schema may want alternative distinguished roots, or to have global declarations that can be plugged in as if it were a library of schema elements.

However, even if these elements are not referenced from the distinguished root, we may still attempt to compile them if the checkAllTopLevel flag is set (which is true be default), in which case these expressions that reach out of the global declaration can lead to schema definition errors that cannot be fixed.

To support these kinds of schemas, this adds a flag to the DPathElementCompileInfo indicating which element is the distinguished root. If a path goes outside its local root but that root isn't the distinguished root, then it must be one of the above mentioned cases, and we can throw a PathNoContextException which ignores the error outputs a warning.

DAFFODIL-2393


tuxji (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -868,11 +868,25 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-    val err = new RelativePathPastRootError(
-      this.schemaFileLocation,
-      "Relative path '%s' past root element.",
-      this.wholeExpressionText,
-    )
+    // This path expression cannot be compiled because we went past the root. This normally
+    // should be an SDE with a RelativePathPastRootError. However, if we don't have any element
+    // compile infos or the current root is not the distinguished root that Daffodil is
+    // compiling, it means this expression is used on a term that is not a descendent of the
+    // distinguish root. The expression might make sense if a different distinguished root were
```
mbeckerle, line 875: typo: shoudl be "// distinguished root"

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions_unused_path_no_context.dfdl.xsd
```diff
@@ -28,20 +28,119 @@
...
+
+
+  <xs:element name="unused_element_4">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="value9" type="xs:string" />
+        <xs:element name="nest1">
+          <xs:complexType>
+            <xs:sequence>
+              <xs:element name="nest2" type="xs:string" dfdl:terminator="{ ../../value9 }" />
```
mbeckerle, line 98: Is this line producing a warning? I see a `<warning>value9</warning>` above.

I don't think this should produce a warning. The ../.. walks up to nest1 then unused_element_4, but not past. 

unused_element_4 is an element. Had it been a type or group then yes, I can see a warning being ok. But it's an element. So why should this warn?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions_unused_path_no_context.dfdl.xsd
```diff
@@ -28,20 +28,119 @@
...
   </xs:annotation>
 
+
   <xs:element name="e1" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="1" />
 
-  <xs:group name="unused_group">
+  <xs:group name="unused_group_1">
     <xs:sequence>
       <xs:element name="value1" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="1" />
       <xs:sequence dfdl:terminator="{ ./value1 }" />
```
mbeckerle, line 37: So, this is a bit odd. The . in ./value1 refers to "some parent element" because this is inside a group. 

But we don't care what element that is. 

I'm ok with this issuing a warning, I think, so long as existing schemas don't pour out tons of such warnings. 
But this is the borderline case where it would be nice if there was a way to avoid the warning in this case. 

I can see how the expression compiler can't cope with this though. It wants that Up step to refer to an element that has DPathCompileInfo. There is no such. 

So this is a corner case where we definitely cannot compile the expression, but we may not want to issue a warning. 

I'd say let's merge this in, and try  it out and if we get too many such warnings deal with it then.

mbeckerle (approved):
+1, but please reconfirm behavior of the value9 warning per comment.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions_unused_path_no_context.dfdl.xsd
stevedlawrence: Yeah, this was something weird I did notice. I think the issue is that unused_element_4 is a GlobalElementDecl, so there is no DPathElementCompileInfo associated with it. I believe only Root, ElementRef, and LocalElementDecl actually get DPathElementCompileInfos.

So in the context of nest2, `../../` goes to nothing (at least nothing that has a DPathElementCompileInfo. nest1 has no parent DPathElementCompileInfo and nest2 has no grandparent. Because of that, we think it does actually go up and out, even though it doesn't really.

Maybe a fix would be when we compile a GlobalElemntDecl because checkAllTopLevel is defined, we instead create a temporary Root (but one that isn't a distinguished root since we need this for the up and out check). This way there would be DPathElementCompileInfo for the unused_element_4, and there would not be a warning. Maybe there is a better approach? The feels kindof hacky just to avoid this warning, but maybe not?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions_unused_path_no_context.dfdl.xsd
stevedlawrence: Yep, that's exactly what's going on. Similar to my other comment, when we compile unused groups we could maybe create a fake "root" element contain a group ref to the unused group. This way `./` would refer to something with a DPathElementCompileInfo and it would work as expected. And if it did actually go outside the fake root, only then would you get a warning. That could potentially reduce a lot of warnings if it's something we did start to see more of them.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions_unused_path_no_context.dfdl.xsd
stevedlawrence: Doing some more playing around with this, a related issue is we don't even compile expression on global element decls that aren't used. For example, we could have a global element with an invalid expression like this:

```xml
  <xs:element name="unused" type="xs:string" dfdl:terminator="{ dfdl:neverCompiled() }" />
```

And no error is created. As above, I think if we created a fake root when compiling this it might solve the problem, but I'm not sure it's worth added complexity.

mbeckerle: Maybe we should just warn that some things (elements, types, groups) are unused and so are not compiled and leave it at that.

PR #974, opened 2023-02-28
Add checks for bit lengths

mike-mcgann:
Daffodil was allowing bit lengths on binary types that exceed the width of the type even though this is not allowed by the DFDL specification. This change adds a check at compile time to see if the bit length value is valid when specified as a constant and at runtime when the bit length is specified an expression.

[DAFFODIL-1704](https://issues.apache.org/jira/browse/DAFFODIL-1704)

### DIFF
--- a/daffodil-codegen-c/src/test/resources/org/apache/daffodil/codegen/c/ex_nums.dfdl.xsd
+++ b/daffodil-codegen-c/src/test/resources/org/apache/daffodil/codegen/c/ex_nums.dfdl.xsd
@@ -138,7 +138,7 @@
                                         dfdl:byteOrder="littleEndian"/>
                             <xs:element name="le_int8" type="xs:byte"
                                         dfdl:byteOrder="littleEndian"/>
-                            <xs:element name="le_int46" type="xs:int"
+                            <xs:element name="le_int46" type="xs:integer"
                                         dfdl:byteOrder="littleEndian"
                                         dfdl:length="46"
                                         dfdl:lengthKind="explicit"/>
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
@@ -705,8 +705,9 @@ trait ElementBase
     // of a variable-width charset, so greater than 0, but we don't know
     // exactly.
     Assert.usage(repElement.isFixedLength)
-    if (repElement.lengthKind =:= LengthKind.Explicit) repElement.lengthEv.optConstant.get
-    else {
+    if (repElement.lengthKind =:= LengthKind.Explicit) {
+      repElement.lengthEv.optConstant.get
+    } else {
       Assert.invariant(repElement.lengthKind =:= LengthKind.Implicit)
       // it's a string with implicit length. get from facets
       schemaDefinitionUnless(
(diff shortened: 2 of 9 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -167,6 +167,16 @@ abstract class BinaryIntegerBaseParser(
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
     if (nBits == 0) return // zero length is used for outputValueCalc often.
+    if (primNumeric.width.isDefined) {
+      val width = primNumeric.width.get
+      if (nBits > width)
+        PE(
+          start,
```
mbeckerle, line 174: Definitely need a test to exercise this.

mbeckerle (approved):
Add a test of the runtime check.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
mike-mcgann: I tried to test this by using an expression such as `{ 2 * 8 }` but I guess that expression could actually be evaluated at compile time and wasn't exercising the runtime check. I updated the test to get the value from the first byte in the data stream.

mbeckerle: Yup. Poor-man's constant folding using an obvious hack - infoset to be a fake infoset that just throws if you try to access any method on it. Eval expression. No infoset access means it can compute the value. If it touches the infoset, it throws, and determines therefore that it can't constant-fold the expression.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -750,21 +751,46 @@ trait ElementBase
...
+        case primNumeric: NodeInfo.PrimType.PrimNumeric =>
+          if (primNumeric.width.isDefined) {
+            val nBits = result.get
+            val width = primNumeric.width.get
+            if (nBits > width) {
+              SDE(
+                "Number of bits %d out of range, must be between 1 and %d bits.",
+                nBits,
+                width,
+              )
```
stevedlawrence, line 787: Might be useful to add "binary" and the type in here to make it clear the the type doesn't match the bits.
```suggestion
              SDE(
                "Number of bits %d out of range for binary %s, must be between 1 and %d bits.",
                nBits,
                primNumeric.globalQName,
                width,
              )
```

stevedlawrence (approved):
+1, minor suggestion for maybe a slightly more helpful error message

daffodil-codegen-c/src/test/resources/org/apache/daffodil/codegen/c/ex_nums.dfdl.xsd
```diff
@@ -138,7 +138,7 @@
                                         dfdl:byteOrder="littleEndian"/>
                             <xs:element name="le_int8" type="xs:byte"
                                         dfdl:byteOrder="littleEndian"/>
-                            <xs:element name="le_int46" type="xs:int"
+                            <xs:element name="le_int46" type="xs:integer"
```
tuxji, line 141: Nice to see that we can test integer and nonNegativeInteger again in ex_nums again.

tuxji (approved):
+1

PR #975, opened 2023-03-01
Warn when annotations placed on group definition

jadams-tresys:
Annotations placed on group definitions will be ignored, so we should warn users that this is occurring.

DAFFODIL-2668

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
@@ -21,6 +21,7 @@ import scala.xml.Comment
 import scala.xml.Node
 import scala.xml.Text
 
+import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.lib.exceptions.Assert
 
 object GlobalGroupDef {
@@ -118,6 +119,7 @@ sealed abstract class GlobalGroupDef(
 
   def groupMembers = {
     validateChoiceBranchKey()
+    checkForGroupDefAnnotations()
     this.groupMembersNotShared
   }
 
@@ -129,6 +131,18 @@ sealed abstract class GlobalGroupDef(
         "dfdl:choiceBranchKey cannot be specified on the choice/sequence child of a global group definition",
       )
     }
+  }
+
+  private def checkForGroupDefAnnotations(): Unit = {
+    // Ensure that the group definition itself does not have any annotations.
+    // Annotations should only be on group references or children of the group
+    // definition
+    val dais = getDFDLAppinfos(defXML \ "annotation" \ "appinfo")
+    if (dais.nonEmpty)
+      SDW(
+        WarnID.InvalidAnnotationPoint,
+        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference, sequence or choice."
+      )
   }
 
   final override lazy val name = defXML
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/AnnotatedSchemaComponent.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/AnnotatedSchemaComponent.scala
@@ -19,6 +19,7 @@ package org.apache.daffodil.core.dsom
 
 import scala.collection.mutable
 import scala.xml.Node
+import scala.xml.NodeSeq
 
 import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.lib.equality._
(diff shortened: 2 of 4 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
```diff
@@ -131,6 +133,15 @@ sealed abstract class GlobalGroupDef(
     }
   }
 
+  private def checkForGroupDefAnnotations(): Unit = {
+    // Ensure that the group definition itself does not have any annotations.
+    // Annotations should only be on group references or children of the group
+    // definition
+    val found = defXML.child.map(_.label).contains("annotation")
```
mbeckerle, line 140: Need to check that the annotation is specifically an appinfo annotation with source="http://www.ogf.org/dfdl/".
Those should create the warning.

You might want to tolerate not a perfect match on the source URI. 

See annotatedSchemaComponent.scala

Look for ``` lazy val dfdlAppInfos = { ... ``` for this logic.

The common logic for finding a DFDL appinfo, tolerating small imperfections in the URI, should be shared between here and there. I'd suggest the common code belongs in XMLUtils, called from both places.

mbeckerle (changes requested):

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
tuxji: Agreed.  Will expect a new commit to check appinfo and fix formatting.

jadams-tresys: I've moved the logic to XMLUtils via a getDFDLAppinfos function.  However, it loses the ability to send schema definition warnings and I'm not sure what a good solution is to get that functionality back. Should there just be another pass through all the appinfos in AnnotatedSchemaComponent to issue warnings as needed?  I thought about having my function return a list of warnings in addition to the list of appinfos and then doing SDW's back in AnnotatedSchemaComponent, but it felt very hacky to have a function returning a tuple with a list containing a tuple for WarnID, String.

mbeckerle (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
mbeckerle: Suggest make dfdlAppInfos return a `Either[Seq[(WarnID, String)], Seq[DFDLAppInfo]]`
where Left is warnings to issue, Right is the appinfo object(s) if all goes well.

stevedlawrence: Maybe another alternative, doesn't `GlobalGroupDef` implement `AnnotatedSchemaComponet` so it has access to dfdlAppInfos? Can the new check function just do something like:
```scala
if (dfdlAppInfos.length > 0) SDW(...)
```
So you reuse the existing dfdlAppInfos logic to find any dfdl annotations (which might warn), but then we can warn that they will be ignored in the check?

jadams-tresys: So it does implement AnnotatedSchemaComponent, but only on the groupXML, which is just the group members I believe.  It doesn't include the actual group definition, which is how the annotations for group definitions were getting ignored before.

stevedlawrence: I see. Maybe instead of making it an `XMLUtils` function it becomes a public function in `AnnotatedMixin` that accepts the XML node to look for annotations as a paramter? This way it could still create warnings like normal, and dfdlAppInfo could pass in `xml` while the GroupDef check could pass in `defXML`? Just trying to avoid having a complex return type.

jadams-tresys: I like this suggestion from Steve.  Changes for it have been pushed up for review.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
```diff
@@ -131,6 +133,18 @@ sealed abstract class GlobalGroupDef(
...
 
+  private def checkForGroupDefAnnotations(): Unit = {
+    // Ensure that the group definition itself does not have any annotations.
+    // Annotations should only be on group references or children of the group
+    // definition
+    val dais = getDFDLAppinfos(defXML \ "annotation" \ "appinfo")
+    if (dais.nonEmpty)
+      SDW(
+        WarnID.InvalidAnnotationPoint,
+        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference",
```
stevedlawrence, line 144: As an alternative to the group reference, could annotations also be placed on the model group (i.e. choice or sequence) of the group definition? I imagine if someone put an annotation on the group def that's maybe what they meant to do? Is it worth saying something like `... should be placed on the group reference or model group`?

stevedlawrence (approved):
+1

mbeckerle:
+1. This is cleaner than what I was suggesting.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
tuxji: Saying "or model group" may confuse a user who isn't familiar with the meaning of "model group."  I suggest either leaving the message the way it is now or changing it to `... should be placed on the group reference, choice, or sequence` if we really want to list all the possible places.

tuxji (approved):
+1

PR #977, opened 2023-03-02
Add WarnID to Schema Definition Warning messages

mike-mcgann:
Warnings in Daffodil can be suppressed by using an identifier but this can be difficult to obtain as the identifier is not displayed to the end user. This change adds this identifier to the message, when available, and contains instructions on how to suppress the warning.

[DAFFODIL-2634](https://issues.apache.org/jira/browse/DAFFODIL-2634)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
```diff
@@ -120,10 +121,19 @@ class SchemaDefinitionWarning(
...
 
   override def isError = false
   def modeName = "Schema Definition"
+
+  override def toString = warnID match {
+    case Some(id) =>
+      super.toString + "\n" +
+        s"To suppress this warning, add '${id}' to the " +
+        "daf:suppressSchemaDefinitionWarnings element of a Daffodil defineConfig " +
+        "used when processing"
```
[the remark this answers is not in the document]
tuxji: I agree we should be as terse as possible and add only the WarnID to the message.  I suspect it'll be easier to append the WarnID identifier after the format string than to insert it between the Schema Definition Warning and the format string.

tuxji (commented):
I think it might be pretty easy to remove all SDW methods and callers which create warnings without a WarnID.  Let's do that in this pull request too.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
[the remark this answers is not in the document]
mike-mcgann: I implemented the second suggestion. It seemed too difficult to add the identifier inline.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
```diff
@@ -120,20 +120,14 @@ class SchemaDefinitionWarning(
...
 
-  override def toString = warnID match {
-    case Some(id) =>
-      super.toString + "\n" +
-        s"To suppress this warning, add '${id}' to the " +
-        "daf:suppressSchemaDefinitionWarnings element of a Daffodil defineConfig " +
-        "used when processing"
-    case None => super.toString
-  }
+  override def getMessage = super.getMessage + s" (id: ${warnID})"
```
stevedlawrence, line 129: I believe this puts the warn id after file/line number info if it's included in the SDW (which is almost always is), which might be confusing? Does the warn id want to be closer to the actual message?

Maybe we do somehing like:
```scala
class SchemaDefinitionWarning(
  ...
) extends SchemaDefinitionDiagnosticBase(
    ...,
    Maybe(kind + s" (id: ${warnID})")
    ...,
) {
```

So we actually append the warnID to the diagnostic format string so it appears before any location information.

tuxji:
Can I get a second committer to agree about 1) removing all SDW macros/methods that don't take a WarnID, and 2) renaming the schemaDefinitionWarningWhen/Unless methods to SDWwhen/SDWunless to make it easier to find all SDW calls with a single grep for "SDW"?

stevedlawrence:
> Can I get a second committer to agree about 1) removing all SDW macros/methods that don't take a WarnID, and 2) renaming the schemaDefinitionWarningWhen/Unless methods to SDWwhen/SDWunless to make it easier to find all SDW calls with a single grep for "SDW"?

Both changes seem like good improvements to me. We might also want to shorten `schemaDefinitionUnless` and `schemaDefintionWhen` and make everything consistent. I also kindof like an alternative capitalization of `sdwWhen` vs `SDWwhen`, but don't feel strongly about it. Though we might want to do all the renaming of stuff in a separate PR though, as it's creeping outside the scope just including warnID in messages.

mike-mcgann:
> Can I get a second committer to agree about 1) removing all SDW macros/methods that don't take a WarnID, and 2) renaming the schemaDefinitionWarningWhen/Unless methods to SDWwhen/SDWunless to make it easier to find all SDW calls with a single grep for "SDW"?

I agree about #1 and started implementing that. I ran into some issues with updating the RuntimeSDWs for some reason. When I run the test suite some UDF tests fail and I'm trying to track that down. It isn't specific. It just errors out saying that the UDF is not found.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
mike-mcgann: I can look into that. I tried to implement the first idea where the ID was right after the SDW text but the information needed seemed to be private.

mike-mcgann: Ok, I updated it to that which doesn't require opening private definitions.

mike-mcgann:
WarnID is now required and that was easy enough to change.

mike-mcgann:
> Though we might want to do all the renaming of stuff in a separate PR though, as it's creeping outside the scope just including warnID in messages.

I'll leave out the renaming changes for now

stevedlawrence (approved):
+1

mike-mcgann:
Is there any additional work we would like on this ticket or is it ready to go?

tuxji (approved):
+1

PR #978, opened 2023-03-03
Appease scalafmt with extraneous comma

jadams-tresys:
In my last PR I missed that scalafmt wasn't happy after I removed what I thought was an extraneous comma.  The PR page said that all checks had passed, but it must have been stale and needed refreshing to see the scalamt fail.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
@@ -141,7 +141,7 @@ sealed abstract class GlobalGroupDef(
     if (dais.nonEmpty)
       SDW(
         WarnID.InvalidAnnotationPoint,
-        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference, sequence or choice."
+        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference, sequence or choice.",
       )
   }
 


jadams-tresys:
Ah, I think I realized what happened.  On my one machine I have my "origin" set to my daffodil fork, but the other machine 'origin' points to the main branch.  My mistake there.

stevedlawrence:
> We should add this command to some setup instructions.

Good call. I've added it to step 6 of the [Code Contributor Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow)

mbeckerle:
Honestly, the only way I stay sane with git is no remote is named "origin". I name them all based on who owns them. My fork is named mbeckerle. If for review I pull down someone else' fork to look at more closely than a github review, the name of that remote is that person's username. The central daffodil repo is named "asf", at my job the central repos for other projects are named "owl", or if they are for clients it's the client/project's name.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/GroupDef.scala
```diff
@@ -141,7 +141,7 @@ sealed abstract class GlobalGroupDef(
     if (dais.nonEmpty)
       SDW(
         WarnID.InvalidAnnotationPoint,
-        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference, sequence or choice."
+        "Annotations placed directly on a group definition will be ignored by DFDL. Any annotation expected to be processed by DFDL should instead be placed on the group reference, sequence or choice.",
```
mbeckerle, line 144: If this comma is truly required, then our scalafmt rules are wrong.

This comma should be tolerated, ... maybe... but not required. 

This is not a list, it's the arguments to a function call. They are fundamentally different concepts that only occasionally overlap.

tuxji (approved):
+1

PR #981, opened 2023-03-08
Update parsing when emptyValueDelimiterPolicy is 'none'

mike-mcgann:
When a value is empty, has an initiator and terminator defined, and the emptyValueDelimiterPolicy is set to 'none', the initiator and terminator should not be found in the input stream and should not be emitted when unparsing. Daffodil was raising a parse error in this case. This update allows parsing to continue when the initiator or terminator are not required to appear.

[DAFFODIL-2205](https://issues.apache.org/jira/browse/DAFFODIL-2205)

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/LocalElementMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/LocalElementMixin.scala
@@ -25,6 +25,7 @@ import org.apache.daffodil.lib.equality._
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.schema.annotation.props.gen._
 import org.apache.daffodil.runtime1.dpath.NodeInfo.PrimType
+import org.apache.daffodil.runtime1.processors.parsers.DelimiterTextType
 
 /**
  * Common to local element decls and element references
@@ -84,6 +85,10 @@ trait LocalElementMixin extends ParticleMixin with LocalElementGrammarMixin {
       else if (
         isDefaultable && !hasTerminator && emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Terminator
       ) true
+      else if (
+        hasInitiator && hasTerminator && emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.None
+      )
+        true
       else if (hasInitiator) {
         //
         // TODO: It is possible that this initiator expression cannot match zero length
@@ -123,5 +128,22 @@ trait LocalElementMixin extends ParticleMixin with LocalElementGrammarMixin {
       } else Assert.impossibleCase()
     res
   }.value
+
+  final def isDelimiterRequiredWhenEmpty(t: DelimiterTextType.Type): Boolean = {
+    val res = t match {
+      case DelimiterTextType.Initiator =>
+        if (!hasInitiator) false
+        else if (emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Both) true
+        else if (emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Initiator) true
+        else false
+      case DelimiterTextType.Terminator =>
+        if (!hasTerminator) false
+        else if (emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Both) true
+        else if (emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Terminator) true
+        else false
+      case _ => false
+    }
+    res
+  }
 
 }
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/DelimiterUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/DelimiterUnparsers.scala
@@ -31,6 +31,7 @@ import org.apache.daffodil.runtime1.processors.unparsers._
 class DelimiterTextUnparser(
   override val context: TermRuntimeData,
   delimiterType: DelimiterTextType.Type,
+  requiredOnEmptyValue: Boolean,
 ) extends TextPrimUnparser {
 
   private def erd = context
@@ -54,6 +55,15 @@ class DelimiterTextUnparser(
       s"Unparsing starting at bit position: ${state.dataOutputStream.maybeAbsBitPos0b}",
     )
 
+    // Do not emit the delimiter if the value is empty and the emptyValueDelimiterPolicy
+    // indicates it should be suppressed
+    if (state.currentInfosetNode.isSimple) {
+      val node = state.currentInfosetNode.asSimple
+      if (node.hasValue && node.isEmpty && !requiredOnEmptyValue) {
+        return
+      }
+    }
+
     val localDelimNode = state.localDelimiters
 
     val delimDFAs = {
@@ -71,7 +81,6 @@ class DelimiterTextUnparser(
 
     try {
       val valueString = delimDFA.unparseValue
-
       val outStream = state.dataOutputStream
       val nCharsWritten = outStream.putString(valueString, state)
       if (nCharsWritten != valueString.length)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
@@ -45,6 +45,8 @@ class DelimiterTextParser(
   delimiterType: DelimiterTextType.Type,
   isDelimited: Boolean,
   mustMatchNonZeroData: Boolean,
+  requiredOnEmptyValue: Boolean,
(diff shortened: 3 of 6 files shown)


tuxji (approved):
+1

Looks fine to me.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/DelimiterUnparsers.scala
```diff
@@ -54,6 +55,15 @@ class DelimiterTextUnparser(
       s"Unparsing starting at bit position: ${state.dataOutputStream.maybeAbsBitPos0b}",
     )
 
+    // Do not emit the delimiter if the value is empty and the emptyValueDelimiterPolicy
+    // indicates it should be suppressed
+    if (state.currentInfosetNode.isSimple) {
+      val node = state.currentInfosetNode.asSimple
+      if (node.hasValue && node.isEmpty && !requiredOnEmptyValue) {
+        return
```
stevedlawrence, line 63: We usually try to avoid return. I think the only way around it is to wrap everyhing in an if-block, which could be messy. So maybe it's fine in this case? Thoughts from others?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -108,12 +109,14 @@ class DelimiterTextParser(
     if (foundDelimiter.isDefined) {
       if (!containsLocalMatch(foundDelimiter.get.matchedDFAs, start)) {
         // It was a remote delimiter but we should have found a local one.
-        PE(
-          start,
-          "Found out of scope delimiter: '%s' '%s'",
-          foundDelimiter.get.matchedDFAs(0).lookingFor,
-          Misc.remapStringToVisibleGlyphs(foundDelimiter.get.matchedDelimiterValue.get),
-        )
+        if (requiredOnEmptyValue) {
```
stevedlawrence, line 112: I'm not sure I understand this. We found a delimiter, but not the one we were looking for. Seems like that should always be a PE?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -145,8 +147,8 @@ class DelimiterTextParser(
         localTypedDelims = localTypedDelims + " " + scannedDelims.next().lookingFor
       }
 
-      PE(start, "%s '%s' not found", delimiterType.toString, localTypedDelims)
-      return
+      if (!proceedOnMissingDelimiter)
+        PE(start, "%s '%s' not found", delimiterType.toString, localTypedDelims)
```
stevedlawrence, line 151: Where do we require that the value was empty? It seems like this would allow a delimiter to be missing when there is a non-empty value?

stevedlawrence (commented):
An overall thought is that we supposedly already support emptyValueDelmiterPolicy for `initiator`, `terminator`, and `both`? Seems like the logic for `none` should just be a combination of the existing initiator/terminator logic? For example `initiator` supports a missing terminator when empty, `terminator` supports a missing initiator when empty, so `none` just wants to support missing both when empty. I wonder if we don't need these new properties, and we can instead tweek/combine the existing logic to make both optional instead of one or the other?

I'm not sure how we currently implement things though, or maybe those other policies are broken too?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/DelimiterUnparsers.scala
tuxji: Good point.  The whole function is about 60 lines long and the return occurs 10 lines into the function's body.  The rest of the function would have to be wrapped in an if-block, or the function refactored to call sub-functions which do the rest of the work.  I'm fine with the return if others agree.

tuxji (changes requested):
Steve also noticed many other good points, so I suspect some more changes may be needed.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/DelimiterUnparsers.scala
mbeckerle: Return is "maintenance unstable" in scala.  You modify the code at some point and an exception gets involved and then the return ends up causing real problems. 

Best to remove them. They don't compose well with exception throw/catch/try or Try objects.

mike-mcgann:
> I'm not sure how we currently implement things though, or maybe those other policies are broken too?

I started with the original ticket, placed that test case in (in the form that I thought it wanted), verified that the error was the one described in the ticket, and worked from there trying to get the test to pass. I didn't find an obvious place in the code where these other cases were being handled so I focused on getting the 'none' case to work first. Once that was working, I then suspected that the other cases may not be handled since I didn't run across them. I added those additional cases as tests and some of them didn't work. I then made some additional changes to get those to pass. 

I just pulled the latest from main and copied over the new tests there to see what the results would be. Only the 'both' test passed--the 'initiator', 'terminator', and 'none' tests failed. So I don't think the emptyValueDelimiterPolicy got put in or in a way that was exercised by the tests. 

Also, I don't have a grasp on the details of the parsing at the moment so I'm making most of my changes through trial and error and seeing what that fixes and what that breaks.

stevedlawrence:
Sounds like maybe `emptyValueDelimiterPolicy` is just completely broken/not implemented then? We might want to consider how to best implement each of these instead of doing one at a time, there might be a very clean solution.

My first insinct is that maybe we need a new combinator? So a new parser that accepts three sub parsers (initiator, content, and terminator). It parses each of these allowing some to fail depending on the `emptyValueDelimiterPolicy` and the whether the parsed content was empty.

And the Unparse combinator would look at the value of content, and if empty then it unparses the initiator/terminator based on the policy.

This way delimiters don't need to know if they should be optional or not, and they don't have to worry about the content being empty or not. They just do what they do, and if they fail, the combinator that controls them handles it accordingly.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/LocalElementMixin.scala
```diff
@@ -84,6 +85,10 @@ trait LocalElementMixin extends ParticleMixin with LocalElementGrammarMixin {
       else if (
         isDefaultable && !hasTerminator && emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.Terminator
       ) true
+      else if (
+        hasInitiator && hasTerminator && emptyValueDelimiterPolicy =:= EmptyValueDelimiterPolicy.None
```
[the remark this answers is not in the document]
mike-mcgann: This was the only place in the code where I found something working with the emptyValueDelimiterPolicy directly. The negated value of this function gets set as "hasKnownRequiredSyntax" so I put this change in to see what would happen. This was to express the case that if there is both a initiator and a terminator but the policy is none, then it does not have a known required syntax. This change ended up having no effect and as I was tracing through the code, it didn't seem to be used in a way related to delimiter handling. I ended up looking elsewhere but forgot to remove this from the code. I just removed it and the tests still pass so it still had no effect.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
mike-mcgann: This is for the case where the data is `John,(Jacob),Doe` when present and `John,,Doe` when not present having an initiator of `(` and a terminator of `)` for the middle name. When the middle name was not present, Daffodil was raising an error as described in the ticket: `Parse Error: Found out of scope delimiter: ',' ','`. When that block is reached, the existing comment says "It was a remote delimiter but we should have found a local one". I put a breakpoint here and in this case it found a comma while it was looking for the parens. In this case, where the delimiters should be absent when the value is empty, parsing can continue. That seemed to work but I'm not sure if that is 100% correct.

mike-mcgann:
> My first insinct is that maybe we need a new combinator? So a new parser that accepts three sub parsers (initiator, content, and terminator). It parses each of these allowing some to fail depending on the `emptyValueDelimiterPolicy` and the whether the parsed content was empty.

This makes more sense because it was getting a bit confusing trying to figure out how to make these changes with the current structure. I'll take a look and implementing it this way but I may have a few questions on how to do this properly.

stevedlawrence:
I wonder if we should just say that `emptyValueDelimiterPolicy` must be `both` (I think our current behavior) and plan to implement it in a future release? Implementing this propery could be considered implementing a new DFDL feature, and the focus for 3.5.0 is bug fixes. We also don't have any formats that needing this property.

We may want to change DAFFODIL-2205 so that it isn't so much about fixing value with "none", but to just implement `emptyValueDelimiterPolicy` values other than "both"?

mike-mcgann:
Ok, I'm closing this PR for now and updating the ticket with this information.

PR #982, opened 2023-03-08
Ensure all primitives use textNumberPattern and infinfity/NaN correctly

stevedlawrence:
Currently, if the primitive type is an integer then text number parsing disallows parsing decimal points, even if the pattern contains a decimal point. Instead, when parsing integers, we should allow decimals as long the fractional part is zero. And when unparsing, we should unparse a decimal point with a zero fractional part according to the pattern.

This changes the behavior so integer parsing uses the same DecimalFormat configuration as non-integer parsing (i.e. decimals are allowed), but we throw a parse error if the fractional part that was parsed is non-zero. This also means that unparsing integers now outputs decimal points according to the pattern.

Additionally, if textNumberCheckPolicy is strict, we enable ICU setDecimalPatternMatchRequired to true so that we allow or disallow decimal points in the data depending on if the pattern does or does not have a decimal point. Note that lax parsing always allows decimal points regardless of the pattern. For this reason, we now always require the grouping/decimal separator DFDL properties in lax mode.

One bug was discovered in ICU (ICU-22303) where if we require the decimal point due to strict mode enabled, then ICU never parses the infinity/NaN representation. A workaround is added to manually check for these representations until this bug is fixed. ICU unit tests are also added which should fail if ICU fixes this bug so we can remove this workaround.

Make sure we always specify infinity and NaN representations from the DFDL schema for all primitives, not just for xs:double/xs:float. There is no way to disable infinity/NaN ICU parsing, so when if we do not specify these values ICU just uses the locale values, which could lead to unwanted locale specific behavior. Related, this modifies NodeInfo types so that fromNumber fails for types that do not support infinity/NaN (i.e. everything except Double/Float) and creates a parse error.

Modifies virtual decimal logic to ensure we handle cases for numbers that do not fit in a Long (should work) or contain decimal points (should be a parse error).

Tests are updated so if they want to differentiate between int and decimal depending on if a decimal exists in the data, then they must specify a pattern with or without a decimal and enable strict mode--lax mode allows a decimal regardless of type so cannot differentiate the types.

DAFFODIL-2158

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ConvertTextStandardNumberParser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ConvertTextStandardNumberParser.scala
@@ -20,6 +20,7 @@ package org.apache.daffodil.runtime1.processors.parsers
 import java.lang.{ Double => JDouble, Float => JFloat, Long => JLong, Number => JNumber }
 import java.math.MathContext
 import java.math.{ BigDecimal => JBigDecimal }
+import java.math.{ BigInteger => JBigInteger }
 import java.text.ParsePosition
 import scala.util.matching.Regex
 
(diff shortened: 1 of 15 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ConvertTextStandardNumberParser.scala
```diff
@@ -171,47 +185,91 @@ case class ConvertTextStandardNumberParser(
...
-            )
-            return
+            d
           }
           case bd: ICUBigDecimal => {
-            // sometimes ICU will return their own custom BigDecimal, even if the
-            // value could be represented as a BigInteger. We only want Java types,
-            // so detect this and convert it to the appropriate type
+            // ICU will return their own custom BigDecimal if the value cannot fit in a Long and
+            // isn't infinity/NaN. We only want Java types, so detect this and convert it to the
```
tuxji, line 239: Asking only for curiosity.  Why does ICU return a custom BigDecimal instead of a BigDecimal?  Why does the difference matter to Daffodil, that is, why isn't the custom BigDecimal compatible enough?

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
```diff
@@ -4039,16 +4081,13 @@
       <tdml:documentPart type="text"><![CDATA[NOTANUMBER]]]></tdml:documentPart>
     </tdml:document>
 
-    <tdml:warnings>
-      <tdml:warning>DFDL property was ignored</tdml:warning>
-      <tdml:warning>textStandardNaNRep</tdml:warning>
-    </tdml:warnings>
+    <tdml:warnings />
```
tuxji, line 4084: Can we just delete the empty tdml:warnings element without changing any semantics?

tuxji (approved):
+1

Looks good although needs review by @mbeckerle.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ConvertTextStandardNumberParser.scala
stevedlawrence: > Asking only for curiosity. Why does ICU return a custom BigDecimal instead of a BigDecimal? 

Why does ICU do anything? It's a mystery wrapped inside an enigma. We just do our best to make it work :)

> Why does the difference matter to Daffodil, that is, why isn't the custom BigDecimal compatible enough?

I'm not really sure what's different and it's not obvious reading the ICU BigDecimal documentation. But we try to use Java types everywhere (avoiding both Scala types and custom ICU types) just so we're consistent.

daffodil-test/src/test/resources/org/apache/daffodil/section13/text_number_props/TextNumberProps.tdml
stevedlawrence: The empty `tdml:warnings` explicitly says "there should be no warnings", whereas not having a `tdml:warnings` element just says "we don't care if there are warnings or not". The idea with changing it like this was that the test used to expect a warning saying this property was ignored, but the property is now used so I wanted to make sure it's not ignored and there's no warning. We do get an error mentioning "NaN" so that should be enough to make it clear the property is used, but I wanted to still keep the intention of the test regarding warnings.

mbeckerle: I wasn't aware that empty tdml:warnings meant that. Does that also work for empty tdml:validationErrors?

I think there is a JIRA ticket [about adding this feature for expecting/precluding warnings.](https://issues.apache.org/jira/browse/DAFFODIL-864) Perhaps this feature exists now.

stevedlawrence: Looks like I was confused. It looks like we have that logic for `tdml:validationErrors` but not `tdml:warnings`. So you can ensure there are no validation errors, but can't do the same for warnings.

So this change does nothing at the moment.

tuxji: @stevedlawrence, you can leave the empty `tdml:warnings` since it might gain the expected semantics later on.

@mbeckerle and @mike-mcgann, this PR needs a second +1 from one of you.

mbeckerle (approved):
+1

PR #983, opened 2023-03-09
Allow appinfo annotations that contain no-namespace XML

stevedlawrence:
The anyOther group is intended to allow any XML content at all. However, the namespace="##other" only allows XML that has a namespace that isn't DFDL. If it doesn't have a namespace at all, it is not allowed.

This changes anyOther to be a choice of any with either ##other or ##local to allow either namespaced or non-namespaced annotations.

DAFFODIL-2800


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #984, opened 2023-03-09
Move lint checks to a separate job

stevedlawrence:
This allows lint checks and tests to run separately, making is easier to distinguish between failed tests or just minor cleanup issues. Should also allow the CI tests to finish 1-2 minutes quicker on average.

All lint checks are run even if other lint checks fail, which should help to find all issues in a single CI run.

DAFFODIL-2799

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,14 +38,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distribution: [ temurin ]
+        java_distribution: [ temurin ]
         java_version: [ 8, 11, 17 ]
         scala_version: [ 2.12.17 ]
         os: [ ubuntu-22.04, windows-2022 ]
         include:
           - os: macos-12
             shell: bash
-            distribution: temurin
+            java_distribution: temurin
             java_version: 17
             scala_version: 2.12.17
             env_cc: cc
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -198,8 +188,55 @@ jobs:
...
+    steps:
+
+      ############################################################
+      # Setup
+      ############################################################
+
+      - name: Check out Repository
+        uses: actions/checkout@v3.3.0
+        with:
+          fetch-depth: 0
```
tuxji, line 213: Can you add a comment why we're checking out with zero fetch-depth?

.github/workflows/main.yml
```diff
@@ -198,8 +188,55 @@ jobs:
...
+      - name: Run Rat Check
+        if: success() || failure()
+        run: $SBT ratCheck || (cat target/rat.txt; exit 1)
+
+      - name: Run OSGI Check
+        if: success() || failure()
+        run: $SBT osgiCheck
+
+      - name: Run scalafmt Check
+        if: success() || failure()
```
tuxji, line 235: Nice use of `success() || failure()` instead of `always()`.  If a user or concurrency: group cancels the run, steps with `always()` still would be run but steps with `success() || failure()` will be skipped.

tuxji (approved):
+1

.github/workflows/main.yml
stevedlawrence: It took some digging to figure out the reason why (our above comment just says to improve reporting, but I wasn't sure what that meant). I found out it's about sonar scanning, briefly documented here:

https://github.com/SonarSource/sonarcloud-github-action

And detailed reasons here:

https://community.sonarsource.com/t/git-fetch-depth-implications/75260

We don't run sonar scans in the lint checker (sonar scan uses coverage reports so needs to be in the main CI job), so I've removed the `fetch-depth: 0` here. I've also updated the above comment to mention sonar scan as the reason for it.

.github/workflows/main.yml
stevedlawrence: Yeah, using `always()` feels like the obvious choice, but the `success() || failure()` idiom is recommended in the GitHub Actions documentation for the reason you mention:

https://docs.github.com/en/actions/learn-github-actions/expressions#always

Seems like `always` should be used very rarely in favor of this `success() || failure()` idiom.

PR #985, opened 2023-03-09
Throw exceptions when scala-xml cannot continue

mike-mcgann:
The error handlers for SAX parsing accumulate fatal errors into a list and allowing parsing to continue. In a certain case with a malformed XML document, the same error is being continiously reported by scala-xml causing an infinite loop. This change now throws an exception in this case to allow processing to continue.

[DAFFODIL-2586](https://issues.apache.org/jira/browse/DAFFODIL-2586)

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilConstructingLoader.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilConstructingLoader.scala
@@ -171,7 +171,17 @@ class DaffodilConstructingLoader private[xml] (
 
   override def reportSyntaxError(pos: Int, msg: String): Unit = {
     val exc = makeSAXParseException(pos, msg)
-    errorHandler.error(exc)
+    errorHandler.fatalError(exc)
+    if (msg == "'<' not allowed in attrib value") {
+      // DAFFODIL-2586
+      // There is a bug in scala-xml which causes an infinite loop when
+      // this error condition is reached. The loop expects the scanner
+      // to advance but in this case it does not. When this error is
+      // seen, an exception needs to be thrown to stop processing.
+      //
+      // See: https://github.com/scala/scala-xml/blob/v2.1.0/shared/src/main/scala/scala/xml/parsing/MarkupParserCommon.scala#L67-L72
+      throw exc
+    }
   }
 
   /*
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/UserSubmittedTests.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/UserSubmittedTests.tdml
@@ -245,4 +245,13 @@
 			<ex:r2>1</ex:r2>
 		</tdml:dfdlInfoset></tdml:infoset>
 	</tdml:unparserTestCase>
+
+	<tdml:parserTestCase name="test_DFDL_2586" root="root" model="invalid_xml.dfdl.xsd"
+						 description="FIXME">
+		<tdml:document>1</tdml:document>
+		<tdml:errors>
+			<tdml:error>not allowed in attrib value</tdml:error>
+		</tdml:errors>
+	</tdml:parserTestCase>
+
 </tdml:testSuite>
--- a/daffodil-test/src/test/resources/org/apache/daffodil/usertests/invalid_xml.dfdl.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/usertests/invalid_xml.dfdl.xsd
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+           targetNamespace="http://example.com">
+
+  <xs:annotation>
+    <xs:appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format
+      textBidi="no"
+      separatorSuppressionPolicy="trailingEmpty"
+      floating="no"
+      encodingErrorPolicy="replace"
+      outputNewLine="%CR;%LF;"
+      leadingSkip="0"
+      trailingSkip="0"
+      alignment="1"
+      alignmentUnits="bytes"
+      textPadKind="none"
+      textTrimKind="none"
+      truncateSpecifiedLengthString="no"
+      escapeSchemeRef=""
+      representation="binary"
+      encoding="ASCII"
+      separator = ""
+      initiator = ""
+      terminator = ""
+      ignoreCase = "yes"
+      sequenceKind="ordered"
+      initiatedContent="no"
+      fillByte="%SP;"
+      lengthKind="implicit"
+      byteOrder="littleEndian
+  />
+    </xs:appinfo>
+  </xs:annotation>
+
+  <xs:element name="root" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="1"/>
+
+</xs:schema>
--- a/daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestUserSubmittedTests.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestUserSubmittedTests.scala
@@ -26,7 +26,6 @@ object TestUserSubmittedTests {
   val testDir = "/org/apache/daffodil/usertests/"
   val runner = Runner(testDir, "UserSubmittedTests.tdml")
   val runner2 = Runner(testDir, "nameDOB_test.tdml")
-
   @AfterClass def shutDown(): Unit = {
     runner.reset
     runner2.reset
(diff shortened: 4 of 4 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDisallowDocType.scala
```diff
@@ -87,7 +86,7 @@ class TestDisallowDocType {
   }
 
   @Test def test_infosetFileMustNotHaveDocType(): Unit = {
-    val e = intercept[TDMLException] {
+    val e = intercept[SAXParseException] {
```
stevedlawrence, line 89: Seems we shouldn't let SAXParseExceptions bubble  to the very top? I think the way the DaffodilConstructorLoader is intended to work is it doesn't throw exceptions but instead gathers up all errors in the error handler diagnostics to we can report them later?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLSchemaFile.scala
```diff
@@ -92,12 +92,12 @@ class DFDLSchemaFileLoadErrorHandler(schemaFileLocation: SchemaFileLocation)
...
-   * Called on a fatal exception. The parser/validator throws the exception after
-   * this call returns.
+   * Called on a fatal exception and throws the exception
    */
-  def fatalError(exception: SAXParseException) = error(
-    exception,
-  ) // same as non-fatal exception.
+  def fatalError(exception: SAXParseException) = {
+    loaderErrors_ :+= exception
+    throw exception
```
stevedlawrence, line 99: Reading the javadoc for fatalError, it seems like we shouldn't need to throw an exception here, and that the parser should throw the exception after calling fatalError if it is unable to continue.

If this is needed to workaround a bug in Scala-xml that's fine, but it should be documented why it's needed. Maybe scala-xml expects fatalError implementations to throw the exceptions?

mike-mcgann: I was reading the SE8 documentation for ErrorHandler and it has this paragraph:

>For XML processing errors, a SAX driver must use this interface in preference to throwing an exception: it is up to the application to decide whether to throw an exception for different types of errors and warnings. Note, however, that there is no requirement that the parser continue to report additional errors after a call to [fatalError](https://docs.oracle.com/javase/8/docs/api/org/xml/sax/ErrorHandler.html#fatalError-org.xml.sax.SAXParseException-). In other words, a SAX driver class may throw an exception after reporting any fatalError. Also parsers may throw appropriate exceptions for non-XML errors. For example, [XMLReader.parse()](https://docs.oracle.com/javase/8/docs/api/org/xml/sax/XMLReader.html#parse-org.xml.sax.InputSource-) would throw an IOException for errors accessing entities or the document.

From that I concluded that if you want to stop execution on a fatal error, an exception should be thrown in `fatalError` since it said the driver *may* throw an exception. I don't think the driver is required to throw an exception in this case and there is flexibility for it to throw one or not.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDisallowDocType.scala
mike-mcgann: It still needs to stop execution on the fatal error here or it will get caught in the infinite loop. I originally had fatalError throwing an SDE but then I found out the TDML runner installs its own custom error handler. So, to mimic the same behavior I tried to throw an SDE in that runner but it appears that the concrete implementations of those are in runtime1 and not lib. That is why I left it as a SAXParseException. Should the TDML runner catch any unhanded SAXParseExceptions and wrap them in TDMLExceptions or is there a better way?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLSchemaFile.scala
stevedlawrence: > it up to the application to decide whether to throw an exception for different types of errors and warnings. 

This makes me think that we (the application) shouldn't need to throw an exception to get the behavior we need from Xerces (the driver). The behavior we want to just add warnings/errors/fatalErrots to a list and then to let the driver continue gathering errors if it can do so.

And no where does it say the fatalError implementation *must* throw the exception. If that's actually needed because of a bug in Xerces/Scala XML/whatever is calling fatalError requires that behavior, we should document that.

> In other words, a SAX driver class may throw an exception after reporting any fatalError.

This is what I would expect. Xerces calls fatal error, we handle it and add the error to our list, and then Xerces throws the exception if can't make progress OR continues calling error/fatalError more if it can. By throwing an exception, we are telling Xerces to stop even if it could continue, which I don't think is what we want--we want as many errors from Xerces as possible.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDisallowDocType.scala
stevedlawrence: >  Should the TDML runner catch any unhanded SAXParseExceptions and wrap them in TDMLExceptions or is there a better way?

It probably depends where the errors are coming from. If the errors are from compiling a DFDL schema used by a TDML test, those error should all end up as `Diagnositc`s that the TDML runner can look for. If they are coming from something else (e.g. TDMLRunner validating a .tdml file), then it probably wants to be caught and turned into a TDML exception.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLSchemaFile.scala
mike-mcgann: This is needed to workaround what appears to be a bug in scala-xml and its trying to stop it because it is getting caught in that loop. Yes, I wouldn't expect that behavior and a comment can be added to explain what is going on. I was just saying that the driver may continue as much as it wants and doesn't have to throw an exception when the error handler is installed. 

As an alternative, we could always keep around the last error seen. If the next error is exactly the same as the last error, no progress is being made, and we could throw the exception at that point. That way we can get as many errors as possible without triggering the infinite loop.

stevedlawrence: > As an alternative, we could always keep around the last error seen...

It's probably not worth the added complexity, I think just commenting about why we have the throw an limitations (e.g. maybe don't get as many errors as possible) is fine.

Also, I grepped through the scala-xml code and couldn't find any calls to `fatalError`. I wonder if under the hood it's just using Xerces or some Java built-in XML parser, and the issue is actually in that parser and not in scala-xml. Can you output a stack trace? It would be nice to open a ticket with whatever is doing this and see if it could be fixed.

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDisallowDocType.scala
stevedlawrence: I wonder if the problem of the leaking exception is here:

https://github.com/apache/daffodil/blob/b1a5ab06c3bd194ecd43f37789654c89c83beb49/daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilConstructingLoader.scala#L477

When parsing fatalError will be called which throws an exception. That is likely caught in this catch block, but then calls fatalError which will throw that exception. Not only does this mean we'll see the exception twice, it means the exception leaks out. It probably also means the exception is duplicated in the diagnostics, unless the exception was thrown by scala-xml without calling fatalError.

mike-mcgann: I found the root cause of the issue. Daffodil was working fine. It is a bug in scala-xml. In this specific case, the infinite loop is caused when the scanner does not advance when this error is reported. There is evidence of error correction elsewhere so I think this was simply overlooked. My proposed solution is to now only throw an exception when that specific error is raised. Yes, looking at the error message is a bit hackish, but I'm thinking that it is a good compromise where it can continue as much as possible without causing the infinite loop. We have a test now and if the message changes, it should be caught.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -273,7 +273,7 @@ class DFDLTestSuite private[tdml] (
     }
 
     def fatalError(exception: SAXParseException) = {
-      error(exception)
+      loadingExceptions += exception
```
mike-mcgann, line 276: I'll change this back to the original for the final commit

daffodil-test/src/test/scala/org/apache/daffodil/section00/general/TestDisallowDocType.scala
stevedlawrence: Do you think the fix in scala-xml would be fairly straightforward? It would be nice if we could submit a PR so we can eventually get rid of the special casing, since I agree it does sound like a bit of a hack.

mike-mcgann: The code for that is as follows:

```scala
  /**
   * attribute value, terminated by either `'` or `"`. value may not contain `<`.
   * @param endCh either `'` or `"`
   */
  def xAttributeValue(endCh: Char): String = {
    val buf = new StringBuilder
    while (ch != endCh && !eof) {
      // well-formedness constraint
      if (ch == '<') reportSyntaxError("'<' not allowed in attrib value")
      else if (ch == SU) truncatedError("")
      else buf append ch_returning_nextch
    }
    ch_returning_nextch
    // @todo: normalize attribute value
    buf.toString
  }
```

I think that when a `<` character is seen, the error recovery needs to be a call to `ch_returning_nextch`  without appending it to `buf`. So I think it is probably straightforward but I'm not sure. The call to `truncatedError` doesn't seem to handle that either. Maybe there is some other expectation here. It might be worth time boxing a fix attempt. If a working PR can be prepared in less than X hours, great, otherwise drop the attempt.

mike-mcgann: Do we want to keep this explicit check in for now and remove it if, and when, a fix gets merged into scala-xml? I'll try to get a fix in over there but there is no guarantee on when that could happen.

stevedlawrence (approved):
+1, seems a bit hack-y, but hopefully scala-xml fixes it and releases and we can remove this shortly. It's also on in an error condition that should hopefully be unlikely.

stevedlawrence:
Taking a second look at things and at the PR you created for scala-xml to fix the underlying issue, it doesn't look like the infinite loop issue is specific to having `<` characters in an attribute like your `msg` check looks for. Could other illegal characters cause this issue? I wonder if we need to check for a more generic error message not specific to `<`?

mike-mcgann:
> Taking a second look at things and at the PR you created for scala-xml to fix the underlying issue, it doesn't look like the infinite loop issue is specific to having `<` characters in an attribute like your `msg` check looks for. Could other illegal characters cause this issue? I wonder if we need to check for a more generic error message not specific to `<`?

That PR isn't for this exact issue. I decided to put in a PR for a similar issue that had an existing bug report. If that goes well, then I'll file another bug report for this issue and a PR along with it. I wanted to test the waters first.

mike-mcgann:
I also accidentally pushed to the wrong branch and just did a force push to revert that.

tuxji (approved):
+1

PR #987, opened 2023-03-13
Evaluate assert/discriminator expressions after groupContent

jadams-tresys:
According to 9.5 of the DFDL spec assert and discriminators with expressions should be processed after the content of their enclosing sequence, group, or choice. Before these expressions were always being processed before the content.

This commit also moves the setVariable expression evaluaiton to the correct place, which is before the enclosing group.

DAFFODIL-1971, DAFFODIL-1590

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
@@ -213,10 +213,11 @@ trait ProvidesDFDLStatementMixin extends ThrowsSDE with HasTermCheck {
     st.testKind != TestKind.Pattern
   }
 
-  final lazy val patternStatements: Seq[DFDLStatement] = patternAsserts ++ patternDiscrims
+  // Discriminator statements must be before asserts
+  final lazy val patternStatements: Seq[DFDLStatement] = patternDiscrims ++ patternAsserts
 
-  final lazy val lowPriorityStatements: Seq[DFDLStatement] =
-    setVariableStatements ++ nonPatternAsserts ++ nonPatternDiscrims
+  final lazy val assertDiscrimExpressionStatements: Seq[DFDLStatement] =
+    nonPatternDiscrims ++ nonPatternAsserts
 
   final protected lazy val localStatements = this.annotationObjs.collect {
     case st: DFDLStatement => st
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
@@ -21,6 +21,7 @@ import org.apache.daffodil.core.compiler.ForParser
 import org.apache.daffodil.core.compiler.ForUnparser
 import org.apache.daffodil.core.dsom.*
 import org.apache.daffodil.lib.exceptions.Assert
+import org.apache.daffodil.runtime1.processors.parsers.AssertExpressionEvaluationParser
 import org.apache.daffodil.runtime1.processors.parsers.NadaParser
 import org.apache.daffodil.runtime1.processors.parsers.SeqCompParser
 import org.apache.daffodil.runtime1.processors.unparsers.SeqCompUnparser
@@ -83,14 +84,27 @@ class SeqComp private (context: SchemaComponent, children: Seq[Gram])
   protected final override def close = ")"
 
   lazy val parserChildren =
-    children.filter(_.forWhat != ForUnparser).map { _.parser }.filterNot {
-      _.isInstanceOf[NadaParser]
-    }
+    children
+      .filter(_.forWhat != ForUnparser)
+      .map { _.parser }
+      .filterNot {
+        _.isInstanceOf[NadaParser]
+      }
+      .toArray
+
+  lazy val assertExpressionChildren = parserChildren.filter {
+    _.isInstanceOf[AssertExpressionEvaluationParser]
+  }.toArray
 
   final override lazy val parser = {
     if (parserChildren.isEmpty) new NadaParser(context.runtimeData)
     else if (parserChildren.length == 1) parserChildren.head
-    else new SeqCompParser(context.runtimeData, parserChildren.toArray)
+    else
+      new SeqCompParser(
+        context.runtimeData,
+        parserChildren,
+        assertExpressionChildren
+      )
   }
 
   lazy val unparserChildren = {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
@@ -227,26 +227,40 @@ abstract class CombinatorParser(override val context: RuntimeData)
   extends Parser
   with CombinatorProcessor
 
-final class SeqCompParser(context: RuntimeData, val childParsers: Array[Parser])
-  extends CombinatorParser(context) {
+final class SeqCompParser(
+  context: RuntimeData,
+  val childParsers: Array[Parser],
+  testAssert: Array[Parser]
+) extends CombinatorParser(context) {
   override def runtimeDependencies = Vector()
   override def childProcessors = childParsers.toVector
 
   override def nom = "seq"
 
-  val numChildParsers = childParsers.size
+  val optDiscrimParser = childParsers.collectFirst {
+    case ae: AssertExpressionEvaluationParser if (ae.discrim) => ae
+  }
+  val nonDiscrimChildren = childParsers.diff(optDiscrimParser.toSeq)
 
   def parse(pstate: PState): Unit = {
-    var i: Int = 0
-    while (i < numChildParsers) {
-      val parser = childParsers(i)
-      parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
-        return
+    var i = 0
+    val numNonDiscrimChildren = nonDiscrimChildren.size
+
+    // Handle all non discriminator child parsers first
+    while ((i < numNonDiscrimChildren) && (pstate.processorStatus eq Success)) {
+      nonDiscrimChildren(i).parse1(pstate)
       i += 1
     }
-  }
 
+    // If a discriminator statement exists always parse it, even if there was a prior failure.
+    // See section 9.5.2 of the DFDL specification.
+    if (optDiscrimParser.isDefined) {
+      if (pstate.processorStatus eq Success)
+        optDiscrimParser.get.parse1(pstate)
+      else
+        pstate.withTempSuccess(optDiscrimParser.get.parse1)
+    }
+  }
 }
 
 class ChoiceParser(ctxt: RuntimeData, val childParsers: Array[Parser])
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
@@ -16,16 +16,16 @@
   limitations under the License.
 -->
 
-<tdml:testSuite 
-  suiteName="Discriminator" 
+<tdml:testSuite
+  suiteName="Discriminator"
   description="Section 7 - Discriminator"
   xmlns:ex="http://example.com" xmlns="http://example.com"
-  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData" 
+  xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/" 
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
   xmlns:xs="http://www.w3.org/2001/XMLSchema"
   xmlns:fn="http://www.w3.org/2005/xpath-functions"
-  defaultRoundTrip="true" 
+  defaultRoundTrip="true"
   defaultValidation="on">
 
   <tdml:defineSchema name="discrimAssertPlacement">
@@ -209,7 +209,7 @@
         <xs:element name="e2" type="xs:int" dfdl:lengthKind="delimited" />
       </xs:sequence>
     </xs:group>
-    
+
     <xs:group name="namedGroup2">
       <xs:sequence>
         <xs:element name="e3" type="xs:string"
@@ -231,7 +231,7 @@
         </xs:group>
       </xs:complexType>
     </xs:element>
-    
+
     <xs:element name="Item2" dfdl:lengthKind="implicit">
       <xs:complexType>
         <xs:choice>
@@ -292,7 +292,7 @@
   <!--
     Test name: discrimOnSimpleType
     Schema: s0, root e0
-    Purpose: This test demonstrates the use of a discriminator on a simple 
+    Purpose: This test demonstrates the use of a discriminator on a simple
     type. The document meets the discriminator and should pass.
   -->
 
@@ -309,7 +309,7 @@
   <!--
     Test name: discrimOnGroupRef
     Schema: s0, root Item
-    Purpose: This test demonstrates the use of a discriminator on a group 
+    Purpose: This test demonstrates the use of a discriminator on a group
     reference. The document meets the discriminator and should pass.
   -->
 
@@ -326,11 +326,11 @@
       </tdml:dfdlInfoset>
     </tdml:infoset>
   </tdml:parserTestCase>
-  
+
   <!--
     Test name: discrimOnGroupRef2
     Schema: s0, root Item
-    Purpose: This test demonstrates the use of a discriminator on a group 
+    Purpose: This test demonstrates the use of a discriminator on a group
     reference. The document meets the discriminator and should pass.
   -->
 
@@ -351,7 +351,7 @@
   <!--
     Test name: discrimOnElementRef
     Schema: s0, root Item_01
-    Purpose: This test demonstrates the use of a discriminator on an element 
+    Purpose: This test demonstrates the use of a discriminator on an element
     reference. The document meets the discriminator and should pass.
   -->
 
@@ -402,7 +402,7 @@
   <!--
     Test name: discriminatorGuidesChoice2
     Schema: choice1, root c1
-    Purpose: The element should fail, and the discriminator should still 
+    Purpose: The element should fail, and the discriminator should still
     be checked. The error message should express
     the discriminator failure.
   -->
@@ -684,7 +684,7 @@
       <tdml:error>Schema Definition Error</tdml:error>
     </tdml:errors>
   </tdml:parserTestCase>
-  
+
   <tdml:defineSchema name="discrimExpression">
     <xs:include schemaLocation="/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
     <dfdl:format ref="ex:GeneralFormat" />
@@ -714,8 +714,8 @@
               <xs:sequence>
                 <xs:annotation>
                   <xs:appinfo source="http://www.ogf.org/dfdl/">
-                    <!-- NOTE That this discriminator executes regardless of whether the 
-                    sequence content (int element) below parses successfully. So if that int parse fails, 
+                    <!-- NOTE That this discriminator executes regardless of whether the
+                    sequence content (int element) below parses successfully. So if that int parse fails,
                     the discriminator still will be true, and we will NOT backtrack to have zero occurrences
                     of the ints array. -->
                     <dfdl:discriminator testKind="expression" test='{ fn:true() }' />
@@ -730,30 +730,79 @@
       </xs:complexType>
     </xs:element>
 
+    <xs:element name="e4">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="ints" dfdl:occursCountKind="implicit" minOccurs="0" maxOccurs="5">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:annotation>
+                  <xs:appinfo source="http://www.ogf.org/dfdl/">
+                    <!-- NOTE That this discriminator executes regardless of whether the
+                    sequence content (int element) below parses successfully. So with the discriminator
+                    set to false we will fail to parse any int's and instead parse everything as a
+                    string -->
+                    <dfdl:discriminator testKind="expression" test='{ fn:false() }' />
+                  </xs:appinfo>
+                </xs:annotation>
+                <xs:element name="int" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="1" dfdl:lengthUnits="bytes" />
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="string" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="3" dfdl:lengthUnits="bytes" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="e5">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="ints" dfdl:occursCountKind="implicit" minOccurs="0" maxOccurs="5">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:annotation>
+                  <xs:appinfo source="http://www.ogf.org/dfdl/">
+                    <!-- NOTE That this discriminator executes regardless of whether
+                    the sequence content (int element) below parses successfully. In
+                    this test the discriminator references a successfully parsing
+                    element in the partial infoset of an otherwise failing sequence body. -->
+                    <dfdl:discriminator testKind="expression" test='{ ex:int1 eq 1 }' />
+                  </xs:appinfo>
+                </xs:annotation>
+                <xs:element name="int1" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="1" dfdl:lengthUnits="bytes" />
+                <xs:element name="int2" type="xs:int" dfdl:lengthKind="explicit" dfdl:length="1" dfdl:lengthUnits="bytes" />
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="string" type="xs:string" dfdl:lengthKind="explicit" dfdl:length="3" dfdl:lengthUnits="bytes" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
   </tdml:defineSchema>
-  
+
   <tdml:parserTestCase name="discrimExpression_01"
     root="e1" model="discrimExpression"
     description="Property Name: test - DFDL-7-076R">
     <tdml:document><![CDATA[123]]></tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
-        <e1>123</e1> 
+        <e1>123</e1>
       </tdml:dfdlInfoset>
     </tdml:infoset>
   </tdml:parserTestCase>
-  
+
   <tdml:parserTestCase name="discrimExpression_02"
     root="e2" model="discrimExpression"
     description="Property Name: test - DFDL-7-076R">
     <tdml:document>cats</tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
-        <e2>cats</e2> 
+        <e2>cats</e2>
       </tdml:dfdlInfoset>
     </tdml:infoset>
   </tdml:parserTestCase>
-  
+
   <tdml:parserTestCase name="discrimExpression_03"
     root="e2" model="discrimExpression"
     description="Property Name: test - DFDL-7-076R">
@@ -765,13 +814,67 @@
 
   <tdml:parserTestCase name="discrimExpression_04"
     root="e3" model="discrimExpression"
-    description="Property Name: test - DFDL-7-076R">
+    description="Verify that a discriminator will prevent backtracking
+      regardless of whether the sequence content parses succesfully">
+    <tdml:document><![CDATA[abc]]></tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>Failed to populate ex:ints</tdml:error>
+      <tdml:error>Unable to parse xs:int from text: a</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="discrimExpression_05"
+    root="e4" model="discrimExpression"
+    description="Verify that a failed discriminator will be handled correctly
+    when the body of the sequence parses successfully">
+    <tdml:document><![CDATA[123]]></tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <e4>
+          <string>123</string>
+        </e4>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="discrimExpression_06"
+    root="e4" model="discrimExpression"
+    description="Verify correct behavior when the body of a sequence fails and
+    the discriminator fails">
     <tdml:document><![CDATA[abc]]></tdml:document>
     <tdml:infoset>
       <tdml:dfdlInfoset>
-        <e3>
+        <e4>
           <string>abc</string>
-        </e3>
+        </e4>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="discrimExpression_07"
+    root="e5" model="discrimExpression"
+    description="Sequence body fails, but discriminator succeeds and references
+    elements in partial body infoset">
+    <tdml:document><![CDATA[1ab]]></tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>Failed to populate ex:ints</tdml:error>
+      <tdml:error>Unable to parse xs:int from text: a</tdml:error>
+      <tdml:error>Schema context: ex:int2</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="discrimExpression_08"
+    root="e5" model="discrimExpression"
+    description="Sequence body fails and so does the discriminator that
+    references partial sequence body infoset">
+    <tdml:document><![CDATA[2ab]]></tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <e5>
+          <string>2ab</string>
+        </e5>
       </tdml:dfdlInfoset>
     </tdml:infoset>
   </tdml:parserTestCase>
@@ -964,7 +1067,7 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
-    
+
     <xs:element name="r2">
       <xs:complexType>
         <xs:sequence>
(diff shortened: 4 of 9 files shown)


jadams-tresys:
This is primarily WIP as it needs more test coverage and I have been informed that I need to hold off on working on this for a little while.

It is fully functional, just needs more tests for edge cases surrounding  evaluating expressions after the content of sequences, groups, and choices.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
```diff
@@ -217,7 +217,7 @@ trait ProvidesDFDLStatementMixin extends ThrowsSDE with HasTermCheck {
   final lazy val patternStatements: Seq[DFDLStatement] = patternAsserts ++ patternDiscrims
 
   final lazy val lowPriorityStatements: Seq[DFDLStatement] =
```
stevedlawrence, line 219: Any idea where "low priority" comes from? I'm not familiar with that as a DFDL term. Wondering if something like "expressionStatements" would be more clear?

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
```diff
@@ -89,10 +90,19 @@ class SeqComp private (context: SchemaComponent, children: Seq[Gram])
...
+
   final override lazy val parser = {
     if (parserChildren.isEmpty) new NadaParser(context.runtimeData)
     else if (parserChildren.length == 1) parserChildren.head
-    else new SeqCompParser(context.runtimeData, parserChildren.toVector)
+    else
+      new SeqCompParser(
+        context.runtimeData,
+        parserChildren.toVector,
+        assertExpressionChildren.toVector,
```
stevedlawrence, line 104: I believe the `paserChildren` vector contains the `assertExpressionChildren` parsers, which I think means the assert parsers will be evaluated twice? Once when all the `parserChildren` are evaluated and again after the parserChildren finis and then we evaluate the asserts Parsers? Seems like parserChildren should not contain the assert children?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -241,8 +244,14 @@ final class SeqCompParser(context: RuntimeData, val childParsers: Vector[Parser]
     while (i < numChildParsers) {
       val parser = childParsers(i)
       parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
+      if (pstate.processorStatus ne Success) {
+        if (testAssert.nonEmpty) {
+          testAssert.foreach { ap =>
+            pstate.withTempSuccess(ap.parse1)
+          }
+        }
```
stevedlawrence, line 252: Ah, I guess I now see why it's okay for `childrenParsers` to also include the assertion parsers. If all the non-assert parsers succeed then we'll don't evaluate the assertion parsers.

However, I think there is still an issue. Say all the non-assertion parsers succeed, then we start evaluating the assertion parsers, and assume one of them fails. The `pstate.processorStatus ne Success` will be true, and then we'll evaluate all the assertion parsers, including ones we've already evaluated.

So I think this approach still feels like it needs a tweak. Almost feels like the assert parsers need to be completely separate from the other parsers or something.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
```diff
@@ -588,6 +589,23 @@ final class PState private (
...
+   * assert expression needs to be parsed regardless of whether or not the
+   * enclosing sequence content parses or not as the assert expression may be
+   * used as a discriminator for the choice branch.
+   */
+  def withTempSuccess(func: (PState) => Unit): Unit = {
+    val priorProcessorStatus = processorStatus
+    setSuccess()
+    func(this)
+    if (processorStatus eq Success)
+      _processorStatus = priorProcessorStatus
```
stevedlawrence, line 606: Conditionally resetting back to the previous status feels like it might be confusing? Feels like if this is temporarily ignoring status it should always reset back to whatever the status was before. Maybe the way this wants to work is it it returns the status of `func`, and the caller can choose to do with it whatever they want? Something more like

```scala
def withTempSuccess(func: (PState) => Unit): ProcessorResult = {
  val priorStatus = processorStatus
  setSuccess()
  func(this)
  val funcStatus = processorStatus
  _processorStatus = priorStatus
  funcStatus
}
```

Also, Is there any value in passing in a lambda? An alternative would be to just pass in a `Parser`, and then instead of calling `func(this)`, it could call `parser.parse1(this)`. That ensures that this is always called with a `Parser` instead of a function that just happens to accept a `PState`, which is probably safer?

stevedlawrence (changes requested):
I *think* the same assert could be evaluated multiple times?

Also, have you run this against the regression suite? I wonder if there are any schemas that rely on the current behavior and could mess up how they use discriminators?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -241,8 +244,14 @@ final class SeqCompParser(context: RuntimeData, val childParsers: Vector[Parser]
     while (i < numChildParsers) {
       val parser = childParsers(i)
       parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
+      if (pstate.processorStatus ne Success) {
+        if (testAssert.nonEmpty) {
+          testAssert.foreach { ap =>
```
stevedlawrence, line 249: Just had another thought about this, this will evaluate all assertions even if one fails. Is that the correct behavior, or should it stop after the first assertion failure? Seems like if one assertion fails it shouldn't continue to evaluate other assertions?

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
```diff
@@ -89,10 +90,19 @@ class SeqComp private (context: SchemaComponent, children: Seq[Gram])
       _.isInstanceOf[NadaParser]
     }
 
+  lazy val assertExpressionChildren = parserChildren.filter {
+    _.isInstanceOf[AssertExpressionEvaluationParser]
```
stevedlawrence, line 94: Reading the spec some more, it says:

> an attempt to evaluate a discriminator MUST be made even if preceding statements or the parse of the schema component ended in a Processing Error. 

I *think* this implies asserts should not be evaluated if the preceding statements fails? Which kindof makes sense since an assert only causes backtracking and does not discriminate points of uncertainty?

So it feels like the logic needs to differentiate between asserts and discriminators? Both should be evaluated at the end, but only discriminators should be evaluated if the prior parses succeed?

Another question, are discriminators always evaluated before asserts? Or are they evaluated depending on the order defined in the DFDL schema? For example, if you have:

```xml
  <dfdl:assert test="..." />
  <dfdl:discriminator test="..." />
  <dfdl:assert test="..." />
</xs:appinfo>
```

Should that evaluate the first assert, then the discriminiator, then second assert. I don't know if the spec clarifies that or if it's implied, but I think we always evaluate asserts before discrims regardless of how they are defined in the schema?

jadams-tresys:
I've now run it against the regression test suite and I am seeing failures in dfdl-edifact, dfdl-gif, and dfdl-jpeg2000 that I don't see in main.  I'll need to spend some time looking into those to confirm that they are relying on the old behavior.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
jadams-tresys: Changed to expressionStatements

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
jadams-tresys: I believe I have fixed this.  I didn't realize (or forgot since starting this pull request a year ago) that we internally handle discriminator expressions as an assert expression just with a discriminator flag set.

I've changed it so that now we are filtering out just the discriminator statements from the rest of the sequence child parsers and run those after the rest of the sequence body.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
jadams-tresys: I think I Have addressed this with the changes I mentioned above by separating out the discriminator statements and running them after the rest of the sequence body.  Nothing should be getting run twice anymore.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/PState.scala
jadams-tresys: I modeled withTempSuccess after the withPointOfUncertainty function, so that is why it is using a labmda here.  I vaguely recall tasking with @mbeckerle about this function, but I could be mis-remembering things.

I'll make the change to not be conditionally changing the processStatus.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
jadams-tresys: This is a good question.  The way I've implemented it we will perform all dfdl:assert statements first and then do all dfdl:discriminator statements, regardless of whether or not any of the other statements or parsers have failed.

mbeckerle: I am raising this issue about the spec with the DFDL workgroup. (We have a call tomorrow). 
The spec is completely silent about order of evaluation of multiple asserts in a single location of the schema, whether they are done in schema order, whether the first failure stops that evaluation, etc. 

The spec does say that the discriminators are evaluated last, even if there are other failures (to parse, or assertions). This allows for them to be optimized and executed earlier, when they might have been evaluated to true before the parser got a failure. The spec also prevents there from being more than one discriminator at any annotation point.

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
```diff
@@ -765,13 +814,66 @@
...
+    elements in partial body infoset">
+    <tdml:document><![CDATA[1ab]]></tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>Failed to populate ex:ints</tdml:error>
+      <tdml:error>Unable to parse xs:int from text: a</tdml:error>
+    </tdml:errors>
+  </tdml:parserTestCase>
+
+  <tdml:parserTestCase name="discrimExpression_08"
```
mbeckerle, line 867: I'm looking for a test where the sequence body fails, the discriminator succeeds and references things in the sequence body that in fact exist and are correct, so the behavior is as if the parse failed after the discrminator already evaluated to true. 

The use case that motivates all this complexity is a bunch of record types, each a sequence of child elements, each record type has a discriminator that looks at one or a few of the fields of the record to decide if the data is in fact of this record type. Those discriminators *could* be hoisted and evaluated as soon as the fields of the record needed by the expressions have been parsed. If parsing of other later fields of the record then fails, the discriminator has already been set true, so those failures don't cause other record types to be attempted. Rather the whole choice fails. 

This is very close conceptually to a choice by dispatch. Once the dispatch key has been computed and the branch selected, a subsequent failure does not cause backtracking of that choice, but a failure of the whole choice.

mbeckerle (changes requested):
I think the discriminators should be evaluated before any asserts. 

Are we missing an important test case? 

Also I want to be clear that if a discriminator fails then no asserts are evaluated. If any assert fails no subsequent asserts are evaluated.

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
jadams-tresys: Is this not what discrimExpression_07 covers?
It has a discriminator expression that references the first element of a sequence (which does parse correctly).  The 2nd element of the sequence then fails, which causes a parse error instead of back tracking.

jadams-tresys:
@mbeckerle I think I've addressed all of your comments except the missing test case.  Can you double check that discrimExpression_07 doesn't cover the scenario you were describing?

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
```diff
@@ -765,13 +814,66 @@
...
+
+  <tdml:parserTestCase name="discrimExpression_07"
+    root="e5" model="discrimExpression"
+    description="Sequence body fails, but discriminator succeeds and references
+    elements in partial body infoset">
+    <tdml:document><![CDATA[1ab]]></tdml:document>
+    <tdml:errors>
+      <tdml:error>Parse Error</tdml:error>
+      <tdml:error>Failed to populate ex:ints</tdml:error>
+      <tdml:error>Unable to parse xs:int from text: a</tdml:error>
```
mbeckerle, line 863: Add a check that the message contains the element name "int2" since that is the one that should be causing the failure.

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
mbeckerle: Yes. Verified. This does appear to cover it. I added one comment about tightening up the error message since the error should be talking specifically about the int2 field being the one that failed.

mbeckerle (approved):
+1 with one minor improvement to a test case

daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
jadams-tresys: Done

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,47 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
   override lazy val runtimeDependencies = Vector()
   override def childProcessors = childParsers
 
   override def nom = "seq"
 
-  val numChildParsers = childParsers.size
+  val assertDiscrimExpressions = childParsers.collect { case ae: AssertExpressionEvaluationParser => ae }
+  val discrimExpressions = assertDiscrimExpressions.filter { _.discrim }
+  val assertExpressions = assertDiscrimExpressions.filterNot { _.discrim }
+  val nonAssertChildren = childParsers.diff(assertExpressions)
```
stevedlawrence, line 242: I think `nonAssertChildren` includes discriminator parsers so could discriminators get evaluated twice, once in the first loop and once in the discrimnator loop?

stevedlawrence (commented):
I think discirminators will be evaluated twice, also a a little concerned about possible regressions if schemas rely on this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mbeckerle: There's a base class shared between discriminators and regular asserts. They're both a kind of assertion. This scala code may be using that terminology so nonAssertChildren may also mean no discriminators.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,47 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
     }
-  }
 
+    // Handle all non discriminator assert statements last
+    i = 0
+    while ((i < assertExpressions.size) && (pstate.processorStatus eq Success)) {
+        assertExpressions(i).parse1(pstate)
+        i += 1
+    }
+  }
```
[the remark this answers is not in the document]
mbeckerle: Yes. Confirmed in a lengthly workgroup discussion yesterday in fact.

daffodil-test/src/test/scala/org/apache/daffodil/section07/discriminators/TestDiscriminators.scala
```diff
@@ -79,7 +79,11 @@ class TestDiscriminators {
   @Test def test_discrimExpression_03() = { runner.runOneTest("discrimExpression_03") }
 
   // DAFFODIL-1971
-  // @Test def test_discrimExpression_04() = { runner.runOneTest("discrimExpression_04") }
+  @Test def test_discrimExpression_04() = { runner.runOneTest("discrimExpression_04") }
+  @Test def test_discrimExpression_05() = { runner.runOneTest("discrimExpression_05") }
+  @Test def test_discrimExpression_06() = { runner.runOneTest("discrimExpression_06") }
+  @Test def test_discrimExpression_07() = { runner.runOneTest("discrimExpression_07") }
+  @Test def test_discrimExpression_08() = { runner.runOneTest("discrimExpression_08") }
```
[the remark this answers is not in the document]
mbeckerle: If this is for the 3.10.0 release, we have to be prepared to back it out should it lead to any significant regressions. 

Changing order of assertion evaluation, and discriminator evaluation (because a failing discriminator is equivalent to a failing assertion) can result in different diagnostics coming out, so tests may regress because they're expecting one error text, but are given another. But negative tests failing due to one diagnostic vs. another is a sort of lower-threat regression vs. positive tests that break.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
jadams-tresys: @stevedlawrence is right.  It should be

```    val nonAssertChildren = childParsers.diff(assertDiscrimExpressions)```

then there won't be duplicates.  Will change.

File: daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala [hunk 12]
```diff
@@ -227,26 +227,47 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
+
+    // Handle all non assert/discriminator child parsers first
+    while ((i < nonAssertChildren.size) && (pstate.processorStatus eq Success)) {
+        nonAssertChildren(i).parse1(pstate)
+        i += 1
+    }
+
+    // Handle all discriminator expressions statements after sequence body
+    i = 0
+    while (i < discrimExpressions.size) {
```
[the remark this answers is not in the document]
jadams-tresys: Is it not possible to have multiple discriminators?  I had just assumed that it was but am having trouble finding something definitive in the spec.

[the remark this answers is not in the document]
mbeckerle: Only one discriminator at each annotation point. In fact I'm not sure we enforce this, but one annotation point is supposed to have only a discriminator OR asserts (1 or more), but not both:

Section 7.6: "The resolved set of statement annotations for a schema component can contain only a single dfdl:discriminator or one or more dfdl:assert annotations, but not both. To clarify: dfdl:assert annotations and dfdl:discriminator annotations are exclusive of each other. It is a Schema Definition Error otherwise."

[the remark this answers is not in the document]
jadams-tresys: I just verified that we already are enforcing the exclusivity of both allowing only one dfdl:discriminator or multiple dfdl:assert statements.  Doing otherwise creates an SDE.

With that in mind I will make some changes to use these expectations to simplify things.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,34 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
-    var i: Int = 0
-    while (i < numChildParsers) {
-      val parser = childParsers(i)
-      parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
-        return
-      i += 1
+    var i = 0
+
+    // Handle all non assert/discriminator child parsers first
```
stevedlawrence, line 245: This comment should say "all non discriminator child parsers", removing "assert". This will evaluate asserts if they exist.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,34 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
+  val childParsers: Vector[Parser],
+) extends CombinatorParser(context) {
   override lazy val runtimeDependencies = Vector()
   override def childProcessors = childParsers
 
   override def nom = "seq"
 
-  val numChildParsers = childParsers.size
+  val optDiscrimParser = childParsers.collectFirst { case ae: AssertExpressionEvaluationParser if (ae.discrim) => ae }
+  val nonDiscrimChildren = if (optDiscrimParser.isDefined) childParsers.diff(Seq(optDiscrimParser.get)) else childParsers
```
stevedlawrence, line 240: I think this could be simplified to
```scala
val nonDiscrimChildren = childParsers.diff(optDiscrimParser.toSeq)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,34 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
-    while (i < numChildParsers) {
-      val parser = childParsers(i)
-      parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
-        return
-      i += 1
+    var i = 0
+
+    // Handle all non assert/discriminator child parsers first
+    while ((i < nonDiscrimChildren.size) && (pstate.processorStatus eq Success)) {
```
stevedlawrence, line 246: It might be worth storing `nonDiscrimChildren.size` as a member val? I assume it's a constant time function so calling it in the loop probably doesn't matter, but `numChildParser` used to be a member val so there might have been a reason for that?

stevedlawrence (commented):
Just minor cleanup

jadams-tresys:
After digging into some of the failures I was seeing in the dfdl-regression test suite, there wasn't a straightforward work around for the few schemas that were relying on the old discriminator behavior.  I am going to hold off on merging this until after the release of Daffodil 3.10.0 and then this can be added to 4.0.0 and a better solution for these problematic schemas can be investigated.

jadams-tresys:
Now that 3.10 has been released and 4.0.0 is underway, any objections to merging these changes now?

stevedlawrence:
Might want to hold off until all the 2.13 changes are in since that's kindof touching everything and might make dealing with merge conflicts difficult?

Also, breaking backwards compatibility in 4.0.0 is less of a concern than usualy, but we should still have backwards compatibility notes, especially explaining how to update schemas that relied on the old behavior. Did you ever come up with a generic way to update schemas to work with this change? I thought one of the issues preventing merging in 3.10 was there wasn't really a good way to do that?

mbeckerle:
At least at first I'd like 4.0.0 branch to have a scala 2.13 or scala 3 version (or one then the other) which are 100% identical functionality and performance to 3.10.0. I want them to run the exact same schemas unmodified just for testing reasons. 

I would suggest these changes should be in 4.1.0, or a 3.11.0 if we truly need them out sooner.

jadams-tresys:
Fair enough.  I'll re-evaluate once the 2.13 changes seem to be wrapped up.

As for a fix for the failing schemas, it seemed like the correct fix was to add choiceDispatch or initiated content to some of these choice branches that were mistakenly being allowed to backtrack.  I know we didn't want to force a change like that in 3.10.0, but I wasn't sure if a change like that would be OK moving into 4.0.0.

stevedlawrence:
In theory, not all formats can use choice dispatch or initiated content and might still have this behavior. We might need to think about what an alternative would look like for those kinds of schemas.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
stevedlawrence: > I am raising this issue about the spec with the DFDL workgroup.

@mbeckerle @jadams-tresys, was there ever clarification on what the correct behavior is?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,38 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
 
-final class SeqCompParser(context: RuntimeData, val childParsers: Array[Parser])
-  extends CombinatorParser(context) {
-  override def runtimeDependencies = Vector()
+final class SeqCompParser(
+  context: RuntimeData,
+  val childParsers: Array[Parser],
+  testAssert: Array[Parser],
+) extends CombinatorParser(context) {
+  override lazy val runtimeDependencies = Vector()
```
stevedlawrence: All runtimeDependencies are currently defs. Is there a reason this was changed to lazy val?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
```diff
@@ -213,10 +213,10 @@ trait ProvidesDFDLStatementMixin extends ThrowsSDE with HasTermCheck {
     st.testKind != TestKind.Pattern
   }
 
-  final lazy val patternStatements: Seq[DFDLStatement] = patternAsserts ++ patternDiscrims
+  final lazy val patternStatements: Seq[DFDLStatement] = patternDiscrims ++ patternAsserts
```
mbeckerle, line 217: Add a comment "// disc must be before asserts." 
Since this was presumably changed because this invariant is depended upon elsewhere.

stevedlawrence (approved):
+1 to the latest rebase

I'm not sure if we ever got clarification on the correct behavior of evaluating multiple asserts + discriminator, but I'm not sure it should block the change, it feels pretty rare to mix and max asserts + discrims so I don't think having to change it later would be a big deal.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
mbeckerle: Discussion in the DFDL Workgroup is here: https://github.com/OpenGridForum/DFDL/issues/67

The upshot is that the order is "implementation defined", so we need to pick, but also state the behavior. 
I believe the right behavior is "schema order", and "first failure stops the evaluation of all dfdl:assert statements" (but not discriminators, as they may be needed to resolve an enclosing PoU.

mbeckerle (changes requested):

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
jadams-tresys: Nope, I think when I wrote this originally they were all lazy vals and I just missed that once all the rebasing was done.  I'll change it to a def

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,25 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
-
   def parse(pstate: PState): Unit = {
-    var i: Int = 0
-    while (i < numChildParsers) {
-      val parser = childParsers(i)
-      parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
-        return
-      i += 1
+    childParsers.foreach { cp =>
```
stevedlawrence: I think we intentionally avoided childParsers.foreach since I think we found that allocated lambda expression and showed up in profiling. I'm not sure if scala3 changed that at all, but we should probably stick what the for loop without profiling to figure it out

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,25 @@ abstract class CombinatorParser(override val context: RuntimeData)
   extends Parser
   with CombinatorProcessor
 
-final class SeqCompParser(context: RuntimeData, val childParsers: Array[Parser])
-  extends CombinatorParser(context) {
+final class SeqCompParser(
+  context: RuntimeData,
+  val childParsers: Array[Parser],
+  testAssert: Array[Parser]
```
stevedlawrence, line 233: This isn't used anymore, I think it can be removed.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
```diff
@@ -83,14 +84,27 @@ class SeqComp private (context: SchemaComponent, children: Seq[Gram])
...
-    }
+    children
+      .filter(_.forWhat != ForUnparser)
+      .map { _.parser }
+      .filterNot {
+        _.isInstanceOf[NadaParser]
+      }
+      .toArray
+
+  lazy val assertExpressionChildren = parserChildren.filter {
```
stevedlawrence, line 95: Can be removed also, the one parser this is passed into doesn't use this anymore.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
```diff
@@ -213,10 +213,11 @@ trait ProvidesDFDLStatementMixin extends ThrowsSDE with HasTermCheck {
...
   }
 
-  final lazy val patternStatements: Seq[DFDLStatement] = patternAsserts ++ patternDiscrims
+  // Discriminator statements must be before asserts
+  final lazy val patternStatements: Seq[DFDLStatement] = patternDiscrims ++ patternAsserts
 
-  final lazy val lowPriorityStatements: Seq[DFDLStatement] =
-    setVariableStatements ++ nonPatternAsserts ++ nonPatternDiscrims
+  final lazy val assertDiscrimExpressionStatements: Seq[DFDLStatement] =
+    nonPatternDiscrims ++ nonPatternAsserts
```
stevedlawrence, line 220: I think this loses the order that asserts and discriminators are defined, always evaluating discriminators before asserts. I think if we are to obey the order they are defined in the schema we really can't separate discriminators and asserts, we kindof have to treat them as one group of things.

That said, the checkDiscriminatorsAssertsDisjoint function below has this check:
```scala
    schemaDefinitionUnless(
      asserts == Nil || discrims == Nil,
      "Cannot have both dfdl:discriminator annotations and dfdl:assert annotations at the same location."
    )
```

So it seems like we can't actually have assertions and discriminators on the same element, so it's impossible for order to matter? We either get a single discriminator or multiple asserts, but not both?

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,25 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
-      parser.parse1(pstate)
-      if (pstate.processorStatus ne Success)
-        return
-      i += 1
+    childParsers.foreach { cp =>
+      if (pstate.processorStatus.isSuccess) cp.parse1(pstate)
+      else cp match {
+        case ae: AssertExpressionEvaluationParser if (ae.discrim) => pstate.withTempSuccess(ae.parse1)
+        case _ => // processorStatus is Failure, do not execute non-discim child parsers
+      }
```
stevedlawrence: One potential issue with this approach is we don't short circuit failures. For example, say there are 10000 parsers associated with this sequence and the first one fails--we now have to iterate over all the remaining parsers and ignore them just in case there is a discriminator at the end. Ideally on error this would skip to the discriminator parser, evaluate that (assuming it hasn't already been evaluated), and then end.

That said, considering I just realized that we don't even support asserts + discrims on the same element right now, then order doesn't matter and your previous approach is probably better since it does short circuit failures. I think we can remove the latest fixup commit and we'll get the correct behavior more efficiently.

We'll need careful changes here and tests if we ever do support asserts + discrims on the same element, but I don't expect that to happen anytime soon.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/Grammar.scala
stevedlawrence: I think I've discovered that we don't even support asserts + discrims on the same element. I agree this is probably the right behavior to use schema order, but until we support asserts + discrims (if ever) I don't think we need to consider this case.

stevedlawrence (approved):
+1, I think the previous approach is actually what we want. The new fixup commit is a bit less efficient and Daffodil doesn't allow what it was intended to support

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mbeckerle: > One potential issue with this approach is we don't short circuit failures. For example, say there are 10000 parsers associated with this sequence and the first one fails--we now have to iterate over all the remaining parsers and ignore them just in case there is a discriminator at the end. Ideally on error this would skip to the discriminator parser, evaluate that (assuming it hasn't already been evaluated), and then end.

This is not right. In a given sequence, a discriminator expressed lexically at the start of that sequence, that is, on that sequence's annotation, is to be evaluated logically at the end of the sequence. But further, the discriminator is supposed to be evaluated even if the sequence fails somewhere. 

But note: This does *not* apply recursively. A discriminator expressed on or inside one of the children of this sequence, does NOT count for this. Only a discriminator expressed on that sequence's annotation point (it could come from a group reference, or the sequence group itself), is evaluated even in the case of failure. 

So there is no walk of the 1000+ sequence children. This is a local action by the sequence combinator which just has to evaluate the discriminator on the unwind of the parse, even if the parse failed. 

Note also, this applies only to discriminators with testKind="expression".

stevedlawrence: My comment was in regards to the changes that Josh made that have since been reverted. If I recall, the change was essentially something like

```scala
childParsers.foreach { p =>
  if (state.isSuccess || p.isDiscriminator ) {
    p.parse(state)
  }
}
```

So if one of the early `p.parse()` failed it would still iterate over every child parser even though it only needs to evaluate the discriminator.

I don't tthink the code in the current PR does that.

daffodil-core/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -227,26 +227,38 @@ abstract class CombinatorParser(override val context: RuntimeData)
...
+    val numNonDiscrimChildren = nonDiscrimChildren.size
+
+    // Handle all non discriminator child parsers first
+    while ((i < numNonDiscrimChildren) && (pstate.processorStatus eq Success)) {
+      nonDiscrimChildren(i).parse1(pstate)
       i += 1
     }
-  }
 
+    if (optDiscrimParser.isDefined) {
```
mbeckerle, line 257: Comments are needed here to motivate this final execution of the optDiscrimParser even if there was a prior failure. 
Cite the DFDL v1.0 spec section.

mbeckerle (approved):
+1

PR #988, opened 2023-03-15
Update parse error message

mike-mcgann:
When there are not enough bits for parsing, an error message will be displayed showing the number of bits available. Sometimes this value uses the number of bits remaining in the limit and doesn't consider that this could be greater than the number of bits actually available in the data stream. This change selects the lesser of the two values when that information is known.

The TDML runner has also been updated to only place a limit on the parser when the length of the document under test doesn't end on a byte boundary.

[DAFFODIL-2645](https://issues.apache.org/jira/browse/DAFFODIL-2645)

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
@@ -128,6 +128,15 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
    */
   def hasReachedEndOfData: Boolean = inputSource.hasReachedEndOfData
 
+  /**
+   * Return the number of currently available bytes.
+   *
+   * This should not be used to determine the length of the data, as more bytes
+   * may become available in the future. This should really only be used for
+   * debug or diagnostic purposes.
+   */
+  def knownBytesAvailable: Long = inputSource.knownBytesAvailable
+
   def setBitPos0b(newBitPos0b: Long): Unit = {
     // threadCheck()
     Assert.invariant(newBitPos0b >= 0)
@@ -789,7 +798,7 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
     // need to call areBytesAvailable first to ensure at least length bytes are
     // buffered if they exist
     val available = inputSource.areBytesAvailable(nBytesRequested)
-    val bytesToRead = if (available) nBytesRequested else inputSource.bytesAvailable.toInt
+    val bytesToRead = if (available) nBytesRequested else inputSource.knownBytesAvailable.toInt
     val array = new Array[Byte](bytesToRead)
     inputSource.get(array, 0, bytesToRead)
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
@@ -20,7 +20,6 @@ package org.apache.daffodil.runtime1.processors.parsers
 import java.nio.file.Files
 import java.nio.file.StandardOpenOption
 
-import org.apache.daffodil.lib.util.MaybeULong
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
 import org.apache.daffodil.runtime1.processors.LengthInBitsEv
 
@@ -37,10 +36,11 @@ trait ByteChunkWriter { self: PrimParser =>
     writeChunk: (Array[Byte], Int) => Unit,
   ): Unit = {
     val dis = start.dataInputStream
+    val startLoc = start.currentLocation
+    val startSfl = start.schemaFileLocation
     var remainingBitsToGet = nBits
 
     val array = new Array[Byte](start.tunable.blobChunkSizeInBytes)
-
     // Tunable restrictions ensure blobChunkSizeInBits still fits in an int.
     // All the places where toInt is called below can be no larger than this
     // value, and so have no issues with potential overflow.
@@ -54,14 +54,7 @@ trait ByteChunkWriter { self: PrimParser =>
         writeChunk(array, bytesToPut.toInt)
         remainingBitsToGet -= bitsToGet
       } else {
-        val remainingBits =
-          if (dis.remainingBits.isDefined) {
-            val totalBitsRead = nBits - remainingBitsToGet
-            MaybeULong(dis.remainingBits.get + totalBitsRead)
-          } else {
-            MaybeULong.Nope
-          }
-        self.PENotEnoughBits(start, nBits, remainingBits)
+        PENotEnoughBits(start, startSfl, startLoc, nBits, start.dataInputStream)
         remainingBitsToGet = 0 // break out of the loop
       }
     }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/HexBinaryLengthParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/HexBinaryLengthParsers.scala
@@ -52,7 +52,7 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
       // create and fill the byte array
       val dis = start.dataInputStream
       if (!dis.isDefinedForLength(nBits)) {
-        PENotEnoughBits(start, nBits, dis.remainingBits)
+        PENotEnoughBits(start, nBits, dis)
       } else {
         val array = start.dataInputStream.getByteArray(nBits.toInt, start)
         currentElement.setDataValue(array)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
@@ -18,13 +18,13 @@
 package org.apache.daffodil.runtime1.processors.parsers
 
 import org.apache.daffodil.io.BacktrackingException
+import org.apache.daffodil.io.InputSourceDataInputStream
 import org.apache.daffodil.lib.api.DataLocation
 import org.apache.daffodil.lib.api.Diagnostic
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.exceptions.SchemaFileLocation
 import org.apache.daffodil.lib.util.Maybe.One
-import org.apache.daffodil.lib.util.MaybeULong
-import org.apache.daffodil.lib.util.Misc
+import org.apache.daffodil.lib.util.{ MaybeULong, Misc }
 import org.apache.daffodil.runtime1.dsom.RuntimeSchemaDefinitionError
 import org.apache.daffodil.runtime1.processors.CombinatorProcessor
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
@@ -65,10 +65,49 @@ sealed trait Parser extends Processor {
     pstate.setFailed(new ParseError(One(sfl), One(dataLoc), s, args: _*))
   }
 
-  def PENotEnoughBits(pstate: PState, neededBits: Long, remainingBits: MaybeULong) = {
-    val remainingStr =
-      if (remainingBits.isDefined) s" but found only ${remainingBits.get} available" else ""
-    PE(pstate, "Insufficient bits in data. Needed %d bit(s)%s.", neededBits, remainingStr)
+  def PENotEnoughBits(
+    pstate: PState,
+    sfl: SchemaFileLocation,
+    dataLoc: DataLocation,
+    neededBits: Long,
+    source: InputSourceDataInputStream,
+  ): Unit = {
+    val startPos = dataLoc.bitPos1b - 1
+    val remainingLimitedBits = {
+      if (source.bitLimit0b.isEmpty) MaybeULong.Nope
+      else {
+        val lim = source.bitLimit0b.get
+        Assert.invariant(lim >= 0)
+        val nBits = lim - startPos
+        MaybeULong(nBits)
+      }
+    }
+    val remainingBits = {
+      if (source.hasReachedEndOfData) {
+        val bitsAvailable = {
+          val fragmentBitsReadFromSourcePos = source.bitPos0b % 8
+          val bitsAvailableFromSourcePos =
+            source.knownBytesAvailable * 8 - fragmentBitsReadFromSourcePos
+          val bitsBetweenStartPosAndSourcePos = source.bitPos0b - startPos
+          val bitsAvailableFromStartPos =
+            bitsAvailableFromSourcePos + bitsBetweenStartPosAndSourcePos
+          bitsAvailableFromStartPos
+        }
+        if (remainingLimitedBits.isEmpty || bitsAvailable < remainingLimitedBits.get)
+          bitsAvailable
+        else
+          remainingLimitedBits.get
+      } else remainingLimitedBits.get
+    }
+
+    PE(
+      pstate,
+      sfl,
+      dataLoc,
+      "Insufficient bits in data. Needed %d bit(s) but found only %d available",
+      neededBits,
+      remainingBits,
+    )
   }
 
   def PENotEnoughBits(
@@ -76,17 +115,27 @@ sealed trait Parser extends Processor {
     sfl: SchemaFileLocation,
     dataLoc: DataLocation,
     neededBits: Long,
-    remainingBits: MaybeULong,
-  ) = {
-    val remainingStr =
-      if (remainingBits.isDefined) s" but found only ${remainingBits.get} available" else ""
+  ): Unit = {
     PE(
       pstate,
       sfl,
       dataLoc,
-      "Insufficient bits in data. Needed %d bit(s)%s.",
+      "Insufficient bits in data. Needed %d bit(s)",
       neededBits,
-      remainingStr,
+    )
+  }
+
+  def PENotEnoughBits(
+    pstate: PState,
+    neededBits: Long,
+    source: InputSourceDataInputStream,
+  ): Unit = {
+    PENotEnoughBits(
+      pstate,
+      pstate.schemaFileLocation,
+      pstate.currentLocation,
+      neededBits,
+      source,
     )
   }
 
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
@@ -1138,7 +1138,7 @@ class Main(
                         if (loc.bitLimit0b.isDefined) {
                           (loc.bitLimit0b.get - loc.bitPos0b).toString
                         } else {
-                          "at least " + (inStream.inputSource.bytesAvailable * 8)
+                          "at least " + (inStream.inputSource.knownBytesAvailable * 8)
                         }
                       Logger.log.error(
                         s"Left over data after consuming 0 bits while streaming. Stopped after consuming ${loc.bitPos0b} bit(s) with ${remainingBits} bit(s) remaining.",
@@ -1176,7 +1176,7 @@ class Main(
                       dumpString
                     } else ""
                     val curBytePosition1b = inStream.inputSource.position + 1
-                    val bytesAvailable = inStream.inputSource.bytesAvailable
+                    val bytesAvailable = inStream.inputSource.knownBytesAvailable
                     val bytesLimit = math.min(8, bytesAvailable).toInt
                     val destArray = new Array[Byte](bytesLimit)
                     val destArrayFilled = inStream.inputSource.get(destArray, 0, bytesLimit)
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/DataInputStreamImplMixin.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/DataInputStreamImplMixin.scala
@@ -18,7 +18,6 @@
 package org.apache.daffodil.io
 
 import org.apache.daffodil.lib.exceptions.Assert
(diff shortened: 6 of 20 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
```diff
@@ -55,9 +55,9 @@ trait ByteChunkWriter { self: PrimParser =>
         remainingBitsToGet -= bitsToGet
       } else {
         val remainingBits =
-          if (dis.remainingBits.isDefined) {
+          if (dis.remainingLimitedBits.isDefined) {
             val totalBitsRead = nBits - remainingBitsToGet
-            MaybeULong(dis.remainingBits.get + totalBitsRead)
+            MaybeULong(dis.remainingLimitedBits.get + totalBitsRead)
```
tuxji, line 60: Please add a comment explaining why out of 26 places calling `remainingBits`, you decided to change this particular place to call `remainingLimitedBits` instead of `remainingBits`.  The `remainingLimitedBits` function returns only the remaining bits until the limit is reached, so you're losing the information that the remaining bits until the end of the input is reached might be even smaller.

tuxji (approved):
+1

One change (noted below) looks strange and needs a comment explaining why it needs to call that particular function (`remainingLimitedBits` instead of `remainingBits`).  Other than that, the changes look good.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
stevedlawrence: Yeah, I think this might actually be incorrect? This use of `remainingBits` here is to calculate a parameter for `PENotEnoughBits`. And since this is for an error message, like the other uses of `remainingBits`, I think we do want to take into account EOF, and so should use `remainingBits`.

Which I think means there are no more uses of `remainingLimitedBits`, and maybe it's logic can just be merge into `remainingBits`?

Another thought, nothing should really use `reaminingBits` to make parsing decision (we don't currently, but we could accidentally in the future). Parsers should always use `isDefinedForLength` if they want to know if bits are available, since that takes into account both bit limit and EOF, and will also read and cache bytes from the input stream as needed. The only place it is reasonable to use is in error messages when `isDefinedForLength` returns false, which is how it's used right now (and it's not used in unparse at all). Since it really shouldn't be used, and is only used for not enough bits diagnostics, I wonder if we can just remove it and move the logic into `PENotEnoughBits` available, or at least rename it so it's clear it should only be used for diagnostics when `isDefinedForLength` fails? This would avoid us accidentally using it for actual parsing decisions.

mike-mcgann: The old `remainingBits` function which only considers end-of-limit got renamed to `remainingLimitedBits`. The new function that returns the length to the end-of-limit or end-of-stream now uses the name of `remainingBits`. I originally considered `reminaingLimitedBits` and `remainingStreamBits` but I didn't think that worked well. So this change isn't really a change to functionality but a refactoring to use the new name.

To fix this bug, I initially found the two locations that output the error message and threw a `RuntimeException` if the order of the numbers was backwards. I updated all instances where the error message was called because it seemed like getting the length to end-of-limit or end-of-file was correct when reporting an error message. 

In this case, it is doing more than reporting an error, so I decided to leave the functionality as-is.

stevedlawrence: > In this case, it is doing more than reporting an error, so I decided to leave the functionality as-is.

I think it is reporting an error, it's just using slightly fancier logic to take into account read data since the blob parsers reads data in chunks.

mike-mcgann: I renamed `bytesAvailable` to `knownBytesAvailable` to help indicate that this is not necessarily a known value and should not be relied on in parsing situations. The `remainingBits`/`remainingLimitBits` have been refactored out and that logic is now in the error reporting function.

tuxji:
> I renamed bytesAvailable to knownBytesAvailable to help indicate that this is not necessarily a known value and should not be relied on in parsing situations. The remainingBits/remainingLimitBits have been refactored out and that logic is now in the error reporting function.

+1

I've reviewed the pull request again and I like the most recent changes.  Much better now.

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
```diff
@@ -129,23 +129,14 @@ final class InputSourceDataInputStream private (val inputSource: InputSource)
...
-   * The number of bits remaining if known. If a limit has not been set,
-   * returns Nope. If a limit has been set, the number of bits available
-   * until the limit as long as the end of data has not been seen. If the
-   * end of data has been seen, the smaller value of the remaining bits
-   * in the limit and the remaining bits in the data stream.
+   * Return the number of currently available bytes.
+   *
+   * This should not be used to determine the length of the data, as more bytes
+   * may become available in the future. This should really only be used for
+   * debug purposes.
```
stevedlawrence, line 136: Might say "for debug or diagnostic purposes", just to make it clear it is safe to use in things like PEs.

Might also be worth mention that this does not take into account the bitLimit just so it's clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -65,7 +66,27 @@ sealed trait Parser extends Processor {
...
+        Assert.invariant(lim >= 0)
+        val nBits = lim - source.bitPos0b
+        MaybeULong(nBits)
+      }
+    }
+
+    val remainingBits = {
+      if (source.hasReachedEndOfData) {
+        val bitsAvailable = source.knownBytesAvailable * 8
+        if (remainingLimitedBits.isDefined && bitsAvailable < remainingLimitedBits.get)
```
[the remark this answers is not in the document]
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -65,7 +66,27 @@ sealed trait Parser extends Processor {
...
+        val lim = source.bitLimit0b.get
+        Assert.invariant(lim >= 0)
+        val nBits = lim - source.bitPos0b
+        MaybeULong(nBits)
+      }
+    }
+
+    val remainingBits = {
+      if (source.hasReachedEndOfData) {
+        val bitsAvailable = source.knownBytesAvailable * 8
```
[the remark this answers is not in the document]
mike-mcgann: Updated and I added an additional test for this condition

daffodil-io/src/main/scala/org/apache/daffodil/io/InputSourceDataInputStream.scala
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
```diff
@@ -54,14 +53,7 @@ trait ByteChunkWriter { self: PrimParser =>
...
       } else {
-        val remainingBits =
-          if (dis.remainingBits.isDefined) {
-            val totalBitsRead = nBits - remainingBitsToGet
-            MaybeULong(dis.remainingBits.get + totalBitsRead)
-          } else {
-            MaybeULong.Nope
-          }
-        self.PENotEnoughBits(start, nBits, remainingBits)
+        self.PENotEnoughBits(start, nBits, dis)
```
stevedlawrence, line 56: This change could potentially lead to confusion because blobs are read in chunks?

As an example, let's say we have 3 bytes in the data, our blob size if 4 bytes, and the chunk size is 1 byte.

So `nBits` will be 4 * 8 = 32 bits to read, with only 24 bits available. Because we chunk the reads, we will read 8 bits at a time and successfully read all 24 bits. Then we will go to read the last byte since we need 32 total. That doesn't exist so we call `PENotENoughBits` passing in 32 as the bits we needed. The remaining bits calculation will say there are zero bits available, so the error would say something like

> bit position 24, 32 bits needed, 0 available

Which is may be confusing since we needed 32 bits starting at bitPos = 0, not 24.

Really, we probably want to say:

> bit position 0, 32 bits needed, 24 available

But that's hard because by the time we get to this point the bit position has changed because of the chunk reads, and there aren't actually 24 bits available at this new position, so PENotENoughBits can't easily report that.

Maybe instead we change this line to this:

```scala
self.PENotEnoughBits(start, bitsToGet, dis)
```

Which would report something like this:

> bit position 24, 8 bits needed, 0 available

That's maybe a bit confusing if users aren't aware chunking is going on. They might say "my blob length is 32, why does it say only 8 are needed". But maybe that's fine? Maybe we just teach users that blobs are chunked and if there isn't enough data the error will be about the last chunk read that failed and at least these numbers are accurate?

mike-mcgann: It sounds like when something is to be consumed from the input stream that the starting position should be saved before starting?

stevedlawrence: We can do capture the bitPosition0b (and in some cases we do), but the PE that gets created still uses the state a the time of creation. So we would need to reset the bitPosition back to the beginning of the blob prior to calling PENotEnoughBits. But, that might cause additional problems. For example, the BucketingInputStream has a maximum size it will store, after which it starts throwing away older bytes--if we reset back into those discarded bytes we might get other errors. And with a big blob that doesn't seem impossible.

Maybe alternatively we just save the DataLocation instead of the bitPosition from the beginning of the blob and manually create a PE with that DataLocation?

There is an `PENotEnoughBits` function that does accept a DataLocation just below the one you modified:
```
  def PENotEnoughBits(
    pstate: PState,
    sfl: SchemaFileLocation,
    dataLoc: DataLocation,
    neededBits: Long,
    remainingBits: MaybeULong,
  ) = {
```
That gets a correct starting position. And we would need to adjust your `remainingBits` logic to use the `bitPosition` from the `DataLocation` instead of the `DataInputstream` and account for differences between the DataInputStream and DataLocation bitPositions, but I *think* that should be doable? Might also make sense to have the PENotEnoughBits function you modified to call this one,  and move your logic into that so they use the same logic.

All doable, but could get messy.

mike-mcgann:
Closing for revisions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
mike-mcgann: The `currentLocation` member in `PState` has `bitPos1b` but not `bitPos0b`. I ended up storing the value of `bitPos0b` at the beginning of the parse and then using that in the error message if need be. The chunk writer now returns the number of bits actually written instead of handling the error there and the caller creates the error if the expectation isn't met.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -65,12 +66,51 @@ sealed trait Parser extends Processor {
...
 
+  def PENotEnoughBits(
+    pstate: PState,
+    neededBits: Long,
+    source: InputSourceDataInputStream,
+  ): Unit = {
+    PENotEnoughBits(pstate, neededBits, source.bitPos0b, source)
+  }
+
   def PENotEnoughBits(
```
stevedlawrence, line 114: I'd suggest the blobs should use this `PENotEnoughBits` (and modify it to have the remaining bits logic) and pass in a `DataLocation` saved from the beginning of the  blob parsing before it started reading chunks. And then the remaining bits logic could adjust bits based on the difference between `DataLocation.bitPos1b - 1` and `pstate.bitPos0b`.  We'll also want to modify this to call the `PE` function that accepts a `DataLocation` so the `ParseError` uses the saved `DataLocation` instead of the one from the `pstate`. This way the generated diagnostics has he correct bitPosition at the beginning of the blob.

The above `PENotENoughBits` that everything else uses can then just call this one, passing in `pstate.currentLocation` as the `DataLocation`. In that case the difference between `DataLocation.bitPos1b - 1` and `pstate.bitPos0b` will be zero, so the adjusting logic just won't change anything.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -65,7 +66,37 @@ sealed trait Parser extends Processor {
...
+      }
+    }
+
+    val remainingBits = {
+      if (source.hasReachedEndOfData) {
+        val bitsAvailable = {
+          if (source.bitPos0b % 8 == 0)
+            source.knownBytesAvailable * 8
+          else {
+            source.knownBytesAvailable * 8 - (source.bitPos0b % 8)
```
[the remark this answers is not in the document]
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mike-mcgann: Updated to now use the DataLocation instead of the bitPos.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -90,6 +85,53 @@ sealed trait Parser extends Processor {
     )
   }
 
+  def PENotEnoughBits(
```
stevedlawrence, line 88: I wonder if we should remove the above `PENotEnoughBits` that accepts a `remainingBits` parameer? Nothing should every calculate their own remaining bits and call that since they'll likely do it wrong and not use information from DataInputStream, DataLocation, etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -90,6 +85,53 @@ sealed trait Parser extends Processor {
...
+        val nBits = lim - startPos
+        MaybeULong(nBits)
+      }
+    }
+
+    val remainingBits = {
+      if (source.hasReachedEndOfData) {
+        val bitsAvailable = {
+          val fragmentBitsRead = startPos % 8
+          source.knownBytesAvailable * 8 - fragmentBitsRead
```
stevedlawrence, line 110: I think this is incorrect? It only uses `startPos` to know how many fragment bits to remove, but doesn't adjust the total known bits available based on `startPos`, since `knownBytesAvailable` is based on `dis.bitPos0b`.

I think we could do something like this:
```scala
val bitsAvailable = {
  val fragmentBitsReadFromSourcePos = source.bitPos0b % 8
  val bitsAvailableFromSourcePos = source.knownBytesAvailable * 8 - fragmentBitsReadFromSourcePos
  val bitsBetweenStartPosAndSourcePos = source.bitPos0b - startPos
  val bitsAvailableFromStartPos = bitsAvailableFromSourcePos + bitsBetweenStartPosandSource
  bitsAvailableFromStartPos
}
```

There might be better variable names, but the long ones do make it clear what's being calculated and why.

So this uses the previous logic to figure out how many bits are available starting from the dis.bitPos0b, then just adds the number of bits between the `DataLocation` and the `pstate currentLocation`, since the two might not be the same. 

In fact, and it might be cleaner, you could maybe have all your old logic to calculate remainingBits (which is all dis.bitPos0b based), and then at the very end do something like this to adjust the remaining bits based on the DataLocation, e.g.

```scala
val remainingBits = ... // all old logic
val bitsBetweenStartPosAndCurPos = (dis.bitPos0b - startPos)
val remainingBitsFromStartPos = remainingBits + bitsBetweenStartPosAndCurPos 
PENotEnoughBits(..., neededBits, remainingBitsFromStartPos)
```
If `dis.bitPos0b` and `sourcePos` are the same (i.e. most cases except for things that use chunkWithBits), then `bitsBetweenStartPosAndCurPos` is just 0 and doesn't have an affect. But in the case of chunkWithBits where they aren't the same, we just add the difference between where chunksWithBits started (startPos) and where it we start calculating remainingBits from (dis.bitPos0b).

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -90,6 +85,53 @@ sealed trait Parser extends Processor {
...
+        }
+        if (remainingLimitedBits.isEmpty || bitsAvailable < remainingLimitedBits.get)
+          MaybeULong(bitsAvailable)
+        else
+          remainingLimitedBits
+      } else remainingLimitedBits
+    }
```
stevedlawrence, line 117: I wonder if `remainingBits` should be a `ULong` instead of a `MaybeULong` now? If we ever call `PENotEnoughBits`, then must have hit the bit limit and `remainingLimitedBits` must be defined, or we hit the end of data, and bitsAvailable must be defined. Either way, we always have some number that represents the number of available bits. So maybe this last part just wants to be:


```scala
     {
        ...
        if (remainingLimitedBits.isEmpty || bitsAvailable < remainingLimitedBits.get)
          bitsAvailable
        else
          remainingLimitedBits.getULong
      } else remainingLimitedBits.getULong
```

And if you do the startPos adjustment at the end, then that becomes a bit cleaner since remainingBits is just a `ULong` instead of a `MaybeULong` and you don't have to check if it's defined.

stevedlawrence (commented):
Just a couple more comments, might still be a couple bugs?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mike-mcgann: There is one place that passes in a None or Nope here so I created a new constructor that doesn't take this argument in that case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/HexBinaryLengthParsers.scala
```diff
@@ -64,9 +64,12 @@ sealed abstract class HexBinaryLengthParser(override val context: ElementRuntime
       // limiting the size of hex binarie data.
       val array = new Array[Byte](nBytes.toInt)
       val buffer = ByteBuffer.wrap(array)
-      writeBitsInChunks(start, nBits) { case (bytes, nBytes) =>
+      val written = writeBitsInChunks(start, nBits) { case (bytes, nBytes) =>
         buffer.put(bytes, 0, nBytes)
       }
+      if (written != nBits) {
+        PENotEnoughBits(start, nBits, start.dataInputStream)
+      }
```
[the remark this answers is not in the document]
mike-mcgann: Updated

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BlobLengthParsers.scala
```diff
@@ -32,15 +31,18 @@ trait ByteChunkWriter { self: PrimParser =>
    * which takes as parameters an array of bytes and the number of bytes in
    * that array that should be written. The chunk size is defined by the
    * blobChunkSizeInBytes tunable.
+   *
+   * @return the actual number of bits written
```
stevedlawrence, line 35: I think this no longer returns number of bits written? Can this comment be removed?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
```diff
@@ -65,21 +65,42 @@ sealed trait Parser extends Processor {
...
+          bitsAvailableFromStartPos
+        }
+        if (remainingLimitedBits.isEmpty || bitsAvailable < remainingLimitedBits.get)
+          bitsAvailable
+        else
+          remainingLimitedBits.get
+      } else remainingLimitedBits.get
+    }
+
+    val remainingStr = s" but found only ${remainingBits} available"
```
stevedlawrence, line 103: Can we make this part of the PE message now that `remainingBits` always exists? And then the PE args are `neededBits` and `remainingBits`?

stevedlawrence (approved):
+1 :+1:  two minor cleanup suggestions

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/Parser.scala
mike-mcgann: Updated

tuxji:
Looks ready to squash and merge.

PR #1005, opened 2023-04-17
Fix min/maxOccurs validation for optional elements

stevedlawrence:
Commit dd60b6067d made changes to validation so that it took into account absent elements that incremented the array iteration count but did not actually change the array size. It did this by just counting the actual elements in the infoset. Unfortunately, that change was very subtly broken for optional elements since it assumed maybeLastChild was always a DIArray, but it could actually be a DISimple/DIComplex for optional elements since they use very similar array logic.

We could modify this logic to take DISimple/DIComplex nodes into account, but later commit 16c738c38f added new state to differentiate between the occurs index and the iteration index. So this just modifies the array validation logic so it is very similar to the old logic, but uses the new state that accounts for absent elements.

DAFFODIL-2808


mike-mcgann (approved):
+1

tuxji (approved):
+1

PR #1013, opened 2023-05-08
Remove use of the IntegrationTest sbt configuration

stevedlawrence:
Upcoming versions of sbt will deprecate the IntegrationTest configuration. Using their suggested migration path, this moves the existing integration tests to a new daffodil-test-integration project. Because of the extra time needed to stage the CLi and run the tests, this project is not aggregeated so must be run manually.

DAFFODIL-2811


mbeckerle (approved):
+1 Very clean fix. I can see why they removed the IntegrationTest feature given that this is the workaround. It's just better done this way.

.github/workflows/main.yml
```diff
@@ -161,7 +161,7 @@ jobs:
         run: $SBT test
 
       - name: Run Integration Tests
-        run: $SBT IntegrationTest/test
+        run: $SBT daffodil-test-integration/Test/test
```
tuxji, line 164: According to the DEVELOP.md changes below, `daffodil-test-integration/test` would work too.  Why do you use the longer `daffodil-test-integration/Test/test` in the CI workflow?

tuxji (approved):
+1

I wonder if this change could help us merge #986 which failed CI due to a conflict in jline jars between sbt's classpath and daffodil's classpath?  Steve opened a sbt ticket [sbt/sbt#7177](https://togithub.com/sbt/sbt/issues/7177) and the offered solution was to fork the tests.  If only one or two tests were impacted, we could move them to daffodil-test-integration.  If too many tests were affected, never mind.

.github/workflows/main.yml
stevedlawrence: Good call, the `Test` isn't needed, will remove.

stevedlawrence:
> I wonder if this change could help us merge https://github.com/apache/daffodil/pull/986

Possibly. It looks like 42 tests failed in the jline PR, so we'd probably need to move all of those here. Doable, but would probably cause CI to take much longer.

The debugger isn't a really a critical part of the CLI, and I don't know of any updates to jline that we really need. Maybe we just avoid upgrading jline until sbt fixes this issue? Unfortunately that might be a while, likely not until 2.0 at the earliest, but I don't think we're in any rush to update jline.

PR #1025, opened 2023-06-06
Fix typo in genVersion resource generator

stevedlawrence:
The comment says the opposite of what is intended, which could be very confusing to developers.

DAFFODIL-2816


mike-mcgann (approved):

tuxji (approved):
+1

PR #1026, opened 2023-06-07
Support text numbers that parse to negative zero double

stevedlawrence:
ICU parses negative zero to a Double type, which we do not currently allow, expecting only Infinifity or NaN. This modifies the assertion to allow negative zero as well.

DAFFODIL-2818

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
@@ -3683,6 +3683,18 @@
     </tdml:errors>
   </tdml:unparserTestCase>
 
+  <tdml:parserTestCase name="double_negativeZero" root="doubleText"
+    model="SimpleTypes-Embedded.dfdl.xsd" description="Section 5 Schema types-double - DFDL-5-009R"
+    roundTrip="false">
+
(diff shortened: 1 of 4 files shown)


mbeckerle (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -3683,6 +3683,18 @@
     </tdml:errors>
   </tdml:unparserTestCase>
 
+  <tdml:parserTestCase name="double_negativeZero" root="doubleText"
+    model="SimpleTypes-Embedded.dfdl.xsd" description="Section 5 Schema types-double - DFDL-5-009R"
```
tuxji, line 3687: (Aside) I am interested in ways to test Daffodil's C code generator more thoroughly.  However, the majority of Daffodil's tests (like this one) parse textual data, not binary data.  Even when I find binary data tests, they might also use hidden groups, input and/or output value calculations, variables, or other unsupported features.  Is there an easy way to identify existing Daffodil tests which could be run with "daffodil test -I daffodilC" successfully (besides just trying it on all tests and seeing which ones pass)?

mbeckerle: I suggest trying all tests and see what ones pass :-)

PR #1029, opened 2023-06-08
Add codegen-c support for runtime length hexBinary

tuxji:
Extend Daffodil's C code generator to support runtime length hexBinary elements.  Also make some cleanups for better consistency.

DAFFODIL-2819

dependency-scan.yml: Saw a better example somewhere (it was called dependency-graph.yml) so modify and rename to match that example.

DaffodilCCodeGenerator.scala: Change comments and names in compileCode and pickCommand to be more similar to another codegen.  Extend generateCode to handle RepOrderedWithMinMaxSequenceChild.

BinaryValueCodeGenerator.scala: Remove restriction on runtime lengths.

CodeGeneratorState.scala: Add cExpression to implement a poor man's C code expression generator until Daffodil adds custom code generation functionality to compiled DFDL expressions.  Remove namespace prefixes in makeLegalForC to ensure they don't appear in C code.  Extend arrayMaxOccurs to handle more cases.  Make cStructFieldAccess do a better job of converting a DFDL expression written as a path to a C struct field dereference.

HexBinaryCodeGenerator.scala: Allow specified length hexBinary elements to use runtime lengths as well as constant lengths.

ex_nums.tdml: Add a new test to check support for runtime length hexBinary elements.

Evaluatable.scala: Allow codegen-c to get compiled expression from EvaluatableExpression.

TestExNums.scala: Make both daffodil and daffodilC run new test.

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -13,30 +13,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Dependency scan
+name: Dependency Graph
 
-# Controls when the workflow will run
+# Run after pushing a commit to the main branch.
+
 on:
-  # Triggers the workflow on push or pull request events but only for the "main" branch
   push:
-    branches: [ "main" ]
-    
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
+    branches:
+      - main
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  # This workflow contains a single job called "build"
-  build:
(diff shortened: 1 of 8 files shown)


stevedlawrence (approved):
+1

mbeckerle (approved):
+1
The code looks fine, but I am somewhat concerned by the codecov gaps. Can you add a test to cover them, or explain why they're present and unexercised?

I'll approve it anyway, but advocate for addressing this first.

tuxji (commented):
I've added explanations for the uncovered lines.  The new test didn't use an idiv operator in the expression which would have covered some lines, and some other changes aren't finished.

PR #1033, opened 2023-06-12
Create a separate file for a portable GeneralFormat

stevedlawrence:
The DFDLGeneralFormat.dfdl.xsd file defines the emptyElementParsePolicy extension property in the GeneralFormat format. Even if a schema uses GeneralFormatPortable, which does not deine this exension property, IBM DFDL still sees it and creates an error. This means that it is impossible to use this file for portability tests.

To resolve this, this creates a new DFDLGeneralFormatPortable.dfdl.xsd file that does not contain any properties that IBM DFDL does not support. Additionally, the define format in this file uses the same GeneralFormat name as DFDLGeneralFormat.dfdl.xsd. This allows schema authors to alway reference the GeneralFormat format, and only change the xs:include to make a schema portable or not.

Compatibility/Deprecation:

The GeneralFormatPortable define format has been removed. Instead, portable schemas should import a new portable file and reference the "GeneralFormat" format. For example, a schema that previously looked like this:

    <dfdl:format ref="GeneralFormatPortable" ... />

Should now look like this:

    <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormatPortable.dfdl.xsd" />

    <dfdl:format ref="GeneralFormat" ... />

DAFFODIL-2820

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/core/api/TestDsomCompiler3.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/core/api/TestDsomCompiler3.scala
@@ -73,7 +73,7 @@ class TestDsomCompiler3 {
       // Verify things still work using specified tmpDir
       //
       val Seq(schema) = sset.schemas
-      val Seq(schemaDoc, _) = schema.schemaDocuments
+      val schemaDoc = schema.schemaDocuments.head
       val Seq(decl) = schemaDoc.globalElementDecls.map { _.asRoot }
       val Seq(ct) = schemaDoc.globalComplexTypeDefs
       assertEquals("example1", ct.name)
--- a/daffodil-cli/src/test/resources/org/apache/daffodil/cli/blob_backtracking.dfdl.xsd
+++ b/daffodil-cli/src/test/resources/org/apache/daffodil/cli/blob_backtracking.dfdl.xsd
@@ -22,11 +22,11 @@
   xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
-  <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+  <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormatPortable.dfdl.xsd"/>
 
   <annotation>
     <appinfo source="http://www.ogf.org/dfdl/">
-      <dfdl:format ref="tns:GeneralFormatPortable"
+      <dfdl:format ref="tns:GeneralFormat"
         representation="binary"
         encodingErrorPolicy="replace" />
     </appinfo>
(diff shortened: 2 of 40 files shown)


stevedlawrence:
Note that a number of tests will fail until https://github.com/apache/daffodil/pull/1032 is merged, but I think this is straightforward enough that it an still be reviewed.

daffodil-core/src/test/scala/org/apache/daffodil/core/api/TestDsomCompiler3.scala
```diff
@@ -73,7 +73,7 @@ class TestDsomCompiler3 {
       // Verify things still work using specified tmpDir
       //
       val Seq(schema) = sset.schemas
-      val Seq(schemaDoc, _) = schema.schemaDocuments
+      val schemaDoc = schema.schemaDocuments.head
```
tuxji, line 76: For other reviewers, this line has been simplified to get the first schemaDoc in a more straightforward way.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
```diff
@@ -19,101 +19,17 @@
...
 
-      <!--
-        This set of properties should never change. It can be added to 
-        if properties are missing, but should not be changed.
-        
-        Changes should be introduced by way of defining new formats in
-        terms of this one, which override definitions. See below in 
-        the file for examples. 
-       -->
-      <dfdl:defineFormat name="GeneralFormatOriginal">
```
tuxji, line 34: I see GeneralFormatOriginal has been moved to its own file called DFDLGeneralFormatBase.dfdl.xsd and that file included above here.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd
```diff
@@ -19,101 +19,17 @@
...
-        />
-      </dfdl:defineFormat>
-      
       <dfdl:defineFormat name="GeneralFormat">
-        <dfdl:format ref="GeneralFormatOriginal"
+        <dfdl:format ref="GeneralFormatBase"
             emptyElementParsePolicy="treatAsEmpty"/>
       </dfdl:defineFormat>
         
-      <dfdl:defineFormat name="GeneralFormatPortable">
```
tuxji, line 110: Likewise, I see GeneralFormatPortable has been moved to its own file called DFDLGeneralFormatPortable.dfdl.xsd.

tuxji (approved):
+1, looks good

daffodil-core/src/test/scala/org/apache/daffodil/core/api/TestDsomCompiler3.scala
stevedlawrence: Yep, because there is now an extra include (GeneralFormatBase.xsd) there are now three docs in the `schema.schemaDocuments` list in a lot of tet. The first is still the root schema that we actually care about, but the pattern match fails. I figure switching to using `.head` will allow us to change the includes again if needed without breaking these tests, for example if IBM DFDL and Daffodil fix compatibility issues and the portable stuff goes away.

mbeckerle (approved):
+1

stevedlawrence:
Yep, I'll wait until the schematron change is merged, rebase this PR to double check all tests pass, and then merge.

PR #1035, opened 2023-06-14
Correctly check for errors when parsing text calendars

stevedlawrence:
The SimpleDateFormat parse() method does not set the error index on the ParsePosition if there is an error. Instead, it only sets the normal index to the last successful parse position of the input string.

Daffodil currently relies on the error index, which means our error checking is incorrect.

Fortunately, in most cases, this doesn't matter because our error checking also requires that we parse the entire input string, so if the index is zero it wouldn't match the string length and we would still see it as an error. The one exception to this is if the input string is zero length, in which case we do not correctly error and just create a date/time with value zero.

This is fixed by correctly checking index for zero instead of inspecting the error index.

This also does a little refactoring to replace the very old cache implementation with the much newer Evaluatable for creating SimpleDateFormat instances.

DAFFODIL-2821

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesDateTime.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesDateTime.scala
@@ -31,6 +31,7 @@ import org.apache.daffodil.lib.schema.annotation.props.gen.CalendarPatternKind
 import org.apache.daffodil.lib.schema.annotation.props.gen.Representation
 import org.apache.daffodil.runtime1.processors.CalendarEv
 import org.apache.daffodil.runtime1.processors.CalendarLanguageEv
+import org.apache.daffodil.runtime1.processors.DateTimeFormatterEv
 import org.apache.daffodil.runtime1.processors.parsers.ConvertBinaryCalendarSecMilliParser
 import org.apache.daffodil.runtime1.processors.parsers.ConvertTextCalendarParser
 import org.apache.daffodil.unparsers.runtime1.ConvertBinaryCalendarSecMilliUnparser
@@ -165,6 +166,12 @@ abstract class ConvertTextCalendarPrimBase(e: ElementBase, guard: Boolean)
     } else {
       p
     }
+
+    schemaDefinitionWhen(
+      patternToCheck.length == 0,
+      "dfdl:calendarPatttern contains no pattern letters",
+    )
+
     patternToCheck.toSeq.foreach(char =>
       if (!validFormatCharacters.contains(char)) {
         if (e.representation == Representation.Binary)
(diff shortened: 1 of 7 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/EvCalendarLanguage.scala
```diff
@@ -96,3 +97,33 @@ class CalendarEv(
...
+
+    // As per ICU4J documentation, "Date formats are not synchronized. If multiple threads
+    // access a format concurrently, it must be synchronized externally." Rather than
+    // synchronzing, we create a ThreadLocal so each thread gets their own copy of the
+    // SimpleDateFormat
+    val dateFormatTL = new ThreadLocal[SimpleDateFormat] with Serializable {
+      override def initialValue = {
+        val formatter = new SimpleDateFormat(pattern, locale)
+        formatter.setCalendar(calendar)
+        formatter.setLenient(true) // TODO: should this use calendarCheckPolicy?
```
stevedlawrence, line 123: I'm pretty sure this is wrong and makes it so calendar parsing is always lenient regardless of the `dfdl:calendarCheckPolicy` property, but this copies the existing logic from the old `tlDataFormatter` so at least doesn't cause any regressions and just fixes the one bug at hand.

I've attempted to make it correctly use calendarCheckPolicy, but it broke a number of tests, including some unexpected failures that I think need more investigation. Once I can create a reproducible test I'll open a ticket, but I think a fix might be best for the next release so we have time to deal with regressions.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesDateTime.scala
```diff
@@ -201,18 +208,34 @@ abstract class ConvertTextCalendarPrimBase(e: ElementBase, guard: Boolean)
     p
   }
 
+  private lazy val dateTimeFormatterEv = {
+    val ev = new DateTimeFormatterEv(
+      calendarEv,
+      localeEv,
```
mbeckerle, line 214: This change may affect/fix https://issues.apache.org/jira/browse/DAFFODIL-2823 also, as the locale is being supplied differently now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/EvCalendarLanguage.scala
mbeckerle: Agree we should fix the one problem, and defer the strict/lenient to next point release.

mbeckerle (approved):
+1

tuxji (approved):
+1 (others, please step up too)

PR #1039, opened 2023-06-28
Specify a locale in units tests

stevedlawrence:
Some unit tests do not specify a locale when creating certain objects, which causes the default system locale to be used. This can cause these tests to fail if run on a system with a different locale. This modifies those tests to specify a locale and ignore the system locale.

This also modifies the GitHub actions so some jobs are run with non-US locales so that we can detect if new tests are added with locale dependent behaviors.

Note that commit 1a0954a7b6 modified integration tests to not be locale dependent--only a small number of unit tests need fixing.

DAFFODIL-2599, DAFFODIL-2612

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,14 @@ jobs:
             shell: msys2 {0}
             env_cc: clang
             env_ar: llvm-ar
+          # configure different languages for some jobs, defaulting to en_US.UTF-8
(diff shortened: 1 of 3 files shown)


mbeckerle (approved):
+1

tuxji (approved):
+1

Fellow contributors, please review PRs - I'm still on vacation.

PR #1047, opened 2023-07-11
Implement Trace functionality for CLI TDML test

olabusayoT:
- fix bug with debugger where we were casting the data processor instead of the debugger

Note, we were unable to test the trace output using CLI tests as the trace output wasn't able to be found using CLI.expect. My guess is it has something to do with output being sent to a PipedOutputStream, which may be different from System.out of the other output.

DAFFODIL-2694

### DIFF
--- a/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLItdml.scala
+++ b/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLItdml.scala
@@ -110,6 +110,46 @@ class TestCLItdml {
     }(ExitCode.Success)
   }
 
+  @Test def test_CLI_Tdml_Trace_singleTest(): Unit = {
+    val tdml = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml",
+    )
+
+    runCLI(args"-t test $tdml byte_entities_6_08") { cli =>
+      cli.expect("parser:")
+      cli.expect("bitPosition:")
+      cli.expect("data:")
+      cli.expect("[Pass] byte_entities_6_08")
+    }(ExitCode.Success)
+  }
+
+  @Test def test_CLI_Tdml_Debug_singleTest(): Unit = {
+    val tdml = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml",
+    )
+
+    runCLI(args"-d test $tdml byte_entities_6_08") { cli =>
+      cli.expect("(debug)")
+      cli.sendLine("continue")
+      cli.expect("[Pass] byte_entities_6_08")
+    }(ExitCode.Success)
+  }
+
+  @Test def test_CLI_Tdml_DebugFile_singleTest(): Unit = {
+    val tdml = path(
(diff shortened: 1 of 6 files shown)


mike-mcgann (approved):

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/processor/tdml/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -190,8 +190,8 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor)
     }
 
   override def withDebugger(db: AnyRef): DaffodilTDMLDFDLProcessor = {
-    Assert.usage(dp ne null)
-    val d = dp.asInstanceOf[Debugger]
+    Assert.usage(db ne null)
+    val d = db.asInstanceOf[Debugger]
```
mbeckerle, line 194: Wow. What a bug. 

We should disallow use of tiny value names like dp and db and spell things out a little more I think. 

db here is a parameter name. That one should definitely be spelled out as 'debugger'.

mbeckerle: Note that code cov says this wasn't a bug, because it's _not called_.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLItdml.scala
```diff
@@ -110,6 +110,16 @@ class TestCLItdml {
     }(ExitCode.Success)
   }
 
+  @Test def test_CLI_Tdml_Trace_singleTest(): Unit = {
+    val tdml = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml",
+    )
+
+    runCLI(args"-t test $tdml byte_entities_6_08") { cli =>
+      cli.expect("[Pass] byte_entities_6_08")
```
mbeckerle, line 119: Please add comments here about what you'd like to test, but how expect isn't letting you do so. 

Also add a JIRA ticket about that issue and reference it from the comments.

mbeckerle (approved):
+1

daffodil-tdml-processor/src/main/scala/org/apache/daffodil/processor/tdml/DaffodilTDMLDFDLProcessor.scala
olabusayoT: Right, I noticed the bug while working on getting the -d functionality to work with "test", it gets called then. This changeset doesn't have that functionality work, but it felt wrong to leave the bug

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLItdml.scala
olabusayoT: Done

mbeckerle (approved):
+1

mike-mcgann (approved):

PR #1051, opened 2023-07-12
Fix NumberFormatException message in hexCharToValue

pkatlic:
Update exception message in Misc.hex2Bytes.hexCharToValue function to include correct range of allowed values.

Misc.scala: Update exception message to include correct values.

TestMisc.scala: Add tests for Misc.hex2Bytes function.

DAFFODIL-2834


olabusayoT (approved):
+1

PR #1052, opened 2023-07-12
Implement Minimal Option for XMLTextInfosetOutputter

olabusayoT:
- exclude xml slug and prefix namespace bindings

DAFFODIL-2732


mike-mcgann (approved):

mbeckerle (approved):
+1

PR #1057, opened 2023-07-19
Tdml gen app

arosien:


NolanMatt:
@arosien I updated it to only apply to the daffodil-tdml-lib portion of the build, but still getting errors, my new thought is that on the xjcCommandLine it's referencing org.apache.daffodil.tdml which seems to me like a bit of a circular reference that might not resolve since this is the project we're building it in.  I'm not sure though, just seems a bit off to me.  Any ideas?

tuxji:
I saw that this PR needed a maintainer's approval to run CI on it, so I looked over the changes and let CI run but unfortunately the changes don't seem to work on GitHub runner machines.  Some runs printed their error messages in foreign languages (I wonder how that happened?), but one run's English [errors](https://github.com/apache/daffodil/actions/runs/5729863413/job/16213034094?pr=1057) seem to indicate that DFDL additions to XML Schema confuse xjc.  Did you run xjc successfully on your local machine?

stevedlawrence:
> Some runs printed their error messages in foreign languages (I wonder how that happened?)

For a while we had a few tests that expected error messages that were generated and internationalized by the OS (e.g. "could not open file"). But the tests expected english error messages so would fail on systems that defaulted to another language. These tests were fixed to not look for these specific error messages, but to avoid new tests that might do something similar, a couple of github action jobs were updated to run with LANG set to de_DE and ja_JP.

https://github.com/apache/daffodil/blob/main/.github/workflows/main.yml#L61-L68

arosien:
FYI the SbtXjcPlugin unfortunately installs itself globally, so it applies to all projects that have xsd schemas in them, hence the need for the `disablePlugins(SbtXjcPlugin)` expressions sprinkled around everywhere except the tdml-lib project.

build.sbt
```diff
@@ -60,6 +61,7 @@ lazy val macroLib = Project("daffodil-macro-lib", file("daffodil-macro-lib"))
   .disablePlugins(OsgiCheckPlugin)
 
 lazy val propgen = Project("daffodil-propgen", file("daffodil-propgen"))
+  .disablePlugins(SbtXjcPlugin)
```
mbeckerle, line 64: Can this disable just become part of commonSettings, then the only clutter would be the one enableSettings for tdml?

michael-hoke: We did something similar - set the Seq containing xsd schemas to run through JAXB to empty in commonSettings - it allows us to remove all of these disablePlugins lines. It builds for me locally, so, hopefully, the CI tests will build now as well.

build.sbt
```diff
@@ -143,7 +144,9 @@ lazy val sapi = Project("daffodil-sapi", file("daffodil-sapi"))
 
 lazy val tdmlLib = Project("daffodil-tdml-lib", file("daffodil-tdml-lib"))
   .dependsOn(macroLib % "compile-internal", lib, io, io % "test->test", slf4jLogger % "test")
+  .enablePlugins(SbtXjcPlugin)
```
arosien, line 147:
```suggestion
```
Since the plugin applies globally, this isn't needed.

build.sbt
```diff
@@ -211,6 +214,67 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+    xjcCommandLine += "org.apache.daffodil.tdml",
+    xjcBindings += "daffodil-tdml-lib/src/main/resources/bindings.xjb",
+    xjcLibs := Seq(
+      "org.glassfish.jaxb" % "jaxb-xjc" % "2.2.11",
+      "com.sun.xml.bind" % "jaxb-impl" % "2.2.11",
+      "javax.activation" % "activation" % "1.1.1",
+    ),
+    xjcJvmOpts ++= extraJvmOptions,
+    Compile / xjc / sources := Seq(
+      // Need to refer to the individual files because we do not want XMLSchema to go through the JAXB compliation
```
arosien, line 250:
```suggestion
      // Need to refer to the individual files because we do not want XMLSchema to go through the JAXB compilation
```
speeling

build.sbt
```diff
@@ -17,6 +17,7 @@
 
 import scala.collection.immutable.ListSet
 
+import com.github.retronym.sbtxjc.SbtXjcPlugin
```
arosien, line 20:
```suggestion
```

build.sbt
michael-hoke: I don't think that's the case - We set `Compile / xjc / sources := Seq()` in `commonSettings` and overwrite it with this - these should be the only files that we generate JAXB for. Potentially, any imports we would also grab, but I believe we have all of the imports explicitly defined here anyways.

I can double check this, but I'm almost positive that either setting `Compile / xjc / sources` to contain the folders or adding XMLSchema won't build because XMLSchema prevents the build from passing (If I'm remembering correctly, there are a lot of clashes, and we would need a lot more definitions in the bindings.xjb file). I also don't see any `xs:include` lines for it, so I don't think it's strictly necessary to include this file for the TDML JAXB bindings to be complete.

michael-hoke: Double checked - it clashes a lot with XMLSchema_for_DFDL.xsd - about 50 additional entries into the bindings.xjb file would be required.

arosien: Ah I see, re: `xs:include` wonkiness. Yeah I was hoping to use the default `Compile / xjc / sources` (https://github.com/sbt/sbt-xjc/blob/master/src/main/scala/com/github/retronym/sbtxjc/SbtXjcPlugin.scala#L64).

The only downside is that if there are any file-level xsd additions or deletions then the build file needs to be adjusted.

michael-hoke: Right. Unless there's a way to use the default and then explicitly exclude XMLSchema.xsd but that isn't perfect either... I'd be okay with either way.

mbeckerle: I think you should only need to generate objects for:

tdml.xsd

The challenge is that some TDML files embed mini-schemas inside using tdml:defineSchema. 

That can hold anything one would find inside a DFDL schema, so.... not sure how to cope with that without dragging all of XMLSchema_for_DFDL.xsd 

These other schema files should not be needed nor used:

dafint.xsd
dafext.xsd
dfdlx.xsd
datatypes.dtd
xml.xsd
XMLSchema.dtd
XMLSchema.xsd

And hopefully not these depending on the tdml:defineSchema issue:

XMLSchema_for_DFDL.xsd
DFDL_part?_*.xsd - 3 files match this (used by XMLSchema_for_DFDL.xsd)

michael-hoke: I just did a build, removing every xsd file other than tdml.xsd and removing the reference to bindings.xjb. I get a lot of undefined element messages. The XJC plugin seems to require every file whose definitions we import to go through the JAXB process as well.

mbeckerle: I suggest maybe this is an approach that could work:

- make a variant of tdml.xsd which omits the tdml:defineSchema element. 

- plan to add back some manual implementation for that object. 

So you need to:
Comment out all the includes/imports.

Comment out line 64 (element ref to tns:defineSchema).

Comment out all other uses of "defineSchema" and the definition of it. 

Will your tool to generate code still work then? That *should* eliminate pulling in all that other stuff and all the excess definitions.

michael-hoke: I ended up having to take out a bit more than just that, but I was able to generate the XSDs without a bindings.xjb file and only having tdml.xsd listed as a source file. Here's the output from `git diff` in a patch file. If you would like me to upload the entire changed document, let me know.

I will go back and attempt to see if this works with our extension, but it looks like it should work for us since we only use a very minimal subset of the TDML schema for now.

[tdml-xsd-update.patch](https://github.com/apache/daffodil/files/12620820/tdml-xsd-update.patch)

stevedlawrence: The content of defineSchema is used for testing schemas emedded in the TDML file itself.

I guess excluding that from jaxb would be fine if the extension only supports external schemas, which is how most real schema projects are used. So maybe that's a fine limitation? Same probably goes for he defineConfig section?

The most important stuff is probably the test cases with inputs and expected outputs.

Is there a plan so we don't have to maintain two separate tdml.xsd files that are 90% the same except for this small patch?

mbeckerle: We could have tdml-core.xsd and tdml.xsd which includes tdml-core.xsd.

The jaxb could use only the core, and the full tdml.xsd could add in the missing stuff like defineSchema and provide the root tdml:testSuite element.

michael-hoke: This change looks like it would work for the extension. There was a change related to namespaces that I'm not thrilled about, but that can more than likely be fixed with a couple more code changes...

We are not planning to support internal schemas in the extension. One of the features we are looking to implement is to provide the ability to package the TDML schema and the DFDL schema in a zip archive, which directly conflicts with having an internal schema, so I don't think we'd be looking at that even for the longer term.

We currently don't support configs in the extension at all, as far as I know... Even if we were to add them in, there's no reason for us to not use them externally, as we're already using the schema.

build.sbt
```diff
@@ -211,6 +212,67 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+    xjcBindings += "daffodil-tdml-lib/src/main/resources/bindings.xjb",
+    xjcLibs := Seq(
+      "org.glassfish.jaxb" % "jaxb-xjc" % "2.2.11",
+      "com.sun.xml.bind" % "jaxb-impl" % "2.2.11",
+      "javax.activation" % "activation" % "1.1.1",
+    ),
+    xjcJvmOpts ++= extraJvmOptions,
+    Compile / xjc / sources := Seq(
+      // Need to refer to the individual files because we do not want XMLSchema to go through the JAXB compilation
+      file("daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd"),
```
mbeckerle, line 249: Use this trick to make the lines less redundant
```
val daflib = "daffodil-lib/src/main/resources/org/apache/daffodil/xsd/"
file(daflib + "tdml.xsd")  
```

mbeckerle (approved):
+1 One suggestion to minimize redundant dir strings in pathnames.

build.sbt
michael-hoke: I'd like to give the splitting out of the tdml schema into tdml + tdml-core a try before we merge this. I'll leave it as a separate commit until after you and Steve have a chance to review it so we can easily back it out if you decide it's something you don't want.

mbeckerle: ok. Specifically tag me for re-review then when you've done that. Thx.

michael-hoke:
@mbeckerle @stevedlawrence 

I refactored the local parts of the TDML schema into tdml-core.xsd. Other than what I showed earlier in the patch file, the only other notable change was to replace the `choice` in `testSuite` with a `substitutionGroup`.

Everything appears to have been properly separated - the generated JAXB classes don't allow the internal schema/configs in `testSuite`, but the compile/test process is still working for me locally.

I will squash once we decide whether this is a direction we want to continue in.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
```diff
@@ -31,86 +31,13 @@
...
-    </restriction>
-  </simpleType>
+  <xsd:include schemaLocation="tdml-core.xsd"/>
   
 <!-- We want to allow an xsd:schema to be named and directly embedded in 
   the TDML thereby allowing a fully-self-contained single file test case as 
   an exchange medium for tests. -->
 
-  <element name="defineSchema" type="tns:defineSchemaType"/>
+  <element name="defineSchema" type="tns:defineSchemaType" substitutionGroup="tns:testSuiteChoices"/>
```
mbeckerle, line 40: Wow. I learned what a substitution group is really for. Cool.

mbeckerle (approved):
+1 

I'm happy with this reorg of the tdml xsd files.

michael-hoke:
One more (hopefully, the last) push. It's been squashed and should pass the formatting check as well.

This also is tracked in JIRA - [DAFFODIL-2830](https://issues.apache.org/jira/browse/DAFFODIL-2830).

build.sbt
```diff
@@ -211,6 +212,43 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
   .dependsOn(tdmlProc % "test")
   .settings(commonSettings, nopublish)
 
+/* Workaround: certain reflection (used by JAXB) isn't allowed by default in JDK 17:
+ * https://docs.oracle.com/en/java/javase/17/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-7BB28E4D-99B3-4078-BDC4-FC24180CE82B
+ *
+ * While we can handle this JVM quirk at build time, at runtime we won't know
+ * a user's JVM version. We'll provide documentation and an extension setting
```
mbeckerle, line 219: We do compile daffodil but then run it with a variety of JVM versions, so it's possible this needs to be a runtime thing. 
E.g., are these options even allowed on earlier JVMs? 

@stevedlawrence please look this over.

build.sbt
```diff
@@ -241,6 +279,10 @@ lazy val commonSettings = Seq(
   resourceManaged := baseDirectory.value / "resource_managed",
   libraryDependencies ++= Dependencies.common,
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "--verbosity=1"),
+  // SbtXjcPlugin is enabled globally, and will, be default, attempt to build xsd schemas if any exist in a project
```
mbeckerle, line 282: typo "by" default

build.sbt
```diff
@@ -211,6 +212,43 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+    Seq(
+      "--add-opens",
+      "java.base/java.lang=ALL-UNNAMED",
+    )
+  else Seq()
+
+lazy val xjcSettings =
+  Seq(
+    libraryDependencies ++= Seq(
+      "com.sun.xml.bind" % "jaxb-impl" % "2.2.11",
```
mbeckerle, line 233: Most of our dependencies are in project/Dependencies.scala. 

Should these be there as well for consistency?

build.sbt
```diff
@@ -211,6 +212,43 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+      "javax.activation" % "activation" % "1.1.1",
+      "org.glassfish.jaxb" % "jaxb-xjc" % "2.2.11",
+    ),
+    xjcCommandLine += "-nv",
+    xjcCommandLine += "-p",
+    xjcCommandLine += "org.apache.daffodil.tdml",
+    xjcLibs := Seq(
+      "org.glassfish.jaxb" % "jaxb-xjc" % "2.2.11",
```
mbeckerle, line 241: Please label which of these are the runtime dependencies, and which are the compile-time dependencies. (I'm assuming there is that separation).

mbeckerle (commented):
I want to see @stevedlawrence comments on this before approval.

build.sbt
```diff
@@ -211,6 +212,43 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+lazy val xjcSettings =
+  Seq(
+    libraryDependencies ++= Seq(
+      "com.sun.xml.bind" % "jaxb-impl" % "2.2.11",
+      "javax.activation" % "activation" % "1.1.1",
+      "org.glassfish.jaxb" % "jaxb-xjc" % "2.2.11",
+    ),
+    xjcCommandLine += "-nv",
+    xjcCommandLine += "-p",
+    xjcCommandLine += "org.apache.daffodil.tdml",
```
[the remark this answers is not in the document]
stevedlawrence: Can we spend some time trying to figure out the reproducible builds issue with jaxb generated files that we saw with vscode? Daffodil currently has reproducible builds for jars (the other zip/tar/etc release artifacts have timestamp issues) and I would like to keep it that way. There's some discussions in the ASF about automating things like release signing, but that requires reproducible builds.

I see there's a `-no-header` xjc option that prevents a header comment containing a timestamp from being added. I wouldn't expect a comment to affect the .class file, so I'm not optimistic it would fix it, but at least it's one less piece of randomness to consider.

stevedlawrence (commented):
Seems reasonable, though I would like to see some tests that show how these generated class are use and make sure it actually works.

I would also like to track down the reproducible build issue seen in vscode jaxb. Daffodil jars are reproducible so if this makes it so they aren't, that's a regression I would like to avoid.

build.sbt
arosien: I also accidentally type "optinos" all the time :)

build.sbt
arosien, line 282:
```suggestion
  // SbtXjcPlugin is enabled globally, and will, by default, attempt to build xsd schemas if any exist in a project
```

build.sbt
arosien: I thought the plugin defined the same set of libs, but the activation jar was [only added recently](https://github.com/sbt/sbt-xjc/commit/2178c47e50020dd09ad55781c932282250b8d752) (2021!) but hasn't been released yet. (Latest release is 0.10, which doesn't even have a tag.)

I'm testing reducing the additional library mentions in this settings block.

stevedlawrence: Oof, I didn't notice it that change is two years old without a release. I guess we could just do this in that case to just fix the issue:
```scala
xjcLibs += "javax.activation"   % "activation" % "1.1.1",
```
Keeps things a little less verbose. And I doubt we have to worry about he plugin being updated and breaking things. Doubt there will be another release anytime soon.

Alternatively, has anyone looked at [scalaxb](https://github.com/eed3si9n/scalaxb)? It looks like it's created by the main SBT maintainer, so imagine it will continue to be maintained for a while and with SBT best practices.

I made changes this PR to use scalaxb instead of xjc and it was pretty straightforward. This was all I had to change to this PR:

```scala
lazy val tdmlLib = Project("daffodil-tdml-lib", file("daffodil-tdml-lib"))
  .enablePlugins(ScalaxbPlugin)
  .dependsOn(macroLib % "compile-internal", lib, io, io % "test->test", slf4jLogger % "test")
  .settings(commonSettings)
  .settings(scalaxbSettings)

lazy val scalaxbSettings = Seq(
  Compile / scalaxb / scalaxbPackageName := "org.apache.daffodil.tdml.scalaxb",
  Compile / scalaxb / sources := Seq(
    (lib / Compile / resourceDirectory).value / "org/apache/daffodil/xsd/tdml-core.xsd",
   )
)
```

Nicely, it doesn't auto enable so it can be enabled for only the one project we need without the `sources := Seq()` hack.

The only issue I ran into is the generator adds imports that aren't used, which Daffodil treats as an error. But disabling fatal warnings, the generated code compiles fine. There's probably a way to fix this without disabling warnings. And if not, I imagine scalaxb would accept a patch and get a release out quickly.

I also noticed it generates immuatable case classes. But it does have a `scalaxbGenerateMutable` setting to make the case members `var`s if that's preferred

Plus, I imagine it is probably more straightforward to move our existing TDML runner functionality into those generated classes since it's already scala code, which would be nice. It would probably reduce boilerplate in our existing TDML runner lib, and make it so you could run tests without having to serialize wih saxb and then reload with the TDML runner. That's a far off goal, but scalaxb makes that more possible I think.

michael-hoke: Re: scalaxb, I had initially attempted to implement the JAXB bindings with that plugin but was unsuccessful. I could not find a way to handle duplicate names for items in the schemas. In the SbtXjc plugin, we handled that with the bindings.xjb file, but I could not find a way to provide anything like that in scalaxb. We won't run into that issue with tdml-core.xsd, but if you want to eventually go back to importing all of the schemas in JAXB, I don't believe scalaxb would currently be able to handle that.

stevedlawrence: > duplicate names for items in the schemas.

Can you clarify what this means? I'm not sure I understand.

I'm also not against the tdml-core split. It would just be a plus if we didn't need it with scalaxb, but I guess we do.

Regardless of xjc vs scalaxb, I prefer tdml-core over bindings.xjb.

michael-hoke: If two items in all imported schemas (elements, attributes, etc.) share a name, JAXB will attempt to create a binding for them at the same Java file location, which, obviously, won't work, so there needs to be a way to tell it how to rename one of them. With the SbtXjcPlugin, that is the bindings.xjb file. With the scalaxb plugin, I could not find a way to do this outside of refactoring the schemas so that everything has a unique name or by submitting a patch to the plugin to support a bindings file... Although, the tdml-core refactor also resolves this as long as we only want JAXB bindings for that subset of the TDML schema.

If you are going to eventually integrate the JAXB bindings into the TDMLRunner, I imagine that would require generating bindings for the base DFDL schemas, which do have clashing names in a few places. I'm not suggesting that we make any changes at this point - just want to point it out so that this isn't a surprise if/when you get to this point.

stevedlawrence: I see, thanks for the explanation. Sounds like the tdml-core split is a better solution.

If we did integrate the TDMLRunner into the JAXB bindings, I'm not sure we would even want bindings for the stuff we ignore now. For example, embedded DFDL schemas just want to be treated as raw strings or XML nodes or something. Those schemas are just written to a file and compiled with Daffodil, so converting them to JAXB bindings just to serialize back to XML is probably unnecessary.

But I don't think there's a plan to integrate  the TDML runner into these bindings, at least in the near term, so whatever the solution is for duplicate identifiers, it can be dealt with then.

arosien:
Yay, all tests pass @NolanMatt. Now on to linting, etc.

project/Dependencies.scala
```diff
@@ -63,4 +63,17 @@ object Dependencies {
   lazy val exi = Seq(
     "com.siemens.ct.exi" % "exificient" % "1.0.7",
   )
+
+  lazy val xjc =
```
stevedlawrence, line 67: This `xjc` dependency list is assigned to both `xjcLibs` and `libraryDependencies`. This is fine for `xjcLibs` since that setting only defines dependencies needed to run the xjc source generator. But I don't think we can have dependencies added to `libraryDependencies` that are conditional on the java version, since the version of Java we build with might be different than the version of Java a used at runtime.

In fact, since releases are built and published with Java 8, these won't be listed as actual dependencies for our official releases. So people using Java 11 or later won't have them as a dependency.

But maybe that's okay and none of these should be added to `libraryDependencies` in the first place? Maybe we document (or maybe it's obvious to JAXB users?) that if a user wants to use these xjc generated files they need to depend on those xjc libraries? Or maybe we just always require them regardless of java version?

Is this the result of https://github.com/apache/daffodil/pull/1057#discussion_r1339338337

Note, another win for scalaxb--I don't think it has any dependency complications like this.

build.sbt
```diff
@@ -211,6 +214,43 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
...
+          "java.base/java.lang=ALL-UNNAMED",
+        )
+      else Seq()),
+    Compile / xjc / sources := Seq(
+      file("daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml-core.xsd"),
+    ),
+    // Auto-generated Java files have invalid Scaladoc :(
+    Compile / doc / sources ~= (_ filterNot (_.getName endsWith ".java")),
+    // Tests use JAXB, which uses reflection, which has restrictions above JVM 17.
+    Test / fork := true,
```
stevedlawrence, line 244: I'm generally not a fan of forking tests, we've seen forking cause things to go much slower. But I guess in this specific case it's only for tdml-lib, which is small enough that it's not really noticeable. So should be fine. Just a heads up to make sure we don't do this globally.

But another win for scalaxb, no need for these Java 17 reflection options (which users running Daffodil on Java 17 would also need to add?), no need to fork tests, no Java conditional dependencies, and no need to mess with doc or xjc sources. Everything just works the way it's expected to.

If xjc is the right technology to use I guess this is fine, but there sure are a lot of hacks being added to get it working that I think just go away if we switch to scalaxb. And with the tdml-core change, the changes are very straightforward. I was able to get it working with just a few settings:

```scala
lazy val scalaxbSettings = Seq(
  Compile / scalaxb / scalaxbPackageName := "org.apache.daffodil.tdml.scalaxb",
  Compile / scalaxb / scalaxbGenerateMutable := true,
  Compile / scalaxb / sources := Seq(
    (lib / Compile / resourceDirectory).value / "org/apache/daffodil/xsd/tdml.xsd",
   ),
  // fixes issue with unused imports
  Compile / scalaxb / scalaxbHttpClientStyle := HttpClientStyle.Sync,
)
```

Granted, I didn't create any tests to see how marshalling/unmarshalling worked, but I imagine it's straightforward.

project/Dependencies.scala
stevedlawrence: Also, this adds new dependencies. Have you verified that the license are ASF compatible and if changes are needed to NOTICE/LICENSE?

michael-hoke: I believe we will need to make updates to the NOTICE/LICENSE files.

SbtXjcPlugin is BSD-3
jaxb-xjc and jaxb-impl are EDL 1.0
activation is CDDL 1.0 - this is the most restrictive, but I think we're okay because we're open source.

On the other hand, if we switch to scalaxb, that uses MIT.

However, there are dependencies listed on the github page that I don't see in Steve's code snippet. In case we need those:
org.dispatchhttp.dispatch-core - LGPL 3.0
javax.xml.bind.jaxb-api - CDDL 1.0
org.scala-lang.modules.scala-xml - Apache 2.0
org.scala-lang.scala-parser-combinators - Apache 2.0.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/JaxbTests.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  @Test def testReading(): Unit = {
+    val testSuite = JAXBContext
+      .newInstance(classOf[TestSuite])
+      .createUnmarshaller()
+      .unmarshal(getClass.getClassLoader().getResourceAsStream("test-suite/ibm-contributed/dpaext1-2.tdml"))
+      .asInstanceOf[TestSuite]
+
+    assertNotNull(testSuite)
+    assertEquals("dpaext", testSuite.suiteName)
```
stevedlawrence, line 18: I'm curious what the `dfdInfoset` part unmarshalls to, since that can contain random XML (or a string for `type="file"`) that isn't defined by the schema, it's just an XSD `any` element. Is it made available as a string, some sort of XML DOM, or is it just ignored? If ignored, that's likely a problem since users of this will probably want access to the infoset part of a TDML file.

It might be useful to modify this test to examine some of the things we think we'll need (like `dfdlnfoset`) to make sure they are unmarshalling in usable ways. And maybe also act as a reference for using this.

A related thought, if XSD `any` does parse to something useful like a string or DOM, could XSD `any` also be used to support the `defineSchema` and `defineConfig` elements? For example, what if we added a new `tdml-jaxb.xsd` file that did something like:

```xml
...
<xsd:include schemaLocation="tdml-core.xsd"/>

<element name="defineSchema" type="tns:defineSchemaType" substitutionGroup="tns:testSuiteChoices"/>
<complexType name="defineSchemaType">
  <sequence>
    <any namespace="##any" processContents="lax" minOccurs="0" maxOccurs="unbounded"/>
  </sequence>
</comlexType>
...
```

So very similar to how `dfdlInfoset` is already defined. We can configure xjc to generate code based on this new jaxb specific tdml file, and the `defineSchema` just becomes a string or DOM or however it treats XSD `any`, just like `dfdlInfoset` does.

Doesn't mean we necessarily have to support embedded configs and schemas in the vscode tool where this is planned to be used, but it at least makes it possible. And also makes it more possible to combin the existing TDML runner with the new jaxb stuff.

project/Dependencies.scala
stevedlawrence: The xjc dependencies are fine, but CDDL is category B we do need to mention CDDL so needs to be mentioned in a readme.

We already depend on scala-xml and scala-parser-combinators. LGPL is category X and is not allowed by ASF, but I don't think it's a mandatory dependency. Dispatch http looks like an http communication library library that scalaxb can generate code to support, but it looks optional to me. For example, I think it can be disabled with:

```scala
Compile / scalaxb / scalaxbGenerateDispatchClient := false
```

Similarly scalaxb also supports http4s and gigahorse, which I think are also optional with similar keys to disabled (disabled by default).  I'm also not sure where scalaxb actually uses the jaxb-api. I don't see it mentioned in the generated code. Maybe that's some an optional dependency as well?

But either way I think we're fine, we just need to make sure to update LICENSE/NOTICE appropriately if CDDL is included.

build.sbt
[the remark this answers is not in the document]
michael-hoke: It looks like the `-no-header` flag solves the issue for the .class and .jar files, but I'm not exactly sure why... I did confirm that all that flag does is strip out a comment header at the beginning of the file, so I don't know why that matters for .class files... I did confirm that the build is not reproducible without the `-no-header` flag. I also found out that there is a .class decompiler in VSCode (I think it might even be native). I decompiled both versions of a class that the `diff` command reported as different and did not see a change.

That being said, even with the `-no-header` flag, I notice differences in the generated .java files. These differences are the order of {@link ...} comments, so they also do not affect the actual code... Completely unsure as to why these comments don't matter and the header's do. The list of files that differ also varies, at least slightly, with every build.

If you can confirm that the current code on this branch creates a reproducible build for you, we'll leave `-no-header` in. If/When we attempt to switch over to scalaxb, we'll verify that we have a reproducible build.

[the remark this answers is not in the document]
stevedlawrence: I am actually not able to reproduce the non-reproducable-build issue, either with or without the `-no-header` flag. So I'm not convinced that was the actual cause (which I think makes sense--I think we agree that comments shouldn't affect the class file).

I do see the different header dates and sometimes different ordering of `@links`, but none of that seems to cause the class files to be different for me. Seems like there must be something else going on. I can't reproduce it so I'm not sure what the cause would be...

The fact that sometimes `@links` show up in a different order or in different files makes me feel like maybe there is some source of randomness or ordering that *sometimes* does cause a change to the class file. Maybe no-header fixes that as a side-effect or maybe it's just not always reproducible, like I'm having trouble with now?

stevedlawrence:
With the changes to scalaxb, is this ready for re-review?

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/JaxbTests.scala
arosien: `dfdlinfoset` is translated into:
```scala
case class DfdlInfosetType(var mixed: Seq[scalaxb.DataRecord[Any]] = Nil,
  var attributes: Map[String, scalaxb.DataRecord[Any]] = Map.empty) {
  def typeValue = attributes("@type").as[Type]
  def typeValue_=(_value: Type)(implicit evidence: scalaxb.CanWriteXML[Type]) = attributes += "@type" -> scalaxb.DataRecord(_value)
}
```

arosien:
> With the changes to scalaxb, is this ready for re-review?

Yes! See some [infoset translation](https://github.com/apache/daffodil/pull/1057#discussion_r1480298851) and here is the top-level test suite:
```scala
case class TestSuite(var testSuiteChoices: Seq[scalaxb.DataRecord[Any]] = Nil,
  var attributes: Map[String, scalaxb.DataRecord[Any]] = Map.empty) {
  def suiteName = attributes.get("@suiteName") map { _.as[String]}
  def suiteName_=(_value: Option[String])(implicit evidence: scalaxb.CanWriteXML[Option[String]]) = attributes += "@suiteName" -> scalaxb.DataRecord(_value)
  def ID = attributes.get("@ID") map { _.as[String]}
  def ID_=(_value: Option[String])(implicit evidence: scalaxb.CanWriteXML[Option[String]]) = attributes += "@ID" -> scalaxb.DataRecord(_value)
  def description = attributes.get("@description") map { _.as[String]}
  def description_=(_value: Option[String])(implicit evidence: scalaxb.CanWriteXML[Option[String]]) = attributes += "@description" -> scalaxb.DataRecord(_value)
  def defaultRoundTrip = attributes.get("@defaultRoundTrip") map { _.as[RoundTripType]}
  def defaultRoundTrip_=(_value: Option[RoundTripType])(implicit evidence: scalaxb.CanWriteXML[Option[RoundTripType]]) = attributes += "@defaultRoundTrip" -> scalaxb.DataRecord(_value)
  def defaultValidation = attributes.get("@defaultValidation") map { _.as[ValidationType]}
  def defaultValidation_=(_value: Option[ValidationType])(implicit evidence: scalaxb.CanWriteXML[Option[ValidationType]]) = attributes += "@defaultValidation" -> scalaxb.DataRecord(_value)
  def defaultConfig = attributes.get("@defaultConfig") map { _.as[String]}
  def defaultConfig_=(_value: Option[String])(implicit evidence: scalaxb.CanWriteXML[Option[String]]) = attributes += "@defaultConfig" -> scalaxb.DataRecord(_value)
  def defaultImplementations = attributes.get("@defaultImplementations") map { _.as[Seq[ImplementationItem]]}
  def defaultImplementations_=(_value: Option[Seq[ImplementationItem]])(implicit evidence: scalaxb.CanWriteXML[Option[Seq[ImplementationItem]]]) = attributes += "@defaultImplementations" -> scalaxb.DataRecord(_value)
}
```

Users of these data structures will need to build up, for example, `ParserTestCaseType` values, serialize them via `scalaxb.toXML`, and push them into the `testSuiteChoices` member.

stevedlawrence:
Can you please rebase to the latest main branch and resolve the conflicts?

stevedlawrence (changes requested):
I think this could just use a bit more extensive test, and maybe some license questions?

Also, it looks like you squashed into a single commit, but it still needs to be rebased to fix conflicts.

Also, can you update the commit message to include a bug reference, right now it has a broken link. I think DAFFODIL-2830 is the right bug?

arosien:
> I think this could just use a bit more extensive test, and maybe some license questions?
> 
> Also, it looks like you squashed into a single commit, but it still needs to be rebased to fix conflicts.
> 
> Also, can you update the commit message to include a bug reference, right now it has a broken link. I think [DAFFODIL-2830](https://issues.apache.org/jira/browse/DAFFODIL-2830) is the right bug?

I don't know why it says build.sbt has conflicts, locally I don't see them. I'll poke around more.

Will update the bug reference.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/ScalaxbTests.scala
```diff
@@ -0,0 +1,23 @@
...
+          getClass
+            .getClassLoader()
+            .getResourceAsStream("test-suite/ibm-contributed/dpaext1-2.tdml"),
+        ),
+      )
+
+    assertNotNull(testSuite)
+    assertEquals(Some("dpaext"), testSuite.suiteName)
+  }
+}
```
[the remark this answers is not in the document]
arosien: While trying to update the tests, I tried to parse IPv4.tdml from another module's tests and discovered scalaxb does not properly ignore XML comments when parsing, failing to parse. This is likely a bug in the generated parsing code of scalaxb, and I'll file an issue, but it may be a blocker for us. What do you think if this is the case @stevedlawrence @mbeckerle?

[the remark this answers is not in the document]
mbeckerle: We have various tools in daffodil-misc/xml/XMLUtils.scala that strip out comments, strip out processing instructions, etc. 
You may need to do that pass first in order to be able to read in TDML.

This is a really common thing in XML processing that people don't accommodate comments, which are part of the XML Infoset. As are EntityRefs, etc. Somewhere in the daffodil tests we have a DFDL schema that has a comment everwhere that a comment is allowed. To make sure we tolerate them in all the places they can appear. 

Are you parsing TDML, or reading and writing it back?

Reading one can strip out this stuff, but if you are rewriting the file this stuff has to be preserved. TDML is not an object model. Ex: reading and writing this:
```
<str><![CDATA[big long string with
significant whitespace
inside it.
]]></str>
```
And getting it to round trip with that CDATA bracketing in it, all whitespace preserved, etc. .... that's going to be tricky.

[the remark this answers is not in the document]
stevedlawrence: > but if you are rewriting the file this stuff has to be preserved

I don't think jaxb/scalaxb has ways to preserve comments/cdata etc. It's all converted to Java objects that don't retain that information.

I think this just needs to be a limitation for people wanting to use jaxb/scalaxb, they need to strip comments prior to unmarshalling, and roundtripping is not guaranteed when marshalling.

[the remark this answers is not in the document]
mbeckerle: @arosien what is the use case here. I had thought the purpose of this scalaxb stuff was to enable adding TDML support to the VSCode extension, and that would mean both reading and writing out TDML.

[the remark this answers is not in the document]
michael-hoke: The use case for the extension is to enable reading and writing of **generated** TDML files, so the comments aren't an immediate issue for us. We don't expect users to have or use non-generated TDML files in order to use the features in the extension.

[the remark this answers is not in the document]
stevedlawrence: Can VS Code extensions use Java/Scala libraries? I thought everything was typescript?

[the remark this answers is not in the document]
michael-hoke: The frontend is typescript. The backend that communicates with Daffodil is Scala.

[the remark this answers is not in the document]
arosien: > The use case for the extension is to enable reading and writing of generated TDML files, so the comments aren't an immediate issue for us. We don't expect users to have or use non-generated TDML files in order to use the features in the extension.

I don't quite think this is true. We happen to most often handle completely generated TDML files, but it should be the case that someone with a pre-existing non-generated TDML file would want to add a generated test case to the file, in which case the app using this lib should either raise an error if there are comments, or warn that we're going to trash them.

> @arosien what is the use case here. I had thought the purpose of this scalaxb stuff was to enable adding TDML support to the VSCode extension, and that would mean both reading and writing out TDML.

It's both reading and writing in the daffodil-vscode project, but I don't think that matters for the changes in daffodil proper. We can document the limitation with comments somewhere and downstream projects can choose to raise an error when encountering comments, filter them out, etc.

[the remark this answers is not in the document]
arosien: > Can VS Code extensions use Java/Scala libraries? I thought everything was typescript?

> The frontend is typescript. The backend that communicates with Daffodil is Scala.

Oh, I forgot, this is my favorite part... We currently launch the daffodil debugger to create and append TDML files, which is IMO rather excessive, given that we could alternatively read and write XML files from the typescript extension itself. (There are a number of typescript-based XSD XML parser generator libraries out there.) But we can perhaps discuss that in a different forum.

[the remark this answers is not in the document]
stevedlawrence: > We currently launch the daffodil debugger to create and append TDML files, which is IMO rather excessive, given that we could alternatively read and write XML files from the typescript extension itself.

Yeah, a pure typescript implementation has some clear advantages. That might be a better approach.

Another thought, there have been some discussions about an alternative DFDL syntax since there's a lot of dislike of XML. Maybe it's also worth considering a different test format? Maybe json (or something similar) would be sufficient, which I believe typescript has much better support for. That probably requires an even longer discussion and is maybe a more long term thing, if even done, but possibly something worth considering.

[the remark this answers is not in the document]
michael-hoke: > it should be the case that someone with a pre-existing non-generated TDML file would want to add a generated test case to the file, in which case the app using this lib should either raise an error if there are comments, or warn that we're going to trash them.

Agreed that we should, but the immediate use case is for generated files. The hope is that people with custom non-generated files that don't work out of the box create an issue and we can deal with the edge cases that way. I think having some sort of message/warning if comments (or anything else that we can't currently handle) are detected is a really good starting point that we're not currently doing - a ticket to track this would probably be useful.

> We currently launch the daffodil debugger to create and append TDML files

This isn't quite as true as it was before. With the GUI implemented, we've integrated the TDML generate/append more into the actual DFDL parse. By default, a TDML file is generated for every parse - it's not a separate action anymore. The generate/append actions work off of the last parse that was ran instead of requiring the user to do another parse.

So, when the user hits the generate/append button, it's all handled on the typescript side now. Even though that's not the complete picture, it's a step towards decoupling the TDML from the backend if that's a direction we eventually move to... Although, I don't see how we can do that as a TDML test case (or at least for how we're using it) is implicitly tied to a DFDL parse. Maybe there's a path to take if we add a way to have the backend send information back to the frontend after a parse?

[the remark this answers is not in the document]
mbeckerle: > It's both reading and writing in the [daffodil-vscode](https://issues.apache.org/jira/browse/DAFFODIL-vscode) project, but I don't think that matters for the changes in daffodil proper. We can document the limitation with comments somewhere and downstream projects can choose to raise an error when encountering comments, filter them out, etc.

So when scalaxb writes out an infoset, strings in the infoset are likely to contain "<", ">", "&", etc. So they need XML escapifying when output. I would assume this would happen automatically. 

But note that if you read in TDML and an infoset is embedded and uses `<![CDATA[...]]>` stuff, that it will write out as the string contents (without the CDATA bracketing) but escapified. 

I'm ok with that. It's just fragile if XML tooling decides it is ok to mess with the whitespace of the TDML file. E.x., if you output the XML back to a file by pretty printing it, then data might be corrupted by whitespace changes. 

This is just a really painful area of experience with XML. It's just not a great data language because it was designed to be a human authored markup language. Hence, the whitespace insensitivity, line-ending changes, etc.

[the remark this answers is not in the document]
arosien: > So, when the user hits the generate/append button, it's all handled on the typescript side now.

ah, i forgot about that, thanks @michael-hoke

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/scalaxb/ScalaxbTests.scala
```diff
@@ -0,0 +1,142 @@
...
+            .getClassLoader()
+            .getResourceAsStream("test-suite/ibm-contributed/dpaext1-2.tdml"),
+        ),
+      )
+
+    assertNotNull(testSuite)
+    assertEquals(Some("dpaext"), testSuite.suiteName)
+  }
+
+  // currently failing, not sure why scalaxb isn't handling comments. is the xsd too strict and somehow excludes allowing comments?
```
arosien, line 39: This test fails due to XML comments in the TDML file. It seems fairly normal to have comments in XML, so this seems like a bad blocker for the use of scalaxb. But as noted, perhaps this is due to an overly strict XSD file? Is that possible?

stevedlawrence: I'mt not sure XSD can allow or disallow comments, I think they're just baked into XML. So it probably means scalaxb just can't handle comments.

Ideally, we could open a ticket with scalaxb to see if they could add the ability to ignored comments.

Alternatively, and maybe an interim solution until scalaxb ignores comments, users of `scalaxb.fromXML[TestSuite](...)`  could filter the NodeSeq to remove comments before passing it to scalaxb. Scala-xml has a [Transformer API](https://github.com/scala/scala-xml/tree/main/shared/src/main/scala/scala/xml/transform) that should make it not too painful. It's definitely annoying to have to worry about that, but I'm not sure it's necessarily a blocker as long as we document it. E.g.
```scala
val node = scala.xml.XML.load(...)
val noComments = CustomRemoveCommentsTransformer.transform(node)
val testSuite = scalaxb.fromXML[TestSuite](noComments)
```

shanedell:
@stevedlawrence I know there was a discuss out for releasing Daffodil 3.7.0. Would this possibly be a PR we could get in before the actual release or where do you stand on what still needs done for this to meet requirements?

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/scalaxb/ScalaxbTests.scala
arosien: I tested the [example xsd](https://scalaxb.org/running-scalaxb) with some XML with various comments inside it, and the scalaxb generated code handled the comments just fine, so I suspect there's something more complicated going on with the TDML xsd. Ick.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/scalaxb/ScalaxbTests.scala
```diff
@@ -0,0 +1,142 @@
...
+          </tdml:infoset>
+        </tdml:parserTestCase>
+      )
+    val actualParserTestCase: ParserTestCaseType =
+      testSuite
+        .testSuiteChoices
+        .headOption
+        .map(_.as[ParserTestCaseType])
+        .getOrElse(throw new AssertionError("first test case is missing")) // fail() returns Unit so mimic JUnit
+    assertEquals(expectedParserTestCase, actualParserTestCase)
```
arosien, line 140: This test case also fails, but it is going to take some time to figure out what is different. I suspect it some kind of namespace problem....

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/scalaxb/ScalaxbTests.scala
mbeckerle: XSD cannot talk about comments in the XML document. They're orthogonal to schema validity.

stevedlawrence:
> @stevedlawrence I know there was a discuss out for releasing Daffodil 3.7.0. Would this possibly be a PR we could get in before the actual release or where do you stand on what still needs done for this to meet requirements?

Based on @arosien's comments, it sounds like there are still some open issues to resolve with this PR? There's also merge conflicts and a potential licensing issue to figure out. So it might be best to push to the next release. Were thinking we might try to get 3.8.0 out quicker than usual with the new layer API. We could try to get this in that release so you wouldn't have to wait too long.

mbeckerle:
Since I'm re-reading this ticket.... I had a thought. Rather than making the VSCode read/write TDML, which we can see using the scalaxb approach, is hard to make compatible with human authored stuff like CDATA bracketing, comments, etc,, why don't we define this to be different from HA (Human Authored) TDML, can explicitly call it MG (Machine Generated) TDML. This is really just a shift in perspective, but we're giving ourselves the freedom to simply be incompatible with Human-Authored TDML, and we can add features like an explicit '<tdml:comment>.' element to overcome the XML-Comments issue.

michael-hoke:
> Since I'm re-reading this ticket.... I had a thought. Rather than making the VSCode read/write TDML, which we can see using the scalaxb approach, is hard to make compatible with human authored stuff like CDATA bracketing, comments, etc,, why don't we define this to be different from HA (Human Authored) TDML, can explicitly call it MG (Machine Generated) TDML. This is really just a shift in perspective, but we're giving ourselves the freedom to simply be incompatible with Human-Authored TDML, and we can add features like an explicit 'tdml:comment.' element to overcome the XML-Comments issue.

This is more or less how it is now. Our use cases currently assume MG TDML files, and we've been testing with MG files. I think we've been trying to get basic parsing working with HA files because, in theory, we should be able to load both in. We should be able append/execute any TDML as well, but, to my knowledge, we haven't tested with that yet.

daffodil-tdml-lib/src/test/scala/org/apache/daffodil/tdml/scalaxb/ScalaxbTests.scala
arosien: @stevedlawrence, you from the past helped me find the problem! https://stackoverflow.com/questions/52984363/scala-xml-child-method-from-an-xml-node-gets-trailling-white-space

This fixes the reading-comments issue, along with my [comparisons-failing issue](https://github.com/apache/daffodil/pull/1057#discussion_r1538102297).

arosien: Also @mbeckerle helped in a previous comment https://github.com/apache/daffodil/pull/1057#discussion_r1483645379 🙇

arosien:
Closing for multiple reasons:
- TDML XML object mapping support isn't used, or can be avoided, in downstream Scala projects like [daffodil-vscode](https://github.com/apache/daffodil-vscode). See [above](https://github.com/apache/daffodil/pull/1057#discussion_r1484667731) for more context.
- The original JAXB solution was deemed to have too particular build configuration given differences between JVM versions, so scalaxb was chosen instead. But since JVM 21 was introduced it also appears to have JVM version dependent build configuration, which is slightly annoying. It's not a blocker, however.
- Given the oddities of XML object binding, the need for an alternative was proposed [above](https://github.com/apache/daffodil/pull/1057#discussion_r1484660294), say, JSON, which may fit more downstream languages like Scala and Typescript.

PR #1058, opened 2023-07-20
Better Error Reporting for Library API

olabusayoT:
- replace abort in usageError/usage with IllegalStateException
- Add diagnostics error message to place where usage is called
- add test

DAFFODIL-2072

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
@@ -110,7 +110,7 @@ final class ProcessorFactory private (
   override def onPath(xpath: String): DFDL.DataProcessor = sset.onPath(xpath)
 
   override def forLanguage(language: String): DFDL.CodeGenerator = {
-    Assert.usage(!isError)
+    checkNotError()
 
     // Do a poor man's pluggable code generator implementation - we can replace
     // it after we observe how the validator SPI evolves and wait for our
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/SchemaSetRuntime1Mixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/SchemaSetRuntime1Mixin.scala
@@ -19,7 +19,6 @@ package org.apache.daffodil.core.runtime1
 
 import org.apache.daffodil.core.dsom.SchemaSet
 import org.apache.daffodil.core.grammar.VariableMapFactory
-import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.util.Logger
 import org.apache.daffodil.runtime1.api.DFDL
 import org.apache.daffodil.runtime1.processors.DataProcessor
@@ -68,7 +67,7 @@ trait SchemaSetRuntime1Mixin {
   }.value
 
   def onPath(xpath: String): DFDL.DataProcessor = {
-    Assert.usage(!isError)
+    checkNotError()
     if (xpath != "/")
       root.notYetImplemented("""Path must be "/". Other path support is not yet implemented.""")
     val rootERD = root.elementRuntimeData
--- a/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
+++ b/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
@@ -1217,6 +1217,22 @@ public class TestJavaAPI {
     }
 
     @Test
+    public void testJavaAPI27() throws IOException {
+        org.apache.daffodil.japi.Compiler c = Daffodil.compiler();
+        java.io.File schemaFile = getResource("/test/japi/mySchema6.dfdl.xsd");
+        ProcessorFactory pf = c.compileFile(schemaFile);
+        assertTrue(pf.isError());
+        try {
+            pf.onPath("/");
+        } catch (Exception e) {
+            Throwable cause = e.getCause();
+            assertTrue(cause.toString().contains("Must call isError"));
+            assertTrue(cause.getCause().toString().contains("Schema Definition Error"));
+            assertTrue(cause.getCause().toString().contains("Cannot resolve the name 'tns:nonExistent'"));
+        }
+    }
+
+    @Test
     public void testJavaAPINullXMLTextEscapeStyle() throws IOException, ClassNotFoundException {
         ByteArrayOutputStream xmlBos = new ByteArrayOutputStream();
         try {
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/api/Diagnostic.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/api/Diagnostic.scala
@@ -308,4 +308,18 @@ trait WithDiagnostics {
    * then one can proceed to run the compiled entity.
    */
   def isError: Boolean
+
+  /**
+   * Helper method to check that isError is false, if not it throws
+   * a usage error caused by illegal state caused by a compilation error
+   */
+  def checkNotError(): Unit = {
+    Assert.usage(
+      !isError,
+      new IllegalStateException(
+        "Must call isError() to ensure there are no errors",
+        getDiagnostics.find(_.isError).get,
+      ),
+    )
+  }
 }
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
@@ -51,7 +51,10 @@ abstract class UnsuppressableException(m: String, th: Throwable) extends Excepti
   def this(th: Throwable) = this(null, th)
 }
 
-class UsageException(m: String) extends UnsuppressableException(m)
+class UsageException(m: String, th: Throwable) extends UnsuppressableException(m, th) {
+  def this(th: Throwable) = this(null, th)
+  def this(m: String) = this(m, null)
+}
 class NotYetImplementedException(m: String)
   extends UnsuppressableException("Not yet implemented: " + m)
 class Abort(m: String, th: Throwable) extends UnsuppressableException(m, th) {
@@ -83,13 +86,18 @@ object Assert extends Assert {
    */
   def usageErrorUnless(testAbortsIfFalse: Boolean, message: String): Unit =
     macro AssertMacros.usageMacro2
+  def usageErrorUnless(testAbortsIfFalse: Boolean, cause: Throwable): Unit =
+    macro AssertMacros.usageMacro2Cause
   def usageErrorUnless(testAbortsIfFalse: Boolean): Unit = macro AssertMacros.usageMacro1
 
   /**
    * Brief form
    */
   def usage(testAbortsIfFalse: Boolean, message: String): Unit = macro AssertMacros.usageMacro2
-  def usage(testAbortsIfFalse: Boolean): Unit = macro AssertMacros.usageMacro1
+  def usage(testAbortsIfFalse: Boolean, cause: Throwable): Unit =
+    macro AssertMacros.usageMacro2Cause
+  def usage(testAbortsIfFalse: Boolean): Unit =
+    macro AssertMacros.usageMacro1
 
   /**
    * test for something that the program is supposed to be ensuring.
@@ -128,7 +136,15 @@ object Assert extends Assert {
   //
 
   def usageError(message: String = "Usage error."): Nothing = {
-    abort(message)
+    toss(new UsageException(message))
+  }
+
+  def usageError(cause: Throwable): Nothing = {
+    toss(new UsageException(cause))
+  }
+
+  def usageError2(message: String = "Usage error.", testAsString: String): Nothing = {
+    usageError(message + " (" + testAsString + ")")
   }
 
   def nyi(info: String): Nothing = {
--- a/daffodil-lib/src/test/scala/org/apache/daffodil/lib/macros/TestAssertMacros.scala
+++ b/daffodil-lib/src/test/scala/org/apache/daffodil/lib/macros/TestAssertMacros.scala
@@ -41,7 +41,7 @@ class TestAssertMacros {
   }
 
   @Test def testUsage1Arg(): Unit = {
-    val e = intercept[Abort] {
+    val e = intercept[UsageException] {
       Assert.usage(if (1 == x) true else false)
     }
     val msg = e.getMessage()
@@ -54,7 +54,7 @@ class TestAssertMacros {
   }
 
   @Test def testUsage2Arg(): Unit = {
-    val e = intercept[Abort] {
+    val e = intercept[UsageException] {
       Assert.usage(if (1 == x) true else false, "foobar")
     }
     val msg = e.getMessage()
@@ -109,5 +109,14 @@ class TestAssertMacros {
     assertTrue(msg.contains("else"))
     assertTrue(msg.contains("false"))
     assertTrue(msg.contains("foobar"))
+  }
+
+  @Test def testUsage2ArgCause(): Unit = {
+    val e = intercept[UsageException] {
+      Assert.usage(if (1 == x) true else false, new Exception("test"))
+    }
+    val cause = e.getCause.toString
+    assertTrue(cause.contains("Exception"))
+    assertTrue(cause.contains("test"))
   }
 }
--- a/daffodil-macro-lib/src/main/scala/org/apache/daffodil/lib/exceptions/AssertMacros.scala
+++ b/daffodil-macro-lib/src/main/scala/org/apache/daffodil/lib/exceptions/AssertMacros.scala
@@ -28,7 +28,17 @@ object AssertMacros {
 
     q"""
     if (!($testAbortsIfFalse)) {
-         Assert.abort2("Usage error: " + $message, $testAsString)
+         Assert.usageError2("Usage error: " + $message, $testAsString)
+    }
+    """
+  }
+
+  def usageMacro2Cause(c: Context)(testAbortsIfFalse: c.Tree, cause: c.Tree): c.Tree = {
+    import c.universe._
+
+    q"""
+    if (!($testAbortsIfFalse)) {
+         Assert.usageError($cause)
     }
     """
   }
@@ -40,7 +50,7 @@ object AssertMacros {
 
     q"""
     if (!($testAbortsIfFalse)) {
-         Assert.abort("Usage error: " + $testAsString)
+         Assert.usageError("Usage error: " + $testAsString)
     }
     """
   }
@@ -100,5 +110,4 @@ object AssertMacros {
     }
     """
   }
-
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
@@ -340,8 +340,7 @@ class DataProcessor(
    * runtime. Instead we deal with success and failure statuses.
    */
   def parse(input: InputSourceDataInputStream, output: InfosetOutputter): DFDL.ParseResult = {
-    Assert.usage(!this.isError)
-
+    checkNotError()
     // If full validation is enabled, tee all the infoset events to a second
     // infoset outputter that writes the infoset to a byte array, and then
     // we'll validate that byte array upon a successful parse.
@@ -501,7 +500,7 @@ class DataProcessor(
   }
 
   def unparse(inputter: InfosetInputter, output: DFDL.Output): DFDL.UnparseResult = {
-    Assert.usage(!this.isError)
+    checkNotError()
     val outStream = java.nio.channels.Channels.newOutputStream(output)
     unparse(inputter, outStream)
   }
--- a/daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
+++ b/daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
@@ -28,7 +28,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.Paths
 import javax.xml.XMLConstants
 
-import org.apache.daffodil.lib.exceptions.Abort
+import org.apache.daffodil.lib.exceptions.UsageException
 import org.apache.daffodil.sapi.Daffodil
 import org.apache.daffodil.sapi.DaffodilParseXMLReader
 import org.apache.daffodil.sapi.DaffodilUnhandledSAXException
@@ -887,7 +887,7 @@ class TestScalaAPI {
     try {
       res = dp.parse(input, outputter)
     } catch {
-      case e: Abort => {
+      case e: UsageException => {
         assertTrue(e.getMessage().contains("Usage error"))
         assertTrue(e.getMessage().contains("invalid input source"))
       }
@@ -1205,6 +1205,28 @@ class TestScalaAPI {
     val ur = dp.unparse(inputter, wbc)
     assertFalse(ur.isError)
     assertEquals(expectedData, bos.toString())
+  }
+
+  @Test
+  def testScalaAPI26(): Unit = {
+    val c = Daffodil.compiler()
+    val schemaFile = getResource("/test/sapi/mySchema6.dfdl.xsd")
+    val pf = c.compileFile(schemaFile)
+    assertTrue(pf.isError())
+    try {
+      pf.onPath("/")
+    } catch {
+      case e: UsageException => {
+        val cause = e.getCause
+        assertTrue(cause.toString.contains("Must call isError"))
+        assertTrue(
+          cause.getCause.toString.contains("Schema Definition Error"),
+        )
+        assertTrue(
+          cause.getCause.toString.contains("Cannot resolve the name 'tns:nonExistent'"),
+        )
+      }
+    }
   }
 
   @Test
--- a/daffodil-io/src/test/scala/org/apache/daffodil/io/TestInputSourceDataInputStream2.scala
+++ b/daffodil-io/src/test/scala/org/apache/daffodil/io/TestInputSourceDataInputStream2.scala
@@ -18,7 +18,7 @@
 package org.apache.daffodil.io
 
 import org.apache.daffodil.lib.Implicits._
-import org.apache.daffodil.lib.exceptions.Abort
+import org.apache.daffodil.lib.exceptions.UsageException
 import org.apache.daffodil.lib.util.MaybeULong
 
 import org.junit.Assert._
@@ -107,7 +107,7 @@ class TestInputSourceDataInputStream2 {
     assertEquals(81, dis.bitLimit1b.get)
     assertEquals(10, dis.bytePos0b)
     dis.reset(m1)
-    intercept[Abort] {
+    intercept[UsageException] {
       dis.reset(m2)
     }
   }
--- a/daffodil-japi/src/test/resources/test/japi/mySchema6.dfdl.xsd
+++ b/daffodil-japi/src/test/resources/test/japi/mySchema6.dfdl.xsd
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<schema xmlns="http://www.w3.org/2001/XMLSchema"
+  targetNamespace="http://example.com" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tns="http://example.com">
+  
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format ref="tns:GeneralFormat" />
+    </appinfo>
+  </annotation>
+  
+  <include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+  
+  <element name="e2" dfdl:lengthKind="delimited" type="tns:nonExistent"/>
+
+</schema>
--- a/daffodil-sapi/src/test/resources/test/sapi/mySchema6.dfdl.xsd
+++ b/daffodil-sapi/src/test/resources/test/sapi/mySchema6.dfdl.xsd
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<schema xmlns="http://www.w3.org/2001/XMLSchema"
+  targetNamespace="http://example.com" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:tns="http://example.com">
+  
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format ref="tns:GeneralFormat" />
+    </appinfo>
+  </annotation>
+  
+  <include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+  
+  <element name="e2" dfdl:lengthKind="delimited" type="tns:nonExistent"/>
+
+</schema>


daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
```diff
@@ -110,7 +110,7 @@ final class ProcessorFactory private (
   override def onPath(xpath: String): DFDL.DataProcessor = sset.onPath(xpath)
 
   override def forLanguage(language: String): DFDL.CodeGenerator = {
-    Assert.usage(!isError)
+    Assert.usage(!isError, getDiagnostics.mkString)
```
stevedlawrence, line 113: `Assert.usage` means that the user calling this API used he API incorrectly. So instead of displaying the diagnostics associated with the failed ProcessorFactory, I think the error message wants to be about incorrect API usage. In this case a usage failure means the user did not check `isError` before calling the `forLanguage` function. So the message probably wants to instead be something like:

> ProcessorFactory forLanguage() must not be called when isError is true.

I'm not sure I love that, but something along those lines.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -68,7 +68,7 @@ trait SchemaSetRuntime1Mixin {
   }.value
 
   def onPath(xpath: String): DFDL.DataProcessor = {
-    Assert.usage(!isError)
+    Assert.usage(!isError, getDiagnostics.mkString)
```
stevedlawrence, line 71: Same comment as above, but about the `onPath` function instead of `forLanguage`.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
```diff
@@ -128,7 +128,7 @@ object Assert extends Assert {
   //
 
   def usageError(message: String = "Usage error."): Nothing = {
-    abort(message)
+    toss(new IllegalStateException(message + "\n" + shortBacktrace))
```
stevedlawrence, line 131: Do we need to include shortBacktrace? I think the `IllegalStateException` should contain a stack trace that API users can output if they want.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/lib/exceptions/AssertMacros.scala
```diff
@@ -28,7 +28,7 @@ object AssertMacros {
 
     q"""
     if (!($testAbortsIfFalse)) {
-         Assert.abort2("Usage error: " + $message, $testAsString)
+         throw new IllegalStateException("Usage error: " + $message + $testAsString)
```
stevedlawrence, line 31: Do we want to remove the testAsString? For an API user, that really isn't going to mean very much. Instead the message should be descriptive enough to tell the user how to fix the issue?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
```diff
@@ -340,7 +340,7 @@ class DataProcessor(
    * runtime. Instead we deal with success and failure statuses.
    */
   def parse(input: InputSourceDataInputStream, output: InfosetOutputter): DFDL.ParseResult = {
-    Assert.usage(!this.isError)
+    Assert.usage(!this.isError, this.getDiagnostics.mkString)
```
stevedlawrence, line 343: Same comment as above, these message want to be about not calling parse/unparse when DataProcessor.isError is true.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
```diff
@@ -501,7 +501,7 @@ class DataProcessor(
   }
 
   def unparse(inputter: InfosetInputter, output: DFDL.Output): DFDL.UnparseResult = {
-    Assert.usage(!this.isError)
+    Assert.usage(!this.isError, this.getDiagnostics.mkString)
```
stevedlawrence, line 504: Grepping through the code, there are a ton of places where use have usage checks. I *think* all of the places except for the ones in this PR indicate a bug in Daffodil if they fail. However, the usage checks in this PR indicate a API user doing something wrong.

So we sort of have two different kinds of usage checks. One where we're doing internal Daffodil checks and one 
doing external API checks. I wonder if the former wants to stay as aborts, and only these user API ones want to become IllegalStateExceptions? Maybe using Assert.usage in the places in this PR is wrong and they should be something else? Or maybe all the internal ones want to be Assert.invariant? Or maybe we need two different kinds of usage functions, one for internal API and one for public API checks?

stevedlawrence (commented):
Suggest changes to the usage messages. I also unsure if we need to differentiate different kinds of usage errors, i.e. public vs private API usage errors?

tuxji (commented):
Does this PR change all Assert.usage calls to Assert.invariant?  Should some places still call Assert.usage or if not, should this PR remove the Assert.usage function too?  That's the only concern I can think of; if Steve is satisfied that you've addressed all of his comments, I'll be satisfied too.

olabusayoT:
> Does this PR change all Assert.usage calls to Assert.invariant? Should some places still call Assert.usage or if not, should this PR remove the Assert.usage function too? That's the only concern I can think of; if Steve is satisfied that you've addressed all of his comments, I'll be satisfied too.

It changes the non-API/internal usage calls to invariant. Any API or user facing code (e.g TDMLRunner), we still use usage.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataInputStreamImplMixin.scala
```diff
@@ -32,7 +32,7 @@ trait DataInputStreamImplMixin
   }
 
   final override def isAligned(bitAlignment1b: Int): Boolean = {
-    Assert.usage(bitAlignment1b >= 1)
+    Assert.invariant(bitAlignment1b >= 1)
```
mbeckerle, line 35: Usage is preferred for arg checks.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -333,12 +333,12 @@ trait DataOutputStreamImplMixin
   def fragmentLastByteLimit = fragmentLastByteLimit_
 
   def setFragmentLastByte(newFragmentByte: Int, nBitsInUse: Int): Unit = {
-    Assert.usage(nBitsInUse >= 0 && nBitsInUse <= 7)
-    Assert.usage(
+    Assert.invariant(nBitsInUse >= 0 && nBitsInUse <= 7)
```
mbeckerle, line 336: Should be usage

tuxji (approved):
+1

OK, if Steve is satisfied with the changes, I'm satisfied as well.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence, line 1: There are a still a number of places in the TDMLRunner where Assert.usage is called. I wonder if those want to be invariants if they should never happen? Or alternatively if they could happen due to a bad TDML file, I wonder if we should be throwing a TDMLException? I think maybe the rule is something like:
- A check that should never fail unless there is a bug in the TDML Runner: Assert.invariant
- A check that should only fail if the TDML file is invalid: throw TDMLException (I think? is there precedence for what to do when we detect invalid TDML content?)

I'm not sure if there are functions in the TMDL library that could be called in a way where it's in an illegal state like with the DataProcessor and ProcessorFactor objects. If we create a Runner without exceptions, all of its public methods should be safe to call. I don't think we ever return a Runner object that represents an error and you cant call certain functions like with DP/PF.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -956,8 +956,8 @@ trait DataOutputStreamImplMixin
     bitLengthFrom1To64: Int,
     finfo: FormatInfo,
   ): Boolean = {
-    Assert.usage(bitLengthFrom1To64 >= 1 && bitLengthFrom1To64 <= 64)
-    Assert.usage(isWritable)
+    Assert.invariant(bitLengthFrom1To64 >= 1 && bitLengthFrom1To64 <= 64)
```
mbeckerle, line 959: Arg checks should be usage.

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -1164,15 +1164,15 @@ trait DataOutputStreamImplMixin
...
 
   override final def resetMaybeRelBitLimit0b(savedBitLimit0b: MaybeULong): Unit = {
-    Assert.usage(isWritable)
+    Assert.invariant(isWritable)
     setMaybeRelBitLimit0b(savedBitLimit0b, true)
   }
 
   final override def isAligned(bitAlignment1b: Int): Boolean = {
-    Assert.usage(bitAlignment1b >= 1)
+    Assert.invariant(bitAlignment1b >= 1)
```
mbeckerle, line 1172: usage

daffodil-io/src/main/scala/org/apache/daffodil/io/DataOutputStreamImplMixin.scala
```diff
@@ -1150,7 +1150,7 @@ trait DataOutputStreamImplMixin
         // pastData from it as part of what it displays.
         areDebugging,
     )
-    Assert.usage(nBytesRequested >= 0)
+    Assert.invariant(nBytesRequested >= 0)
```
mbeckerle, line 1153: usage

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -296,15 +296,15 @@ class DirectOrBufferedDataOutputStream private[io] (
   }
 
   override def setJavaOutputStream(newOutputStream: java.io.OutputStream): Unit = {
-    Assert.usage(newOutputStream ne null)
+    Assert.invariant(newOutputStream ne null)
```
mbeckerle, line 299: usage

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIinputStreams.scala
```diff
@@ -99,8 +99,8 @@ object BoundaryMarkLimitingStream {
     targetChunkSize: Int = 32 * 1024,
   ) = {
 
-    Assert.usage(targetChunkSize >= 1)
-    Assert.usage(boundaryMark.length >= 1)
+    Assert.invariant(targetChunkSize >= 1)
```
mbeckerle, line 102: both usage

daffodil-io/src/main/scala/org/apache/daffodil/io/LimitingJavaIinputStreams.scala
```diff
@@ -162,8 +162,8 @@ class RegexLimitingStream(
   targetChunkSize: Int = 32 * 1024,
 ) extends InputStream {
 
-  Assert.usage(targetChunkSize >= 1)
-  Assert.usage(maximumLengthDelimiterExample.length >= 1)
+  Assert.invariant(targetChunkSize >= 1)
```
mbeckerle, line 165: both usage

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
```diff
@@ -128,7 +128,7 @@ object Assert extends Assert {
   //
 
   def usageError(message: String = "Usage error."): Nothing = {
-    abort(message)
+    toss(new IllegalStateException(message))
```
mbeckerle, line 131: I don't think of a usage error as an illegal state exception. If you want to use a precise Java exception class for this I think there is an IllegalArgumentException.

I *do* think of an invariant failure as an Illegal state exception. 

But we have quite a bit of code depending on the fact that we always throw only one of our own exceptions so that they can be handled uniformly (or explicitly NOT handled when somebody wants to be sure they are not being captured.)

So I think the right thing is to use a Java IllegalArgumentException as the cause of a Daffodil-specific exception like Abort. 

Let's improve the scaladoc on usage-related assert methods. They're about checking args. 

They're also primarily intended for *us* daffodil developers using our own internal methods.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/BitOrder.scala
```diff
@@ -44,7 +44,7 @@ object Bits {
   def asUnsignedByte(b: Long): Int = (b & 0xff).toInt
 
   def asSignedByte(i: Long): Byte = {
-    Assert.usage(i >= 0)
+    Assert.invariant(i >= 0)
```
mbeckerle, line 47: usage

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/BitOrder.scala
```diff
@@ -86,7 +86,7 @@ object Bits {
   }
 
   def reverseBitsWithinBytes(bb: ByteBuffer): Unit = {
-    Assert.usage(bb.position() == 0)
+    Assert.invariant(bb.position() == 0)
```
mbeckerle, line 89: usage

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/BitOrder.scala
```diff
@@ -109,7 +109,7 @@ object Bits {
   }
 
   def reverseBytes(bb: ByteBuffer): Unit = {
-    Assert.usage(bb.position() == 0)
+    Assert.invariant(bb.position() == 0)
```
mbeckerle, line 112: I'm not going to try to flag all these any more. 

If it is checking an arg, that's what Assert.usage is for. 

Perhaps we should have a synonym Assert.argCheck(...) ?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
stevedlawrence, line 1: If the `usage` functions are now for checking public API usage, then we probably always want a message to be required as a parametr. So it is maybe worth removing the functions that don't accept a message? For example, maybe this want's to be removed:

```scala
def usageErrorUnless(testAbortsIfFalse: Boolean): Unit = macro AssertMacros.usageMacro1
```

We should also add scaladoc to the usage functions to make it clear they are only for checking public API usage, and that invariants are for checking internal state/parameters.

Should we also remove the `UsageException` class if we are going to use the IllegalStateException?

daffodil-lib/src/test/scala/org/apache/daffodil/lib/macros/TestAssertMacros.scala
```diff
@@ -41,7 +41,7 @@ class TestAssertMacros {
   }
 
   @Test def testUsage1Arg(): Unit = {
-    val e = intercept[Abort] {
+    val e = intercept[IllegalStateException] {
```
mbeckerle, line 44: Not correct. Usage is not illegal state.

daffodil-macro-lib/src/main/scala/org/apache/daffodil/lib/exceptions/AssertMacros.scala
```diff
@@ -28,7 +28,7 @@ object AssertMacros {
 
     q"""
     if (!($testAbortsIfFalse)) {
-         Assert.abort2("Usage error: " + $message, $testAsString)
+         throw new IllegalStateException("Usage error: " + $message)
```
mbeckerle, line 31: If you want this to use a standard java exception it should throw an Abort with IllegalArgumentException as the cause of the exception. Otherwise I fear this is going to break quite a lot of stuff. 

Changing the class being thrown is a big change. You need to scrutinize every catch in the code base to see if it has to change correspondingly.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/LineFoldedTransformer.scala
```diff
@@ -395,9 +395,9 @@ class LineFoldedOutputStream(mode: LineFoldMode, jos: OutputStream) extends Outp
 
   override def write(bInt: Int): Unit = {
     val b = bInt.toChar
-    Assert.usage(!closed)
+    Assert.invariant(!closed)
     Assert.invariant(line.length <= lineLength)
-    Assert.usage(b >= 0)
+    Assert.invariant(b >= 0)
```
mbeckerle, line 400: This bInt check is a usage check.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/StringLengthUnparsers.scala
```diff
@@ -72,7 +72,7 @@ sealed abstract class StringSpecifiedLengthUnparserTruncateBase(
   erd: ElementRuntimeData,
 ) extends StringSpecifiedLengthUnparserBase(erd) {
 
-  Assert.usage(stringTruncationType ne TextTruncationType.None)
+  Assert.invariant(stringTruncationType ne TextTruncationType.None)
```
mbeckerle, line 75: usage

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/InfosetInputter.scala
```diff
@@ -657,7 +657,7 @@ object InfosetAccessor {
 final class Info private (val v: AnyRef) extends AnyVal with Serializable {
   def node = element
   def element = {
-    Assert.usage(v ne null)
+    Assert.invariant(v ne null)
```
mbeckerle, line 660: usage

mbeckerle (changes requested):
Not sure what to recommend here. 

Some of the changes from usage to invariant checks are sensible, but a huge number are not. 

It looks like you headed down a path of making usage error about published Daffodil API users, and invariant checks about internal Daffodil developer errors.

But that's not the intention of usage and invariant. Both are for us daffodil devs to make the code more explicit about its assumptions. 

I think the crux of fixing the bug is making sure when an API call for a Daffodil API user does not follow the required discipline (such as calling isError before calling parse or unparse) that the exception we throw should contain the list of diagnostics (the existence of which is why you are supposed to call isError first)

For example we could throw an IllegalStateException giving the cause as an exception that is a DiagnosticsListException that carries the list of diagnostics. (If there is more than one.)

mbeckerle:
> Just a couple minor comments/questions.
> 
> As to changing most usage calls to invariants, that doesn't seem totally unreasonable to me. However, I think it could be argued that the two are different things that could warrant different names. One check is bad parameters passed into internal functions, and one is checking for bad internal state. You could argue different function names might make it more clear what's being checked (e.g. internalUsage() vs invariant() or something), but I'm not sure there's a lot of extra benefit. It both cases, a failure represents an internal bug in Daffodil, so calling invariant seems reasonable, and is definiely simpler. But I don't feel strongly about it. I could be swayed either way if there is a preference for separate function names.
> 
> @mbeckerle any thoughts?

usage is "argCheck", and invariant is  "stateCheck". Both are for our own internal use. These pre-date our API modules (JAPI/SAPI). 

Arguably those should document and throw specific exceptions from each public method. And we should define the exceptions they throw and convenient checker methods directly *in those modules themselves*.

stevedlawrence:
Is there much value from having them separate function names? With a goal of simplifying things, what if we changed them all to something like `Assert.assert(...)`. For both the argCheck and stateCheck cases, we are really just doing a bugCheck, so maybe a more generic name would be reasonable for both? Based on some of your comments there are places where usage should be invariant or vice versa so there's clearly some confusing between the different names that this would alleviate.

Then the usage() function can be this "public API usage thing", or maybe a new name would be better to make it more clear that it's about incorrect public API usage.

mbeckerle:
> Is there much value from having them separate function names? With a goal of simplifying things, what if we changed them all to something like `Assert.assert(...)`. For both the argCheck and stateCheck cases, we are really just doing a bugCheck, so maybe a more generic name would be reasonable for both? Based on some of your comments there are places where usage should be invariant or vice versa so there's clearly some confusing between the different names that this would alleviate.

Yeah, in hind-sight now I think almost all those changes of usage to invariant are wrong. Just because something isn't an arg doesn't mean it's not a usage error.  

> 
> Then the usage() function can be this "public API usage thing", or maybe a new name would be better to make it more clear that it's about incorrect public API usage.

argCheck makes it clear that it's about checking arguments for proper usage. That is perhaps a useful new method. But precondition checks are another kind of usage-error check. We have lots of lazy vals in our code where there are no arguments, but there are still usage errors in that one can "use them wrong" and we want to make clear when you are using them wrong. 

If that kind of check is  not going to be usage(...) they should be either argCheck(...) or precondition(...) both of which, when they fail, issue a kind of usage error. They indicate a programmer's fairly immediate local mistake in the way the thing was called or nearby context of that call. I prefer usage(...) myself, because in lazy code often things being checked are not arguments. 

invariant() is for things that are potentially more distant from the local calling context. Really its just about making assumptions of the code explicit so if under maintenance somebody changes the way things work and things code is relying on change, that there are explicit checks to detect this. It's also used for postcondition checks - when a certain method finishes some condition should hold. 

So I'm ok with argCheck, preCondition, and invariant as names of different methods. But usage an invariant are sufficient I think. 

We should improve the scaladoc on those sorts of methods, but I'm very opposed to changing the name of 'invariant' to anything else or losing the distinction of usage errors (arg or precondition) vs. invariant failure.

stevedlawrence:
> So I'm ok with argCheck, preCondition, and invariant as names of different methods. But usage an invariant are sufficient I think.

Just to clarify, do you think combining them all into a single generic name is not sufficient? Does it matter if a bug is a usage bug or an invariant breaking bug or an invalid precondition bug?

I'm just wondering if we can simplify our lives so that every time we want to add a check for a potential bug we don't have think about what kind of check we are adding. We're simply making an assertion that some check must pass, and if at some point it doesn't then we need to investigate why.

mbeckerle:
> > So I'm ok with argCheck, preCondition, and invariant as names of different methods. But usage an invariant are sufficient I think.
> 
> Just to clarify, do you think combining them all into a single generic name is not sufficient? Does it matter if a bug is a usage bug or an invariant breaking bug or an invalid precondition bug?
> 
> I'm just wondering if we can simplify our lives so that every time we want to add a check for a potential bug we don't have think about what kind of check we are adding. We're simply making an assertion that some check must pass, and if at some point it doesn't then we need to investigate why.

I'm opposed to collapsing these kinds of assertions to just one. 

I think the distinction of usage error from invariant is helpful. Making everything just an undifferentiated assert isn't an improvement. 

"You're using it wrong" vs. "This code only makes sense if X is true" are different thought processes. I think of argCheck and precondition as things that should be explained in the scaladoc/javadoc of a method unless they're obvious from just looking at the argCheck/precondition code. E.g., putting both argCheck(x ne null) and scaladoc saying x cannot be null is just redundant. 

I think of invariant as something you probably wouldn't discuss in the scaladoc/javadoc because it's not about how to use it, it's about how the software works.

stevedlawrence:
> I'm opposed to collapsing these kinds of assertions to just one.

Fair enough. So maybe the correct approach is to create a new assertion function specifically for making these public API usage assertions that throws IllegalStateException so the API user knows to fix their code?

mbeckerle:
> > I'm opposed to collapsing these kinds of assertions to just one.
> 
> Fair enough. So maybe the correct approach is to create a new assertion function specifically for making these public API usage assertions that throws IllegalStateException so the API user knows to fix their code?

I agree. I would suggest put this right into the SAPI and JAPI modules.

stevedlawrence:
Also, since this is a change to the Scala/Java APIs, please remember to add a "Deprecation/Compatibility" section to your commit. As a reminder from the [code contributor workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow) since I always forget:

> If there are any changes that deprecate functionality or are non-backwards compatible, a section should follow labeled with the "Deprecation/Compatibility:" keyword with a description that can be copy/pasted into release notes. This should be more user focused, including what was deprecated/non-backwards compatible and a migration guide.

olabusayoT:
Noted! So I noticed we have an InvalidUsageException class in SAPI/JAPI...would this be preferable to use over java's IllegalStateException for the new assert function?

daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
```diff
@@ -110,7 +110,10 @@ final class ProcessorFactory private (
   override def onPath(xpath: String): DFDL.DataProcessor = sset.onPath(xpath)
 
   override def forLanguage(language: String): DFDL.CodeGenerator = {
-    Assert.usage(!isError)
+    Assert.argCheck(
```
stevedlawrence, line 113: Does this want to be something other than "argCheck"? To me argCheck implies we are making sure the arg passed in is valid. For example, you might do something like `argCheck(language != null)`. Feels like we still want this to be `usage()` but a public API usage variant.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -886,3 +887,18 @@ object DaffodilXMLEntityResolver {
...
+
+  /**
+   * Argument check for API functions
+   * @param testFailsIfFalse test that causes failure when false
+   * @param message message to be displayed on failure of assertion
+   */
+  def argCheck(testFailsIfFalse: Boolean, message: String): Unit = {
+    Assert.argCheck(testFailsIfFalse, message)
+  }
+}
```
stevedlawrence, line 904: Is this used anywhere? There are no coverage errors but I dont' see where it's used. Can it be removed if not used?

Note that I think if this function is actually used it will force evaluation of `message` even if the test passes because it must be passed into this function. We want to avoid that since there could be some overhead with generating those strings. The use of macros avoids that.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
stevedlawrence: I think we want to revert this change? The `usage` functions are now about internal API checks that indicate bugs in Daffoil, so an abort is probably the right thing to do.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
```diff
@@ -91,6 +91,12 @@ object Assert extends Assert {
   def usage(testAbortsIfFalse: Boolean, message: String): Unit = macro AssertMacros.usageMacro2
   def usage(testAbortsIfFalse: Boolean): Unit = macro AssertMacros.usageMacro1
 
+  /**
+   * Test for the validity of the arguments of the API functions
+   */
+  def argCheck(testThrowsIfFalse: Boolean, message: String): Unit =
```
stevedlawrence, line 97: Do we want a name other than "argCheck" for checking public API stuff? The isError check isn't checking an argument, it's just checking internal state.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
```diff
@@ -340,8 +340,10 @@ class DataProcessor(
    * runtime. Instead we deal with success and failure statuses.
    */
   def parse(input: InputSourceDataInputStream, output: InfosetOutputter): DFDL.ParseResult = {
-    Assert.usage(!this.isError)
-
+    Assert.argCheck(
+      !this.isError,
+      s"Must check isError before calling DataProcessor parse. ${this.getDiagnostics.mkString}",
```
stevedlawrence, line 345: This error message is making an assumption that the user isn't calling isError at all. But a user could call isError and have a bug where they just don't check the result correctly or call parse anyways.. So they might be confused--"but I am calling isError". Maybe a minor tweak like ?

> DataProcessor parse() must not be called if isError() returns true

It also feels odd to include diagnostics the message. Seems like this could hide the real issue of bad API usage. This could be more of an issue if there are alot of warnings mixed in, which might make it very hard to see this one like about bad API usage. Maybe we could filter the diags and show only the error? Might be even better to make the one error a `cuase` to the IllegalArgumentException. But I wonder if we should even do that? A user might see that diagnostic and fix the schema, when really they need to fix their code.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/SchemaSetRuntime1Mixin.scala
```diff
@@ -68,7 +68,10 @@ trait SchemaSetRuntime1Mixin {
   }.value
 
   def onPath(xpath: String): DFDL.DataProcessor = {
-    Assert.usage(!isError)
+    Assert.argCheck(
+      !isError,
+      s"Must check isError before calling onPath. ${getDiagnostics.mkString}",
+    )
```
stevedlawrence, line 74: Had a thought, these files are in daffodil-core and so I kind of consider them as private API. So maybe they still want to call usage()? And then the onPath/parse/unparse/etc functions in the daffodil-sapi/japi use a public usage variant that throws an IllegalStateException?

Or maybe these functions are so close to the public API that they are still considered public?

stevedlawrence (commented):
Just some minor comments

daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
olabusayoT: Thoughts on precondition()?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
olabusayoT: Does this mean for argCheck we should be throwing IllegalArgumentException?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
olabusayoT: The diagnostics were included based on this comment: https://github.com/apache/daffodil/pull/1058#pullrequestreview-1541309384, should I disregard?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
olabusayoT: Assert.argCheck calls the macro. Should we call the macro directly? I thought we needed a version of the call per this comment https://github.com/apache/daffodil/pull/1058#issuecomment-1648663730

daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
stevedlawrence: What if we are just very direct with the name? This is a function that throws an IllegalStateException, so what about `illegalState()`? This way it can be used for either public or private functions where a failed test just wants to throw this one specific exception?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
stevedlawrence: Assert.argCheck doesn't really call the macro, it **is** the macro. This function ends up being converted to something like this once the macro is expanded:

```scala
def argCheck(testFailsIfFalse: Boolean, message: String): Unit = {
  if (!testFailsIfFalse) {
    throw new IllegalStateException(message)
  }
 }
```
So calling something like
```scala
Assertions.argCheck(!isError, s"Some $expensive string to evaluate")
```
Will force evaluation of of the test and the string and pass the results into the argCheck where it then does the test/throw. So it's forcing evaluating of a string that might be expensive and likely will never even be used since the condition will rarely fail.

However, if something calls `Assert.argCheck` directly, e.g.
```scala
Assert.argCheck(!isError, s"Some $expensive string to evaluate")
```
Then macro expansion causes this to become something like this:
```scala
if (!isError) {
  throw new IllegalStateException(s"Some $expensive string to evaluate")
}
```
So if the condition never fails, we never evaluate that string.

Maybe @mbeckerle was suggesting this new function should be called from the daffodil-sapi instea of daffodil-core, so the checks are done in the public API themselves? Or if we want o have the `Assertions` thing here as part of the public API, then it needs to be something like
````scala
object Assertions {
  def argCheck(...) = macro argCheck
}
```
The macro keyword makes this a special macro, and calls of the function don't actuall call the function be instead expand the macro where the function is called. It's sort of like in-lining the function.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
stevedlawrence: Yeah, an argCheck function probably wants to either call abort or throw an IllegalArgumentException. But our uses aren't really about checking args, so I'd suggest we rename it. See other comment.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
stevedlawrence: I don't feel strongly about whether we include Diagnostics or not. I'm still not convinced it makes sense, but if @mbeckerle or others think it's important then it's fine with me.

But I would suggest that it should be a `cause` to the `IllegalStateException` and not part of the exception message. The diagnostics are only a secondary reason for throwing this exception and don't belong in the message itself.

Maybe the `illegalState` function/macro accepts a third cause parameter, and it's called something like this:
```scala
Assert.illegalState(isError, "...", diagnostics.find(_.isError).get)
```

The macro would expand to
```scala
if (isError) {
  throw new IllegalStateException("...", diagnostics.find(_.isError).get)
}
```
So the message and `diagnostics.find` (or whatever logic is needed) would only be evaluated if `isError` is true.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/SchemaSetRuntime1Mixin.scala
mbeckerle: It would not hurt to call checkNotError() here. But since this is internal code, other usage checks like that args are non-null should just be unadorned `usage(test)` only with no other clutter AFAIC. An important goal for this Assert.usage(...) and similar is no code clutter. If you even put a message string in there that's already too much.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
mbeckerle: I think this is the wrong approach. We should not have different checkers for public/non-public API.

I don't know why usage doesn't throw UsageException. Seems to me it should throw that. 

The definition of UsageException is incomplete and that's probably why. It needs to be just like Abort. 
```
class UsageException(m: String, th: Throwable) extends UnsuppressableException(m, th) {
  def this(th: Throwable) = this(null, th)
  def this(m: String) = this(m, null)
}
``` 
That way it can be thrown with a message or a cause. I think this is the right pattern any time one derives an exception/error class - you derive from the 2-arg version, and offer two single-arg constructors for message and cause.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
```diff
@@ -340,8 +340,11 @@ class DataProcessor(
    * runtime. Instead we deal with success and failure statuses.
    */
   def parse(input: InputSourceDataInputStream, output: InfosetOutputter): DFDL.ParseResult = {
-    Assert.usage(!this.isError)
-
+    Assert.illegalState(
```
mbeckerle, line 343: This line is repeated multiple places. Should just be:
```
checkNotError()
```

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
```diff
@@ -1207,6 +1207,26 @@ class TestScalaAPI {
...
+    val schemaFile = getResource("/test/sapi/mySchema6.dfdl.xsd")
+    val pf = c.compileFile(schemaFile)
+    assertTrue(pf.isError())
+    try {
+      pf.onPath("/")
+    } catch {
+      case e: IllegalStateException => {
+        val msg = e.getMessage
+        val cause = e.getCause.toString
+        assertTrue(msg.contains("ProcessorFactory onPath() must not"))
```
mbeckerle, line 1222: These couple lines checking the message... I wouldn't bother with these first two checks. Just that you got Schema Definition Error and "nonExistant" is enough.

daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
```diff
@@ -1216,6 +1216,24 @@ public void testJavaAPI26() throws IOException, ClassNotFoundException, External
...
+        org.apache.daffodil.japi.Compiler c = Daffodil.compiler();
+        java.io.File schemaFile = getResource("/test/japi/mySchema6.dfdl.xsd");
+        ProcessorFactory pf = c.compileFile(schemaFile);
+        assertTrue(pf.isError());
+        try {
+            pf.onPath("/");
+        } catch (IllegalStateException e) {
+            String message = e.getMessage();
+            String cause = e.getCause().toString();
+            assertTrue(message.contains("ProcessorFactory onPath() must not"));
```
mbeckerle, line 1230: Just test for "Schema Definition Error" and "nonExistant".

daffodil-lib/src/main/scala/org/apache/daffodil/lib/exceptions/Assert.scala
```diff
@@ -91,6 +91,12 @@ object Assert extends Assert {
   def usage(testAbortsIfFalse: Boolean, message: String): Unit = macro AssertMacros.usageMacro2
   def usage(testAbortsIfFalse: Boolean): Unit = macro AssertMacros.usageMacro1
 
+  /**
+   * Test for the validity of the state of the API functions
+   */
+  def illegalState(testThrowsIfFalse: Boolean, message: String, cause: Throwable): Unit =
```
mbeckerle, line 97: Do we have other illegal state checks than just the isError issue? 
I suggest we replace this with checkNotError().

mbeckerle (changes requested):
I really don't like crafting message strings for this sort of stuff. 

I think we're missing the ability to just pass the cause object and let the error system assemble the message string.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/api/Diagnostic.scala
```diff
@@ -308,4 +308,12 @@ trait WithDiagnostics {
    * then one can proceed to run the compiled entity.
    */
   def isError: Boolean
+
+  /**
+   * Helper method to check that isError is false, if not it throws
+   * a usage error caused by illegal state caused by a compilation error
+   */
+  def checkNotError(): Unit = {
+    Assert.usage(!isError, new IllegalStateException(getDiagnostics.find(_.isError).get))
```
stevedlawrence, line 317: We've lost the message that says the API is being used incorrectly so a developer using the API wont know what to change. The user will just see something like:

> Usage Error: InvalidStateException, cause: Schema Defition Error: ....

The SDE found in the diagnostic is not the problem, it's more like a symptom. The underlying problem is incorrect API usage, so that needs to be included in the error message. The core issue of this bug is that it's not clear what to do if the check fails and the user gets an exception, and I don't think this does anything to make that more clear.

daffodil-sapi/src/test/scala/org/apache/daffodil/example/TestScalaAPI.scala
```diff
@@ -887,7 +887,7 @@ class TestScalaAPI {
     try {
       res = dp.parse(input, outputter)
     } catch {
-      case e: Abort => {
+      case e: UsageException => {
```
stevedlawrence, line 890: This is an API change, please remember to add a Deprecation/Compatibility section to your commit message.

stevedlawrence (changes requested):

daffodil-lib/src/main/scala/org/apache/daffodil/lib/api/Diagnostic.scala
mbeckerle: We can pass the constructor of IllegalStateException both a message string AND a cause.
Then the string can explain that you must call 'isError' to ensure there are not errors, but the error creating that situation is still featured also.

olabusayoT: How's changing the definition to 

```scala
  def checkNotError(): Unit = {
    Assert.usage(!isError, new IllegalStateException(
      "Must call isError() to ensure there are no errors",
      getDiagnostics.find(_.isError).get,
    ))
  }
```

olabusayoT: The new cause will be:
```
cause = {IllegalStateException@2841} "java.lang.IllegalStateException: Must call isError() to ensure there are no errors"
 detailMessage = "Must call isError() to ensure there are no errors"
 cause = {SchemaDefinitionError@2853} Schema Definition Error: Error loading schema due to org.xml.sax.SAXParseException; systemId: file:/home/okilo@columbia.tresys.com/git/owlsvcs/data/tools/daffodil/daffodil-sapi/target/scala-2.12/test-classes/test/sapi/mySchema6.dfdl.xsd; lineNumber: 32; columnNumber: 74; src-resolve: Cannot resolve the name 'tns:nonExistent' to a(n) 'type definition' component.\nSchema context: file:/home/okilo@columbia.tresys.com/git/owlsvcs/data/tools/daffodil/daffodil-sapi/target/scala-2.12/test-classes/test/sapi/mySchema6.dfdl.xsd Location in .../test-classes/test/sapi/mySchema6.dfdl.xsd
 ```

Comment (stevedlawrence) [reply daffodil-lib/src/main/scala/org/apache/daffodil/lib/api/Diagnostic.scala:317]: Seems reasonable to me.

I forget why we added `IllegalStateException`, but maybe we don't even need it anymore? Now that `UsageException` accepts a cause maybe this is just:

```
 def checkNotError(): Unit = {
    Assert.usage(!isError,
      "Must call isError() to ensure there are no errors",
      getDiagnostics.find(_.isError).get,
    )
  }
```

I don't feel strongly about it. Either works for me if having `IllegalStateException` provides extra useful information.

mbeckerle: I think illegal state adds the information that the usage is not that something is wrong with an argument, it's that you called it in the wrong state. Differentiates from IllegalArgumentException.

PR #1060, opened 2023-07-24
Update scala-xml to 2.2.0

stevedlawrence:
Update code to work with changes to the scala-xml library. This also includes removing a workaround for an infinite loop bug in reportSyntaxError.

Note that scala-steward deleted the branch associated with PR https://github.com/apache/daffodil/pull/1049 when the PR was closed, so I can't reop/update that PR. The scala-xml devs suggested an approach to resolve the issues mention in that, which is implemented in this new PR.

### DIFF
--- a/daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestXMLCatalogAndValidate.scala
+++ b/daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestXMLCatalogAndValidate.scala
@@ -24,7 +24,6 @@ import scala.collection.mutable
 import scala.language.reflectiveCalls
 import scala.xml.Attribute
 import scala.xml.Elem
-import scala.xml.InputSource
 import scala.xml.MetaData
 import scala.xml.NamespaceBinding
 import scala.xml.Node
@@ -45,6 +44,13 @@ import org.apache.xml.resolver.CatalogManager
 import org.junit.Assert.assertTrue
 import org.junit.Assert.fail
 import org.junit.Test
+import org.xml.sax.ContentHandler
+import org.xml.sax.DTDHandler
+import org.xml.sax.EntityResolver
+import org.xml.sax.ErrorHandler
+import org.xml.sax.InputSource
+import org.xml.sax.XMLReader
+import org.xml.sax.helpers.XMLFilterImpl
 
 object Implicits {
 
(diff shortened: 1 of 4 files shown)


daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestXMLCatalogAndValidate.scala
```diff
@@ -446,48 +452,56 @@ class SchemaAwareFactoryAdapter() extends NoBindingFactoryAdapter {
...
+  // We must use XMLReader setProperty() function to set the entity resolver--calling
+  // setEntityResolver with the Xerces XML reader causes validation to fail for some
+  // reason. We call the right function below, but unfortunately, scala-xml calls
+  // setEntityResolver in loadDocument(), which cannot be disabled and scala-xml does not
+  // want to change. To avoid this, we wrap the Xerces XMLReader in an XMLFilterImpl and
+  // override setEntityResolver to a no-op. However, XMLFilterImpl parse() calls
+  // setEntityResolver() on the XMLReader, which for the same reason as before causes
+  // issues. To fix this, we can override parse() to just pass through to the parent, but
+  // that means we must override the various set/get handler functions to also pass
+  // through to the parent.
```
mbeckerle, line 466: This is really jumping through hoops. 

I'm glad we have something that works, but I'd say scala-xml is a mess if it requires this.  The legacy of this XSD decision not to specify what schemaLocation really does, ... is causing quite a bit of havoc. 

The real comment here is this whole system is mostly duplicate of our loader, so we're not testing that loader here. 

I'm all for tests that characterize how a library works, particularly if the way it works is flakey and may evolve. But in this case we have almost the same piece of code in our library and here. 

Shouldn't we be calling our own library here to be sure it does this right? 

Or are there sufficient test elsewhere of that, and the purpose of this really is to illustrate usage for this foolish resolver crud in scala-xml?

mbeckerle (approved):
+1 Just a question about the code duplication between test code and our library.

daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestXMLCatalogAndValidate.scala
stevedlawrence: > I'm glad we have something that works, but I'd say scala-xml is a mess if it requires this. 

I'd agree, however, it also wouldn't surprising me if we're just doing something really wrong in how we're setting features/properties on the SAX Parser/XMLReader. The fact that we set a property with name "**internal**/entity-resolver" in it, makes me feel like we're missing something that would allow `setEntitytResolver` to work. I did play around with it a bit, but couldn't figure it out.

> Or are there sufficient test elsewhere of that, and the purpose of this really is to illustrate usage for this foolish resolver crud in scala-xml?

I think this is the case. We do have a few tests (e.g. [TestXMLLoader.scala](https://github.com/apache/daffodil/blob/main/daffodil-lib/src/test/scala/org/apache/daffodil/lib/xml/test/unit/TestXMLLoader.scala)  that are unit tests specifically around the use of the DaffodilXMLLoader. I think this file does act as a pretty good example of how to hook together scala-xml and Xerces + catalog manager/validation stuff and should probably stick around.

tuxji: I'm fine with keeping this test code around as well.  I also wonder if it's worth writing a test of the way you want to call the scala xml loader code (which doesn't currently work) so that if scala xml does change to support that way, the test would detect the change in behavior and alert us?  Maybe not worth the time, but just a thought.

tuxji (approved):
+1

daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestXMLCatalogAndValidate.scala
stevedlawrence: I think the way we called it was probably a bit of a hack, so I'm not sure we really want to go back to that. We essentially skipped the intended load xml entrypoint (`loadDocument()`) which initializes a number of important things, and probably isn't something really we want to undo. I think we just got lucky before and got by without it, but they are now required.

Probably what we really need to do is figure out how we can use `setEntityResolver` instead of `setProperty` with the Xerces `XMLReader`. I image it must be possible, but I think we're just doing something wrong that's messing up its internals. I didn't spend a ton of time on it, but I couldn't figure out what we are doing wrong. If we could figure that out this `XMLFilterImpl` stuff can just go away and we can use scala xml as the library intended.

PR #1064, opened 2023-08-03
Allow changing blob attributes when full/custom validation is enabled

stevedlawrence:
When full or custom validation is enabled, Daffodil creates a new TeeInfosetOutputter which proxies infoset events to the user provided InfosetOutputter and another one used for validation. This means that calls to getBlobDirectory, getBlobSuffix, setBlobPath, etc. go to the TeeInfosetOutputter, and ignore any values that a user may have set on their provided InfosetOutputter. This essentially means you cannot change the directory, suffix, or prefix of blobs if full or custom validation is enabled. It also means that if calling getBlobPaths on the user provided InfosetOutputter never returns anything, even if blobs were created.

To fix this, if full or custom validation is enabled, we copy the blob attributes to the TeeInfosetOutputter used for parsing. We also ensure that at the end of parsing, we call setBlobPaths on the InfosetOutputter the user passed in instead of the TeeInfosetOutputter.

DAFFODIL-2837

### DIFF
--- a/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
+++ b/daffodil-japi/src/test/java/org/apache/daffodil/example/TestJavaAPI.java
@@ -35,7 +35,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


tuxji (approved):
+1, looks good to me

mbeckerle (approved):
+1

PR #1068, opened 2023-08-14
Fix DPath conversion from double to unsigned long

stevedlawrence:
A conversion from a source type of xs:double to a target type of xs:unsignedLong correctly inserts a DoubleToUnsignedLong converter, but follows that with an additional LongToUnsignedLong converter. This leads to an exception because LongToUnsignedLong expects to get a Long, but the DoubleToUnsignedLong converter has converted the value to a BigInt (the representation for an UnsignedLong).

The additional converter is not needed and is likely a copy/paste error. This removes the unneeded converter and allows the conversion to succeed.

DAFFODIL-2839


mbeckerle (approved):
+1

tuxji (approved):
+1

Agree, seems best to just remove the unneeded conversion.

PR #1070, opened 2023-08-18
Remove dfdlx:inputTypeCalc and dfdlx:outputTypeCalc

stevedlawrence:
This removes the dfdlx:inputTypeCalc and dfdlx:outputTypeCalc properties and DPath functions. This allows for removal of a number of related member variables and functions used only by the expression type calc functions. For example, the typeCalcMap is only used by the functions to lookup a calculator by QName at runtime. All type calculators are known statically at compile time, so everything related to runtime type calculator maps goes away.

Also does some minor refactoring:
- Moves code around so similar logic is all in one place or removes conditionals
- Changes a tuple where only element has a meaningful value to an Either so we enforce that restriction
- Fixes typo in IdentityTypeCalc name
- Changes some error messages so they are a bit more less developer centric

Deprecation/Compatibility:

The dfdlx:inputTypeCalc and dfdlx:outputTypeCalc properties and
functions have been removed. Instead, one should use dfdlx:repType with
dfdlx:repValues and dfdlx:repValueRanges.

DAFFODIL-2211

----
Note that this is only part of of DAFFODIL-2211. I'm making this a separate PR so it's more clear which parts are about removing input/outputValueCalc and which parts are about updating the diagnostics. It also means if we ever want to bring back typeValueCalc, we can just revert this PR without undoing the diagnostic changes


stevedlawrence:
To do before merged, this needs a `Deprecation/Compatibility` section in the commit message to go in the release notes.

tuxji (approved):
+1

Would benefit from a couple of clarifying comments

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SimpleTypes.scala
```diff
@@ -312,40 +308,49 @@ abstract class SimpleTypeDefBase(xml: Node, lexicalParent: SchemaComponent)
     }
   }
 
-  lazy val optInputTypeCalc = findPropertyOption("inputTypeCalc", expressionAllowed = true)
-  lazy val optOutputTypeCalc = findPropertyOption("outputTypeCalc", expressionAllowed = true)
-
   lazy val optTypeCalculator: Option[TypeCalculator] = LV('optTypeCalculator) {
```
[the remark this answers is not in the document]
stevedlawrence: Will do, I had to learn about type calculators too so this would definitely be useful. Quickly, type calculators are used to convert a parsed value (based on the repType) to a value that actually ends up in the infoset.  There are different type calculators for supporting conversions based on enumerations and unions, and previously expressions, which is now removed.

So a parser actually parses the data (based on the repType), and the type calculator converts that parsed value to something else (based on the enumeration/union), usually with a different type, and that ends up in the infoset.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -183,9 +146,9 @@ class KeysetValueTypeCalculatorOrdered(
   override def inputTypeCalc(
     x: DataValuePrimitive,
     xType: NodeInfo.Kind,
-  ): (DataValuePrimitiveNullable, Maybe[Error]) = {
+  ): Either[Error, DataValuePrimitiveNullable] = {
```
mbeckerle, line 149: Ah, so you also changed this return type from tuple to Either. That's cleaner also.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -223,159 +186,34 @@ class KeysetValueTypeCalculatorUnordered(
...
-   */
-  lazy val maybeInputTypeCalc = maybeInputTypeCalcDelay.value
-  lazy val maybeOutputTypeCalc = maybeOutputTypeCalcDelay.value
-
-  // The class TypeValueCalc will verify that supports(Un)Parse is true when nessasary
-  // Therefore, if we ever call the below functions, we know that the relevent Maybe object is defined.
-
-  // TODO, pass x into DPath state
-
+class IdentityTypeCalculator(srcType: NodeInfo.Kind) extends TypeCalculator(srcType, srcType) {
```
mbeckerle, line 208: Curious as to why we still need this. For unit tests?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -223,159 +186,34 @@ class KeysetValueTypeCalculatorUnordered(
...
-    xType: NodeInfo.Kind,
-  ): (DataValuePrimitiveNullable, Maybe[Error]) = (x, Maybe.Nope)
-  override def outputTypeCalc(
-    x: DataValuePrimitive,
-    xType: NodeInfo.Kind,
-  ): (DataValuePrimitiveNullable, Maybe[Error]) = (x, Maybe.Nope)
+  ): Either[Error, DataValuePrimitiveNullable] = Right(x)
 }
 
 class UnionTypeCalculator(
```
mbeckerle, line 219: I suspect this might be able to go away.  I'm not understanding any feature of unions. I mean do we allow simple type unions as a repType or can one combine enums with a rep type using unions? 

This change set is complex enough, so if you want to put off further investigation of this I'm good with that.

And I also could be entirely mistaken and this stuff may need to be here.

mbeckerle (approved):
+1 This is a good set of changes. Other than the added doc, I just added comments that perhaps there are a few *more* things that can be removed also, but likely that should be looked at after this is merged.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -21,135 +21,98 @@ import scala.collection.immutable.HashMap
...
 import org.apache.daffodil.runtime1.processors.parsers.ParseError
 import org.apache.daffodil.runtime1.processors.unparsers.UState
 
+/**
+ * A TypeCalculator is a class that performs some kind of conversion from one DataValuePrimitive
+ * to another DataValuePrimitive. In general, this is done to convert a parsed physical
+ * representation from the data to a logical representation that goes in the infoset, and
+ * reverse on unparse. The former is done in the inputTypeCalc functions, and the latter in the
+ * outputTypeCalc functions. Generally the conversion is to a different type (e.g. xs:int to
+ * xs:string), but this need not be the case.
```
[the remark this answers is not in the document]
stevedlawrence: I think we support all int-like types to string. I think only decimal, float, double, calendar, etc. are not supported. We do also support string to string using the identity type calculator, which was untested but I did add a test for. I'm not sure what the use case for this was, but the logic was there.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -21,135 +21,98 @@ import scala.collection.immutable.HashMap
...
+ *
+ * The different TypeCalculator implementations allow for different conversions based on
+ * different properties provided in the schema (e.g. repType, enumerations, unions, repValues,
```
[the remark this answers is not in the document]
stevedlawrence: You can have two different simple types with enumerations/repValues defined that are unioned by another simple type. Here's a test we have of this:

https://github.com/apache/daffodil/blob/main/daffodil-test/src/test/resources/org/apache/daffodil/extensions/type_calc/inputTypeCalc.tdml#L66-L80

I'm not sure if we actually use unions with repValues in any schemas, but I do see use a xs:unions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
stevedlawrence: It was previously used when the repType and simpleType bases were the same, for example both are xs:string. I'm not actually sure if int to int works, but I can add a test for it.

I'm also not sure the use case for it, but I was thinking maybe it might be useful to use the same rep type for multiple logical types. E.g.

```xml
<simpleType name="rep" dfdl:properties="...">
  <restriction base="xs:int" />
</simpleType>

<simplType name="intToString" dfdl:repType="rep">
  <restriction base="xs:string">
    <enumeration value="1" dfdl:repValues="1" />
    ...
  </restriction>
</simpleType>

<simplType name="intToInt" dfdl:repType="rep">
  <restriction base="xs:int">
    <enumeration value="1" dfdl:repValues="1" />
    ...
  </restriction>
</simpleType>
```
I'm not sure if this is actually useful, or even if it works, but I assumed there was a reason for the identity transform.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
stevedlawrence: > Do we allow simple type unions as a repType or can one combine enums with a rep type using unions?

We allow a simple type to be a union of other simple types which can have repTypes defined. We only have one tdml test for it (and a couple of unit tests) so I'm not sure how heavily it's tested or used, but it seems to work.

Daffodil also supports unions in general (see union.tdml), so maybe it was added for that reason?

If we think this repType unions is unnecessary I can remove it to, I think it would be pretty straightforward and I'd rather do it now while it's all fresh in my memory. But I can do it in a separate PR if that's preferred.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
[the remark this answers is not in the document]
mbeckerle: I'd prefer to not support anything other than int->string. But we can make further changes later to restrict that. 

I do not think string transformation by lookup tables (which is what string-string really kind of is) would ever become part of the DFDL standard because it's too far down the slippery slope of transformation.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
mbeckerle: I'd rather take these union things out.  A repType should just be an integer, and typically isn't going to have facets, just DFDL properties. It's not visible at all in the infoset so there is no notion of an invalid repType value. 

If a simple type that uses enumerations and repValue(s) is a union... that feels very problematic to me and I'd like to eliminate that possibility.

mbeckerle: That said, are we sure we're not introducing yet another backward incompatibility.. We need to look at the complex schemas that were using inputTypeCalc et al, and see if they're using these union features.

stevedlawrence: I've grepped through our regression suite, and of the handful of schemas that use `xs:union`s, none of those unions or the simple types they reference use `repType`. We do use unions, but it looks like they are used to combine various restrictions.

So I don't think we would break anything by removing repType union support. I'll remove union repType support as a separate PR and test the regression suite first to make sure there are no breakages.

PR #1071, opened 2023-08-24
Remove support for dfdlx:repType/Values from unions

stevedlawrence:
dfdlx:repTypex on simpleTypes references from unions are now ignored. Likewise, dfdlx:repValues and dfdlx:repValueRanges are also ignored. Unions are now only used for combining restrictions from their simple types.

Deprecation/Compatibility:

dfdlx:repType, dfdlx:repValues, and dfdlx:repValueRanges are now ignored if references on xs:simpleTypes that are referenced from xs:unions. Instead, unions should be combined into a single simple type with the combined representation type and values properties.

DAFFODIL-2211

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SimpleTypes.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SimpleTypes.scala
@@ -317,7 +317,7 @@ abstract class SimpleTypeDefBase(xml: Node, lexicalParent: SchemaComponent)
    * multiple elements and input/outputValueCalc, but using repType allows it to be done by
    * Daffodil. The task to convert a physical type to a logical type during parse (or in reverse
    * during unparse) is done via a TypeCalculator. The goal of this optTypeCalculator is to
-   * inspect the various properties (e.g. repType, enumeration, unions, repValues,
+   * inspect the various properties (e.g. repType, enumeration, repValues,
    * repValueRanges) and create a TypeCalculator that does the necessary conversions. This type
    * calculator will later be provided to a parser/unparser to be evaluated at runtime after the
    * physical representation is parsed or prior to being unparsed.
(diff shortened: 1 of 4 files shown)


stevedlawrence:
For the record, I've run this PR against the regression suite and had no failures due to this change. Some schemas failed due to the removal of input/outputTypeCalc in commit 17f3fa8a1206f8801ab96a262f2219fb3eb6c72f, but that is expected and those schemas will either be deprecated or fixed.

tuxji (approved):
+1

Looks fine, although I don't know what to look for to check that the PR deletes everything needing to be removed.  That would be my only concern.

stevedlawrence:
> although I don't know what to look for to check that the PR deletes everything needing to be removed.

Yeah, I mostly did this through code inspection so it's hard to know for sure.

I did find codcov a bit useful since it will show code that has lost coverage even if not changed in the PR. In this case we only had two tests for unions, so I'm less confident we had good coverage in the first place, but codecov did find one small function I had missed.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SimpleTypes.scala
```diff
@@ -405,90 +386,39 @@ abstract class SimpleTypeDefBase(xml: Node, lexicalParent: SchemaComponent)
...
-    /*
-     * Note that there is no fromRestriction option here
-     * In theory, we could consider every restriction type without an explicit repType to be
-     * either a restriction or identity transform.
-     * In practice, this would introduce the overhead of a transform to almost every derived type.
-     * Instead, when a user needs a restriction transform, they must simply provide the reptype explitly,
-     * which is arguably a good design decision from a readability standpoint of the schema as well.
-     */
-    optRepTypeFromSelf.orElse(optRepTypeFromUnion)
+    PrimitiveType
```
stevedlawrence, line 390: This `PrimitiveType` line shouldn't be here. I'm not sure where it came from, maybe a weird copy/paste issue? But it doesn't do anything and should be removed.

mbeckerle (approved):
+1

PR #1075, opened 2023-08-29
Require that dfdlx:repType references a global simple type definition

stevedlawrence:
Primitive types do not carry any DFDL properties and so do not make for very good repTypes. To support primitive types as a repType, all DFDL proprties would need to come from the general format, which would only make sense if a few specific cases.

Instead, we simply disallow primitive repTypes, instead requiring that repType be a QName to a simple global type definition, which DFDL properties can be put.

This creates a new function to create a helpful diagnostic when it is detected that a repType references a primitive type. Additionally, dfdl:prefixLengthType already has this restriction, and is modified to use this new helper function.

Deprecation/Compatibility:

dfdlx:repType can no longer reference primitive types (e.g. xs:int). Instead, a global simple type definition should be created with that has a base of the primitive type.

DAFFODIL-2211

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
@@ -425,6 +425,21 @@ final class SchemaSet private (
     _.getGlobalSimpleTypeDef(refQName.local)
   }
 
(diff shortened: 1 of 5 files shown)


mbeckerle (approved):
+1

tuxji (approved):
+1

PR #1083, opened 2023-09-11
Ensure UTF-8 is used during CLI tests

stevedlawrence:
Daffodil should ignore the Java file.encoding property, only using the encoding specified in a DFDL schema. This goes for all of our tests as well. The one place where this isn't the case is with CLI tests, which can lead to some tests failing or hanging if run on a system with a different encoding.

To fix this, this modifies the CLI test infrastructure to run CLI tests with a UTF-8 encoding, regardless of the file.encoding property provided to Java, since parts of the CLI, specifically the debugger, use special UTF-8 characters and have tests that expect them.

This also modifies the Windows/Java 8 GitHub action to set the Java file.encoding property to US-ASCII. This ensures that if we ever do add file.encoding specific behavior to Daffodil then hopefully a test will fail. Note that newer versions of SBT do not get the file.encoding property from LANG so this also modifies the GitHub actions so that we can specify the encoding and LANG separately.

This also found tests for stringAsXml that relied on file.encoding instead of specifying an encoding. Fortunately, this was just broken tests and not a broken stringAsXml implementation.

DAFFODIL-2612

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,15 +41,15 @@ jobs:
         java_distribution: [ temurin ]
         java_version: [ 8, 11, 17 ]
         scala_version: [ 2.12.18 ]
-        os: [ ubuntu-22.04, windows-2022 ]
-        include:
+        os: [ ubuntu-22.04, windows-2022, macos-12 ]
+        exclude:
+          # only run macos on java 17
           - os: macos-12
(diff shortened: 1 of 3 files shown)


tuxji (approved):
+1

Looks correct.

mbeckerle (approved):
+1

PR #1085, opened 2023-09-13
Improve and narrow dfdlx:repType extension implementaion

stevedlawrence:
- The dfdlx:repType property is now resolved just like any other DFDL property, it no longer *must* be on a global simple type definition.
- The dfdlx:repType property now must reference a type with an underlying primitive type derived from xs:integer, and elements that define dfdlx:repType must have an underlying primitive type of xs:string
- Enumerations no longer need to be immediately on the element or simpleType associate with a repType. Multiple levels of indirection via restriction bases are now supported.
- Most of the complexity related to type calc is removed, now that it does not need to be generalized for other uses of type calculators. In fact, the concept of a "type calculator" and rep value sets are removed. The needed data structures for quick conversion from rep type to logical type are created and passed directly to the RepType parser/unparsers. We lose some extensibility, but that only adds unneeded complexity.
- repValueRanges are no longer converted to BigInts. Instead, where necessary DPath functions are used to compare values.
- repType logic is moved out of expression related files into their own specific files, since repType no longer has any relation to expressions now that type calculators are removed.
- The repType mixin is moved from DSOM to grammar, since the properties are more related to grammar usage than DSOM usage. This simplifies the implementation a bit
- EnumerationDef now does not convert repValues to their repType. Previously it could do this because it just use a BigInt, but now that the correct data primitive is used, we must do it elsewhere, sine the EnumerationDef does not know what the repType primitive is. Only the element where the repType is defined knows that information.
- Adds new logic to make sure that repValues and repValueRanges do not overlap

Deprecation/Compatibility:

The dfdlx:repType property now must reference a type with an underlying primitive type derived from xs:integer, and elements that define dfdlx:repType must have an underlying primitive type of xs:string

DAFFODIL-2738, DAFFODIL-2809


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -985,6 +985,7 @@ trait ElementBase
       "Facets minLength and maxLength are allowed only on types string and hexBinary.",
     )
     typeDef match {
+      case _ if hasRepType => (repTypeElementDecl.minLength, repTypeElementDecl.maxLength)
```
mbeckerle, line 988: With this change, it's clearly computing the min/max length of the representation. 

Add that to the scaladoc.

But is this correct? Is this only used when we need the max/min length of the representation? 
What about limited validation? Then we're dealing with the infoset value, and the max/min length are those of the logical data, not the rep.

Perhaps we need to distinguish infosetMinLength and repMinLength (similarly for max) ?

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/RepTypeMixin.scala
```diff
@@ -0,0 +1,213 @@
...
+      .toMap
+    val parseRangesMap: Seq[(DataValueNumber, DataValueNumber, DataValueString)] =
+      enumAttributes
+        .filterNot(_._2.isEmpty)
+        .flatMap { case (_, repRanges, enumValue) =>
+          repRanges.map { case (low, high) => (low, high, enumValue) }
+        }
+    val unparseMap: Map[DataValueString, DataValueNumber] = enumAttributes.map {
+      case (repValues, repRanges, enumValue) =>
+        val canonicalRep = None
```
mbeckerle, line 200: None.orElse(x.headOption).orElse(...)

I think can be simplified to

x.headOption.orElse(...)

daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/Cookers.scala
```diff
@@ -101,8 +101,6 @@ object ChoiceDispatchKeyCooker extends StringLiteralNonEmptyNoCharClassEntitiesN
 object ChoiceBranchKeyCooker
   extends ListOfStringLiteralNonEmptyNoCharClassEntitiesNoByteEntities()
 
-object RepValueCooker extends ListOfStringLiteralNonEmptyNoCharClassEntitiesNoByteEntities()
```
mbeckerle, line 104: If this is deleted, is that because repValues and range end points can only be integers?

mbeckerle: Follow up. I've never seen it, but are our repValues integers nonNegative only, or can things like -1 -2, etc. be enum repValues?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType.tdml
```diff
@@ -0,0 +1,708 @@
...
+    <xs:simpleType name="uint8" dfdl:lengthKind="explicit" dfdl:length="1">
+      <xs:restriction base="xs:unsignedInt"/>
+    </xs:simpleType>
+
+    <xs:simpleType name="keysetValue_01" dfdlx:repType="tns:uint8">
+      <xs:restriction base="xs:string">
+        <xs:enumeration value="zero" dfdlx:repValues="0" />
+        <xs:enumeration value="one" dfdlx:repValues="1" />
+        <xs:enumeration value="2-100" dfdlx:repValueRanges="2 100" />
+        <xs:enumeration value="101 103-110 115 120-125" dfdlx:repValues="101 115" dfdlx:repValueRanges="103 110 120 125"/>
```
mbeckerle, line 62: Nice test case. I would not have thought to put both repValues and ranges on one enum.

mbeckerle (approved):
+1 

This is a big improvement and simplification. There's one code-cov failure I saw where it would I think be of value to add a test to cover.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
```diff
@@ -17,288 +17,9 @@
...
-    }
-  }
-
-  def compileIdentity(srcType: NodeInfo.Kind): TypeCalculator = new IdentityTypeCalculator(
-    srcType,
-  )
-
-}
 
 object Range {
```
tuxji, line 24: Should we rename TypeCalculator.scala to Range.scala or RangeBound.scala since it defines only those types now?

tuxji (approved):
+1 after you answer Mike's comments

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
stevedlawrence: This change is basically the same as before, the case just changes a little bit (see the line removed below), so functionally nothing was really changed. And it looks like things to work as expected.

If I add a min/maxLength restriction to  the logical type, then limited validation does correctly check the length of the calculated enumeration string. Note that it seems unlikely that schemas will use both min/maxLength and enums on a restriction, but if they to it seems to work. Maybe this min/maxLength caclulation is not what is used for facet checking?

Also, if I add min/maxLength to the repType simple type, then it erros. min/maxLength facets are only valid on xs:string types, and repType must be a string type.

I'll add some tests to make sure this works as expected and isn't broken in the future.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/Cookers.scala
stevedlawrence: I think I can actually bring back the cooker. I thought this cooked everything to a BigInt and so avoided it (one of the updates in the PR is to use the correct DataPrimitive type). But I now see it just cooks to a list of strings. This is better than the `.split` stuff I do elsewhere. I'll give it a shot.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/TypeCalculator.scala
stevedlawrence: Makes sense. Will do.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/Cookers.scala
stevedlawrence: > Follow up. I've never seen it, but are our repValues integers nonNegative only, or can things like -1 -2, etc. be enum repValues?

The code currently allows anything derived from xs:integer, so signed types should work as well. I'll add a test or two to make sure it works as expected.

stevedlawrence:
> There's one code-cov failure I saw where it would I think be of value to add a test to cover.

I think the only way to trigger this code is if an element with dfdlx:repType is defined is in a hidden group while unparsing. However, this would mean that element also needs a dfdl:outputValueCalc. For example:

```xml
    <xs:simpleType name="rep" dfdl:lengthKind="explicit" dfdl:length="1">
      <xs:restriction base="xs:byte"/>
    </xs:simpleType>

    <xs:group name="hiddenField">
      <xs:sequence>
        <xs:element name="field" dfdlx:repType="tns:rep" dfdl:outputValueCalc="{ ../ex:visField }">
          <xs:simpleType>
              <xs:enumeration value="one" dfdlx:repValues="1" />
              <xs:enumeration value="two" dfdlx:repValues="2" />
              <xs:enumeration value="three" dfdlx:repValues="3" />
            </xs:restriction>
          </xs:simpleType>
        </xs:element>
      </xs:sequence>
    </xs:group>

    <xs:element name="root1">
      <xs:complexType>
        <xs:sequence>
          <xs:group ref="ex:hiddenField" />
          <xs:element name="visField" type="xs:string" dfdl:inputValueCalc="{ ../ex:field }" />
        </xs:sequence>
      </xs:complexType>
    </xs:element>
```

But so far I haven't been able to get unparsing to work with this. There seems to be some weird suspension issue where the outputValueCalc expression can't find the visField element. I'm guessing it has something to do with the detached quasi element messing with the infoset, but haven't dug much further so far.

That said, I wonder if it's even worth trying to support dfdlx:repType + dfdl:ouputValueCalc? A repType inside hidden groups seems unlikely since the whole point of repTypes it to convert an int to a more human friendly string. If it's hidden is it worth doing that conversion at all? And for non-hidden fields, it feels like it would be unlikely to want to calculate the string value on unparse via OVC.

Maybe it should be an SDE if the two properties are defined together?

mbeckerle:
Agree. Add the SDE check.

If you are using input/output value calc, you already have a decoupling of the rep from the logical, mixing that with the enum/repType system is unwise. 

Hidden groups don't compose orthogonally with everything else in DFDL.

PR #1086, opened 2023-09-15
Replace Scala Maps with Java Maps to in choice parsers to avoid bad diagnostics

stevedlawrence:
When Scala serializes a Map, it uses writeReplace() to actually serialize a HashMap$SerializationProxy object. There appears to be a bug in the deserialization of this proxy object which can lead to very confusing diagnostics if classpaths are not set up correctly when reloading a saved parser.

For example, if a layer transformer is not on the classpath when reloading, you could get an error mentioning only this HashMap$SerializationProxy and the ChoiceCombinator parser. It is not clear at all that the core issue is a layer transformer is not on the classpath.

To fix this, this changes the choice combinator parsers and unparsers to use Java Maps instead. This do not have the deserialization issues that Scala Maps have. Now, if a layer transformer is not on the classpath you get a much more helpful diagnostic stating the saved parser was "created with a different set of dependencies containing a class no longer on the classpath", and includes the actual transformer factory class name, making it much more clear what the problem is.

DAFFODIL-2728

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/ChoiceAndOtherVariousUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/ChoiceAndOtherVariousUnparsers.scala
@@ -17,23 +17,26 @@
 
 package org.apache.daffodil.unparsers.runtime1
 
+import scala.collection.JavaConverters._
+
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.Maybe._
 import org.apache.daffodil.lib.util.MaybeInt
+import org.apache.daffodil.lib.util.ProperlySerializableMap._
 import org.apache.daffodil.runtime1.infoset._
 import org.apache.daffodil.runtime1.processors.RuntimeData
 import org.apache.daffodil.runtime1.processors._
 import org.apache.daffodil.runtime1.processors.unparsers._
 
 case class ChoiceBranchMap(
-  lookupTable: Map[ChoiceBranchEvent, Unparser],
+  lookupTable: ProperlySerializableMap[ChoiceBranchEvent, Unparser],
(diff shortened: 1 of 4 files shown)


stevedlawrence:
Note that I was only able to reproduce this using a very large complicated schema. Even having a pretty good guess at what the underlying issue is (direct dispatch choices failing to deserialize because they refer to missing layers), I was still unable to reproduce the problem with a smaller schema. That said, even if I found one I'm not sure how easy it would be to create a regression test since it requires messing saving/reloading parsers with different classpaths and missing layers.

But I did confirm this change results it much better diagnostic for the large schema where it is reproducible.

tuxji (approved):
+1

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/ChoiceAndOtherVariousUnparsers.scala
```diff
@@ -26,14 +28,14 @@ import org.apache.daffodil.runtime1.processors._
 import org.apache.daffodil.runtime1.processors.unparsers._
 
 case class ChoiceBranchMap(
-  lookupTable: Map[ChoiceBranchEvent, Unparser],
+  lookupTable: java.util.Map[ChoiceBranchEvent, Unparser],
```
mbeckerle, line 31: Suggest centralizing all of this around this type definition
which you then put all the scaladoc/comment documentation on it. 
```
type ProperlySerializableMap = java.util.Map
```
Then use ProperlySerializableMap in all the places you currently java.util.Map explicitly.

mbeckerle (approved):
+1 But I suggest some revisions to centralize all the knowledge about this issue as much as possible.

stevedlawrence:
All comments resolved. I also added an implicit class that decorates `Map`s with a `toProperlySerializableMap` function which handles the logic to correctly convert a Scala Map to a Java Map.

PR #1090, opened 2023-09-23
Changes for Java 21 (LTS)

mbeckerle:
Ready for review

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -211,6 +211,12 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
   .dependsOn(tdmlProc % "test")
   .settings(commonSettings, nopublish)
 
+// Choices here are Java LTS versions, 8, 11, 17, 21,...
+// However 8 is deprecated as of Java 21, so will be phased out.
+val minSupportedJavaVersion: String =
+  if (scala.util.Properties.isJavaAtLeast("21")) "11"
+  else "8"
+
 lazy val commonSettings = Seq(
   organization := "org.apache.daffodil",
   version := "3.6.0-SNAPSHOT",
@@ -245,7 +251,7 @@ lazy val commonSettings = Seq(
 
 def buildScalacOptions(scalaVersion: String) = {
   val commonOptions = Seq(
-    "-target:jvm-1.8",
+    s"-release:$minSupportedJavaVersion", // scala 2.12 can only do Java 8, regardless of this setting.
     "-feature",
     "-deprecation",
     "-language:experimental.macros",
@@ -268,13 +274,19 @@ def buildScalacOptions(scalaVersion: String) = {
     case _ => Seq.empty
   }
 
-  val javaVersionSpecificOptions =
-    if (scala.util.Properties.isJavaAtLeast("9"))
-      Seq("-release", "8") // ensure Java backwards compatibility (DAFFODIL-2579)
-    else
-      Seq.empty
+  commonOptions ++ scalaVersionSpecificOptions
+}
 
-  commonOptions ++ scalaVersionSpecificOptions ++ javaVersionSpecificOptions
+val javaVersionSpecificOptions = {
+  val releaseOption = // as of Java 11, they no longer accept "-release". Must use "--release".
+    if (scala.util.Properties.isJavaAtLeast("11")) "--release" else "-release"
+
+  // Java 21 deprecates Java 8 and warns about it.
+  // So if you are using Java 21, Java code compilation will specify a newer Java version
+  // to avoid warnings.
+  if (scala.util.Properties.isJavaAtLeast("11")) Seq(releaseOption, minSupportedJavaVersion)
+  else if (scala.util.Properties.isJavaAtLeast("9")) Seq(releaseOption, "8")
+  else Nil // for Java 8 compilation
 }
 
 // Workaround issue that some options are valid for javac, not javadoc.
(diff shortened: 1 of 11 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
```diff
@@ -70,6 +70,34 @@ object XMLUtils {
...
+   */
+  def strToDouble(s: String): Double = {
+    s match {
+      case PositiveInfinityString => Double.PositiveInfinity
+      case NegativeInfinityString => Double.NegativeInfinity
+      case NaNString => Double.NaN
+      case _ => s.toDouble
+    }
+  }
+
```
stevedlawrence, line 100: The `NodeInfo.Float.fromXMLString` and `NodeInfo.Double.fromXMLString` functions do something similar with a little extra error checking.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
```diff
@@ -850,6 +886,14 @@ Differences were (path, expected, actual):
     arrayCounters
   }
 
+  private def getXSIType(a: Elem) = {
+    val res = a.attribute(XSI_NAMESPACE.toString, "type").map(_.head.text).orElse {
+      val a2 = a.attributes.asAttrMap.get("xsi:type")
```
stevedlawrence, line 891: This `a2` is new. Is this because `xmlns:xsi` isn't define in some infosets? Should we just define them so they are valid XML?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
```diff
@@ -1113,6 +1189,29 @@ Differences were (path, expected, actual):
         val b = DFDLDateTimeConversion.fromXMLString(dataB)
         a == b
       }
+      case Some("xs:double") => {
+        val a = strToDouble(dataA)
+        val b = strToDouble(dataB)
```
stevedlawrence, line 1194: Mentioned above, but this could be `NodeInfo.Double.fromXMLString(...)`. Same with xs:float below.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
```diff
@@ -103,38 +102,99 @@ object Misc {
     // more time is wasted by people forgetting that the initial "/" is needed
     // to get classpath relative behavior... Let's make sure there is a leading "/"
     val resPath = if (resourcePath.startsWith("/")) resourcePath else "/" + resourcePath
-    val res = this.getClass().getResource(resPath)
-    if (res == null) {
-      (None, resPath)
-    } else (Some(res.toURI), resPath)
+    val res = Option(this.getClass().getResource(resPath))
+    (res.map.(_.toURI), resPath)
```
stevedlawrence, line 106: Typo `.` here.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
mbeckerle: Actually NodeInfo is in daffodil-runtime1, and this is in daffodil-lib, so to avoid code duplication I can convert NodeInfo.Double.fromString to call this. That's the right thing because the NodeInfo things construct JDouble boxed things anyway.

build.sbt
```diff
@@ -268,28 +272,30 @@ def buildScalacOptions(scalaVersion: String) = {
     case _ => Seq.empty
   }
 
-  val javaVersionSpecificOptions =
-    if (scala.util.Properties.isJavaAtLeast("9"))
-      Seq("-release", "8") // ensure Java backwards compatibility (DAFFODIL-2579)
-    else
-      Seq.empty
+  commonOptions ++ scalaVersionSpecificOptions
```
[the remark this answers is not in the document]
mbeckerle: Added per comment. 

But... It's a bit trickier. Scala 2.12 can't specify a release other than 8. So we can't just default to specifying java 11 class files when we are using Java 21, because scala warns about that, and we're fatal warnings. 

So I did the minSupportedJavaVersion thing, but it's only used by java compilation for now. We have very small amounts of java code today, but those would complain about java 8, so those will use Java 11 classes IF the JVM is Java 21 only. Scala compilation does not specify any "-target" option.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
mbeckerle: Same resolution. NodeInfo.Double.fromString calls this.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
mbeckerle: Removed. Requires an xmlns:xsi definition now.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
mbeckerle: Fixed.

tuxji (approved):
I'm on vacation so I had to review on my phone, but looks good to me. 

Maybe a future PR can extend the type lookup to retrieve schema component type data (element "a" has the base type "xs:double") in addition to checking for an explicit xsi:type attribute in the XML document?  This change would be better than modifying both daffodil and daffodilC to write xsi:type=”xs:double" into XML infoset documents.

stevedlawrence (approved):
+1 :+1:

PR #1092, opened 2023-10-14
WIP: Metadata walker for JAPI

mbeckerle:
For preliminary review.

Enables systems that want to interface tightly with Daffodil via JAPI, to walk the runtime schema information so as to construct their own corresponding metadata.

Done to enable Apache Drill but should be useful generally. 

For use case, see PR in Apache Drill: https://github.com/apache/drill/pull/2836

DAFFODIL-2832

### DIFF
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
@@ -22,49 +22,36 @@ import java.net.URI
 import java.nio.channels.ReadableByteChannel
 import java.nio.channels.WritableByteChannel
 import scala.collection.JavaConverters._
-
-import org.apache.daffodil.core.compiler.{ Compiler => SCompiler }
-import org.apache.daffodil.core.compiler.{ InvalidParserException => SInvalidParserException }
-import org.apache.daffodil.core.compiler.{ ProcessorFactory => SProcessorFactory }
+import org.apache.daffodil.core.compiler.{Compiler => SCompiler}
+import org.apache.daffodil.core.compiler.{InvalidParserException => SInvalidParserException}
+import org.apache.daffodil.core.compiler.{ProcessorFactory => SProcessorFactory}
 import org.apache.daffodil.core.dsom.ExpressionCompilers
 import org.apache.daffodil.core.dsom.walker.RootView
 import org.apache.daffodil.japi.debugger._
 import org.apache.daffodil.japi.infoset._
 import org.apache.daffodil.japi.io.InputSourceDataInputStream
 import org.apache.daffodil.japi.packageprivate._
+import org.apache.daffodil.japi.schema.{RuntimeSchemaHandler, RuntimeSchemaWalker}
 import org.apache.daffodil.lib.api.URISchemaSource
 import org.apache.daffodil.lib.api.Validator
-import org.apache.daffodil.lib.api.{ DataLocation => SDataLocation }
-import org.apache.daffodil.lib.api.{ Diagnostic => SDiagnostic }
-import org.apache.daffodil.lib.api.{ LocationInSchemaFile => SLocationInSchemaFile }
-import org.apache.daffodil.lib.api.{ WithDiagnostics => SWithDiagnostics }
+import org.apache.daffodil.lib.api.{DataLocation => SDataLocation}
+import org.apache.daffodil.lib.api.{Diagnostic => SDiagnostic}
+import org.apache.daffodil.lib.api.{LocationInSchemaFile => SLocationInSchemaFile}
+import org.apache.daffodil.lib.api.{WithDiagnostics => SWithDiagnostics}
 import org.apache.daffodil.lib.xml.DFDLCatalogResolver
 import org.apache.daffodil.lib.xml.XMLUtils
-import org.apache.daffodil.runtime1.api.DFDL.{
-  DaffodilUnhandledSAXException => SDaffodilUnhandledSAXException,
-}
-import org.apache.daffodil.runtime1.api.DFDL.{
-  DaffodilUnparseContentHandler => SDaffodilUnparseContentHandler,
-}
-import org.apache.daffodil.runtime1.api.DFDL.{
-  DaffodilUnparseErrorSAXException => SDaffodilUnparseErrorSAXException,
-}
+import org.apache.daffodil.runtime1.api.DFDL.{DaffodilUnhandledSAXException => SDaffodilUnhandledSAXException}
+import org.apache.daffodil.runtime1.api.DFDL.{DaffodilUnparseContentHandler => SDaffodilUnparseContentHandler}
+import org.apache.daffodil.runtime1.api.DFDL.{DaffodilUnparseErrorSAXException => SDaffodilUnparseErrorSAXException}
 import org.apache.daffodil.runtime1.debugger.Debugger
-import org.apache.daffodil.runtime1.debugger.{ InteractiveDebugger => SInteractiveDebugger }
-import org.apache.daffodil.runtime1.debugger.{ TraceDebuggerRunner => STraceDebuggerRunner }
-import org.apache.daffodil.runtime1.processors.{
-  DaffodilParseXMLReader => SDaffodilParseXMLReader,
-}
-import org.apache.daffodil.runtime1.processors.{ DataProcessor => SDataProcessor }
-import org.apache.daffodil.runtime1.processors.{
-  ExternalVariableException => SExternalVariableException,
-}
-import org.apache.daffodil.runtime1.processors.{
-  InvalidUsageException => SInvalidUsageException,
-}
-import org.apache.daffodil.runtime1.processors.{ ParseResult => SParseResult }
-import org.apache.daffodil.runtime1.processors.{ UnparseResult => SUnparseResult }
-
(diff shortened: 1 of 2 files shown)


daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
+package org.apache.daffodil.japi.schema
```
mbeckerle, line 1: apache banner

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Called for simple type element declarations.
+   *
+   * @param qnameString
+   * @param primTypeName
+   * @param isArray only true if more than one occurrence is possible. Disjoint with isOptional
+   * @param isOptional true if 0 or 1 occurrence only. Allows for a different representation of optional from a 0..1 array
+   */
+  def elementSimple(
```
mbeckerle, line 28: Name: simpleElement ?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
...
+  ElementRuntimeData,
+  ErrorERD,
+  SequenceRuntimeData,
+  TermRuntimeData,
+}
+
+/**
+ * Base class used by japi clients who want to walk the runtime schema information.
+ */
+abstract class RuntimeSchemaHandler() {
```
mbeckerle, line 18: Name: make clear this is walking the schema, but not the DSOM schema, it's walking the schema as the Runtime1 sees it, which is the metadata needed to interface Runtime1 with other data frameworks. It's the metadata of the data that will be presented to that framework. 

"RuntimeSchema" doesn't seem sufficient.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
...
+    handler.endSequence()
+  }
+
+  private def walkChoice(handler: RuntimeSchemaHandler, crd: ChoiceRuntimeData): Unit = {
+    handler.startChoice()
+    crd.groupMembers.map { trd =>
+      walkTerm(handler, trd)
+    }
+    handler.endChoice()
+  }
```
mbeckerle, line 133: Needs unit tests or other tests.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.daffodil.japi.{ DataProcessor => JDataProcessor }
+import org.apache.daffodil.lib.exceptions.Assert
+import org.apache.daffodil.runtime1.dpath.NodeInfo.PrimType
+import org.apache.daffodil.runtime1.processors.{
+  ChoiceRuntimeData,
+  DataProcessor => SDataProcessor,
+  ElementRuntimeData,
+  ErrorERD,
+  SequenceRuntimeData,
+  TermRuntimeData,
```
tuxji, line 12: In standard Daffodil code style, each type is imported in a separate import statement (not collected and grouped like this).

tuxji (commented):
Since this PR is still a work in progress, will only make some comments.

I agree that this new RuntimeSchemaWalker.scala's API needs unit tests with examples of how to call the API properly.  I agree that some API names such as `elementSimple` could be better named `simpleElement` or `startSimpleElement` (without an end function).

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -401,6 +388,12 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
   extends WithDiagnostics(dp)
   with Serializable {
 
+  /**
+   * For use by the JAPI implementation only.
+   * @return the underlying data processor
+   */
+  private [japi] def getUnderlyingDataProcessor = dp
```
stevedlawrence, line 395: Does scala's package private thing apply to the generated java docs? If not, we might want to add a doc that says this shouldn't be used externally.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -514,6 +507,11 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
    */
   def save(output: WritableByteChannel): Unit = dp.save(output)
 
+  def walkRuntimeSchema(handler: RuntimeSchemaHandler): Unit = {
+    val walker = new RuntimeSchemaWalker(dp)
```
stevedlawrence, line 511: I think we already have a couple different walkers. We have an` AbsractDSOMWalker` that is used by the experimental NiFi Daffodil Record processor. We also have the walker used by the C Code generator. I think they walk different things. Does this also walk something different? Is there any chance these could all be combined? Seems we might be doing something wrong if we have a bunch of different ways to walk different things?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
stevedlawrence, line 1: I wonder if this wants to be in some experimental namespace? I consider japi and sapi as being very strict about backwards compatibility. It might be nice to put this in a namespace that makes it clear this maybe not be final and is subject to change--I don't want to tie us down to an API that maybe isn't completely fleshed out.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
stevedlawrence: We already have a DSOM walker used by NiFi, is that not sufficient? Can you add comments about how the two differ?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
```diff
@@ -0,0 +1,135 @@
...
+  // we wouldn't need this otherwise.
+  def this(jdp: JDataProcessor) = this(jdp.getUnderlyingDataProcessor)
+
+  private lazy val rootERD = dp.ssrd.elementRuntimeData
+
+  def walk(handler: RuntimeSchemaHandler): Unit = {
+    walkTerm(handler, rootERD)
+  }
+
+  private def walkTerm(handler: RuntimeSchemaHandler, trd: TermRuntimeData): Unit = {
```
stevedlawrence, line 74: I'm not sure where the logic should be, but one difference between this walker and the DSOM walker is the DSOM walker has a `isHidden` boolean to avoid walking into hidden group refs that won't end up in the infoset. Does this need something similar, or beause this is walking runtime1 constructs are hidden elements implicitly skipped over?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Right, so I think that decision, to use a DSOM walker for NiFi, was a mistake. 

When we are interfacing Daffodil's runtime-1 output to a different data fabric, it makes sense to use the metadata structures of runtime-1 rather than DSOM. Everything we need seems to be available on the RuntimeData objects. We don't need the complexity of DSOM.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
mbeckerle: How about japix which we document as japi-experimental?

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
mbeckerle: I will have to verify the RuntimeSchemaWalker (or whatever we end up calling it) does not walk into hidden groups. We don't need the hidden groups because the infoset outputter from Daffodil via JAPI already has them removed from it. So we never will encounter an actual element for conversion to Drill that is hidden.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/schema/RuntimeSchemaWalker.scala
mbeckerle: Be sure to make one test use a hidden group

mbeckerle:
Closing. Superceded by PR:

https://github.com/apache/daffodil/pull/1112

Relevant comments on this have been recorded there with links back to this PR for context.

PR #1104, opened 2023-10-27
Use the correct namespace for the quasi repType element

stevedlawrence:
When a repType property exists for an element, we create a new quasi schema element that has its type attribute set to the value of the repType attribute. This quasi element is then used to generate a repType parser from.

Currently, the namespace used for this quasi element is the in-scope namespace for the element that has a the repType property. But it is possible that that repType property is defind on a simple type in another file with a completely different namespace. This means this quasi element could use the wrong namspace, which can lead to failure to find the repType and confusing diagnostics.

To fix this, instead of using the in-scope namespace from the element using a repType, we instead use the in-scope namespace from where the repType was actually defined.

DAFFODIL-2857

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_a.dfdl.xsd
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_a.dfdl.xsd
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<schema
+  xmlns="http://www.w3.org/2001/XMLSchema"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:a="http://example.com"
+  targetNamespace="http://example.com"
+  elementFormDefault="unqualified">
+
+  <include schemaLocation="/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
+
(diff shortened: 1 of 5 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_b.dfdl.xsd
```diff
@@ -0,0 +1,47 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<schema
+  xmlns="http://www.w3.org/2001/XMLSchema"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:b="http://example.com"
```
mbeckerle, line 24: We don't want people to do this in real practice, but I think we want to use prefix "a" here so to show that the namespaces are being used, not the prefixes.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_b.dfdl.xsd
```diff
@@ -0,0 +1,47 @@
...
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format ref="b:GeneralFormat" representation="binary" />
+    </appinfo>
+  </annotation>
+
+  <simpleType name="repType" dfdl:lengthKind="explicit" dfdl:length="1">
+    <restriction base="xs:int" />
+  </simpleType>
+
+  <simpleType name="valueType" dfdlx:repType="b:repType">
```
mbeckerle, line 40: If this says "a:repType", but "a" is this file's namespace, then I'd be more convinced this works.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_a.dfdl.xsd
```diff
@@ -0,0 +1,39 @@
...
+
+  <include schemaLocation="repType_01_b.dfdl.xsd" />
+
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:format ref="a:GeneralFormat"/>
+    </appinfo>
+  </annotation>
+
+  <element name="value" type="a:valueType" />
```
mbeckerle, line 37: You need to define a simple type in this file named repType to prove that it does NOT get used.

mbeckerle (approved):
+1

Minor suggestions.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/repType/repType_01_b.dfdl.xsd
stevedlawrence: The issue is that this file uses a "b" prefix and the other file uses the "a" prefix. They must be different to trigger the bug.

mbeckerle (approved):
+1
There is an alternate way this could be illustrated, but having worked through it, I think your original testing is sufficient.

bsloane1650:
+1

PR #1105, opened 2023-10-27
Allow empty leading/trailing whitespace for list-of-strings properties

stevedlawrence:
Currently we SDE if a property that accepts of list of DFDL string literals starts or end with whitespace. This removes that restriction in the list cooker by splitting on whitespace and removing any strings that are the empty string.

This has one side effect that the resulting list can now be empty--previously an empty property would have a single element with the empty string. An empty list is easier to reason about and detect than a list with a single zero-length string, but it does require changing the logic in a few places.

DAFFODIL-2858

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/EntityReplacer.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/EntityReplacer.scala
@@ -722,20 +722,9 @@ sealed abstract class ListOfStringLiteralBase(
   private lazy val olc = oneLiteralCooker
 
   protected def cook(raw: String, context: ThrowsSDE, forUnparse: Boolean): List[String] = {
-    if (raw.length != 0 && (raw.head.isWhitespace || raw.last.isWhitespace)) {
-      val ws = if (raw.head.isWhitespace) raw.head else raw.last
-      val wsVisible = Misc.remapCodepointToVisibleGlyph(ws.toChar).toChar
-      val hexCodePoint = "%04x".format(ws.toInt)
-      context.SDE(
-        "The property '%s' cannot start or end with the string \"%s\"(Unicode hex code point U+%s), or consist entirely of whitespace."
-          + "\nDid you mean to use character entities like '%%SP;' or '%%NL;' to indicate whitespace in the data format instead?",
-        propName,
-        wsVisible,
-        hexCodePoint,
-      )
-    }
 
-    val rawList = raw.split("\\s+").toList
+    // ignore leading, trailing, and repeating whitespae
+    val rawList = raw.split("\\s").filterNot(_ == "").toList
 
     val cooked = {
       val cookedList: ListBuffer[String] = ListBuffer.empty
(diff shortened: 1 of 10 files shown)


daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/EntityReplacer.scala
```diff
@@ -774,18 +763,14 @@ class NonEmptyListOfStringLiteral(pn: String, allowByteEntities: Boolean)
 
   override def testCooked(cookedList: List[String], context: ThrowsSDE) = {
     context.schemaDefinitionUnless(
-      cookedList.exists { _.length > 0 },
+      cookedList.length > 0,
       "Property dfdl:%s cannot be empty string. Use dfdl:nilValue='%%ES;' for empty string as nil value.",
```
mbeckerle, line 767: This messsage mentions dfdl:nilValue but at least in theory it is not the only property that would use this cooker.

Just drop the property name. I think the message is ok without it. Or pass it in as a parameters somehow.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/EntityReplacer.scala
```diff
@@ -889,16 +874,16 @@ class NonEmptyListOfStringLiteralCharClass_ES_WithByteEntities(pn: String)
...
```
mbeckerle, line 878: Same issue. Hard coded nilValue in message.

mbeckerle (approved):
+1

I did find 2 places where a generic message mentions a specific property, so fix those. But other than those quick fixes this looks good

daffodil-lib/src/main/scala/org/apache/daffodil/lib/cookers/EntityReplacer.scala
stevedlawrence: Changed to use `%s` and `propName` so the error message should stay the same so not break existing tests.

Semi-related, in making these changes, I noticed a lot of duplication in our cookers (e.g. these multiple nil cookers). I'm thinking at some point we may want to refactor these. For example, we have some really complicated cookers that are like:

```scala
object ChoiceDispatchKeyCooker extends StringLiteralNonEmptyNoCharClassEntitiesNoByteEntities()
```
I imagine we can replace these just just a bunch of mixins, e.g.

```scala
object ChoiceDispatchKeyCooker extends StringLiteral
 with NonEmptyLiteral
 with NoCharClassEntities
 with NoByteEntities
```

So we just create a bunch of tiny specific traits that are easy to write and verify, and stack them all together. It makes the cooker definitions a bit longer, but makes it much easier to fix/change them without having to create new 50+ character classes when a new property needs something slightly different.

For another day though.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -2010,12 +2010,13 @@
...
-    <tdml:errors>
-      <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>property 'initiator'</tdml:error>
-      <tdml:error>cannot start or end with the string</tdml:error>
-      <tdml:error>U+0020</tdml:error>
-    </tdml:errors>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <expr_space5>
+          <value>21blastoff</value>
```
bsloane1650, line 2016: This output is correct, as the newlines in the initiator are whitespace, and therefore delimeters. However, just looking at the test, it looks wrong for the "21" to not be part of the prefix. Maybe have the initiator be
```
"{'
321
blast
' }"
```

That way the example is intuitive, and you still demonstrate that the newline is being interperated as a delimeter.

bsloane1650 (approved):
+1 
Minor suggestion to make a test case a bit more intuitive.

PR #1112, opened 2023-11-09
Refactoring to expose clean metadata and infoset walkers

mbeckerle:
(Supercedes PR https://github.com/apache/daffodil/pull/1092, which is still open because I linked the comments there from comments in this PR.)

Version 3.7.0-SNAPSHOT

This was designed to support integration of Daffodil directly (without intermediate XML or JSON or even EXI) to other data handling tools, specifically, Apache Drill.

InfosetElement and related InfosetNode traits are now in runtime1.api package.

InfosetOutputter now has methods which use InfosetElement and InfosetArray traits as the objects passed to the handler methods. This improves the API over making DIArray, DISimple, and DIComplex visible.

Added Metadata, ElementMetadata, etc. (also runtime1.api package) which provide limited exposure to the RuntimeData and CompilerInfo information.

Added MetadataHandler - which is walked by MetadataWalker which can be called from DataProcessor.

Walking the runtime1 metadata is easier than walking the DSOM tree. And these data fabrics like Apache Drill are interfacing to the runtime, not the compiler, so it's natural for the runtime1 metadata structures and data structures to be the ones driving the interfacing.

The InfosetNode types were always supposed to be the API, the DINodes the implementation. This solves the issue of what classes should show through to SAPI and JAPI about the infoset nodes. It should be the InfosetNode types, not the DINode types.

Furthermore, the InfosetNode types can have methods to access the needed runtime metadata information needed by walkers, InfosetOutputters, etc.

These hide our infoset implementation and runtime metdata (RuntimeData classes) implementions.

Added unit test for metadata and data walking to core module

Note: Nothing has changed with InfosetInputters, as those are not needed for Apache Drill integration - which is parse-only.

Refactored InfosetOutputter to enable sharing InfosetOutputterImpl.

BlobMethodImpl is now a class so must be extended first, InfosetOutputter is a trait so is implemented (second) in Java declarations.

Removed use of DIComplex, DISimple from SAPI and JAPI tests

NodeInfo.PrimType is n.g. for Java access.
Using PrimTypeNode which is a class.

TODO: maybe just use type name strings for this.

Added features to SchemaUtils to avoid "tns" prefix definition (which is now officially frowned upon)

Added isHidden to SequenceRuntimeData. Needed to avoid walking hidden elements that appear in metadata structures, but aren't relevant as they do not appear in InfosetOutputter events.

TODO: ensure these are not shared between hidden/non-hidden groups. (PropEnv stuff)
That is: isHidden on SequenceRuntimeData may be broken in the case where a sequence definition is shared both hidden and non-hidden.

Added child access to InfosetComplexElement. This is used by tests.

It is questionable whether this should be allowed, because we don't have fast O(1) hash-based access to all children, only to children referenced by expressions. This allows linear search for now.

Added QName.toPrefix_localName method. Why allocate these on the fly if they're static compile time constants.

DAFFODIL-2832


daffodil-core/src/test/scala/org/apache/daffodil/core/general/TestRuntimeProperties.scala
```diff
@@ -37,7 +37,8 @@ import org.junit.Test
  */
 class RedactingScalaXMLInfosetOutputter extends ScalaXMLInfosetOutputter {
 
-  override def startSimple(diSimple: DISimple): Unit = {
+  override def startSimple(se: InfosetSimpleElement): Unit = {
+    val diSimple = se.asInstanceOf[DISimple]
```
mbeckerle, line 41: This should be able to be done without need for DISimple.

daffodil-core/src/test/scala/org/apache/daffodil/core/util/TestUtils.scala
```diff
@@ -40,13 +40,13 @@ import org.apache.daffodil.lib.xml._
...
 import org.apache.daffodil.runtime1.infoset.InfosetInputter
-import org.apache.daffodil.runtime1.infoset.InfosetOutputter
 import org.apache.daffodil.runtime1.infoset.ScalaXMLInfosetInputter
 import org.apache.daffodil.runtime1.infoset.ScalaXMLInfosetOutputter
 import org.apache.daffodil.runtime1.processors.DataProcessor
 import org.apache.daffodil.runtime1.processors.VariableMap
 
 import org.apache.commons.io.output.NullOutputStream
+import org.apache.daffodil.runtime1.infoset.InfosetOutputter
 import org.junit.Assert.assertEquals
```
mbeckerle, line 51: Revert. Just a re-ordering of imports.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/Metadata.scala
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Called for complex type element metadata (for declarations or references)
+   *
+   * Subsequent calls will be for the model group making up the content
+   * of the element.
+   */
+  def startComplexElementMetadata(m: ComplexElementMetadata): Unit
+  def endComplexElementMetadata(m: ComplexElementMetadata): Unit
+
+  //  def startSequenceMetadata(m: SequenceMetadata): Unit
```
mbeckerle, line 74: Decide if these exist or not.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/InfosetImpl.scala
```diff
@@ -14,13 +14,18 @@
...
  * limitations under the License.
  */
-
 package org.apache.daffodil.runtime1.infoset
 
-import java.lang.{ Boolean => JBoolean }
-import java.lang.{ Number => JNumber }
-import java.math.{ BigDecimal => JBigDecimal }
-import java.util.HashMap
+import java.lang.{
```
mbeckerle, line 19: Revise these imports into standard style.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/InfosetOutputter.scala
```diff
@@ -1,22 +1,9 @@
-/*
```
mbeckerle, line 1: Restore banner.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
```diff
@@ -514,6 +508,13 @@ class DataProcessor private[japi] (private var dp: SDataProcessor)
    */
   def save(output: WritableByteChannel): Unit = dp.save(output)
 
+  /**
+   * Walks the handler over the runtime metadata structures
+   *
+   * @param handler - the handler is called-back during the walk as each metadata structure is encountered.
+   */
+  def walkMetadata(handler: MetadataHandler) = dp.walkMetadata(handler)
```
mbeckerle, line 516: Comment from prior PR: 
https://github.com/apache/daffodil/pull/1092/files#r1360829216

Substance of it is why do we need another walker. Answer is in the commit message, but in short - because it makes more sense to do it on the runtime1 metadata objects since we're interfacing to runtime1, not to the schema compiler. 

This may need more study. The NiFi integration uses a DSOM walker. Consolidation may be possible.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/MetadataWalker.scala
```diff
@@ -0,0 +1,76 @@
+package org.apache.daffodil.runtime1.processors
```
mbeckerle, line 1: Apache banner missing

mbeckerle, line 1: Comment from prior PR:
https://github.com/apache/daffodil/pull/1092/files#r1360834018

Asked if this should be in some specific different package like perhaps "dfdlx.japi" or japix ?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/MetadataWalker.scala
```diff
@@ -0,0 +1,76 @@
...
+  private def walkSimpleElement(
+    handler: MetadataHandler,
+    erd: ElementRuntimeData,
+  ): Unit = {
+    handler.simpleElementMetadata(erd)
+  }
+
+  private def walkSequence(handler: MetadataHandler, sm: SequenceMetadata): Unit = {
+    val srd = sm.asInstanceOf[SequenceRuntimeData]
+    if (!srd.isHidden) {
```
mbeckerle, line 59: Prior PR comment:
https://github.com/apache/daffodil/pull/1092/files#r1360846271

Was observing the need for this isHidden handling. 

The commit comment has a TODO about verifying that we don't share groups between hidden/not.

daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
mbeckerle: Actually I thought of a better argument for why walking the runtime1 metadata is better than walking the DSOM objects.

You don't need the Schema source. It works on the pre-compiled binary schema just as well as if the schema was just compiled. This bypasses the need for Daffodil's schema compiler to be involved at all in interfacing to say, Apache Drill or other data fabrics. A pre-compiled DFDL schema is all that is needed.

There's a further advantage. At runtime, when you are actually parsing data, the metadata attached to the infoset objects is the runtime1 metadata objects, and the fact that those are the exact same objects you walked if you walk the metadata first in a preparation/compilation step is helpful, because at runtime it is useful to sometimes use that metadata - one would be for built-in-self-checking - make sure the metadata at run time is the expected metadata, but the other is that at runtime sometimes you actually need metadata.

The example I can think of where this is useful is if a DFDL schema has a choice, and two branches of the choice both contain an element named "message", but these are of different complex types. We actually have this in a number of large message format schemas. Now, the system we are interfacing with may not have the ability for different choice branches to have children with overlapping names like this. So it may need to call these message1, message2,.... and so on. At runtime when Daffodil dispatches a startComplex event call to the InfosetOutputter, we need to look at the actual runtime metadata (ERD) of the message element to determine which one it actually is, so we can choose to populate the correct one of message1, message2, etc.

This is very much like the next-element resolver that is in the runtime1 metadata. (Exposing that so it can be reused meaningfully in this case is the obvious next thought.)

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JDOMInfosetOutputter.scala
```diff
@@ -46,7 +49,8 @@ class JDOMInfosetOutputter extends InfosetOutputter with XMLInfosetOutputter {
     result = Maybe(root.asInstanceOf[org.jdom2.Document])
   }
 
-  def startSimple(diSimple: DISimple): Unit = {
+  override def startSimple(se: InfosetSimpleElement): Unit = {
+    val diSimple = se.asInstanceOf[DISimple]
```
stevedlawrence, line 53: Is casting to a DISimple done to avoid having to update other outputters and deal with that later? I assume the goal is these infoset outputters should only use InfosetFooEelements? Doing the conversion would be good evidence that the InfosetElement API interface makes everything needed available.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/ScalaXMLInfosetOutputter.scala
```diff
@@ -17,20 +17,23 @@
...
 import org.apache.daffodil.lib.xml.XMLUtils
+import org.apache.daffodil.runtime1.api.InfosetArray
+import org.apache.daffodil.runtime1.api.InfosetComplexElement
+import org.apache.daffodil.runtime1.api.InfosetSimpleElement
 import org.apache.daffodil.runtime1.dpath.NodeInfo
 
+import scala.collection.mutable.ListBuffer
+import scala.xml.MetaData
+import scala.xml.Null
+import scala.xml.UnprefixedAttribute
```
stevedlawrence, line 32: You might need to change your IDE settings. I think scala imports are supposed to be before daffodil imports. Can you configure your IDE to use the .scalafmt.conf file?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/XMLInfosetOutputterMixin.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  def remapped(dataValueAsString: String) =
+    XMLUtils.remapXMLIllegalCharactersToPUA(dataValueAsString)
+
+  /**
+   * String suitable for use in the text of a Processing Instruction.
+   *
+   * The text is a pseudo-XML string.
+   */
+  protected final def fmtInfo(diTerm: DITerm): Maybe[String] = {
```
stevedlawrence, line 34: Should we just drop the fmtInfo stuff? Seems like nothing uses it, and my guess it's is only used for debugging? If we do need access to fmtInfo, we should make that available in the CLI or VS Code debugger.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/MetadataWalker.scala
```diff
@@ -0,0 +1,76 @@
...
+      // handler.startSequence()
+      srd.groupMembers.map { trd =>
+        walkTerm(handler, trd)
+      }
+      // handler.endSequence()
+    }
+  }
+
+  private def walkChoice(handler: MetadataHandler, crd: ChoiceRuntimeData): Unit = {
+    // handler.startChoice()
```
stevedlawrence, line 69: Seems a walker handler would want to know that the children are options of a choice and could be converted to a union or something. Can they know that without start/endChoice?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/RuntimeData.scala
```diff
@@ -689,6 +698,10 @@ sealed class ElementRuntimeData(
       name
     }
   }
+
+  override def namespace: NS = dpathElementCompileInfo.namedQName.namespace
+  override def optNamespacePrefix: Option[String] = dpathElementCompileInfo.namedQName.prefix
+  override def sscd: String = dpathElementCompileInfo.sscd
```
stevedlawrence, line 704: I've forgotten what does sscd stands for. Which makes me thinks it's used rarely enough that maybe it needs a more descriptive name? Could be done in another PR, since it's not related to this.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/Infoset.scala
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * API access to array objects in the DFDL Infoset
+ */
+trait InfosetArray extends InfosetItem {
+  override def metadata: ElementMetadata // the element that is an array
+  def getOccurrence(occursIndex: Long): InfosetElement
+  def length: Long
+  final def apply(occursIndex: Long): InfosetElement = getOccurrence(occursIndex)
+  final def apply(occursIndex: Int) : InfosetElement = getOccurrence(occursIndex.toLong)
```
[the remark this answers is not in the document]
mbeckerle: I will do this last.

mbeckerle:
@stevedlawrence This is ready for a next review pass. I've addressed all the prior comments I believe, and cleaned up several other things. 

Once merged, I want to run a large schema regression test with as many schemas as possible to verify this hasn't broken things.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -70,6 +70,11 @@ abstract class SequenceTermBase(
 
   def isOrdered: Boolean
 
+  /**
+   * Overridden in sequence group ref
+   */
+  def isHidden: Boolean = false
```
mbeckerle, line 76: This requires a test to ensure that if the same group definition is used twice, one from a dfdl:hiddenGroupRef, the other from a regular group ref, that Daffodil recognizes this and makes two disjoint copies.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/InfosetOutputterImpl.scala
```diff
@@ -0,0 +1,47 @@
...
+abstract class InfosetOutputterImpl() extends BlobMethodsImpl with InfosetOutputter {
+
+  import Status._
+
+  private def status: Status = READY
+
+  override def getStatus(): Status = {
+    // Done, Ready (Not started), Visiting (part way done - can retry to visit more)...
+    status
+  }
```
stevedlawrence, line 31: This staus stuff doesn't even seem to be use? Can we remove it? I dont' remember what it was originally for.

stevedlawrence (commented):
Looks really good. A number of comments with fresh eyes. 

I think my main question is what this all looks like from a Javadoc and Java API implementation perspective, so that weird scala stuff doesn't leak into it and that all the members are visible in the doc. I think all the new MetaData and Infoset traits aren't defined anywhere where are java doc is generated, so we need to figure out how that will work. I dont' really want more proxy classes.

Maybe we can plan to drop the japi/sapi stuff, and merge that and the new stuff into a the api package and finally get rid of the proxy stuff. 4.0.0 is nearing, so maybe it's time to discuss fixing our API.

mbeckerle:
@stevedlawrence @tuxji  This should be very close to ready now. 
Please give this a review again.

tuxji:
> @stevedlawrence @tuxji This should be very close to ready now. Please give this a review again.

I'll start reviewing this PR.  Quick question, though.  Is this new walker API intended to be used by the C code generator too, instead of the C generator walking the compiled schema by hand?

mbeckerle:
Not intended for that use case.

This new metadata walker is for java (typically) applications to interface
to runtime1 specifically.

The c backend is more of a compiler integration.

daffodil-core/src/test/scala/org/apache/daffodil/core/infoset/TestInfoset.scala
```diff
@@ -183,16 +191,18 @@ class TestInfoset1 {
...
+      case arr: InfosetArray => {
         assertEquals(2, arr.length)
-        var a = arr(1).asInstanceOf[InfosetSimpleElement]
-        assertEquals(w_erd, a.runtimeData)
+        var a = arr(1).asInstanceOf[InfosetSimpleElement] // 1-based
+        assertEquals(w_erd, a.metadata)
 
-        assertEquals(4, a.dataValue.getAnyRef)
-        assertEquals(infoset, a.parent)
+        assertEquals(4, a.getUnsignedLong.longValue().toInt)
```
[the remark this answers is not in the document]
mbeckerle: These return java types. java.math.BigInteger didn't seem to have a direct conversion to int. Hence this.

mbeckerle:
> +1
> 
> I looked at the javadoc and it doesn't include any of the new stuff since it's not in the japi project. That's expected, and doesn't have to be fixed for this PR, but we should make sure that is addressed at some point, maybe once the drill integration is complete and we are confident that the new API is sufficient.

Yeah, we have to look into how to make the doc "hang together" despite some of the code being defined in java and some (most really) in scala, and some being in japi/sapi packages and some in runtime1/api package. 

We may have to put in our own relative hyperlinks between the two bodies of documentation.

We also really need a way to document and exclude some methods/members from the API without having to have a sharp separation between things in the sapi/japi packages which are *all* API, and things not. We need to avoid the overhead of this proxy japi/sapi structure. It's too hard to maintain.  I'd be fine with just adding to the java/scaladoc a standard comment and naming convention indicating that a method is internal not to be considered part of the API. These methods do need to be public. Even just naming them with prefix "_" and always putting them last in the trait/class/object would be ok.

PR #1121, opened 2023-11-20
Make code gen support EmptyGram

tuxji:
Fix the Daffodil C code generator to stop crashing on someone's private DFDL schema.  Their schema is private and too large to include, but somehow it creates an EmptyGram object.  Make sure the code gen recognizes the EmptyGram object instead of crashing when seeing it.

c/DaffodilCCodeGenerator.scala: Add clause to support EmptyGram.

DAFFODIL-2863

### DIFF
--- a/daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
+++ b/daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
@@ -259,6 +259,9 @@ object DaffodilCCodeGenerator
    */
   def generateCode(gram: Gram, cgState: CodeGeneratorState): Unit = {
     gram match {
+      // Skip empty grams
+      case g: Gram if g.isEmpty => noop(g)
+      // Handle non-empty grams
       case g: AlignmentFill => alignmentFillGenerateCode(g, cgState)
       case g: AssertBooleanPrim => noop(g)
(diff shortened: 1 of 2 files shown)


daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
```diff
@@ -275,6 +276,7 @@ object DaffodilCCodeGenerator
       case g: ElementParseAndUnspecifiedLength =>
         elementParseAndUnspecifiedLengthGenerateCode(g, cgState)
       case g: ElementUnused => noop(g)
+      case g: Gram if g eq EmptyGram => noop(g)
```
stevedlawrence, line 279: It might better to add this as the first case instead:
```scala
case g: Gram if g.isEmpty => noop(g)
```
This will still noop `EmptyGram`'s like you have, but also noop non-`EmptyGram` grammars that determine that they are empty. I'm not sure if it's possible for codgen-c to walk into those or if you'll only ever see empty grammars that are `EmptyGram`, but it wouldn't hurt.

tuxji: Thanks, I have improved the fix as suggested and added a new test case with a test schema containing an empty gram.  Ready for review again.

stevedlawrence (approved):
+1 :+1:

PR #1126, opened 2023-12-06
Run SonarCloud using Java 17

pkatlic:
Use Java 17 instead of Java 11 for SonarCloud.

DAFFODIL-2856

main.yml: Update Java version.


stevedlawrence:
+1, we only run the sonarcloud scan when PRs are merged, so we'll want to make sure to double check the github action once that happens

PR #1127, opened 2023-12-08
Give codegen-c somewhat limited dfdl:assert support

tuxji:
On the basis that a little support is better than none, allow codegen-c to generate C code for a limited subset of dfdl:assert expressions.  The use case is to allow generated code to validate or reject messages using dfdl:assert expressions as well as ranges and enumerations.  The dfdl:assert support is limited to replacing "../foo/x" and "."  with the appropriate pointers and replacing DFDL named operators ("eq") with C-compatible operators ("==") using find and replace operators.  In the long term, a better approach would be to add methods to the compiled expression classes to generate C code directly from the compiled expressions with fuller understanding of what the parts of the expressions represent.

DAFFODIL-2865

errors.[ch]: Add new ERR_DFDL_ASSERT error code for assertion expressions which fail.

validators.[ch]: Add new validate_dfdl_assert function for checking assertion expressions.  Handle both recoverableError and processingError failure types appropriately.

DaffodilCCodeGenerator.scala: Call AssertStatementGenerateCode to generate C code for assertion expressions in both empty sequences and simple elements.

AssertStatementGenerateCode.scala: Generate C code for two types of assertion expressions: "{ ../../foo/x eq 2 }" as used in empty sequences and "{ . eq 2 }" as used in simple elements.  Anything else likely won't produce valid C code at this time.

CodeGeneratorState.scala: Make the cExpression function find and replace even more DFDL named operators with C-compatible operators.

NestedUnion/generated_code.c: Regenerate to show new C code for assertion expressions in simple elements.

data/nested.struct.{diag{1,2},err{1,2}}.dat: Add new data files which will deliberately fail assertion expression checks.

infosets/nested.struct.{diag{1,2},err{1,2}}.dat.xml: Add new infoset files which will deliberately fail assertion expression checks.

{data,infosets}/nested.union.bar.dat{,.xml}: Modify to pass assertion expression checks.

nested.dfdl.xsd: Add both kinds of assertion expression checks (empty sequences and simple elements) to test the new codegen-c support.

nested.tdml: Add new validation test cases to check both failure types of assertion expression checks.

TestNested.scala: Call the new validation test cases.


daffodil-codegen-c/src/main/resources/org/apache/daffodil/codegen/c/files/libruntime/validators.c
```diff
@@ -38,6 +38,32 @@ validate_array_bounds(const char *name, size_t count, size_t minOccurs, size_t m
...
+
+void
+validate_dfdl_assert(bool assertion, const char *expression, const char *failure_type,
+                     ParserOrUnparserState *pu)
+{
+    // The boolean expression that determines if the assertion is true
+    // is evaluated in generated code and the result passed here, as
+    // well as failure_type which determines whether to issue a diagnostic
+    // which is printed later or an error which terminates processing now.
+    if (!assertion && failure_type[0] == 'r')
```
stevedlawrence, line 51: Thoughts on converting `failure_type` to a C enum instead of checking the first letter of a string? This seems a bit fragile.

Or maybe replace the `failure_type` parameter with a `bool recoverable` and then pass in `assert.failureType == FailureType.RecoverableError`?

daffodil-codegen-c/src/main/resources/org/apache/daffodil/codegen/c/files/libruntime/validators.c
```diff
@@ -38,6 +38,32 @@ validate_array_bounds(const char *name, size_t count, size_t minOccurs, size_t m
...
+    {
+        Diagnostics *diagnostics = get_diagnostics();
+        const Error error = {ERR_DFDL_ASSERT, {.s = expression}};
+
+        add_diagnostic(diagnostics, &error);
+        pu->diagnostics = diagnostics;
+    }
+    else if (!assertion)
+    {
+        static Error error = {ERR_DFDL_ASSERT, {0}};
```
stevedlawrence, line 61: Unrelated to this PR, but just out of curiosity, is there a reason you use static `Error`'s for pu error? I guess an alternative would be to make the `struct Error` in `ParserOrUnparserState` be a non-pointer and do something like:
```c
pu->error.code = ERR_DFDL_ASSERT;
pu->error.s = expression;
```
I have no preference either way, just wondering since I feel using static is less common in C. But I also haven't done C development in a while, so maybe I'm forgetting a subtle difference. Or is it just preference?

And after asking, I now see that `pu->error == NULL` means no error, which you can't do if error is a non-pointer. I guess with the non-pointer approach you would need a special code of ERR_NO_ERROR or something, and do `pu->error.code == ERR_NO_ERR`, which is too verbose. That could be wrapped in a function, e.g. `if (is_error(pstate->pu)) return;`, but it's probably not worth the refactor.

So nevermind. I'll leave this in case others have the same question as I did.

daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
```diff
@@ -306,6 +310,17 @@ object DaffodilCCodeGenerator
   private def elementCombinator(g: ElementCombinator, cgState: CodeGeneratorState): Unit = {
     cgState.pushElement(g.context)
     DaffodilCCodeGenerator.generateCode(g.subComb, cgState)
+    // Also generate code for the element's assert statements
+    g.context.assertStatements.foreach { assert =>
+      assert.testKind match {
+        case TestKind.Pattern =>
```
stevedlawrence, line 316: Is the convention for the C code generator to silently ignore things it doesn't support? I imagine an SDE or Warning might be useful in some cases, but if this is what you've been doing this is fine.

stevedlawrence (approved):
+1 :+1:  minor comments

mike-mcgann (approved):

PR #1129, opened 2023-12-11
Use encodingErrorPolicy replace, even when set to error

stevedlawrence:
If encodingErrorPolicy="error" we print a warning letting the user know it is not supported and that we will use "replace" instead. This is so that we can support schemas written for IBM DFDL (which only supports "error") and have the exact same behavior as IBM DFDL except for when data has encoding errors, which is usually rare.

However, even though we warn that we will use "replace", we do not actually do so, which leads to an assertion exception if there are encoding errors. This fixes that so our behavior matches the warning message and we always use "replace".

DAFFODIL-2861

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
@@ -678,6 +678,7 @@
 
 		<tdml:document>
 			<tdml:documentPart type="text">1234</tdml:documentPart>
+			<tdml:documentPart type="byte">80</tdml:documentPart> <!-- invalid in US-ASCII -->
 		</tdml:document>
 
     <tdml:warnings>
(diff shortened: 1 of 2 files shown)


tuxji (approved):
+1

TermEncodingMixin.scala comment looks good.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
```diff
@@ -693,6 +694,7 @@
           <ex:char>2</ex:char>
           <ex:char>3</ex:char>
           <ex:char>4</ex:char>
+          <ex:char>�</ex:char>
```
mbeckerle, line 697: Nit, but I think a character entity would be better here than this literal unicode character. `&#xFFFD;` along with an XML comment saying this is the unicode replacement character.

mbeckerle (approved):
+1
One minor suggestion

PR #1130, opened 2023-12-11
Return FAIL if tests are not found

pkatlic:
Include specified test names in test list even if not found in test file in order to print names of tests not found. Return FAIL if no tests pass or any tests fail or are not found.

DAFFODIL-2699

Main.scala: Include all test names for listing even if not found, update return code logic to check for no passing tests and any tests failing or not found.

TestCLItdml.scala: Update expected exit codes based on new return code logic, note that test_CLI_Tdml_Listing_negativeTest01 now returns FAIL due to tests not found and test_CLI_catch_TestNotCompatible now returns FAIL due to no passing tests.


tuxji (approved):
+1 assuming CI passes

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -1729,7 +1732,7 @@ class Main(
             ),
           )
 
-          if (fail == 0) ExitCode.Success else ExitCode.TestError
+          if (pass > 0 && fail == 0 && notfound == 0) ExitCode.Success else ExitCode.TestError
```
stevedlawrence, line 1735: Should we remove the `pass > 0` condition? I'm thinking of an an edge case where a tdml file has no tests defined and a user runs `daffodil test no_tests.tdml`. If we require `pass > 0` to be considered a success, then that command would exit with failure even though nothing technically failed. Maybe there's an argument that running no tests should be a failure? But using SBT as an example, running a test suite with no tests is considered a success. Maybe there are other counter examples?

stevedlawrence (approved):
+1, one question about the `pass > 0` condition

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
tuxji: OK, agreed that pass > 0 can be removed.

PR #1131, opened 2023-12-14
Verify that debuggers are reset after TDML tests are run

stevedlawrence:
Current behavior in Daffodil is to automatically disable debugging after a call to runOneTest, which helps to avoid errors where enabling debugging in one test might cause failures in another test.

This updates a test to use the latest TDML API and confirm that this behavior actually happens. Note that the schema that originally found the issue with long-lived debugger state is no longer valid in Daffodil, so it needed some tweaks. This also simplifies the custom debugger to avoid storing strings that are never needed.

DAFFODIL-790

### DIFF
--- a/daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestUserSubmittedTests.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestUserSubmittedTests.scala
@@ -17,9 +17,13 @@
 
 package org.apache.daffodil.usertests
 
+import org.apache.daffodil.core.dsom.ExpressionCompilers
+import org.apache.daffodil.runtime1.debugger.InteractiveDebugger
+import org.apache.daffodil.runtime1.debugger.TraceDebuggerRunner
 import org.apache.daffodil.tdml.Runner
 
 import org.junit.AfterClass
+import org.junit.Assert._
 import org.junit.Test
 
 object TestUserSubmittedTests {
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/usertests/TestUserSubmittedTests.scala
```diff
@@ -52,49 +56,38 @@ class TestUserSubmittedTests {
...
-  def fini(): Unit = dr.fini
-}
+    // run a test with the debugger and debugging enabled so that we count the lines. runOneTest
+    // will disable debugging when the test completes
+    runner.runOneTest("test_DFDL_782")
+    assertTrue(crunner.numLines > 0)
 
-class CustomTraceRunner extends TraceRunner {
-  private var _lines = List.empty[String]
+    // reset the numeLines counter to 0
```
mbeckerle, line 71: Typo numeLines

mbeckerle (approved):
+1

I found a typo in a comment.

tuxji (approved):
+1

I'm not familiar with this approach to debugging Daffodil.  Do I understand that the InteractiveDebugger class here does not mean that you interact with a live interactive debugger command line, rather, it lets you supply a callback "listener" class which can trace parts of Daffodil execution and print or accumulate data to help you debug a problem?

stevedlawrence:
> Do I understand that the InteractiveDebugger class here does not mean that you interact with a live interactive debugger command line, rather, it lets you supply a callback "listener" class which can trace parts of Daffodil execution and print or accumulate data to help you debug a problem?

Yep. There's a number of commands that the interactive debugger understands (e.g. "break", "step", "continue", "eval"). The "listener" or "debugger runner" is what provides the commands to run and receives the results to display. The name "interactive" is probably not the best since it really doesn't need to have user interaction at all (e.g the `TraceDebuggerRunner` just uses a list of predetermined commands). But I think it originally did require user input, but was eventually generalized and the name stuck.

PR #1132, opened 2023-12-22
Add test that illustrates float text parse bug

mbeckerle:
This is a representation="text" xs:float that is
delimited. Parsing Float.MaxValue i.e., this text string:
```
3.4028235E38
```
I get this error:
```
Parse Error: Value '340282350000000000000000000000000000000' is out of range for type: xs:float
```

No similar issue for type xs:double.

DAFFODIL-2867

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
@@ -139,6 +139,9 @@
       dfdl:textNumberPattern="000000000000000000000" />
     <xs:element name="floatTextFail" type="xs:float"
       dfdl:lengthKind="implicit" />
+    <xs:element name="floatTextDelim" type="xs:float"
+                dfdl:representation="text" dfdl:lengthKind="delimited"
+                dfdl:terminator="%SP;"/>
 
     <xs:element name="nonNegIntBin" type="xs:nonNegativeInteger"
       dfdl:representation="binary" dfdl:lengthKind="explicit" dfdl:length="{ 16 }" />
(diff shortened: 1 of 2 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -3719,6 +3722,17 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
+  <tdml:parserTestCase name="float_text_delim" root="floatTextDelim"
+                       model="SimpleTypes-Embedded.dfdl.xsd" description="Section 5 Simple type-float - DFDL-5-008R">
+
+    <tdml:document><![CDATA[3.4028235E38]]></tdml:document>
```
tuxji, line 3728: The document does not end with the terminator (%SP;) expected by the floatTextDelim type.  Can you comment on what the implications are?

tuxji (approved):
+1 with a question

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
stevedlawrence: Fields with `lengthKind="delimited"` can be delimited by more than just separators, including EOF. From the spec:

> On parsing, the length of an element with dfdl:lengthKind 'delimited' is determined by scanning the data stream for the delimiter.
> 
> The data stream is scanned for any of
> ·         the element's terminator (if specified)
> ·         an enclosing construct's separator or terminator
> ·         the end of an enclosing element designated by its known length
> ·         the end of the data stream

So the dfdl:separator="%SP;" isn't really needed in this case, but it doesn't hurt.

stevedlawrence (approved):
+1

PR #1133, opened 2024-01-03
Fix xs:float text number parse/unparse edge cases

stevedlawrence:
When parsing text floats, ICU sometimes returns a BigInteger, such as the case when parsing the maximum single precision floating-point value (i.e. 3.40282347E+38).

The current logic for checking if these parsed BigIntegers are in range of a float is to convert it to a double and then compare that against the maximum value of a float (promoted to a double). However, converting a BigInteger to a double can lead to changes in precision, making this comparison inaccurate. Looking at bit representations makes this more clear:

Float.MaxValue.doubleValue converted to bits is:

    0 10001111110 1111111111111111111111100000000000000000000000000000

But BigInteger(Float.MaxValue).doubleValue converted to bits is:

    0 10001111110 1111111111111111111111100101010011011010111111111000

The rules for converting a float to a double are different than converting a BigInteger to a double, which leads to changes in precision and a double value that is slightly bigger than the actual float max value, leading to out of range errors.

Because converting a BigInteger to a double is unreliable, this adds a new case for BigInteger, which converts it to a BigDecimal and compares that against the max value, which avoids precision issues. Note that this also changes the min/maxBD values to be created from a string rather than a double, since converting a double to a BigDecimal also has related precision issues.

Additionally, when unparsing text floats, ICU does not have a way to unparse single precision floats. Instead, it converts the float to a double and then unparses that. However, this leads to text numbers with extra precision that implies more accuracy than exists for floats, and leads to errors with some round trip tests. To fix this, this converts the float to a BigDecimal using the String constructor (which maintains prcision), and then has ICU format the BigDecimal. This is more expensive, but leads to more correct text floats.

DAFFODIL-2867

### DIFF
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
@@ -140,8 +140,7 @@
     <xs:element name="floatTextFail" type="xs:float"
       dfdl:lengthKind="implicit" />
     <xs:element name="floatTextDelim" type="xs:float"
-                dfdl:representation="text" dfdl:lengthKind="delimited"
-                dfdl:terminator="%SP;"/>
+                dfdl:representation="text" dfdl:lengthKind="delimited" />
 
     <xs:element name="nonNegIntBin" type="xs:nonNegativeInteger"
       dfdl:representation="binary" dfdl:lengthKind="explicit" dfdl:length="{ 16 }" />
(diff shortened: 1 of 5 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -3722,10 +3721,10 @@
     </tdml:infoset>
   </tdml:parserTestCase>
 
-  <tdml:parserTestCase name="float_text_delim" root="floatTextDelim"
+  <tdml:parserTestCase name="float_text_max" root="floatTextDelim"
                        model="SimpleTypes-Embedded.dfdl.xsd" description="Section 5 Simple type-float - DFDL-5-008R">
 
-    <tdml:document><![CDATA[3.4028235E38]]></tdml:document>
+    <tdml:document><![CDATA[340,282,350,000,000,000,000,000,000,000,000,000,000]]></tdml:document>
```
mbeckerle, line 3727: I had to stare at this a while and go look at GeneralFormatBase to decide this is ok.

Insert a comment that this is 39 digits long because the textNumberPattern is "#,###.###" which doesn't allow for use of an exponent, so it has to output all the integer digits.

mbeckerle (approved):
+1 minor comments only

Great that this gets rid of a bunch of twoPass tests that were highly unmotivated.

tuxji (approved):
+1

Looks fine.

PR #1137, opened 2024-01-09
fn:count(non-array) must issue SDE

pkatlic:
Throw a SDE during compilation if the fn:count function is called on a path
expression that ends with a unique scalar element.

DAFFODIL-2711

Expression.scala: Remove unused isTypeCorrect and checkTypeCorrectness properties. Add functions to check for Array or Optional elements and combine common code, update naming to include OrOptional. Rename isOptional to isArrayOrOptional.  Check for isLastStep in downwardStep. Add checkArgArrayOrOptional function to check for an array or optional element and call for FNCount and FNExactlyOneExpr expressions.

ElementBaseRuntime1Mixin.scala: Add isOptional property.

dafext.xsd: Remove unused enumeration.

ArrayRelated.scala: Modify FNCount run implementation to handle Optional elements.

DState.scala: Remove isAnArray function since it is now redundant.

FNFunctions.scala: Fix typo.

CompiledExpression1.scala: Add isOptional property.

RuntimeData.scala: Add isOptional to constructor.

testWarnings.tdml: Update error message check.

TestTDMLRunnerWarnings.scala: Update error message check.

PrefixedTests.tdml: Update element to use fn:exists instead of fn:count.

expressions.tdml: Update element to use fn:exists instead of fn:count.

Functions.tdml: Add tests for fn:count, update error message checks and comments. Note that using a group reference for an optional when combined with an array will throw a ClassCastException, this may be a bug where an optional is evaluated as a DownArray instead of a DownElement. The issue DAFFODIL-2879 contains more details.

TestDFDLExpressions.scala: Add new test cases for Functions.tdml.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
@@ -89,17 +89,6 @@ abstract class Expression extends OOLAGHostImpl() with BasicComponent {
     res
   }
 
-  // This split allows overrides of this lazy val to still reuse this
-  // (super.isTypeCorrect doesn't work inside an overridden lazy val.)
-  lazy val isTypeCorrect: Boolean = checkTypeCorrectness
-  //
-  // override for other checking beyond what is needed to do conversions
-  //
-  protected lazy val checkTypeCorrectness =
-    children.forall { child =>
-      child.isTypeCorrect
-    }
-
   def text: String
 
   def hasReferenceTo(elem: DPathElementCompileInfo): Boolean = {
@@ -744,21 +733,28 @@ abstract class PathExpression() extends Expression {
    * about an array such as its current count.
    */
   lazy val isPathToOneWholeArray: Boolean = {
-    if (steps == Nil) false // root is never an array
-    if (steps == Nil) false // root is never an array
-    else if (isSelf) false // self path is never an array
-    else
-      steps.last.isArray &&
-      !steps.last.pred.isDefined && // last cannot have a [N] pred
-      steps.dropRight(1).forall {
-        // for all the prior steps
-        // if they mention an array element, there
-        // must be a [N] predicate.
-        step =>
-          if (step.isInstanceOf[Up]) true
-          else if (step.isArray) step.pred.isDefined
-          else true
-      }
+    isNotRootOrSelf &&
+    steps.last.isArray &&
+    steps.last.pred.isEmpty
+  }
+
+  /**
+   * Path to a single array element without a [N] predicate qualifying it
+   * or to an optional element.
+   *
+   * That is, the kind of path expression used to access information
+   * about an array or optional such as its current count.
+   */
+  lazy val isPathToOneWholeArrayOrOptional: Boolean = {
+    isNotRootOrSelf &&
+    steps.last.isArrayOrOptional &&
+    steps.last.pred.isEmpty
+  }
+
+  private lazy val isNotRootOrSelf: Boolean = {
+    if (steps == Nil) false
+    else if (isSelf) false
+    else true
   }
 
   def isSelf = steps match {
@@ -969,16 +965,27 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
   }
 
   final lazy val isArray: Boolean = {
-    val (arrays, scalars) = stepElements.partition { _.isArray }
-    (arrays.length, scalars.length) match {
+    checkAmbiguousPath
+    stepElements.exists(_.isArray)
+  }
+
+  final lazy val isArrayOrOptional: Boolean = {
+    checkAmbiguousPath
+    stepElements.exists(_.isArray) || stepElements.exists(_.isOptional)
+  }
+
+  final lazy val checkAmbiguousPath: Unit = {
+    val (arraysOrOptionals, scalars) = stepElements.partition { se =>
+      se.isArray || se.isOptional
+    }
+    (arraysOrOptionals.length, scalars.length) match {
       case (a, s) if (a > 0 && s > 0) =>
-        arrays.head.SDE(
-          "Path step is ambiguous. It can be to an array or a non-array element.\n" +
+        arraysOrOptionals.head.SDE(
+          "Path step is ambiguous. It can reference both an array or a non-array element, which is not allowed.\n" +
             "One of the non-arrays is %s",
           scalars.head.schemaFileLocation.toString,
         )
-      case (a, s) if (a == 0) => false
-      case _ => true
+      case _ => // do nothing
     }
   }
 
@@ -1215,7 +1222,7 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
         Assert.invariant(pred.get.targetType == NodeInfo.ArrayIndex)
         val indexRecipe = pred.get.compiledDPath
         new DownArrayOccurrence(nqn, indexRecipe)
-      } else if (targetType == NodeInfo.Exists) {
+      } else if (isLastStep && targetType == NodeInfo.Exists) {
         new DownArrayExists(nqn)
       } else {
         schemaDefinitionUnless(
@@ -2295,6 +2302,20 @@ abstract class FunctionCallBase(
     res
   }
 
+  protected def checkArgArrayOrOptional(n: Int): Unit = {
+    lazy val isArrayOrOptional = expressions(n) match {
+      case pe: PathExpression => pe.isPathToOneWholeArrayOrOptional
+      case _ => false
+    }
+    if (!isArrayOrOptional) argArrayOrOptionalErr()
+  }
+  protected def argArrayOrOptionalErr() = {
+    SDE(
+      "The %s function requires an array or optional path expression.",
+      functionQName.toPrettyString,
+    )
+  }
+
   final def checkArgCount(n: Int): Unit = {
     if (expressions.length != n) argCountErr(n)
   }
@@ -2335,7 +2356,7 @@ abstract class FunctionCallBase(
 /**
  * Tells the sub-expression that we want an array out of it.
  */
-abstract class FunctionCallArrayBase(
+abstract class FunctionCallArrayOrOptionalBase(
   nameAsParsed: String,
   fnQName: RefQName,
   args: List[Expression],
@@ -2348,11 +2369,6 @@ abstract class FunctionCallArrayBase(
     case _ => subsetError("The %s function must contain a path.", funcName)
   }
 
-  override lazy val isTypeCorrect = {
-    checkArgCount(1)
-    arrPath.isPathToOneWholeArray
-  }
-
   override def targetTypeForSubexpression(subExp: Expression): NodeInfo.Kind = NodeInfo.Array
 
 }
@@ -2361,7 +2377,7 @@ abstract class FunctionCallArrayBase(
  * Tells the sub-expression that we want an array out of it.
  */
 case class FNCountExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expression])
-  extends FunctionCallArrayBase(nameAsParsed, fnQName, args) {
+  extends FunctionCallArrayOrOptionalBase(nameAsParsed, fnQName, args) {
 
   val funcName: String = "fn:count"
 
@@ -2369,6 +2385,7 @@ case class FNCountExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expre
 
   override lazy val compiledDPath = {
     checkArgCount(1)
+    checkArgArrayOrOptional(0)
     val arg0Recipe = args(0).compiledDPath
     val arg0Type = args(0).inherentType
     val c = conversions
@@ -2378,7 +2395,7 @@ case class FNCountExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expre
 }
 
 case class FNExactlyOneExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expression])
-  extends FunctionCallArrayBase(nameAsParsed, fnQName, args) {
+  extends FunctionCallArrayOrOptionalBase(nameAsParsed, fnQName, args) {
 
   val funcName: String = "fn:exactly-one"
 
@@ -2386,6 +2403,7 @@ case class FNExactlyOneExpr(nameAsParsed: String, fnQName: RefQName, args: List[
 
   override lazy val compiledDPath = {
     checkArgCount(1)
+    checkArgArrayOrOptional(0)
     subsetError("fn:exactly-one is not supported.")
   }
 }
@@ -2797,6 +2815,7 @@ sealed abstract class LengthExprBase(
         // $COVERAGE-ON$
       }
     }
+    // checks specifically for an array, but does not apply to an optional
     if (path.isPathToOneWholeArray)
       SDE(s"First argument to ${fnQName} cannot be a path to an array")
     val steps = path.steps
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DState.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DState.scala
@@ -41,7 +41,6 @@ import java.math.{ BigInteger => JBigInt }
 
 import org.apache.daffodil.lib.api.DaffodilTunables
 import org.apache.daffodil.lib.api.DataLocation
-import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.runtime1.dsom.DPathCompileInfo
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValuePrimitiveNullable
 
@@ -118,6 +117,7 @@ case class DState(
   tunable: DaffodilTunables,
   val parseOrUnparseState: Maybe[ParseOrUnparseState],
 ) {
+
   import org.apache.daffodil.lib.util.Numbers._
 
   var isCompile = false
@@ -140,6 +140,7 @@ case class DState(
   def setMode(m: EvalMode): Unit = {
     _mode = m
   }
+
   def mode = _mode
 
   //
@@ -203,10 +204,12 @@ case class DState(
     _currentValue = v
     _currentNode = null
   }
+
   def setCurrentValue(v: Long): Unit = {
     _currentValue = v
     _currentNode = null
   }
+
   def setCurrentValue(v: Boolean): Unit = {
     _currentValue = v
     _currentNode = null
@@ -215,45 +218,26 @@ case class DState(
   def booleanValue: Boolean = currentValue.getBoolean
 
   def longValue: Long = asLong(currentValue.getAnyRef)
+
   def intValue: Int = longValue.toInt
+
   def doubleValue: Double = asDouble(currentValue.getAnyRef)
 
   def integerValue: JBigInt = asBigInt(currentValue.getAnyRef)
+
   def decimalValue: JBigDecimal = asBigDecimal(currentValue.getAnyRef)
+
   def stringValue: String = currentValue.getString
 
   def isNilled: Boolean = currentElement.isNilled
 
-  private def isAnArray(): Boolean = {
-    if (!currentNode.isInstanceOf[DIArray]) {
-      Assert.invariant(errorOrWarn.isDefined)
-      if (currentNode.isInstanceOf[DIElement]) {
-        errorOrWarn.get.SDW(
-          WarnID.PathNotToArray,
-          "The specified path to element %s is not to an array. Suggest using fn:exists instead.",
-          currentElement.name,
-        )
-      } else {
-        errorOrWarn.get.SDW(
-          WarnID.PathNotToArray,
-          "The specified path is not to an array. Suggest using fn:exists instead.",
-        )
-      }
-      false
-    } else {
-      true
-    }
-  }
-
   def arrayLength: Long =
-    if (isAnArray()) currentArray.length
-    else 1L
+    currentArray.length
 
-  def finalArrayLength: Long =
-    if (isAnArray()) {
-      currentArray.requireFinal
-      currentArray.length
-    } else 1L
+  def finalArrayLength: Long = {
+    currentArray.requireFinal()
+    currentArray.length
+  }
 
   def exists: Boolean = true // we're at a node, so it must exist.
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
@@ -167,7 +167,7 @@
         <xs:appinfo source="http://www.ogf.org/dfdl/">
           <dfdl:assert test="{ fn:count( /ex:more_count1 ) eq 3 }"
             message="Assertion 
-            failed for fn:count( /ex:more_count ) eq 3" />
+            failed for fn:count( /ex:more_count1 ) eq 3" />
         </xs:appinfo>
       </xs:annotation>
       <xs:complexType>
@@ -182,7 +182,7 @@
         <xs:appinfo source="http://www.ogf.org/dfdl/">
           <dfdl:assert test="{ fn:count( /ex:more_count1b ) eq 1 }"
             message="Assertion 
-            failed for fn:count( /ex:more_count ) eq 1" />
+            failed for fn:count( /ex:more_count1b ) eq 1" />
         </xs:appinfo>
       </xs:annotation>
       <xs:complexType>
@@ -3323,14 +3323,11 @@
     <tdml:document>
       <tdml:documentPart type="text"><![CDATA[#:1]]></tdml:documentPart>
     </tdml:document>
-    <tdml:infoset>
-      <tdml:dfdlInfoset xmlns:xs="http://www.w3.org/2001/XMLSchema"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-        <more_count1b>
-          <i1>1</i1>
-        </more_count1b>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>fn:count</tdml:error>
+      <tdml:error>array</tdml:error>
+    </tdml:errors>
   </tdml:parserTestCase>
 
 <!--
@@ -5146,6 +5143,145 @@
           <xs:element name="term" type="xs:string"
             dfdl:lengthKind="explicit" dfdl:length="0"
             dfdl:terminator="{ if (fn:count(../ex:int) gt 1) then 'x' else 'y' }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countArray">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" minOccurs="2" maxOccurs="2" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countInput">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countOptional">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countOutput">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:outputValueCalc="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countOccurs">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:occursCountKind="expression" dfdl:occursCount="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countIfArrayOptional">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" minOccurs="2" maxOccurs="2" dfdl:representation="binary" />
+          <xs:element name="int2" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:outputValueCalc="{ fn:count(if (../ex:int[1] ne 0) then ../ex:int else ../ex:int2) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:element name="countIfArrayScalar">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" minOccurs="2" maxOccurs="2" dfdl:representation="binary" />
+          <xs:element name="int2" type="xs:int" minOccurs="1" maxOccurs="1" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:outputValueCalc="{ fn:count(if (../ex:int[1] ne 0) then ../ex:int[1] else ../ex:int2) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+
+    <xs:group name="countGroup">
+      <xs:sequence>
+        <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:foo) }"/>
+      </xs:sequence>
+    </xs:group>
+    <xs:element name="scalarOptional">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="scalar">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" dfdl:representation="binary"/>
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="optional">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="0" dfdl:representation="binary"/>
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+    <xs:element name="scalarArray">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="scalar">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" dfdl:representation="binary"/>
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="array">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="2" maxOccurs="2" dfdl:representation="binary"/>
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
+    <xs:element name="arrayOptional">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="array">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="2" maxOccurs="2" dfdl:representation="binary"/>
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="optional">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary"/>
+                <!-- throws ClassCastException, possible bug in Daffodil group evaluation -->
+                <!-- <xs:group ref="countGroup"/> -->
+                <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:foo) }"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
         </xs:sequence>
       </xs:complexType>
     </xs:element>
@@ -11045,18 +11181,11 @@
     <tdml:document>
       <tdml:documentPart type="text"></tdml:documentPart>
     </tdml:document>
-    <tdml:infoset>
-      <tdml:dfdlInfoset>
-        <count02>
-          <calced>calculated string</calced>
-          <count>1</count>
-        </count02>
-      </tdml:dfdlInfoset>
-    </tdml:infoset>
-    <tdml:warnings>
-      <tdml:warning>Runtime Schema Definition Warning</tdml:warning>
-      <tdml:warning>fn:exists</tdml:warning>
-    </tdml:warnings>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>fn:count</tdml:error>
+      <tdml:error>array</tdml:error>
+    </tdml:errors>
   </tdml:parserTestCase>
 
 <!--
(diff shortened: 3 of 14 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -11390,6 +11417,110 @@
...
+  <tdml:parserTestCase name="count_14" root="count09" roundTrip="true"
+                       model="XPathFunctions" description="Section 23 - Functions - fn:count - DFDL-23-122R">
+    <tdml:document>
+      <tdml:documentPart type="byte">00 00 00 01</tdml:documentPart>
+    </tdml:document>
+    <tdml:errors>
+      <tdml:error>Schema Definition Error</tdml:error>
+      <tdml:error>fn:count</tdml:error>
+      <tdml:error>array</tdml:error>
+    </tdml:errors>
```
tuxji, line 11481: Oops, the count_14 test should not expect a SDE (I'm sorry I missed that).  The issue says fn:count() should be an SDE if the path is not for an array or optional element.  The element for count_14 is an optional element so the count should return 0 or 1 depending on whether the input document is empty or has 4 bytes in it (both cases should be tested).

tuxji (changes requested):
Oops, the checkArgArray function in Expression.scala needs to return true if the last step ends in an optional element as well as an array.  Once that's fixed, and if no one else sees a problem, we can approve and merge the PR.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2295,6 +2295,18 @@ abstract class FunctionCallBase(
     res
   }
 
+  final def checkArgArray(): Unit = {
+    lazy val isArray = expressions.head match {
```
stevedlawrence, line 2299: This `checkArgArray` function is only used for `fn:count`, which only has one argument so it's fine to just check `expressions.head`, but without knowing that it could be confusing why this only checks head and not all of the function arguments. I wonder if this could be made more generic somehow? Maybe it takes an index into the expressions list and checks that that specific expression is an array? And then the usage in FnCountExpr would be `checkArgArray(0)`? Or maybe it takes an actual expression, so the usage would be `checkArgArray(args(0))`? I'm not sure if a similar check we do somewhere else that we could copy the pattern from?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2295,6 +2295,18 @@ abstract class FunctionCallBase(
...
+      case rpe: RelativePathExpression => rpe.steps.last.isArray
+      case rpe: RootPathExpression => rpe.steps.last.isArray
```
stevedlawrence, line 2301: Can these two be combined into this?
```scala
case pe: PathExpression => pe.steps.last.isArray
```
I think these are the only classes that extend that base class?

Also, is `isArray` sufficient? I know optional elements (i.e. minOccurs=0 maxOccurs=1) are represented in the infoset as an array, but sometimes aren't considered as actual arrays in other parts of the code. I'm not sure if the StepExpression isArray handles that? I think we want `fn:count` to still work on optional elements?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2295,6 +2295,18 @@ abstract class FunctionCallBase(
...
+      case _ => true
```
stevedlawrence, line 2302: One thing to consider is this doesn't catch expressions like this:
```xquery
fn:count(if (some_condition) then /some/path else /some/other/path)
```
I think that's legal and I think would be considered invalid because of this case.

I'm not sure if there's a good way to check that all branches of a subexpression result in an array. We must be able to get some basic information since we can statically check that all branches have the same type. I think we do that via `inherentType`, but I'm not sure the result would return information about being an array or not. Maybe this is just a limitation of our checking and it's good enough? And if someone whats the above, they would need to do something like this:
```xquery
if (some_condition) then fn:count(/some/path) else fn:count(/some/other/path)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DState.scala
```diff
@@ -215,45 +218,26 @@ case class DState(
...
-          WarnID.PathNotToArray,
-          "The specified path is not to an array. Suggest using fn:exists instead.",
-        )
-      }
-      false
-    } else {
-      true
-    }
-  }
-
```
stevedlawrence, line 247: So this is an 100% static check now? Nice! It might be worth adding a comment to the `arrayLength` and `finalArrayLength` functions that we must statically know current dpath result is an array, otherwise these might return some weird results or throw an exception. They are currently only used by FNCount so it should be safe, but if another function ever uses those we need to make sure they are doing the checkArgArray thing.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
stevedlawrence: Agreed, I think that is the correct behavior. Only minOccurs=maxOccurs=1 should trigger the SDE.

stevedlawrence (commented):
Nice change, it'll nice to have this all be static checking and no runing timechecking. A couple minor comments/questions. Also, before we merge this we'll need to rebase to remove the merge commit.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DState.scala
stevedlawrence: I'm wondering if we can get away with just removing the warnID? The only place these warn IDs show up is with users disabling the warning. And if someone is disabling this specific warning then it probably means they have paths that don't resolve to an array. But this change now forces them to fix that since it's now an SDE, so also forcing them to remove references to that warning seems reasonable.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: > One thing to consider is this doesn't catch expressions like this:
> 
> ```xq
> fn:count(if (some_condition) then /some/path else /some/other/path)
> ```
> 
> I think that's legal and I think would be considered invalid because of this case.
> 
> I'm not sure if there's a good way to check that all branches of a subexpression result in an array. We must be able to get some basic information since we can statically check that all branches have the same type. I think we do that via `inherentType`, but I'm not sure the result would return information about being an array or not. Maybe this is just a limitation of our checking and it's good enough? And if someone whats the above, they would need to do something like this:
> 
> ```xq
> if (some_condition) then fn:count(/some/path) else fn:count(/some/other/path)
> ```

Pretty sure the expression type system treats "array or optional" as a type, and so the type inference on the expressions requires all branches of an IF to evaluate to an array/optional in order to satisfy the requirements of fn:count. It's supposed to propagate the requirement from the fn:count into the expression, which would then propagate it to the then and else branch expressions as the "target type" that the expression is trying to create.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DState.scala
```diff
@@ -215,45 +218,26 @@ case class DState(
...
-  def finalArrayLength: Long =
-    if (isAnArray()) {
-      currentArray.requireFinal
-      currentArray.length
-    } else 1L
+    currentArray.length
+
+  def finalArrayLength: Long = {
+    currentArray.requireFinal()
+    currentArray.length
```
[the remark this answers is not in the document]
mbeckerle: Um. That change to standardize this arrayOpt thing should probably be a separate PR/change set, since otherwise that renaming might clutter this PR.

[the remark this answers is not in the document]
mbeckerle: I can help yes. Send me a meeting request for an afternoon Teams meeting Wed, Thu, or Fri this week.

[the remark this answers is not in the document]
mbeckerle: Note: since I helped with writing this fix, I am recusing myself from review of it.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +744,41 @@ abstract class PathExpression() extends Expression {
...
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
+    steps.last.pred.isEmpty && // last cannot have a [N] pred
+    steps.dropRight(1).forall {
+      // for all the prior steps
+      // if they mention an array element, there
+      // must be a [N] predicate.
+      step =>
+        if (step.isInstanceOf[Up]) true
+        else if (step.isArray) step.pred.isDefined // no special case to check for optional
```
stevedlawrence, line 779: Optional elements do not need a predicate? I thought DPath treated them like arrays so they must have a `[1]` predicate?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +744,41 @@ abstract class PathExpression() extends Expression {
...
+  }
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
+    steps.last.pred.isEmpty && // last cannot have a [N] pred
+    steps.dropRight(1).forall {
+      // for all the prior steps
+      // if they mention an array element, there
+      // must be a [N] predicate.
+      step =>
+        if (step.isInstanceOf[Up]) true
```
stevedlawrence, line 778: What about `Up2`, `Self`, and `Self2`? Do we need exceptions for those other kinds steps?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -969,16 +989,27 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-    val (arrays, scalars) = stepElements.partition { _.isArray }
-    (arrays.length, scalars.length) match {
+    checkAmbiguousPath
+    stepElements.exists(_.isArray)
+  }
+
+  final lazy val isOptional: Boolean = {
+    checkAmbiguousPath
+    stepElements.exists(_.isOptional)
+  }
```
stevedlawrence, line 999: These feel potentially confusing. I feel like our definition is usually an element is either an array OR an optional, but never both. But with step expressions, because of ambiguities, it's possible for isArray and isOptional to both be true. That seems reasonable considering the way dpath works and supports this ambiguity, but it might be worth adding a comment.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
stevedlawrence: I don't think my original comment is addressed? Or are you saying there is a check somewhere else within the type system that makes sure things work as expected? Are there tests for this?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -5150,6 +5147,51 @@
...
+    </xs:element>
+
+    <xs:element name="count11">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="int" type="xs:int" dfdl:representation="binary" />
+          <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:occursCountKind="expression" dfdl:occursCount="{ fn:count(../ex:int) }" />
+        </xs:sequence>
+      </xs:complexType>
+    </xs:element>
```
stevedlawrence, line 5193: Can you add a test that has a conditional inside the fn:count expression? Maybe some with good or bad ambiguities and mixing arrays and optionals? e.g.
```xquery
fn:count(if (../foo) then ../someArray else ../someOptional)
```

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +744,41 @@ abstract class PathExpression() extends Expression {
...
-          if (step.isInstanceOf[Up]) true
-          else if (step.isArray) step.pred.isDefined
-          else true
-      }
+    else true
+  }
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
+    steps.last.pred.isEmpty && // last cannot have a [N] pred
+    steps.dropRight(1).forall {
```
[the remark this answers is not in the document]
tuxji: Method Definition: def init: List[A]
Return Type: It returns all the elements of the list except the last one.

// For all the initial steps, if they mention an array element, there must be a [N] element.

I agree with Steve, let's call steps.init and rename this function to initStepsHaveIndexingPredicates.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +744,41 @@ abstract class PathExpression() extends Expression {
...
-        step =>
-          if (step.isInstanceOf[Up]) true
-          else if (step.isArray) step.pred.isDefined
-          else true
-      }
+    else true
+  }
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
+    steps.last.pred.isEmpty && // last cannot have a [N] pred
```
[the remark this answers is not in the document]
tuxji: Mike, do you concur?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
tuxji: Mike, please weigh in.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
tuxji: Mike, please weigh in.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
tuxji: Agreed.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
tuxji: Best to add a test to check so we can know for sure.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
tuxji: Makes sense.

tuxji (changes requested):
Good comments by Steve, let's update the PR one more time.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: Well, we don't implement "query style" expressions,....

Technically a path like foo/barArray/baz/quuxArray/value returns every value for every index of quuxArray for every index of barArray. I.e., it is like an implied array. That's what it means if treated like an XPath anyway. 

But the point of the DFDL expression language was *not* to enable querying nor to be a clone of XPath, but to express formats, and I know of no need for these sorts of queries in expressing data formats, so that's why we left it out. There is some confusion of whether the DFDL spec even allows this at all. I can't think of a context where it makes sense in DFDL. 

I'm ok with requiring [N] for fixed N for every array path step (except the last), for any path at all. Keeping in mind that ".." never moves you up to a whole array, but to an instance within an array. Hence "..[N]" is never needed (nor allowed).

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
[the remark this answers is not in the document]
mbeckerle: -1 on this change to use "init" terminology.

I have never heard of "init" function before. 

In English, 'initial' means first, not 'all but last'. So I claim 'init' is an extremely really poor choice of name for this concept and not a mistake I would choose to repeat in our terminology.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: Yes. I think those (Up2, Self, Self2) are missing. Neither up steps ".." nor self steps "." should have indexing on them. 

The DFDL spec doesn't rule out "..[1]" nor ".[1]" except to say they are equivalent to the same expression without the [1] because ".." and "." always refer to a single infoset element, not to a whole array of them. 

Tolerating ".[1]" isn't doing anybody any favors however.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: I am not sure we should allow this kind of polymorphism. We don't allow array and scalar polymorphism, nor optional and scalar polymorphism. I think allowing array and optional polymorphism is unwise in this context. 

If you have a path. There should be only one way it needs to be compiled - the path steps should be exactly the same on all paths or it should be an error. That's certainly the way to avoid what is otherwise a quite subtle situation that I would claim, is likely a mistake.  Arrays and optionals have different downward steps (DownArrayOccurrence vs. DownElement).

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: Looking at the code, [exp] is supposed to be tolerated so long as exp evaluates to 1, even on scalars. 
There is a TODO in the code saying "implement this".

Blech. 

This was part of the misguided (I now think) approach the DFDL work group used where there was this notion that DFDL expressions could be implemented by re-purposing an XPath interpreter. When you have no schema, then you have no idea what's an array, what's optional, what's scalar. Every XPath expression is potentially a query that returns a node list. Every element is potentially an array or optional. 

When you have a schema compiler it is decidedly NON helpful to allow expressions which are almost always mistakes when you have the information from the schema which tells you that you are indexing something where that's not meaningful. 

So the DFDL spec says an implementation can issue a warning about these things. I would go further and say that these things should have 3 states: allow, warning, error. But I would default these to error, and make people express that they want warning or allow. 

Frankly, i don't want to bother implementing warning nor allow unless someone complains about it.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
pkatlic: This is fixed.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: The RelativePathExpression and RootPathExpression cases are now combined as PathExpression.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: There are two new tests using subexpressions containing optional and scalar elements.

mbeckerle: An if-then-else propagates the type needed by the surrounding context of the if-then-else and repeats it for the 'then' and 'else' subexpressions:
```  
override def targetTypeForSubexpression(subexp: Expression) = {
    Assert.invariant(children.contains(subexp))
    if (subexp == predicate)
      NodeInfo.Boolean
    else
      this.targetType
  }
```
But I looked at every occurrence of "count(" in the code base and none of them contain an if-then-else, so we need to add a test.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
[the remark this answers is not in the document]
pkatlic: Given the discussion we left this as named containing "prior".

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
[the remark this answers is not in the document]
pkatlic: We check that the last step has an empty predicate and throw SDEs for Up2, Self, and Self2 elements.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: Warnings for optional elements having predicates are not addressed by this branch.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: This check is updated to throw SDEs for Up2, Self, and Self2 steps.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: The isOptional function is now renamed to isArrayorOptional to be clearer about its purpose.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
pkatlic: There are two new tests with conditionals inside a fn:count expression for both optional and scalar elements.

mbeckerle: Yes, we should have 3 tests that combine each pair from the choices of scalar, optional, and array. 

A group with an upward path that escapes the group like 
```
<group name="g">
   <sequence>
      <element name="e" type="xs:int" dfdl:inputValueCalc="{ ../foo }"/>
   </sequence>
</group>
```
then use that group in situations where foo is scalar in one context, optional in the other, optional in one, array in the other, and scalar in one, array in the other. Would catch the analogous problem for the ambiguous upward path situation. Like
```
    ... 
    <element name="scalarOptional">
        <complexType><sequence>
                <element name="scalar">
                     <complexType><sequence>
                          <element name="foo" type="xs:int"/><!-- scalar -->
                          <group ref="g"/>
                     </sequence></complexType></element>
                <element name="optional">
                     <complexType><sequence>
                          <element name="foo" type="xs:int" minOccurs="0"/><!-- optional -->
                          <group ref="g"/>
                     </sequence></complexType></element>
        </complexType></sequence>
   </element>
```
and similar for scalarArray and optionalArray.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: To address @stevedlawrence original question, optionals do not need indexing. A path step that is the name of an optional element generates DownElement just like for a scalar. This is part of the clarification of what does isArray mean which is now distinct from isArrayOrOptional. This piece of code calls isArray and generates array downward steps, the else clause generates DownElement, and this use of isArray is the strong isArray which excludes optionals. 
The DownElement RecipeOp has this comment:
```
/**
 * Down to a non-array element. Can be **optional or scalar**.
 */
case class DownElement(nqn: NamedQName) extends RecipeOp { 
...
```

tuxji (commented):
We are getting very close to approving this PR!  Just one last thing to debug to make sure the logic is correct.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +753,45 @@ abstract class PathExpression() extends Expression {
...
+    steps.dropRight(1).forall {
+      // for all the prior steps
+      // if they mention an array element, there
+      // must be a [N] predicate.
+      // neither Up (..) nor Self (.) steps should have indexing
+      step =>
+        if (step.isInstanceOf[Up]) true
+        else if (step.isInstanceOf[Up2]) SDE("Up2 step must not have indexing")
+        else if (step.isInstanceOf[Self]) SDE("Self step must not have indexing")
+        else if (step.isInstanceOf[Self2]) SDE("Self2 step must not have indexing")
```
stevedlawrence, line 791: It might be cleaner to write this as a match/case, we can also combine Up/Up2 with the parent UpStepExpression and similarly for Self/Self2. Maybe something like this:

```scala
step match {
  case _: UpStepExpression => SDE("up step must not have indexing")
  case _: SelfStepExpression => SDE("self step must not have indexing")
  case _ => !step.isArray || step.pred.isDefined
}
```

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2329,6 +2342,11 @@ abstract class FunctionCallBase(
   protected def checkArgArrayOrOptional(): Unit = {
     lazy val isArrayOrOptional = expressions.head match {
       case pe: PathExpression => pe.isPathToOneWholeArrayOrOptional
+      case ie: IfExpression => {
+        val tp = ie.thenPart
+        val ep = ie.elsePart
+        tp.isPathToOneWholeArrayOrOptional || ep.isPathToOneWholeArrayOrOptional
```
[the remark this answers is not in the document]
stevedlawrence: Agreed, I would expect an `&&` here too.

Another thing maybe worth considering is that this doesn't allow nested ifs, since it only checks one level deep. So if the `then` or `else` branches contain another if-statement, even if those resolve to a path to a whole array/optional, it will fail.

Though, I'm not sure if that matters, and in fact, I'm wondering if we should just require that `fn:count()` accepts exactly a path--no conditionals, no other functions, etc. So we don't allow `fn:count(if(cond) then /path1 else /path2)` and that fails with an SDE like "fn:count argument must be a path". If you want conditional counts then you do it like this: `if (cond) then fn:count(/path1) else fn:count(/path2)`. It's a bit more verbose, but is more clear and avoids potential edge cases.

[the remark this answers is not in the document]
tuxji: I like the simplicity of allowing only a path as fn:count's argument.  I checked the DFDL specification and it was not very clear whether we must allow any DFDL expression to be given to fn:count as argument:

| Function | Meaning |
|--------|--------|
| fn:count($arg) | Returns the number of items in the value of $arg as an xs:integer. Returns 0 if $arg is the empty sequence. |

The following are Schema Definition Errors, regardless of whether they are detected in advance of processing or once processing begins:

  -   Expression value is not single node
    -  Most DFDL expression contexts require an expression to identify a single node, not an array (aka sequence of nodes). There are a few exceptions such as the fn:count(…) function, where the path expression must be to an array or optional element.
  -   Expression value is not array element or optional element.
    -  Some DFDL expression contexts require an array or an optional element.
    -  Example: The fn:count(...) function argument must be to an array or optional element. It is a Schema Definition Error if the argument expression is otherwise.

Mike, please advise us with your deeper knowledge of DPath and XPath.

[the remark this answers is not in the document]
mbeckerle: Ok. I think the DFDL spec is not clear on whether fn:count must take a path (not a DPath expression, a literally step/step/step thingy) or if it can be if/then/else so long as the then/else are both arrays or both optionals. 

I think this code doesn't work right for nested ifs. To get that to work this would have to compose like the way the type system composes. I.e., the 
```
tp.isPathToOneWholeArrayOrOptional || ep.isPathToOneWholeArrayOrOptional
```
(which should be && not ||) that would need to inductively be computed for the then-part and else part. 
Putting in that infrastructure, and then testing it sufficiently feels very not worth it. 

I think I'd rather keep it simpler and just say fn:count's argument must literally be a path (i.e., relative or absolute path steps). It's a fair interpretation of the DFDL spec to put this restriction in.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -5212,6 +5212,80 @@
...
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="optional">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary"/>
+                <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:foo) }"/>
+                <!-- throws ClassCastException -->
```
pkatlic, line 5280: When referencing a group here a ClassCastException is thrown, which may be a bug.

class org.apache.daffodil.runtime1.infoset.DISimple cannot be cast to class org.apache.daffodil.runtime1.infoset.DIArray

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/walker/AbstractDSOMWalker.scala
```diff
@@ -35,6 +35,7 @@ trait CommonContextView {
 trait TermView extends CommonContextView {
   def isArray: Boolean
   def isOptional: Boolean
+  final def isArrayOrOptional: Boolean = isArray || isOptional
```
tuxji, line 38: Ask your IDEA if there are any uses of` this `isArrayOrOptional` function.  Code coverage says this function is not covered by tests, which implies that this function is never used anywhere else (if so, we should remove it).

tuxji (approved):
+1

If the tests are checking the right outcomes, then the code is working correctly.  Good job!

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/walker/AbstractDSOMWalker.scala
pkatlic: This was added as a potentially-useful helper in the future, but can be removed as it is currently not used.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +753,38 @@ abstract class PathExpression() extends Expression {
...
+    else true
+  }
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
+    steps.last.pred.isEmpty && // last cannot have a [N] pred
+    steps.dropRight(1).forall {
+      case _: Up => true
+      case _: Up2 => SDE("Up2 step must not have indexing")
+      case _: SelfStepExpression => SDE("Self step must not have indexing")
+      case step => !step.isArray || step.pred.isDefined
```
stevedlawrence, line 786: I would expect `Up` and `Up2` to have the exact same logic. They are basically the same thing, just one asserts the parent QName must be a specific name. I'm not sure why one is `true` and the other is an SDE? I would expect these could be combined into a single `case _: UpStepExpression`.

Also, "Up2" in the error message is an internal identifier that probably won't make sense to users. Maybe just call it an "up step".

Though, thinking about this more, I don't think I understand these Up2 and SelfStep case. Say we have an expression like `fn:count(../parent::foo/bar/baz)`. I believe this function will throw an SDE here because of the `parent:foo` up-step? But this path is perfectly fine.

Based on the error message, I think this case trying to check against something like `fn:count(../..[2]/some/other/element)', where the `..[2]` isn't legal? In fact, a related issue is mentioned in a comment in `SelfStepExpression` and references DAFFODIL-2182. So we probably aren't dealing with predicates with up- and self-steps correctly, but I'm not sure this is the right function, or PR, to fix it. The goal of this function just wants to make sure any steps to arrays have a predicate, which I think the last case correctly checks.

Up- and self-steps can technically go to arrays (i.e. isArray can be true for them) but they always resolve to a single instance in the array, so they don't actually need predicates, so instead of SDE'ing, I think they just want to be `true`, like the Up-step case. So I think those other cases just want to be something like:

```scala
// up and self expressions do not need a predicate because they always reference a single element, even if that element is part of an array
case _: UpStepExpression | _: SelfStepExpression => true
// either not an array, or is an array with a predicate
case step => !step.isArray || step.pred.isDefined
```

Though, we already have that check here:

https://github.com/apache/daffodil/blob/main/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala#L1223

I wonder if the issue is that the `targetType` for the `fn:count` is `NodeInfo.Exists`, which means that part of code that normally requires predicates isn't doing it? I think that is a bug? Maybe that wants to be changed to something like:

```scala
} else if (!isLast && targetType == NodeInfo.Exists) {
  new DownArrayExists(nqn)
} else {
```

This way, even with fn:exists or fn:count, we still require a predicate for all intermediate named steps. The only one that doesn't require a predicate is the last step. This means you can no longer do something like this:

```
fn:exists(/path/arrayOne/foo/bar/arrayTwo)
```
Where arrayOne is an array without a predicate. You could do that in the past, but that doesnt make any sense. Only the arrayTwo step can not have a predicate. I'm not even sure what would currently happen in that case, since we require a specific instance of arrayOne to look for the foo child.

Sorry for the long wall of text, but in sumary, I think we just need the `!isLast` change, and then priorStepsHavingIndexingPredicates goes away, since that is now enforced by the isLast change.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2295,6 +2332,20 @@ abstract class FunctionCallBase(
     res
   }
 
+  protected def checkArgArrayOrOptional(): Unit = {
```
stevedlawrence, line 2335: If this function is going to be in FunctionCallBase, then it would be useful if it could be generic enough to work for any possible function. Right now, it only looks at expression.head, so it's only useful for functions that require array/optional as the first function parameter. I'd suggest changing it to allow it to pass in an index that can be used to pick which expression it checks. The fn:count function would call it like `checkArgArrayOrOption(0)`.

Alternatively, move this logic into FunctionCallArrayBase, since it's likely the only thing that need will need this kindof check. It already has `arrPath.isPathToOneWholeArray`, maybe that just needs to be updated to require this?

That said, I'm not sure why arrayPath.isPathToOneWholeArray in FunctionCallArrayBase doesn't already result in an SDE. Seems like that's already doing the check we want. I wonder if there's just a bug there and that's not getting checked correctly, and we don't actually need this new checkArgArrayOrOptional check?

stevedlawrence (commented):
I'm trying to understand our existing code more, and if feels like we already have parts of the pieces implemented such that some of the new code isn't actually need, we just have bugs where the existing logic isn't getting hit? I wonder if we just need to fix those bugs? I possibly identified two.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
stevedlawrence: Note that I think I have the name and negation wrong above. It should be
```scala
} else if (isLastStep && targetType == NodeInfo.Exists) {
  new DownArrayExists(nqn)
} else {
```

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
stevedlawrence: I took another scan at this, and it looks like `isTypeCorrect` (which already calls isPathToOneWholeArray) is never actually used. I suggest we just delete this val and the couple of references to it, and then move your `checkArgArrayOrOptional` function to `FunctionCallArrayBase`. It can then use arrPath (instead of expression.head) to verify that is goes to a whole array. And then update implementations of FunctionCallArrayBase to call this function, since they should all go to a whole array.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -744,21 +742,38 @@ abstract class PathExpression() extends Expression {
...
+    priorStepsHaveIndexingPredicates
+  }
+
+  private lazy val isNotRootOrSelf: Boolean = {
+    if (steps == Nil) false
+    else if (isSelf) false
+    else true
+  }
+
+  private lazy val priorStepsHaveIndexingPredicates: Boolean = {
```
stevedlawrence, line 769: Can this val be removed? With the `isLastStep` addition in `NamedStep`, I think we should now SDE if any prior steps are arrays without an index predicate, so now the `steps.dropRight(1).forAll` check should always evaluate to true an isn't really needed. I think we do still need the `steps.last.pred.isEmpty` check but that probably makes more sense as an added check in `isPathToOneWholeArrayOrOptional` and `isPathToOneWholeArray`.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -234,6 +223,15 @@ abstract class Expression extends OOLAGHostImpl() with BasicComponent {
...
   }
+
+  /**
+   * Path to a single array element without a [N] predicate qualifying it
+   * or to an optional element.
+   *
+   * That is, the kind of expression used to access information
+   * about an array or optional such as its current count.
+   */
+  lazy val isPathToOneWholeArrayOrOptional: Boolean = false
```
stevedlawrence, line 234: Can we remove this from `Expression` so it is only a member of `PathExpression`? Seems odd to ever ask a generic `Expression` if it is a special kind of path. Note that `isPathToOneWholeArray` is only a member of PathExpression, so removing this would make the two functions consistent.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2378,7 +2414,7 @@ case class FNCountExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expre
 }
 
 case class FNExactlyOneExpr(nameAsParsed: String, fnQName: RefQName, args: List[Expression])
-  extends FunctionCallArrayBase(nameAsParsed, fnQName, args) {
+  extends FunctionCallArrayOrOptionalBase(nameAsParsed, fnQName, args) {
```
stevedlawrence, line 2418: Can we update this compiledDPath in this class to also call `checkArgArrayOrOptional(0)`? We don't currently supported it, but if we do add support it would be good to already have this check.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
stevedlawrence: Does this ClassCastException still occur, or did newer changes fix this?

stevedlawrence (commented):
Looks good, one question abou the class cast exception and minor cleanup

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
pkatlic: This exception still occurs for this case with an array and optional, as it seems that the optional is evaluated as a DownArray rather than a DownElement.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: Yes, moved this to PathExpression.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: This priorStepsHaveIndexingPredicates expression is required in order to properly process array elements. With this removed, we saw several tests fail.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: This call to checkArgArrayOrOptional(0) now made by FNExactlyOneExpr.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
stevedlawrence: Did you add `steps.last.pred.isEmpty` to `isPathToOneWholeArrayOrOptional` and `isPathToOneWholeArray`?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -5150,6 +5147,145 @@
...
+                <xs:group ref="countGroup"/>
+              </xs:sequence>
+            </xs:complexType>
+          </xs:element>
+          <xs:element name="optional">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="foo" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary"/>
+                <!-- throws ClassCastException, possible bug in Daffodil group evaluation -->
+                <!-- <xs:group ref="countGroup"/> -->
```
stevedlawrence, line 5280: This definitely sounds like a bug. Can you open a ticket so we remember to look into it? If possible, include steps to reproduce (e.g. uncomment this group reference, run XYZ test) and the exception?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
pkatlic: Good catch!  After this update, the priorStepsHaveIndexingPredicates expression is no longer needed.

tuxji (approved):
This PR looks ready to be approved once the new bug is created in the JIRA database.

stevedlawrence (approved):
+1, agreed, looks good to me. Just needs to be squashed to a single commit

tuxji:
Peter has created DAFFODIL-2879 as requested. PR is approved and ready for merge, will merge it.

PR #1139, opened 2024-01-11
Use the negative pattern of textNumberPattern to resolve padding ambiguities

stevedlawrence:
> [!NOTE]
> This is an alternative approach to #1138, which does feel like a bit of a hack and does not support other padding styles. Like that one, I've marked this as an RFC/Draft since it uses behavior not specified by the DFDL specification and gives meaning to something ICU normally ignores.
>
> In the previous PR, there was some discussion about making the negative pattern mean something more. Although DAFFODIL-2870 suggests we shouldn't do that, I think this use is small enough that it is potentially acceptable. It also doesn't require changes to Daffodil parsing/unparsing (with possible performance implications) or extension properties mentioned in the other PR (e.g. `dfdlx:textNumberPositivePatttern`/`dfdlx:textNumberNegativePattern`/`dfdlx:icuPadPosition`). We definitely do need to be careful about what, if any, meaning we extract from normally ignored parts of the negative pattern, but the pad position seems like a reasonable exception to me considering possible ambiguities with the positive pattern.  

ICU only uses the positive pattern of the textNumberPattern property to determine pad character and position, completely ignoring the negative pattern. If the positive pattern has no affix associated with the pad character, then there is an ambiguity if the pad character should appear before or after the negative affix when unparsing negative numbers. In this case, ICU defaults to before the affix, with no way to change it via the pattern. This effectively means it is not possible for ICU number padding to appear after a negative affix if there is no positive affix.

To resolve this ambiguity and allow configuring where pad characters appear, we inspect the negative pattern. If both negative and positive patterns define padding on the same affix, and the positive pattern has an empty string for that affix, then we use the pad position from the negative pattern. In all other cases, the pad character in the negative pattern is ignored following usual ICU behavior.

For example, a textNumberPattern of "*0####0;-*00" formats a negative number with zero padding after the hyphen, whereas normal ICU behavior would ignore the negative pattern and zero pad before the hyphen.

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesTextNumber.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesTextNumber.scala
@@ -28,6 +28,7 @@ import org.apache.daffodil.lib.schema.annotation.props.gen.TextNumberRounding
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.Maybe._
 import org.apache.daffodil.lib.util.MaybeDouble
+import org.apache.daffodil.lib.util.MaybeInt
 import org.apache.daffodil.runtime1.dpath.NodeInfo.PrimType
 import org.apache.daffodil.runtime1.processors.Delimiter
 import org.apache.daffodil.runtime1.processors.TextNumberFormatEv
@@ -38,6 +39,10 @@ import org.apache.daffodil.runtime1.processors.unparsers.Unparser
 import org.apache.daffodil.unparsers.runtime1.ConvertTextCombinatorUnparser
 import org.apache.daffodil.unparsers.runtime1.ConvertTextNumberUnparser
 
+import com.ibm.icu.impl.number.AffixPatternProvider
+import com.ibm.icu.impl.number.Padder.PadPosition
+import com.ibm.icu.impl.number.PatternStringParser
+import com.ibm.icu.impl.number.PatternStringParser.ParsedPatternInfo
 import com.ibm.icu.text.DecimalFormat
 
 case class ConvertTextCombinator(e: ElementBase, value: Gram, converter: Gram)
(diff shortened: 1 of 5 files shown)


tuxji (commented):
Yes, I like this proposed approach better than the previously discussed ideas.  I especially like that people can keep using their same textNumberPattern property and only need to add a character (was it *?) to the negative half of the pattern to solve their problem with the way ICU interprets the pattern.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesTextNumber.scala
```diff
@@ -586,6 +597,29 @@ case class ConvertTextStandardNumberPrim(e: ElementBase)
         Nope
       }
 
+    // ICU does not have a way to set the pad position to after an affix if the positive pattern
+    // does not have that affix. For example, "* 0", will always have a pad position of
```
mbeckerle, line 601: Needs release note. This is a backward incompatible change (very minor one) because a "*0" in the negative pattern would have been always ignored before, but now means something that changes the behavior.

The warnings suggested in [DAFFODIL-2870](https://issues.apache.org/jira/browse/DAFFODIL-2870) are still meaningful, they just need to allow for the cases where the pad specifier in the negative pattern is used rather than ignored.

mbeckerle (approved):
+1

This needs release note.

This also needs documentation as a Daffodil extension to DFDL.

This should be presented to the DFDL workgroup as a needed fix to the DFDL spec. I will start a discussion thread there about this.

stevedlawrence:
With no objections, I've removed "draft" from this, created an issue in Jira, and updated the commit message to reference the issue and mention the backwards incompatibility change.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesTextNumber.scala
tuxji: Where do we capture and record a phrase or sentence which needs to be added to the release notes so we don't forget to include it later?

tuxji (approved):
+1

PR #1149, opened 2024-01-18
Fix and tests added DAFFODIL-2873 zoned ebcdic bug

mbeckerle:
DAFFODIL-2873

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
@@ -27,13 +27,13 @@ import org.apache.daffodil.core.grammar.primitives.ConvertTextBooleanPrim
 import org.apache.daffodil.core.grammar.primitives.LeadingSkipRegion
 import org.apache.daffodil.core.grammar.primitives.LiteralValueNilOfSpecifiedLength
 import org.apache.daffodil.core.grammar.primitives.SimpleNilOrValue
-import org.apache.daffodil.core.grammar.primitives._ // there are too many to show individually
+import org.apache.daffodil.core.grammar.primitives._
 import org.apache.daffodil.core.runtime1.ElementBaseRuntime1Mixin
 import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.schema.annotation.props.Found
 import org.apache.daffodil.lib.schema.annotation.props.NotFound
-import org.apache.daffodil.lib.schema.annotation.props.gen._ // there are too many to show individually
+import org.apache.daffodil.lib.schema.annotation.props.gen._
 import org.apache.daffodil.lib.util.PackedSignCodes
 import org.apache.daffodil.lib.xml.GlobalQName
 import org.apache.daffodil.lib.xml.XMLUtils
@@ -699,6 +699,22 @@ trait ElementBaseGrammarMixin
     prod("textZonedNumber", textNumberRep == TextNumberRep.Zoned) {
       ConvertZonedCombinator(this, stringValue, textZonedConverter)
     }
+
+  /**
+   * True if the encoding is known to be an EBCDIC one, as in the encoding is not  a runtime expression
+   * and it's some ebcdic flavor. If it's any ascii flavor or a runtime expression this is false.
+   */
+  lazy val isKnownEBCDICEncoding: Boolean =
+    charsetEv.optConstant.map { _.isEbcdicFamily() }.getOrElse(false)
+
+  /**
+   * Avoids requesting the textZonedSignStyle property if we know the encoding
+   * is an EBCDIC flavor.
+   */
+  lazy val optTextZonedSignStyle = {
+    if (isKnownEBCDICEncoding) None
+    else Some(textZonedSignStyle)
+  }
 
   private lazy val textConverter = {
     primType match {
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
@@ -386,6 +386,45 @@ object DecimalUtils {
     outArray
   }
 
+  /**
+   * These are the java characters corresponding to EBCDIC B0 to B9.
+   * These are commonly "^£¥·©§¶¼½¾" but we compute them by decoding to
+   * avoid literal charset issues.
+   *
+   * Note that in Java chars (which are unicode), these do not have adjacent code points.
+   * They are (hex) 5e, a3, a5, b7, a9, a7, b6, bc, bd, be.
+   */
+  private val B0_to_B9_chars = "^£¥·©§¶¼½¾"
+
+  /**
+   * Despite the name, this does not actually convert the digit from EBCDIC encoding.
+   * The characters have already been decoded from whatever charset into Unicode code points.
+   * What this does is deal with overpunched positive and negative sign character mappings.
+   * In this case positive are 0-9 and "{ABCDEFGHI", negative are "}JKLMNOPQR" and
+   * what we call the B0_to_B9 characters, which are "^£¥·©§¶¼½¾".
+   * @param digit
+   * @return A pair of the integer the digit represents, and a boolean indicating if negative.
+   */
+  def convertFromZonedEBCDIC(digit: Char): (Int, Boolean) = {
+    if ((digit >= '0') && (digit <= '9')) // positive 0-9
+      (digit - 48, false)
+    else if (digit == '{') // positive 0 aka hex C0 ebcdic
+      (0, false)
+    else if ((digit >= 'A') && (digit <= 'I')) // positive 1-9 as C1 to C9 i.e, "ABCDEFGHI"
+      (digit - 'A' + 1, false)
+    else if (digit == '}') // negative 0 aka hex D0 ebcdic
+      (0, true)
+    else if ((digit >= 'J') && (digit <= 'R')) // negative 1-9 as hex D1 to D9.
+      (digit - 'J' + 1, true)
+    else {
+      val index = B0_to_B9_chars.indexOf(digit)
+      if (index >= 0)
+        (index, true)
+      else
+        throw new NumberFormatException("Invalid zoned digit: " + digit)
+    }
+  }
+
   def convertFromAsciiStandard(digit: Char): (Int, Boolean) = {
     if ((digit >= '0') && (digit <= '9')) // positive 0-9
       (digit - 48, false)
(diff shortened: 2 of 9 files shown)


daffodil-test/src/test/resources/org/apache/daffodil/section13/zoned/zoned2.tdml
```diff
@@ -0,0 +1,96 @@
...
+        <maxInclusive value="9999999.99"/>
+        <minInclusive value="-9999999.99"/>
+      </restriction>
+    </simpleType>
+
+    <element name="money7" type="ex:displayMoney7Digits"/>
+
+  </tdml:defineSchema>
+
+  <tdml:parserTestCase name="ZonedEBCDICLeadingSign" root="money7" model="ebcdic1"
```
mbeckerle, line 72: tdml: prefixes can all be removed.

daffodil-test/src/test/resources/org/apache/daffodil/section13/zoned/zoned2.tdml
```diff
@@ -0,0 +1,96 @@
...
+    </simpleType>
+
+    <element name="money7" type="ex:displayMoney7Digits"/>
+
+  </tdml:defineSchema>
+
+  <parserTestCase name="ZonedEBCDICLeadingSign" root="money7" model="ebcdic1"
+    description="Test of a negative number with ebcdic charset, leading overpunched sign of R should be -9">
+    <document>
+      <documentPart type="text" encoding="ebcdic-cp-us">R99999999</documentPart>
```
stevedlawrence, line 75: It's there any value in defining the document part in bytes instead of text with ebcdic-cp-us encoding? That would show that the issue doesn't have anything to do with ebcdic encoding in the documentPartt? I feel like that isn't used very often so there's a chance the bug is in that?

stevedlawrence (approved):
+1 needs  `sbt scalafmtAll` to quiet the linter.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/RuntimePropertyMixins.scala
```diff
@@ -123,7 +123,7 @@ trait TermRuntimeValuedPropertiesMixin
     ev
   }
 
-  final lazy val charsetEv = {
+  final lazy val charsetEv: CharsetEv = {
```
mbeckerle, line 126: Revert

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -700,6 +701,30 @@ trait ElementBaseGrammarMixin
...
+  /**
+   * True if the encoding is known to be an EBCDIC one, as in the encoding is not  a runtime expression
+   * and it's some ebcdic flavor. If it's any ascii flavor or a runtime expression this is false.
+   */
+  lazy val isKnownEBCDICEncoding: Boolean = {
+    charsetEv.optConstant
+      .map { bcs: BitsCharset =>
+        val nom = bcs.name.toUpperCase()
+        val res = (nom == "IBM037") ||
+          nom.startsWith("EBCDIC")
```
mbeckerle, line 713: This may not be sufficient to recognize EBCDIC. There are numrous IBM code pages that are EBCDIC besides IBM037. There are ones for the international variants of EBCDIC. All of them do the exact same thing for zoned numbers. 

We need something that will work regardless of the aliases vs. the "official" name. If the name, or any alias contains "ebcdic" (case insensitive) then it is an ebcdic variant.

stevedlawrence: Should we add an `isEBCDIC` member to `BitsCharset` that defaults to `false`, and any EBCDIC encodings we implement would override that to `true`? Then this just becomes

```scala
charsetEv.optConstant.map { _.isEBCDIC }.getOrElse(false)
```

Also, doing some grepping for EBCDIC I see we also have a `Misc.isAsciiBased` function. That's only used for dumping debug data and works on Java charsets instead of our BitsCharset, but maybe it provides extra motivation that we might eventually want to have other `is*` members so we can ask questions about our supported charsets without having to make guesses based on the name.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -700,6 +701,30 @@ trait ElementBaseGrammarMixin
...
+  }
+
+  /**
+   * Avoids requesting the textZonedSignStyle property if we know the encoding
+   * is an EBCDIC flavor.
+   */
+  lazy val optTextZonedSignStyle = {
+    if (isKnownEBCDICEncoding) None
+    else Some(textZonedSignStyle)
+  }
```
stevedlawrence, line 726: This doesn't do the right thing if the charset is non-constant be evaluates to EBCIDC at runtime. Is that something we care about? Does it make sense to convert this to a evaluatable?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
```diff
@@ -386,6 +386,25 @@ object DecimalUtils {
...
+   */
+  def convertFromZonedEBCDIC(digit: Char): (Int, Boolean) = {
+    if ((digit >= '0') && (digit <= '9')) // positive 0-9
+      (digit - 48, false)
+    else if (digit == '}') // negative 0
+      (0, true)
+    else if ((digit >= 'J') && (digit <= 'R')) // negative 1-9
+      (digit - 'J' + 1, true)
+    else
+      throw new NumberFormatException("Invalid zoned digit: " + digit)
```
stevedlawrence, line 405: The DFDL spec says this about `textNumberRep`:
> In EBCDIC-based encodings, code points 0xC0 to 0xC9 or 0xF0 to 0xF9 represent a positive sign and digits 0 to 9 (these byte ranges correspond typically to characters '{ABCDEFGHI'* or '0123456789'), and code points 0xD0 to 0xD9 or 0xB0 to 0xB9 represent a negative sign and digits 0 to 9 (these byte ranges correspond typically to characters '}JKLMNOPQR' or  '^£¥·©§¶¼½¾ ' ). On parsing both ranges are accepted. On unparsing the range 0xC0 to 0xC9 are produced for positive signs and the range 0xD0 to 0xD9 are produced for negative signs.

Should this function also support `{ABCDEFGHI` for positive numbers and `^£¥·©§¶¼½¾ ` for negatives?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
```diff
@@ -417,6 +436,25 @@ object DecimalUtils {
       throw new NumberFormatException("Invalid zoned digit: " + digit)
   }
 
+  /**
+   * Does not encode to the EBCDIC charset, but converts negative digits
+   * to their EBCDIC Zoned overpunched representation where negative 0 to 9
+   * are mapped to "}JKLMNOPQR".
+   * @param digit - the digit, as a char '0' to '9'
+   * @param positive - true if positive, false if negative
+   * @return the character needed to represent this character as an overpunched sign digit.
```
stevedlawrence, line 445: From the spec:

> In EBCDIC-based encodings, code points 0xC0 to 0xC9 or 0xF0 to 0xF9 represent a positive sign and digits 0 to 9 (these byte ranges correspond typically to characters '{ABCDEFGHI' or '0123456789'), and code points 0xD0 to 0xD9 or 0xB0 to 0xB9 represent a negative sign and digits 0 to 9 (these byte ranges correspond typically to characters '}JKLMNOPQR' or  '^£¥·©§¶¼½¾ ' ). On parsing both ranges are accepted. On unparsing the range 0xC0 to 0xC9 are produced for positive signs and the range 0xD0 to 0xD9 are produced for negative signs.

This makes it sound like on unparsing the positive 0-9 digits should be unparsed as `{ABCDFGHI` instead of 0-9?

daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestDecimalUtils.scala
```diff
@@ -1846,733 +1989,1620 @@ class TestDecimalUtils {
...
-      zonedFromNumber("-9", TextZonedSignStyle.AsciiTandemModified, OverpunchLocation.Start),
+      zonedFromNumber(
+        "-9",
+        Some(TextZonedSignStyle.AsciiTandemModified),
+        OverpunchLocation.Start,
+      ),
       "",
     )
   }
+
```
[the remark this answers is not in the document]
stevedlawrence: Same question as above, should the positives be `{` and `A-I` instead of `0-9`? And do we need tests for the other range of positive/negative values?

stevedlawrence (approved):
+1

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
mbeckerle: You are correct. Fixed.

daffodil-lib/src/test/scala/org/apache/daffodil/lib/util/TestDecimalUtils.scala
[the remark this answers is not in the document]
mbeckerle: You are correct. Fixed and tests added.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/DecimalUtils.scala
mbeckerle: Correct. Fixed.

mbeckerle:
I believe this now addresses all the comments by @stevedlawrence.

Needs a second reviewer. @jadams-tresys can you review this? I believe the ascii zoned stuff was first done by you.

jadams-tresys:
+1

Looks good to me.  Reading the Wikipedia page I'm not surprised I didn't get this right on the first try.  This section was a fun read though: https://en.wikipedia.org/wiki/EBCDIC#Criticism_and_humor

jadams-tresys (approved):

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
stevedlawrence: But if the charset evaluates to EBCDIC at runtime we'll still use textZonedSignStyle. So in that case we still need the property at compile time (like you have) but we need to ignore it at runtime.

So maybe the parser/unparsers need to have the charsetEv passed in, and if at runtime if it's non-constant and evaluates to EBCDIC then it ignores `optTextZonedSignStyle`?

mbeckerle: Correct. That is what happens. So this optional thing is just about not reqiring what is an ascii-oriented property at all in the case where it is known for certain to be EBCDIC.

PR #1150, opened 2024-01-24
Update daffodil-release-candidate to support daffodil-sbt

stevedlawrence:
- Build, sign, and publish a source zip to Apache dist and a packaged jar to Apache's maven repo
- Change how we calculate checksums/signature so it does not require a bin/ directory since daffodil-sbt does not have any convenience binaries aside from those published to maven--it only publishes a source to Apache dist

Closes apache/daffodil-sbt#16


mbeckerle (approved):
+1

tuxji (approved):
+1

PR #1158, opened 2024-02-13
Refactor NodeInfo.PrimType - part of layer API work

mbeckerle:
Supports a simple Java enum (JPrimType).

Bunch of cleanups of NodeInfo dead or little-used code.

optPrimType is now available on all TypeNodes and associated Kind Scala types. (eliminates list searching to find primType)

This is part of work towards reinventing the layering API, but it has impact on the metadata interface being used from Drill.

DAFFODIL-2845


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/JPrimType.java
mbeckerle, line 1: I have no idea why it thinks this is a rename of TestNodeInfo.scala

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/JPrimType.java
```diff
@@ -14,16 +14,29 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package org.apache.daffodil.runtime1.api;
 
-package org.apache.daffodil.runtime1.dpath
+public enum JPrimType {
```
mbeckerle, line 19: Javadoc. 

This enum is easy to use from Java code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -85,16 +82,16 @@ sealed abstract class TypeNode private (
 ) extends Serializable
   with NodeInfo.Kind {
 
-  def this(sym: Symbol, parents: => Seq[NodeInfo.Kind], children: => Seq[NodeInfo.Kind]) =
```
mbeckerle, line 88: Did you know type Symbol was removed from Scala in Scala 3. Possibly earlier.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -298,36 +279,36 @@ object NodeInfo extends Enum {
 
   def fromObject(a: Any) = {
     a match {
-      case x: String => NodeInfo.String
```
mbeckerle, line 301: This could be reverted. 

I was thinking of removing NodeInfo's aliases for all the PrimType objects, but it was too much trouble.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/JPrimType.java
mbeckerle, line 1: This file is illustrative of the approach to stable APIs that are usable from Java.

The idea is that enums needed to shine through the API are defined in Java so that they're natural to use from Java. Scala enums then embed them (java enums are classes, but they're always final) when they cannot be used directly.

I'm open to suggestions on naming here. I chose "J" prefix but "E" for enum might be better.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -198,58 +199,39 @@ object NodeInfo extends Enum {
...
+     * The prim type corresponding to this type.
+     * So if this Kind is derived from a PrimType kind, then
+     * this kind should return that prim type object.
+     *
+     * @return None if there is no corresponding prim type.
+     */
+    def optPrimType: Option[PrimType] = None
 
-    final lazy val selfAndAllParents: Set[Kind] = parents.flatMap {
+    private final lazy val lcaseName = name.toLowerCase()
```
stevedlawrence, line 227: Is there value in storing the lowercase version of this as a member? I feel like memory is our big problem, everytime we add a new member we are taking up just a little bit more memory. In fact, it's only used by `parentList`, and that is only used by `doesParentListContain`, which just calls `toLowerCase` again, so we aren't even saving any CPU by storing it. Probably a premature optimization, but seems like `doesParentListContain` could be this:

```scala
    def doesParentListContain(typeName: String): Boolean = {
      selfAndParents.exists { _.name.equalsIngoreCase(typeName) }
    }
```
Avoids both he lcaseName and parentList members, and I imagine is just as fast.

All that said, I now see `doesParentListContain` is never even used, so I suggest we just drop that function and the associated members it uses.

stevedlawrence (approved):
+1, just a number of minor thoughts/suggestions

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Delay.scala
```diff
@@ -65,7 +65,7 @@ import org.apache.daffodil.lib.exceptions.Assert
  *
  * }}}
  */
-final class Delay[T] private (private var box: Delay.Box[T], sym: Symbol)
+final class Delay[T] private (private var box: Delay.Box[T], sym: AnyRef)
```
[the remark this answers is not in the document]
mbeckerle: These symbols were just used for debugging to identify what was going on. 

I wanted to eliminate use of symbols here and instead pass other kinds of context objects. (Java enum objects) 
All it needs to have is a toString method, so AnyRef is a reasonable type for this. 

In general yes getting rid of Symbol should be a separate matter. 

I will add Scaladoc explaining that these are just context objects used for debugging.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/JPrimType.java
```diff
@@ -14,16 +14,29 @@
...
+    Byte,
+    Short,
+    Long,
+    Integer,
+    Decimal,
+    UnsignedInt,
+    UnsignedByte,
+    UnsignedShort,
+    UnsignedLong,
+    NonNegativeInteger,
```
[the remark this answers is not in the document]
mbeckerle: These are a Java Enum representation of DFDL's simple types. 

The fact that there are no unsigned types in Java is a separate issue that is addressed in trait InfosetSimpleElement where there are methods like getUnsignedLong which returns a JBigInt, and getUnsignedShort returns a JInt.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/Metadata.scala
```diff
@@ -155,58 +151,11 @@ trait ComplexElementMetadata extends ElementMetadata {
...
-  val NonNegativeInteger: PrimitiveType = PrimType.NonNegativeInteger
-  val Double: PrimitiveType = PrimType.Double
-  val Float: PrimitiveType = PrimType.Float
-  val HexBinary: PrimitiveType = PrimType.HexBinary
-  val AnyURI: PrimitiveType = PrimType.AnyURI
-  val Boolean: PrimitiveType = PrimType.Boolean
-  val DateTime: PrimitiveType = PrimType.DateTime
-  val Date: PrimitiveType = PrimType.Date
-  val Time: PrimitiveType = PrimType.Time
+  def jPrimType: JPrimType
```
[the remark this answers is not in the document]
mbeckerle: I'd prefer not to rename Daffodil's NodeInfo.PrimType object and trait/type, as it is referenced all over the code base, as are methods like optPrimType and primType. 

So we need to distinguish this enum from that. 

Idea: rename JPrimType to PrimTypeEnum or EPrimType, and the methods correspondingly.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
mbeckerle: I will drop this method and associated.

Note that this graph of NodeInfo TypeNode objects is basically a tiny static data structure. There are fewer than 30 such objects. Everything else just points at these objects or uses their types (called Kinds)  to represent a DFDL type or other static characteristic of data or an argument/result of an expression.

stevedlawrence: Yeah, I don't think memory is really an issue in this case. Just something I try to be aware of. I don't have any numbers, but I feel like we have a lot of things that are calculated once and stored in a val and never used again just taking up memory.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/CompiledExpression.scala
```diff
@@ -251,33 +252,35 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
...
+              )
+            }
+          }
 
-    new ConstantExpression[T](qn, nodeInfoKind, logical.getAnyRef.asInstanceOf[T])
+        new ConstantExpression[T](qn, nodeInfoKind, logical.getAnyRef.asInstanceOf[T])
+      }
+      case _ => {
+        val msg = "No known primitive type to convert logical value to: %s"
+        compileInfoWhereExpressionWasLocated.SDE(msg, nodeInfoKind)
```
mbeckerle, line 282: Change to an assert, and put the coverage suppression around it.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Delay.scala
```diff
@@ -104,7 +104,7 @@ final class Delay[T] private (private var box: Delay.Box[T], sym: Symbol)
       else {
         box
       }
-    "Delay(" + sym + ", " + objString + bodyString + ")"
+    "Delay(" + sym.toString + ", " + objString + bodyString + ")"
```
mbeckerle, line 107: Add coverage suppression. These are only for debugging purposes.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Delay.scala
```diff
@@ -130,23 +130,27 @@ object Delay {
...
+  private class Box[T](val obj: AnyRef, val context: AnyRef, delayedExpression: => T) {
     lazy val value = delayedExpression
 
     override def toString() = {
       val desc = obj match {
         case d: NamedMixinBase => "(" + d.diagnosticDebugName + ")"
         case _ => ""
       }
-      "box(" + obj.hashCode() + desc + ")"
+      "box(" + obj.toString + desc + ")"
```
mbeckerle, line 153: Add coverage suppression. These are only for debugging.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/Metadata.scala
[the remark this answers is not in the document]
mbeckerle: That's worth a try.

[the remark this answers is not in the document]
mbeckerle: Currently I have DFDLPrimTypeEnum as the enum name, and dfdlType as the method that returns one of those.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JDOMInfosetOutputter.scala
```diff
@@ -56,7 +56,7 @@ class JDOMInfosetOutputter extends InfosetOutputter {
 
     if (!simple.isNilled) {
       val text =
-        if (simple.metadata.primitiveType == PrimitiveType.String) {
+        if (simple.metadata.jPrimType == JPrimType.String) {
```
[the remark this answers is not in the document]
mbeckerle: method name is now dfdlType, enum name is DFDLPrimTypeEnum

mike-mcgann (approved):

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLPrimTypeEnum.java
```diff
@@ -14,16 +14,31 @@
...
-
-class TestNodeInfo {
-
-  @Test def test_nodeInfoCanBeConstructed(): Unit = {
-    assertTrue(NodeInfo.isXDerivedFromY("Byte", "Long"))
-  }
+/**
+ * An enumeration of all DFDL's simple types.
+ */
+public enum DFDLPrimTypeEnum {
```
stevedlawrence, line 22: Is it standard to end a Java enum type with `Enum`? I'm not sure I've seen that before, but it's been a while since I've done Java. I don't think we have anything like `DFLDPrimType` that this could be confused with?

stevedlawrence (approved):
+1 to latest changes, I especially like `dfdlType`/`DFDLPrimTypeEnum`--makes it more clear that it is different from java primitives which might help avoid possible confusion. Only question is if `Enum` suffix is necessary, but not strong preference either way.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLPrimTypeEnum.java
stevedlawrence: Also, I think Java convention is to capitalize enum names? Should we do that too?

mbeckerle: I am disinclined to all-caps enum names that come from Java enum definitions. It will make them very ugly to use in Scala code. It's never been a convention I liked. Honestly, I think it's a holdover from C days of early computing when ALL CAPS NAMES were a way of shouting that you did get the point that you shouldn't hard-code magic numbers in your code. But it wasn't necessary even then when we didn't have fancy IDEs that would take you to the definition in one click. 

The suffix "Enum" was intended to make clear that "this is just an enum", but I think I will remove it, as Java enums are classes and can have methods and other richness, which we may want to add at some point.

PR #1159, opened 2024-02-15
Improve performance of xml and json infoset outputters

stevedlawrence:
The XMLTextInfosetOutputter and JSONInfosetOutputter do not use any buffering when writing data. Modifying these to wrap a BufferedWriter around the existing OutputStreamWriter gives significant performance improvements. Using a BufferedWriter also allows us to use its built-in newLine() function for pretty printing.

This also modifies the "Standard" XML escape style in the xml infoset outputter so that it first checks if there are any characters that need to be escaped, similar to what we do for CDATA escape style. In most cases, there will not be any characters that need escaping, so we can avoid Scala XML's escape utility, which has noticeable overhead, even if nothing needs escaping.

With these changes tested on a large file with lots of strings, this saw total parse + infoset output time drop from about 125 seconds to 93 seconds, about a 25% decrease. Note that parsing with the null infoset outputter takes about 78 seconds, so the xml infoset outputter overhead went from about 37% of the total parse time down to about 20%.

DAFFODIL-2872

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetOutputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetOutputter.scala
@@ -29,12 +29,16 @@ import org.apache.daffodil.runtime1.api.InfosetSimpleElement
 
 import com.fasterxml.jackson.core.io.JsonStringEncoder
 
-class JsonInfosetOutputter private (writer: java.io.Writer, pretty: Boolean)
+class JsonInfosetOutputter private (writer: java.io.BufferedWriter, pretty: Boolean)
   extends InfosetOutputter
   with Indentable {
 
   def this(os: java.io.OutputStream, pretty: Boolean) = {
-    this(new java.io.OutputStreamWriter(os, StandardCharsets.UTF_8), pretty)
+    // using a BufferedWriter provides significant performance improvements
+    this(
+      new java.io.BufferedWriter(new java.io.OutputStreamWriter(os, StandardCharsets.UTF_8)),
+      pretty,
+    )
   }
 
   // Keeps track of if the next element we see is the first child or not of a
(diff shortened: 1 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetOutputter.scala
```diff
@@ -29,12 +29,15 @@ import org.apache.daffodil.runtime1.api.InfosetSimpleElement
...
 
-class JsonInfosetOutputter private (writer: java.io.Writer, pretty: Boolean)
+class JsonInfosetOutputter private (writer: java.io.BufferedWriter, pretty: Boolean)
   extends InfosetOutputter
   with Indentable {
 
   def this(os: java.io.OutputStream, pretty: Boolean) = {
-    this(new java.io.OutputStreamWriter(os, StandardCharsets.UTF_8), pretty)
+    this(
+      new java.io.BufferedWriter(new java.io.OutputStreamWriter(os, StandardCharsets.UTF_8)),
```
mbeckerle, line 38: Add a comment here that this buffered writer provides significant performance improvement. 

My first thought was that you should only add a Buffered writer if os argument is non-buffered, so you should conditionally check if the os output stream is already buffered. 

But we need a writer wrapped around the output stream anyway. And even if the output stream is buffered, use of a buffered writer may still  speed things up.

mbeckerle (approved):
+1
I had only one thought about a comment being added.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetOutputter.scala
stevedlawrence: > Add a comment here that this buffered writer provides significant performance improvement.

WIll do.

> But we need a writer wrapped around the output stream anyway.

Correct. When I was looking at the profiler, every write required spinning up a stream encoder, which has a lot of overhead. It's much more efficient to encode a bunch of characters at once. So we need a `BufferedWriter` regardless of what the `os` parameter is.

I just tested what happens if the `os` parameter is already a `BufferedOutputStream` (i.e. essentially double buffering), and it doesn't make a noticeable difference in parse time. So it's unnecessary to pass in a `BufferedOutuptStream`, but it doesn't hurt.

tuxji (approved):
+1

PR #1164, opened 2024-02-23
Include the input buffer in ExpectIt failure messages

stevedlawrence:
When ExpectIt fails to find a match in CLI tests, it adds the input buffer that scanned in the exception object it throws. Unfortunately, it does not include that buffer in the actual message, likely because it could be pretty verbose.

Most of our CLI test outputs are fairly small, so we add the input buffer to the actual exception message so it can be seen by users. This should make debugging failing CLI tests a bit easier without having to enable the debug parameter in runCLI and re-run the test.

DAFFODIL-2878


mbeckerle (approved):
+1

tuxji (approved):
+1

PR #1165, opened 2024-02-23
First Cut at new API for Layers - for review

mbeckerle:
This PR won't build or run. Ignore that.

I've stripped away the implentation of all this so as to focus review on how the layers came out from an API and supportability of the API perspective only.

So what is here are all the built-in and test layers that are part of Daffodil, and the API files defined by Daffodil that these layers use/need.

A later commit will make this all runnable (done, and all tests pass).

DAFFODIL-2825


daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/Base64MimeLayer.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.collection.JavaConverters._
+
+import org.apache.daffodil.runtime1.layers.api.JLayerLengthKind
+import org.apache.daffodil.runtime1.layers.api.Layer
+import org.apache.daffodil.runtime1.layers.api.LayerRuntime
+
+final class Base64MimeLayer
+  extends Layer(
+    layerName = "base64_MIME",
+    supportedLayerLengthKinds = Seq(JLayerLengthKind.BoundaryMark).asJava,
```
stevedlawrence, line 32: I'll review this more in depth next week, but a single high level question: Is there a fundamental difference between limiting layers and a encode/decode layers? For example,  could layers + lengths instead be something like:

```xml
<sequence dfdl:layer="boundaryMark"  ...>
  <sequence dfdl:layer="base64_MIME" ...>
    ...
  </sequence>
</sequence>
```

So, could a "limiting layer" just be a layer that just reads/writes bytes 1-to-1, but stops transferring bytes at some point in the data. And an "encoding layer" is just a layer that transforms bytes up until it runs out of data? Which could be limited by a parent element with lengthKind="..." or something more complex layer like a boundary mark layer. And that could mean a single layer could be both a "limiting layer" and an "encode/decode" layer, i.e. a layer reads bytes up until some point that it determines it's done *and* transforms the bytes that it read.

mbeckerle: Hmmm. Perhaps. 

One thing I did is make it so layers mostly don't have to deal with length limiting. They just declare whether they want implicit, explicit, or boundaryMark, and the system does the rest. That said, Some of the layers needed more than this. Line folding specifically. 

This idea that the ideas can be decoupled is kind of interesting. boundaryMark is a kind of delimited behavior which (1) can be used on sequences (not elements like regular dfdl:lengthKind delimited) and (2) doesn't deal with escape schemes. For parsing the boundaryMark is found and removed, unparsing it is inserted. 

Explicit is just a thing that halts data delivery after N bytes for parsing, does nothing at all on unparsing.

Implicit just does nothing, parsing or unparsing.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/LineFoldedTransformer.scala
```diff
@@ -75,157 +73,62 @@ import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
...
-    s
-  }
-}
+  /**
+   * This is CRLF not followed by space or tab.
+   * Note that if the CR is at end-of-data that is NOT a match. (See the $ in the negative lookahead)
+   *
+   * Note: one must not use capturing groups in these regex.
+   */
+  override protected def regexForBoundaryMarkMatch: String = "\\r\\n(?!(?:\\t|\\ |$))"
```
mbeckerle, line 120: Remove "$" from the regex. Left-over from an experiment I was doing.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/LineFoldedTransformer.scala
```diff
@@ -393,11 +298,25 @@ class LineFoldedOutputStream(mode: LineFoldMode, jos: OutputStream) extends Outp
     }
   }
 
+  // TODO: This is broken because an isolated \r or \n is output as \r\n, but without verifying that the next character
+  //  is space or tab. If that ends up as a \r\n not followed by a space or tab it has broken the data, as on a reparse
+  //  that \r\n will be interpreted as an end-of-layer marker.
```
mbeckerle, line 303: That's true only if the \r\n is being used as a boundaryMark. If the layerLengthKind is 'implicit' then this insertion of \r\n doesn't harm anything. 

Maybe line folded layers don't need boundaryMark support at all? Maybe implicit is enough?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerException.scala
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package org.apache.daffodil.runtime1.layers
+
+abstract class LayerException(msg: String, cause: Throwable) extends Exception(msg, cause) {
+  def this(msg: String) = this(msg, null)
+  def this(cause: Throwable) = this(null, cause)
+}
+
+class LayerCompilerException(msg: String, cause: Throwable) extends LayerException(msg, cause) {
```
mbeckerle, line 24: Some of these belong in the api package because layer authors may want to throw them?

Or maybe we let the user throw whatever they want, we catch and encapsulate as either compile or runtime layer exceptions?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerImplicits.scala
mbeckerle, line 1: I had hoped that Scala implicits would let scala code ignore the difference between Scala Option and Java Optional, and scala code wouldn't have the clutter. 

I didn't achieve that, however. You still have to invoke ".asJava" or ".asScala" in lots of places.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/JLayerLengthKind.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.daffodil.runtime1.layers.api;
+
+public enum JLayerLengthKind {
```
mbeckerle, line 19: I don't like the name JLayerLengthKind.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/JLayerLengthUnits.java
mbeckerle, line 1: The whole concept of LayerLengthUnits can go away. It's not needed. It was put in by analogy to dfdl:lengthUnits, but for layers, they're byte centric, so this property is either ignored, or must be 'bytes'. 

Should just drop the whole concept.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerPropertyInfo.java
```diff
@@ -0,0 +1,86 @@
...
+   * These can be suppressed using the warn ID "layerCompileWarning"
+   *
+   * @param msg  A string which is included with the diagnostic information.
+   * @param args any number of arguments which are substituted into the msg via the same
+   *             mechanism as String.format
+   */
+  // TODO: Remove if not needed. Hard to implement, as the LayerRuntimeInfo object
+  // Doesn't have compiler context for issuing warnings.
+  //  void schemaDefinitionWarning(String msg, Object... args);
+}
```
mbeckerle, line 86: Remove SDW. Haven't needed it yet.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/AISTransformer.scala
```diff
@@ -20,130 +20,78 @@ package org.apache.daffodil.runtime1.layers
...
-    newJOS
-  }
+  override def layerLimiter(layerPropertyInfo: LayerPropertyInfo): Optional[LayerLimiter] =
+    Optional.of(new BoundaryMarkRegexLimiter {
+      override protected def regexForBoundaryMarkMatch: String = ","
+
+      override protected def boundaryMarkToInsert: String = ","
+
+      override protected def maximumLengthBoundaryMark: String = ","
+    })
```
mbeckerle, line 75: The above is just unnecessary. This is just regular boundaryMark behavior, where the boundary mark is a constant built into the layer code. Why bother. Just in the AIS layer DFDL file, define dfdlx:boundaryMark="," and then drop this stuff.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/CheckDigitLayer.scala
mbeckerle, line 1: CheckDigit as a layer example should go away. Really check digits computation should be a UDF. 

For now it's useful as just another layer example.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/IPv4ChecksumLayer.scala
```diff
@@ -18,48 +18,77 @@
...
+  with LayerChecksumMixin // Checksums are easily expressed as a compute method on a byte buffer.
+  {
 
   /**
    * This layer is always exactly 20 bytes long.
    */
-  override protected def layerBuiltInConstantLength = Some(20L)
+  private def lenInBytes = 20 // bytes
+
+  override def optLayerBuiltInConstantLength: Optional[Int] = Optional.of(lenInBytes)
```
mbeckerle, line 69: Could just punt on this built in fixed length stuff and just use layerLengthKind explicit for these. 

Or, could just use layerLengthKind 'implicit' and implement 'explicit' using an element of specified length that surrounds the part that needs to have the fixed length.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/TestLayers.scala
```diff
@@ -37,7 +37,7 @@ class TestLayers {
 
   @Test def test_gzipLayer1(): Unit = { runner.runOneTest("gzipLayer1") }
   @Test def test_foldedIMFBase64Layers1(): Unit = {
-    runner.runOneTest("foldedIMFBase64Layers1")
+    runner.trace.runOneTest("foldedIMFBase64Layers1")
```
mbeckerle, line 40: Revert

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
```diff
@@ -0,0 +1,275 @@
...
+ * @param supportedLayerLengthKinds list of the layer length kinds the layer supports
+ * @param supportedLayerLengthUnits list of the layer length units the layer supports
+ * @param isRequiredLayerEncoding true if the layer is textual and so needs the layerEncoding property
+ * @param optLayerVariables a LayerVariables structure describing the variables the layer accesses
+ */
+abstract class Layer(
+  val layerName: String,
+  val supportedLayerLengthKinds: java.util.List[JLayerLengthKind],
+  val supportedLayerLengthUnits: java.util.List[JLayerLengthUnits],
+  val isRequiredLayerEncoding: Boolean,
```
stevedlawrence, line 92: Is layer encoding needed anymore? It looks like CheckDigit and AIS are the only ones that set this to true. I don't see AIS using the layerCharset. CheckDigit layer uses it, but maybe we can get away with it just always being ascii?

Or alternatively, maybe this this is just another variable? For example, checkDigit already has a `checkDigitParams` variable, maybe an additional variable could be used to specify encoding?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
```diff
@@ -0,0 +1,275 @@
...
+   * Called exactly once when the schema is compiled to do extra checking that the layer is being used properly.
+   * The thrown exception becomes a SchemaDefinitionError at schema compile time.
+   *
+   * Example checks are:
+   * - layerEncoding is constant and is a single-byte charset
+   * - layerLength, if constant, is within a maximum value range
+   * - layerBoundaryMark string, if constant, is not too long and contains only allowed characters.
+   * These things can be required to be constant by this check, or it can check their values for legality
+   * if they happen to be constant. Since these are runtime-valued properties (can be expressions), then if the
+   * layer allowed that, they must also be checked at runtime.
```
stevedlawrence, line 109: In thinking more about how things might work if limiting and transform layers were all the same thing, we need to consider how layerLength, layerBoundaryMark, etc. settings are provided to layers. I guess these would have to be variables? For example, a boundary mark might look like:

```xml
<sequence>
  <annotation>
    <appinfo source="http://www.ogf.org/dfdl/">
      <dfdl:newVariableInstance ref="bm:boundaryMark" defaultValue="\r\n" />
    </appinfo>
  </annotation>
  <sequence dfdlx:layer="boundaryMark">
     <sequence dfdlx:layer="lineFolded_AIS" >
       ...
     </sequence>
  </sequence>
  ...
</sequence>
```

It's a bit more verbose, but it's much more general. I guess the idea is there are no other `dfdlx:layer*` properties, there is only `dfdlx:layer`. Layers can be either transform or limiting (or both), and all configuration of those layers must be done via variables.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
```diff
@@ -0,0 +1,275 @@
...
+  final def name(): String =
+    layerName // name() method with empty args is required by SPI loader
+
+  /**
+   * Called exactly once when the schema is compiled to do extra checking that the layer is being used properly.
+   * The thrown exception becomes a SchemaDefinitionError at schema compile time.
+   *
+   * Example checks are:
+   * - layerEncoding is constant and is a single-byte charset
+   * - layerLength, if constant, is within a maximum value range
```
stevedlawrence, line 105: Another thought, can layerLengthKind="explicit" and layerLength be removed, and just use a parent element with dfdl:length

For example, what is the difference between these two:

```xml
<sequence dfdlx:layer="foo" dfdlx:layerLengthKind="explicit" dfdl:lengthLength="100">
  ...
</sequence>
```

```xml
<element name="foo" dfdl:lengthKind="explicit" dfdl:length="100">
  <complexType>
    <sequence dfdlx:layer="foo">
      ...
    </sequence>
  </complexType>
</element>
```

The latter is more verbose, but I think it acheives the same thing? The idea being layers don't have support a specific kind of layer length. A layer just always reads until it hits the end of data, which might be EOF or might be a length provided by a parent element, or might be restricted by a surrounding layer that limits the length.

stevedlawrence (commented):
I really like where this is heading :+1:  I wonder if it can be simplified even more by removing the concept of layerLengthKind and the differentiation between liimting layers and encode/decode layers. It feels like the different layerLengthKinds can be implement as normal layers that just don't do a transform, but instead write EOF at some point. And variables can be used for things like layerLength, layerBoundaryMark, layerEncoding, etc.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/LineFoldedTransformer.scala
```diff
@@ -75,157 +73,62 @@ import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
...
+   */
+  override def layerLimiter(layerPropertyInfo: LayerPropertyInfo): Optional[LayerLimiter] =
+    layerPropertyInfo.layerLengthKind() match {
+      case JLayerLengthKind.Implicit =>
+        Optional.empty() // use the default implicit layer length implementation.
+      case JLayerLengthKind.BoundaryMark =>
+        Optional.of(new LineFoldedLayerBoundaryMarkLimiter(this))
+      case JLayerLengthKind.Explicit =>
+        Assert.invariantFailed("layerLengthKind 'explicit' not allowed.")
     }
```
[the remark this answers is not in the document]
mbeckerle: I think this is what I'll try next. I think all properties except dfdlx:layer go away. Regular DFDL variables provide everything including layer length, boundary marks, etc.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerException.scala
mbeckerle: I will follow this UDF style.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/JLayerLengthKind.java
mbeckerle: These properties go away in the "it's all just DFDL variables" formulation of this stuff.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
```diff
@@ -0,0 +1,275 @@
...
+ *
+ * @param prefix preferred namespace prefix for the namespace of the variables
+ * @param namespace namespace of the variables
+ * @param variables list of pairs, each is the name (an NCName, that is without any
+ *                  namespace prefix) and type of a variable
+ */
+final case class LayerVariables(
+  prefix: String,
+  namespace: String,
+  variables: java.util.List[(String, DFDLPrimType)],
```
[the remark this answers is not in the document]
mbeckerle: The string in this case can be one of these "{urn:someNamespace}localName" type strings to incorporate a namespace.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
```diff
@@ -0,0 +1,275 @@
...
+)
+
+/**
+ * This is the primary API class for writing layers.
+ *
+ * All layers are derived from this class, and must have no-args default constructors.
+ *
+ * Derived classes will be dynamically loaded by Java's SPI system.
+ * The names of concrete classes derived from Layer are listed in a resources/M.services file
```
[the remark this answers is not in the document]
mbeckerle: The name in my source tree is M.services.  I was wondering how it decides what files to read. The string M.services is not in our code base, so we aren't telling the SPI to look for a file of that name. It must be by convention that this file has "services" in its name or something like that.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
mbeckerle: I think this can just be a DFDL variable, if a layer needs it then it can define a variable to use.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
mbeckerle: Yes, but there is a difference in behavior. 

The layerLengthKind 'explicit' is only explicit on parse. On unparse it ignores the layerLength. 

This was to avoid the problem DFDL v1.0 has where the output wants to be perhaps a different length than the input, but to recompute that on output you have to measure the length not of the original element, but an inner element inside it, to avoid the circularity. 

But I'm inclined to try getting rid of dfdl:layerLengthKind entirely and see how it goes.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.scala
mbeckerle: This is what I will try next. More verbose, but that's mostly XSD's fault.  Only one thing to learn: variables.

mbeckerle:
I have addressed the comments here in a new PR, which includes not just the layers themselves but the new implementation also. 

Closing.

PR #1170, opened 2024-02-27
Second Cut at new API for Layers - for review

mbeckerle:
layer API is all defined in Java clases and interfaces now.

Reworked to use DFDL variables for all
communications into the layer code.

Split layers so that layer limiting is its own layer (for boundary mark and there is also a fixedLength layer).

There are two ways to do explicit layer limiting, that differ in an important way.

If you use 'implicit' layer length (no restriction layer), and constraint the layer length by a surrounding element of specified length, then the layer length is controlled when parsing, but NOT controlled when unparsing.

If you use the layer "fixedLength", that dictates the parse and unparse length to be the value of the layerLength variable. Similarly if you use a checksum layer built with the ChecksumLayer base class, the length is controlled for both parsing and unparsing.

DAFFODIL-2845


daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/ByteSwapTransformer.scala
mbeckerle, line 1: Should rename from "Transformer" to "Layer"

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dfdlx.xsd
```diff
@@ -123,106 +123,10 @@
 
   <xs:attributeGroup name="ExtLayeringAGQualified">
     <xs:attribute form="qualified" name="layerTransform" type="dfdlx:LayerTransformType" />
```
mbeckerle, line 125: Should we rename layerTransform to just layer now, since it is the only property left in the layer system?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/PackedBinaryUnparserTraits.scala
```diff
@@ -94,9 +94,9 @@ abstract class PackedBinaryDecimalBaseUnparser(
     val bigDec = number.asInstanceOf[JBigDecimal]
     if (bigDec.movePointRight(binaryDecimalVirtualPoint).scale != 0) {
       e.schemaDefinitionError(
-        "Decimal point of number '%s' does not match the binaryVirtualDecmialPoint: %d",
+        "Decimal point of number '%s' does not match the binaryVirtualDecmialPoint: %s",
         bigDec,
-        binaryDecimalVirtualPoint,
+        binaryDecimalVirtualPoint.toString,
```
mbeckerle, line 99: This is a java-ism about varargs. Can't pass primitive types from scala code to java varargs code.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * You don't have to check that the variables are defined and declared in matching manner, that happens automatically.
+   *
+   * @throws LayerException on any error.
+   *
+   */
+  public void check(LayerCompileInfo layerPropertyInfo) throws LayerException { /* nothing */ }
+
+
+  public InputStream wrapLayerDecoder(InputStream jis, LayerRuntime lr) throws IOException { return null; }
```
mbeckerle, line 87: These wrap methods should not have default implementations.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerCompileInfo.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public interface LayerCompileInfo {
+
+  String layerName();
+
+  SchemaFileLocation schemaFileLocation();
+
+  /**
+   * Obtains the LayerVariable object that can be used to access the variable at runtime.
+   *
```
mbeckerle, line 36: LayerVariable objects are serializable (they are a view trait on VariableRuntimeData objects) - these could be computed and saved as part of the LayerTransformer object. 

Right now there's a lot of check(...) code dealing with variables. It would be good to eliminate that complexity.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerRuntime.java
```diff
@@ -0,0 +1,57 @@
...
+ * when parsing, but are fatal when unparsing.
+ *
+ * Also provides the ability to cause runtime schema definition errors, which are always
+ * fatal.
+ *
+ * This object contains the processor state, but hidden behind an API so that only relevant
+ * aspects of the processor state are visible.
+ */
+public interface LayerRuntime extends LayerCompileInfo {
+
```
mbeckerle, line 35: Java doc needed for all methods. Should inherit them from LayerCompileInfo for most of the methods.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/FixedLengthLayer.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.daffodil.runtime1.layers.api.LayerCompileInfo
+import org.apache.daffodil.runtime1.layers.api.LayerRuntime
+import org.apache.daffodil.runtime1.layers.api.LayerVariable
+
+import org.apache.commons.io.IOUtils
+
+/**
+ * Suitable only for checksums computed over small sections of data, not large data streams or whole files.
+ *
+ * The entire region of data the checksum is being computed over, will be pulled into a byte buffer in memory.
```
stevedlawrence, line 37: Is this just for ease of implementation? I don't think the checksum layer has access to this byte array right? Might be worth adding a TODO to fix that, seems this layer could be implemented in a streaming manner so it doesn't need to copy everything to another buffer.

Should maybe also remove mentions of "checksum" since this could really be used for anything (that is relatively small).

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerRuntimeImpl.scala
```diff
@@ -0,0 +1,116 @@
...
+    state.setVariable(
+      variable.asInstanceOf[VariableRuntimeData],
+      DataValue.toDataValue(v),
+      srd,
+    )
+
+  override def schemaDefinitionError(msg: String, args: AnyRef*): Unit =
+    runtimeSchemaDefinitionError(msg, args)
+
+  override def getCharset(layerEncoding: String): Charset = Charset.forName(layerEncoding)
```
stevedlawrence, line 96: Should we remove this from LayerRuntime? Layers that need a charset can call `Charset.forName` themeselves?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerTransformerFactory.scala
```diff
@@ -17,15 +17,49 @@
...
  * Factory for a layer transformer.
  *
  * This is the serialized object which is saved as part of a processor.
  * It constructs the layer transformer at runtime when newInstance() is called.
  *
- * This must be implmented as part of implementation of a daffodil layer.
+ * This allows the layer instance itself to be stateful.
  */
-abstract class LayerTransformerFactory(val name: String) extends Serializable {
+class LayerTransformerFactory(val layerCompileInfo: LayerCompileInfo) extends Serializable {
```
stevedlawrence, line 32: Wondered if we could get rid of this in another comment. If we don't, I'd suggest we rename these transformer classes. I'm not sure they really do any transforming anymore, they really just a wrappers to allow Layers to store state in the `Layer` instance. Maybe `LayerWrapper` or something?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public OutputStream wrapLayerEncoder(OutputStream jos, LayerRuntime lr) throws IOException { return null; }
```
stevedlawrence, line 89: Thoughts on naming these both `wrap` with either `InputStream` or `OutputStream` parameter types? Limiting layers don't really encode/decode so it's maybe confusing?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerRuntime.java
```diff
@@ -0,0 +1,57 @@
...
+  void setString(LayerVariable variable, String s);
+
+  int getInt(LayerVariable variable);
+
+  void setInt(LayerVariable variable, int v);
+
+  long getLong(LayerVariable variable);
+
+  void setLong(LayerVariable variable, long v);
+}
```
stevedlawrence, line 57: A conseequence of the above comment if variables are just turned into function arguments/return value, much of this goes away. We wouldn't need the LayerVariable stuff at all (and could support more than just Int/String/Long, we could support anything a UDF supports.

I also wonder if we could get rid of the processing error an sde functions? Could layers just be expected to throw `LayerProcessorErrorException` or `LayerSchemaDefinitionErrorException`, and then the sequence layer parser/unparse could catch those? I guess maybe you lose context about where the error occurs? Since the exception would be caught by some random parser/unparse, and not the layer parser/unparse. I was thinking maybe the LayerRuntime could be completely removed, but I think not since you do need the layer context, I think.

stevedlawrence (commented):
Awesome change! It would be nice if we could come up with a way to remove the "check" function. I suggested one possibility, I'm not sure if I like it, but maybe something similar would work.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerCompileInfo.java
mbeckerle: I agree with the goal. Get rid of check() entirely (or most layers wouldn't need one). 

A layer has a namespace, all variables in that namespace are made available for reading and writing by the layerRuntime. The layer expresses which variables are input, which are output, and what the expected types for them are. Nothing is saved by the check() routine for re-use.

mbeckerle (commented):
I'm still working on this. No need to re review the code. I'm just saving a checkpoint for backup.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/ByteSwapTransformer.scala
```diff
@@ -22,53 +22,27 @@ import java.io.OutputStream
...
-    }
+import org.apache.daffodil.runtime1.layers.api.Layer
+import org.apache.daffodil.runtime1.layers.api.LayerRuntime
 
-    layerCompileInfo.SDEUnless(
-      layerCompileInfo.optLayerLengthOptConstantValue.isDefined,
-      "The property dfdlx:layerLength must be defined.",
-    )
+final class TwoByteSwapLayer extends ByteSwap("twobyteswap", 2)
+final class FourByteSwapLayer extends ByteSwap("fourbyteswap", 4)
```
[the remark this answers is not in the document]
stevedlawrence: It looks like the byte swap algorithm works for any number. So it should work fine for 3, 5 ,1000, etc. without any changes. I would be surprised if any format actually needed any of those, but I could imagine a format having a word size of 8 or 16 or something. I think the only check we would need is that it's non-zero, which could be enforced by making the variable an unsignedInt, or unsignedByte if there are concerns with very large swaps.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerRuntime.java
mbeckerle: Yes, the primary purpose of the LayerRuntime object is to be the context object of that use of the layer so that thrown exceptions get associated with that context and diagnostics messaging can point at that layer use point in the schema. 

The problem is that the API being called is just InputStream.read or OutputStream.write, and I don't want to add a proxy and try-catch on every such call just so that if an error occurs we can associate that thrown exception with the daffodil layer context. 

For parsing we could avoid this as it proceeds in an orderly traversal, so the LayerParser can have the catches. 

For unparsing.... the unordered nature of unparsing mean that the LayerUnparser is long gone from the stack by the time the layer is potentially being written to and closed, which is at the time suspensions are finally being completed allowing the blocked unparsing that writes to them to finish. This could be in some context where I'm not sure what exactly would be catching the exceptions. 

This needs more thought.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/LayerCompileInfo.java
mbeckerle: Trying to simplify this idea slightly. What do you think of this.

Each layer has a target namespace. All variables in that namespace must either be parameters to, or return results from, the layer. 

The layer class must have a default constructor so SPI can load it.

It also must have a constructor with an arg for each parameter variable - name of which matches the variable name, type of which is the Java/Scala type corresponding to the mapping from DFDLPrimType. (ex: xs:date becomes icu.Calendar, unsignedShort becomes JInt, etc. runtime1.api.InfosetSimpleElement defines this corresondence all in one place.)

It also must have a getter for each result variable, the name of which is "get_" appended to the variable name (or maybe it is just exactly the variable name). The return type is determined the same way as the arg types to the constructor. 

At layer schema compile time each variable in the layer namespace must have a corresponding constructor arg, or a corresponding getter, and the types have to match up or it's an SDE. 

At runtime the layer is constructed using the variable values, when layer processing completes the framework will call the getters for each result variable and populate those variables. 

This is not too bad. The inputter and ouputter streams can share any variable logic via methods on the layer class itself.

mbeckerle:
Closing for now. I may reopen when ready for review again.

PR #1174, opened 2024-02-29
Pin github actions to a commit hash

stevedlawrence:
Pinning to a tag can lead to breaking builds or malicious actors compromising our actions if they are able to rename or delete a tag. Pinning to a commit hash avoids these issues. There is no loss of functionality--dependabot is able to update commit hashs and comments containing the associated tag.

DAFFODIL-2881

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Check out Repository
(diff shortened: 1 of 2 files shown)


tuxji (approved):
+1

Verified all hashes are correct.

stevedlawrence:
This is just replacing tags with equivalent commit hash so I don't think an additional +1 is necessary. I'll merge.

PR #1176, opened 2024-03-03
Third Cut at Simplfying Layers - uses reflection to simpify variable handling

mbeckerle:
Please look first at the code for the layers themselves in daffodil-runtime1-layers, and in daffodil-test.

Then look at the ".api" packages for anything in them that needn't be there.

Still TBD:
- a possible namespace bug: bswap prefix vs. bs prefix for byteSwapLayer.dfdl.xsd
- change dfdlx:layerTransform to dfdlx:layer and vals/vars/args named layerTransformer to either layer or layerFactory
- review TODO/FIXME and  a few commented out code sections I've not yet revisited. 
-


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/VariableMap1.scala
```diff
@@ -278,8 +280,8 @@ class VariableMap private (vrds: Seq[VariableRuntimeData], vTable: Array[Seq[Var
   }
 
   def cloneForSuspension(): VariableMap = {
-    val newTable = new Array[Seq[VariableInstance]](vTable.size)
-    Array.copy(vTable, 0, newTable, 0, vTable.size)
+    val newTable = new Array[Seq[VariableInstance]](vTable.length)
+    Array.copy(vTable, 0, newTable, 0, vTable.length)
     new VariableMap(vrds, newTable)
```
stevedlawrence, line 285: Is here a different between size vs length?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/LayeredSequenceParser.scala
```diff
@@ -17,56 +17,48 @@
...
-      case le: LayerNotEnoughDataException =>
-        PENotEnoughBits(
-          state,
-          le.schemaFileLocation,
-          le.dataLocation,
-          le.nBytesRequired * 8,
-        )
+      case pe: ParseError =>
+        state.setFailed(pe)
       case e: Exception =>
```
stevedlawrence, line 60: Could a runtime SDE be caught here? Should that be treated differently than LayerUnexpectedException?

project/Dependencies.scala
```diff
@@ -25,6 +25,7 @@ object Dependencies {
     "com.lihaoyi" %% "os-lib" % "0.9.3", // for writing/compiling C source files
     "org.scala-lang.modules" %% "scala-xml" % "2.2.0",
     "org.scala-lang.modules" %% "scala-parser-combinators" % "2.3.0",
+    "org.scala-lang.modules" %% "scala-java8-compat" % "1.0.2",
```
stevedlawrence, line 28: What is this for? This has a NOTICE file so we'll need to update our NOTICE files.

stevedlawrence (commented):
Really happy with how this. Just a bunch of thoughts. Only real question is if things are made more simple by using a setDFDLVariables function for input variables instead of using alternative constructor.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerFactory.scala
```diff
@@ -0,0 +1,326 @@
...
+      // least as far as the number and type of the DFDL variables it consumes and writes goes.
+      // However, we want to do this exactly once, not every time this method is called,
+      // So there is a cache of instances that have already been through these checks,
+      // at runtime.
+      // We compute this data structure once only at the time the SPI Loads the layer
+      // class.
+      // In addition, this process of verifying the DFDL variables used by the layer
+      // pre-computes some data structures that facilitate fast run-time processing
+
+      val lvr = computeLayerVarsRuntime(lri.layerRuntimeData, spiLayerInstance)
```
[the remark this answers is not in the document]
mbeckerle: This one I have to think about. Perhaps we can do this. Once we have the layer we have the layer namespace, and that will let us get the variables if we have the data processor. That's the issue. Do we have the data processor object (with schema set runtime data) already in hand at this point, or is that 'in process' of being created?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerDriver.scala
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Parsing works as a tree traversal, so when the parser unwinds from
+   * parsing the layer we can invoke this to handle cleanups, and
+   * finalization issues like assigning the result variables
+   */
+  final def removeInputLayer(
+    s: InputSourceDataInputStream,
+    layerRuntimeImpl: LayerRuntimeImpl,
+  ): Unit = {
+    layerVarsRuntime.callGettersToPopulateResultVars(layer, layerRuntimeImpl)
```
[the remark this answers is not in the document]
mbeckerle: The LayerFactory has to _get_ variable values. The LayerDriver has to call the getters to assign the output variables. 

This is ordinary I think. One could do something more java-beans flavored, and have setters for the args and getters for the results, but I think that's clumsier. Never did like java beans approach. So many lines of code to do nothing much.

daffodil-runtime1-layers/src/main/resources/org/apache/daffodil/layers/xsd/base64_MIMELayer.dfdl.xsd
```diff
@@ -0,0 +1,37 @@
...
+  <!--
+  This layer does not define any variables for parameters or results
+  -->
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+      <dfdl:defineFormat name="base64_MIME">
+        <dfdl:format dfdlx:layerTransform="b64:base64_MIME"/>
+      </dfdl:defineFormat>
+    </appinfo>
+  </annotation>
```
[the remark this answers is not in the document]
mbeckerle: Getting rid of the dfdl:ref. They add no value over just dfdlx:layer with a QName arg.

daffodil-core/src/main/scala/org/apache/daffodil/core/layers/LayerCompiler.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  private def srd = sq.sequenceRuntimeData
+  private def lrd = sq.optionLayerRuntimeData.get
+
+  private def layerQName: RefQName = lrd.layerQName
+  private def layerName = layerQName.local
+  private def layerNamespace = layerQName.namespace
+
+  private val layeredSequenceAllowedProps = Seq(
+    "ref",
```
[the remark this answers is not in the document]
mbeckerle: Removing.

mbeckerle (commented):
I'm addressing most of your comments in the next commits. 

There is obviously layer doc to write, and tests to cover many areas must be created.

I plan to first investigate other layer implementations that exist in house at Owl to make sure those will convert easily. 

Then I will return to the testing, doc, and the last few unresolved issues for which the comments are not resolved.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -932,6 +936,33 @@ object NodeInfo extends Enum {
...
+        case DFDLPrimType.UnsignedLong => classOf[JBigInt]
+        case DFDLPrimType.NonNegativeInteger => classOf[JBigInt]
+        case DFDLPrimType.Double => classOf[JDouble]
+        case DFDLPrimType.Float => classOf[JFloat]
+        case DFDLPrimType.HexBinary => classOf[Array[Byte]]
+        case DFDLPrimType.AnyURI => classOf[java.net.URI]
+        case DFDLPrimType.Boolean => classOf[JBoolean]
+        case DFDLPrimType.DateTime => classOf[ICUCalendar]
+        case DFDLPrimType.Date => classOf[ICUCalendar]
+        case DFDLPrimType.Time => classOf[ICUCalendar]
```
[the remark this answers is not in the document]
stevedlawrence: Maybe we have the same supports as user defined functions? I think that's basically just all numerics and byte array. So we exclude date/time types and anyURI. I think there is value in supporting aditional things like double/float, boolean so I'd prefer not to limit it to just string and int.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerDriver.scala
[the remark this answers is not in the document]
stevedlawrence: The more I think about this, I actually like setters for the args and getters for the results. Using a setter gives a single place to do input error detection and avoids the dummy constructor stuff in scala implementations. Implementing a setter in scala feels a bit dirty, but this is a Java API where setters are normal. And you need mutable state for any getters, so more mutable state for setters is at least consistent.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerFactory.scala
```diff
@@ -0,0 +1,326 @@
...
+          lrd.vmap.getOrElse(
+            p.getName,
+            lrd.context.SDE(s"No layer DFDL variable named '$p.getName' was found."),
+          )
+        }.toSeq
+
+        // Now we deal with the result getters and the corresponding vars
+        //
+        val allLayerVRDs = ListSet(lrd.vmap.toSeq.map(_._2): _*)
+        val returnVRDs = allLayerVRDs -- paramVRDs // set subtraction
```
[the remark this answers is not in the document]
stevedlawrence: A thought, is there any value in saying that input variables are defined by by the constructor parameters and output variables are defined by the getters, and additional variables are ignored. This would allow layers to have variables that are sort of "helper" variables? For example, what about something like this:

**myLayer.dfdl.xsd**
```xml
...
<defineVariable name="foo" type="xs:int" defaultValue="2" />
<defineVariable name="bar" type="xs:int" defaultValue="{ $foo * 100}" />
...
```

The constructor might only accept and use `bar`, but a user to do something like this:
`daffodil parse -Dlyr:foo=10 ...`

So they could change the default value of `foo`, which would change the value of `bar`, which is what the layer actually uses. Or the user could just set `lyr:bar` directory if they want that extra control.

I'm not sure if that adds value or not. But it might simplify the implementation of this function since some of the logic here is making sure all defined variables are used?

I guess the argument against this is that a user might forget to use a variable and we want to alert them of that?

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/FixedLengthLayer.scala
```diff
@@ -0,0 +1,144 @@
...
+      isOpen = false
+      val ba = baos.toByteArray
+      val baLen = ba.length
+      if (baLen != layerLength)
+        lr.processingError(
+          s"Insufficient output data for fixed-length layer. Needed $layerLength bytes, but only $baLen were unparsed.",
+        )
+      jos.write(ba)
+
+      // TODO: Consider if this close should happen in the framework instead of here.
```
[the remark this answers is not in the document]
mbeckerle: This close is now in the framework. User can ignore closing anything.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/ChecksumLayer.java
```diff
@@ -0,0 +1,53 @@
...
+   *                   If the bytes in this buffer are modified by the compute method, those modified bytes are what
+   *                   the parsing will parse from, and the unparsing will output.
+   * @return the checksum value as an Int (32-bit signed integer)
+   */
+  public abstract int compute(
+    LayerRuntime layerRuntime,
+    boolean isUnparse,
+    ByteBuffer byteBuffer
+  );
+}
```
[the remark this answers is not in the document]
mbeckerle: Yes that's how it works. I plan to change this to a "group" setter call that sets all the parameters, and then a single result getter  retrieval call  that way the constructor is just a plain no-arg constructor. 

One annoyance is that this has to return a JInt now, not just int. Same with the parameter args: they can't be int, they have to be JInt.  The types have to be the boxed ones. I will look into whether that is truly necessary or not. 

I think the prefix "getDFDLResultVariable_" is probably wrong. It should mention that it's about layers, so maybe "getDaffodilLayerResult_"?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerDriver.scala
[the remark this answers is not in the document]
mbeckerle: I think I would prefer one setter for all params. Only a zero-arg default constructor is needed then. 

But we still need to compute the call to the setter at schema compile time, which checks the types and number and names of the parameters against the variable definintions. 

It's much closer to the java-beans approach, ie., 100% of making sure all the parameters are correct is on the user's back, vs the constructor approach, but both are viable.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/VariableMap1.scala
mbeckerle: It seems length is preferred to size in Scala code style.

daffodil-runtime1-layers/src/main/resources/org/apache/daffodil/layers/xsd/boundaryMarkLayer.dfdl.xsd
```diff
@@ -0,0 +1,39 @@
...
+  xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+  xmlns:bm="urn:org.apache.daffodil.layers.boundaryMark"
+  targetNamespace="urn:org.apache.daffodil.layers.boundaryMark">
+
+  <annotation>
+    <appinfo source="http://www.ogf.org/dfdl/">
+
+      <dfdl:defineVariable name="boundaryMark" type="xs:string"/>
+      <dfdl:defineVariable name="layerEncoding" type="xs:string"/>
```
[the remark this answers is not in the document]
mbeckerle: Agree a README is the right thing. 

We have to make sure the README is included in the delivered jars though.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/ChecksumLayer.java
```diff
@@ -23,29 +23,29 @@
...
- *
- *     Int getDFDLResultVariable_digest() { return getChecksum() }
- *
+ * <p/>
+ *     Int getLayerVariableResult_digest() { return getChecksum() }
+ * <p/>
  * This will be called automatically to retrieve the integer value that was returned from the `compute` method, and
  * the DFDL variable 'digest' will be assigned that value.
  */
 public abstract class ChecksumLayer extends ChecksumLayerBase {
```
stevedlawrence, line 45: Do we need a comment that implementations must call `setLength` prior to wrap being called?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -63,11 +63,11 @@
  * <p/>
```
stevedlawrence, line 63: THis doc needs updating to use the new setVariable method.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -63,11 +63,11 @@
...
  * A layer that wants to return a value after the layer algorithm completes defines a special recognizable
  * getter method. The name of the getter is formed from prefixing the DFDL variable name with the string
- * 'getDFDLResultVariable_'. The return type of the getter must match the type of the variable.
+ * 'getLayerVariableResult_'. The return type of the getter must match the type of the variable.
  * <p/>
  * For example, a result value getter for a DFDL variable named 'checksum' of type xs:unsignedShort would be:
  * <pre>
- *      Int getDFDLResultVariable_checksum() {
+ *      Int getLayerVariableResult_checksum() {
```
stevedlawrence, line 70: Do these set/getVariable functions need o be public, or can reflection still find them if they are package private? 

Also, it might be worth using `int` instead of `Int` since this is javadoc.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/CheckDigitLayer.scala
```diff
@@ -47,33 +47,37 @@ object CheckDigitLayer {
...
+    this.digitEncoding = digitEncoding
+  }
+
   /**
    * Result DFDL variable value getter. This getter is called to obtain the value for,
    * and populate the DFDL variable named checkDigit, in the layer's target namespace.
    * @return the check digit value
    */
-  def getDFDLResultVariable_checkDigit: JInt =
+  def getLayerVariableResult_checkDigit: JInt =
```
stevedlawrence, line 78: Just curious if this must be a Jint or would a primitive Int work too? I think it would work with the new changes? Should our examples use primitive int since I imagine that what most Java devs would find more natural?

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/IPv4ChecksumLayer.scala
```diff
@@ -36,38 +36,42 @@ final class IPv4ChecksumLayer()
...
    * @param byteBuffer contains the 20 bytes to be used to compute this checksum.
    * @return the computed checksum value as a 32-bit signed Int.
    */
   override def compute(
-    layerCompileInfo: LayerRuntime,
+    layerRuntime: LayerRuntime,
     isUnparse: Boolean,
     byteBuffer: ByteBuffer,
   ): Int = {
+    init(layerRuntime)
```
stevedlawrence, line 74: I'm not a huge fan of this init, it feels an awful lot like the check() function that we tried to remove. Maybe it's required in this case? I did mention a possible alternative in another comment where the set variable function calls setLength, but I guess the issue here is this layer don't have the a set variable function to call setLength since this layer has no variables?

A thought, does it work to have an empty set variable functions? For example, maybe this checksum layer implementation could do:
```scala
def setLayerVariableParameters(): Unit = {
  setLength(20)
}
```

So the documentation for ChecksumLayer api says all implementations must implement a setLayerVariableParameters function and must call setLength(). If the layer has no variables, the function can have an empty parameter list.

So setLayerVariablParameters kindof becomes a layers `init()` function, and it can be parameterized by variables or not?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/LayerDriver.scala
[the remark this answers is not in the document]
mbeckerle: This is cleaned up next push of changes.  The LayerDriver class now does all the invocation of the setter for vars and getter for results. The LayerFactory just invokes the LayerRuntimeCompiler and creates the Layer object instance, and LayerDriver that encapsulates it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/ChecksumLayer.java
mbeckerle: fixed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/ChecksumLayer.java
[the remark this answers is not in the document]
mbeckerle: Turns out the code is tolerant of the unboxed types. I just had to make our own type checking tolerate it. 

Javadoc for Layer has been updated to document the getter.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
mbeckerle: updated in next commits

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
mbeckerle: package private works also. I've changed all ours, since those will be used as examples.

daffodil-runtime1-layers/src/main/resources/org/apache/daffodil/layers/xsd/boundaryMarkLayer.dfdl.xsd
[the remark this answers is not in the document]
mbeckerle: I have NOT addressed this yet.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/FixedLengthLayer.scala
```diff
@@ -42,33 +40,35 @@ import org.apache.commons.io.IOUtils
...
-
-  private def maxFixedLength = Short.MaxValue
-
-  override def wrapLayerInput(jis: InputStream, lr: LayerRuntime): InputStream = {
+  def setLayerVariableParameters(fixedLength: Long): Unit = {
+    this.fixedLength = fixedLength
+  }
 
+  private def check(lr: LayerRuntime): Unit = {
+    if (fixedLength < 1)
```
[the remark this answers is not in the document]
mbeckerle: Yes, this is gone. In the IPV4Checksum example, the setLayerVariableParameters has no args and just serves to initialize.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/ChecksumLayerBase.scala
```diff
@@ -38,11 +38,35 @@ import org.apache.commons.io.IOUtils
...
+
+  /**
+   * Initializes the layer.
+   *
+   * This method is called to ensure that `setLength` has been called before the streams are created.
+   * It can be overridden in subclasses, but those must run the `super.init(lr)`
+   *
+   * @param lr The LayerRuntime instance.
+   */
+  protected def init(lr: LayerRuntime): Unit = {
```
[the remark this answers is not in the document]
mbeckerle: not calling setLength is an SDE now, and check is gone.

mbeckerle (commented):
This is ready for a significant review now.
@tuxji now would be worthwhile to add your perspective
@pkatlic please take a look
@olabusayoT please take a look - focus on the built in and example layers, and the APIs they use more than the implementation, though of course comments are welcome on that also.

Not done: 

- user documentation of these layers
- clean up of this AIS example - which is using internals of BitsCharset so isn't really a clean user example
- tests that pass every type of variable, multiple getters that get every type of result
- negative tests to excite all the error situations (which codecov will spot when those error paths aren't used.
- at Owl we have several layers that are in-house things, and I haven't yet re-done those to match this, but I'm pretty confident they will work fine, as they are just a variety of proprietary checksums, and one interesting zip-file-layer one.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * This is the primary API class for writing layers.
+ * <p/>
+ * All layers are derived from this class, and must have no-args default constructors.
+ * <p/>
+ * Derived classes will be dynamically loaded by Java's SPI system.
+ * The names of concrete classes derived from Layer are listed in a resources/M.services file
```
tuxji, line 31: *resources/META-INF/services

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/CheckDigitLayer.scala
tuxji: Question has not been resolved yet.

I got this far only by skimming; I didn't have time to review 2/3s of the PR thoroughly.  I will accept that the changes work since the tests pass.

tuxji (commented):
Seems to be getting to a good point for merge, although unfinished parts should be completed first.  I do wish that there was a way to split this PR into a series of several smaller PRs with cohesive changes in each PR since that would make the numerous changes easier to review.  I skimmed 2/3s of the PR's changes.

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
```diff
@@ -90,53 +92,53 @@ private[io] class ByteArrayOrFileOutputStream(
...
               "Unable to create temporary file in %s: %s".format(
                 tempDirPath.getPath,
-                e.getMessage(),
+                e.getMessage,
               ),
             )
         }
       val newStream: java.io.OutputStream = new java.io.FileOutputStream(maybeFile.get)
-      stream.flush
+      stream.flush()
```
pkatlic, line 100: What is the best practice for inclusion of parentheses in parameterless function calls? It seems like we don't include them for most uses (see line 95 above), but is this a special case?

daffodil-io/src/test/scala/org/apache/daffodil/layers/TestLimitingJavaIOStreams.scala
```diff
@@ -162,52 +136,52 @@ ZyBzb2x1dGlvbnMuCg=="""
...
-    val afterBaos = new ByteArrayOutputStream()
-    while ({
-      c = bais.read()
-      c != -1
-    }) {
-      afterBaos.write(c)
-    }
-    afterBaos.close()
-    val actualAfterDelim = new String(afterBaos.toByteArray())
+    val rls = new RegexLimitingInputStream(bais, "\\r\\n(?!(?:\\t|\\ ))", "\r\n", utf8)
```
pkatlic, line 139: Since these pattern strings are used several times here, should they be defined as a constant?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
tuxji: The inclusion of empty parens in function definitions/calls is a semantic signal to all programmers that the function is not a pure function, that is, it performs a side effect or changes mutable state.  If a Scala function is declared/implemented with empty parens, the IDE will warn if calls are made without empty parens.  Likewise, if the function is defined as a parameterless function (without empty parens), the IDE will warn if function calls are made with empty parens.  However, Java doesn't implement these semantics itself, so Scala 2 is very liberal when it comes to mixing parameterless and empty-paren methods and calls.  You can add or remove empty parens in any place and the code will compile/run anyway although other programmers may misunderstand the function's semantics.  I think the Scala 3 compiler has become stricter and will check/enforce these semantics to a greater extent.

daffodil-runtime1-layers/src/test/scala/org/apache/daffodil/layers/runtime1/TestLengthLimitedLineFoldingStreams.scala
```diff
@@ -59,19 +60,9 @@ esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat\r
     //
     // regex is CRLF not followed by a tab or space.
     //
-    val rls = new RegexLimitingStream(bba, "\\r\\n(?!(?:\\t|\\ ))", "\r\n", iso8859)
+    val rls = new RegexLimitingInputStream(bba, "\\r\\n(?!(?:\\t|\\ ))", "\r\n", iso8859)
```
pkatlic, line 63: Similar question as above, should these pattern strings be moved to a common constant?

daffodil-io/src/main/scala/org/apache/daffodil/io/DirectOrBufferedDataOutputStream.scala
mbeckerle: One other wrinkle here is IDEs. My IntelliJ IDEA recommends scala style of including the () on a call if the method returns Unit (or void for Java), and not including the () on the call if the method returns a value. 

So toString is recommended to call without the parens because it returns a value. 
Ditto for getters of any kind. 

Of course for methods that both have side-effects, AND return a value, and have no args. In that case IntelliJ IDEA will recommend removing them, despite the method having side effects (because it really doesn't know about the side-effects, and I disagree.

daffodil-test/src/test/resources/META-INF/services/org.apache.daffodil.io.processors.charset.BitsCharsetDefinition
```diff
@@ -15,4 +15,4 @@
 #
 
 org.apache.daffodil.charsets.BitsCharset_ISO_8859_1_Reverse_Definition
-org.apache.daffodil.charsets.BitsCharsetTest_ISO_8859_13_Definition
+org.apache.daffodil.charsets.BitsCharsetTest_ISO_8859_13_Definition
```
pkatlic, line 18: Looks to be missing a newline at the end of the file.

olabusayoT (approved):
Looks great +1 with a small change

mbeckerle:
Closing. These comments have been fixed/addressed. 

This PR subsumed by 

https://github.com/apache/daffodil/pull/1187 

which was then subsumed by 

https://github.com/apache/daffodil/pull/1191

PR #1179, opened 2024-03-12
Add test showing optional elements cannot have predicates in expressions

stevedlawrence:
Optional elements used to be implemented as arrays with a max size of one. A side effect of this was that accessing these elements in an expression required a predicate (e.g. /path/to/optional[1]).

However, at some point we changed the implementation so optional elements were no longer arrays and were added to the infoset just like scalars. This change also meant they no longer needed, or were allowed to have, predicates in expressions. This adds a test to confirm that behavior.

This also adds a colon to "subset" error messages to make the actual error message more clear.

DAFFODIL-2013


tuxji (approved):
+1

mbeckerle (approved):
+1

PR #1180, opened 2024-03-12
Only allow ambiguous path steps between scalar and optional elements

stevedlawrence:
Scalar and optional elements are both implemented in the infoset exactly the same as DIElements, the only real difference is optional elements might not appear in the infoset. Arrays on the other hand are implemented differently as a DIArray. Because of this, it adds complexity if we allow expression path steps that can ambiguously reference either an array or an optional. We currently allow this, but this complexity leads to bugs and thrown exceptions.

In fact, in most cases ambiguously referencing an array or an optional doesn't even work since arrays need a predicate and optionals do not allow predicates. The only case it could potentially work is as the last step as an argument to a function like fn:count. The added complexity is not worth supporting this edge case.

Instead of allowing this, we should instead only allow ambiguous paths steps to scalar and optional elements. This modifies the path step logic to enforce this. A step must now unambiguously reference an array or it must reference scalars/optionals.

Note that this slightly affects the behavior of the fn:count() function. This function must still reference an array or optional as before, but it can now reference a scalar only if it is an ambiguous path that could also reference an optional. If the argument unambiguously references a scalar, it is an SDE as it was before. It can also no longer ambiguously reference an array or optional, but that was previously broken.

DAFFODIL-2879


stevedlawrence:
@pkatlic, can you take a look at this? This makes a tweak to the changes you made to fn:count

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -975,18 +975,16 @@ sealed abstract class StepExpression(val step: String, val pred: Option[Predicat
...
-          "Path step is ambiguous. It can reference both an array or a non-array element, which is not allowed.\n" +
-            "One of the non-arrays is %s",
-          scalars.head.schemaFileLocation.toString,
-        )
-      case _ => // do nothing
-    }
+    val (arrays, nonArrays) = stepElements.partition { _.isArray }
+    schemaDefinitionWhen(
+      arrays.length > 0 && nonArrays.length > 0,
+      "Path step %s is ambiguous. It can reference both an array or a non-array element, which is not allowed.\n" +
```
tuxji, line 981: I suggest changing `or` to `and` in the error message above, even though it may cause some tests to require changes to their expected output.  When you say "both", that implies "and" rather than "or".

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -11693,11 +11693,19 @@
     <tdml:document>
       <tdml:documentPart type="byte">00 00 00 01</tdml:documentPart>
     </tdml:document>
-    <tdml:errors>
-      <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>ex:foo</tdml:error>
-      <tdml:error>ambiguous</tdml:error>
-    </tdml:errors>
+    <tdml:infoset>
```
tuxji, line 11696: @pkatlic: As @stevedlawrence asked, please check these changes to test results carefully.

tuxji (approved):
+1

I expanded the test results diffs to see the tests' comments, and the changes to the results appear to make sense (scalar & optional from ambiguous to good, array & optional from good to ambiguous).

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
```diff
@@ -5276,9 +5278,7 @@
             <xs:complexType>
               <xs:sequence>
                 <xs:element name="foo" type="xs:int" minOccurs="0" maxOccurs="1" dfdl:representation="binary"/>
-                <!-- throws ClassCastException, possible bug in Daffodil group evaluation -->
-                <!-- <xs:group ref="countGroup"/> -->
-                <xs:element name="count" type="xs:int" dfdl:representation="binary" dfdl:inputValueCalc="{ fn:count(../ex:foo) }"/>
+                <xs:group ref="countGroup"/>
```
pkatlic, line 5281: Yes, this now uses the group reference instead of standalone count element.

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_functions/Functions.tdml
pkatlic: The new test results appear to correct the behavior, with scalarOptional parsing properly and arrayOptional throwing the SDE regarding ambiguous path steps.

pkatlic:
+1

The tests now have the expected results for the scalar/optional and array/optional element groups.

pkatlic (approved):

PR #1182, opened 2024-03-13
Fix mapping from separatorPolicy to separatorSuppressionPolicy

stevedlawrence:
Daffodil allows use of the deprecated separatorPolicy property and internally maps it to the new separatorSuppressionPolicy property. Unfortunately, according to the OGF Errata document our mapping is wrong, which means suppressedAtEndLax behaves like trailingEmptyStrict and suppressedAtEndStrict behaves like trailingEmpty. This is reversed from how it should be. This fixes that mapping so they have the correct behavior.

Deprecation/Compatibility:

The deprecated separatorPolicy property values of "suppressedAtEndLax" and "suppressedAtEndStrict" had incorrect behaviors, with each property behaving like the other. Schemas using one value should be updated to use the other. Alternatively, schemas should be updated to use the new separatorSuppressionPolicy property with the following mappings from the deprecated property:

* required -> never
* suppressed -> anyEmpty
* suppressedAtEndLax -> trailingEmpty
* suppressedAtEndStrict -> trailingEmptyStrict

DAFFODIL-2882

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/schema/annotation/props/ByHandMixins.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/schema/annotation/props/ByHandMixins.scala
@@ -197,8 +197,8 @@ trait SeparatorSuppressionPolicyMixin extends PropertyMixin {
         spString match {
           case "required" => SeparatorSuppressionPolicy.Never
           case "suppressed" => SeparatorSuppressionPolicy.AnyEmpty
-          case "suppressedAtEndStrict" => SeparatorSuppressionPolicy.TrailingEmpty
-          case "suppressedAtEndLax" => SeparatorSuppressionPolicy.TrailingEmptyStrict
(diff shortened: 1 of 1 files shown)


mbeckerle (approved):
+1

tuxji (approved):
+1

Is it concerning that no tests needed any changes after switching these two mappings?

stevedlawrence:
> Is it concerning that no tests needed any changes after switching these two mappings?

We just don't have any tests that use the deprecated `separatorPolicy` property with either `trailingEmptyLax` or `trailingEmptyStrict`, so this part of the mapping was never actually hit. Most of our separator suppression tests use the newer `separatorSuppressionPolicy` property values of `trailingEmpty` and `trailingEmptyStrict` are tested.

We could add some tests to make sure this mapping works correctly, but I'm not sure it's worth it. We know the core functionality works, and by inspection it's pretty clear the mapping is correct now. And this is a deprecated property that should go away eventually anyways.

PR #1185, opened 2024-03-15
Add ability to support or ignore restrictions not allowed by DFDL

stevedlawrence:
A new tunable call invalidRestrictionPolicy is added to control how Daffodil handles restrictions that are not allowed by DFDL. This has values of "error", "ignore", "validate, which tell Daffodil to SDE (current behavior), allow but do not use the restriction for validation, or allow and use it for validation, respecitvely.

This tunable currently only affects the xs:pattern restriction on non-string types. This allows extra validation capabilities using just XML Schema that would otherwise require a different validation mechanism like schematron.

Note that Daffodil's pattern validation implementation already works for non-string types, so relatively few changes are needed to support the "validation" option. The only real change is to move the logic related to this new tunable into the Restriction class, since the patternValues variable in that class is what the SimpleTypeRuntimeData uses for validation.

DAFFODIL-2877


daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
```diff
@@ -811,4 +832,78 @@
...
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:int>201</ex:int>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:validationErrors>
+      <tdml:error>failed facet checks</tdml:error>
+      <tdml:error>pattern</tdml:error>
+      <tdml:error>.*[05]</tdml:error>
+    </tdml:validationErrors>
```
tuxji, line 892: Actually, if you need to double check whether it is necessary to match validation errors, this test is easier to edit than any of the other tests.  The only change you need is to comment out lines 888 to 892 and check if the test still passes (and whether you see any of the validation messages in the test's output).

tuxji (approved):
+1 with thoughts.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
```diff
@@ -811,4 +832,78 @@
...
+    model="intPatternRestriction"
+    config="cfg_invalidRestrictionPolicyIgnore"
+    validation="limited">
+    <tdml:document>201</tdml:document>
+    <tdml:infoset>
+      <tdml:dfdlInfoset>
+        <ex:int>201</ex:int>
+      </tdml:dfdlInfoset>
+    </tdml:infoset>
+    <tdml:validationErrors /> <!-- there should be no validation errors -->
```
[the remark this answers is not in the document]
stevedlawrence: If a test case does not have a `<tdml:validationErrors>` element, then validation errors created by the test are ignored. Validation errors created by the test have no affect on the success/failure of the test.

If a test case has a `<tdml:validationErrors>` element and that element has `<tdml:error>` children, then the test *must* create validation errors and each of the `<tdml:error>` strings *must* appear in those messages. If any do not, the test fails.

If a test case has a `<tdml:validationErrors>` element but that element does not have children, then the test case *must not* create validation errors. If there are any validation errors the test fails.  

In each of these cases, the infoset must still match the expected infoset or the test is a failure, regardless of any validation errors.

So this test technically doesn't need an empty `<tdml:validationErrors />` element to pass. But it does ensure that we are correctly ignoring the pattern restriction and not trying to validate anything. Without it, things could be broken and we could still be validating but not know it (in fact, my first change was broken and wasn't ignoring things and this test caught it).

I'll update to the comment as suggested.

> Also, if you changed the config to validate the pattern and removed line 874 checking there are no validation errors, would the "daffodil test" command print the validation errors in its test output, or would you never know that the test passed despite producing validation errors?

I believe the behavior of "sbt test" and the CLI "daffodil test" is to hide all output unless something fails. So if there were validation errors but no `<tdml:validationErrors />` element in the test case, then the validation error woudl be ignored, the test would be considered a success, and you would never know. I think enabling verbose test output does output the validation errors even on success, so it's not completely hidden, but you do need to do something special to see it.

daffodil-test/src/test/resources/org/apache/daffodil/section00/general/tunables.tdml
stevedlawrence: Mentioned above, but validation errors are ignored if there is no `<tdml:validationErrors>` element, so with those lines removed this test would still pass but would not verify that validation occurred.

mbeckerle (approved):
+1

PR #1188, opened 2024-03-19
Error if -T and -P options are used together in the CLI

stevedlawrence:
When a parser is created with the CLI "save-parser" command, the values of the tunables are serialized along with the parser. When using the "parse", "unparse", or "performance" commands with the -P option to reload a saved parser, we use the serialized tunable values and ignore the -T option if provided. This can be confusing since there is no indication that the -T option is ignored.

To avoid this, we now error if the -P and -T options are used together. Saved parsers must be rebuilt if you want to change tunable values.

DAFFODIL-2880


tuxji (approved):
+1

pkatlic (approved):
+1

PR #1189, opened 2024-03-20
Enable dfdl:extraEscapedCharcaters to take an expression

olabusayoT:
- implement Ev val and class for extraEscapedCharacters similar to  escapeCharacter and escapeEscapeCharacter
- add tests for literal and expression examples

DAFFODIL-2876

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/EvEscapeSchemes.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/EvEscapeSchemes.scala
@@ -41,6 +41,15 @@ class EscapeEscapeCharEv(expr: CompiledExpression[String], ci: DPathCompileInfo)
   with InfosetCachedEvaluatable[String] {
   override lazy val runtimeDependencies = Vector()
 }
+class ExtraEscapedCharsEv(expr: CompiledExpression[String], ci: DPathCompileInfo)
+  extends EvaluatableConvertedExpression[String, Seq[String]](
+    expr,
+    ExtraEscapedCharactersCooker,
+    ci,
+  )
+  with InfosetCachedEvaluatable[Seq[String]] {
+  override lazy val runtimeDependencies = Vector()
+}
 
 trait EscapeSchemeCommonEv {
   def optEscapeEscapeChar: Maybe[EscapeEscapeCharEv]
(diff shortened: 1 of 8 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/EvEscapeSchemes.scala
```diff
@@ -114,7 +123,12 @@ class EscapeSchemeCharUnparseEv(
   def compute(state: ParseOrUnparseState) = {
     val escChar = escapeChar.evaluate(state).charAt(0)
     val optEscEscChar = evalAndConvertEEC(state)
-    new EscapeSchemeCharUnparserHelper(escChar, optEscEscChar, extraEscapedCharsCooked, ci)
+    new EscapeSchemeCharUnparserHelper(
+      escChar,
+      optEscEscChar,
+      extraEscapedCharsCooked(state),
```
stevedlawrence, line 129: Just to make our code more consistent with escEscChar, can you move this up and change the name, something like:

```scala
val extraEscChars = evalAndConvertExtraEscapedChars(state)
new EscpeSchemeCharUnparserHalper(..., extraEscChars, ...)
```

Same with below use.

daffodil-test/src/test/resources/org/apache/daffodil/section07/escapeScheme/escapeScheme.tdml
```diff
@@ -314,6 +314,24 @@
...
+
+    <dfdl:defineEscapeScheme name="fromData5">
+      <dfdl:escapeScheme
+              escapeCharacter="{ xs:string(xs:int(../tns:escapeChar) - 3) }"
+              escapeKind="escapeCharacter"
+              escapeEscapeCharacter="{ xs:string(xs:int(fn:ceiling(xs:float(../tns:escapeEscapeChar)))) }"
+              extraEscapedCharacters="1 3"
+              generateEscapeBlock="whenNeeded" />
+    </dfdl:defineEscapeScheme>
+
```
stevedlawrence, line 334: It might be useful to also have a test that makes sure extraEC works with escapeKind="block". By code inspection it's clear it does, but that would prevent us accidentally breaking it in the future.

stevedlawrence (approved):
+1 :+1: just some very minor suggestions

stevedlawrence:
+1 to latest changes

mbeckerle (approved):
+1

tuxji (approved):
+1

PR #1190, opened 2024-03-22
Remove DAFFODIL_HOME since IDEs don't need it

olabusayoT:
- use "." for daffodilRoot instead

DAFFODIL-2836

### DIFF
--- a/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/Util.scala
+++ b/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/Util.scala
@@ -52,22 +52,20 @@ object Util {
 
   private val isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows")
 
-  private val daffodilRoot = sys.env.getOrElse("DAFFODIL_HOME", ".")
-
   private val daffodilBinPath = {
     val ext = if (isWindows) ".bat" else ""
-    Paths.get(daffodilRoot, s"daffodil-cli/target/universal/stage/bin/daffodil$ext")
+    Paths.get(s"daffodil-cli/target/universal/stage/bin/daffodil$ext")
   }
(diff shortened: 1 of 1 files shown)


daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/Util.scala
```diff
@@ -52,7 +52,7 @@ object Util {
 
   private val isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows")
 
-  private val daffodilRoot = sys.env.getOrElse("DAFFODIL_HOME", ".")
+  private val daffodilRoot = "."
```
stevedlawrence, line 55: Can we delete daffodilRoot and it's uses? Looks like its only used in a couple of places, where a value of "." is essentially a no-op.

stevedlawrence (approved):
+1

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/Util.scala
olabusayoT: Done

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/Util.scala
```diff
@@ -67,7 +65,7 @@ object Util {
    * operating system
    */
   def path(string: String): Path = {
-    Paths.get(daffodilRoot, string)
+    Paths.get(string)
```
stevedlawrence, line 68: Can you remove daffodilRoot from the comment too? Also, while you're in this section can you replace "line separator" with "path separator" in this comment, I think "line" was probably a typo?

mbeckerle (approved):
+1

PR #1191, opened 2024-03-24
New Layer API & Improved Implementation

mbeckerle:
This commit fixes the remaining issues open about Daffodil's Layers extension/plugins. Partial progress was in commit ff9ba4. This finishes that work.

- DAFFODIL-2845 - Layer Improvements (umbrella ticket)
- DAFFODIL-2844 - LayerNotEnoughDataException ...
- DAFFODIL-2825 - Define Supported Layering API
- DAFFODIL-2841 - Layers do not support created SDE or PE/UE Errors
- DAFFODIL-2843 - How to create PE/UE from a layer

The layer API is all defined in Java clases and interfaces now.  One of the built-in layers (Gzip) is written in Java now.

There is only one layer property now: dfdlx:layer which takes a QName of the layer class.

Layer classes are derived from

    org.apache.daffodil.runtime1.layers.api.Layer

A layer a name and target namespace, and hence a QName is required to identify it. This namespace is OWNED by the layer. All DFDL variables defined in that namespace are either used to pass parameters to the layer code, or receive results (such as a checksum) back from the layer code. This is enforced. A layer that has no DFDL variables does not have to define a DFDL schema that defines the layer's target namespace, but any layer that uses DFDL variables *must* define a schema with the layer's namespace as its target namespace, with the variables declared in it (using dfdl:defineVariable).

There is also an abstract base class for defining checksum layers called,

    org.apache.daffodil.runtime1.layers.api.ChecksumLayer

All the layer properties about layer length kinds etc. are gone.

dfdlx:layerLengthKind='boundaryMark' and dfdlx:boundaryMark properties are replaced by a layer named 'boundaryMark'.

dfdlx:layerLengthKind='explicit' and dfdlx:layerLength properties are replaced by a layer named 'fixedLength'.

dfdlx:layerLengthKind='implicit' is replaced by just not using one of the two above layers.

dfdlx:layerTransform is replaced by just dfdlx:layer and is now a QName.  Layers no longer have both transformers and length-limiters mixed together.

There are two ways to do explicit layer limiting, that differ in an important way.

If you use 'implicit' layer length (no restriction layer), and constrain the layer length by a surrounding element of specified length, then the layer length is controlled when parsing, but NOT controlled when unparsing.

If you use the layer "fixedLength", that dictates the parse and unparse length to be the value of the layerLength variable. Similarly if you use a checksum layer built with the ChecksumLayer base class, the length is controlled for both parsing and unparsing as it works like the fixedLength layer.

Layer Variables:

Each variable is either a parameter passed to a special setter named setLayerVariableParameters, or is a return variable which is populated from a special getter. The special setter args and getters taken together must use up all the DFDL variables defined in the layer's namespace. (See the javadoc for the Layer base class.)

A special getter example might be:

   int getLayerVariableResult_checksum() { .... }

This corresponds to, and returns the value that populates, the DFDL variable named 'checksum' in the layer's namespace. That DFDL variable must have a DFDL type corresponding to the int type (xs:unsignedShort or xs:int).

Layers could have multiple result variables, but we have no actual examples of more than one value being returned.  (Returning multiple variables is tested. We just have no real use cases for it at this time.)

Layer Exception Handling:

Added setProcessingErrorException(...) method to Layer

This allows the layer to specify that if the layer throws specific exceptions or runtime exceptions that they are converted into processing errors. This eliminates most need for layers to contain try-catches.

Additional Fixes in this Commit:

Fixed gzip layer which was not constructing on Java 17 or 21 due to static initializer differences of some sort.

Removed restriction that a layer can only have a single term inside it.

Added close() to InputSourceDataInputStream, as we weren't closing these. They now implement java.io.Closeable so can use try-with-resources for auto closing in Java.

There are a large number of simplifications in the "layering" of the various layer-related implementation classes.

LayerRegistry is gone, merged into LayerFactory, which is gone, merged into LayerDriver.

In the examples/tests for layering: Changed check digit bad digit style to construct an always-invalid element instead of an assertion failure.

About Testing:

The testing is quite extensive now. There are tests for every way that a user can goof up the definition of a layer class, and there are tests for processingError, runtimeSchemaDefinitionError, and throwing an Exception from every place a user-defined Layer could cause these. Parse errors cause backtracking in all sensible cases. Nothing aborts.

Protecting every place that layer code is called from any sort of error/throw requires quite a number of try/catch blocks. This may have performance implications.

Add some fuzz testing around gzip layer

GZip layer converts IOExceptions to PE explicitly.

Fuzz tests show that gzip does a good job converting various bad data issues into IOException, so
really that's the only one the GzipLayer has to deal with.

Changed boundaryMarkLayer and gzip layer and AIS (test) layer to use setProcessingErrorException to handle IOExceptions.

(and the test SimpleBombOutLayer.scala used to drive negative tests.)

Added bug-related comments to other files about bugs discovered while fuzz-testing of other non-daffodil layer schemas.

DEPRECATION/COMPATABILITY

This new Layer API is 100% incompatible with schemas or layer code from Daffodil 3.6.0 and all prior versions of Daffodil. The layer feature was just an experimental feature in Daffodil, so we reserved the right to change it.

However, it is our hope that the new Layer API introduced by this commit will prove to be stable and supportable indefinitely.

DAFFODIL-2845, DAFFODIL-2844, DAFFODIL-2825, DAFFODIL-2841, DAFFODIL-2843


olabusayoT (approved):
Looks good! Some minor suggestions. +1

tuxji:
I started reviewing yesterday and got through 33 of the 163 files before resuming my project work. I have only 2 very minor comments so far but I will review the rest of the files over time.  (I just noticed I was reviewing an older PR, will review this PR instead.)

mbeckerle:
> I started reviewing yesterday and got through 33 of the 163 files before resuming my project work. I have only 2 very minor comments so far but I will review the rest of the files over time. (I just noticed I was reviewing an older PR, will review this PR instead.)

Note that both PRs should be identical. I just squashed/rebased this one to get github to re-run testing for me and to get up-to-date codecov results. So you can continue on the other if that's where you started.

daffodil-test/src/test/resources/META-INF/services/org.apache.daffodil.runtime1.layers.api.Layer
```diff
@@ -27,8 +27,31 @@
...
+org.apache.daffodil.runtime1.layers.STL_BadNotALayer # error intentional. This isn't a layer class
+org.apache.daffodil.runtime1.layers.STL_BombOutLayer
+
+
+
+
+
+
+
+
```
tuxji, line 55: Are the empty lines at the end intentional?

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/AISPayloadArmoringLayer.scala
```diff
@@ -183,17 +113,25 @@ class AISPayloadArmoringOutputStream(jos: java.io.OutputStream) extends OutputSt
...
-        IOUtils.write(cb, jos, iso8859)
-        cb.clear()
+      val ba = baos.toByteArray
+      using(InputSourceDataInputStream(ba)) { dis =>
+        val finfo = FormatInfoForAISDecode
+        val cb = CharBuffer.allocate(256)
+        //
+        // TODO: This is not a supportable API. We want to reuse the bitsCharset features of daffodil
+        //  for this non-byte-sized charset decoding. But this finfo object (a trait on the ParseOrUnparseState)
+        //  should not be part of the API
```
tuxji, line 123: Are you talking only about the AIS Payload Armoring Layer API, or about the layer API in general?  I haven't noticed the finfo object used elsewhere in this PR, just in this function.

tuxji (approved):
+1 with a few suggestions

I wonder if layers will become so important and widely used that Daffodil code generators for the smallest possible deterministic subset of DFDL will have to implement layers?  Probably not, but speculate away.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -901,7 +901,7 @@ object Main {
     val BadExternalVariable = Value(33)
     val UserDefinedFunctionError = Value(34)
     val UnableToCreateProcessor = Value(35)
-    val LayerExecutionError = Value(36)
+    val LayerExecutionError = Value(36) // TODO: remove if remains unused.
```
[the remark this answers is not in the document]
stevedlawrence: Maybe fatal errors end up converted into a LayerExecutionError?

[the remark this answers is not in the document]
stevedlawrence: Ah, it looks like fatal layer errors end up with the BugFound exit code. While fatal layer errors are bugs, they are bugs in the layer implementation. It might be better to execit with LayerExecutionError jsut to make it clear that it's not a bug in daffodil but a bug in a layer.

daffodil-test/src/test/resources/META-INF/services/org.apache.daffodil.runtime1.layers.api.Layer
mbeckerle: Inadvertent. It's another IDE setting to always truncate final empty lines.

daffodil-test/src/test/scala/org/apache/daffodil/runtime1/layers/AISPayloadArmoringLayer.scala
mbeckerle: I will clarify. I'm talking about how this AIS layer is using the BitsCharset which is an internal thing (now anyway) so while the layer could be built in to Daffodil, it can't be written by an outsider without depending on Daffodil internals.

mbeckerle:
> +1 with a few suggestions
> 
> I wonder if layers will become so important and widely used that Daffodil code generators for the smallest possible deterministic subset of DFDL will have to implement layers? Probably not, but speculate away.

They will. Checksums and CRCs are important. Decompression?, base64 decoding? Maybe not.

Though important wire formats that uses say, mil-std-2045 headers, have a 2 big field in that header that indicates the body is compressed via one of 3 algorithms, or none.  That said, I've never seen data with that header that uses the compression feature. It's just theoretically possible.

daffodil-runtime1-layers/src/main/scala/org/apache/daffodil/layers/runtime1/FixedLengthLayer.scala
```diff
@@ -0,0 +1,134 @@
...
+    this.fixedLength = fixedLength.toInt
+    Assert.invariant(fixedLength >= 0) // variable is unsignedInt, so this can't be negative
+    if (fixedLength > maxFixedLength)
+      processingError(
+        s"fixedLength value of $fixedLength is above the maximum of $maxFixedLength.",
+      )
+  }
+
+  override def wrapLayerInput(jis: InputStream): InputStream = {
+    new FixedLengthInputStream(jis)
```
stevedlawrence, line 72: Note, I just came across Apache CommonsIO `BoundedInputStream` which does basically the same thing as this but streams data. So it fixes the "make this streaming" TODO, at least for the input part. It think it would mean the `nRead < fixedLength` would need to be done at close though, maybe something like this:

```scala
new BoundedInputStream(jis, fixedLength) {
  override def close(): Unit = {
    if (getCount() < fixedLength) {
      processingError(new LayerNotEnoughDataException(fixedLength, getCount))
    }
    super.close()
  }
}
```

Note that this changes the PE to occur at the end of the layer instead of at the beginning. Could that affect backtracking?

Which raises a question to me, if a layer throws a PE how does backtracking work? Does it backtrack to the very beginning of the layer? Or is the PE just handled by whatever is currently parsing the layer data? Which means a parser could potentially ignore a layer PE by just backtracking a little and keep reading on?

It  feels like if a layer creates a PE, we want to backtrack all the way back to before that layer started, since that implies the data wasn't valid layer data. Is how it works?

mbeckerle: Using the IOUtils library instead of our own - makes sense.  I'll consider changing this or add a ticket to do this as a second change set. 

The second issue is more important. We need some tests to characterize what it means for each place a layer can cause a parse error. E.g., what happens if a parse error occurs on close of the layer stream, what happens if it happens on establishment of the layer stream, what happens if it occurs during access to layer data. 

The error on close issue is snarly for unparsing, but all processing errors during unparsing are fatal, so we only really case that these errors cannot be masked in that case.

For parsing however, we need a failure to establish or to close the stream to cause the parse of the sequence having the layer to fail with a parse error.  I will look into a couple more tests to verify this precisely, or document which tests we have now which verify this. Tests will need to know not just that a parse error occurred, but exactly what construct in the DFDL schema got the parse error. Some of these errors must be parse errors of the layered sequence, not of something inside the layered sequence which could be suppressed by a choice (for example).

mbeckerle: I decided to leave this as is with a TODO that says it could be made streaming but not to change it until/unless there is demand for the feature. Basically too much new testing and debug is required that I don't have time for. 

As for what is the backtrack behavior, keep in mind that the start of the layer is quite possibly NOT a point of uncertainty. There's no notion of going back to the start of the layer separate from if there happens to be a point of uncertainty within the layer right at the start, then backtracking can reset the position of the PState to 0 and that will backup in the layer to position 0 of those bytes. But this is exactly the same thing as regular non-layer data streams. Backtracking inside a  layer is no different from backtracking inside a regular data stream that is associated with finite input such as a file. The layer being too short causing an error is the same as a file being too short. The layer throwing some random exception because the data appears to be corrupted is the same as for example trying to parse a text number and hitting non-digit characters. The data can always look like noise because the parser is speculating down the wrong branch, and that's as true for layer decoders as it is for number parsers. 

If there is no point of uncertainty until somewhere *before* the layer start, then backtracking can reset to before the layer even existed, keep in mind the layer being used at all is part of what is being reset by backtracking, so backtracking to before the layer removes the layer, discarding the objects (which are reclaimed by garbage collection eventually).

PR #1192, opened 2024-03-25
Add support for validation of length facet

olabusayoT:
- add support for length, also set minLength and maxLength with the value of length, when hasLength is true, and those variables are queried
- ensure length and minLength and maxLength facets are not declared together (Xerces Error, but we have asserts incase Xerces validation is turned off)
- add tests for validation=limited and validation=on

DAFFODIL-2842

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
@@ -83,8 +83,8 @@ trait ElementBase
   requiredEvaluationsIfActivated(isSimpleType)
   requiredEvaluationsIfActivated(if (hasPattern) patternValues)
   requiredEvaluationsIfActivated(if (hasEnumeration) enumerationValues)
-  requiredEvaluationsIfActivated(if (hasMinLength) minLength)
-  requiredEvaluationsIfActivated(if (hasMaxLength) maxLength)
+  requiredEvaluationsIfActivated(if (hasMinLength || hasLength) minLength)
+  requiredEvaluationsIfActivated(if (hasMaxLength || hasLength) maxLength)
   requiredEvaluationsIfActivated(if (hasMinInclusive) minInclusive)
   requiredEvaluationsIfActivated(if (hasMaxInclusive) maxInclusive)
   requiredEvaluationsIfActivated(if (hasMinExclusive) minExclusive)
@@ -711,8 +711,8 @@ trait ElementBase
       Assert.invariant(repElement.lengthKind =:= LengthKind.Implicit)
       // it's a string with implicit length. get from facets
       schemaDefinitionUnless(
-        repElement.hasMaxLength,
-        "String with dfdl:lengthKind='implicit' must have an XSD maxLength facet value.",
+        repElement.hasMaxLength || repElement.hasLength,
+        "String with dfdl:lengthKind='implicit' must have a length or maxLength facet value.",
       )
       val ml = repElement.maxLength
       ml.longValue()
@@ -941,8 +941,9 @@ trait ElementBase
 
   private lazy val hasPattern: Boolean = typeDef.optRestriction.exists(_.hasPattern)
   private lazy val hasEnumeration: Boolean = typeDef.optRestriction.exists(_.hasEnumeration)
-  protected lazy val hasMinLength = typeDef.optRestriction.exists(_.hasMinLength)
-  protected lazy val hasMaxLength = typeDef.optRestriction.exists(_.hasMaxLength)
+  protected lazy val hasLength: Boolean = typeDef.optRestriction.exists(_.hasLength)
+  protected lazy val hasMinLength: Boolean = typeDef.optRestriction.exists(_.hasMinLength)
+  protected lazy val hasMaxLength: Boolean = typeDef.optRestriction.exists(_.hasMaxLength)
   private lazy val hasMinInclusive = typeDef.optRestriction.exists(_.hasMinInclusive)
   private lazy val hasMaxInclusive = typeDef.optRestriction.exists(_.hasMaxInclusive)
   private lazy val hasMinExclusive = typeDef.optRestriction.exists(_.hasMinExclusive)
@@ -963,18 +964,19 @@ trait ElementBase
   /**
    * Compute minLength and maxLength together to share error-checking
    * and case dispatch that would otherwise have to be repeated.
+   *
+   * Also set them to the value of length, in the case we've used the length facet
    */
   final lazy val (minLength: java.math.BigDecimal, maxLength: java.math.BigDecimal) =
     computeMinMaxLength
-  // TODO: why are we using java.math.BigDecimal, when scala has a much
-  // nicer decimal class?
+
   private val zeroBD = new java.math.BigDecimal(0)
   private val unbBD = new java.math.BigDecimal(-1) // TODO: should this be a tunable limit?
 
   private def computeMinMaxLength: (java.math.BigDecimal, java.math.BigDecimal) = {
     schemaDefinitionUnless(
       isSimpleType,
-      "Facets minLength and maxLength are allowed only on types string and hexBinary.",
+      "The length facet or minLength/maxLength facets are not allowed on complex types",
     )
     typeDef match {
       case _ if hasRepType => {
@@ -986,7 +988,7 @@ trait ElementBase
         val pt = prim.primType
         schemaDefinitionWhen(
           (pt == PrimType.String || pt == PrimType.HexBinary) && lengthKind == LengthKind.Implicit,
-          "Facets minLength and maxLength must be defined for type %s with lengthKind='implicit'",
+          "The length facet or minLength/maxLength facets must be defined for type %s with lengthKind='implicit'",
           pt.name,
         )
         //
@@ -1001,12 +1003,17 @@ trait ElementBase
         val pt = st.primType
         val typeOK = pt == PrimType.String || pt == PrimType.HexBinary
         schemaDefinitionWhen(
-          !typeOK && (hasMinLength || hasMaxLength),
-          "Facets minLength and maxLength are not allowed on types derived from type %s.\nThey are allowed only on typed derived from string and hexBinary.",
+          !typeOK && (hasLength || hasMinLength || hasMaxLength),
+          "The length facet or minLength/maxLength facets are not allowed on types derived from type %s.\nThey are allowed only on types derived from string and hexBinary.",
           pt.name,
         )
-        val res = (hasMinLength, hasMaxLength, lengthKind) match {
-          case (true, true, LengthKind.Implicit) => {
+        val res = (hasLength, hasMinLength, hasMaxLength, lengthKind) match {
+          case (true, false, false, _) => (r.lengthValue, r.lengthValue)
+          case (true, _, _, _) =>
+            Assert.invariantFailed(
+              "Facet length cannot be defined with minLength and maxLength facets",
+            )
+          case (false, true, true, LengthKind.Implicit) => {
             schemaDefinitionUnless(
               r.minLengthValue.compareTo(r.maxLengthValue) == 0,
               "The minLength and maxLength must be equal for type %s with lengthKind='implicit'. Values were minLength of %s, maxLength of %s.",
@@ -1016,7 +1023,7 @@ trait ElementBase
             )
             (r.minLengthValue, r.maxLengthValue)
           }
-          case (true, true, _) => {
+          case (false, true, true, _) => {
             schemaDefinitionWhen(
               r.minLengthValue.compareTo(r.maxLengthValue) > 0,
               // always true, so we don't bother to specify the type in the message.
@@ -1026,13 +1033,13 @@ trait ElementBase
             )
             (r.minLengthValue, r.maxLengthValue)
           }
-          case (_, _, LengthKind.Implicit) =>
+          case (false, _, _, LengthKind.Implicit) =>
             SDE(
               "When lengthKind='implicit', both minLength and maxLength facets must be specified.",
             )
-          case (false, true, _) => (zeroBD, r.maxLengthValue)
-          case (false, false, _) => (zeroBD, unbBD)
-          case (true, false, _) => (r.minLengthValue, unbBD)
+          case (false, false, true, _) => (zeroBD, r.maxLengthValue)
+          case (false, false, false, _) => (zeroBD, unbBD)
+          case (false, true, false, _) => (r.minLengthValue, unbBD)
           case _ => Assert.impossible()
         }
         res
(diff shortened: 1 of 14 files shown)


mbeckerle (approved):
+1

tuxji (approved):
+1 with a question

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -960,12 +962,73 @@ trait ElementBase
     typeDef.optRestriction.flatMap { _.enumerationValues }
   }
 
+  final lazy val length: java.math.BigDecimal = {
+    Assert.invariant(hasLength)
+    schemaDefinitionUnless(
+      isSimpleType,
+      "Facet length is allowed only on types string and hexBinary.",
```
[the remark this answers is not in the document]
olabusayoT: You're right! I updated the message to reflect the check in both length and computeMinMaxLength (where this was copied from)

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/Facets.scala
```diff
@@ -387,7 +413,7 @@ trait Facets { self: Restriction =>
         errorOnLocalLessThanBaseFacet(theLocalFacet, theRemoteFacet, facetType)
         localFacet
       }
-      case Facet.maxLength | Facet.fractionDigits => {
+      case Facet.length | Facet.maxLength | Facet.fractionDigits => {
         errorOnLocalGreaterThanBaseFacet(theLocalFacet, theRemoteFacet, facetType)
```
stevedlawrence, line 417: This means that a local length facet must be less than the base length facet, is that correct? Have we confirmed with the XSD spec that that is the right behavior? Feels like make a local length should always be the same as a base length, since there is no way to narrow a single value instead of a range like min/maxValue have.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1616,29 +1616,33 @@ trait ElementBaseGrammarMixin
...
         value,
       )
-      if (maxLengthLong != -1 && len > maxLengthLong) warn("max", maxLengthLong)
-      Assert.invariant(minLengthLong >= 0)
-      if (minLengthLong > 0 && len < minLengthLong) warn("min", minLengthLong)
+      if (hasLength && len != lengthLong) warn("l", lengthLong)
+      else if (hasMinLength || hasMaxLength) {
+        if (maxLengthLong != -1 && len > maxLengthLong) warn("maxL", maxLengthLong)
+        Assert.invariant(minLengthLong >= 0)
+        if (minLengthLong > 0 && len < minLengthLong) warn("minL", minLengthLong)
```
stevedlawrence, line 1644: Suggest we just pass in the string "minLength", "maxLength" and "length", the above `%sength` is kindof hard to parse and only saves us a few characters.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/RuntimeData.scala
```diff
@@ -345,7 +346,15 @@ final class SimpleTypeRuntimeData(
         return Error("facet enumeration(s): %s".format(e.enumerationValues.mkString(",")))
       }
     }
-
+    // Check length
+    e.length.foreach { length =>
+      val lenAsLong = length.longValue()
+      val isLengthGreaterThanEqToZero = lenAsLong.compareTo(0L) >= 0
+      if (isLengthGreaterThanEqToZero) {
```
stevedlawrence, line 353: I don't really understand why we do a length >= 0 check. I see it's just copied from min/maxLength, but seems like that should never happen and should be caught during complication? length, minLength, and maxLength can never be negative.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/RuntimeData.scala
```diff
@@ -407,7 +416,32 @@ final class SimpleTypeRuntimeData(
...
+        val dataLen = data.length.toLong
+        val isDataLengthEqual = dataLen.compareTo(lenAsLong) == 0
+        if (isDataLengthEqual) java.lang.Boolean.TRUE
+        else java.lang.Boolean.FALSE
+      }
+      case PrimType.HexBinary => {
+        val data = diNode.dataValue.getByteArray
 
+        val dataLen = data.length.toLong
+        val isDataLengthEqual = dataLen.compareTo(lenAsLong) == 0
```
stevedlawrence, line 438: Should length compare against the actual number of bytes, or the number of chars in the string representation of the infoset?

Note that min/maxLength have the exact same logic, but shouldn't they have less-than or greather-than? I wonder if those are unrelated bugs?

stevedlawrence (approved):
+1 minor suggestions and things to confirm with the XSD spec

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/Facets.scala
stevedlawrence: Found this in the XSD spec:
> Schema Component Constraint: length valid restriction
It is an [·error·](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#dt-error) if [length](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#f-l) is among the members of [{facets}](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#std-facets) of [{base type definition}](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#std-base_type_definition) and [{value}](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#f-l-value) is not equal to the [{value}](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#f-l-value) of the parent [length](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#f-l).

So I think this is wrong. This needs to error if local is not equal to base.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/RuntimeData.scala
mbeckerle: For hexBinary, it's definitely number of bytes, not number of characters. 

For strings,...it's supposed to be characters, but a little DFDL tricky thing is that if dfdl:lengthKind is 'implicit' then the length is in characters, but the charset must be SBCS (single-byte character set) when dfdl:lengthUnits='bytes' so that the maxLength facet, which is used to obtain the length, will be correct for either bytes or characters. Otherwise DFDL's interpretation of the maxLength value could be inconsistent with XSDs interpretation.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/RuntimeData.scala
```diff
@@ -407,7 +416,32 @@ final class SimpleTypeRuntimeData(
...
+        if (isDataLengthEqual) java.lang.Boolean.TRUE
+        else java.lang.Boolean.FALSE
+      }
+      case _ => e.SDE("Length facet is only valid for string and hexBinary.")
```
tuxji, line 442: This line is not covered by tests, codecov says, so an earlier check prevents execution from coming here.  I'd still leave the check in for the fallback "match everything else" case, but put the word "Facet" first and spell length with a lowercase l.

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
```diff
@@ -407,8 +424,33 @@
...
+        </xs:simpleType>
+    </xs:element>
+
+    <xs:element name="e2_8" dfdl:lengthKind="delimited" dfdl:encoding="iso-8859-1">
+        <xs:simpleType>
+            <xs:restriction base="xs:hexBinary">
+                <xs:length value="0"/>
+            </xs:restriction>
+        </xs:simpleType>
+    </xs:element>
```
tuxji, line 451: Please note that e2_8 is a type derived from hexBinary like the others, but the comment on the test case that uses e2_8 says the type "int" which indicates either e2_8 should be a type derived from int or that comment is incorrect.  

Also please note that element names like e2_2, etc., are not well chosen or easily maintainable names.  You can't insert new elements between them easily without having to renumber or tack on new numbers.  In fact, the numbers tacked on the original element names seem like they were intended to provide some information about the facets, not just give some kind of ordering to the elements, see this example below:

```scala
    <xs:element name="e2_2" dfdl:lengthKind="delimited">
      <xs:simpleType>
        <xs:restriction base="ex:stMinLength_1">
          <xs:minLength value="2" />
        </xs:restriction>
      </xs:simpleType>
    </xs:element>
```

A name like `ex:stMinLength_1` is much more clear that it's defining a string with a minLength of 1.  So, these tests would become easier to maintain if they used similarly descriptive names such as `hex_length_0` instead of `e2_8`.

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
```diff
@@ -643,7 +685,63 @@
...
+        <tdml:infoset>
+            <tdml:dfdlInfoset>
+                <ex:e2_7>DEADBEEFDAFADA</ex:e2_7>
+            </tdml:dfdlInfoset>
+        </tdml:infoset>
+    </tdml:parserTestCase>
+
+    <tdml:parserTestCase name="validation_testFacets_03"
+                         root="e2_8" model="TestFacets"
+                         description="checks that int doesn't work with length facet"
```
tuxji, line 724: Either this description or the entire test is incorrect since `e2_8` is really a `hex_length_0`.

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
```diff
@@ -1921,7 +2019,139 @@
...
+    <tdml:parserTestCase name="validation_testFacets2_01"
+                         root="e2_5" model="TestFacets2"
+                         description="checks that length cannot be used with min/maxLength"
+                         validation="limited">
+
+        <tdml:document>
+            <tdml:documentPart type="byte">deadbeef</tdml:documentPart>
+        </tdml:document>
+
+        <!-- this is a Xerces Error -->
```
tuxji, line 2125: Is Xerces used in both validation modes "limited" and "on"??  I thought the point of "limited" validation was to validate faster by using only Daffodil itself, skipping calling Xerces since the user doesn't want to wait for Xerces to validate the infoset.

tuxji (approved):
+1, but could be improved with some changes.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -960,12 +962,73 @@ trait ElementBase
     typeDef.optRestriction.flatMap { _.enumerationValues }
   }
 
+  final lazy val length: java.math.BigDecimal = {
```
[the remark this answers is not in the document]
stevedlawrence: We might not even need length in the tuple? Maybe deep down in this function we can do something like

```scala
if (hasLength)
  (length, length)
else
  (minLength, maxLength)
```

So the fact that length was used kindof disappears, and length just means minLength == maxLength

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -960,12 +962,73 @@ trait ElementBase
...
+  final lazy val (minLength: java.math.BigDecimal, maxLength: java.math.BigDecimal) = {
+    // if hasLength is true, and min/maxLength is queried, set it to length
+    if (hasLength) {
+      (length, length)
+    } else {
+      computeMinMaxLength
+    }
+  }
   // TODO: why are we using java.math.BigDecimal, when scala has a much
   // nicer decimal class?
```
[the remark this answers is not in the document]
stevedlawrence: I think we moved away from Scala types in most cases a while ago. I don't remember why, but for example all of our Infoset primitives just java types.

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
stevedlawrence: I think this comes from validating the DFDL schema, not the infoset. So limited/on/full doesn't affect this.

The Java and Scala API's don't allow turning off validating the DFDL schema. It can be disabled using the internal ProcessorFactory:

https://github.com/apache/daffodil/blob/main/daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala#L64

I *think* the TDML runner can also disable validating dfdl schemas, but that's only done in special cases I think.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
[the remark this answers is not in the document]
mbeckerle: I think this was just avoiding boxes on the boxes on the boxes. Scala's decimal is built on java's decimal so one more layer of object between the actual data and operations on it. We did not strictly speaking need the additional functionality of the scala decimal class, so we decided to go with Java's boxed number types as our representation types in the infoset. I rather wish we had gone with ULong for xs:unsignedLong because it is far more efficient, but this decision is 10 years ago and probably before we learned of the ULong library.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -1026,13 +1034,15 @@ trait ElementBase
...
               "When lengthKind='implicit', both minLength and maxLength facets must be specified.",
             )
-          case (false, true, _) => (zeroBD, r.maxLengthValue)
-          case (false, false, _) => (zeroBD, unbBD)
-          case (true, false, _) => (r.minLengthValue, unbBD)
+          case (false, false, true, _) => (zeroBD, r.maxLengthValue)
+          case (false, false, false, _) => (zeroBD, unbBD)
+          case (false, true, false, _) => (r.minLengthValue, unbBD)
+          case (true, _, _, _) =>
+            SDE("Facet length cannot be defined with minLength and maxLength facets")
```
stevedlawrence, line 1045: This is a minor organizational thing, but it might be helpful to move this case right after the `(true, false,false, _)`case. That way the first two cases are when `hasLength` is true--we either use it or error if either minLength/maxLength is defined. That handles all the cases where hasLength is true, so it's maybe a little more mentally easier to think about the more complex minLength/maxLength cases that follow, since we don't have to consider hasLength anymore.

Also, should this be an assert.invariantFailed, since xerces should fail before we get here if length is defined with a minLength/maxLength?

stevedlawrence (approved):
+1 to latest changes

PR #1193, opened 2024-03-26
Update Apache Daffodil KEYS

pkatlic:


KEYS
```diff
@@ -349,3 +349,62 @@ gDY5Gv25FBHRWluYzimk48n2tLZPk15mxqFS4u1e8CdPdE5Sh5i0wUs1uT/vp4PG
...
+pub   rsa4096 2024-03-25 [SC]
+      20F7E90D97FFA8AE1578143A3352C3E55AB7A478
+uid           [ultimate] Peter Katlic <pkatlic@apache.org>
+sig 3        3352C3E55AB7A478 2024-03-25  [self-signature]
+sub   rsa4096 2024-03-25 [E]
+sig          3352C3E55AB7A478 2024-03-25  [self-signature]
+
+-----BEGIN PGP PUBLIC KEY BLOCK-----
+
+mQINBGYB7TUBEAC8yI6PMaylbVjETvZWJc3YHt0Rx3T8hS9+W3n3+rbU1DVPmF0x
```
stevedlawrence, line 362: Downloading the keys file and running `gpg --list-packets KEYS | grep -E '(user ID|digest algo)'`, it looks like your key is algorithm 8, which is sha256. Although that should work fine, ASF prefers sha-512. It might be best to regenerate your key with sha512 in case apache or anything ever deprecates use of sha256. We had an issue where RPM deprecated sha1 and we had to regenerate some keys, so it's not unheard of.

I think the settings mentioned in "Add some lines to your gpg.conf file:" in this file will cause it to generate sha512.

stevedlawrence (changes requested):

stevedlawrence (approved):
+1, looks good!

stevedlawrence:
We usually require 2 +1's from a committer/PMC member before merging a pull request. This one is small enough that it's not a big deal that you merged it, but for future PR's please wait for 2 +1's.

mbeckerle (commented):
+1

pkatlic:
@stevedlawrence Got it, will make sure to wait for the 2 +1s on future PRs.

PR #1195, opened 2024-03-26
Improve diagnostic from delimiter scanning

mbeckerle:
DAFFODIL-2345

### DIFF
--- a/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
+++ b/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
@@ -151,14 +151,15 @@ class TestCLIParsing {
     }(ExitCode.BadExternalVariable)
   }
 
-  @Test def test_CLI_Parsing_SimpleParse_DFDL1197_fix(): Unit = {
+  @Test
+  def test_CLI_Parsing_SimpleParse_DFDL1197_fix(): Unit = {
     val schema = path(
       "daffodil-test/src/test/resources/org/apache/daffodil/section12/delimiter_properties/testOptionalInfix.dfdl.xsd",
     )
 
     runCLI(args"-vv parse -s $schema") { cli =>
       cli.sendLine("1/3", inputDone = true)
-      cli.expectErr("<Sequence><Separator/><RepMinMax name='s1'>")
+      cli.expectErr("<Sequence><separator/><RepMinMax name='s1'>")
     }(ExitCode.LeftOverData)
   }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
@@ -19,6 +19,7 @@ package org.apache.daffodil.runtime1.processors.dfa
 
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.exceptions.SchemaFileLocation
+import org.apache.daffodil.lib.util.Misc
 import org.apache.daffodil.runtime1.processors.parsers.DelimiterTextType
 
 /**
@@ -133,6 +134,17 @@ trait DFAField extends DFA {
   final override def run(r: Registers): Unit = runLoop(r, DFA.EndOfData, StateKind.EndOfData)
 }
 
+object DFADelimiter {
+  private val controlOrWhitespace = "\\p{C}|\\p{Z}".r
+
+  private def containsCtrlOrWS(s: String) = controlOrWhitespace.findFirstMatchIn(s).isDefined
+
+  def strForDiagnostic(s: String) =
+    if (containsCtrlOrWS(s))
+      s"'$s' ('${Misc.remapStringToVisibleGlyphs(s)}')"
+    else s"'$s'"
+}
+
 trait DFADelimiter extends DFA {
   def delimType: DelimiterTextType.Type
   def lookingFor: String
@@ -147,6 +159,9 @@ trait DFADelimiter extends DFA {
   final val isES = lookingFor == "%ES;"
 
   def unparseValue: String
+
+  lazy val strForDiagnostic: String = DFADelimiter.strForDiagnostic(lookingFor)
+
 }
 
 /**
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
@@ -28,10 +28,14 @@ import org.apache.daffodil.runtime1.processors.LocalTypedDelimiterIterator
 import org.apache.daffodil.runtime1.processors.RemoteTerminatingMarkupAndLocalTypedDelimiterIterator
 import org.apache.daffodil.runtime1.processors.TermRuntimeData
 import org.apache.daffodil.runtime1.processors.dfa.DFADelimiter
+import org.apache.daffodil.runtime1.processors.dfa.ParseResult
 import org.apache.daffodil.runtime1.processors.dfa.TextParser
 
 object DelimiterTextType extends Enum {
-  abstract sealed trait Type extends EnumValueType
+  abstract sealed trait Type extends EnumValueType {
+    override lazy val toString =
+      Misc.initialLowerCase(getClass().getSimpleName.replace("$", ""))
+  }
   case object Initiator extends Type
   case object Separator extends Type
   case object Terminator extends Type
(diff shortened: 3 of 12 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -73,6 +76,35 @@ class DelimiterTextParser(
...
+  private def localDelimiters(state: PState): Seq[DFADelimiter] = {
+    var foundLocalDFAIndex = -1
+    val localIndexStart = state.mpstate.delimitersLocalIndexStack.top
+    val inScopeDelimiters = state.mpstate.delimiters
+    val res = inScopeDelimiters.slice(localIndexStart, inScopeDelimiters.length)
+    res
+  }
+
+  private def didNotFindExpectedDelimiter(foundDelimiter: ParseResult, start: PState): Unit = {
+    val delimString = foundDelimiter.matchedDelimiterValue.get
```
olabusayoT, line 88: This val doesn't appear to be used

olabusayoT (approved):
Looks great! Minor note! +1

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -73,6 +76,35 @@ class DelimiterTextParser(
     foundLocalDFAIndex >= 0
   }
 
+  private def localDelimiters(state: PState): Seq[DFADelimiter] = {
+    var foundLocalDFAIndex = -1
```
pkatlic, line 80: While foundLocalDFAIndex is used in containsLocalMatch above, it is unused here in localDelimiters.

pkatlic (approved):
+1 after checking unused variables.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -73,6 +76,35 @@ class DelimiterTextParser(
...
+    val foundDFA = foundDelimiter.matchedDFAs(0)
+    PE(
+      start,
+      s"""Found enclosing delimiter: %s during scan for local delimiter(s): %s.
+         | The expected delimiter(s) were: %s.
+         | The enclosing delimiter was from %s.
+         |""".stripMargin,
+      foundDFA.strForDiagnostic,
+      localDelims.map { d => d.strForDiagnostic }.mkString(", "),
+      localDelims
```
stevedlawrence, line 99: This duplicates localDelims. THoughts on an error message like

```scala
PE(
  "Found enclosing delimiter %s from %s.
  |Expected delimiter(s) were:
  |%s",
  foundDFA.strForDiagnostic,
  foundDFA.locationDescription,
  localDelims.map { d =>
    s"  ${d.delimType.toString} ${d.strForDiagnostic} from location ${d.location}."
  }
)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
```diff
@@ -73,6 +76,35 @@ class DelimiterTextParser(
...
+        .map { d =>
+          s"  ${d.delimType.toString} ${d.strForDiagnostic} from location ${d.location}."
+        }
+        .mkString("\n", "\n", ""),
+      foundDFA.location,
```
stevedlawrence, line 104: Thoughts on including `locationDescription` either instead of or in addition to `.location`? `.location` just outputs the element name which can sometimes make it hard to figure out where in the schema it actually is. `locationDescription` includes file/line/col info so makes it real easy.

stevedlawrence (approved):
+1, minor suggestions

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
```diff
@@ -133,6 +134,17 @@ trait DFAField extends DFA {
   final override def run(r: Registers): Unit = runLoop(r, DFA.EndOfData, StateKind.EndOfData)
 }
 
+object DFADelimiter {
+  private val controlOrWhitespace = "\\p{C}|\\p{Z}".r
+
+  private def containsCtrlOrWS(s: String) = controlOrWhitespace.findFirstMatchIn(s).isDefined
+
+  def strForDiagnostic(s: String) =
+    if (containsCtrlOrWS(s))
```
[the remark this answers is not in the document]
mbeckerle: My thought was not about performance. This is only invoked when we're about to cause an error though as a PE it could get backtracked, so performance is an issue. This was mostly to avoid showing a second version of the characters which convert whitespace and control characters to visible things but not annoying the user with that if they aren't present.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
```diff
@@ -147,6 +159,9 @@ trait DFADelimiter extends DFA {
   final val isES = lookingFor == "%ES;"
 
   def unparseValue: String
+
+  def strForDiagnostic = DFADelimiter.strForDiagnostic(lookingFor)
```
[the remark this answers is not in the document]
mbeckerle: Control characters like NUL and Ctrl-B, STX/ETX, STT, EOT, ESC, DEL, etc. are also used as delimiters quite a bit. 

ALso, I think you can do this: dfdl:terminator="&#x20; &#x09;" which is a space and a tab character. XML allows it. I am not sure DFDL won't normalize that away and reject the definition, but either way you can have other quasi-whitespace control chars that would be confusing to display.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
mbeckerle: I was thinking getting what was not found and what was found instead onto the first line of the message would be helpful with other details on later lines.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/DelimiterParsers.scala
mbeckerle: Ah. I was wondering why I wasn't getting file/line num info. I though it might just be due to TDML tests with embedded schema.

Will fix.

mbeckerle (commented):
Thanks for the thorough review.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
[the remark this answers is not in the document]
stevedlawrence: Seems odd to use actual control characters instead of character class equivalents, but I guess it's possible?  Should we then change `lookingFor` so that is the remapped version, an remap it a compile time? lookingFor seems to be the diagnostic, or is it use for other things?

Ca we also make this a val or pass it in as a constructor paramter?  We probably shouldn't remap this everytime it ends up in diagnostics, it's normal behavior to find an out of scope delimiter and need to create a PE.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
```diff
@@ -151,7 +151,11 @@ class TestCLIParsing {
     }(ExitCode.BadExternalVariable)
   }
 
-  @Test def test_CLI_Parsing_SimpleParse_DFDL1197_fix(): Unit = {
+  // FIXME FIXME FIXME
```
mbeckerle, line 154: @stevedlawrence  I can't figure out why this test doesn't pass any more. 
The string it is searching for is there in the very verbose output. 
Can you take a quick look at this?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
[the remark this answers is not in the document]
mbeckerle: This has been addressed in latest commit. I elimiinated the duplicate code in Misc for various remappers and they all use sensible non-copying loops now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/dfa/Runtime.scala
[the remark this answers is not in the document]
mbeckerle: This should be resolved. I changed this to lazy val so there is no allocation per use. I did not chase down all the lookingFor to see if they are all just diagnostic display, as there were a couple cases where it wasn't obvious at first glance.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
stevedlawrence: Looks like the error now outputs
```
<Sequence><separator/><RepMinMax name='s1'>
```
Note the lowercase `s` in "separator". I guess changing the toString of the enums modified one of the parser names?

mbeckerle: Did not spot that. I guess the searches I was doing default to case insensitive. 

Thanks!

PR #1196, opened 2024-03-27
Add support for passing exception directly into DaffodilUnhandledSAXException

olabusayoT:
- add constructors for throwable only and message only
- add unit tests that create the DaffodilUnhandledSAXException and compares them to the message, the throwable or both

DAFFODIL-2433

### DIFF
--- a/daffodil-core/src/test/scala/org/apache/daffodil/core/processor/TestSAXUnparseAPI.scala
+++ b/daffodil-core/src/test/scala/org/apache/daffodil/core/processor/TestSAXUnparseAPI.scala
@@ -23,8 +23,10 @@ import java.io.ByteArrayOutputStream
 import org.apache.daffodil.lib.Implicits.intercept
 import org.apache.daffodil.lib.xml.DaffodilSAXParserFactory
 import org.apache.daffodil.lib.xml.XMLUtils
+import org.apache.daffodil.runtime1.api.DFDL.DaffodilUnhandledSAXException
 
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.xml.sax.InputSource
@@ -200,4 +202,37 @@ class TestSAXUnparseAPI {
     assertTrue(m.contains("prior to end"))
     assertTrue(m.contains("{http://example.com}list"))
   }
+
+  @Test def testDaffodilUnhandledSAXException_creation_bothMessageAndCause(): Unit = {
+    val message = "Error Message"
+    val expectedException = new IllegalArgumentException("Illegal Argument Message")
+    val actualException = new DaffodilUnhandledSAXException(message, expectedException)
+    assertEquals(message, actualException.getMessage)
+    assertEquals(expectedException, actualException.getCause)
+  }
+
+  @Test def testDaffodilUnhandledSAXException_creation_onlyMessage(): Unit = {
+    val message = "Error Message"
+    val actualException = new DaffodilUnhandledSAXException(message)
+    assertEquals(message, actualException.getMessage)
+    assertNull(actualException.getCause)
+  }
+
+  @Test def testDaffodilUnhandledSAXException_creation_onlyCause(): Unit = {
+    val expectedException = new IllegalArgumentException("Illegal Argument Message")
+    val actualException = new DaffodilUnhandledSAXException(expectedException)
+    // when the detailMessage is null as is the case when no message is passed in,
+    // getMessage returns the detailMessage from the embedded exception
+    assertEquals(expectedException.getMessage, actualException.getMessage)
+    assertEquals(expectedException, actualException.getCause)
+  }
+  @Test def testDaffodilUnhandledSAXException_creation_onlyCauseNoCauseMessage(): Unit = {
+    val expectedException = new IllegalArgumentException()
+    val actualException = new DaffodilUnhandledSAXException(expectedException)
+    // when the detailMessage is null as is the case when no message is passed in,
+    // getMessage returns the detailMessage from the embedded exception
+    assertNull(actualException.getMessage)
+    assertEquals(expectedException, actualException.getCause)
+  }
+
 }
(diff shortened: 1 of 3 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLParserUnparser.scala
```diff
@@ -236,8 +236,12 @@ object DFDL {
    * Thrown by the DaffodilUnparseConentHandler when an unexpected error
    * occurs, this usually represents a bug in Daffodil
    */
-  class DaffodilUnhandledSAXException(description: String, cause: Exception)
-    extends SAXException(description, cause)
+  class DaffodilUnhandledSAXException(description: String, cause: Throwable)
+    extends SAXException(description, new Exception(cause)) {
```
mbeckerle, line 240: Why the Exception wrap? Is that because SAXException will only accept an Exception as the cause argument? 

I would suggest that unless we need our constructors to accept Throwable, we should also accept only Exception and then we wouldn't need to wrap Exception to satisfy SAXException's constructor.

mbeckerle (changes requested):

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLParserUnparser.scala
tuxji: Correct, SAXException accepts only an Exception as its cause argument.  Agreed, DaffodilUnhandledSAXException should accept only Exception and refrain from wrapping the cause twice.

daffodil-core/src/test/scala/org/apache/daffodil/core/processor/TestSAXUnparseAPI.scala
```diff
@@ -200,4 +202,34 @@ class TestSAXUnparseAPI {
...
   }
+
+  @Test def testDaffodilUnhandledSAXException_creation_bothMessageAndCause(): Unit = {
+    val message = "Error Message"
+    val thr = new IllegalArgumentException("Illegal Argument Message")
+    val e = new DaffodilUnhandledSAXException(message, thr)
+    assertEquals(message, e.getMessage)
+    // we call getCause twice here, because the cause is wrapped in an Exception before
+    // being returned
+    assertEquals(thr, e.getCause.getCause)
```
tuxji, line 213: All these tests will need to be changed to call getCause only once.

tuxji (changes requested):
Looks good except for wrapping the exception twice.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLParserUnparser.scala
```diff
@@ -236,8 +236,12 @@ object DFDL {
...
+    def this(msg: String) = this(msg, null)
+
+    def this(cause: Throwable) = this(null, cause)
```
mbeckerle, line 243: I recently discovered (today) that our SLF4J logging system doesn't properly deal with constructors of Exceptions that take only a cause. If such a class gets to top level you get "[error] null" where the null is because no message string was supplied. So I recently changed some exception constructors to do this:

    def this(cause: Throwable) = 
      this({
                 if (cause.getMessage() == null)
                    Misc.getMessageFromClass(cause)
                 else
                     cause.getMessage()
             },
             cause)

This ensures if the exception object is passed to SLF4J's error method that at least the class name of the cause shows up, and if the cause has a message, that is used rather than 'null'. 

This still doesn't deal with a bunch of chained cause exceptions where one of them eventually has a message, or perhaps none of them do, but it avoids the SLF4J 'null' output.

stevedlawrence: That seems so messy, are we just using/accessing exceptions wrong?

olabusayoT: I don't know the use case where we'd have an exception with no message, but isn't null the expected thing if you have no message passed in to the exception and no cause messaged passed in with the cause? As in the case below?

```scala
  @Test def testDaffodilUnhandledSAXException_creation_onlyCauseNoCauseMessage(): Unit = {
    val expectedException = new IllegalArgumentException()
    val actualException = new DaffodilUnhandledSAXException(expectedException)
    // when the detailMessage is null as is the case when no message is passed in,
    // getMessage returns the detailMessage from the embedded exception
    assertNull(actualException.getMessage)
    assertEquals(expectedException, actualException.getCause)
  }
```

tuxji (approved):
+1

stevedlawrence (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/api/DFDLParserUnparser.scala
mbeckerle: I think Java exceptions with messages and causes do not have an agreed/ideal usage pattern. 

I've looked and not found it. 

The encapsulation pattern: where you wrap your own exception class around an underlying library's thrown exception - is a useful pattern, but not one that is necessarily fully supported by other software. In particular, if you use this pattern, and don't provide any message text, because well, the class name itself says it all, then you find some things, like SLF4J, make no effort to create any useful message from your classes. 

Some frameworks/systems are used in a way where exception messages are needed, other situations causes are used, but there seem to be many contexts where things assume messages are present when they are not. 

There are also goofy things that happen with cause chains. I've seen circular cause chains, for example, where the cause of an exception points back to itself. Anything that chases the cause chain needs to cope with this. Loops in this chain are clearly meaningless mistakes. It's possible for the same exception class to appear in the chain as two different instances, but not a cycle.  

To me it's very desirable to have a reliable single call which always creates a explanatory string from any throwable regardless of whether any given encapsulation or exception class in the chain provides a message, has a cycle, etc. Often the class names of the exceptions in the chain are the important information .  

Unfortunately, that isn't what getMessage() does. Many things that handle exceptions appear to believe getMessage has that contract, when it really is just a getter on a message if provided at construction time. 

And ... you override getMessage() at your peril. Voice of experience as you can imagine. It is very easy to get stack overflows if you override getMessage() and anything goes wrong. E.g., overriding get message to call super.getMessage, and if null look at the cause.... turns out to be a very bad idea. 

I think our Misc.getSomeMessage(e) is supposed to be the thing that always gives you *some* message from an exception - even if it's just the class name. I don't think it deals with chained exceptions quite right though. 

But SLF4J doesn't call our Misc.getSomeMessage(e) of course. It seems to just call the getMessage() getter and ends up ignoring the cause object.

PR #1198, opened 2024-03-28
Allow backtracking when parsing prefix length fails with prefixIncludesPrefixLength

stevedlawrence:
> [!NOTE]  
> This might want to wait until 3.8.0 or a later release where we can fix all the prefix length issues 

If parsing a prefix length fails, we just pretend the value was zero and rely on callers to check the processor status and ignore the zero value if it failed. However, even if the parse fails we still do the negative prefix length SDE checks. And if the prefix has an adjustment due to the prefixIncludesPrefixLength property, that zero is a adjusted to become negative and causes an SDE instead of a PE.

Really, we shouldn't do these SDE checks at all if we failed to parse a prefix length since the zero is meaningless and should be ignored. This rearranges the logic so we only check for negative prefix values if the prefix parse actually succeeded.

This also skips calculating contentLength/valueLength if the parse fails.

DAFFODIL-2688

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
@@ -82,31 +82,34 @@ trait PrefixedLengthParserMixin {
 
     state.infoset = plElement
 
-    val parsedLen: JLong =
-      try {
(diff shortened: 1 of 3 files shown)


tuxji (approved):
+1

I'm fine with this change going in immediately, unless it's non backward compatible.

stevedlawrence:
This isn't a backwards compatibility issue, and it is a fairly minor bug. I have no strong preference whether it goes in 3.7 or the next.

PR #1199, opened 2024-03-28
Add support for dfdl:checkRangeInclusive/checkRangeExclusive functions

olabusayoT:
- add implementation for checkRangeInclusive/Exclusive in Expressions and DFDLFunctions
- add tests for checkRangeInclusive/Exclusive for numeric types and non-numeric types

DAFFODIL-1515

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
@@ -2016,6 +2016,12 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
       case (RefQName(_, "checkConstraints", DFDL), args) => {
         DFDLCheckConstraintsExpr(functionQNameString, functionQName, args)
       }
+      case (RefQName(_, "checkRangeInclusive", DFDL), args) => {
+        DFDLCheckRangeInclusiveExpr(functionQNameString, functionQName, args)
+      }
+      case (RefQName(_, "checkRangeExclusive", DFDL), args) => {
+        DFDLCheckRangeExclusiveExpr(functionQNameString, functionQName, args)
+      }
       case (RefQName(_, "decodeDFDLEntities", DFDL), args) => {
         FNOneArgExpr(
           functionQNameString,
@@ -3039,6 +3045,101 @@ case class DFDLCheckConstraintsExpr(
 
   override lazy val inherentType = NodeInfo.Boolean
 
+}
+
+case class DFDLCheckRangeInclusiveExpr(
+  nameAsParsed: String,
+  fnQName: RefQName,
+  args: List[Expression],
+) extends FunctionCallBase(nameAsParsed, fnQName, args) {
+
+  lazy val List(arg1, arg2, arg3) = { checkArgCount(3); args }
+
+  override lazy val children = args
+
+  override lazy val compiledDPath = {
+    checkArgCount(3)
+    val argDPath = args(0).compiledDPath
+    val rangeFrom = args(1).compiledDPath
+    val rangeTo = args(2).compiledDPath
+    val c = conversions
+    val res = new CompiledDPath(DFDLCheckRangeInclusive(argDPath, rangeFrom, rangeTo) +: c)
+    res
+  }
+  override def targetTypeForSubexpression(subexpr: Expression): NodeInfo.Kind = {
+    val targetType = (arg1.inherentType, arg2.inherentType, arg3.inherentType) match {
+      case (
+            testType: NodeInfo.Numeric.Kind,
+            minType: NodeInfo.Numeric.Kind,
+            maxType: NodeInfo.Numeric.Kind,
+          ) => {
+        val rangeType = NodeInfoUtils.generalizeArgTypesForComparisonOp(
+          "checkRangeInclusive",
+          minType,
+          maxType,
+        )
+        val targetType = NodeInfoUtils.generalizeArgTypesForComparisonOp(
+          "checkRangeInclusive",
+          testType,
+          rangeType,
+        )
+        targetType
+      }
+      case (test, min, max) =>
+        SDE(s"Cannot call $nameAsParsed with non-numeric types: $test, $min, $max")
+    }
+    targetType
+  }
+
+  override lazy val inherentType: NodeInfo.Kind = NodeInfo.Boolean
+
+}
+
+case class DFDLCheckRangeExclusiveExpr(
+  nameAsParsed: String,
+  fnQName: RefQName,
+  args: List[Expression],
+) extends FunctionCallBase(nameAsParsed, fnQName, args) {
+
+  lazy val List(arg1, arg2, arg3) = { checkArgCount(3); args }
+
+  override lazy val children = args
+
+  override lazy val compiledDPath = {
+    checkArgCount(3)
+    val argDPath = args(0).compiledDPath
+    val rangeFrom = args(1).compiledDPath
+    val rangeTo = args(2).compiledDPath
+    val c = conversions
+    val res = new CompiledDPath(DFDLCheckRangeExclusive(argDPath, rangeFrom, rangeTo) +: c)
+    res
+  }
+  override def targetTypeForSubexpression(subexpr: Expression): NodeInfo.Kind = {
+    val targetType = (arg1.inherentType, arg2.inherentType, arg3.inherentType) match {
+      case (
+            testType: NodeInfo.Numeric.Kind,
+            minType: NodeInfo.Numeric.Kind,
+            maxType: NodeInfo.Numeric.Kind,
+          ) => {
+        val rangeType = NodeInfoUtils.generalizeArgTypesForComparisonOp(
+          "checkRangeExclusive",
+          minType,
+          maxType,
+        )
+        val targetType = NodeInfoUtils.generalizeArgTypesForComparisonOp(
+          "checkRangeExclusive",
+          testType,
+          rangeType,
+        )
+        targetType
+      }
+      case (test, min, max) =>
+        SDE(s"Cannot call $nameAsParsed with non-numeric types: $test, $min, $max")
+    }
+    targetType
+  }
+
+  override lazy val inherentType: NodeInfo.Kind = NodeInfo.Boolean
 }
 
 /**
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
@@ -17,10 +17,13 @@
 
 package org.apache.daffodil.runtime1.dpath
 
+import java.math.BigInteger
+
 import org.apache.daffodil.lib.calendar.DFDLCalendarConversion
 import org.apache.daffodil.lib.cookers.EntityReplacer
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.runtime1.dsom.SchemaDefinitionError
+import org.apache.daffodil.runtime1.infoset.DataValue
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValueBool
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValuePrimitive
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValueShort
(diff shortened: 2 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
```diff
@@ -39,6 +42,94 @@ case class DFDLCheckConstraints(recipe: CompiledDPath) extends RecipeOpWithSubRe
...
+  override def run(dstate: DState): Unit = {
+    val saved = dstate.currentNode
+    dataRecipe.run(dstate)
+    val dataVal = dstate.currentValue
+    dstate.setCurrentNode(saved)
+    minRecipe.run(dstate)
+    val minVal = dstate.currentValue
+    dstate.setCurrentNode(saved)
+    maxRecipe.run(dstate)
+    val maxVal = dstate.currentValue
```
tuxji, line 103: I'm not familiar with evalution of expressions.  Why does this code call dstate.setCurrentNode(saved) after both dataRecipe.run(state) and minRecipe.run(dstate), but not after maxRecipe.run(dstate)?  Why doesn't dstate need to have its saved state restored after the last maxRecipe,run(dstate)?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
```diff
@@ -39,6 +42,94 @@ case class DFDLCheckConstraints(recipe: CompiledDPath) extends RecipeOpWithSubRe
...
+
+  def executeCheck(
+    dataVal: DataValue.DataValuePrimitiveNullable,
+    minVal: DataValue.DataValuePrimitiveNullable,
+    maxVal: DataValue.DataValuePrimitiveNullable,
+  ): Boolean = {
+    (dataVal.value, minVal.value, maxVal.value) match {
+      case (data: Integer, min: Integer, max: Integer) => data >= min && data < max
+      case (data: java.lang.Double, min: java.lang.Double, max: java.lang.Double) =>
+        data >= min && data < max
```
tuxji, line 121: This range check looks like it's still doing an inclusive comparison on the min side and doing an exclusive comparison only on the max side.  Is that how the DFDL function "checkRangeExclusive" is supposed to work?

daffodil-test/src/test/resources/org/apache/daffodil/section23/dfdl_expressions/expressions.tdml
```diff
@@ -7684,4 +7684,217 @@ blastoff
...
+<!--      <tdml:dfdlInfoset>-->
+<!--        <expMinMax_mixedIntAndFloat>-->
+<!--          <data>1</data>-->
+<!--          <min>1.9</min>-->
+<!--          <max>2.5</max>-->
+<!--          <isInRangeInc>false</isInRangeInc>-->
+<!--          <isInRangeExc>false</isInRangeExc>-->
+<!--        </expMinMax_mixedIntAndFloat>-->
+<!--      </tdml:dfdlInfoset>-->
+<!--    </tdml:infoset>-->
```
tuxji, line 7897: Commented out code shouldn't be left around for a long time.

tuxji (commented):
Some questions first...

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
stevedlawrence: The `currentNode` DState can be used by a `run` implementation to know where to evaluate a relative path expression from (e.g. `../foo/bar/baz`).

But calling `run` can also potentially mutate `currentNode`, changing it to something else, which can interfere with other calls to run, so it needs to be reset.

For example, say we had something like this:

```
dfdl:checkRangeExclusive(../foo/bar/baz, ../uno/dos/tres, ../one/two/three)`
```

Before we call `run()`, let's say `currentNode` points to an element called `root`.

After we call `dataRecipe.run()` which evaluates `../foo/bar/baz` relative to `root`, the `currentNode` will point the `baz` element, because DState was mutated. If we do not reset `currentNode` back to `root`, then when we call `minRecipe.run()` to evaluate `../uno/dos/tres` that relative path will be evaluated relative to the `baz` element instead of the `root` element. So that's why we need to save and restore curentNode after each call to run.

Note that we don't have to restore `currentNode` after calling `maxReipce.run()` because we aren't personally calling anymore runs, so whatever `currentNode` ended up as doesn't matter to us. If some other expression was called that led to this run() being called, then they were responsible for saving and restoring currentNode if they needed it.

stevedlawrence (changes requested):
Small suggestions to reduce duplication.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
```diff
@@ -39,6 +42,94 @@ case class DFDLCheckConstraints(recipe: CompiledDPath) extends RecipeOpWithSubRe
...
+    (dataVal.value, minVal.value, maxVal.value) match {
+      case (data: Integer, min: Integer, max: Integer) => data >= min && data <= max
+      case (data: java.lang.Double, min: java.lang.Double, max: java.lang.Double) =>
+        data >= min && data <= max
+      case (data: java.lang.Float, min: java.lang.Float, max: java.lang.Float) =>
+        data >= min && data <= max
+      case (data: java.math.BigDecimal, min: java.math.BigDecimal, max: java.math.BigDecimal) =>
+        data.compareTo(min) >= 0 && data.compareTo(max) <= 0
+      case (data: BigInteger, min: BigInteger, max: BigInteger) =>
+        data.compareTo(min) >= 0 && data.compareTo(max) <= 0
```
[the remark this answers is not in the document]
stevedlawrence: An improvement, maybe `forType` can be implemented as a `Map` for constant lookup and reduced allocations? e.g.

```scala
object ComparisonOps {
  val forType: Map[NodeInfo.Kind, (CompareOpBase, CompareOpBase, CompareOpBase, CompareOpBase, CompareOpBase, CompareOpBase)] =
    Map(
       PrimType.Boolean -> (EQ_Compare, NE_Compare, LT_Boolean, LE_Boolean, GT_Boolean, GE_Boolean),
       PrimType.Integer -> (EQ_Compare, NE_Compare, LT_Integer, LE_Integer, GT_Integer, GE_Integer),
       ...
    )
}
```

The use is still the same, e.g. `ComparisonOps.forType(somePrimType)`

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
stevedlawrence: I don't remember, but maybe we do it to mimic PState, which is mutable? Back in the early days PState wasn't mutable and we just copied it each time something changed, but they overhead of copies started becoming very noticeable so we switched to a mutable PState.

DState might be small enough that the overhead of creating multiple DStates might not be noticeable? Hard to say with out making the change, and it would probably be a pretty significant change--all the run() functions would need to be changed to allocate a new DState. I would be much cleaner and less prone to error though.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
olabusayoT: There wasn't any information on the DFDL spec about how to treat exclusive, and the redmine link on the tickets leads to a 404 error and attempts to find the issue referenced were not successful so I used the mathematical definition of exclusive, which deals with just the upper value (as seen in this SO post: https://stackoverflow.com/questions/39010041/what-is-the-meaning-of-exclusive-and-inclusive-when-describing-number-ranges)

stevedlawrence: Here's the latest DFDL spec: https://github.com/OpenGridForum/DFDL/blob/master/docs/current/gwdrp-dfdl-v1.0.8-GFD-R-P.240.pdf

Page 184 talks about these functions, but doesn't clarify what it means by inclusive/exclusive.

An alternative interpretation is that the min/max values are treated similar to minInclusive, maxInclusive, minExclusive, and maxExclusive facet restrictions, which is more like what tuxji (and me) were expecting. But your reasoning makes sense too.

@mbeckerle, do you know the intention of these functions?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -3041,6 +3047,101 @@ case class DFDLCheckConstraintsExpr(
...
+
+  override lazy val children = args
+
+  override lazy val compiledDPath = {
+    checkArgCount(3)
+    val argDPath = args(0).compiledDPath
+    val rangeFrom = args(1).compiledDPath
+    val rangeTo = args(2).compiledDPath
+    val c = conversions
+    val res = new CompiledDPath(DFDLCheckRangeExclusive(argDPath, rangeFrom, rangeTo) +: c)
```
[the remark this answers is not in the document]
olabusayoT: Just a note that the above code is run treats the minVal as inclusive, or exclusive depending on the function called, answering the question within this comment: https://github.com/apache/daffodil/pull/1199#discussion_r1544650251

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/DFDLFunctions.scala
mbeckerle: Inclusive means both ends are part of the range, Exclusive means both ends are outside the range. 

This is only seems ambiguous in English as in "1 to 10 inclusive" where it becomes a question as to whether you mean inclusive to apply to "1 to 10", or just "10".  But as "inclusive" is an adjective that describes a characteristic of an interval/range, not a single number. So I claim this is not ambiguous. It applies to both ends of the range, always. It requires more words to talk about the mixed combinations of including and excluding. 

In the DFDL functions, both ends of the ranges are definitely what was intended.

stevedlawrence (approved):
+1 :+1: just a couple small suggestion

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/ComparisonOps.scala
```diff
@@ -17,10 +17,42 @@
...
 package org.apache.daffodil.runtime1.dpath
 
+import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.util.Numbers._
+import org.apache.daffodil.runtime1.dpath.NodeInfo.PrimType
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValueBool
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValuePrimitive
 
+object ComparisonOps {
+  lazy val forType: Map[
```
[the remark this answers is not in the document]
stevedlawrence: Another option might be to create a new case class and make that the value of the map, e.g.: 

```scala
case class ComparisonOps(eq: ComparisonOp, ne: ComparisonOp, lt: ComparisonOp, ...) 

...

Map(
  PrimType.Boolean -> ComparisonOps(EQ_Compare, NE_Compare, LT_Boolean, ...),
  ...
)
```

Then uses become something like:
```scala
val compOps = ComparisonOps.forType(targetTypeForSubExpr)
val compare = if (isExclusive) compOps.lt else compOps.le
```
It's a little more verbose, but using it is a bit more clear and less prone to errors since it's impossible for callers of `forType` to get the order of the 6-tuple wrong.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -258,116 +258,21 @@ case class ComparisonExpression(op: String, adds: List[Expression])
...
+      case ">=" => subsetError("Unsupported operation '%s'. Use 'ge' instead.", op)
+      case "=" => subsetError("Unsupported operation '%s'. Use 'eq' instead.", op)
+      case "!=" => subsetError("Unsupported operation '%s'. Use 'ne' instead.", op)
+      case "eq" => eq
+      case "ne" => ne
+      case "lt" => lt
+      case "gt" => gt
+      case "le" => le
+      case "ge" => ge
       case _ => subsetError("Unsupported operation '%s' on type %s.", op, convergedArgType)
```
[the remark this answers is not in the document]
olabusayoT: Created https://issues.apache.org/jira/browse/DAFFODIL-2886

Though I think for ease of creating TDML tests, we should leave the error as a subsetError that results in a SDE rather than an Assert.invariantFailed. I can easily see someone creating a schema that tries to compare hexBinary elems and an SDE would be a better error for them to get until this is implemented.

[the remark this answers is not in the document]
stevedlawrence: > we should leave the error as a subsetError that results in a SDE rather than an Assert.invariantFailed

Yeah, the way you made the change this should stay a subset error. In my head the change would be something like this:

```scala
if (convergedArgType == PrimType.HexBinary && (op == "lt || op == "gt" || op == "le" || op == ge")) {
  subsetError(...)
}
op match {
  ...
}
```
So the hexBinary test would be outside the match case. If it were done that way the default case would want to be an Assert.invariant. But they way did in the match case you're right that it should be a subset error. And your way is probably more scala-y so is better than what I was thinking. And your new hexBinary tests confirm its doing the right thing :+1:

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/ComparisonOps.scala
```diff
@@ -17,10 +17,176 @@
...
 import org.apache.daffodil.runtime1.infoset.DataValue.DataValuePrimitive
 
+/**
+ * Case class used for ordering the return of the appropriate comparison operations for each primitive kind
+ * @param eq represents the equality comparison
+ * @param ne represents the unequality comparison
+ * @param lt represents the less than comparison
+ * @param le represents the less than and equal to comparison
+ * @param gt represents the greater than comparison
+ * @param ge represents the greater than and equal to comparison
```
tuxji, line 33: greater than or equal to

PR #1204, opened 2024-04-03
Add ability to specify errors that should not occur

olabusayoT:
- using warnings/errors match attribute, specify whether all/any/none of the messages should match
- add verifyAnyOfDiagnosticsFound and verifyNoneOfDiagnosticsFound for any/none match attributes
- add tdml tests

DAFFODIL-864

### DIFF
--- a/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
+++ b/daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
@@ -75,6 +75,7 @@
       <attribute name="defaultValidation" type="tns:validationType" use="optional"/>
       <attribute name="defaultConfig" type="xs:string" use="optional"/>
       <attribute name="defaultImplementations" type="tns:implementationsType" use="optional"/>
+      <attribute name="defaultIgnoreUnexpectedWarnings" type="xs:boolean" use="optional"/>
     </complexType>
     <unique name="unique-parserTestCase-name">
       <selector xpath="parserTestCase"/>
@@ -184,6 +185,7 @@
     <attribute name="unsupported" type="xs:boolean" use="optional" default="false"/>
     <attribute name="validation" type="tns:validationType" use="optional"/>
     <attribute name="implementations" type="tns:implementationsType" use="optional"/>
+    <attribute name="ignoreUnexpectedWarnings" type="xs:boolean" use="optional"/>
 
   </attributeGroup>
 
@@ -348,7 +350,7 @@
       <xs:simpleType>
         <xs:restriction base="xs:string">
           <xs:enumeration value="all"/>
-          <xs:enumeration value="any"/>
+          <xs:enumeration value="none"/>
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -73,11 +73,11 @@ import org.apache.daffodil.lib.util.Misc.uriToDiagnosticFile
 import org.apache.daffodil.lib.util.SchemaUtils
 import org.apache.daffodil.lib.xml.DaffodilXMLLoader
 import org.apache.daffodil.lib.xml.XMLUtils
+import org.apache.daffodil.tdml.DiagnosticType.DiagnosticType
 import org.apache.daffodil.tdml.processor.AbstractTDMLDFDLProcessorFactory
 import org.apache.daffodil.tdml.processor.TDML
 import org.apache.daffodil.tdml.processor.TDMLDFDLProcessor
 import org.apache.daffodil.tdml.processor.TDMLParseResult
-import org.apache.daffodil.tdml.processor.TDMLResult
 import org.apache.daffodil.tdml.processor.TDMLUnparseResult
 
 import org.apache.commons.io.IOUtils
@@ -169,6 +169,9 @@ private[tdml] object DFDLTestSuite {
  *
  * shouldDoWarningComparisonOnCrossTests controls whether test warning messages  are compared
  * during cross testing.
+ *
+ * defaultIgnoreUnexpectedWarningsDefault specifies whether a test should ignore unexpected warnings (i.e it creates
+ * warnings, but there is no tdml:warnings elements)
  */
 
 class DFDLTestSuite private[tdml] (
@@ -181,7 +184,8 @@ class DFDLTestSuite private[tdml] (
   val defaultValidationDefault: String,
   val defaultImplementationsDefault: Seq[String],
   val shouldDoErrorComparisonOnCrossTests: Boolean,
-  val shouldDoWarningComparisonOnCrossTests: Boolean
+  val shouldDoWarningComparisonOnCrossTests: Boolean,
+  val defaultIgnoreUnexpectedWarningsDefault: Boolean
 ) {
 
   val TMP_DIR = System.getProperty("java.io.tmpdir", ".")
@@ -415,6 +419,10 @@ class DFDLTestSuite private[tdml] (
       str.split("""\s+""").toSeq
     }
   }
+  lazy val defaultIgnoreUnexpectedWarnings = {
+    val str = (ts \ "@defaultIgnoreUnexpectedWarnings").text
+    if (str == "") defaultIgnoreUnexpectedWarningsDefault else str.toBoolean
+  }
 
   lazy val embeddedSchemas = {
     val res = (ts \ "defineSchema").map { node => DefinedSchema(node, this) }
@@ -610,6 +618,7 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
 
   lazy val defaultRoundTrip: RoundTrip = parent.defaultRoundTrip
   lazy val defaultValidationMode: ValidationMode.Type = parent.defaultValidationMode
+  lazy val defaultIgnoreUnexpectedWarnings: Boolean = parent.defaultIgnoreUnexpectedWarnings
 
   private lazy val defaultImplementations: Seq[String] = parent.defaultImplementations
   private lazy val tcImplementations = (testCaseXML \ "@implementations").text
@@ -634,15 +643,20 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   lazy val optExpectedOrInputInfoset = (testCaseXML \ "infoset").headOption.map { node =>
     new Infoset(node, this)
   }
-  lazy val optExpectedErrors: Option[ExpectedErrors] = (testCaseXML \ "errors").headOption.map {
-    node => ExpectedErrors(node, this)
+  lazy val optExpectedErrors: Option[Seq[ExpectedErrors]] = {
+    val s = (testCaseXML \ "errors").map { node => ExpectedErrors(node, this) }
+    if (s.nonEmpty) Some(s) else None
   }
-  lazy val optExpectedWarnings: Option[ExpectedWarnings] =
-    (testCaseXML \ "warnings").headOption.map { node => ExpectedWarnings(node, this) }
-  lazy val optExpectedValidationErrors: Option[ExpectedValidationErrors] =
-    (testCaseXML \ "validationErrors").headOption.map { node =>
+  lazy val optExpectedWarnings: Option[Seq[ExpectedWarnings]] = {
+    val s = (testCaseXML \ "warnings").map { node => ExpectedWarnings(node, this) }
+    if (s.nonEmpty) Some(s) else None
+  }
+  lazy val optExpectedValidationErrors: Option[Seq[ExpectedValidationErrors]] = {
+    val s = (testCaseXML \ "validationErrors").map { node =>
       ExpectedValidationErrors(node, this)
     }
+    if (s.nonEmpty) Some(s) else None
+  }
 
   val tcName = (testCaseXML \ "@name").text
   lazy val tcID = (testCaseXML \ "@ID").text
@@ -712,6 +726,11 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   lazy val tcRoundTrip: String = (testCaseXML \ "@roundTrip").text
   lazy val roundTrip: RoundTrip =
     if (tcRoundTrip == "") defaultRoundTrip else DFDLTestSuite.standardizeRoundTrip(tcRoundTrip)
+  lazy val tcIgnoreUnexpectedWarnings: String = (testCaseXML \ "@ignoreUnexpectedWarnings").text
+  lazy val ignoreUnexpectedWarnings: Boolean =
+    if (tcIgnoreUnexpectedWarnings == "") defaultIgnoreUnexpectedWarnings
+    else tcIgnoreUnexpectedWarnings.toBoolean
+
   lazy val description = (testCaseXML \ "@description").text
   lazy val unsupported = (testCaseXML \ "@unsupported").text match {
     case "true" => true
@@ -724,17 +743,26 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   }
 
   lazy val shouldValidate = validationMode != ValidationMode.Off
+  lazy val ignoreUnexpectedValidationErrors = if (optExpectedValidationErrors.isDefined) {
+    optExpectedValidationErrors.get.exists(
+      _.hasDiagnostics
+    ) // ignore unexpected validation errors will be false if <tdml:validationErrors /> or true
+    // if there are children under validationErrors
+  } else {
+    true // ignore unexpected validation errors if there is no validationErrors element
+  }
   lazy val expectsValidationError =
-    if (optExpectedValidationErrors.isDefined) optExpectedValidationErrors.get.hasDiagnostics
+    if (optExpectedValidationErrors.isDefined)
+      optExpectedValidationErrors.get.exists(_.hasDiagnostics)
     else false
 
   protected def runProcessor(
     compileResult: TDML.CompileResult,
     expectedData: Option[InputStream],
     nBits: Option[Long],
-    errors: Option[ExpectedErrors],
-    warnings: Option[ExpectedWarnings],
-    validationErrors: Option[ExpectedValidationErrors],
+    errors: Option[Seq[ExpectedErrors]],
+    warnings: Option[Seq[ExpectedWarnings]],
+    validationErrors: Option[Seq[ExpectedValidationErrors]],
     validationMode: ValidationMode.Type,
     roundTrip: RoundTrip,
     implString: Option[String]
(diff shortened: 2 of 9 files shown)


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1844,12 +1944,99 @@ object VerifyTestCase {
...
+    }
+    if (wasFound) {
+      throw TDMLException(
+        s"""Found one of diagnostic ${expectedDiags.head.diagnosticType} message """" +
+          expectedDiagMsgs.mkString("\n") +
+          """" in some of the actual diagnostic messages: """ + "\n" +
+          actualDiagMsgs.mkString("\n"),
+        implString,
+      )
+    }
```
stevedlawrence, line 2039: It might be useful to reduce duplication by combining these functions into a single one that accepts a mode. It feels like are the core they are pretty similar just with some minor differences. For example, you could maybe do something like:

```scala
val diagsFound = expectedDiags.map { diag => 
  val wasFound = actualDiags.contains(diag)
  (diag, wasFound)
}

matchAttr match {
  case "all" if (!diagsFound.forall(_._2)) => {
     throw TDMLException(...)
  }
  case "any" if (!diagsFound.exists(_._2)) => {
     throw TDMLException(...)
  }
  case "none" if (diagsFound.exists(_._2)) => {
    throw TDMLException(...)
  }
}
```

So the logic to determine which diags are found is exactly the same with out copy/paste, and we create a Seq that determines which diags are found/not found. Then a match/case does the logic to figure out which of those were failures depending on the match mode and to create the appropriate exception.

daffodil-lib/src/main/resources/org/apache/daffodil/xsd/tdml.xsd
```diff
@@ -349,6 +349,7 @@
         <xs:restriction base="xs:string">
           <xs:enumeration value="all"/>
           <xs:enumeration value="any"/>
+          <xs:enumeration value="none"/>
```
stevedlawrence, line 352: I didn't even know this "match" attribute existed. I guess we never implemented it?

I'm wonder if it would be useful to have something a big more complex? For example, I could imagine a test might want to do something like diags "a", "b", and "c", must all appear, one of "m" or "n" must apear, and none of "x", "y", and "z" must appear. Maybe we need to allow muptilpe <tdml:error> tags? Then this would be described like

```xml
<parserTestCase ...>
  ... 
  <errors match="all">
    <error>a</error>
    <error>b</error>
    <error>c</error>
  </errors>
  <errors match="any">
    <error>m</error>
    <error>n</error>
  </errors>
  <errors match="none">
    <error>x</error>
    <error>y</error>
    <error>z</error>
  </errors>
</parserTestCase>
```
Maybe that's unneeded complexity though?

daffodil-tdml-processor/src/test/resources/test/tdml/testTDMLErrorsWarningsMatchAttribute.tdml
```diff
@@ -0,0 +1,150 @@
...
+        <A>A</A>
+      </ex:causesValidationError>
+    </tdml:dfdlInfoset>
+  </tdml:infoset>
+
+  <tdml:validationErrors match="any">
+    <tdml:error>value 'A' of element 'A' is not valid</tdml:error>
+    <tdml:error>WILL NOT MATCH ON ANY</tdml:error>
+  </tdml:validationErrors>
+</tdml:parserTestCase>
```
stevedlawrence, line 148: Does this support a way to say that no diagnostics of a type should exist? Or maybe a way to say there must be warnings or errors, but that we don't necessarily care what they are?

Maybe an empty tag with match="all" means an diag type must exist, but we don't care about the message.
```xml
<tdml:errors match="all" />
```

And a empty tag with match="none" means a diag type must not exist:
```xml
<tdml:warnings match="none" />
```

I'm not sure if that logic makes sense or if it should be reversed, but something along those lines might be useful.

mbeckerle (changes requested):
Needs a way to specify no warnings from a test or a test suite. 

Needs to cover errors (match all and match none) as well as warnings (match all and match none). I don't think match 'any' is needed.

mbeckerle (approved):
+1

A separate task once this is committed is to go through the Daffodil test suites and convert them all to default that no-warnings are allowed, adding expected warnings, or configs with warning suppression, to each test that needs it. 

We should avoid adding lots of expected warnings to tests where the warning is not related to the test. It's better to suppress those warnings via a tweak to the test schema, or adding a suppression config for that class of warning in that case. 

I think this will end up being a fairly big job however, and I'm guessing done incrementally, not all at once.

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
```diff
@@ -1902,7 +1902,7 @@
         
       <!-- Validation error for branch1:e1  -->
       <tdml:error>Validation Error</tdml:error>
-      <tdml:error>nonNegA failed facet checks due to: facet minInclusive (30)</tdml:error>
+      <tdml:error>e1 failed facet checks due to: facet minInclusive (30)</tdml:error>
```
stevedlawrence, line 1905: This is unexpected, any idea what cause this change in message?

stevedlawrence (commented):
Looks good, especially since TDML runner has really grown very complex, handful of questions about how things work that are probably worth a discussion since there are clearly many different ways expected results can be interpreted.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1330,6 +1415,14 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
       case NoRoundTrip | OnePassRoundTrip => {
+        val allDiags = firstParseResult.getDiagnostics ++ compileWarnings
+        checkDiagnosticMessages(
+          allDiags,
+          Seq.empty,
+          optWarnings = warnings,
+          optValidationErrors = validationErrors,
+          implString = implString)
+
         verifyParseResults(compileWarnings, firstParseResult, testInfoset, implString)
```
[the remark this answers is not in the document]
olabusayoT: You're right...It looks like we only need it for onePass, since it's the only one that doesn't call verifyParseResults.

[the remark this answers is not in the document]
olabusayoT: Ooops never mind, missed the additional matchCase

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -634,15 +642,33 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   lazy val optExpectedOrInputInfoset = (testCaseXML \ "infoset").headOption.map { node =>
     new Infoset(node, this)
   }
-  lazy val optExpectedErrors: Option[ExpectedErrors] = (testCaseXML \ "errors").headOption.map {
-    node => ExpectedErrors(node, this)
+  lazy val optExpectedErrors: Option[Seq[ExpectedErrors]] = {
+    val e = (testCaseXML \ "errors" \\ "error")
```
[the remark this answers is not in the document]
olabusayoT: I'm of the mind to have them all behave consistently. So if we have ` <tdml:validationErrors />` mean that there should be no validation errors, it should be the same for `<tdml:errors />` and `<tdml:warnings />`. But if we have that functionality, then I don't think we need the noWarnings/noValidationErrors attributes. Thoughts?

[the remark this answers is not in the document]
stevedlawrence: The problem if we didn't support `noWarnings` is it would be a huge effort to add `<tdml:warnings />` to all tests to ensure warnings are never  ignored. We eventually want get to the point where no tests have unexpected warnings and your new property is definitely the preferred way to enforce that globally over `<tdml:warnings />` which can't be added globally.

If we do want consistency with validation, we may want to consider deprecating `<tdml:validation />` and replacing it with a property similar to the noWarnings one . Just like with warnings, we probably want to get the the point where there are no unexpected validation errors, and it's much easier to enforce that by flipping a default flag than adding `<tdml:validaiton />` to all the tests. If we do this, I'd suggest it be a separate PR since there's backwards compatibility issues to consider.

Also, I don't think it's worth the added complexity right now to support something like `<tdml:errors />` means "some errors exist but I don't care about the message". That might be useful at some point, but that could also be a separate PR if we want it. The key thing we really need is the noWarnings capability.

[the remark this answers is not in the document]
stevedlawrence: Some more thoughts. I would suggest we make it so `<tdml:warnings />` and `<tdml:errors />` is not allowed--each one must have child elements. That keeps things simple an intuitive. We don't really need a way to say "there must be an error/warning, but I don't care what it is"--that just isn't really that necessary and adds complication. 

For now, we still allow `<tdml:validation />` for backwards compatibility, but we should eventually plan to deprecate that in favor of a new property similar to the "no warnings" property. And when we add that property, we can support backwards compatibility by making it so `<tdml:validationErrors />` enables the "noValidationErrors" flag, but is otherwise ignored. Eventually we can drop that backwards compatibility and make `<tdml:validationErrors />` invalid to match `<tdml:warnings />` and <tdml:errors />`. These changes to how validationErrors works can all be done in future PRs though. It's fine if there is a slight inconsistency for now.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -959,32 +988,49 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
...
     if (
       !isCrossTest(implString.get) ||
       parent.shouldDoWarningComparisonOnCrossTests
-    )
-      VerifyTestCase.verifyAllDiagnosticsFound(
+    ) {
+      lazy val warningDiagnostics = diagnostics.filterNot(d => d.isError || d.isValidation)
+      lazy val nonNoneAttribWarnings = optWarnings.filter {w => !w.forall(_.matchAttrib.contains("none")) }.getOrElse(Nil)
+      if (noWarnings && warningDiagnostics.nonEmpty) {
+        throw TDMLException(s"Test should not create warnings, but created the following: ${warningDiagnostics.mkString("\n")}",
```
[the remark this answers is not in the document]
olabusayoT: I updated the logic and renamed noWarnings to noWarningsAllowed

daffodil-test/src/test/resources/org/apache/daffodil/section02/validation_errors/Validation.tdml
stevedlawrence: I just noticed that this test expects `<tdml:errors>`. Does it even make sense to expect validation errors if we don't even get an infoset? I wonder if this was just previously ignored since the tests didn't expect success? Should the expected validationErrors just be removed here, and in general tdml:validationErrors + tdml:errors in the same test should maybe be an TDMLException? We can open a bug for that if that makes sense.

olabusayoT: Good point, I did just add validationErrors to checkDiagnostics, which explains why the test just now failed.

I think it makes sense to make validationErrors errors in the same test a TDMLException in a future PR

olabusayoT: For the time being, I'll remeove the expected validation error

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1806,68 +1881,101 @@ object VerifyTestCase {
    * @param expectedDiags                             Expected diagnostics from test cases
    * @param implString                                Implementation string
    */
-  def verifyAllDiagnosticsFound(
+  def verifyDiagnosticsFound(
     actualDiags: Seq[Diagnostic],
-    expectedDiags: Option[ErrorWarningBase],
+    expectedDiags: Option[Seq[ErrorWarningBase]],
+    ignoreUnexpectedWarnings: Boolean,
```
stevedlawrence, line 1887: Can this parameter be changed to a more generic `ignoreUnexpectedDiags: Boolean`? For example, when the `expectedDiags` are `ExpectedErrors` we would always pass in false (we never ignore unexpected errors). When `expectedDiags` are `ExpectedWarnings` we would pass in the value of `ignoreUnexpectedWarnings`. That way this function doesn't need to have have specific logic for warnings (or validationErrors when we add ignoreUnexpectedValidationErrors). It just has something like:

```scala
val hasUnexpectedDiags = actualDiagsFiltered.isNonEmpty && expectedDiags.isEmpty
if (hasUnexpectedDiags && !ignoreUnexpectedDiags)
  throw new TDMLExecption(found unexpected diags)
```

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1806,68 +1881,101 @@ object VerifyTestCase {
...
     implString: Option[String]
-  ) = {
+  ): Option[Seq[Unit]] = {
```
stevedlawrence, line 1889: Can this just return a `Unit`? I think this function either throws an exception if there are problems or returns nothing?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1806,68 +1881,101 @@ object VerifyTestCase {
...
+    }
+    val actualDiagsFiltered = if (expectedDiagnosticType == "validation error") {
+      actualDiags.filter(diag => diag.isValidation)
+    } else if (expectedDiagnosticType == "warning") {
+      actualDiags.filterNot(d => d.isValidation || d.isError)
+    } else if (expectedDiagnosticType == "error") {
+      actualDiags.filter(_.isError)
+    } else {
+      Seq.empty
+    }
```
stevedlawrence, line 1910: I general prefer to avoid comparing strings, they tend to be fragile. For example, if we one day decide to capitalize these this stuff will break. Instead, we can use Scala type system, e.g.

```
val expectedDiagType = expectedDiags.headOption
val actualDiagsFiltered = expectedDiagType match {
  case Some(_: ExpectedValidationError) => actualDiags.filter(diag => diag.isValidation)
  case Some(_: ExpectedWarning) => actualDiags.filterNot(d => d.isValidation || d.isError)
  case Some(_: ExpectedError) => actualDiags.filter(_.isError)
  case None => Seq.empty
}
```
That said, it might be useful to pass in an enum or a filter function or something as a paramter, or do the filtering in checkDiagnosticMessages (assuming that can be updated so everyone calls it). Something like that might be preferred, since this ends up with `Seq.empty` if we don't know the type, which makes it look like there are no actual diagnostics to check.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1806,68 +1881,101 @@ object VerifyTestCase {
...
+      !actualDiags.forall(d => d.isValidation || d.isError)
+      && expectedDiags.isEmpty
+      && !ignoreUnexpectedWarnings
+    ) {
+      throw TDMLException(
+        s"ignoreUnexpectedWarnings = false and test does not expect warnings, but created the following: " +
+          s"${actualDiags.filterNot(d => d.isValidation || d.isError).mkString("\n")}",
+        implString
+      )
+    }
```
stevedlawrence, line 1943: Yeah, here it might be nice to make this a more generic check.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1806,68 +1881,101 @@ object VerifyTestCase {
...
+      diags.map { diag =>
+        val matchAttr = diag.matchAttrib
+        val diagsFound = diag.messages.map { d =>
+          // make case insensitive
+          val wasFound =
+            actualDiagsFilteredMessages.exists(ad => ad.toLowerCase.contains(d.toLowerCase))
+          (d, wasFound)
+        }
+        matchAttr match {
+          case "all" if (!diagsFound.forall(_._2)) => {
```
stevedlawrence, line 1955: Technically, you probably want the condition inside the case block. For example, currently if matchAttr is "all" but the condition fails then it will do the "none" and efault cases, which shouldn't happen. And that way, the default case implies the user had something like `match="foo"` and it should be an error.

stevedlawrence (commented):
Looks good, just and handful of thoughts for more possible cleanup so this complex test runner isn't so complex.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -959,32 +990,45 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
 
   protected def checkDiagnosticMessages(
     diagnostics: Seq[Diagnostic],
-    errors: ExpectedErrors,
-    optWarnings: Option[ExpectedWarnings],
+    errors: Seq[ExpectedErrors],
```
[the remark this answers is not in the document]
olabusayoT: Yes, making that change will allow us to call checkDiagnosticMessages everywhere we currently call verifyDiagnosticsFound. We would just pass in None for what we don't wish to check.

Ex to check only warnings or only validationErrors, we do
```
        checkDiagnosticMessages(
          allDiags,
          None,
          optWarnings,
          None,
          implString
        )
    //or 
        checkDiagnosticMessages(
          actual.getDiagnostics,
          None,
          None,
          optValidationErrors,
          implString
        )
    
```

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
olabusayoT: actualDiagsFiltered will be empty if expectedDiags.isEmpty...since it uses the type of the expectDiags for filtering, but I think we can just change that to actualDiags and the logic works.

Unless we want actualDiagsFiltered to return unfiltered actualDiags when we have no expected diags, thoughts on that?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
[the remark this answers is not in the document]
stevedlawrence: Seems reasonable. But are there cases where we don't what to check all expected diagnostics? Seems like during each test we do a parse/unparse/whatever and then check that all actual diagnostics match the expected diagnostics.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
stevedlawrence: If we use actualDiags then I think it will error if we have any diagnostics, not just unexpected diagnostics of the right type. Feels like we should pass in an an enum that says which diag we are checking, or pass in a filter function that is passed in that is used to pick wich diagnostics to check. Then we don't have to rely on the expected diags to know which actual diags to compare against.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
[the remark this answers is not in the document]
olabusayoT: I think so, in verifyParseResult for example, it looks like we check validation errors separate from warnings

```
(shouldValidate, expectsValidationError) match {
      case (_, true) => {
        // Note that even if shouldValidate is false, we still need to check
        // for validation diagnostics because failed assertions with
        // failureType="recoverableError" are treated as validation errors
        checkDiagnosticMessages(
          actual.getDiagnostics,
          None,
          None,
          optExpectedValidationErrors,
          implString
        )
        Assert.invariant(actual.isValidationError)
      }
      case (true, false) => {
        VerifyTestCase.verifyNoValidationErrorsFound(
          actual,
          implString
        ) // Verify no validation errors from parser
        Assert.invariant(!actual.isValidationError)
      }
      case (false, false) => // Nothing to do here.
    }

    val allDiags = compileWarnings ++ actual.getDiagnostics
    checkDiagnosticMessages(
      allDiags,
      None,
      optExpectedWarnings,
      None,
      implString
    )
```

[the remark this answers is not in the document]
stevedlawrence: I imagine some of that can be refactored and simplified now that you've cleaned up much of this and made things more generic and combined functionality. Seems all three of those checks can be done with one call to checkDiagnosticMessages.

I guess we do still need `None` sometimes, since some functions (e.g. runParseExpectSuccess) don't have access to certain diagnostics because they expect success and really do need to pass in None. I think maybe the separate expectSuccess/expectFailure funtions are complicating things a bit, but that clean up can be done for another day.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
olabusayoT: Hmm you're right...ok, will add enum called diagnosticType

stevedlawrence: Another alternative, if you update everything to calls checkDiagnostisMesssages you could do the filter in there, since each call to verifyDiagnosticFound knows which type it's verifying.

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -1199,37 +1259,26 @@ case class ParserTestCase(ptc: NodeSeq, parentArg: DFDLTestSuite)
...
-        // for validation diagnostics because failed assertions with
-        // failureType="recoverableError" are treated as validation errors
-        VerifyTestCase.verifyAllDiagnosticsFound(
-          actual.getDiagnostics,
-          optExpectedValidationErrors,
-          implString
-        ) // verify all validation errors were found
-        Assert.invariant(actual.isValidationError)
-      }
       case (true, false) => {
```
stevedlawrence, line 1262: Do we need this case anymore? You've updated checkDiagnosicts to check validation errors, including whether or not to ignore unexpected errors. I think this match case goes away completely and we just use the default case? And verifyNoValidatioNErrorsFound goes away?

stevedlawrence (approved):
+1 looks great, this is a big improvement and I think will help alot to make the test runner less complex.

Just one more place where maybe something might not be needed anymore, but I'm not positive.

PR #1210, opened 2024-04-10
Pin libmxml to 3.3.1 in macOS CI

stevedlawrence:
Homewbrew upated to 4.x and currently fails CI

DAFFODIL-2889

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,10 @@ jobs:
 
       - name: Install Dependencies (macOS)
         if: runner.os == 'macOS'
-        run: brew install libmxml
+        run: |
+          brew tap-new local/libmxml
+          brew extract --version 3.3.1 libmxml local/libmxml
+          brew install libmxml@3.3.1
 
       - name: Install Dependencies (Linux)
         if: runner.os == 'Linux'


.github/workflows/main.yml
```diff
@@ -102,7 +102,10 @@ jobs:
 
       - name: Install Dependencies (macOS)
         if: runner.os == 'macOS'
-        run: brew install libmxml
+        run: |
+          brew tap-new local/libmxml
+          brew extract --version 3.3.1 libmxml local/libmxml
+          brew install libmxml@3.3.1
```
tuxji, line 108: I'm curious why the first two commands are needed.  Are they always required even though the page I saw didn't mention them, or is there something about the GitHub runner which makes them necessary?  I have not set up a programming environment on macOS so I don't know much about brew.

tuxji (approved):
+1

.github/workflows/main.yml
stevedlawrence: Yeah, I found that the normal @xyz thing only works if the homebrew-core repo includes multiple versions. Homebrew only seems to do this for popular packages where people might need multiple versions, like node.js. For less popular packages, like libmxml, they only support the latest version. Seems to be a pretty basic package manager.

So essentially what this does is it creates a local repo (`brew tap-new`), finds and extracts 3.3.1 out of the git history of homebrew-core into that local repo (`brew extract`), and then installs that local version.

This is the only way I could find to do it. It seems kindof developer-y/experimental, so it feels like it we might want it to just be a short term solution. In the long term, we probably need to figure out why 4.x doesn't work and see if we can support that. But this at least allows our CI to pass.

stevedlawrence: Note, I also don't know much about macOS and homebrew. This was all just based on googling and lots of trial and error until I found something that worked. There might be a better way to do it.

pkatlic: It does seem that this is the recommended approach to specifying an older version due to the way that homebrew is managed.

PR #1214, opened 2024-04-15
Set SOURCE_DATE_EPOCH for build reproducability when creating release candidates

stevedlawrence:
A common source of differences in our release artifacts are embedded timestamps. The latest version of sbt-native-packager, used to create these artifacts now supports the SOURCE_DATE_EPOCH environment variable which provides control over these embedded timestamps.

When building the release candidate, we now set the SOURCE_DATE_EPOCH to the UNIX timestamp of when the commit being built was merged (i.e. the git "committer" time), allowing for reproducible builds. There are some caveats:

* The MSI installer includes one UUID and timestamp that cannot be changed. Fortunately, msidiff shows this is the only difference so is straightforward to verify
* The RPM created by the release candidate script embeds a GPG signature which has a timestamp of when the signature was created, which cannot be changed. To verify RPM reproducibility, the signature must removed with rpmsign --delsign
* Zip files still include a timestamp in an extended header. I'll report to sbt-native-packager and see if a fix can be included in the next release.

DAFFODIL-2890

### DIFF
--- a/containers/release-candidate/src/daffodil-release-candidate
+++ b/containers/release-candidate/src/daffodil-release-candidate
@@ -234,6 +234,8 @@ DAFFODIL_TUTORIALS_DIR=$DAFFODIL_SITE_DIR/site/tutorials
 DAFFODIL_DIST_DIR=$REPO_ROOT/$DAFFODIL_DIST
 DAFFODIL_RELEASE_DIR=$DAFFODIL_DIST_DIR/$VERSION-$PRE_RELEASE
 
+export SOURCE_DATE_EPOCH=$(git show --no-patch --format=%ct HEAD)
+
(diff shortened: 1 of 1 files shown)


tuxji (commented):
+1

> The RPM created by the release candidate script embeds a GPG signature which has a timestamp of when the signature was created, which cannot be changed. To verify RPM reproducibility, the signature must removed with rpmsign --delsign

Should the PR make the release candidate script call rpmsign --delsign?  On second thought, only the verification itself (which is outside the scope of the release script) needs to call rpmsign --delsign on both rpms to obtain two rpms which can be compared bit-for-bit.

stevedlawrence:
Right, we only want to remove the signature if we're checking for reproducibility. Otherwise we want to keep it so that users installing via the jfrog yum repo will be able to verify the RPM signature.

pkatlic (approved):
+1

PR #1218, opened 2024-04-17
Always create an unparse error when a required array is missing

stevedlawrence:
Currently, if a required array in an unseparated sequence is missing from an infoset then we create an unparse error. However, if that required but missing array is followed by another array in the infoset then we abort. This is incorrect--we should create an unparse error just like other cases.

To fix this, this modifies the unseparated sequence unparser logic so that we first check if this event is even for this array. If it is, then it must be an start event and we perform the normal logic. If it isn't, then it means we have zero occurrences of this array and the event must have been the start event for a following element or the end event for a parent element. Either way, we check if zero occurrences is valid for this array and create an unparse error if not. With this refactored logic, there is now no difference between a missing array followed by a scalar, optional, or another array.

DAFFODIL-2891


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/UnseparatedSequenceUnparsers.scala
```diff
@@ -107,78 +106,58 @@ class OrderedUnseparatedSequenceUnparser(
...
-                }
-
-                unparser.checkFinalOccursCountBetweenMinAndMaxOccurs(
-                  state,
-                  unparser,
-                  numOccurrences,
-                  maxReps,
-                  state.arrayIterationPos - 1,
-                )
-                unparser.endArrayOrOptional(erd, state)
```
stevedlawrence, line 147: Note that this logic is all unchanged, this is just an indentation change. It might be helpful to hide whitespace when reviewing this.

tuxji (approved):
+1

pkatlic (approved):
+1

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/UnseparatedSequenceUnparsers.scala
```diff
@@ -107,78 +106,58 @@ class OrderedUnseparatedSequenceUnparser(
...
+              maxReps,
+              state.arrayIterationPos - 1,
+            )
+            unparser.endArrayOrOptional(erd, state)
           } else {
-            // no event (state.inspect returned false)
-            Assert.invariantFailed("No event for unparing.")
+            // this is either a start event for a following element or an end event for a
+            // parent element. Either way, we never saw a start event for this array/optional,
+            // which means there were zero occurrenes. Make sure that is valid for this array
```
mbeckerle, line 153: The concept of 'valid' here vs. 'well formed' gives me some concern. 
When dfdl:occursCountKind='parsed', then any number of occurrences, including zero, is 'well formed', just not valid. 
Further, when unparsing, if the number of occurrences is < minOccurs we're supposed to check if the element is defaultable (rules are quite complex for this, esp. for an element of complex type which is why this is not yet implemented I believe.)

I believe the change here is an improvement to the code (avoids an abort at least). 

I'm just trying to make sure the comment isn't misleading in this quite complex area where we are knowingly partial in what we've implemented of DFDL defaulting behavior. 

Maybe just change the "Make sure that is valid for this array" to "Make sure that is **_allowed_** for this array" ?

I also feel like we should somehow comment this to be revisited when defaulting during unparsing is considered.

PR #1219, opened 2024-04-18
Use an empty sequence for the xml of implied choice branch sequences

stevedlawrence:
When we need to add an implied sequence for a choice branch (like when the branch is an array), we wrap the branch Term in a ChoiceBranchImpliedSequence, which is essentially a Sequence that passes logic through to the underlying Term. This allows us to use existing sequence logic such as creating repetition parsers for arrays.

However, when we create this ChoiceBranchImpliedSequence we set its xml parameter to the xml of the Term it wraps. And because it extends Sequence, it does all the normal sequence logic of parsing that xml, even though much of it is never used because most logic passes through to the underlying Term.

In most cases, this works fine (although its wasted effort) since all that extra work is simply ignored. But, if that XML contains any expressions, things can quickly go off the rails as it tries to compile expressions with the context of this implied sequence.

To fix this, instead of passing in the XML of the underlying Term, we just pass in an empty sequence XML. This means there is no longer any real work for the ChoiceBranchImpliedSequence to do, except for be a tool to create the necessary repetition parsers to wrap around the Term. Not only does this avoid issues with incorrect contexts, but it avoids unnecessary work that is just ignored.

DAFFODIL-2887

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
@@ -346,7 +346,7 @@ object ChoiceBranchImpliedSequence {
  * handled as a degenerate sequence having only one element decl within it.
  */
 final class ChoiceBranchImpliedSequence private (rawGM: Term)
-  extends SequenceTermBase(rawGM.xml, rawGM.optLexicalParent, rawGM.position)
+  extends SequenceTermBase(<sequence />, rawGM.optLexicalParent, rawGM.position)
   with SequenceDefMixin
   with ChoiceBranchImpliedSequenceRuntime1Mixin {
 
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/discriminators/discriminator.tdml
@@ -1131,4 +1131,59 @@
     </tdml:errors>
   </tdml:parserTestCase>
 
+  <tdml:defineSchema name="discrimOnChoiceArray">
+    <xs:include schemaLocation="/org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"/>
+    <dfdl:format ref="ex:GeneralFormat" lengthKind="explicit" length="1" />
+
+    <xs:element name="r1" dfdl:lengthKind="implicit">
+      <xs:complexType>
+        <xs:sequence>
+          <xs:element name="type" type="xs:string" />
+          <xs:element name="max" type="xs:int" />
+          <xs:element name="values" dfdl:lengthKind="implicit">
+            <xs:complexType>
+              <xs:choice dfdl:choiceDispatchKey="{ ../ex:type }">
+                <xs:element name="a" dfdl:choiceBranchKey="a" type="xs:int" maxOccurs="unbounded">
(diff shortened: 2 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -346,7 +346,7 @@ object ChoiceBranchImpliedSequence {
  * handled as a degenerate sequence having only one element decl within it.
  */
 final class ChoiceBranchImpliedSequence private (rawGM: Term)
-  extends SequenceTermBase(rawGM.xml, rawGM.optLexicalParent, rawGM.position)
+  extends SequenceTermBase(<sequence />, rawGM.optLexicalParent, rawGM.position)
```
stevedlawrence, line 349: An alternative approach I considered but did not investigate might be to remove `ChoiceBranchImpliedSequence` entirely and instead have logic when creating the choice parser/unparser to inspect the branches and wrap arrays with the right Sequence parser/unparser.

This might simplify some logic (there are a number of places where we have to check if something is the implied sequence and do something different), but possibly makes the logic to create choice parser/unparser a bit more complex.

Thoughts?

mbeckerle: To me ChoiceBranchImpliedSequence is a natural thing to have.  Taking it out would require adding conditional logic to deal with choice branches that are just single non-sequence terms (i.e., elements - which can be repeating, and can have separators - by inheriting properties) in many places where now it just uses SequenceTermBase and doesn't care where the sequence came from. 

Tricking it to have dummy empty sequence XML is a natural fix to me. 

An alternative I did consider way back, was to make all such implied sequences be actual manifest sequences. This would be done when parsing the schema in the ModelGroup code I think. When it parses a choice it would look for a branch that is not a sequence, and create (and nest) the XML properly at that point before parsing the schema. 

That would establish the invariant that there are NO implied sequences for choices, they're all in the schema, albeit by inserting them. But I think this would also be sub-optimal, as errors could complain about a sequence that the user knows nothing of. 

I think a ChoiceBranchImpliedSequence class is the right device to enable error messages to refer properly to the content of a choice branch, using sequence-like language when sensible, and avoiding such language when it is not.

tuxji: Sounds good to me!  I appreciate having this look at how choices work under the hood, as it answers something I was wondering about (how does Daffodil encapsulate choice branches which are not complex element references or sequences, but rather single simple-type elements?).  Now I know Daffodil encapsulates them in a ChoiceBranchImpliedSequence with a dummy empty sequence XML.

tuxji (approved):
+1

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
stevedlawrence: Actually, I think we only wrap choice branches with an implied sequence for branches that are arrays, since array occurrences are handle by the sequence parser. If there's only a single element, or an explicit sequence, then we already have a single parser for that branch and we don't need anything extra.

Agreed that ChoiceBranchImpliedSequence is likely the best approach.

mbeckerle (approved):
+1

PR #1223, opened 2024-04-22
Add support for nested jars

stevedlawrence:
Daffodil supports resolving schemaLocations relative to a jar URI, which only looks for the schemaLocation in that same jar. Such jar URI's usually look like this:

  jar:file:/path/to/some.jar!/path/to/schema.dfdl.xsd

To resolve a path relative to this, we split this into its components: a jar part and the resource path part inside that jar and resolve the schemaLocation relative to the path part.

Although Java does not support loading resources within nested jars, tools like Spring Boot have added such support. This can result in jar URIs that look like this:

  jar:file:/path/to/outer.jar!/path/to/inner.jar!/path/to/schema.dfdl.xsd

This breaks our current code because we expect there to only be two components, but we end up with three because of the multiple exclamation points.

To fix this and to support nested jars, this changes the logic to only split on the last exclamation point. Everything before that is considered the jar component and everything after is considered the path component in that jar, which could be a nested jar. We then resolve relative paths to this path component as usual and look for the result to be in that same jar component.

DAFFODIL-2892

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
@@ -160,6 +160,26 @@ object Misc {
   }
 
   /**
+   * Split a jar URI into its two component parts: 1) the path to a jar file 2) the resource
+   * path within that jar file. By definition, these two parts are delimited by the last
+   * occurrence of an exclamation point. If the resource path part contains an exclamation point
+   * it should have already been escaped using "%21", which is the URI escape code for an
+   * exclamation point.
+   *
+   * Note that it is important that we look for the last index of an exclamation point, since it
+   * is possible some tools have added support for nested jars, and could have multiple
+   * exclamation points for each nesting.
+   */
+  def splitJarUri(uri: URI): (String, String) = {
+    Assert.invariant(uri.getScheme == "jar")
+    var uriStr = uri.toString
+    val exclamIdx = uriStr.lastIndexOf("!")
+    val jarPart = uriStr.substring(0, exclamIdx)
+    val pathPart = uriStr.substring(exclamIdx + 1)
+    (jarPart, pathPart)
+  }
+
+  /**
    * Java 20 deprecated the 2-arg URL constructor which worked to create relative URIs
    * within the same Jar file.
    *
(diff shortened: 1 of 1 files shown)


stevedlawrence:
Note that this is difficult to test because nested jars are not supported by Java by default--they are a feature added by things like Spring Boot. I'm not sure of a good way to regression test this without a lot of code to create and support nested jars. But the reporter of this issue has verified that this PR fixed the issue for them.

tuxji (approved):
+1

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
```diff
@@ -159,6 +159,25 @@ object Misc {
     }
   }
 
+  /**
+   * Split a jar URI into its two component parts: 1) the path to a jar file 2) the resource
+   * path within that jar file. By definition, these two parts are delimited by the last
+   * occurrence of an exclamation point. If the resource path part contains an exclamation point
+   * it should have already been escaped using %21.
```
mbeckerle, line 166: wording. escaped as %21 or escaped using %21. Is %21 a exclamation point, or an escape character?

mbeckerle (approved):
+1

Apache Commons really should contain all this sort of URI stuff. Every application needs this sort of thing.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
stevedlawrence: I updated the comment to make this more clear. 

I think maybe one reason why this isn't part of something like apache commons is that it really isn't standard. Nested jars is not something supported by java and the "jar" URI natively. Instead, things like Spring Boot add that support. And because we kindof mess with jar URIs (which may not be standard Java jar URIs) we start running into problems like this.

Really, there should be a standard API for resolving a path relative to a jar URI. The URI class does have the `resolve` method, but that doesn't seem to work like you would expect for jar URIs. But if that did work, then however Spring Boot adds support for nested jars, it could probably also add support for resolving relative paths in nested jars.

So all this code we have should just be `uri.resolve(relativePath)` and it would just work. But unfortunately that's doesn't seem to be the case. Or we're using the URI class completely wrong...

PR #1228, opened 2024-05-01
Only validate if a simple element parse succeeds

stevedlawrence:
Currently, restriction validation logic takes place in a finally block of the ElementParserBase combinator when the parse status is "success". However, if an exception is thrown (e.g. Runtime SDE) then we stil hit this finally block, parse status could still be "success", but we likely won't have any data to validate. This can lead to a confusing exception that hides the original exception.

To fix this, this moves validation to the end of the try block, so that we only ever attempt validation if all aspects of a parse succeed. And if an exception is thrown, then this code is simply skipped.

This also modifies the TDML runner so it correctly sets validationMode even when errors are expected. This allows us to test this case where we expect a Runtime SDE error when validation is enabled.

DAFFODIL-2895

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
@@ -29,10 +29,9 @@ import org.apache.daffodil.runtime1.infoset._
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
 import org.apache.daffodil.runtime1.processors.Processor
 import org.apache.daffodil.runtime1.processors.Success
-import org.apache.daffodil.runtime1.processors.TermRuntimeData
 
 abstract class ElementParserBase(
-  rd: TermRuntimeData,
+  erd: ElementRuntimeData,
   name: String,
   patDiscrimParser: Maybe[Parser],
   patAssertParser: Array[Parser],
@@ -43,7 +42,7 @@ abstract class ElementParserBase(
   eParser: Maybe[Parser],
   eAfterParser: Maybe[Parser],
   eRepTypeParser: Maybe[Parser],
-) extends CombinatorParser(rd) {
+) extends CombinatorParser(erd) {
 
   override lazy val runtimeDependencies = Vector()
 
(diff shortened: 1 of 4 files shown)


tuxji (approved):

stevedlawrence:
Agreed, I don't think I want to mess with early returns in this PR. We do have DAFFODIL-1492 to find all the uses of returns and replace them. We'll probably want to do that all at the same PR in a consistent way.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
```diff
@@ -198,6 +197,14 @@ abstract class ElementParserBase(
         eAfterParser.get.parse1(pstate)
 
       if (pstate.processorStatus ne Success) return
+
+      val shouldValidate = erd.isSimpleType &&
+        pstate.dataProc.isDefined && pstate.dataProc.value.validationMode != ValidationMode.Off
```
mbeckerle, line 202: I think when "full" validation is on, we should not incur the overhead of 'limited' validation, we should leave it up to the Xerces validation at the end of whole parse.

So this should, I think, look for ValidationMode.Limited specifically.

mbeckerle (approved):
+1 One comment to consider - not sure we should do daffodil's internal validation if Full validation has been requested.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
stevedlawrence: That's a change in behavior that might break some tests? We might have some tests that set validationMode=full/on but expect specific error messages from Daffodil's limited validation. But that might be considered a bug in those tests?

Note we also have a similar check that would need to be changed here: https://github.com/apache/daffodil/blob/main/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceChildBases.scala#L532-L533. Might have others as well? So maybe if we want to change the behavior we might want to create a ticket and do it as as separate PR?

mbeckerle: Agreed. Separate ticket. Also current behaviour tests the facets for daffodil even when validation is full, and making sure the facets work in both can be important.

pkatlic (approved):

PR #1233, opened 2024-05-06
Change to eliminate the ugly trailing commas.

mbeckerle:
This bulk change touches everything, so be cautious when you rebase onto it any outstanding changes.

The merge of this can be deferred until current outstanding changes are merged first.

DAFFODIL-2897


mbeckerle:
Looks like 'sbt scalafmtAll scalafmtsbt' is not enough. There are still other files that this doesn't hit.

tuxji (approved):
+1

stevedlawrence:
> Looks like 'sbt scalafmtAll scalafmtsbt' is not enough. There are still other files that this doesn't hit.

Yep, for reference, the `daffodil-test-integration` project is not aggregated so that slow tests are not run when you call `sbt test`. But this skips all sbt commands for the subproject, not just `sbt test`, so all sbt commands (such as scalafmt) must be run specifically on that project with `sbt daffodil-test-integration/<command>`. Our CI does this when compile, run tests, and check scala formatting. Ideally we could just disable "test" aggregation for this subproject, but sbt doesn't provide that control.

stevedlawrence (approved):
+1

PR #1236, opened 2024-05-09
Support Java 21

pkatlic:
DAFFODIL-2898

main.yml: Add Java 21 to CI.

Misc.scala: Disambiguate method references.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         java_distribution: [ temurin ]
-        java_version: [ 8, 11, 17 ]
+        java_version: [ 8, 11, 17, 21 ]
         scala_version: [ 2.12.19 ]
         os: [ ubuntu-22.04, windows-2022, macos-12 ]
         exclude:
@@ -48,6 +48,8 @@ jobs:
             java_version: 8
           - os: macos-12
             java_version: 11
+          - os: macos-12
+            java_version: 21
         include:
           # configure shell/cc/ar for all OSes
           - os: ubuntu-22.04
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
@@ -547,8 +547,8 @@ object Misc {
       .onMalformedInput(CodingErrorAction.REPORT)
     val bb = ByteBuffer.wrap((0 to 255).map { i => i.toByte }.toArray)
     val cb = dec.decode(bb)
-    assert(cb.position == 0)
-    assert(cb.limit == 256)
+    assert(cb.position() == 0)
+    assert(cb.limit() == 256)
     // These 5 are unmapped by Windows-1252 but we want to turn any
     // byte into a legit character. So these We add 0x100
     // to get unicode codepoints.
--- a/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIDebugger.scala
+++ b/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIDebugger.scala
@@ -58,7 +58,9 @@ class TestCLIDebugger {
     "-Dorg.jline.terminal.provider=dumb",
     "-Dorg.jline.terminal.dumb=true",
     "-Dorg.jline.terminal.dumb.color=false",
-    "-Dfile.encoding=UTF-8"
+    "-Dfile.encoding=UTF-8",
+    "-Dstderr.encoding=UTF-8",
+    "-Dstdout.encoding=UTF-8"
   )
 
   val envs = Map(


.github/workflows/main.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         java_distribution: [ temurin ]
-        java_version: [ 8, 11, 17 ]
+        java_version: [ 8, 11, 17, 21 ]
```
tuxji, line 42: CI takes between 20 to 30 minutes to complete even before adding Java 21 as another platform to test.  We may want to consider dropping one or two older platforms.  I suggest we keep 8 (I assume we still need to verify compatibility with legacy systems using Java 8) but drop both 11 and 17.  It seems to me, from an API point of view, that keeping only the earliest and the latest platforms is enough to cover all the API changes since the API changes are cumulative from 8 to 21. 

If there are any concerns, we can discuss dropping 11 & 17 on the dev list first and do it in a separate PR instead of doing it in this PR (or hold this PR until we reach consensus on the dev list).

mbeckerle (approved):
+1 I use java 21 in development mostly. It is noticibly faster than java 8 and 11, I'm not sure about contrast with Java 17.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/util/Misc.scala
```diff
@@ -547,8 +547,8 @@ object Misc {
       .onMalformedInput(CodingErrorAction.REPORT)
     val bb = ByteBuffer.wrap((0 to 255).map { i => i.toByte }.toArray)
     val cb = dec.decode(bb)
-    assert(cb.position == 0)
-    assert(cb.limit == 256)
+    assert(cb.position() == 0)
+    assert(cb.limit() == 256)
```
tuxji, line 551: Oh, good, wish all changes were that easy :-).

tuxji (approved):
+1 with a proposal for discussion

.github/workflows/main.yml
tuxji: Actually, the different platforms build in parallel anyway so I now realize that dropping 11 and 17 may not decrease the total CI time significantly.  The 21 / windows job is taking an oddly longer time (it's still running while the others have finished) but perhaps it will cache something and run faster next time.

mbeckerle: Hopefully the data center running this CI is powered by renewable energy. Otherwise each Java version has a carbon footprint. Note: many data centers in US.TX are so powered. 

I suggest merge this, and let's discuss on dev list and users about Java version needs.

stevedlawrence: Looks like something is broken on Windows on Java 21. Some CLI tests are possibly hanging. Tests don't normally take this long.

As to what to drop, I would prefer we keep 11 and 17 since they are LTS and we claim to support them. I think the plan is to drop Java 8 support for Daffodil 4.0.0, so we'll be back to the normal 3 versions of java when that happens.

stevedlawrence (changes requested):
-1 we need to figure out the windows failures before merging this

.github/workflows/main.yml
tuxji: > Looks like something is broken on Windows on Java 21. Some CLI tests are possibly hanging. Tests don't normally take this long.

I used [SDKMAN on Windows](https://blog.vvauban.com/blog/install-sdkman-on-windows) to build Daffodil and run its CLI tests with the following JDKs:

* 22.0.1-tem (tests hang)
* 21.0.3-tem (tests hang)
* 20.0.2-tem (tests hang)
* 19.0.2-tem (tests hang)
* 18.0.2.1-tem (tests hang)
* 17.0.11-tem (tests run to completion)

Every single JDK after 17 hangs while running the CLI tests and has to be killed.  I scanned the [Consolidated JDK 18 Release Notes](https://www.oracle.com/java/technologies/javase/18all-relnotes.html) but didn't see anything that might explain the hanging.  

Steve, please troubleshoot this hanging problem if you have time.  I recommend making the following changes to build.sbt to build daffodil without any C compiler or C code generator dependencies, so that you will need only Windows, java, and sbt with the compile & daffodil-test-integration/test goals:

```diff
diff --git a/build.sbt b/build.sbt
index fc3b804a7..a0bf6faac 100644
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,6 @@ import sbtcc._
 lazy val genManaged = taskKey[Unit]("Generate managed sources and resources")
 lazy val genProps = taskKey[Seq[File]]("Generate properties scala source")
 lazy val genSchemas = taskKey[Seq[File]]("Generate DFDL schemas")
-lazy val genCExamples = taskKey[Seq[File]]("Generate C example files")
 lazy val genVersion = taskKey[Seq[File]]("Generate VERSION file")

 lazy val daffodil = project
@@ -30,7 +29,6 @@ lazy val daffodil = project
   .enablePlugins(JavaUnidocPlugin, ScalaUnidocPlugin)
   .aggregate(
     cli,
-    codeGenC,
     core,
     io,
     japi,
@@ -52,7 +50,7 @@ lazy val daffodil = project
     tutorials,
     udf
   )
-  .settings(commonSettings, nopublish, ratSettings, unidocSettings, genCExamplesSettings)
+  .settings(commonSettings, nopublish, ratSettings, unidocSettings)

 lazy val macroLib = Project("daffodil-macro-lib", file("daffodil-macro-lib"))
   .settings(commonSettings, nopublish)
@@ -99,29 +97,6 @@ lazy val runtime1Layers = Project("daffodil-runtime1-layers", file("daffodil-run
   .dependsOn(runtime1, lib % "test->test", slf4jLogger % "test")
   .settings(commonSettings)

-val codeGenCLib = Library("libruntime.a")
-lazy val codeGenC = Project("daffodil-codegen-c", file("daffodil-codegen-c"))
-  .enablePlugins(CcPlugin)
-  .dependsOn(core, core % "test->test", slf4jLogger % "test")
-  .settings(commonSettings)
-  .settings(
-    Compile / cCompiler := sys.env.getOrElse("CC", "cc"),
-    Compile / ccArchiveCommand := sys.env.getOrElse("AR", "ar"),
-    Compile / ccTargets := ListSet(codeGenCLib),
-    Compile / cSources := Map(
-      codeGenCLib -> ((Compile / resourceDirectory).value / "org" / "apache" / "daffodil" / "codegen" / "c" / "files"
-        * GlobFilter("lib*") * GlobFilter("*.c")).get()
-    ),
-    Compile / cIncludeDirectories := Map(
-      codeGenCLib -> Seq(
-        (Compile / resourceDirectory).value / "org" / "apache" / "daffodil" / "codegen" / "c" / "files" / "libcli",
-        (Compile / resourceDirectory).value / "org" / "apache" / "daffodil" / "codegen" / "c" / "files" / "libruntime"
-      )
-    ),
-    Compile / cFlags := (Compile / cFlags).value
-      .withDefaultValue(Seq("-Wall", "-Wextra", "-Wpedantic", "-std=gnu11"))
-  )
-
 lazy val core = Project("daffodil-core", file("daffodil-core"))
   .dependsOn(
     runtime1Unparser,
@@ -146,13 +121,12 @@ lazy val tdmlLib = Project("daffodil-tdml-lib", file("daffodil-tdml-lib"))
   .settings(commonSettings)

 lazy val tdmlProc = Project("daffodil-tdml-processor", file("daffodil-tdml-processor"))
-  .dependsOn(tdmlLib, codeGenC, core, slf4jLogger)
+  .dependsOn(tdmlLib, core, slf4jLogger)
   .settings(commonSettings)

 lazy val cli = Project("daffodil-cli", file("daffodil-cli"))
   .dependsOn(
     tdmlProc,
-    codeGenC,
     sapi,
     japi,
     schematron % Runtime,
@@ -172,7 +146,7 @@ lazy val schematron = Project("daffodil-schematron", file("daffodil-schematron")
   .settings(libraryDependencies ++= Dependencies.schematron)

 lazy val testDaf = Project("daffodil-test", file("daffodil-test"))
-  .dependsOn(tdmlProc % "test", codeGenC % "test->test", udf % "test->test")
+  .dependsOn(tdmlProc % "test", udf % "test->test")
   .settings(commonSettings, nopublish)
 //
 // Uncomment the following line to run these tests
@@ -214,7 +188,7 @@ lazy val testStdLayout = Project("daffodil-test-stdLayout", file("test-stdLayout
 // Choices here are Java LTS versions, 8, 11, 17, 21,...
 // However 8 is deprecated as of Java 21, so will be phased out.
 val minSupportedJavaVersion: String =
-  if (scala.util.Properties.isJavaAtLeast("21")) "11"
+  if (scala.util.Properties.isJavaAtLeast("20")) "11"
   else "8"

 lazy val commonSettings = Seq(
@@ -436,41 +410,3 @@ lazy val unidocSettings = Seq(
       }
   }
 )
-
-lazy val genCExamplesSettings = Seq(
-  Compile / genCExamples := {
-    val cp = (codeGenC / Test / dependencyClasspath).value
-    val inSrc = (codeGenC / Compile / sources).value
-    val inRSrc = (codeGenC / Test / resources).value
-    val stream = (codeGenC / streams).value
-    val filesToWatch = (inSrc ++ inRSrc).toSet
-    val cachedFun = FileFunction.cached(stream.cacheDirectory / "genCExamples") { _ =>
-      val forkCaptureLogger = ForkCaptureLogger()
-      val forkOpts = ForkOptions()
-        .withOutputStrategy(Some(LoggedOutput(forkCaptureLogger)))
-        .withBootJars(cp.files.toVector)
-      val mainClass = "org.apache.daffodil.codegen.c.DaffodilCExamplesGenerator"
-      val outdir = (codeGenC / Test / sourceDirectory).value / "examples"
-      val args = Seq(mainClass, outdir.toString)
-      val ret = Fork.java(forkOpts, args)
-      forkCaptureLogger.stderr.foreach { stream.log.error(_) }
-      if (ret != 0) {
-        sys.error("failed to generate C example files")
-      }
-      val files = forkCaptureLogger.stdout
-        .filterNot(_.startsWith("WARNING"))
-        .map { f =>
-          new File(f)
-        }
-        .toSet
-      stream.log.info(s"generated ${files.size} C examples to $outdir")
-      files
-    }
-    cachedFun(filesToWatch).toSeq
-  },
-  Compile / compile := {
-    val res = (Compile / compile).value
-    (Compile / genCExamples).value
-    res
-  }
-)
```

pkatlic: @stevedlawrence Excellent investigation, will add the patch and see if this fixes the Windows build.

stevedlawrence (approved):
+1, looks like the stderr.encoding/stdout.encoding fixed it :+1:

PR #1237, opened 2024-05-14
Enable javadoc and scaladoc from runtime1 API

mbeckerle:
This commit is not about the content of that javadoc nor scaladoc, but about the build.sbt and other changes to enable API doc from things other than our JAPI and SAPI modules.

DAFFODIL-2900


build.sbt
```diff
@@ -431,9 +444,20 @@ lazy val unidocSettings = Seq(
...
+        .filter { source =>
+          //
+          // This filter limits javadoc to only files with
+          // api, sapi, or japi directories in their paths.
+          //
+          val str = source.toString
+          str.contains("/api/") ||
+          str.contains("/sapi/") ||
+          str.contains("/japi/")
+        }
```
olabusayoT, line 460: I believe this will exclude the UDFs , as they don't have api in their path. Is that intentional?

olabusayoT (commented):
Question about filtering unidocAllSources

build.sbt
mbeckerle: Unintentional.

build.sbt
```diff
@@ -413,14 +413,27 @@ lazy val ratSettings = Seq(
...
-  JavaUnidoc / unidoc / unidocProjectFilter := inProjects(japi, udf),
+  ScalaUnidoc / unidoc / unidocAllSources :=
+    (ScalaUnidoc / unidoc / unidocAllSources).value.map { sources =>
+      sources.filter { source =>
+        //
+        // This filter limits the scaladoc to only files which have
+        // api, sapi, or japi as directory names in their paths.
+        //
+        val str = source.toString
+        str.contains("/api/") ||
```
stevedlawrence, line 432: For reference, these are the new classes in the API:

```
root
  org
   apache
     daffodil
       runtime1
         api
           ChoiceMetadata
           ComplexElementMetadata
           DFDL
           DFDLPrimType
           ElementMetadata
           InfosetArray
           InfosetComplexElement
           InfosetDocument
           InfosetElement
           InfosetItem
           InfosetSimpleElement
           InfosetTypeException
           Metadata
           MetadataHandler
           ModelGroupMetadata
           SequenceMetadata
           SimpleElementMetadata
           TermMetadata
         layers 
           api
             ChecksumLayer
             Layer
```
I think all the Metadata ones are new public APIs (are they considered experimental or are we officially supporting them an requiring backwards compatibility?).

But I consider `DFDL` to be an internal API that people shouldn't use, at least until we get rid of SAPI/JAPI with DAFFODIL-1747. Im not sure about the others. Are any of these considered internal and should be filtered out?

build.sbt
```diff
@@ -413,14 +413,27 @@ lazy val ratSettings = Seq(
...
   ),
-  JavaUnidoc / unidoc / unidocProjectFilter := inProjects(japi, udf),
+  ScalaUnidoc / unidoc / unidocAllSources :=
+    (ScalaUnidoc / unidoc / unidocAllSources).value.map { sources =>
+      sources.filter { source =>
+        //
+        // This filter limits the scaladoc to only files which have
+        // api, sapi, or japi as directory names in their paths.
+        //
+        val str = source.toString
```
stevedlawrence, line 430: `str` contains an absolute path to the sources, so there's a rare edge case that a directory that the daffodil repo is in contains one of these strings, which would result in not filtering anything out. For example, a path like `/home/user/api/repos/daffodil.git/`. I can't think of a reason why that would happen, but who knows.

We could avoid this by doing something like:
```scala
val str = source.toString.stripPrefix(baseDirectory.value.toString)
```

build.sbt
```diff
@@ -431,9 +445,21 @@ lazy val unidocSettings = Seq(
...
-        source.toString.contains("$") || source.toString.contains("packageprivate")
-      }
+      sources
+        .filterNot { source =>
+          source.toString.contains("$") || source.toString.contains("packageprivate")
+        }
+        .filter { source =>
+          //
+          // This filter limits javadoc to only files with
+          // api, sapi, or japi directories in their paths.
```
olabusayoT, line 455: Comments need to be updated, but +1

olabusayoT (approved):
+!

build.sbt
```diff
@@ -413,14 +413,28 @@ lazy val ratSettings = Seq(
...
+      sources.filter { source =>
+        //
+        // This filter limits the scaladoc to only files which have
+        // api, sapi, or japi as directory names in their paths.
+        //
+        val str = source.toString
+        str.contains("/udf/") ||
+        str.contains("/api/") ||
+        str.contains("/sapi/") ||
+        str.contains("/japi/")
```
stevedlawrence, line 434: /japi/ doesn't exist in scaladoc. It doesn't hurt to have it I guess, and it makes the logic the same, so maybe this is fine?

build.sbt
```diff
@@ -431,9 +445,21 @@ lazy val unidocSettings = Seq(
...
+        .filter { source =>
+          //
+          // This filter limits javadoc to only files with
+          // api, sapi, or japi directories in their paths.
+          //
+          val str = source.toString
+          str.contains("/udf/") ||
+          str.contains("/api/") ||
+          str.contains("/sapi/") ||
+          str.contains("/japi/")
```
stevedlawrence, line 461: Same thoughts as above, note that sapi doesn't exist in javadoc but probably doesn't hurt to keep for consistency.

build.sbt
```diff
@@ -431,9 +445,21 @@ lazy val unidocSettings = Seq(
...
+      sources
+        .filterNot { source =>
+          source.toString.contains("$") || source.toString.contains("packageprivate")
+        }
+        .filter { source =>
+          //
+          // This filter limits javadoc to only files with
+          // api, sapi, or japi directories in their paths.
+          //
+          val str = source.toString
```
stevedlawrence, line 457: The generated javadoc doesn't contain the metadata an other classes mentioned above. I think because they are written in scala so would need the genjavadoc plugin. Or do we want to exclude those and consider them experimental? Maybe we should only allow "/layers/api/" or something?

stevedlawrence (commented):
Very cool this worked. Some quesions about what is internal vs public, and differences between java and scaladoc.

build.sbt
stevedlawrence: Another option might be to use full paths, e.g.
```
str.conains("/org/apache/daffodil/udf/") ||
...
```
This might be better so we are more explicitly about exactly which package namespaces are considered public API.

mbeckerle: I will include this when I centralize the filtering.

build.sbt
```diff
@@ -53,6 +54,8 @@ lazy val daffodil = project
     udf
   )
   .settings(commonSettings, nopublish, ratSettings, unidocSettings, genCExamplesSettings)
+  .settings(javadocSettings: _*) // for genjavadoc to create javadoc from scala code
```
mbeckerle, line 57: @stevedlawrence this line is one of the permutations of how to make genjavadoc work from the github readme for that tool. I also tried using `.config(Javadoc)` from their instructions also, with no effect. 

This doesn't cause an error, but neither does it work to cause the genjavadoc and unidoc to work together.

stevedlawrence: My guess is we need something similar to `daffodil-japi/build.sbt` copied into `daffodil-runtime1` to enable the GenJavadoc plugin for sbt-unidoc to use.

mbeckerle: Please take another look. 

This runs, but errors in genjavadoc becuase the filters do not apply to runtime1 files until after it fails with a zillion javadoc errors trying to convert scaladoc to javadoc for files we do not care about. We need to exclude various directories before genjavadoc converts them, but currently genjavadoc by way of unidoc, is getting handed whole sub-projects like "runtime1" and "japi" to convert.

stevedlawrence: I've just pushed a change that I think fixes the CI, looks like the CI is passing so far.

This fixes it by disabling publishing of -doc jar artifacts, including genjavadoc artifacts. We don't really need these and people probably shouldn't reference them anyways. They should use the public API published on our website. The reason this works:

There is no way to filter out what GenJavadoc generates. When SBT compiles a scala file with the genjavadoc compiler plugin enabled it is going to generate a java file for it, no matter what. For complex scala files like in runtime1, this can sometimes generate invalid java files that cause scala/javadoc to fail. And when scaladoc/javadoc is run it's going to cause a failure.

However, we use unidoc which runs on a subset of files (thanks to the new filtering) so these invalid files should be ignored. However, when publishing jars (which a number a tasks can trigger like publishLocal, daffodil-cli/stage, etc.), it also publishes -doc artifacts for each project. These doc artifacts do not use the same filtering as unidoc, so they end up using these invalid files and cause the failure.

Removing the PublishJavaDocPlugin is probably sufficient to cause it to ignore these invalid java files when publishing -doc artifacts, but really we don't need to be publishing any docs.

build.sbt
```diff
@@ -17,7 +17,8 @@
 
 import scala.collection.immutable.ListSet
 
-import sbtcc._
+import Dependencies.genjavadoc
+import sbtcc.*
```
stevedlawrence, line 21: Is this import * correct, shouldn't it be _ in scala?

project/Dependencies.scala
```diff
@@ -62,4 +62,11 @@ object Dependencies {
   lazy val exi = Seq(
     "com.siemens.ct.exi" % "exificient" % "1.0.7"
   )
+
+  // Scala Steward may try to update this version to include the Scala version,
+  // for example 0.18_2.12.15. This is incorrect because the unidoc plugin uses
+  // crossVersion to figure out the Scala version. This should be set to just the
+  // version of the genjavadoc plugin, without the Scala version.
+  lazy val genjavadoc: String = "0.19"
```
stevedlawrence, line 70: I don't know how scala steward knows how to update this version, but based on the comment it can. But I wonder if it does it because it knows about the "unidocGenjavadocVersion" key, which is no longer right next to this version? Maybe we should change this to something like:

```scala
lazy val genjavadoc = Seq(
  unidocGenjavadocVersion := "0.19"
)
```

and

```scala
 .enablePlugins(GenJavadocPlugin)
 .settings(Dependencies.genjavadoc)
}
```
I would hope that makes it more like it can find the dependency string since it's very similar to what it was before.

build.sbt
```diff
@@ -412,15 +415,48 @@ lazy val ratSettings = Seq(
   ratFailBinaries := true
 )
 
+/**
+ * Used on projects where we want javadoc even though the code is written in scala.
+ */
+def withScalaDocConvertedToJavaDoc(p: Project): Project = {
+  p.enablePlugins(GenJavadocPlugin)
+    .settings(unidocGenjavadocVersion := Dependencies.genjavadoc)
+}
```
stevedlawrence, line 424: I think we could actually just move the unidocGenjavadocVersion setting to commonSettings, then any projects that enable GenJavadocPlugin will automatically get it. And since enablePlugins is the only thing this does, we can just move `enablePlugins(GenJavadocPlugin)` to the individual projects.

build.sbt
```diff
@@ -429,12 +465,15 @@ lazy val unidocSettings = Seq(
...
-    sources =>
-      sources.filterNot { source =>
-        source.toString.contains("$") || source.toString.contains("packageprivate")
+  JavaUnidoc / unidoc / unidocAllSources :=
+    (JavaUnidoc / unidoc / unidocAllSources).value
+      .map(apiDocSourceFilter(_))
+      .map {
+        _.filterNot { source =>
+          // For JAPI so that some things are excluded from the javadoc because they are internal
+          source.toString.contains("$") || source.toString.contains("packageprivate")
```
stevedlawrence, line 474: Could we move these to the `excluded` check in apiSourceDocFilter? No scala files have dollar signs or "packageprivate" in them so it shouldn't affect scala docs.

build.sbt
```diff
@@ -246,7 +249,9 @@ lazy val commonSettings = Seq(
   sourceManaged := baseDirectory.value / "src_managed",
   resourceManaged := baseDirectory.value / "resource_managed",
   libraryDependencies ++= Dependencies.common,
-  testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "--verbosity=1")
+  testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "--verbosity=1"),
+  Compile / packageDoc / publishArtifact := false,
+  Dependencies.genjavadocVersion
```
stevedlawrence, line 254: This now results in a bunch of warnings like this:
```
[warn] there are 20 keys that are not used by any other settings/tasks:
[warn]  
[warn] * daffodil / unidocGenjavadocVersion
[warn]   +- /home/runner/work/daffodil/daffodil/project/Dependencies.scala:73
...
```

I guess my idea of adding this globally wasn't a great idea. Maybe we just add
```scala
.settings(Dependencies.genjavadocVersion)
```
after we enable the JavadocPlugin for each project? Its a bit repetitive, but I personally find it a bit cleaner than the `withScalaDocConvertedToJavaDoc` function. But I don't feel strongly about it.

project/Dependencies.scala
```diff
@@ -15,7 +15,9 @@
  * limitations under the License.
  */
 
-import sbt._
+import sbt.*
+import sbt.Def
+import sbtunidoc.GenJavadocPlugin.autoImport.unidocGenjavadocVersion
```
stevedlawrence, line 20: Do we actually nee these changes? We don't use `sbt.Def` in this class, and I think `unidocGenjavadocVersion` should be autoimported and not needed. And the `sbt.*` import isn't standard scala is it? This feels like an auto-importer gone wrong.

stevedlawrence (approved):
+1, minor comments

project/Dependencies.scala
mbeckerle: Well I get clear errors if I don't have the import of unidocGenjavadocVersion. It has no type information for that object otherwise.

stevedlawrence: Yep, you're right, I guess SBT *.scala files don't work like .sbt files.

build.sbt
mbeckerle: Tools are suggesting * now. I think scala 3 preferred style is * not _ for several things _ was being used for. Scala 2 did have far too many overloads of things on the _ character.

PR #1244, opened 2024-05-20
WIP: checkpoint on preserve CR issue

mbeckerle:
checkpoint - possibly unnecessary files

DAFFODIL-1559

### DIFF
--- a/daffodil-cli/src/it/scala/org/apache/daffodil/xml/TestXMLConversionControl.scala
+++ b/daffodil-cli/src/it/scala/org/apache/daffodil/xml/TestXMLConversionControl.scala
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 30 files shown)


stevedlawrence:
Does Daffodil ever really need the CRL -> LF canonicalization? If CR is part of an actual field and is not used as a delimiter or something, it seems we should never do this lossy mapping. Maybe we should just always disable CR -> LF canonicalization and convert CR to PUA. It makes infosets messier, but enures we don't lose data.

stevedlawrence:
> The default scheme needs to be exactly what we do today. 

I thought mapping CR to LF was a regression that should be fixed? Or have we always done this mapping?

> we need to settle on the scheme for how users parameterize the InfosetOutputters and Inputters needs to be agreed upon

Agreed. Interesting idea to have more complex but more human friendly mappings. It does potentially cause issues with validation since changing the mapping requires different validation pattern facets and thus a different schema.

mbeckerle:
> 
> I thought mapping CR to LF was a regression that should be fixed? Or have we always done this mapping?
> 
Actually, I am not sure about this. We need to look at the Daffodil release before Jan 2023 and test it there.

stevedlawrence:
> > I thought mapping CR to LF was a regression that should be fixed? Or have we always done this mapping?
> 
> Actually, I am not sure about this. We need to look at the Daffodil release before Jan 2023 and test it there.

I tested as far back as 3.1.0 and all of them did the CR -> LF mapping. So this is not a regression. Maybe not the preferred behavior, but probably a change we wait for 4.0.0 to change.

mbeckerle:
> I tested as far back as 3.1.0 and all of them did the CR -> LF mapping. So this is not a regression. Maybe not the preferred behavior, but probably a change we wait for 4.0.0 to change.

Thanks for doing that. Fixing this functionally is very easy. We will need a switch to control it, and putting in a switch/param to enable/disable/choose the behavior is the stumbling block. 

I believe you suggested we do this with a new properties-file method, as opposed to extending the current tunables system to include this behavior. 

If you can write that up for the dev list then it's probably pretty easy to agree on and we can implement it with this as the driving use case.

PR #1246, opened 2024-05-28
Updated javadoc for layering.

mbeckerle:
DAFFODIL-2569


pkatlic (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -27,51 +27,231 @@
...
+ *         <td>xs:short</td>
+ *         <td>short</td>
+ *     </tr>
+ *     <tr>
+ *         <td>xs:long</td>
+ *         <td>long</td>
+ *     </tr>
+ *     <tr>
+ *         <td>xs:int</td>
+ *         <td>java.math.BigInteger</td>
```
stevedlawrence, line 166: Does xs:int map to a Java int? I thought xs:integer maps to BigInteger? And I think this list is missing xs:integer? Is that an accident or do layers not support xs:integer?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
```diff
@@ -96,23 +275,34 @@ public Layer(String localName, String targetNamespace) {
...
    * <p>
    * This method and the string it returns are required by the SPI loader.
    * @return A unique indentifier for the kind of layer. Contains both local and namespace components of the layer's complete name.
    */
   public final String name() { return LayerUtils.spiName(localName, targetNamespace); }
 
+  /** For framework use. */
   public final String localName() { return this.localName; }
+
+  /** For framework use. */
```
stevedlawrence, line 288: We might want to consider an update (in a later release) to make these package private, which I think should exclude them from javadoc? I guess that would mean the classes calling these functions would need to be in the same package--maybe that refactoring would be too complicated.

stevedlawrence (approved):
+1, minor comments

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
mbeckerle: Yeah, package private requires use of reflection, or lots of other code has to be in the layers.api package. 

However, ChatGPT4o says there are ways to add one's own annotations to java code that the javadoc tool can see:

https://chatgpt.com/share/ed29c152-9bfa-4598-8a46-bf590abefb7d

Perhaps in the future. 

Note that if you are pointing an IDE at the jar and sources, this wouldn't work. only the generated javadoc HTML would see the exclusions.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/layers/api/Layer.java
mbeckerle: Hmmm. I will check the code and/or fix this doc. xs:integer should have a mapping to java.lang.BigInteger and xs:int to int.

PR #1249, opened 2024-05-30
Update style on bitOrder tutorial.

mbeckerle:
This improvement is not really necessary, but I did it as part of other work, then found it wasn't needed. 

This ticket is an excuse to integrate this change anyway, eventhough it turned out not to be needed for a different change set. 

Uses default namespaces to avoid prefixes on more than just the tutorial elements.

Avoids TNS convention. Uses elementFormDefault="unqualified", and defineSchemas do not use the default namespace.

DAFFODIL-2906


mbeckerle:
Too much whitespace changing. Will revert and recreate.

PR #1252, opened 2024-06-03
Allow DataProcessors to be garbage collected in TDML tests

stevedlawrence:
DataProcessors are currently stored in a "processor" variable in each TestCase for easy access while a TestCase is running. However, once a TestCase finishes running, it really shouldn't hold on to this DataProcessor anymore so that it can be garbage collected. We now set that variable to null when finished to allow this. Note that this does not remove DataProcessors from the compile cache--it only allows garbage collection of DataProcessor created with one of the withXYZ functions to tweak the cached processor for a specific test.

Allowing garbage collection of DataProcessors is especially important when Xerces validation is enabled, since each DataProcessor hold a unique instance of a possibly very large Xerces validator. We likely need to change how we store Xerces validators so that they can be shared, but in the meantime this change at least allows a large number of TDML tests to run with validation enabled without running into OutOfMemory errors.

DAFFODIL-2901


mbeckerle (approved):
+1

pkatlic (approved):
+1

PR #1256, opened 2024-06-06
Upgrade nodejs in the release candidate container

stevedlawrence:
The version of nodejs that ships with Fedora 40 has a bug that causes a SIGILL error when runing npm on some architectures. This installs nodejs/npm from the Fedora "updates" repo which includes a fix.


mbeckerle:
Based on this, which I can't verify against the machine that was getting the SIGILL as it's at home, I'm just going to create the 3.8.0-rc1 using fedora 40 on a machine which does NOT get the SIGILL issue. 

That way this can be merged later along with a few other improvements to the release container.

mbeckerle (approved):
+1

PR #1273, opened 2024-07-31
Optimize BucketingInputSource for performance

stevedlawrence:
The BucketingInputSource has a bytePositionToIndicies function that returns a tuple containing the bucket index and index within that bucket to find a given byte position. This function should be inlined and in theory the tuple allocation could be optimized out since we immediately take it apart into separate variables, but that doesn't seem to be the case, and leads to noticeable Overhead.

This removes the tuple allocation by replacing the one function with two separate functions. This means there is now an extra function call but it avoids the tuple allocation, which appears to be the main overhead.

This also is more careful about which variables are Int's and Long's to minimize the number of toInt calls. This is unlikely to make a performance difference, but does make the code cleaner. This also switches from integer/modular division to shifts and masks which should also be more efficient. This does now require the bucket size to be specified as a power of two.

In basic testing, these changes reduced the overhead of the BucketingInputSource compared to the ByteBufferInputSource from about 15% to 5%.

DAFFODIL-2920

### DIFF
--- a/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
+++ b/daffodil-io/src/main/scala/org/apache/daffodil/io/InputSource.scala
@@ -212,14 +212,17 @@ abstract class InputSource {
  * can be freed when the reference count goes to zero.
  *
  * @param inputStream the java.io.Inputstream to read data from
- * @param bucketSize the size of each individual bucket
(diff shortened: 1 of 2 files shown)


pkatlic (approved):
+1, looks like the changes to use bit operations in getBucketIndex and getByteIndex will result in performance improvements

jadams-tresys (approved):
+1

Changes look good, but didn't see much performance difference on my system.  Only saw about a 1% improvement averaging 5 runs both before and after.

PR #1274, opened 2024-07-31
mmap files when possible to improve CLI parse performance

stevedlawrence:
Daffodil currently supports two different input sources: a BucketingInputSource backed by an InputStream and ByteBufferInputSource backed by a ByteBuffer. The CLI currntly always uses the BucketingInputSource because the ByteBufferInputSource does not support stdin or files larger than 2GB. Although the gap is closing due to other optimizations, the BucketingInputSource still has overhead compared to ByteBufferInputSource due to added complexity.

This changes the CLI logic to use a ByteBufferInputSource where possible (parsing files <= 2GB) using mmap and a MappedByteBuffer to efficiently create a ByteBuffer.

Basic testing shows about a 5% increase over the BucketingInputSource for a large file with many small reads.

DAFFODIL-2921

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
@@ -25,7 +25,10 @@ import java.io.PrintStream
 import java.net.URI
 import java.nio.ByteBuffer
 import java.nio.channels.Channels
+import java.nio.channels.FileChannel
+import java.nio.file.Files
 import java.nio.file.Paths
+import java.nio.file.StandardOpenOption
 import java.util.Scanner
 import java.util.concurrent.Executors
 import javax.xml.parsers.SAXParserFactory
@@ -1165,13 +1168,24 @@ class Main(
           case Some(processor) => {
             Assert.invariant(!processor.isError)
             val input = parseOpts.infile.toOption match {
-              case Some("-") | None => STDIN
+              case Some("-") | None => InputSourceDataInputStream(STDIN)
               case Some(file) => {
-                val f = new File(file)
-                new FileInputStream(f)
+                // for files <= 2GB, use a mapped byte buffer to avoid the overhead related to
+                // the BucketingInputSource. Larger files cannot be mapped so we cannot avoid it
+                val path = Paths.get(file)
+                val size = Files.size(path)
+                if (size <= Int.MaxValue) {
+                  val fc = FileChannel.open(path, StandardOpenOption.READ)
+                  val bb = fc.map(FileChannel.MapMode.READ_ONLY, 0, size)
+                  fc.close() // we no longer need the channel now that we've mapped it
+                  InputSourceDataInputStream(bb)
+                } else {
+                  val is = Files.newInputStream(path, StandardOpenOption.READ)
+                  InputSourceDataInputStream(is)
+                }
               }
             }
-            using(InputSourceDataInputStream(input)) { inStream =>
+            using(input) { inStream =>
               val output = parseOpts.output.toOption match {
                 case Some("-") | None => STDOUT
                 case Some(file) => new FileOutputStream(file)
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/io/InputSourceDataInputStream.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/io/InputSourceDataInputStream.scala
@@ -26,7 +26,30 @@ import org.apache.daffodil.io.{ InputSourceDataInputStream => SInputSourceDataIn
  * Provides Daffodil with byte data from an InputStream, ByteBuffer, or byte
  * Array.
  *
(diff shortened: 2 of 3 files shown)


pkatlic (approved):
+1

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -1165,13 +1168,24 @@ class Main(
...
-              case Some("-") | None => STDIN
+              case Some("-") | None => InputSourceDataInputStream(STDIN)
               case Some(file) => {
-                val f = new File(file)
-                new FileInputStream(f)
+                // for files <= 2GB, use a mapped byte buffer to avoid the overhead related to
+                // the BucketingInputSource. Larger files cannot be mapped so we cannot avoid it
+                val path = Paths.get(file)
+                val size = Files.size(path)
+                if (size <= Int.MaxValue) {
```
mbeckerle, line 1177: doc for the map method says: 

    For most operating systems, mapping a file into memory is more expensive than reading or writing a 
    few tens of kilobytes of data via the usual read

So should there be a floor check also e.g., below some size we just read it into a byte buffer and avoid the map?

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -1165,13 +1168,24 @@ class Main(
...
+                  val fc = FileChannel.open(path, StandardOpenOption.READ)
+                  val bb = fc.map(FileChannel.MapMode.READ_ONLY, 0, size)
```
mbeckerle, line 1179: This is in the CLI. Could this be done inside the API so that all applications benefit from it? 

E.g, InputSourceDataInputStream(is) analyzes the input stream to see if it is a file and of the needed size?

mbeckerle (approved):
+1

I just have some questions for discussion. As is the change is an improvement certainly.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
stevedlawrence: Possibly. Though, I imagine if you're parsing a small file with the CLI then the overhead of mmap is going to be relatively small compared to the overhead of starting up a JVM and maybe the it won't make a difference? I'm not sure. We can do some experiments to see if there's a benefit for smaller files.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
stevedlawrence: We could, but I'm a little hesitant to force something on a API user if we can't say for sure it will be faster in 100% of cases, especially if there are cases where it could be slower (e.g. like with small files you mentioned).

Maybe an alternative might be to instead just provide better API documentation, maybe something like:

> The InputStream variant has potential overhead due to streaming capabilities and support for unlimited data sizes. In some cases, better performance might come from using the ByteBuffer variant instead. For example, if your data is already in a byte array, one should use the Array[Byte] or ByteBuffer variants instead of wrapping it in a ByteArrayInputStream. As another example, instead of using a FileInputStream one could consider mapping the File to a MappedByteBuffer, keeping in mind that MappedByteBuffers might have different performance characteristics depending on the file size and system.

And then we leave it up to the API users to figure out what works best for their system/environment?

mbeckerle: Add to that comment a link to the code lines in the CLI as an illustrative example of how to do it, or just put an example in the javadoc, and I agree that would be sufficient.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
mbeckerle: Could leave this as you have it, and we can look at nightly performance stuff to see if it slows down noticably. Lots of those will use files that are small.

stevedlawrence: The nightlies don't use the `parse` command so won't see any change. They use the `performance` command which reads test files into a byte array before testing to avoid overhead related to disk I/O.

We could create some patches that run on the nightlies, one patch change the performance command to use FileInputStream and one to use a MappedByteBuffer, which would give us an idea of mmap vs file input stream. But that's feels like a decent amount of work just to figure out an optimal size where mmap overhead > bucketing overhead. Also, based on my bucketing vs non-bucketing tests, I feel like bucketing overhead is probably more than mmap-overhead, even with small files and so we should always avoid bucketing when possible.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
stevedlawrence: Done.

PR #1280, opened 2024-08-21
Minor fixes to the release candidate container script

stevedlawrence:
Add additional changes to commit a13501a90d to improve the release candidate container.

- Appends a colon and space to the prompt in the get_non_empty_input function instead of in caller parameters. The callers did already add a colon but not a space so the input looked off. This ensures the colon and space always exist and callers do not have to think about it
- The "Apache Password" prompt does not provide the "-s" option to hide the password. This modifies get_non_empty_input so that any parameters after the prompt are passed directly to "read" and the "Apache Password" call is modified to pass in "-s"

DAFFODIL-2914

### DIFF
--- a/containers/release-candidate/src/daffodil-release-candidate
+++ b/containers/release-candidate/src/daffodil-release-candidate
@@ -47,13 +47,16 @@ USAGE
 }
 
 # Function to get a non-empty string as input
+# $1 is the prompt to use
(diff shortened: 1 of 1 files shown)


pkatlic (approved):
+1

mbeckerle (approved):
+1

PR #1282, opened 2024-08-22
Improve reload diagnosics when a character set plugin is missing

stevedlawrence:
Currently, reloading a saved parser when a neccessary character set plugin is not on the classpath results in an exception with a very helpful message that has nothing to do with character sets or plugins, for example:

    java.lang.ClassCastException: cannot assign instance of
    scala.collection.immutable.List$SerializationProxy to field
    org.apache.daffodil.runtime1.processors.ModelGroupRuntimeData.groupMembers
    of type scala.collection.Seq in instance of
    org.apache.daffodil.runtime1.processors.SequenceRuntimeData

This is a known issue with scala and its use of proxy when serializing data structures like Lists and Maps.

To fix the exception and improve diagnostics, this uses writeReplace to serialize a BitsCharset as a BitsCharsetSerializationProxy with information about the needed BitsCharset. This proxy implements readResolve to look up the BitsCharset from the bits charset registry and restore the original BitsChraset. If not found in the registry, then we throw an helpful exception that bubbles up to the reload function and we can output a helpful diagnostic. Now we get something like:

    [error] The saved parser was created with a different set of
    dependencies containing a class no longer on the classpath: Charset
    plugin com.example.MyCustomCharset for X-DFDL-MY-CUSTOM-CHARSET

Add an integraion test to make sure this errors with a reasonable diagnostic, which require splitting a schema file into valid and invalid parts so we can reuse it with the CLI.

DAFFODIL-2915

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/compiler/Compiler.scala
@@ -20,6 +20,7 @@ package org.apache.daffodil.core.compiler
 import java.io.File
 import java.io.FileInputStream
 import java.io.InvalidClassException
+import java.io.InvalidObjectException
 import java.io.ObjectInputStream
 import java.io.StreamCorruptedException
 import java.net.URI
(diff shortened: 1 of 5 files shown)


stevedlawrence:
I thought it would probably be wise to add tests that reload a saved parser with missing layers and UDFs on the classpath, in addition to charsets. UDF's work as is and give a good diagnostic. Layers also work, but create an SDE that must be caught when reloading or else the CLI thinks it's a bug. PR is upated.

mbeckerle (approved):
+1

jadams-tresys (approved):
+1

This definitely would have saved me some headaches when trying to figure out why a schema stopped working all of a sudden when I didn't realize the charset was no longer in the JAVA/DAFFODIL path.

PR #1283, opened 2024-08-22
Do not show counterintuitive placement for testKind=expression

jadams-tresys:
The counterintuitive placement warning should only be displayed for asserts/disciminators that have testKind=expression.

DAFFODIL-2912

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
@@ -33,6 +33,7 @@ import org.apache.daffodil.lib.schema.annotation.props.SeparatorSuppressionPolic
 import org.apache.daffodil.lib.schema.annotation.props.gen.OccursCountKind
 import org.apache.daffodil.lib.schema.annotation.props.gen.SeparatorPosition
 import org.apache.daffodil.lib.schema.annotation.props.gen.SequenceKind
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -164,9 +165,10 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
   protected final lazy val checkIfNonEmptyAndDiscrimsOrAsserts: Unit = {
     val msg = "Counterintuitive placement detected. Wrap the discriminator or assert " +
       "in an empty sequence to evaluate before the contents."
-    if (groupMembers.nonEmpty && discriminatorStatements.nonEmpty)
+    // Only show warning if the testKind=expression, see DFDL spec Section 9.5 Evaluation Order for Statement Annotations
+    if (groupMembers.nonEmpty && discriminatorStatements.filter(_.testKind == TestKind.Expression).nonEmpty)
```
stevedlawrence, line 169: Might be a little more clear to do `discriminatorStatements.exists(_.testKind == TestKind.Expression)`

jadams-tresys: Agreed.  I always forget about exists()

mbeckerle (commented):
+1

PR #1285, opened 2024-08-27
Add compileResource API method

stevedlawrence:
Some API users want to find a DFDL schema resource on the classpath (e.g. by using Class.getResource). However, this results in an absolute URI which when passed to compileSource cannot be depersonalized in some cases and leads to verbose diagnostic paths. This is also one issue that prevents reproducible saved parsers since these absolute paths are included in the saved parser and are specific to the build environment.

To fix this, this adds a new compileResource API function that accepts a resource path. This new API calls getResource instead of the user having to call it, and it creates a URISchemaSource with the resulting URI and diagnostic path context set to the resource path. This allows the diagnostic paths to be depersonalized and is one step towards reproducible saved parsers.

DAFFODIL-2919

### DIFF
--- a/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
+++ b/daffodil-japi/src/main/scala/org/apache/daffodil/japi/Daffodil.scala
@@ -172,6 +172,40 @@ class Compiler private[japi] (private var sCompiler: SCompiler) {
   }
 
   /**
+   * Compile DFDL resource name into a [[ProcessorFactory]]
+   *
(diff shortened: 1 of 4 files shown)


stevedlawrence:
Note that I have tested modifying the sbt-daffodil plugin to use this new API and it works as expected, resulting in depersonalized and less verbose diagnostic paths when using the `packageDaffodilBin` command.

mbeckerle (approved):
+1

jadams-tresys (approved):
+1

PR #1286, opened 2024-08-28
Remove absolute URI from SchemaFileLocation to improve reproducibility

stevedlawrence:
When we create a saved parser, the absolute URI of the schema context is stored in the ScemaFileLocation. This absolute URI is very likely to be different if compiled on different machines or in different directories. This makes verify reproducible saved processors difficult.

To fix this, this removes the URI from the SchemaFileLocation. SchemaFileLocation is only used for diagnostics so does not need the absolute URI--diagnosticFile should be used for that. The URI is still stored in some objects where it is needed (e.g. import/include), but not in any places that are serialized when we created a saved parser.

Previous places that used the URI from SchemaFileLocation are modified to either get the URI from somewhere else (that isn't serialized) or use diagnosticFile, which is depersonalized and allows for reproducible builds.

To verify correctness, this adds a CLI test that builds the same saved parser in two different randomly created directories and ensures they have the same hash. Some CLI test utilities were updated to support this, including the ability to specific a different working directory to run the CLI command from.

DAFFODIL-2918

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
@@ -225,9 +225,10 @@ class DataProcessor(
       case ValidationMode.Limited => null
       case ValidationMode.Custom(cv) => cv
       case ValidationMode.Full => {
-        val cfg = XercesValidatorFactory.makeConfig(
-          ssrd.elementRuntimeData.schemaURIStringsForFullValidation
-        )
+        // we only need to provide the main schema, Xerces will find imported/included schemas
+        // using the Daffodil resolver set in the XercesValidatorFactory
+        val sources = Seq(ssrd.mainSchemaUriForFullValidation.toString)
+        val cfg = XercesValidatorFactory.makeConfig(sources)
         XercesValidatorFactory.makeValidator(cfg)
       }
     }
--- a/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIMisc.scala
+++ b/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIMisc.scala
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.cli.cliTest
+
+import java.nio.file.Files
+
+import org.apache.daffodil.cli.Main.ExitCode
+import org.apache.daffodil.cli.cliTest.Util._
+
+import org.junit.Assert.assertEquals
+//import org.junit.Test
+
+class TestCLIMisc {
+
+  /**
+   * Create a saved parser for the same schema but in two different directories, the resulting
+   * saved parsers should be exactly the same
+   *
+   * Note that this test is disabled because JVM optimizations (e.g. serializing zero-length
+   * arrays to singletons) are somewhat inconsistently applied and can lead to saved parsers
+   * that are functionally the same but not bit-for-bit identical, causing CI to randomly fail.
+   * This should be enabled when DAFFODIL-2925 is fixed.
+   */
+  /*@Test*/
+  def test_CLI_SaveParser_reproducible(): Unit = {
+    val schema = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/charClassEntities.dfdl.xsd"
+    )
+    val schemaName = schema.getFileName
+
+    val md5sums: Seq[String] = (1 to 2).map { _ =>
+      withTempDir { dir =>
+        withTempFile { parser =>
+          val newSchema = dir.resolve(schemaName)
+          Files.copy(schema, newSchema)
+          runCLI(args"save-parser -s $schemaName $parser", directory = Some(dir)) { cli => }(
+            ExitCode.Success
+          )
+          Util.md5sum(parser)
+        }
+      }
+    }
+
+    assertEquals(md5sums(0), md5sums(1))
+  }
+}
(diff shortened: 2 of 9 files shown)


daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIMisc.scala
```diff
@@ -0,0 +1,55 @@
...
+   * Create a saved parser for the same schema but in two different directories, the resulting
+   * saved parsers should be exactly the same
+   */
+  @Test def test_CLI_SaveParser_reproducible(): Unit = {
+    val schema = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/charClassEntities.dfdl.xsd"
+    )
+    val schemaName = schema.getFileName
+
+    val md5sums = (1 to 2).map { _ =>
```
mbeckerle, line 40: This is an awesome test.

I would declare the type of the md5sums variable just because you are depending on it being a type that assertEquals properly iterates, and I'm not sure what this type is.

stevedlawrence: The md5sum is just string, but agreed it would make it more clear. I'll add it.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/DataProcessor.scala
```diff
@@ -225,9 +225,10 @@ class DataProcessor(
       case ValidationMode.Limited => null
       case ValidationMode.Custom(cv) => cv
       case ValidationMode.Full => {
-        val cfg = XercesValidatorFactory.makeConfig(
-          ssrd.elementRuntimeData.schemaURIStringsForFullValidation
-        )
+        // we only need to provide the main schema, Xerces will find imported/included schemas
+        // using the Daffodil resolver set in th XercesValidatorFactory
```
pkatlic, line 229: Either change "th" to "the" or remove.

pkatlic (approved):
+1 after comment is updated

stevedlawrence:
I've made the suggested updates, but I might hold off on merge this right away. This new test has failed GitHub CI a couple of times on the Java 11 + Windows build. Seems like there's still some source of randomness somewhere. I'll see if I can track it down before merging this to avoid randomly breaking CI.

stevedlawrence:
After a number of runs, I was able to reproduce the issue of different saved parsers in GitHub twice (I'm still unable to do it locally) and download the differing saved parsers. I was able to dump the serialized processors to text (using [NickstaDB/SerializationDumper](https://github.com/NickstaDB/SerializationDumper)) and do a diff, and found the only difference in both cases was this:
```diff
  delimiters
    (array)
-     TC_ARRAY - 0x75
-       TC_REFERENCE - 0x71
-         Handle - 8258633 - 0x00 7e 04 49
-       newHandle 0x00 7e 04 c6
-       Array size - 0 - 0x00 00 00 00
-       Values
+     TC_REFERENCE - 0x71
+       Handle - 8258634 - 0x00 7e 04 4a
```
The `delimiters` member is the `Array[DelimiterParseEv]` in the `DelimiterStackParser`. I'm not 100% sure how to interpret this, but I think in the first case the delimier array is a uniquely allocated zero-length array, where the reference is to the type (i.e. the class description for `DelimiterParseEv)`. In the second case, it just points to an already existing array zero-length array. So in one case the zero-length array is shared, in the other it's not. So functionally exactly the same, just one has extra allocations.

I wonder if maybe there is some optimization somewhere something detects zero-length arrays and can share them instead of allocating new arrays? And sometimes that happens and sometimes it doesn't, which leads to differences in serialization?

I imagine we could maybe do something like manually detect when an array is zero length and make sure we use Array.empty() or some globally allocated zero-length array, but I imagine we're unlikely to ever find all cases so reproducibility will still be an issue.

On a positive note, I have confirmed the difference has nothing to do with paths, so that issue preventing reproducibility is definitely fixed. But I wonder if we should disable this test to avoid random CI failures? Maybe in the future we can switch to some serialization format that is more reliably consistent?

Any thoughts?

mbeckerle:
Maybe see if stack-overflow folks can clarify the "I 'm not 100% sure how
to interpret this." , problem?

stevedlawrence:
I've done a bunch more research an experiments to figure out what's going on.

---
TDLR; Even though some zero-length arrays are separately allocated and do not have reference equality, Java serializes them in a way that makes them the same reference. This is very consistent with one particular array (delimiters member in DelimiterStackParser) but not 100%. And I cannot ever reproduce this behavior (non-reference equal things serialized to reference equal things) in any other cases, even with arrays of the same type.

---

I wrote a test scala file that serialized an object containing zero length arrays in different ways (i.e sometimes have reference equaltiy, sometimes uniquely allocated and not having reference equality) , looked that serialization dump, and confirmed that my interpretation was correct. An array that has reference equality to a previously serialized array looks like this:

```
fieldName
  (array)
    TC_REFERENCE - 0x71
      Handle - 8258634 - 0x00 7e 04 4a
```

But an array does not have reference equality serializes to this:
```
fieldName
  (array)
    TC_ARRAY - 0x75
      TC_REFERENCE - 0x71
        Handle - 8258633 - 0x00 7e 04 49
      newHandle 0x00 7e 04 c6
      Array size - 0 - 0x00 00 00 00
    Values
```

In the first case the TC_REFERENCE is to the pre-serialized array, in the second case the TC_REFERENCE is to the type since the type is the same, but it's a completely separate array.

What's confusing is that I've confirmed that none of the serialized delimiter arrays have reference equality when they are created. They are all separately allocated arrays. But when inspecting the serialized dump, the zero length arrays do have reference equality. I've also confirmed that when deserializing the arrays do have reference equality. I've also confirmed the fields do not have reference equality when the serialization function writeObject() is called. So it must be somewhere in the actual serialization code (i.e. ObjectOutputStream.defaultWriteObject) doing this optimization.

What's also weird is I add a new (unused) member to a non-delimiter parser and allocated zero-length `Array[DelimterParseEv]` and those were not changed at all during serialization. They all stayed separately allocated arrays without reference equality. So whatever is doing this optimization doesn't do it in this case. But it almost always does it for the delimiters field in the `DelimiterStackParser`, but not 100%.

Seems like some kind of deep Java optimization to save some bytes during serialization, but it is not very consistent. I'm also surprised Java is even allowed to do that. Changing something to have reference equality seems like it could potentially break things if something relied things having meaning if they when not reference equal.

mbeckerle:
I suggest just comment out the test, put comments in there pointing to this issue discussion, or exerpt from this in comments there, and then commit the commented out test. No point in getting stuck on this. 

I can't think of any other real way to test this other than building new infrastructure to walk a processor so we could do our own equality checker. 

We may need that anyway for some other purposes - specifically, creating tests that determine if optimizations worked or did not work - we need to be able to create sensible queries about the processors coming out that traverse them looking for things (or the lack of things) to know if optimizations worked or not.

stevedlawrence:
I tried to do more digging, and really couldn't find anything. The best I found find is a conversation I had with ChatGPT where it claims there are JVM optimizations where it will sometimes convert zero-length arrays to singletons. I can't find any actual references to this, and sometimes ChatGPT gave conflicting responses, so I'm not sure how much I trust it, but it does explain what I am seeing and was what I suspected was happening. It's still odd that is does this optimization very consistently in one specific place, and no where else that I see, but I guess optimizations can be odd and hard to predict.

Assuming this is the actual issue, I think we probably need to comment out this test since it sometimes leads to CI failures. It also means reproducibility in saved parsers is not guarantee, but it is at least much better with this change, so I still think it's worth adding. If this becomes a priority in the future, we can investigate alternative serialization mechanisms that are more reproducible.

PR #1287, opened 2024-08-30
Daffodil 2899

parkera101093:
SDW now properly gives VariableRead WarnID for case VariableRead in setVariable funcion

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/VariableMap1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/VariableMap1.scala
@@ -458,9 +458,10 @@ class VariableMap private (
           WarnID.VariableSet,
           "Cannot set variable %s after reading the default value. State was: %s. Existing value: %s",
           variable.rd.globalQName,
-          VariableSet,
+          VariableRead,
           variable.value
         )
+
         variable.setValue(newValue)
         variable.setState(VariableSet)
       }
--- a/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
+++ b/daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
@@ -24,7 +24,7 @@
   xmlns:fn="http://www.w3.org/2005/xpath-functions"
   xmlns:ex="http://example.com"
   defaultRoundTrip="true">
-  
+
   <tdml:defineSchema name="v">
     <dfdl:defineVariable name="v_no_default" type="xsd:int" />
     <dfdl:defineVariable name="v_with_default" type="xsd:int"
@@ -685,7 +685,7 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
-    
+
     <xs:group name="g8">
       <xs:sequence>
         <xs:element name="month" type="xsd:string" dfdl:lengthKind="delimited"
@@ -1251,7 +1251,7 @@
         </e8>
       </tdml:dfdlInfoset>
     </tdml:infoset>
-    
+
     <tdml:document>23</tdml:document>
   </tdml:unparserTestCase>
 
@@ -1336,7 +1336,7 @@
         </xs:sequence>
       </xs:complexType>
     </xs:element>
-    
+
     <xs:element name="r2">
       <xs:complexType>
         <xs:sequence dfdl:separator=",">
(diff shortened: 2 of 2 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/VariableMap1.scala
```diff
@@ -455,14 +455,14 @@ class VariableMap private (
          * the way we evaluate escapeSchemes it could lead us to setting the
          * variable read too early */
         pstate.SDW(
-          WarnID.VariableSet,
+          WarnID.VariableRead,
           "Cannot set variable %s after reading the default value. State was: %s. Existing value: %s",
           variable.rd.globalQName,
-          VariableSet,
+          VariableRead,
```
stevedlawrence, line 461: Is line 461the only line that needs to be changed?

This warning is about setting a variable, so I think the WarnID should probably still be `VariableSet`.

And the variable state needs to be set to "VariableSet" below since we are changing the variable state from Read to Set. I think all that needs to change is the warning from "State was VariableSet" to "State was VariableRead", which this line does.

Also, please rebase/squash your changes into a single commit and include the bug number in the commit message. The [Code Contributor Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow) has commands and an example commit message if you're unsure.

stevedlawrence (changes requested):

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -735,6 +735,7 @@
           <xs:enumeration value="unsupportedAttributeFormDefault" />
           <xs:enumeration value="unsupportedAttributeSchemaLocation" />
           <xs:enumeration value="variableSet" />
+          <xs:enumeration value="variableRead" />
```
stevedlawrence, line 738: Can this be removed? I don't think this new warn ID is used or neeeded?

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -24,7 +24,7 @@
   xmlns:fn="http://www.w3.org/2005/xpath-functions"
   xmlns:ex="http://example.com"
   defaultRoundTrip="true">
-  
+
```
stevedlawrence, line 27: In general, I find it's best to avoid these kinds of unrelated whitespace changes to make reviewing the actual changes easier. Sometimes this requires tweaks to IDE settings to disable this. We probably should fix the whitespace in our TDML files at some point, but we should that to everything at once like we did when we standardized our scala formatting.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -2217,6 +2217,9 @@
         </root>
       </tdml:dfdlInfoset>
     </tdml:infoset>
+    <tdml:warnings>
+      <tdml:warning>State was: VariableRead</tdml:warning>
```
stevedlawrence, line 2221: This should be moved up into the other `<tdml:warnings>` tag, makes it easier if all the warnings are in the same place, and there really isn't anything gained by having them separate--I didn't even realize we supported that.

daffodil-test/src/test/resources/org/apache/daffodil/section07/variables/variables.tdml
```diff
@@ -2646,6 +2649,9 @@
...
+    </tdml:warnings>
```
stevedlawrence, line 2654: I think this `<tdml:warnings>` can be removed. Although this test might result in that warning, it really isn't specifically about that warning--it's really just making sure some escape character thing works. Whether or not there is a warning doesn't really matter to this tests.

The above tests that you've modified is specifically about this warning and making sure it's reasonable and correct.

stevedlawrence (commented):
Looks good, just a couple minor this to clean up.

Also, the commit message says it corrects the WarnID, but that hasn't changed. All that has changed is the warning message to include the correct state.

stevedlawrence (approved):
+1

mbeckerle (approved):
+1 nothing to add beyond Steve's comments.

PR #1289, opened 2024-09-05
Use targetNamespace when resolving global references without a namespace prefix or default namespace

stevedlawrence:
When resolving global references (e.g. global element declaration, complex types, DFDL variables, DFDL formats), if the QName does not have a prefix then we currently use the default namespace (i.e. xmlns="...") when resolving the reference. And if the schema does not define a default namespace then we just assume NoNamespace.

This works fine except when a schema with a targetNamespace includes a schema with no targetNamespace and default namespace. In that case, resolving global references in the included schema must use the targetNamespace of the including schema because all the global components have been chameleoned into a different namespace.

To fix this, we now pass around targetNamespace to a number of places where resolving a global reference might be done (which is a lot of places), and use that targetNamespace when resolving a QName without a prefix and no default namespace.

Note that the logic to resolve defineFormat references already did something similar using its custom "adjustNamespace" logic. This is removed so all global references use the same logic when resolving references--no post-resolution adjustment is needed.

A number of functions were refactored a bit to make the logic and style more consistent, making it easier to see where logic differed. Some unused functions were discovered and removed.

DAFFODIL-2916


daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesExpressions.scala
```diff
@@ -197,6 +202,7 @@ case class SetVariable(stmt: DFDLSetVariable, override val term: Term)
 
   override lazy val exprText = stmt.value
   override lazy val exprNamespaces = stmt.xml.scope
+  override lazy val exprTargetNamespace = stmt.annotatedSC.targetNamespace
```
mbeckerle, line 205: This is pretty paindful that everywhere we have xyzNamespaces to get the scopes we have in parallel xyzTargetNamespace also. 

These should be one context object, not two both of which must be maintained in parallel in all these different places. 

I.e., one tuple of these two things, not two separate things passed all the time.

mbeckerle (approved):
+1

I'm "ok" with this change, and I can see how you got here by pulling this thread, and finding one place after another after another to change and also provide the targetNamespace info.

In hindsight, it's clear the context for name resolution needs to be a more complex thing, and we should be creating an XSDContext object (a glorified tuple) of the scope info for namespace bindings for prefixes, and the right target namespace. 

That said, this is not just some xsdContext lazy val on each DSOM object, because that would preclude any sharing of DSOM objects that are for DECLS that are included into multiple target namespaces. 

I'm a bit concerned that even as coded we're wired into copying here. I.e., pretty sure we actually make new DSOM objects every time a schema file is included into a different targetNamespace. A way to share instead would be better of course. 

Honestly though, I think we should be adopting (and enforcing via an optional switch) a new DFDL schema style which avoids 100% of all element namespacing issues by requiring that global elements are only declared in schemas with no namespace. I.e.,  all root schemas have no namespace. The complex types and groups that actually contain the schema guts can all have namespaces to avoid collisions and such, but no element namespaces at all. 

Solves the JSON compatibility problem entirely, and similarly so many other data interfacing problems where there is no namespace concept. 

Given that, I definitely do *not* want to over invest in code to cope with XML schema nuances of namespace hell. 

So this code is fine, I just wish we'd pushed back on XML namespace hell sooner.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/RestrictionUnion.scala
```diff
@@ -102,10 +107,10 @@ final class Restriction private (xmlArg: Node, val simpleTypeDef: SimpleTypeDefB
    * Exclusive - restriction either has a baseType or a direct primType.
    */
   lazy val (optDirectPrimType, optBaseTypeDef: Option[GlobalSimpleTypeDef]) = {
-    val optPT = schemaSet.getPrimitiveType(baseQName)
+    val optPT = PrimType.fromQName(baseQName)
```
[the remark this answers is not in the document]
stevedlawrence: > The schemaSet.getPrimitiveType(baseQName) should take the baseQName and run up it's definition chain until it finds the eventual primitive type it is derived from

That's not what getPrimitiveType currently does. It has basically the same logic as the new `PrimType.fromQName` in that it returns a `Some` if the QName is the XSD namspace and the localName is one of the primitive types, and it returns None otherwise.  We don't currently have function that does the chain thing to find a primitive type. Right now you would have call getGlobaSimpleTypeDef and then look at primType on the result to find the base primitive type. And if a property can resolve to a global simple type def OR a primitive type, you can to call both getGlobalSImpleTypeDef and getPrimType.

We could change add a new function that could do what you describe, but it's maybe outside the scope of this change?

> Is this code really trying to both chase the chain back to the primitive base type while also assuring that the referenced types all exist?

It looks like restriction unions have different logic depending on if the type is a primitive type or a global simple type def, looks like something about combining facets from different type defs? I'm not too sure, as I didn't really look to deep into this since I didn't change the logic, just changed how resolve a string to a QName. What we do with that QName hasn't changed.

Note that I forget exactly why I moved fromQName from SchemaSet to PrimType (I went down a few different rabbit holes before came to this solution) but the way it works kind of made more sense to me sinc it doesn't take into account the schema set at all. It's purely about "is this QName the XSD namespace" and "is his a valid XSD primitive local name". I can move it back to SchemaSet if you think it makes sense, but the logic is the same.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
```diff
@@ -413,87 +411,52 @@ final class SchemaSet private (
...
-    context: ThrowsSDE
-  ): GlobalSimpleTypeDef = {
-    val gstd = getGlobalSimpleTypeDef(refQName)
-    gstd.getOrElse {
-      val isPrimitive = getPrimitiveType(refQName).isDefined
-      val msg =
-        if (isPrimitive) s"The $prop property cannnot resolve to a primitive type: $refQName"
-        else s"Failed to resolve $prop to a global simpleType definition: $refQName"
-      context.schemaDefinitionError(msg)
+  def getGlobalSimpleTypeDef(refQName: RefQName) = {
```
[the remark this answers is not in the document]
stevedlawrence: Yeah, they're all Option types, that wasn't changed. Agreed they should have return types. Will add.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesExpressions.scala
stevedlawrence: Agreed, especially if the solution for DAFFODIL-2917 is to add support for xpathDefaultNamespace, which would be another piece of context for resolving things. And maybe the value of unqualifiedPathStepPolicy (which may or may not go away?) also wants to be part of that. Should we add the `XSDContext` thing you suggest and put all those pieces in it? And anything that resolves QNames needs an XSDContext, and any new context needed for such resolution is added to that object?

mbeckerle: I suggest merging this as is. There are a variety of other namespace-related things I'd like to consider before designing the complete solution.  Minimization of namespace bindings is one of them, for example, which involves assembling more than one such context and potentially creating unique namespace prefix bindings to resolve ambiguity.

jadams-tresys (approved):
+1

PR #1290, opened 2024-09-05
Fixed test test_date_constructor_01 in TestDFDLExpressions

parkera101093:
Fixed test "test_date_constructor_01" in testDFDLExpressions. This involved uncommenting the test and changing the expected error message in date_constructor_01 in Functions.tdml

This test checks that single digit days are not allowed in dates.

This test has been disabled for a long time due to a now fixed bug where single digit days were being allowed in dates. This bug has been fixed and single digit days are no longer allowed, so the test has been renenabled, and the expected error message has been updated

 DAFFODIL-1124


stevedlawrence (approved):
+1 :+1:

mbeckerle (approved):
+1

PR #1293, opened 2024-09-06
Fix check for explicit length within length facet range

stevedlawrence:
When we check if an explicit dfdl:length is within range of the minLength/maxLength/length facets, we do not take into account units or character encoding, which can lead to incorrect and confusing warnings.

This modifies the check to use the existing LengthInBitsEv to calculate the length in bits of an element (which uses lengthUnits), and uses CharsetEv for string types to correctly convert the explicit bit length to infoset length, where possible

DAFFODIL-2888

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
@@ -1611,37 +1611,47 @@ trait ElementBaseGrammarMixin
         )
       case _ => // ok
     }
+
     /*
-     * When lengthKind is explicit and length is a constant, it is a warning if
-     * the type is a type that respects minLength and maxLength, and the constant length
-     * is not in range.
+     * Warn if a type respects minLength/maxLength/length facets and we can calculate that the
+     * infoset length will be out of range of the facet values. Note that we can only do this in
+     * specific cases, like when the length and encoding properties are constant and the
+     * encoding is fixed width characters.
      */
-    val isTypeUsingLengthOrMinMaxLengthFacets = typeDef.typeNode match {
-      case s: NodeInfo.String.Kind => true
-      case s: NodeInfo.HexBinary.Kind => true
-      case _ => false
-    }
-    if (
-      (lengthKind eq LengthKind.Explicit) &&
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1611,37 +1611,47 @@ trait ElementBaseGrammarMixin
...
-        value
-      )
-      if (hasLength && len != minLengthLong && len != maxLengthLong)
-        warn("length", minLengthLong)
-      else if (hasMinLength || hasMaxLength) {
-        if (maxLengthLong != -1 && len > maxLengthLong) warn("maxLength", maxLengthLong)
-        Assert.invariant(minLengthLong >= 0)
-        if (minLengthLong > 0 && len < minLengthLong) warn("minLength", minLengthLong)
+    if ((lengthKind eq LengthKind.Explicit) && (hasLength || hasMinLength || hasMaxLength)) {
+      val optInfosetLen = elementLengthInBitsEv.optConstant.flatMap { maybeKnownLenInBits =>
```
mbeckerle, line 1622: naming convention. Normally the prefix "maybe" is used for things of type Maybe. Is the type of optConstant's result a Maybe? If so then all good. I just want to make sure we're following those conventions if all these types are getting omitted.

mbeckerle (approved):
+1 with a style comment

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1611,37 +1611,47 @@ trait ElementBaseGrammarMixin
...
       case _ => // ok
     }
+
     /*
-     * When lengthKind is explicit and length is a constant, it is a warning if
-     * the type is a type that respects minLength and maxLength, and the constant length
-     * is not in range.
+     * Warn if a type respects minLength/maxLength/length facets and we can calculate that the
+     * infoset length will be out of range of the facet values. Note that we can only do this in
+     * specific cases, like when the length and encodig properties are constant and the encoding
```
olabusayoT, line 1618: encoding*

olabusayoT (approved):
Minor typo in comment

+1

PR #1296, opened 2024-09-10
Removed argType parameter from FNZeroArgExpr

parkera101093:
argType parameter wasn't needed for FNZeroArgExpr since it's supposed to implement dfdl functions that take no argument those functions are fn:local-name and fn:namespace-uri functions.

argType is used for targetTypeForSubexpression which should never be called in FNZeroArgExpr since their are zero arguments. However, targetTypeForSubexpression still has be implemented in the class since FNZeroArgExpr extends FunctionCallBase, so targetTypeForSubexpression is set to Node.Exists in the FNZeroArgExpr class definition.

In addtion, parameter were removed from FNLocalName0 and FNNamespaceUri0 since these parameters were no longer needed after FNZeroArgExpr was updated. Also, to support this change, both classes now extend RecipeOp instead of RecipeOpWithSubRecipes since RecipeOp needs no parameters and the subRecipes feature wasn't being used in these functions anyways.

 DAFFODIL-2331

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
@@ -1508,8 +1508,7 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           functionQNameString,
           functionQName,
           NodeInfo.String,
-          NodeInfo.Exists,
-          FNLocalName0(_, _)
+          FNLocalName0
         )
 
       case (RefQName(_, "local-name", FUNC), args) if args.length == 1 =>
@@ -1527,8 +1526,7 @@ case class FunctionCallExpression(functionQNameString: String, expressions: List
           functionQNameString,
           functionQName,
           NodeInfo.String,
-          NodeInfo.Exists,
-          FNNamespaceUri0(_, _)
+          FNNamespaceUri0
         )
 
       case (RefQName(_, "namespace-uri", FUNC), args) if args.length == 1 =>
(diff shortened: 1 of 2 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -2546,18 +2544,17 @@ case class FNZeroArgExpr(
...
   resultType: NodeInfo.Kind,
-  argType: NodeInfo.Kind,
-  constructor: (CompiledDPath, NodeInfo.Kind) => RecipeOp
+  constructor: () => RecipeOp
 ) extends FunctionCallBase(nameAsParsed, fnQName, Nil) {
 
   override lazy val inherentType = resultType
 
-  override def targetTypeForSubexpression(childExpr: Expression): NodeInfo.Kind = argType
+  override def targetTypeForSubexpression(childExpr: Expression): NodeInfo.Kind = NodeInfo.Exists
```
stevedlawrence, line 2552: This function should never be called since `FNZeroArgExpr`'s never have sub expressions. So instead of returning `NodeInfo.Exists`, we probably should just call `Assert.impossible()` wrapped with `// $COVERAGE-OFF$` and `// $COVERAGE-ON$` comments to to disable code coverage.

stevedlawrence (approved):
+1 with one a minor suggestion and fixes for the failing lint formatting check

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
parkera101093: Done!

stevedlawrence:
Just FYI, I pushed formatting updates since I believe Alex is on PTO this week and this is small enough that it can probably make it into the next release

mbeckerle (approved):
+1

PR #1298, opened 2024-09-13
JSON output no longer quotes all values

parkera101093:
JSONInfosetInputter and Outputter were treating all values as strings. This meant everyhing was quoted and was messier for the user to read. In addition, not all json comes from dfdl, so it makes sense to support numbers.

As of the time of this commit 156 tests fail as the changes cause some issues with dates and NaN, INF, and -INF values as well as numbers with a large amount of decimals such as 0.1234567898765432123456789 from test decimal_constructor_06.

 DAFFODIL-2362

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetInputter.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetInputter.scala
@@ -85,7 +85,9 @@ class JsonInfosetInputter(input: java.io.InputStream) extends InfosetInputter {
       jsp.getCurrentToken() match {
         case JsonToken.START_OBJECT => if (objectDepth == 1) StartDocument else StartElement
         case JsonToken.END_OBJECT => EndElement
-        case JsonToken.VALUE_STRING | JsonToken.VALUE_NULL => {
+        case JsonToken.VALUE_STRING | JsonToken.VALUE_NUMBER_INT |
+            JsonToken.VALUE_NUMBER_FLOAT | JsonToken.VALUE_TRUE | JsonToken.VALUE_FALSE |
+            JsonToken.VALUE_NULL => {
           // we don't want to start faking element end yet, but signify that
           // after a call to next(), we will want to fake it
           nextEventShouldBeFakeEnd = true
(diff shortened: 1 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetInputter.scala
```diff
@@ -172,7 +180,13 @@ class JsonInfosetInputter(input: java.io.InputStream) extends InfosetInputter {
...
 
           // start of a simple type or null
-          case JsonToken.VALUE_STRING | JsonToken.VALUE_NULL => exitNow = true
+          case JsonToken.VALUE_STRING |
+               JsonToken.VALUE_NUMBER_INT |
+               JsonToken.VALUE_NUMBER_FLOAT |
+               JsonToken.VALUE_TRUE |
+               JsonToken.VALUE_FALSE |
+               JsonToken.VALUE_NULL =>
+            exitNow = true
```
stevedlawrence, line 189: It looks like there is a `JsonToken.isScalar()` function that I think returns true if it's one of these VALUE_* tokens. So this case can probably be simplified to `case token if token.isScalar => exitNow = true`. Need to double check that's the case.

stevedlawrence (changes requested):
Looks good, just some suggestions for minor clean up and add a couple of tests to make sure we handle an error case correctly. I don't think we did before this change so it would be good to verify it's fixed and get a reasonable diagnostic now.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetInputter.scala
```diff
@@ -118,10 +123,11 @@ class JsonInfosetInputter(input: java.io.InputStream) extends InfosetInputter {
     primType: NodeInfo.Kind,
     runtimeProperties: java.util.Map[String, String]
   ): String = {
-    if (jsp.getCurrentToken() == JsonToken.VALUE_NULL) {
+    if (!jsp.getCurrentToken().isScalarValue()) {
+      throw new NonTextFoundInSimpleContentException("Unexpected array or object '" + getLocalName + "' on line " + jsp.getTokenLocation().getLineNr())
```
stevedlawrence, line 127: Can you add tests for this to make sure this works as expected?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/JsonInfosetOutputter.scala
```diff
@@ -112,14 +112,40 @@ class JsonInfosetOutputter private (writer: java.io.BufferedWriter, pretty: Bool
...
+      case DFDLPrimType.Date => true
+      case DFDLPrimType.Time => true
+
+      // json does not support inf/nan double/float so they must be quoted
+      case DFDLPrimType.Double => {
+        val d = simple.getDouble
+        d.isInfinite || d.isNaN
+      }
+      case DFDLPrimType.Float => {
+        val f = simple.getFloat.toDouble
```
stevedlawrence, line 142: You don't need to convert the float to a double, the Float class has `isInfinite` and `isNan` functions.

stevedlawrence (approved):
+1, I think it would be good to have a couple tests for the getSimpleText exception, also needs to reformat the files with `sbt scalafmtAll`

stevedlawrence:
I've updated this PR for minor cleanups and to improve test coverage. This is ready for review.

mbeckerle (approved):
+1

PR #1302, opened 2024-09-17
Support locally suppressing schema definition warnings

stevedlawrence:
We currently only support supressing schema definition warnings globally with the use of tunables. However, some uses of Daffodil may not be able to set tunables. Or in some cases we want to suppress a warning if it is caused by a specific location in a schema, but all others should not be suppressed.

Tos support this, this adds support for a new
"daf:suppressSchemaDefinitionWarnings" attribute, where the value is a whitespace separated list of schema definition warnings to suppress if the schema component it is placed on leads to a schema definition warning. Note that this property does not follow morning DFDL property scoping rules. So for example, you cannot have a default value defined in a DFDL format that applies globally to a schema.

Additionaly, we need to treat the new property similar to how to treat dfdl:ref (i.e. exclude it from certain logic), so logic related to excluding dfdl:ref is made more generic to support the new property, reduce special cases, and make surrounding code more consistent.

DAFFODIL-2638


jadams-tresys (approved):
+1

mbeckerle (approved):
+1

olabusayoT (approved):
+1

PR #1309, opened 2024-09-20
Flip ignoreUnexpectedValidationErrors flag

olabusayoT:
- we want to ignore unexpected validation error as the default

DAFFODIL-2927


stevedlawrence (approved):
+1, is it worth adding a test that relies on the default value being true so that if we unintentionally flip this flag it will fail and make us double check things?

mbeckerle (approved):
+1

PR #1310, opened 2024-09-20
Disable constant length facet check for inputValueCalc elements

stevedlawrence:
It is possible to have dfdl:lengthKind="explicit" defiend for elements with dfdl:inputValueCalc. In that case, ther lengthKidproperty should just be ignored. But the logic for checking if a constant length element is within facet length bounds did not ignore the property for IVC elements, which led to it expecting a dfdl:length and errorer.

To fix this, and addition restriction is added so we only do the check for represented elements, i.e. non-IVC elements.

DAFFODIL-2888


olabusayoT (approved):
+1

mbeckerle (approved):
+1

PR #1312, opened 2024-09-23
Add expressionAllowed param to getPropertyOption

olabusayoT:
- update function calls
- update tests to ensure we no longer get warnings about textStandardExponentRep

DAFFODIL-2928


mbeckerle (approved):
+1

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/PropertyGenerator.scala
```diff
@@ -430,41 +430,44 @@ trait CurrencyMixin extends PropertyMixin {
...
-      "occursCount",
-      "inputValueCalc",
-      "outputValueCalc",
       "textStandardInfinityRep",
       "textStandardNaNRep",
       "textStandardZeroRep",
       "nilValue",
       "textStringPadCharacter",
       "textNumberPadCharacter",
       "textBooleanPadCharacter",
```
stevedlawrence, line 468: The ones now in the `runtimeValuedProperties` are not really runtime valued properties, they are just done by hand for some reason (i'm not sure why?). I wonder if they really belong in the `excludedBecauseDoneByHand` member up above? And then `excludeRuntimeProperties` really only has to look at `expressionAllowedProperities`

daffodil-propgen/src/main/scala/org/apache/daffodil/propGen/PropertyGenerator.scala
```diff
@@ -642,12 +645,18 @@ object Currency {
...
       }
     })"""
-    val res = template.replaceAll("currency", propName)
+    val res = if (expressionAllowedProperties.contains(propName)) {
+      template
+        .replaceAll("currency", propName)
+        .replaceAll("false", "true") // add expression allowed for these
+    } else {
+      template.replaceAll("currency", propName)
+    }
```
stevedlawrence, line 659: Thoughts on changing `false` to `expressionAllowed` and then just always do something like this?
```scala
val expressionAllowStr = expressionAllowedProperties.contains(propName).toString
template
  .replace("currency", propName)
  .replace("expressionAllowed", expressionAllowedStr)
```
Avoids the replace duplicate code and matches on something less common to avoid possible incorrect replacements if we ever tweak this template.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/schema/annotation/props/ByHandMixins.scala
```diff
@@ -451,7 +451,7 @@ trait TextStandardExponentRepMixin extends PropertyMixin {
 
   lazy val textStandardExponentRep: Found = {
     val tsec = getPropertyOption("textStandardExponentCharacter")
-    val tser = getPropertyOption("textStandardExponentRep")
+    val tser = getPropertyOption("textStandardExponentRep", expressionAllowed = true)
```
stevedlawrence, line 454: The bug mentioned several properties that needed this. Did you confirm this is the only case that's missing?

stevedlawrence (approved):
+1 :+1: just minor suggestions for possible cleanup

PR #1314, opened 2024-09-24
Fix how unprefixed QNames are resolved

stevedlawrence:
Commit 2590f54ce9 change it so that if no default namespace is defined then an unprefixed QName reference is resolved using the targetNamespace as its namespace. However, this logic was not correct.

Instead, we need different logic depending on if the schema containing such a reference is included vs imported. In the former case, we should use the targetNamespace of the including schema. In the latter, we should just use no-namespace.

This modifies the logic to add a new 'noPrefixNamespace' member which keeps track of what namespace to use when resolving a unprefixed QName reference using include/import information. This member replaces previous changes of targetNamespace.

This found a number of tests where references to defined formats were not actually correct but happened to work due to previously flawed logic. This corrects these tests, and adds additional tests to make sure include/import logic works as expected.

Deprecation/Compatibility:

A bug was fixed so Daffodil now correctly handles resolving unprefixed QName references. However, this means that some unprefixed references that relied on the previous incorrect behavior may not resolve anymore. Although there are a number of ways to fix this, defining a namespace prefix and adding that prefix to broken reference should work in all cases.

DAFFODIL-2916


mbeckerle:
I think the ticket is DAFFODIL-2916, not 2961.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/QNameBase.scala
```diff
@@ -168,19 +168,22 @@ object QName {
...
-    // qualified then we probably do want to use targetNamespace to look it up since those
-    // elements likely have a namespace. But for non-qualified local steps, we probably do just
-    // want to use NoNamespace since the step won't have a namespace. However, this is sort of a
-    // chicken-and-egg problem--we need to know the namespace to how to find the step, but we
-    // don't know which namespace to use (NoNamespace or targetNamespace) unless we've already
-    // found the same. Instead, we use NoNamespace and let the flexibility that
-    // unqualfiedPathStepPolicy gives use help to find the step element. Possibly related to
-    // DAFFODIL-2917
+    // It is not clear what namespace to use when a path step does not have a prefix and there
+    // is no default namespace defined. The step could be to a global element in the same
```
mbeckerle, line 175: Comment isn't right. Or reflects the confusion that this fix straightens out. 

The reasoning is expressed in the definition of the noPrefixNamespace member.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaDocIncludesAndImportsMixin.scala
```diff
@@ -93,6 +93,39 @@ trait SchemaDocIncludesAndImportsMixin { self: XMLSchemaDocument =>
...
+        // targetNamespace, then it can only reference elements that are imported into it
+        // without a namespace
+        NoNamespace
+      }
+      case Some(_) => Assert.impossible()
+      case None => {
+        // if we weren't included or imported it means we are the bootstrap schema. There are
+        // no references in a bootstrap schema, only an import of the real schema, so this
+        // value doesn't really matter and should never be used.
+        NoNamespace
```
mbeckerle, line 124: If it "should never be used" I'd expect codecov to tell us it is uncovered. But it doesn't. So I suspect the comment "should never be used" isn't quite right.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/GrammarTerm.scala
```diff
@@ -56,7 +56,7 @@ abstract class Gram(contextArg: SchemaComponent)
   final override lazy val tunable = context.tunable
 
   final override def namespaces = context.namespaces
-  final override def targetNamespace = context.targetNamespace
+  final override def noPrefixNamespace = context.noPrefixNamespace
```
mbeckerle, line 59: So this means at the level of Gram objects, nobody ever asks for noPrefixNamespace. At least in our tests. 

We should either have a test that exercises this, or we should remove the member.  It's likely this is just bypassed, as in the places that could call it are themselves calling noPrefixNamespace on the context object themselves.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/QNameBase.scala
```diff
@@ -518,28 +521,25 @@ protected trait RefQNameFactoryBase[T] {
   def resolveRef(
```
mbeckerle, line 521: Suggest improve the scaladoc above because this code is handling both prefixed and non-prefixed refs.  The existing scaladoc suggests this is only about QNames with an explicit prefix.

mbeckerle (changes requested):
Very minor changes mostly to comments/scaladoc are suggested, assuming I have understood this all properly. 

One or two code coverage improvements are needed, as I'm not sure why those things were not covered. It seems they should have been.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaDocIncludesAndImportsMixin.scala
stevedlawrence: It looks like codecov doesn't understand this code and can't properly detect that there is no coverage. I've confirmed that this bit of code is never hit. I'll change it to an Assert.impossible and update the comment.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/QNameBase.scala
stevedlawrence: Can you clarify what isn't right about this comment? My goal with this is to explain why we don't use the `noPrefixNamespace` parameter when calling resolveRef and instead pass in `NoNamespace`.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/QNameBase.scala
```diff
@@ -581,9 +585,9 @@ object RefQNameFactory extends RefQNameFactoryBase[RefQName] {
     val resolvedNS = (preNS, ns) match {
       case (None, UnspecifiedNamespace) => {
         // if neither prefix nor namespace was specified, use the default namespace (i.e.
-        // xmln="...") if defined. If not defined, use the targetNamespace.
+        // xmlns="...") if defined. If not defined, use the noPrefixNamespace.
         val defaultNS = resolveDefaultNamespace(scope, unqualifiedPathStepPolicy)
-        defaultNS.map(NS(_)).getOrElse(targetNamespace)
+        defaultNS.getOrElse(noPrefixNamespace)
```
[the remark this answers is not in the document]
stevedlawrence: This entire function has zero coverage and is not used anywhere. This is just dead code that I think we should probably just remove.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/GrammarTerm.scala
stevedlawrence: It looks like not only does nothing ask for `noPrefixNamespace` at the level of `Gram` objects, nothing also asks for `namespaces` and `unqualifiedPathStep` policy, two things also needed for resolving references. It looks like codecov only sees no coverage on noPrefixNamespace for some reason.

I think what that means is we need to move `ResolvesQNames` to a trait down a level so that is only on a DSOM objects but not any Gram objects. Which I think makes sense? DSOM should be resolving everything and then Grams are created based on those results.

I think that means `ResolvesQNames` should be on `SchemaComponent` instead. I'll move it and see what happens.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/QNameBase.scala
mbeckerle: Withdrawn.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/GrammarTerm.scala
mbeckerle: This makes sense to me. DSOM should be the only thing that knows anything about namespaces.  That said, some stuff happens at Gram level that probably should not.

mbeckerle (approved):
+1 with the various fixups (if feasible) to remove namespace stuff from Gram objects.

You might consider doing that in a separate PR if it starts to unravel too many things.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/GrammarTerm.scala
stevedlawrence: I was able to remove ResolvesQNames from the Gram level without issue, it was a relatively small change.

olabusayoT (approved):
Minor questions for clarity but everything looks good

+1

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -191,6 +191,7 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
       minimizedScope,
       defaultBitOrder,
       optPrimType,
+      targetNamespace,
```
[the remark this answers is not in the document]
stevedlawrence: This was removed by the original patch the avoid duplicating `targetNamespace` in the runtime data here:

https://github.com/apache/daffodil/commit/2590f54ce93e262d02660bd9d57bc7979a7d06d9#diff-65004632a05e8533fdb90c6f2fc0d9e840ffa93293ae171e54e1cd99591bc540L657

This PR replaces `targetNamespace` with `noPrefixNamespace` but we still need the original `targetNamespace`, so this change essentially reverts the original change back to where `targetNamespace` was.

PR #1315, opened 2024-09-24
Enable debug/trace for parse, unparse and test only

olabusayoT:
- make debug/trace non global and sub option for parse, unparse and test only
- update CLI tests

Deprecation/Compatibility:
--trace and --debug are no longer global arguments for daffodil, and are instead arguments for parse, unparse and test only.

DAFFODIL-1141

### DIFF
--- a/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
+++ b/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
@@ -323,23 +323,16 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
   shortSubcommandsHelp()
 
   // Global Options
-  val debug = opt[Option[String]](
-    argName = "file",
-    descr =
-      "Enable the interactive debugger. Optionally, read initial debugger commands from [file] if provided."
-  )(optionalValueConverter[String](a => a))
-  val trace = opt[Boolean](descr = "Run this program with verbose trace output")
   val verbose = tally(descr = "Increment verbosity level, one level for each -v")
   val version = opt[Boolean](descr = "Show Daffodil's version")
 
   // Parse Subcommand Options
   object parse extends scallop.Subcommand("parse") {
-    banner("""|Usage: daffodil parse (-s <schema> [-r <root>] | -P <parser>)
-              |                      [-c <file>] [-D<variable>=<value>...] [-I <infoset_type>]
-              |                      [-o <output>] [--stream] [-T<tunable>=<value>...] [-V <mode>]
-              |                      [infile]
+    banner("""|Usage: daffodil parse (-s <schema> | -P <parser>) [PARSE_OPTS] [--] [infile]
               |
               |Parse a file, using either a DFDL schema or a saved parser
+              |
+              |-- can be used to separate command-line options from trailing arguments
               |
               |Parse Options:""".stripMargin)
 
@@ -380,8 +373,9 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
     val rootNS = opt[RefQName](
       "root",
       argName = "node",
-      descr =
-        "Root element to use. Can be prefixed with {namespace}. Must be a top-level element. Defaults to first top-level element of DFDL schema."
+      descr = "Root element to use. Can be prefixed with {namespace}. " +
+        "Must be a top-level element. Defaults to first top-level element of DFDL schema. " +
+        "Only valid with the --schema option."
     )
     val schema =
       opt[URISchemaSource](
@@ -402,13 +396,24 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
       name = 'T',
       keyName = "tunable",
       valueName = "value",
-      descr = "Tunable configuration options to change Daffodil's behavior"
+      descr = "Tunable configuration options to change Daffodil's behavior. " +
+        "Only valid with the --schema option."
     )
     val validate: ScallopOption[ValidationMode.Type] = opt[ValidationMode.Type](
       short = 'V',
       default = Some(ValidationMode.Off),
       argName = "mode",
       descr = "Validation mode. Use 'on', 'limited', 'off', or a validator plugin name."
+    )
+    val debug = opt[Option[String]](
+      argName = "file",
+      descr =
+        "Enable the interactive debugger. Optionally, read initial debugger commands from [file] if provided. " +
+          "Cannot be used with --trace option."
+    )(optionalValueConverter[String](a => a))
+    val trace = opt[Boolean](
+      descr = "Run this program with verbose trace output. " +
+        "Cannot be used with the --debug option."
     )
 
     val infile = trailArg[String](
@@ -418,6 +423,8 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
 
     // must have one of --schema or --parser
     requireOne(schema, parser)
+
+    mutuallyExclusive(trace, debug) // cannot provide both --trace and --debug
 
     // if --parser is provided, cannot also provide --root or -T
     conflicts(parser, List(rootNS, tunables))
@@ -450,12 +457,11 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
 
   // Unparse Subcommand Options
   object unparse extends scallop.Subcommand("unparse") {
-    banner("""|Usage: daffodil unparse (-s <schema> [-r <root>] | -P <parser>)
-              |                        [-c <file>] [-D<variable>=<value>...] [-I <infoset_type>]
-              |                        [-o <output>] [--stream] [-T<tunable>=<value>...] [-V <mode>]
-              |                        [infile]
+    banner("""|Usage: daffodil unparse (-s <schema> | -P <parser>) [--] [infile]
               |
               |Unparse an infoset file, using either a DFDL schema or a saved parser
+              |
+              |-- can be used to separate command-line options from trailing arguments
               |
               |Unparse Options:""".stripMargin)
 
@@ -495,8 +501,9 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
     val rootNS = opt[RefQName](
       "root",
       argName = "node",
-      descr =
-        "Root element to use. Can be prefixed with {namespace}. Must be a top-level element. Defaults to first top-level element of DFDL schema."
+      descr = "Root element to use. Can be prefixed with {namespace}. " +
+        "Must be a top-level element. Defaults to first top-level element of DFDL schema. " +
+        "Only valid with the --schema option."
     )
     val schema =
       opt[URISchemaSource](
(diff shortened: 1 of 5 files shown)


mbeckerle (approved):
+1

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -617,7 +629,7 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
   // Test Subcommand Options
   object test extends scallop.Subcommand("test") {
     banner(
-      """|Usage: daffodil test [-I <implementation>] [-l] [-r] [-i] <tdmlfile> [testnames...]
+      """|Usage: daffodil test [-I <implementation>] [-l] [-r] [-i] [-d <file> | -d -- | -t] <tdmlfile> [testnames...]
```
stevedlawrence, line 632: `-d --` isn't really the right way to signify the ability to use `--`, since it sort of implies `--` is an argument to the `-d` option. But that's not really case. The `--` can go anywhere and just says that everything following it should not be treated as an option or an option argument, and should instead be treated as a trailing argument.

We happen to only really need it for the -d option since there's sometimes an ambiguity if the thing following it is the file with debug commands or a trailing argument. This ambiguity is whys why it's needed in some of our tests that do something like this:

```
daffodil test -d foo.tdml`
```
In this case you do need the `--` to say that foo.tdml is not actually an argument to the -d option, e.g.
```
daffodil test -d -- foo.tdml
```

But note that `--` isn't always necessary if there is no ambiguity. For example, if you do this
```
daffodil test -d -r foo.tdml 'test_foo_.*'
```
Then scallop knows that `-r` is an option so it can infer the `-d` option did not have an argument, and so the `--` isn't needed at all. This is also why you don't need `-d --` for the parse/unparse tests. In those tests there is no ambiguity because `-d` is immediately followed by another option.

But note that the same problem can still happen for the parse/unparse commands, it's not specific to the test command. For example, if you did this:
```
daffodil parse -s foo.dfdl.xsd -d input.bin
```
Scallop would think "input.bin" is the file of debug commands, there would be no trailing args, and so daffodil would parse stdin. Instead, you would need to do: 

```
daffodil parse -s foo.dfdl.xsd -d -- input.bin
```

The `--` is also useful in rare edge cases where the file you want to parse looks like an argument. For example, if you had a file that started with a hyphen you would also need to use `--`, e.g.
```
daffodil parse -s foo.dfdl.xsd -- --file_that_looks_like_an_argument.bin
```
(there's good reason we don't start files with hyphens).

So all this to say that `--` isn't really specific to `-d` so shouldn't be added to the usage like that. If we really want to make it clear that it can be used, probably the best way to do it is to add `[--]` before the trailing arguments. That would imply that `--` is optional, but if you want it to deal with ambiguity then put it before trailing arguments.

Git does something similar for commands like `git add`, and the man page even says:
```
       --
           This option can be used to separate command-line options from the list of files, (useful when filenames might be mistaken for command-line options).
```

I'm not sure if scallop would let us add that to the option description though.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -638,12 +650,22 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
       tally(descr = "Increment test result information output level, one level for each -i")
     val list = opt[Boolean](descr = "Show names and descriptions instead of running test cases")
     val regex = opt[Boolean](descr = "Treat <testnames...> as regular expressions")
+    val debug = opt[Option[String]](
+      argName = "file",
+      descr =
+        "Enable the interactive debugger. Optionally, read initial debugger commands from [file] if provided. " +
+          "To use debug without a file, one must supply '--' inplace of file " +
+          "i.e 'daffodil test -d -- path/to/tests.tdml'"
```
stevedlawrence, line 658: This description isn't quite right. See above comment.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -1012,30 +1033,41 @@ class Main(
     }
   }
 
-  def withDebugOrTrace(proc: DFDL.DataProcessor, conf: CLIConf): DFDL.DataProcessor = {
-    (conf.trace() || conf.debug.isDefined) match {
+  def withDebugOrTrace(
+                        proc: DFDL.DataProcessor,
+                        traceArg: ScallopOption[Boolean],
+                        debugArg: ScallopOption[Option[String]]
```
stevedlawrence, line 1039: Indention got weird here. scalafmt should fix it?

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -453,6 +456,7 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
     banner("""|Usage: daffodil unparse (-s <schema> [-r <root>] | -P <parser>)
               |                        [-c <file>] [-D<variable>=<value>...] [-I <infoset_type>]
               |                        [-o <output>] [--stream] [-T<tunable>=<value>...] [-V <mode>]
+              |                        [--trace | --debug <file>]
               |                        [infile]
```
stevedlawrence, line 460: These usage strings are starting to get a bit verbose. I'm wondering they are getting too complicated to be useful and if we should simplify the usage string to just something like this:

```
Usage: daffodil unparse (-s <schema> | -P <parser>) [UNPARSE_OPTS] [--] [infile]

Unparse Options:
  -- scallop generated stuff--
```

So it shows only the required and commonly used options (-s and -P) and trailing args, and anything else is described in more detail below that. We would probably want to also modify some descriptions to mention restrictions. For example, '--root' would need to say that it's only valid with the --schema option. But that and trace/debug mutual exclusion are the only restrictions I think.

It also makes the `[--]` stand out a bit more so might make it's availability more clear. And it means we don't have to update this usage and make it even more complicated every time we add a new option.

stevedlawrence (changes requested):
Useful change, just some comments about update the usage strings.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
olabusayoT: I decided to add the following comment to all the banners

`-- can be used to separate command-line options from trailing arguments`

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -402,14 +396,25 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
...
       short = 'V',
       default = Some(ValidationMode.Off),
       argName = "mode",
       descr = "Validation mode. Use 'on', 'limited', 'off', or a validator plugin name."
     )
+    val debug = opt[Option[String]](
+      argName = "file",
+      descr =
+        "Enable the interactive debugger. Optionally, read initial debugger commands from [file] if provided. " +
+          "Cannot be used with --trace option. They are mutually exclusive options."
```
stevedlawrence, line 412: Remove "they are mutually exclusive options, think "Cannot be used with ---trace option" is sufficient.

daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala
```diff
@@ -323,24 +323,17 @@ class CLIConf(arguments: Array[String], stdout: PrintStream, stderr: PrintStream
...
   object parse extends scallop.Subcommand("parse") {
-    banner("""|Usage: daffodil parse (-s <schema> [-r <root>] | -P <parser>)
-              |                      [-c <file>] [-D<variable>=<value>...] [-I <infoset_type>]
-              |                      [-o <output>] [--stream] [-T<tunable>=<value>...] [-V <mode>]
-              |                      [infile]
+    banner("""|Usage: daffodil parse (-s <schema> | -P <parser>) [PARSE_OPTS] [--] [infile]
               |
               |Parse a file, using either a DFDL schema or a saved parser
               |
+              |-- can be used to separate command-line options from trailing arguments
```
stevedlawrence, line 335: I'll open a ticket with scallop to see if they can add a way for "--" to show up in the --help output. It's usage is rare enough that it probably doesn't belong in the main description, but I agree that a description is needed. I think this is unfortunately the only place it can exist without a scallop change. This is fine for now, but hopefully scallop can support it.

stevedlawrence (approved):
+1, just some suggestions about removing extra verbosity. Keeping --help output as minimal as possible it important to keep it from being overwhelming.

mbeckerle:
Does trace in fact work on daffodil test? I'd swear I tried this the other day and got no trace output. Do you perhaps need to do daffodil -t trace -i (both -t and -i in the older syntax) to see the trace output? I definitely wanted this to work and was surprised when it did not. I could not tell you which daffodil version I was using. 3.5.0 or 3.8.0, but I'm not sure which.

olabusayoT:
> Does trace in fact work on daffodil test? I'd swear I tried this the other day and got no trace output. Do you perhaps need to do daffodil -t trace -i (both -t and -i in the older syntax) to see the trace output? I definitely wanted this to work and was surprised when it did not. I could not tell you which daffodil version I was using. 3.5.0 or 3.8.0, but I'm not sure which.

It definitely works for the cli...for both 3.5.0 and 3.8.0 without needing to do anything other than:
daffodil --trace parse -s schema.xsd data.bin 
or 
daffodil -t parse -s schema.xsd data.bin 

I also confirmed Runner.trace works on both 3.5.0 and 3.8.0

stevedlawrence:
I confirmed it also works with the `test` subcommand with 3.9.0, e.g. `daffodil --trace test foo.tdml`.

olabusayoT:
Oh I misunderstood your comment. Attempts to use "daffodil --trace test csv/test/csv.tdml" doesn't work @stevedlawrence  , what tdml file did you use?

olabusayoT:
> Oh I misunderstood your comment. Attempts to use "daffodil --trace test csv/test/csv.tdml" doesn't work @stevedlawrence , what tdml file did you use?

Ah looks like that's because csv.tdml fails. It works when I try "daffodil --trace test daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml" for 3.8.0 but not for 3.5.0

stevedlawrence:
For CSV I had to run it from the src/ directory so that Daffodil could find the non-embedded schema, e.g.

```
cd path/to/csv/src/
daffodil --trace test ../test/csv.tdml
```

mbeckerle:
I learned --trace or -t with daffodil test DOES work with 3.8.0, does NOT work with 3.5.0. 
```
mbeckerle@orca:~/dataiti/opensource/DFDLSchemas/faketdl$ daffodil --version
Apache Daffodil 3.5.0
mbeckerle@orca:~/dataiti/opensource/DFDLSchemas/faketdl$ daffodil -t test test/TestFakeTDL.tdml test_msg_01
[Pass] test_msg_01

Total: 1, Pass: 1, Fail: 0, Not Found: 0
mbeckerle@orca:~/dataiti/opensource/DFDLSchemas/faketdl$ daffodil -t test -i test/TestFakeTDL.tdml test_msg_01
[Pass] test_msg_01

Total: 1, Pass: 1, Fail: 0, Not Found: 0
```

PR #1317, opened 2024-09-25
Add new validation mode: xerces

olabusayoT:
- update tdml schema
- add new tests using none attribute to ensure no limited validation happens with xerces validation

DAFFODIL-2896

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
@@ -82,7 +82,7 @@ abstract class ElementParserBase(
       } else {
         val failureMessage = ccfResult.errMsg
         Logger.log.debug(
-          "Validation failed for ${context.toString} due to ${failureMessage}. The element was ${currentElement.namedQName}."
+          s"Validation failed for ${context.toString} due to ${failureMessage}. The element was ${currentElement.namedQName}."
         )
         pstate.validationError(
           "%s failed facet checks due to: %s",
@@ -199,7 +199,8 @@ abstract class ElementParserBase(
       if (pstate.processorStatus ne Success) return
 
       val shouldValidate = erd.isSimpleType &&
-        pstate.dataProc.isDefined && pstate.dataProc.value.validationMode != ValidationMode.Off
+        pstate.dataProc.isDefined && (pstate.dataProc.value.validationMode != ValidationMode.Off
+          && pstate.dataProc.value.validationMode != ValidationMode.Xerces)
 
       if (shouldValidate) {
         validate(pstate)
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceChildBases.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceChildBases.scala
@@ -530,7 +530,9 @@ trait EndArrayChecksMixin {
     if (state.processorStatus eq Success) {
 
       val shouldValidate =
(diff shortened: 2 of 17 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
```diff
@@ -199,7 +199,8 @@ abstract class ElementParserBase(
       if (pstate.processorStatus ne Success) return
 
       val shouldValidate = erd.isSimpleType &&
-        pstate.dataProc.isDefined && pstate.dataProc.value.validationMode != ValidationMode.Off
+        pstate.dataProc.isDefined && (pstate.dataProc.value.validationMode != ValidationMode.Off
+          && pstate.dataProc.value.validationMode != ValidationMode.Xerces)
```
jadams-tresys, line 203: Is this correct here as it seems that shouldValidate will always be false if ValidationMode.Xerces.  Is there a reason we need it to be false when validating with Xerces here?

Also this might be another case where a match statement makes things more clear.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SequenceChildBases.scala
```diff
@@ -530,7 +530,9 @@ trait EndArrayChecksMixin {
     if (state.processorStatus eq Success) {
 
       val shouldValidate =
-        state.dataProc.isDefined && state.dataProc.value.validationMode != ValidationMode.Off
+        state.dataProc.isDefined &&
+          (state.dataProc.value.validationMode != ValidationMode.Off
+            && state.dataProc.value.validationMode != ValidationMode.Xerces)
```
jadams-tresys, line 535: Same question here

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -928,8 +928,14 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
...
-        if (validationMode == ValidationMode.Full) false
-        else if (defaultValidationMode == ValidationMode.Full) false
+        if (
+          validationMode == ValidationMode.Full
+          || validationMode == ValidationMode.Xerces
+        ) false
+        else if (
+          defaultValidationMode == ValidationMode.Full
+          || defaultValidationMode == ValidationMode.Xerces
+        ) false
```
jadams-tresys, line 938: I think might be clearer in a match case:
```
(validationMode, defaultValidationMode) match {
    case (ValidationMode.Full | ValidationMode.Xerces, _) => false
    case (_, ValidationMode.Full | ValidationMode.Xerces) => false
    case (_, _) if (optExpectedWarnings.isDefined) => false
    case (_, _) => true
}
```

jadams-tresys (commented):
I may not be familiar with how the validation is handled, but I'm not seeing any code that is actual running the Xerces validation.  Is that going to be in a separate pull request or is that a completely external process?

stevedlawrence:
There are a number of comments in DAFFODIL-2896, we may want to consider more substantial changes to how validation works.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
olabusayoT: Yes, this shouldValidate is what controls whether we do limited validation, so this condition is saying do limited validation when validation mode is limited and full!

olabusayoT:
Closing this in favor of https://github.com/apache/daffodil/pull/1326

PR #1321, opened 2024-09-26
Warn about Multiple Child Elements with Same Name

olabusayoT:
- currently when we have multiple child elements in a sequence with the same name, daffodil allows it, while choice branches return a UPA error and unordered sequences return an SDE. We want to generate a warning when this happens in an ordered sequence so we refactor code to warn if the local names are the same. This has the effect of warning whether namespaces are supported or unsupported
- we group using elementChildren instead of groupMembers since we want to consider all children since regardless of the model groups, their immediate children end up as siblings in the infoset and can cause potential ambiguities for some infoset types
- add test to check we get UPA error with choices
- add tests to exercise new warning (same/different namespaces, but same local name)

DAFFODIL-2736

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
@@ -103,6 +103,7 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent, 
 
   requiredEvaluationsIfActivated(checkIfValidUnorderedSequence)
   requiredEvaluationsIfActivated(checkIfNonEmptyAndDiscrimsOrAsserts)
+  requiredEvaluationsIfActivated(checkIfMultipleChildrenWithSameName)
 
   protected def apparentXMLChildren: Seq[Node]
 
@@ -181,7 +182,7 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent, 
     if (!isOrdered) {
       checkMembersAreAllElementOrElementRef
       checkMembersHaveValidOccursCountKind
-      checkMembersHaveUniqueNamesInNamespaces
+      checkUnorderedSequenceMembersHaveUniqueNamesInNamespaces
     }
   }
 
@@ -215,20 +216,34 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent, 
     }
   }
 
-  private lazy val checkMembersHaveUniqueNamesInNamespaces: Unit = {
-    val childrenGroupedByQName = groupMembers.groupBy { gm =>
-      // previous checks should ensure that all group members are either local
-      // elements or element references
-      Assert.invariant(gm.isInstanceOf[ElementBase])
-      gm.asInstanceOf[ElementBase].namedQName
-    }
-    childrenGroupedByQName.foreach { case (qname, children) =>
-      if (children.length > 1) {
+  private lazy val checkUnorderedSequenceMembersHaveUniqueNamesInNamespaces: Unit = {
+    // previous checks should ensure that all element children for unordered sequences are either local
+    // elements or element references
+    elementChildrenInNonHiddenContext
+      .groupBy(_.namedQName)
+      .filter(_._2.length > 1)
+      .values
+      .foreach { children =>
         children.head.SDE(
           "Two or more members of an unordered sequence have the same name and the same namespace"
         )
       }
-    }
+  }
+
+  private lazy val checkIfMultipleChildrenWithSameName: Unit = {
+    elementChildrenInNonHiddenContext
+      .groupBy(_.name)
+      .filter(_._2.length > 1)
+      .values
+      .foreach { children =>
+        children.head.SDW(
+          WarnID.MultipleChildElementsWithSameName,
+          "Two or more members of the sequence have the same name. " +
+            "Since not all infosets support namespaces or siblings with the same name, " +
+            "this could cause issues. QNames are: %s",
+          children.map(c => c.namedQName).mkString(", ")
+        )
+      }
   }
 
   final lazy val isOrdered: Boolean = this.sequenceKind match {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ModelGroup.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ModelGroup.scala
@@ -341,6 +341,22 @@ abstract class ModelGroup protected (index: Int)
     }
   }
 
+  final lazy val elementChildrenInNonHiddenContext: Seq[ElementBase] =
+    LV('elementChildrenInNonHiddenContext) {
+      val ebs = groupMembers
+        .filter {
+          case gr: GroupRef => !gr.isHidden
+          case _ => true
+        }
+        .flatMap { gm =>
+          gm match {
+            case eb: ElementBase => Seq(eb)
+            case gb: ModelGroup => gb.elementChildren
+          }
+        }
+      ebs
+    }.value
+
   final lazy val elementChildren: Seq[ElementBase] = LV('elementChildren) {
     val gms = groupMembers
     val echls = gms.flatMap { gm =>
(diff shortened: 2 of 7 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -215,19 +216,39 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
...
+      )
     }
-    childrenGroupedByQName.foreach { case (qname, children) =>
-      if (children.length > 1) {
-        children.head.SDE(
-          "Two or more members of an unordered sequence have the same name and the same namespace"
-        )
+  }
+
+  private lazy val checkMembersHaveUniqueNamesInNamespaces: Map[Boolean, Seq[Term]] = {
```
stevedlawrence, line 229: Might suggest a different name for this since it isn't really checking anything, it's just grouping things.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -215,19 +216,39 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
...
+        gm.asInstanceOf[ElementBase].namedQName
       }
+    childrenGroupedByQName.map { case (qname, children) =>
+      (children.length > 1, children)
+    }
+  }
+
+  private lazy val checkIfMultipleChildrenWithSameName: Unit = {
+    val nonUniqueNameChildren =
+      checkMembersHaveUniqueNamesInNamespaces.filter(_._1 == true).values
```
stevedlawrence, line 245: I wonder if this needs a different grouping than with the unorderd sequences?

The goal of this check is to warn in case the schema is used with an infoset that don't support duplicate fields, and those kinds of infosets probably also don't support namespaces. So we probably want to warn even if they have different namespaces, which means we need to group by name only and not by QName.

Also, we kindof already have a similar check with `WarnID.NamespaceDifferencesOnly`. I wonder if that check should be removed in place of this one, since this new check is really just a more general version of that check? The existing one is about same-name elements next to each other, this one is about same-name elements that are siblings. Anything that this new check warns about will also warn for the existing check.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -215,19 +216,39 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
...
+    nonUniqueNameChildren.foreach { children =>
+      children.head.SDW(
+        WarnID.MultipleChildElementsWithSameName,
```
stevedlawrence, line 248: Maybe this wants to be WarnID.NamespaceDifferencesOnly?

olabusayoT: I think I prefer the idea of MultipleChildElementsWithSameName subsuming NamespaceDifferencesOnly instead of the other way around, because in sequence group, the namespaces could be different or the same with the same name and it should still trigger  MultipleChildElementsWithSameName. Actually I'm wondering if we ougt to leave NamespaceDifferenceOnly alone since it's very clear that it's for the case where the name is the same and the namespace is different, and MultipleChildElementsWithSameName is for the case where the name is the same regardless of whether the namespace is different or the same?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -215,19 +216,44 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
...
-        children.head.SDE(
-          "Two or more members of an unordered sequence have the same name and the same namespace"
-        )
+  }
+
+  private def groupedMembersWithSameName(
+    includeNamespace: Boolean = true
+  ): Map[Boolean, Seq[Term]] = {
+    val childrenGroupedByName = groupMembers
+      .filter { m => m.isInstanceOf[LocalElementDecl] || m.isInstanceOf[ElementRef] }
```
stevedlawrence, line 231: This filter is new from the old implementation. I think it's fine for the unordered sequences check since unordered sequences only allow local element decls or refs. But I don't think it works for the `checkIfMultipleChildrenWithSameName` check. Normal ordered sequences are allowed to have things other than elements (e.g. choice, group, sequence), and their immediate children need to be considered for the multiple children with same name check, since they all end up as siblings in the infoset of local element decl's and references and cause potential ambiguities for some infoset types.

I think we really need to consider all children regardless of model groups. We probably already have a member for this, we have lots of members about siblings and children. Maybe `elementChildren` is the right thing?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
stevedlawrence: I guess there are kindof few different cases that are are being considered for warnings here:

1) Infoset representations that do not support namespaces. We warn here if **immediate** siblings have the same **name** (but different namespaces). This is the existing NamespaceDifferencesOnly check.
2) Infoset representations that do not support siblings with the same name (but do support namespaces). We warn here if **any** siblings have the same **QName**.
3) Infoset representations that do not support namespaces OR siblings with the same name. We warn here if **any** siblings have the same **name** (but different namespaces).

I think we definitely need to warn for 3 (infosets that support neither namespaces nor same sibling names are common). If we implement 3, that will automatically match 1 and 2. If we have separate checks for 1, 2, and 3, we would end up with duplicate warnings. For example, something that warns for 3, will also warn for 2 and maybe 1 if the siblings are adjacent. Maybe this is fine, which would allow you to suppress 3 warnings but not 1, for example. Or maybe we only warn for 3, and the warning message can be descriptive enough to mention the different implications depending on what an infoset does and does no support?

Agreed NamespaceDifferencesOnly probably isn't the right id for this warning. There are a number of other kinds of things we'll eventually want to warn about that are all about related to infosets that don't support certain things, for example anonymous choices, nested choices, nullable elements. Do we want to eventually plan to warn for all of these kinds of tings under a single warn ID (e.g. LimitedInfosetRepresentation)? Or do we want specific IDs for each type of limitation?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
olabusayoT: I think elementChildren will work for the generic groupMembersBy since we do checkMembersAreAllElementOrElementRef before checkUnorderedSequenceMembersHaveUniqueNamesInNamespaces, so we won't get to the latter if the former fails. So I removed the filter and changed groupMembers to elementChildren

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
```diff
@@ -215,20 +216,37 @@ abstract class SequenceGroupTermBase(xml: Node, lexicalParent: SchemaComponent,
...
-      // elements or element references
-      Assert.invariant(gm.isInstanceOf[ElementBase])
-      gm.asInstanceOf[ElementBase].namedQName
-    }
-    childrenGroupedByQName.foreach { case (qname, children) =>
-      if (children.length > 1) {
+  private lazy val checkUnorderedSequenceMembersHaveUniqueNamesInNamespaces: Unit = {
+    // previous checks should ensure that all group members for unordered sequences are either local
+    // elements or element references
+    groupChildrenBy(_.namedQName)
```
stevedlawrence, line 222: I forget why we added the `groupChildrenBy` function, but it only replaces `elementChildren.groupBy`. The former isn't significantly shorter and I don't think it is any more clear. Is it worth removing `groupChildrenBy`?

stevedlawrence (approved):
+1 :+1: just a minor comment

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SequenceGroup.scala
olabusayoT: It used to do alot more than elementChildren.groupBy, but I moved out the filter stuff, since that wasn't related to grouping

jadams-tresys (approved):
+1

PR #1322, opened 2024-09-26
Add support for escalating SDWs to Errors

olabusayoT:
- add tunable 'escalateWarningsToErrors' to control escalation of all warnings to errors
- add test to demonstrate tunable

DAFFODIL-2810

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
@@ -24,7 +24,6 @@ import org.apache.daffodil.lib.exceptions.SchemaFileLocation
 import org.apache.daffodil.lib.exceptions._
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.Maybe._
-import org.apache.daffodil.runtime1.processors.CompileState
 import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
 
 class SchemaDefinitionError(
@@ -49,6 +48,20 @@ class SchemaDefinitionError(
 
 }
 
+class SchemaDefinitionErrorFromWarning(sdw: SchemaDefinitionWarning)
+  extends SchemaDefinitionWarning(
+    sdw.warnID,
+    sdw.schemaContext,
+    sdw.annotationContext,
+    sdw.kind,
+    sdw.args: _*
+  ) {
+
+  override def isError = true
+  override def modeName = super.modeName + " Warning Escalated"
+
+}
+
 /**
  * Specific class used for this specific error, because we need to pick this off
  * in the debugger for special handling.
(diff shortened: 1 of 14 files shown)


stevedlawrence (approved):
+1 :+1:

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
```diff
@@ -262,14 +262,26 @@ trait ImplementsThrowsOrSavesSDE extends ImplementsThrowsSDE with SavesErrorsAnd
...
-        warnID,
-        Some(schemaFileLocation),
-        NoAnnotationContext,
-        fmt,
-        args: _*
-      )
-      warn(sdw)
+      if (tunable.escalateWarningsToErrors) {
+        val msg = "warnings escalated to errors: " + fmt
+        val sde = new SchemaDefinitionError(
```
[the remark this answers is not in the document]
olabusayoT: I think includinf the warnID is a great idea

To add a new SDE class that supports both RuntimeSDW and regular SDW, we'd need to expose the arguments of SchemaDefinitionDiagnosticBase (SDDB) and make sure we can access them from the passed in sdw. Is this acceptable? I'm thinking adding vals to the args of SDDB and then doing something like:

```
class SchemaDefinitionErrorFromWarning(
  sdw: SchemaDefinitionDiagnosticBase
) extends SchemaDefinitionError(
  (if (sdw.sc.isDefined) Some(sdw.sc.get) else None),
...
) {
```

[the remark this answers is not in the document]
stevedlawrence: Yeah, something like that could probably work.

Another approach, could maybe just extend SchemaDefinitionWarning, and overide a few memers,e.g

```scala
class SchemaDefinitionErrorFromWarning(...) extends SchemaDefinitionWarning(...) {
  override def isError = "true"
  override def modeName = super.modeName + " Escalated"
}
```
I'm not sure what the best appraoch is, but something that can minimize duplication and will automatically inherit any future changes made to SDW would be nice.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
```diff
@@ -49,6 +49,26 @@ class SchemaDefinitionError(
 
 }
 
+class SchemaDefinitionErrorFromWarning(
+  sdw: SchemaDefinitionDiagnosticBase,
```
stevedlawrence, line 53: Can this sdw have the type of `SchemaDefinitionWaring` instead of `SchemaDefinitionDiagnosticBase`? This should only be used for warnings so seems like that would be on okay limitation? And would that simplify the `extends` at all? For example, we wouldn't need to pass in warnID since that's already available in sdw. Maybe it just becomes something like:
```scala
SchemaDefinitionErrorFromWarning(
  sdw: SchemaDefinitionWarning
) extends SchemaDefinitionWarning(
    sdw.warnId,
    sdw.schemaContext,
    sdw.annotationContext,
    sdw.kind,
    sdw.args
  )
```
Keeps things as simple as possible and makes the warning message the exact same as a normal warning, just with a different mode name to make it clear it's been escalated to an error?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
```diff
@@ -49,6 +49,26 @@ class SchemaDefinitionError(
...
+  warnID: WarnID
+) extends SchemaDefinitionWarning(
+    warnID,
+    sdw.sc.toScalaOption,
+    sdw.annotationContext.toScalaOption,
+    ({
+      val origMsg = if (sdw.mfmt.isDefined) sdw.mfmt.get else ""
+      val msg = s"warning escalated to error: ${warnID}. " + origMsg
```
stevedlawrence, line 61: I think the modeName is already "Schema Definition Warning Escalated Error". Could we remove the duplicate "warning escalated to error" in the message?

jadams-tresys: +1

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/ProcessorStateBases.scala
```diff
@@ -574,7 +575,15 @@ abstract class ParseOrUnparseState protected (
     if (!suppress) {
       val rsdw =
         new RuntimeSchemaDefinitionWarning(warnID, ctxt.schemaFileLocation, this, str, args: _*)
-      diagnostics = rsdw :: diagnostics
+      if (tunable.escalateWarningsToErrors) {
+        val sde = new SchemaDefinitionErrorFromWarning(
+          rsdw,
+          warnID
+        )
```
jadams-tresys, line 582: Just curious if this was auto-formatted to move everything to a new line.  Seems a little over-zealous in this case to me.

jadams-tresys (approved):
+1

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
olabusayoT: I don't think this would work for RuntimeSchemaDefinitionWarning since it extends SchemaDefinitionDiagnosticBase rather than SchemaDefinitionWarning

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/ProcessorStateBases.scala
olabusayoT: Oh no, I just wrote it that way...it wasn't autoformatted to look like that. I can change it to one line tho

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dsom/SDE.scala
stevedlawrence: We could probably change it to extend SDW if it makes sense.

olabusayoT: Done...I also made RSDE an SDE to keep things consistent

PR #1324, opened 2024-09-30
Generate warning when node indexed like array

olabusayoT:
- expression '.[...anything...]'' generates warning
- update tests with expression to demonstrate warning

DAFFODIL-2773

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
@@ -773,6 +773,16 @@ case class RelativePathExpression(stepsRaw: List[StepExpression], isEvaluatedAbo
 sealed abstract class StepExpression(val step: String, val pred: Option[PredicateExpression])
   extends Expression {
 
+  final override def text: String = step + pred.map(_.text).getOrElse("")
+
+  def checkIfNodeIndexedLikeArray(): Unit = {
+    schemaDefinitionWhen(
+      pred.isDefined,
+      "Indexing is only allowed on arrays. Offending path step: '%s'.",
+      this.text
+    )
+  }
+
   def relPathErr() = {
     // This path expression cannot be compiled because we went past the root. This normally
     // should be an SDE with a RelativePathPastRootError. However, if we don't have any element
@@ -1012,14 +1022,13 @@ sealed abstract class DownStepExpression(s: String, predArg: Option[PredicateExp
   }
 }
 
-// TODO: Is ".[i]" ever a valid expression in DFDL?
-// Perhaps. Doesn't work currently though. See DAFFODIL-2182
-
 sealed abstract class SelfStepExpression(s: String, predArg: Option[PredicateExpression])
   extends DownStepExpression(s, predArg) {
 
-  override lazy val compiledDPath = new CompiledDPath(SelfMove)
-  override def text = "."
+  override lazy val compiledDPath = {
+    checkIfNodeIndexedLikeArray()
+    new CompiledDPath(SelfMove)
+  }
 
   protected def stepElementDefs: Seq[DPathElementCompileInfo] = {
     if (this.isFirstStep) {
@@ -1037,7 +1046,7 @@ sealed abstract class SelfStepExpression(s: String, predArg: Option[PredicateExp
   }
 }
 
-case class Self(predArg: Option[PredicateExpression]) extends SelfStepExpression(null, predArg)
+case class Self(predArg: Option[PredicateExpression]) extends SelfStepExpression(".", predArg)
 
 /**
  * Different from Self in that it verifies the qName (s) is
@@ -1056,12 +1065,11 @@ case class Self2(s: String, predArg: Option[PredicateExpression])
 sealed abstract class UpStepExpression(s: String, predArg: Option[PredicateExpression])
   extends StepExpression(s, predArg) {
 
-  override def text = ".."
-
   final override lazy val compiledDPath = {
     val areAllArrays = isLastStep && stepElements.forall {
       _.isArray
     } && targetType == NodeInfo.Array
+    checkIfNodeIndexedLikeArray()
     if (areAllArrays) {
       new CompiledDPath(UpMoveArray)
     } else {
@@ -1092,7 +1100,7 @@ sealed abstract class UpStepExpression(s: String, predArg: Option[PredicateExpre
   override lazy val inherentType: NodeInfo.Kind = NodeInfo.Complex
 }
 
-case class Up(predArg: Option[PredicateExpression]) extends UpStepExpression(null, predArg)
+case class Up(predArg: Option[PredicateExpression]) extends UpStepExpression("..", predArg)
 
 /**
  * Different from Up in that it verifies the qName (s) is the
@@ -1100,8 +1108,6 @@ case class Up(predArg: Option[PredicateExpression]) extends UpStepExpression(nul
  */
 case class Up2(s: String, predArg: Option[PredicateExpression])
   extends UpStepExpression(s, predArg) {
-
-  override def text = ".."
 
   override protected def stepElementDefs: Seq[DPathElementCompileInfo] = {
     val cis = super.stepElementDefs
@@ -1146,23 +1152,14 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
         new DownArray(nqn)
       }
     } else {
-      //
-      // Note: DFDL spec allows a[exp] if a is not an array, but it's a processing
-      // error if exp doesn't evaluate to 1.
-      // TODO: Implement this.
-      if (pred.isDefined)
-        subsetError(
-          "Indexing is only allowed on arrays. Offending path step: '%s%s'.",
-          step,
-          pred.get.text
-        )
+      // a[exp] is only supported on arrays, because Daffodil no longer treats
+      // optional elements as arrays
+      checkIfNodeIndexedLikeArray()
       // the downward element step must be the same for all the possible elements, so
       // we can pick the first one arbitrarily.
       new DownElement(nqn)
     }
   }
-
-  override def text = step
 
   /*
    * The ERDs of the elements that correspond to this path step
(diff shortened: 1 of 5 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/CompiledExpression.scala
```diff
@@ -235,6 +236,16 @@ class ExpressionCompiler[T <: AnyRef] extends ExpressionCompilerBase[T] {
       compileInfoWhereExpressionWasLocated.SDE(msg, exprOrLiteral)
     }
 
+    if (expr.contains(".[") && expr.matches(".*\\.\\[.*].*")) {
```
stevedlawrence, line 239: I try to avoid string comparisons when possible. Especially with regular expression where it's easy to have unintentional matches. The DFDLExpressionParser already parses steps and predicates and turns them into objects, so it might be better to do this checks in the objects it creates.

For example the, `SelfStepExpression` class in Expression.scala has an `Option[PredicateExpression]`. Can we just do something like:
```scala
if (predArg.isDefined) SDW(...)
```
Do we want the same warning for `UpStepExpression`?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
```diff
@@ -721,6 +721,7 @@
           <xs:enumeration value="layerCompileWarning" />
           <xs:enumeration value="multipleChoiceBranches" />
           <xs:enumeration value="namespaceDifferencesOnly" />
+          <xs:enumeration value="nodeIndexedLikeArray" />
```
stevedlawrence, line 724: Note that the current check only detects when self expression is indexed like an array. It doesn't detect all single nodes indexed like an array. For example, you could have an expression like `foo/bar/baz[1]`, where baz is not an array. In that case we probably also want a similar `nodeIndexLikeArray` warning. We might want to also to detect those cases in this PR if it's similar enough, could be done separately though. I believe Expression.scala already knows which steps are arrays so it might be as simple as
```scala
if (!step.isArray && predArg.isDefined) SDW(...)
```

stevedlawrence (changes requested):
I think there might be a cleaner way to detect the self expression + predicate?

daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd
mbeckerle: Steve makes a good point. If you have elements and arrays, and you change the nesting, and in `foo/bar/baz[1]` baz was an array before, but now it's an enclosing scalar element or inside the array items, but it's not an array anymore, something else is, then, the expression wants to be broken (or at least warn) because you're not indexing the array anymore, you're indexing in a no-op manner, a scalar. 

I've actually made exactly this error, and had to search all over the place for where I was indexing the wrong element, having introduced a level of element someplace.

olabusayoT: Does this want to be an error or remain a warning? I found the following in the DFDL spec under Section 35 under Expression Errors. So should we be SDEing here if step is not an array or an optional element?

```
 Indexing of non-array non-optional element

§  Example: x[1] when x is declared and has both XSD minOccurs="1" and XSD maxOccurs="1" explicitly, or by not stating either or both of them.
```

stevedlawrence: It looks like we actually might already have the SDE check somewhere? If I create an optional element and try to index it with `[1]` I get the SDE:

> [error] Schema Definition Error: Subset: Indexing is only allowed on arrays. Offending path step: 'foo[1]'.

So somewhere we make sure you can only have a predicate on arrays, and we go one step further and say you can't even have it on optionals. I think maybe allowing it on optionals was mentioned in the spec because Daffodil used to treat optionals as arrays, but I don't think that's the case anymore?

If we are going to be consistent with that behavior, I think maybe `.[1]` and `..[1]` want to be changed to SDE's also, since self-step and up-step can never reference an whole array, they always reference a single element (which at best could be a single element of an array).

I grepped through the regression suite and I don't see any schemas that do something like .[1], so I don't think this is something we need to support. Maybe at some point it was needed, but that doesn't seem to be the case anymore.

mbeckerle: Agree. Let's change the goal of this ticket and make them SDEs. This notion that a DFDL implementation can just use XPath is a fiction that was never realistic, and it will be entirely moot once there is a non-XSD syntax for DFDL (which is planned for DFDL v2.0)

XPath was designed to work without schemas, as was almost everything about XML. In a strongly typed world with schemas, you don't need to have this much squishyness about what's meaningful, and what isn't.

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -1065,6 +1074,7 @@ sealed abstract class UpStepExpression(s: String, predArg: Option[PredicateExpre
     if (areAllArrays) {
       new CompiledDPath(UpMoveArray)
```
stevedlawrence, line 1075: Does the areAllArrays == true block need the predicate check too? This block means it the last up step with a targetType of array, so the expression could be something like `fn:count(..)`. If there is a predicate on this, e.g. `fn:count(..[1])` then I think that should have the same node-as-array an error since the the predicate would resolve to a single element of the array, and won't match the target type of array. I think? So it feels like regardless if the up-step is to an array or not we still never want to allow predicates?

Sidenote, UpMoveArray actually feels a little questionable based on these discussions--an up-step normally only goes to a single element so a special case where `..` goes to array feels odd? Seems like if you wanted the parents array you would have to do something like `fn:count(../../array)`. But maybe since we expect an array it's okay to assume the user meant the actual array and not the array element that up-step normally resolves to?

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -773,6 +773,15 @@ case class RelativePathExpression(stepsRaw: List[StepExpression], isEvaluatedAbo
 sealed abstract class StepExpression(val step: String, val pred: Option[PredicateExpression])
   extends Expression {
 
+  def checkIfNodeIndexedLikeArray(): Unit = {
+    if (pred.isDefined) {
+      SDE(
+        "Expression contains indexing on a node element: %s.",
```
stevedlawrence, line 779: I might recommend "non-array" instead of "node". I think we use that language in a few other places.

stevedlawrence (approved):
+1 looks good, maybe possible cleanup and a question about the UpArray case

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -773,6 +773,15 @@ case class RelativePathExpression(stepsRaw: List[StepExpression], isEvaluatedAbo
 sealed abstract class StepExpression(val step: String, val pred: Option[PredicateExpression])
   extends Expression {
 
+  def checkIfNodeIndexedLikeArray(): Unit = {
+    if (pred.isDefined) {
```
stevedlawrence, line 777: We could also use:
```scala
schemaDefinitionWhen(
  pred.isDefined,
  ...
)
```

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
```diff
@@ -1146,10 +1156,8 @@ case class NamedStep(s: String, predArg: Option[PredicateExpression])
         new DownArray(nqn)
       }
     } else {
-      //
-      // Note: DFDL spec allows a[exp] if a is not an array, but it's a processing
-      // error if exp doesn't evaluate to 1.
-      // TODO: Implement this.
+      // a[exp] is only supported on arrays , because Daffodil no longer treats
+      // optional elements as arrays
       if (pred.isDefined)
```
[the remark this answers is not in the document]
olabusayoT: It looks like subset errors were intended for things we chose not to implement until a future release...since we are not planning on implementing this functionality (i.e .[1] or ..[1] or OptionalElement[1]) in the future...I'll get rid of the subset error, but keep the original error message "Indexing is only allowed on arrays. Offending path step: '%s%s'."

jadams-tresys (approved):
+1

Agree with Steve's suggestions

daffodil-core/src/main/scala/org/apache/daffodil/core/dpath/Expression.scala
mbeckerle: Why allow "..[index]" at all? You could, in this case, do "../../array1_1[index]" I.e., use the array name. 
This is monomorphic in that the expression requires use of a specific array element name, but I'm not sure we lose much there. The only case I can think of where "../[index]" would be helpful is if you are creating some reusable element or group that wants to be polymorphic and work regardless of the name of the enclosing array.  

Polymorphism is good, but we have other reasons why we need "../a/*/b" i.e, wildcard stars, which is a more general way to get similar polymorphism. 

So I do not have a use case for this. "..[index]". It gives us some polymorphism without a need to extend the expression language with "*" wildcards. But it's only of theoretical concern unless there is some use case.

olabusayoT: Ok, so do we change `..` so it doesn't resolve to an array? Or should we leave it resolving to an array, but then disallow `..[index]` as Steve suggested?

stevedlawrence: I think that makes sense. So `..` should never allow a predicate, whether it's an array or not.

But I think that also means that we need to remove `UpMoveArray`? That might be a more involved change and need some good testing, probably wants to be a separate bug/PR.

PR #1326, opened 2024-10-02
Disable Limited Validation when Validation mode is Full

olabusayoT:
- update comments and tests

Deprecation/Compatibility:
Full Validation no longer performs limited validation in addition to Xerces validation. It now only performs Xerces validation.

DAFFODIL-2896


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/ElementCombinator1.scala
```diff
@@ -198,8 +198,11 @@ abstract class ElementParserBase(
 
       if (pstate.processorStatus ne Success) return
 
+      // if we should do limited validation via CheckConstraints
       val shouldValidate = erd.isSimpleType &&
-        pstate.dataProc.isDefined && pstate.dataProc.value.validationMode != ValidationMode.Off
+        pstate.dataProc.isDefined &&
+        pstate.dataProc.value.validationMode != ValidationMode.Off &&
+        pstate.dataProc.value.validationMode != ValidationMode.Full
```
stevedlawrence, line 205: Suggest we just do

```scala
pstate.dataProc.isDefined && pstate.dataProc.value.validationMode == ValidationMode.Limited
```
So a single positive test instead of a bunch of negative tests. If we use negative checks and ever add more validation modes the list of != could get kindof big. And this way if we ever add more validation modes, this won't have to change since this should only ever be enabled with limited validation.

stevedlawrence (approved):
+1 good change, and good for testing since it makes it much easier to check if validation errors come from daffodil or xerces

PR #1330, opened 2024-10-08
Fix NamespaceDifferencesOnly Warning when namespaces are same

olabusayoT:
- currently the code doesn't check if the namespaces are the same and reports that namespaces are different as long as the local names are duplicated. This is incorrect and we add a fix that only generates the error if the namespaces are actually different
- fix incorrect de-duplication of Sibs where we were overwriting the initial duplicates with the later duplicates. Now we prioritize the first "duplicates" over any secondary ones

DAFFODIL-2938

### DIFF
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/PartialNextElementResolver.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/PartialNextElementResolver.scala
@@ -304,9 +304,9 @@ class SeveralPossibilitiesForNextElement(
     hasNamespace: Boolean
   ): Maybe[ElementRuntimeData] = {
     Logger.log.debug(s"""
-      NextERDResovler -> trd: $trd\n
-      NextERDResovler -> looking for: $local\n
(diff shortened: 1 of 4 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/infoset/PartialNextElementResolver.scala
```diff
@@ -304,9 +304,9 @@ class SeveralPossibilitiesForNextElement(
     hasNamespace: Boolean
   ): Maybe[ElementRuntimeData] = {
     Logger.log.debug(s"""
-      NextERDResovler -> trd: $trd\n
-      NextERDResovler -> looking for: $local\n
-      NextERDResovler -> nextERDMap: $nextERDMap\n""")
+      NextERDResolver -> trd: $trd\n
+      NextERDResolver -> looking for: $local\n
+      NextERDResolver -> nextERDMap: $nextERDMap\n""")
```
jadams-tresys, line 309: I stared at this way too hard trying to figure out what changed here, I might be dyslexic

jadams-tresys (approved):
+1

PR #1331, opened 2024-10-08
Ensure hidden groups children excluded from siblings list

olabusayoT:
- currently we don't exclude the children of hidden groups from the siblings list. This is because we do the check for when a groupRef is hidden after ChoiceTermBase. The issue is ChoiceGroupRef and SequenceGroupRef, which are GroupRefs, also extend ChoiceTermBase and SequenceTermBase, so it's imperative that we do the hidden check before the *TermBases check
- add unit test to verify the hidden group's children are excluded from the sibling list

DAFFODIL-2937


jadams-tresys (approved):
+1

PR #1332, opened 2024-10-08
Add support for lengthUnit=bits for any type

olabusayoT:
- we currently report a warning when we use lengthUnit=bits with certain types ex date/time types. Per the discussions on dev (https://lists.apache.org/thread/01sn3fk2wk7jtos2svw4995f7dxxt7g1), we want to open up use of lengthUnit=bits to all types.
- verify test that uses lengthUnit=bits with calendar type no longer produces warning
- verify test that uses lengthUnit=bits with integer type no longer produces warning
- verify test that uses lengthUnit=bits with decimal type no longer produces warning
- deprecate allowBigIntegerBits tunable and the DeprecatedBigIntegerBits WarnID

Deprecation/Compatibility
- allowBigIntegerBits tunable no longer has any effect. All types are allowed to use lengthUnit=bits.
- WarnID DeprecatedBigIntegerBits has been deprecated. It is no longer in use in the codebase.

DAFFODIL-2931

### DIFF
--- a/daffodil-test/src/test/scala/org/apache/daffodil/section12/lengthKind/TestLengthKindExplicit.scala
+++ b/daffodil-test/src/test/scala/org/apache/daffodil/section12/lengthKind/TestLengthKindExplicit.scala
@@ -98,14 +98,33 @@ class TestLengthKindExplicit {
     runner.runOneTest("denseBit_lengthKind_explicit")
   }
 
-  @Test def test_invalidLengthUnitsIntegerWarning_explicit(): Unit = {
-    runner.runOneTest("invalidLengthUnitsIntegerWarning_explicit")
+  // DFDL-2931
+  @Test def test_lengthUnitsBitsForInteger_explicit(): Unit = {
+    runner.runOneTest("lengthUnitsBitsForInteger_explicit")
   }
-  @Test def test_invalidLengthUnitsIntegerError_explicit(): Unit = {
-    runner.runOneTest("invalidLengthUnitsIntegerError_explicit")
+  @Test def test_lengthUnitsBitsForInteger_explicit2(): Unit = {
+    runner.runOneTest("lengthUnitsBitsForInteger_explicit2")
   }
-  @Test def test_invalidLengthUnitsDecimalWarning_explicit(): Unit = {
-    runner.runOneTest("invalidLengthUnitsDecimalWarning_explicit")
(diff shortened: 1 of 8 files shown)


daffodil-test/src/test/scala/org/apache/daffodil/section12/lengthKind/TestLengthKindExplicit.scala
```diff
@@ -98,14 +98,14 @@ class TestLengthKindExplicit {
...
-  @Test def test_invalidLengthUnitsIntegerError_explicit(): Unit = {
-    runner.runOneTest("invalidLengthUnitsIntegerError_explicit")
+  @Test def test_lengthUnitsBitsForInteger_explicit2(): Unit = {
+    runner.runOneTest("lengthUnitsBitsForInteger_explicit2")
   }
-  @Test def test_invalidLengthUnitsDecimalWarning_explicit(): Unit = {
-    runner.runOneTest("invalidLengthUnitsDecimalWarning_explicit")
+  @Test def lengthUnitsBitsForDecimal_explicit(): Unit = {
+    runner.runOneTest("lengthUnitsBitsForDecimal_explicit")
   }
```
stevedlawrence, line 110: We might want some tests for xs:double, xs:float, and xs:date/time/dateTime with binaryCalendarRep set to binarySeconds and binaryMilliseconds.

For example, `binaryValue` in `ElementBaseGrammarMixin.scala` has logic for knownLength float/double, but I'm not sure it handles an expression for dfdl:length, unless that's handled somewhere else. Maybe we don't support expressions at all since those need to be 32/64 bits exactly?

Also, in that same val, the date/time/dateTime case errors that require lenghtUnits must be bytes for binarySeconds/Milliseconds, which this PR should fix, as long as the length is correct number of bits, we sholudn't care whether it was specified in bits or bytes.

stevedlawrence (approved):
+1, might need more tests and code changes since we've relaxed some restrictions

daffodil-test/src/test/scala/org/apache/daffodil/section12/lengthKind/TestLengthKindExplicit.scala
olabusayoT: Added additional tests, and we don't support runtime valued length for binary float/double.

jadams-tresys (approved):
+1

PR #1333, opened 2024-10-08
Add test as dependency to daffodil-test-integration

olabusayoT:
- update failing udf test

DAFFODIL-2933

### DIFF
--- a/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
+++ b/daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
@@ -47,7 +47,7 @@ class TestCLIUdfs {
       "daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd"
     )
 
-    runCLI(args"-v parse -s $schema -r fn_func") { cli =>
+    runCLI(args"-v parse -s $schema -r fn_func", fork = true) { cli =>
       cli.send("strng", inputDone = true)
       cli.expect("<fn_func>")
       cli.expect("<data>strng</data>")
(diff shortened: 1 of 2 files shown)


daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
```diff
@@ -64,8 +64,9 @@ class TestCLIUdfs {
     val schema = path(
       "daffodil-udf/src/test/resources/org/apache/daffodil/udf/genericUdfSchema.xsd"
     )
+    val classpath = udfClasspath()
 
-    runCLI(args"-v parse -s $schema -r user_func1") { cli =>
+    runCLI(args"-v parse -s $schema -r user_func1", classpath) { cli =>
```
stevedlawrence, line 69: Was it intentional to not add the udfClasspath for this tests since this test is about not having any UDF's? Probably worth adding a comment to make that clear.

Though, I would have expected the CLI to succeed with UDF's on the classpath so maybe I'm missing something?

stevedlawrence (approved):
+1, thought it's not clear to me why the test needed to be updated

daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLIUdfs.scala
olabusayoT: So I think the addition of the daffodil-test as a dependency caused the internal classpath to have the UDFs available which caused the test to fail with the changes, so we had to explicitly set the the udfClasspath to be empty for it to pass.

Before we didn't need to because the daffodil-test (which contains the UDF dependencies was missing)

I can update with a comment for clarity though

jadams-tresys (approved):
+1

PR #1334, opened 2024-10-09
Add annotations and facet check for PrefixLengthType

olabusayoT:
- currently we silently ignore facets on the prefixLengthType simple type and we don't execute asserts that try to force facet checking, thereby not doing any validation on prefixLengthType. This ticket aims to fix that issue by  performing the checks and SDEing when we have statement annotation on the simpleType of the prefixLengthType, which are not supported according to the DFDL workgroup.
- per the comments on the ticket, PrefixLengthType simpletypes cannot have annotations (SDE) and it is a processing error if their value fails facet checks

DAFFODIL-2660

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
@@ -192,6 +192,12 @@ trait ElementBaseGrammarMixin
         "%s is specified as a dfdl:prefixLengthType, but specifies a dfdl:trailingSkip other than 0",
         prefixLengthType
       )
+      schemaDefinitionWhen(
+        detachedElementDecl.statements.nonEmpty,
+        "%s is specified as a dfdl:prefixLengthType, but specifies one or more statement annotations (%s)",
+        prefixLengthType,
+        detachedElementDecl.statements.mkString(", ")
+      )
 
       if (
         detachedElementDecl.lengthKind == LengthKind.Prefixed &&
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
@@ -339,6 +339,20 @@ trait KnownPrefixedLengthUnparserMixin {
 
     plElement.setDataValue(java.lang.Integer.valueOf(adjustedLenInUnits.toInt))
 
+    // do checks on facets expressed on prefixLengthType
+    val optSTRD = plElement.erd.optSimpleTypeRuntimeData
+    if (optSTRD.isDefined) {
+      val strd = optSTRD.get
+      val check = strd.executeCheck(plElement)
+      if (check.isError) {
+        UnparseError(
+          One(state.schemaFileLocation),
+          One(state.currentLocation),
+          s"The calculated value of ${prefixedLengthERD.namedQName} ($adjustedLenInUnits) failed check due to ${check.errMsg}"
+        )
+      }
+    }
+
     // unparse the prefixed length element
     state.currentInfosetNodeStack.push(One(plElement))
     prefixedLengthUnparser.unparse1(state)
@@ -389,6 +403,19 @@ trait CalculatedPrefixedLengthUnparserMixin {
     }
     val adjustedLenInUnits = lenInUnits + prefixedLengthAdjustmentInUnits
     plElem.setDataValue(java.lang.Integer.valueOf(adjustedLenInUnits.toInt))
+    // do checks on facets expressed on prefixLengthType
+    val optSTRD = plElem.erd.optSimpleTypeRuntimeData
+    if (optSTRD.isDefined) {
+      val strd = optSTRD.get
+      val check = strd.executeCheck(plElem)
+      if (check.isError) {
+        UnparseError(
+          One(state.schemaFileLocation),
+          One(state.currentLocation),
+          s"The calculated value of ${elem.namedQName} ($adjustedLenInUnits) failed check due to ${check.errMsg}"
+        )
+      }
+    }
   }
 }
 
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
@@ -103,6 +103,18 @@ trait PrefixedLengthParserMixin {
         "Prefixed length result must be non-negative after dfdl:prefixIncludesPrefixLength adjustment , but was: %d",
         adjustedLen
       )
+      // do checks on facets expressed on prefixLengthType
+      val optSTRD = plElement.erd.optSimpleTypeRuntimeData
+      if (optSTRD.isDefined) {
+        val strd = optSTRD.get
+        val check = strd.executeCheck(plElement)
+        if (check.isError) {
+          val pe = state.toProcessingError(
+            s"The value of ${prefixedLengthERD.namedQName} ($parsedLen) failed check due to ${check.errMsg}"
+          )
+          state.setFailed(pe)
+        }
+      }
       adjustedLen
     } else {
       // Return zero if there was an error parsing the prefix length, the caller of this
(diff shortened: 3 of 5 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -192,6 +192,11 @@ trait ElementBaseGrammarMixin
         "%s is specified as a dfdl:prefixLengthType, but specifies a dfdl:trailingSkip other than 0",
         prefixLengthType
       )
+      schemaDefinitionWhen(
+        detachedElementDecl.statements.nonEmpty,
+        "%s is specified as a dfdl:prefixLengthType, but specifies one or more statement annotations",
```
stevedlawrence, line 197: Can you update the message to include which "statement annotations" it contains. Users that aren't familiar with the term "statement annotation" (which sounds kindof generic so is understandable) might be confused what the problem is. If we instead say it contains "dfdl:assert", for example, that would be more useful.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -103,6 +103,18 @@ trait PrefixedLengthParserMixin {
...
         adjustedLen
       )
+      // do checks on facets expressed on prefixLengthType
+      val optSTRD = plElement.erd.optSimpleTypeRuntimeData
+      if (optSTRD.isDefined) {
+        val strd = optSTRD.get
+        val check = strd.executeCheck(plElement)
+        if (check.isError) {
+          val pe = state.toProcessingError(
+            s"The calculated value of ${prefixedLengthERD.namedQName} ($adjustedLen) failed check due to ${check.errMsg}"
```
stevedlawrence, line 113: I wonder if this message should include `parsedLen` instead of `adjustedLen`?

For example, say we have `prefixIncludesPrefixLength="yes"` in the `PrefixFacetsCheck` test below. In that case the adjusted length would be 4 (since we need to subtract the length of the 1-byte prefix element). The facet check would still fail since it's checked against the parsedLength in the infoset (5), but we would output an error like"

>  s"The calculated value of prefix2 (4) failed check due to facet maxInclusive (4)"

Which would be confusing since that looks fine.

That said, maybe the facet checks should be run on the adjusted value and not the parsed value? In which case we would need to change the infoset value to the adjusted length prior to calling executeCheck. I think the spec implies facet checks are run on the parsed value since there's no mention of adjustment.

@mbeckerle, can you clarify?

stevedlawrence (approved):
+1, minor suggested to make diagnostics a little more clear, also wondering if we need a spec clarification if facet checks are done on pre or post adjustment values.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
mbeckerle: Hmmm. DFDL Workgroup discussion did not catch this level of detail that I recall. I'll search around a bit or ask and report back.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -194,8 +194,9 @@ trait ElementBaseGrammarMixin
       )
       schemaDefinitionWhen(
         detachedElementDecl.statements.nonEmpty,
-        "%s is specified as a dfdl:prefixLengthType, but specifies one or more statement annotations",
-        prefixLengthType
+        "%s is specified as a dfdl:prefixLengthType, but specifies one or more statement annotations (%s)",
+        prefixLengthType,
+        detachedElementDecl.statementsAsStrings.mkString(", ")
```
stevedlawrence, line 199: I believe `toString` on any DFDLAnnotation already does this conversion to a pretty string, so I don't think we need `statementsAsString`. I *think* you can just do 
```scala
detachedElementDecl.statemtents.mkString(", ")
```
And it'll do the right thing.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -103,6 +103,18 @@ trait PrefixedLengthParserMixin {
...
         adjustedLen
       )
+      // do checks on facets expressed on prefixLengthType
+      val optSTRD = plElement.erd.optSimpleTypeRuntimeData
+      if (optSTRD.isDefined) {
+        val strd = optSTRD.get
+        val check = strd.executeCheck(plElement)
+        if (check.isError) {
+          val pe = state.toProcessingError(
+            s"The calculated value of ${prefixedLengthERD.namedQName} ($parsedLen) failed check due to ${check.errMsg}"
```
mbeckerle, line 113: This check is also supposed to happen on unparsing. 

We don't normally validate facets while unparsing (yet), so it might not be quite so easy as the parse side. 
If deferring the unparse feature, please add a JIRA ticket.

mbeckerle (approved):
+1 with comments

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
mbeckerle: Note, due to some issues with the DFDL-WG email lists, this inquiry to the group did not get sent until morning of 2024-10-14.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
olabusayoT: It does indeed!

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
olabusayoT: Added in validation check for CalculatedPrefixedLengthUnparserMixin and KnownPrefixedLengthUnparserMixin

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
```diff
@@ -339,6 +339,19 @@ trait KnownPrefixedLengthUnparserMixin {
...
+    // do checks on facets expressed on prefixLengthType
+    val optSTRD = plElement.erd.optSimpleTypeRuntimeData
+    if (optSTRD.isDefined) {
+      val strd = optSTRD.get
+      val check = strd.executeCheck(plElement)
+      if (check.isError) {
+        val ue = state.toProcessingError(
+          s"The calculated value of ${prefixedLengthERD.namedQName} ($adjustedLenInUnits) failed check due to ${check.errMsg}"
+        )
+        state.setFailed(ue)
```
stevedlawrence, line 351: I'm not sure setFailed does the right thing when unparsing? Uparsing doesn't have a concept of backtracking so I think any processing errors should be thrown. I'm not sure setFailed does that. I think most unparsers use the `UE` function?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
olabusayoT: @mbeckerle , any word from the WG on this?

mbeckerle: > @mbeckerle , any word from the WG on this?

No, I've not seen any comments on my email on this subject. 

I think we should just proceed with the decision that the facets are about the value that is stored in the prefix integer, not the length of the data. So if prefixIncludesPrefixLength is 'true' the value stored is the length of the data plus the length of the prefix, and that sum is what the facets are constraining. 

So suppose the data is a string with a 2 byte unsigned prefix and prefixIncludesPrefixLength is true and the max string length is 1000. Then I believe these things hold:

- when parsing: a prefix value of 0 or 1 is malformed and causes a parse error
- when parsing: a prefix value of 2 means zero-length
- The maxInclusive facet should be 1002, as that means a string  of maximum length 1000. 
- a prefix value of 1003 or above is malformed and causes a parse error.

mbeckerle: DFDL Workgroup has discussed and there is general agreement this is the correct interpretation. 
Spec will be clarified. See: https://github.com/OpenGridForum/DFDL/issues/64

PR #1335, opened 2024-10-10
Fix Resolve Schema Location for xsi:SchemaLocation in Config files

olabusayoT:
- currently we check resolvedUri from namespaces endWith the passed in systemId. With the Xerces bug, the absolutized systemId will never match, so we convert the systemId to a URI and grab its path, and use that for the comparison
- add comment about specific validation error from xerces that requires workaround
- add schema location to config to prove it doesn't cause failure

DAFFODIL-2339

### DIFF
--- a/daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilXMLLoader.scala
+++ b/daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilXMLLoader.scala
@@ -216,12 +216,41 @@ class DFDLCatalogResolver private ()
     // because the nsURI will resolve to the including schema file.
     // This will cause the including schema to be repeatedly parsed resulting in a stack overflow.
 
+    lazy val systemIdUri = if (systemId != null) {
+      new URI(systemId)
+    } else {
+      null
+    }
+
+    /**
+     * Xerces has a bug where it absolutizes systemId i.e the user supplies
+     *  {{{
+     *    <xs:schema...
+     *    ... xsi:schemaLocation="urn:some:namespace /some/path.xsd"
+     *  }}}
+     * Xerces takes that schemaLocation URI and absolutizes it to {{{ file:/some/path.xsd }}}
+     * and passes that to our resolveEntity and in turn resolveCommon, which while it's able
+     * to find the namespace, fails to set the resolvedUri since the file:/some/path.xsd will
+     * never match anything resolved from our catalog since that'd return something like
+     * {{{ file:/some/absolute/path/to/some/path.xsd }}}
+     *
+     * This is a workaround to that bug where we convert systemId to a URI and check if the
+     * path (from URI.getPath) matches the end of resolvedUri. Note: This can ignore absolute
+     * URIs passed in for schemaLocation, but those are edge cases where the user expects
+     * the namespace to match a different file (i.e what they provide in the schemalocation)
+     * than what we find in the catalog.
+     */
+    lazy val systemIdPath = if (systemIdUri != null && systemIdUri.getScheme == "file") {
+      systemIdUri.getPath
+    } else {
+      systemId
+    }
     val resolvedId = {
       if (resolvedSystem != null && resolvedSystem != resolvedUri) {
         resolvedSystem
       } else if (
         resolvedUri != null && ((systemId == null) || (systemId != null && resolvedUri.endsWith(
-          systemId
+          systemIdPath
         )))
       ) {
         resolvedUri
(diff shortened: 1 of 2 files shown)


jadams-tresys (approved):
+1

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
```diff
@@ -1459,34 +1459,27 @@ Differences were (path, expected, actual):
...
-        // throw an IllegalArgumentException if that's not the case
-        val msg =
-          s"Non-absolute schemaLocation URI can only contain a path component: $schemaLocation"
-        throw new IllegalArgumentException(msg)
-      } else if (uri.getPath.startsWith("/")) {
+      }
+      // we want to attempt to resolve the URI whether the non-jar uri has a scheme or not,
+      // this is relevant for when we are validating with Xerces, and it calls resolvesEntity
+      // we get URIs that look like file:/path/to/not/absolute/path ex: file:/org/apache/daffodil/xsd/dafext.xsd
+      // that fail to be found in the above case, so we have to look them up
```
stevedlawrence, line 1481: Just to clarify, if we have something like this:
```
xsi:schemaLocation="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext /org/apache/daffodil/xsd/dafext.xsd"
```
Then Xerces changes that to `file:/org/apache/daffodil/xsd/dafext.xsd` and sends that to the EntityResolver? Feels like a bug in Xerces. It doesn't do that for include/import schemaLocation, not sure why it would do that for xsi:schemaLocation.

Note this also means means that if we ever have something legit like this:
```xml
<xs:include schemaLocation="file:/foo/bar/baz.dfdl.xsd">
```
Then we will look on the class path for `/foo/bar/baz.dfdl.xsd`, which feels wrong, we should only ever look at the filesystem for that kind of schemaLocation. Anything else feels like a bug. And I'm not sure I want to trade a Xerces bug for a Daffodil bug.

I'm wondering if we should that xsi:schemaLocation won't resolve anything to a classpath, unless there is someway we can get Xerces to not munge the xsi:schemaLocation. Or maybe we have a bug somewhere else where we are doing that munging?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
```diff
@@ -1459,34 +1459,27 @@ Differences were (path, expected, actual):
...
-        uri.getAuthority == null &&
-        uri.getQuery == null &&
-        uri.getFragment == null &&
-        uri.getPath != null
-
     val optResolved: Option[(URISchemaSource, Boolean)] =
-      if (uri.isAbsolute) {
-        // an absolute URI is one with a scheme. In this case, we expect to be able to resolve
-        // the URI and do not try anything else (e.g. filesystem, classpath). Since this function
+      if (uri.isAbsolute && uri.getScheme.contains("jar")) {
```
stevedlawrence, line 1463: This should be `uri.getScheme == "jar"`. Avoids ever accidentally matching some other scheme like "foojar" if that ever happens.

stevedlawrence (commented):
Feels like a Xercs bug?

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/XMLUtils.scala
olabusayoT: So I looked into this a bit more, and the conversion to a URI is definitely coming from Xerces, though I'm not sure that's a bug. According to this link, the location is supposed to be a URI...so it might be a Daffodil Bug to not properly interprete it. Thoughts? https://www.w3.org/TR/xmlschema-1/#schema-loc

stevedlawrence: `/foo/bar/baz.dfdl.xsd` is a valid URI. I'm not sure Xerces should assume the URI references a file on the filesystem. Feels like it should be up to the entity resolver to determine how to interpret the URI.

daffodil-lib/src/main/scala/org/apache/daffodil/lib/xml/DaffodilXMLLoader.scala
```diff
@@ -221,7 +221,7 @@ class DFDLCatalogResolver private ()
         resolvedSystem
       } else if (
         resolvedUri != null && ((systemId == null) || (systemId != null && resolvedUri.endsWith(
-          systemId
+          (new URI(systemId)).getPath
```
stevedlawrence, line 224: We need a detailed comment that we do this to get around Xerces and why this is okay. Maybe even an example of schema in the code so it's very clear. At face value it looks wrong since if someone includes a schemaLocation of `file:/foo/bar/baz` we will ignore the fact that this is an absolute URI as long as the path matches the resolveUri.

We should probably also only do this if the URI is an absolute `file:` URI, to make it clear it is only trying to work around a very specific Xercses bug and wont' accidentally break other cases. For example, if the URI is a jar: uri then we probably don't want to do this logic.

stevedlawrence (approved):
+1 but needs a comment why this workaround is okay

jadams-tresys (approved):
Adding a 2nd +1 since this went in a different direction from when I gave my first approval.

PR #1336, opened 2024-10-10
Fix Incorrect Lengths SDEs on Computed Element with LengthKind Implicit

olabusayoT:
- add condition that checks for IVC before doing the check
- add tests

DAFFODIL-2806

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
@@ -1007,13 +1007,19 @@ trait ElementBase
           "The length facet or minLength/maxLength facets are not allowed on types derived from type %s.\nThey are allowed only on types derived from string and hexBinary.",
           pt.name
         )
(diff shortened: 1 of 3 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -1007,13 +1007,19 @@ trait ElementBase
...
         )
-        val res = (hasLength, hasMinLength, hasMaxLength, lengthKind) match {
-          case (true, false, false, _) => (r.lengthValue, r.lengthValue)
-          case (true, _, _, _) =>
+        val res = (
+          hasLength,
+          hasMinLength,
+          hasMaxLength,
+          lengthKind,
+          inputValueCalcOption.isDefined
```
stevedlawrence, line 1015: It's better to use isRepresented to determine if something is IVC or not.

stevedlawrence (approved):
+1, minor cleanup to use isRepresented instead of explicitly checking the IVC property

mbeckerle (approved):
+1

jadams-tresys (approved):
+1

PR #1337, opened 2024-10-11
Parse Error on Out of Range Binary Integers

olabusayoT:
- add checks for signed/unsigned in BinaryIntegerBaseParser and BinaryIntegerBaseUnparser
- add tests for parse and unparse
- update failing tests

DAFFODIL-2297

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
@@ -774,11 +774,37 @@ trait ElementBase
     if (
       result.isDefined && repElement.isSimpleType && representation == Representation.Binary
     ) {
+      val nBits = result.get
       primType match {
         case primNumeric: NodeInfo.PrimType.PrimNumeric =>
-          if (primNumeric.width.isDefined) {
-            val nBits = result.get
-            val width = primNumeric.width.get
+          if (primNumeric.minWidth.isDefined) {
+            val minWidth = primNumeric.minWidth.get
+            if (nBits < minWidth) {
+              val isSigned = primNumeric.isSigned
+              val signedStr = if (isSigned) "a signed" else "an unsigned"
+              val outOfRangeFmtStr =
+                "Minimum length for %s binary integer is %d bit(s), number of bits %d out of range. " +
+                  "An unsigned integer with length 1 bit could be used instead."
+              if (isSigned && tunable.allowSignedIntegerLength1Bit && nBits == 1) {
+                SDW(
+                  WarnID.SignedBinaryIntegerLength1Bit,
+                  outOfRangeFmtStr,
+                  signedStr,
+                  minWidth,
+                  nBits
+                )
+              } else {
+                SDE(
+                  outOfRangeFmtStr,
+                  signedStr,
+                  minWidth,
+                  nBits
+                )
+              }
+            }
+          }
+          if (primNumeric.maxWidth.isDefined) {
+            val width = primNumeric.maxWidth.get
             if (nBits > width) {
               SDE(
                 "Number of bits %d out of range for binary %s, must be between 1 and %d bits.",
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
@@ -24,13 +24,16 @@ import org.apache.daffodil.io.FormatInfo
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.schema.annotation.props.gen.LengthUnits
 import org.apache.daffodil.lib.schema.annotation.props.gen.YesNo
+import org.apache.daffodil.lib.schema.annotation.props.gen.YesNo.Yes
 import org.apache.daffodil.lib.util.Maybe._
 import org.apache.daffodil.lib.util.MaybeInt
 import org.apache.daffodil.lib.util.Numbers._
+import org.apache.daffodil.runtime1.dpath.NodeInfo
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
 import org.apache.daffodil.runtime1.processors.Evaluatable
 import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
 import org.apache.daffodil.runtime1.processors.Processor
+import org.apache.daffodil.runtime1.processors.parsers.BinaryNumberCheckWidth
 import org.apache.daffodil.runtime1.processors.parsers.HasKnownLengthInBits
 import org.apache.daffodil.runtime1.processors.parsers.HasRuntimeExplicitLength
 import org.apache.daffodil.runtime1.processors.unparsers._
@@ -56,12 +59,7 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
     val nBits = getBitLength(state)
     val value = getNumberToPut(state)
     val dos = state.dataOutputStream
-    val res =
-      if (nBits > 0) {
-        putNumber(dos, value, nBits, state)
-      } else {
-        true
-      }
+    val res = putNumber(dos, value, nBits, state)
 
     if (!res) {
       Assert.invariant(dos.maybeRelBitLimit0b.isDefined)
@@ -78,8 +76,11 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
 
 }
 
-abstract class BinaryIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
-  extends BinaryNumberBaseUnparser(e) {
+abstract class BinaryIntegerBaseUnparser(e: ElementRuntimeData)
+  extends BinaryNumberBaseUnparser(e)
+  with BinaryNumberCheckWidth {
+
+  private val primNumeric = e.optPrimType.get.asInstanceOf[NodeInfo.PrimType.PrimNumeric]
 
   override def putNumber(
     dos: DataOutputStream,
@@ -87,8 +88,18 @@ abstract class BinaryIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
+    val state = finfo.asInstanceOf[UState]
+    if (primNumeric.minWidth.isDefined) {
+      val minWidth = primNumeric.minWidth.get
+      val isSigned = primNumeric.isSigned
+      checkMinWidth(state, isSigned, nBits, minWidth)
+    }
+    if (primNumeric.maxWidth.isDefined) {
+      val maxWidth = primNumeric.maxWidth.get
+      checkMaxWidth(state, nBits, maxWidth)
+    }
     if (nBits > 64) {
-      dos.putBigInt(asBigInt(value), nBits, signed, finfo)
+      dos.putBigInt(asBigInt(value), nBits, primNumeric.isSigned, finfo)
     } else {
       dos.putLong(asLong(value), nBits, finfo)
     }
@@ -97,9 +108,8 @@ abstract class BinaryIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
 
 class BinaryIntegerKnownLengthUnparser(
   e: ElementRuntimeData,
-  signed: Boolean,
   override val lengthInBits: Int
-) extends BinaryIntegerBaseUnparser(e, signed)
+) extends BinaryIntegerBaseUnparser(e)
   with HasKnownLengthInBits {
 
   override lazy val runtimeDependencies = Vector()
@@ -108,10 +118,9 @@ class BinaryIntegerKnownLengthUnparser(
 
 class BinaryIntegerRuntimeLengthUnparser(
   val e: ElementRuntimeData,
-  signed: Boolean,
   val lengthEv: Evaluatable[JLong],
   val lengthUnits: LengthUnits
-) extends BinaryIntegerBaseUnparser(e, signed)
+) extends BinaryIntegerBaseUnparser(e)
   with HasRuntimeExplicitLength {
 
   override val runtimeDependencies = Vector(lengthEv)
@@ -122,11 +131,12 @@ class BinaryIntegerPrefixedLengthUnparser(
   override val prefixedLengthUnparser: Unparser,
   override val prefixedLengthERD: ElementRuntimeData,
   maybeNBits: MaybeInt,
-  signed: Boolean,
   override val lengthUnits: LengthUnits,
   override val prefixedLengthAdjustmentInUnits: Long
-) extends BinaryIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
+) extends BinaryIntegerBaseUnparser(e: ElementRuntimeData)
   with KnownPrefixedLengthUnparserMixin {
+
+  private val primNumeric = e.optPrimType.get.asInstanceOf[NodeInfo.PrimType.PrimNumeric]
 
   override def childProcessors: Vector[Processor] = Vector(prefixedLengthUnparser)
   override lazy val runtimeDependencies = Vector()
@@ -140,7 +150,7 @@ class BinaryIntegerPrefixedLengthUnparser(
       // bytes needed to represent the number
       val value = getNumberToPut(s.asInstanceOf[UState])
       val len = Math.max(asBigInt(value).bitLength, 1)
-      val signedLen = if (signed) len + 1 else len
+      val signedLen = if (primNumeric.isSigned) len + 1 else len
       (signedLen + 7) & ~0x7 // round up to nearest multilpe of 8
     }
   }
@@ -241,7 +251,8 @@ abstract class BinaryDecimalUnparserBase(
   e: ElementRuntimeData,
   signed: YesNo,
   binaryDecimalVirtualPoint: Int
-) extends BinaryNumberBaseUnparser(e) {
+) extends BinaryNumberBaseUnparser(e)
+  with BinaryNumberCheckWidth {
 
   override def getNumberToPut(state: UState): JNumber = {
     val node = state.currentInfosetNode.asSimple
@@ -268,6 +279,10 @@ abstract class BinaryDecimalUnparserBase(
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
-    dos.putBigInt(asBigInt(value), nBits, signed == YesNo.Yes, finfo)
+    val state = finfo.asInstanceOf[UState]
+    val isSigned: Boolean = signed == Yes
+    val minWidth: Int = if (isSigned) 2 else 1
+    checkMinWidth(state, isSigned, nBits, minWidth)
+    dos.putBigInt(asBigInt(value), nBits, isSigned, finfo)
   }
 }
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
@@ -534,7 +534,9 @@ object NodeInfo extends Enum {
     }
 
     trait PrimNumeric { self: Numeric.Kind =>
-      def width: MaybeInt
+      def isSigned: Boolean
+      def minWidth: MaybeInt
+      def maxWidth: MaybeInt
       def isValid(n: Number): Boolean
       protected def fromNumberNoCheck(n: Number): DataValueNumber
       def fromNumber(n: Number): DataValueNumber = {
@@ -648,7 +650,9 @@ object NodeInfo extends Enum {
       // toString would have a precision different than Float.MaxValue.toString
       override val minStr = "-" + JFloat.MAX_VALUE.toString
       override val maxStr = JFloat.MAX_VALUE.toString
-      override val width: MaybeInt = MaybeInt(32)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(32)
+      override val maxWidth: MaybeInt = MaybeInt(32)
     }
 
     protected sealed trait DoubleKind extends SignedNumeric.Kind
@@ -667,7 +671,9 @@ object NodeInfo extends Enum {
       override val max = JDouble.MAX_VALUE
       override val minStr = "-" + JDouble.MAX_VALUE.toString
       override val maxStr = JDouble.MAX_VALUE.toString
-      override val width: MaybeInt = MaybeInt(64)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(64)
+      override val maxWidth: MaybeInt = MaybeInt(64)
     }
 
     protected sealed trait DecimalKind extends SignedNumeric.Kind
@@ -686,8 +692,9 @@ object NodeInfo extends Enum {
           case _ => true
         }
       }
-
-      override val width: MaybeInt = MaybeInt.Nope
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt.Nope
+      override val maxWidth: MaybeInt = MaybeInt.Nope
 
       override def isInteger = false
     }
@@ -708,8 +715,9 @@ object NodeInfo extends Enum {
           case _ => true
         }
       }
-
-      override val width: MaybeInt = MaybeInt.Nope
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
+      override val maxWidth: MaybeInt = MaybeInt.Nope
 
       override def isInteger = true
     }
@@ -725,7 +733,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueLong = n.longValue
       override val min = JLong.MIN_VALUE
       override val max = JLong.MAX_VALUE
-      override val width: MaybeInt = MaybeInt(64)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
+      override val maxWidth: MaybeInt = MaybeInt(64)
     }
 
     protected sealed trait IntKind extends Long.Kind
@@ -739,7 +749,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueInt = n.intValue
       override val min = JInt.MIN_VALUE.toLong
       override val max = JInt.MAX_VALUE.toLong
-      override val width: MaybeInt = MaybeInt(32)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
+      override val maxWidth: MaybeInt = MaybeInt(32)
     }
 
     protected sealed trait ShortKind extends Int.Kind
@@ -753,7 +765,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueShort = n.shortValue
       override val min = JShort.MIN_VALUE.toLong
       override val max = JShort.MAX_VALUE.toLong
-      override val width: MaybeInt = MaybeInt(16)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
+      override val maxWidth: MaybeInt = MaybeInt(16)
     }
 
     protected sealed trait ByteKind extends Short.Kind
@@ -767,7 +781,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueByte = n.byteValue
       override val min = JByte.MIN_VALUE.toLong
       override val max = JByte.MAX_VALUE.toLong
-      override val width: MaybeInt = MaybeInt(8)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
+      override val maxWidth: MaybeInt = MaybeInt(8)
     }
 
     protected sealed trait NonNegativeIntegerKind extends Integer.Kind
@@ -786,8 +802,9 @@ object NodeInfo extends Enum {
         case f: JFloat if f.isInfinite || f.isNaN => false
         case _ => n.longValue >= 0
       }
-
-      override val width: MaybeInt = MaybeInt.Nope
+      override val isSigned: Boolean = false
+      override val minWidth: MaybeInt = MaybeInt(1)
+      override val maxWidth: MaybeInt = MaybeInt.Nope
 
       override def isInteger = true
     }
@@ -814,7 +831,9 @@ object NodeInfo extends Enum {
       }
       val max = new JBigInt(1, scala.Array.fill(8)(0xff.toByte))
       val maxBD = new JBigDecimal(max)
-      override val width: MaybeInt = MaybeInt(64)
+      override val isSigned: Boolean = false
+      override val minWidth: MaybeInt = MaybeInt(1)
+      override val maxWidth: MaybeInt = MaybeInt(64)
 
       override def isInteger = true
     }
@@ -834,7 +853,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueLong = n.longValue
       override val min = 0L
       override val max = 0xffffffffL
-      override val width: MaybeInt = MaybeInt(32)
+      override val isSigned: Boolean = false
+      override val minWidth: MaybeInt = MaybeInt(1)
+      override val maxWidth: MaybeInt = MaybeInt(32)
     }
 
     protected sealed trait UnsignedShortKind extends UnsignedInt.Kind
@@ -848,7 +869,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueInt = n.intValue
       override val min = 0L
       override val max = 0xffffL
-      override val width: MaybeInt = MaybeInt(16)
+      override val isSigned: Boolean = false
+      override val minWidth: MaybeInt = MaybeInt(1)
+      override val maxWidth: MaybeInt = MaybeInt(16)
     }
 
     protected sealed trait UnsignedByteKind extends UnsignedShort.Kind
@@ -862,7 +885,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueShort = n.shortValue
       override val min = 0L
       override val max = 0xffL
-      override val width: MaybeInt = MaybeInt(8)
+      override val isSigned: Boolean = false
+      override val minWidth: MaybeInt = MaybeInt(1)
+      override val maxWidth: MaybeInt = MaybeInt(8)
     }
 
     protected sealed trait StringKind extends AnyAtomic.Kind
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
@@ -26,6 +26,7 @@ import org.apache.daffodil.lib.equality.TypeEqual
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.MaybeChar
+import org.apache.daffodil.runtime1.dpath.NodeInfo
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
 import org.apache.daffodil.runtime1.processors.FieldDFAParseEv
 import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
@@ -40,18 +41,52 @@ trait PackedBinaryConversion {
   def toBigDecimal(num: Array[Byte], scale: Int): JBigDecimal
 }
 
+trait PackedBinaryLengthCheck {
+  def PE(state: PState, str: String, args: Any*): Unit
+  def checkLengthNotEqualToZero(nBits: Int, start: PState, packedType: String): Boolean = {
+    if (nBits == 0) {
+      PE(
+        start,
+        s"Number of bits %d out of range for a packed $packedType.",
+        nBits
+      )
+      false
+    } else {
+      true
+    }
+  }
+
+  def checkLengthIsMultipleOf4(nBits: Int, start: PState): Boolean = {
+    if ((nBits % 4) != 0) {
+      PE(
+        start,
+        "The given length (%s bits) must be a multiple of 4 when using packed binary formats",
+        nBits
+      )
+      false
+    } else {
+      true
+    }
+  }
+}
+
 abstract class PackedBinaryDecimalBaseParser(
   override val context: ElementRuntimeData,
   binaryDecimalVirtualPoint: Int
 ) extends PrimParser
-  with PackedBinaryConversion {
+  with PackedBinaryConversion
+  with PackedBinaryLengthCheck {
   override lazy val runtimeDependencies = Vector()
 
   protected def getBitLength(s: ParseOrUnparseState): Int
 
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    val lengthEqualsZero = !checkLengthNotEqualToZero(nBits, start, packedType = "decimal")
+    if (lengthEqualsZero) return
+    val lengthNotMultipleOf4 = !checkLengthIsMultipleOf4(nBits, start)
+    if (lengthNotMultipleOf4) return
+
     val dis = start.dataInputStream
 
     if (!dis.isDefinedForLength(nBits)) {
@@ -69,17 +104,28 @@ abstract class PackedBinaryDecimalBaseParser(
 }
 
 abstract class PackedBinaryIntegerBaseParser(
-  override val context: ElementRuntimeData,
-  signed: Boolean = false
+  override val context: ElementRuntimeData
 ) extends PrimParser
-  with PackedBinaryConversion {
+  with PackedBinaryConversion
+  with PackedBinaryLengthCheck {
   override lazy val runtimeDependencies = Vector()
 
+  val signed = {
+    context.optPrimType.get match {
+      case n: NodeInfo.PrimType.PrimNumeric => n.isSigned
+      // context.optPrimType can be of type date/time via ConvertZonedCombinator
+      case _ => false
+    }
+  }
   protected def getBitLength(s: ParseOrUnparseState): Int
 
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    val lengthEqualsZero = !checkLengthNotEqualToZero(nBits, start, packedType = "integer")
+    if (lengthEqualsZero) return
+    val lengthNotMultipleOf4 = !checkLengthIsMultipleOf4(nBits, start)
+    if (lengthNotMultipleOf4) return
+
     val dis = start.dataInputStream
 
     if (!dis.isDefinedForLength(nBits)) {
--- a/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedDecimalParsers.scala
+++ b/daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedDecimalParsers.scala
@@ -84,11 +84,10 @@ class PackedDecimalPrefixedLengthParser(
 
 class PackedIntegerRuntimeLengthParser(
   val e: ElementRuntimeData,
-  signed: Boolean,
   packedSignCodes: PackedSignCodes,
   val lengthEv: Evaluatable[JLong],
   val lengthUnits: LengthUnits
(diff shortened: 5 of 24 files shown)


daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -166,7 +166,22 @@ abstract class BinaryIntegerBaseParser(
...
+      PE(
+        start,
+        "Minimum length for an unsigned binary integer is 1 bit, number of bits %d out of range.",
+        nBits
+      )
+      return
+    }
     if (primNumeric.width.isDefined) {
       val width = primNumeric.width.get
       if (nBits > width)
```
stevedlawrence, line 187: Just below is a PE without a return, which I think is a bug? According to codecov we already have coverage for this, so I imagine the missing return just means we do extra unnecessary work to parse a number that is too big and then just ignore it. Can you confirm if this is needed and open a bug if so?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -93,6 +93,27 @@ abstract class BinaryIntegerBaseUnparser(e: ElementRuntimeData, signed: Boolean)
       dos.putLong(asLong(value), nBits, finfo)
     }
   }
+
+  override def unparse(state: UState): Unit = {
+    val nBits = getBitLength(state)
```
stevedlawrence, line 98: This is going to call getBitLength twice, once here and once when super is called. We should probably avoid that.

Also, it looks like we don't have any checks on the max size for numbers like we do in parse. Maybe we should just do all the min/max length checks in the main unparse function like we do with parse? Or maybe in BinaryIntegerBaseUnparser.putNumber?

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
```diff
@@ -134,7 +134,7 @@
     </element>
 
     <simpleType name="myBit" dfdl:lengthKind="explicit" dfdl:length="1">
-      <restriction base="xs:byte"/>
+      <restriction base="xs:unsignedByte"/>
```
stevedlawrence, line 137: I'm concerned we will have a lot of regressions due to this change, and wonder if we need a process of deprecating this via tunable. It might be worth running this change with the regression suite and seeing how many things rely on this behavior.

stevedlawrence (changes requested):

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
olabusayoT: Indeed it is, created DAFFODIL-2942

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: Hmm, the issue is putNumber doesn't have access to state, which we need for Unparse error and BinaryNumberBaseUnparser is used by non Binary Integer classes

mbeckerle: When there is no access to state, I think the technique is to throw a thin-exception to someplace that catches it that has the state needed to report the error in context properly.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
mbeckerle: Pretty sure we're going to need the compatibility tunable. 

Particularly in Cyberian applications, it's quite likely that users have schemas with signed integers of length 1 bit, since we've been tolerating that, and their application really doesn't care if the values were 0 and -1 or 0 and 1. 
But their test cases might have expected infoset XML that assumes that 0 and -1 are the values for a signed integer of length 1. 

Our regression test set doesn't include all schemas, and I think we have to assume many others are using DFDL now, and have large sets of schemas. (I recently learned of a group which has 300+ DFDL schemas that we have no access to.)

So  we will need a compatibility tunable like "allowSignedIntegerLength1Bit" or something like that. False means we strictly enforce min length 2 for signed integers. "true" means we issue a warning (which can be suppressed by usual means) but implement the current behavior (which I think creates values 0 and -1. 

Note that issuing a warning, then assuming the user intended to use an unsigned integer instead would be a change in behavior (producing values 0 and 1, not 0 and -1), so we definitely don't want to do that. 

The warning message should suggest converting to use an unsigned integer, and the SDE message if disallowing should also suggest that an unsigned integer is the fix for length 1 bit.

mbeckerle: Note that there are obscure corner cases involving charset encodings like "bits" where the number of bits could turn out to be 0 or 1 using lengthKind 'pattern'. Detecting these at runtime only is fine with me.  Ex: 
```
<element name="b" type="xs:byte" dfdl:representation="text" dfdl:encoding="bits" dfdl:lengthKind="pattern"
    dfdl:lengthPattern=".{0,1}(?=11111111)" />
```
That captures any zeros and ones followed by a run of 8 consecutive "1"s. 

This is just a corner case, and so long as we don't abort, and if the length turns out to be 0 or 1, we issue an error, then I'm good with it.

mbeckerle (changes requested):

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
olabusayoT: There were no regressions in our test suite, but Mike makes a great point, so I'll add in the tunable. Do we want to add in a tunable for zero length unsigned integers too i.e allowZeroLengthUnsignedInteger?

mbeckerle: No. Zero length unsigned integers always fail now (no value), so we don't want to start allowing them since their min length is supposed to be 1 bit.

mbeckerle: The place where regressions might be found is in running the big regression of 80-some schemas.

olabusayoT: > The place where regressions might be found is in running the big regression of 80-some schemas.

Right the tests in dfdl-regression-test (which is what I think you're referring to) all passed while using the snapshot version with this change

olabusayoT: > <element name="b" type="xs:byte" dfdl:representation="text" dfdl:encoding="bits" dfdl:lengthKind="pattern"
>     dfdl:lengthPattern=".{0,1}(?=11111111)" />

I'm not sure I follow. In this case, the code changes wouldn't apply since representation is text, not binary. If we have the following element e1

```
    <xs:element name="e1">
      <xs:complexType>
        <xs:sequence>
          <xs:element name="b" type="xs:byte" dfdl:representation="text" dfdl:encoding="X-DFDL-BITS-LSBF"
                      dfdl:lengthKind="pattern"
                      dfdl:lengthPattern=".{0,1}(?=11111111)" />
          <xs:element name="c" type="xs:string" dfdl:representation="text" dfdl:encoding="X-DFDL-US-ASCII-7-BIT-PACKED"
                      dfdl:lengthKind="explicit" dfdl:length="1" />
        </xs:sequence>
      </xs:complexType>
    </xs:element>
```
and the following data: 011111111, it parses fine and you get

<ex:e1 xmlns:ex="http://example.com"><ex:b>0</ex:b><ex:c></ex:c></ex:e1>

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/ElementBase.scala
```diff
@@ -774,10 +774,32 @@ trait ElementBase
...
+      schemaDefinitionWhen(
+        isUnsignedIntegerType && nBits < 1,
+        "Minimum length for an unsigned binary integer is 1 bit, number of bits %d out of range.",
+        nBits
+      )
       primType match {
         case primNumeric: NodeInfo.PrimType.PrimNumeric =>
           if (primNumeric.width.isDefined) {
-            val nBits = result.get
             val width = primNumeric.width.get
```
stevedlawrence, line 803: Thoughts on renaming `width` to `maxWidth` and adding a new `minWidth` member in NodeInfo. Avoids some of the special logic around figuring out if stuff is signed or not. And then checking maxWidth and minWidth is very similar logic, just something like
```scala
if (primNumeric.minWdith.isDefined) {
   val minWidth = primNumeric.minWidth.get
   if(nBits < minWidth) {
     SDE(...)
   }
}
```
And then the minWidth can be reused in the parser/unparser logic.

Maybe we also add a `signed` member so we don't have to have the isSigned/isUnsigned stuff? And then parser/unparsers, which should already have access to the primType can just reference the signed member? We pass around isSigned in a bunch of places, wondering if maybe that can all be simplified.

daffodil-test/src/test/resources/org/apache/daffodil/extensions/enum/enums.tdml
olabusayoT: > I think you can have `representation="binary"` with `lengthKind="pattern"`.

So we get an SDE if we try that. "Schema Definition Error: Binary data elements cannot have lengthKind='pattern'."

stevedlawrence: I guess I was wrong. Seems like lengthKind="pattern" can't be used to get around the binary length limit.

mbeckerle: If you use encoding="bits" it is representation="text" You are treating each bit as a letter/character. It's text.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -78,17 +76,61 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
...
+            "An unsigned integer with length 1 bit could be used instead."
+        if (isSigned && state.tunable.allowSignedIntegerLength1Bit) {
+          state.SDW(
+            WarnID.SignedBinaryIntegerLength1Bit,
+            outOfRangeFmtStr,
+            signedStr,
+            minWidth,
+            nBits
+          )
+          return false
```
stevedlawrence, line 107: Is this `return false` correct? If we are only warning then should we still unparse something? For example, if nBits is 1 and we are signed but allowSignedIntegerLength1Bit, then we should output 1 bit. Seems like we want to warn but then drop through and still put a number.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -163,7 +204,11 @@ class BinaryFloatUnparser(e: ElementRuntimeData) extends BinaryNumberBaseUnparse
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
-    dos.putBinaryFloat(asFloat(value), finfo)
+    if (nBits > 0) {
```
stevedlawrence, line 207: I believe binary float must be exactly 32 bits, anything else should be an error. In fact, I think we have a compile time restriction the binary float must have a known compile time length of 32 bits or its an SDE, so I think it's impossible for nBits to not be correct and we should just always call putBinaryFloat.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -180,7 +225,11 @@ class BinaryDoubleUnparser(e: ElementRuntimeData) extends BinaryNumberBaseUnpars
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
-    dos.putBinaryDouble(asDouble(value), finfo)
+    if (nBits > 0) {
+      dos.putBinaryDouble(asDouble(value), finfo)
```
stevedlawrence, line 229: Same as the float case, but it's always 64 bits.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -268,6 +317,10 @@ abstract class BinaryDecimalUnparserBase(
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
-    dos.putBigInt(asBigInt(value), nBits, signed == YesNo.Yes, finfo)
+    if (nBits > 0) {
+      dos.putBigInt(asBigInt(value), nBits, signed == YesNo.Yes, finfo)
+    } else {
+      true
```
stevedlawrence, line 323: Should this be `false`? I think the purpose of the bug is to make it change it so zero-length integers are now an error? If you put the nBits == 0 check in the `unparse` method we can handle this case for all binary numbers.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -268,6 +317,10 @@ abstract class BinaryDecimalUnparserBase(
     nBits: Int,
     finfo: FormatInfo
   ): Boolean = {
-    dos.putBigInt(asBigInt(value), nBits, signed == YesNo.Yes, finfo)
+    if (nBits > 0) {
+      dos.putBigInt(asBigInt(value), nBits, signed == YesNo.Yes, finfo)
```
stevedlawrence, line 321: Does BinaryDecimal have the same restriction about signed/unsigned and 1-bit/2-bit? For example, if `signed == Yes` and `nBits == 1`, should that be a warning or UE depending on the tunable value?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -56,12 +58,8 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
     val nBits = getBitLength(state)
     val value = getNumberToPut(state)
     val dos = state.dataOutputStream
-    val res =
-      if (nBits > 0) {
-        putNumber(dos, value, nBits, state)
-      } else {
-        true
-      }
+    val res = putNumber(dos, value, nBits, state)
```
stevedlawrence, line 61: Seems this should be
```scala
if (nBits == 0) UE(...)
else putNumber(...)
```
I think the original purpose of this bug was to make it so a length of 0 changes from being ignored to being an error. Does it make sense to handle the zero case here and handle the min/maxWidth checks in putNumber, since those are specific to the number type?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -78,17 +76,61 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
...
+          return false
+        } else {
+          UE(
+            state,
+            outOfRangeFmtStr,
+            signedStr,
+            minWidth,
+            nBits
+          )
+          return false
```
stevedlawrence, line 116: `return false` is unnecessary, UE throws an exception and bubbles up to the top to end the unparse.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -78,17 +76,61 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
...
+    if (primNumeric.maxWidth.isDefined) {
+      val width = primNumeric.maxWidth.get
+      if (nBits > width) {
+        UE(
+          state,
+          "Number of bits %d out of range, must be between 1 and %d bits.",
+          nBits,
+          width
+        )
+        return false
```
stevedlawrence, line 129: Same here, `return false` can be removed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -534,7 +534,9 @@ object NodeInfo extends Enum {
     }
 
     trait PrimNumeric { self: Numeric.Kind =>
-      def width: MaybeInt
+      val isSigned: Boolean
```
stevedlawrence, line 537: Make this a def. Def's can be overridden with a val, but not the other way around, so def is a bit more flexible for implementations.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -686,8 +692,9 @@ object NodeInfo extends Enum {
           case _ => true
         }
       }
-
-      override val width: MaybeInt = MaybeInt.Nope
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(2)
```
stevedlawrence, line 696: Decimal is a bit of a weird case. It can be either signed or unsigned depending on the `dfdl:decimalSigned` property. And if that is false, then minWidth could be 1. Maybe we change this to a Nope to make it clear asking the minWidth of decimal doesn't really make sense, and you need to special case width checking fro decimals?

And actually, the spec says that the minimum width for xs:decimal is 8 bits, because lengthUnits="bits" is not allowed for xs:decimal. But we've removed that restriction for all the types so maybe this doesn't apply for us.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/dpath/NodeInfo.scala
```diff
@@ -767,7 +781,9 @@ object NodeInfo extends Enum {
       protected override def fromNumberNoCheck(n: Number): DataValueByte = n.byteValue
       override val min = JByte.MIN_VALUE.toLong
       override val max = JByte.MAX_VALUE.toLong
-      override val width: MaybeInt = MaybeInt(8)
+      override val isSigned: Boolean = true
+      override val minWidth: MaybeInt = MaybeInt(1)
```
stevedlawrence, line 785: Byte is signed so I think minWidth should be 2?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -70,11 +71,16 @@ abstract class PackedBinaryDecimalBaseParser(
...
   override val context: ElementRuntimeData,
-  signed: Boolean = false
 ) extends PrimParser
   with PackedBinaryConversion {
   override lazy val runtimeDependencies = Vector()
 
+  val signed = {
+    context.optPrimType.get match {
+      case n: NodeInfo.PrimType.PrimNumeric => n.isSigned
+      case _ => false
```
stevedlawrence, line 81: I'm surprised the default case has coverage. This doesn't seem possible to me. A packed binary number should always have a numeric type. I wonder if codecov just can't detect this correctly?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -70,11 +71,16 @@ abstract class PackedBinaryDecimalBaseParser(
...
+    }
+  }
   protected def getBitLength(s: ParseOrUnparseState): Int
 
   def parse(start: PState): Unit = {
```
stevedlawrence, line 86: Just below we have this line:
```scala
if (nBits == 0) return // zero length is used for outputValueCalc often.
```
The purpose of this bug was to make it so zero length numbers is an error. I don't understand why outputValueCalc would have zero length, but based on the bug it sounds like that shouldn't be allowed any more. Should that change to an unparse error?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
stevedlawrence: Also, this makes me concern the sign change wasn't correct? Previously we passed in `false` to a handful of these packed parsers, regardless of the type. I wonder if we don't allow unsigned primitives for those packed parsers, so using primType.isSigned is functionally the same?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -166,9 +163,36 @@ abstract class BinaryIntegerBaseParser(
 
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
-    if (primNumeric.width.isDefined) {
-      val width = primNumeric.width.get
+    if (primNumeric.minWidth.isDefined) {
+      val isSigned = primNumeric.isSigned
+      val signedStr = if (isSigned) "signed" else "unsigned"
```
stevedlawrence, line 168: Move isSigned/signedStr to the inner block where it's used.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberParsers.scala
```diff
@@ -166,9 +163,36 @@ abstract class BinaryIntegerBaseParser(
...
+          )
+        } else {
+          PE(
+            start,
+            outOfRangeFmtStr,
+            signedStr,
+            minWidth,
+            nBits
+          )
+          return
```
stevedlawrence, line 190: We should probably have test for the missing code coverage.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: Hmm, so I have a test that exercises it, and we go from an infoset containing -1 to unparsing 1, since it is only 1 bit. Is that ok?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: I think the reason I moved the check out of the unparse was we wouldn't be able to reach the code in BinaryIntegerBaseUnparser.puNumber in the case that nBits was 0

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: I'm not certain since the ticket only specified Binary Integers. @mbeckerle  thoughts?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: Hmm, currently for unsigned ints, the minWidth check is the nBits == 0 check, so I think moving it out to the unparse would increase duplication since the UE uses the same outOfRangeFmtStr that the SDW uses, reducing duplication.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
stevedlawrence: I think that's probably correct? If we think about the min/max values of a 2's complement value, the min value is given as -(2<sup>n-1</sup>) and the max value is (2<sup>n-1</sup>)-1, where n is the number of bits. That gives us a range of -1 to 0. So -1 unparsing to a single 1 bit seems reasonable. Hopefully our parse logic is consistent with that when this tunabled is enabled.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
stevedlawrence: I thought we never need to unparse 0 bits? Isn't unparsing 0 bits always illegal now?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
stevedlawrence: Gotcha, as long as we make sure all the putNumber calls are doing the correct range checking and throwing a UE when it's out of range that's fine. Feels like float, double, and decimal aren't currently erroring on zero, but maybe zero is impossible for float/double, so only decimal needs it?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
olabusayoT: So this is used by non-PrimNumeric types like date/time via ConvertZonedCombinator...so that's why this has coverage

stevedlawrence: Can you add a comment so that's clear in the future?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
mbeckerle: type xs:decimal with binaryNumberRep "binary" is a signed integer that is then scaled by binaryDecimalVirtualPoint. 

Before the scaling, I think it behaves exactly like xs:integer. For signed, it must have 2 bits. Unsigned (which is what property dfdl:decimalSigned controls - despite the DFDL spec not providing adequate description of the property) means you must have at least 1 bit.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
olabusayoT: So it looks like we parse data that is -1 to an infoset containing 0, which I think we should be unparsing to 0 instead of 1 here since the max value is 0?

mbeckerle: > So it looks like we parse data that is -1 to an infoset containing 0, which I think we should be unparsing to 0 instead of 1 here since the max value is 0?

What does "parse data that is -1" mean? How can the data aka bits be -1. 

Let's use this terminology: data means the representation, aka the "bit stream". 
The value, logical value, or infoset value is a number with no concept of representation. 

Unparse logical value -1 should give a single '1' bit in the data stream. Unparse 0 should give a single '0' bit in the data stream. Parsing is symmetric.

stevedlawrence: Hmm, seems like we have a bug in that parse and unparse are not symmetric.

For parse,
* 0x1 parses to 1
* 0x0 parses to 0

For unparse,
* 1 unparses to 0x0
* 0 unparse to 0x0
* -1 unparses to 0x1

Since parse will never create a -1 in the infoset, it means it will always unparse to 0 unless someone modifies the infoset, which is unlikely.

Considering this, I wonder if we should remove the new tunable and just drop support for signed 1-bit integers completely? If someone was already using this and roundtripping data, things were going to be subtly buggy.  So maybe changing this to an error is going to make it more obvious that things are broken and people can fix their schema?

And maybe we just add a deprecation/compatibility note that signed 1-bit integers were broken and shouldn't be supported anyways, so they are removed.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
olabusayoT: Above this we have PackedDecimal that also returns if nBits is 0, should we also make that a PE? Also PackedDecimal doesn't pass along decimalSigned to the base, is it logical to do something like the below? If so, they I think we need to update the PackedDecimal classes  to pass along decimalSigned

```xml
          <xs:element name="num" type="xs:decimal"  dfdl:representation="binary" dfdl:binaryNumberRep="ibm4690Packed" dfdl:decimalSigned="yes" dfdl:binaryDecimalVirtualPoint="0"/>
```

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
mbeckerle: Truncation of a number because it doesn't fit in the available space is supposed to be an unparse error.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -78,17 +76,58 @@ abstract class BinaryNumberBaseUnparser(override val context: ElementRuntimeData
...
+    if (primNumeric.minWidth.isDefined) {
+      val minWidth = primNumeric.minWidth.get
+      if (nBits < minWidth) {
+        val isSigned = primNumeric.isSigned
+        val signedStr = if (isSigned) "a signed" else "an unsigned"
+        val outOfRangeFmtStr =
+          "Minimum length for %s binary integer is %d bit(s), number of bits %d out of range. " +
+            "An unsigned integer with length 1 bit could be used instead."
+        if (isSigned && state.tunable.allowSignedIntegerLength1Bit && nBits == 1) {
+          state.SDW(
```
stevedlawrence, line 100: Are there any concerns about creating SDW's at parse time? Diagnostic overhead should be small, but imagine an array of a bunch of 1-bit length elements--that could create a lot of diagnostics. Also consider that this same warning was already output during compilation unless nbits was calculated at runtime. Maybe the compile time warning is sufficient and we should just be silent at runtime? Or maybe we need a flag so we only warn here only if we're in the `BinaryIntegerKnownLengthUnparser`?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
```diff
@@ -268,6 +306,32 @@ abstract class BinaryDecimalUnparserBase(
...
+          nBits
+        )
+      } else {
+        UE(
+          state,
+          outOfRangeFmtStr,
+          signedStr,
+          minWidth,
+          nBits
+        )
```
stevedlawrence, line 332: I think this same chunk of code with very minor variations appears in multiple places. Can we add trait to avoid all this duplication?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -51,7 +52,15 @@ abstract class PackedBinaryDecimalBaseParser(
...
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    if (nBits == 0) {
+      PE(
+        start,
+        "Number of bits %d out of range for a binary decimal. " +
+          "An unsigned decimal with length 1 bit could be used instead.",
```
stevedlawrence, line 59: I'm not sure this error message is correct. I believe packed numbers have different rules about minimum lengths. It's not the same as 2's complement.

Also, this message references "binary decimal", which I think implies 2's complement. Any error messages here should probably reference packed decimals, or leave it up to individual packed parsers to detect bitLength errors and give specific errors.

I think what was intended with the packed implementation is we just call `toBigDecimal`/`toBigInteger` below and the different packed implementations do their parsing/validation and throw an exception if it's not a valid length/value, and that exception is converted into a PE.

So I think all this 1-bit signed stuff done elsewhere doesn't even apply to packed decimals. Packed decimals should already be doing bit length checking specific to the packed format. If they handle the zero case correctly, we could probably even remove this check/PE. If not, we probably still need it, but should have an error specific to zero length packed decimals not allowed.

It might also be worth adding a test for some zero length packed things so we get coverage of this edge case.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -69,17 +78,35 @@ abstract class PackedBinaryDecimalBaseParser(
...
+      PE(
+        start,
+        "Minimum length for %s binary integer is %d bit(s), number of bits %d out of range. " +
+          "An unsigned integer with length 1 bit could be used instead.",
+        signedStr,
+        minWidth,
+        nBits
+      )
+      return
+    }
```
stevedlawrence, line 109: Same question as above, I'm not sure this error makes sense for packed numbers and can possibly be removed?

stevedlawrence (approved):
+1, nice improvements. I think my only comments is do reduce a bunch of copy/pasting, and I think we need different behavior/error messages for packed numbers (which we might already have?)

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/BinaryNumberUnparsers.scala
mbeckerle: I'm in favor of silent success at runtime if the flag is set to allow 1-bit signed (for compatibility), and parse-error if the flag is not set.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
mbeckerle: "binary decimal" means type xs:decimal dfdl:binaryNumberRep="binary" i.e., twos complement. 

A fraction part is specified using dfdl:binaryDecimalVirtualPoint="3" for example to get 999.999 (3 digits after the decimal point)

So "binary decimal" behaves like a twos-complement signed integer vis a vis this 1-bit compatibility mode. 

The packed representations are entirely different, and 1-bit stuff does not apply. They must be aligned 4-bit and a multiple of 4-bits length.

olabusayoT: It doesn't look like the indvidual packed parsers/unparser handle the nBits == 0 case, Before we were just returning when nBits was 0, should we revert the code, or just update the error to remove the suggestion of what could be used instead, and update the messaging to state nBits == 0 is not allowed.

mbeckerle: Well, we may be depending on this zero-length behavior. See the comment in the code about zero length is used by outputValueCalc. But this doesn't really make sense to me. Some things, like strings and hexBinary, can be zero length, but many data types zero length makes no sense.  There is defaulting, e.g., when we are trying to parse an integer and we get zero-length back we may provide a default value for the integer. But in the absence of defaulting, most simple types require at least 1 bit.

olabusayoT: I'm not sure about that comment. Isn't OVC only used on unparsing? Also, none of our OVC tests failed when we changed the Return to a PE

mbeckerle: Yes, OVC is only for unparsing. If no test fail when making this a PE, then I would also delete the comment about OVC, which makes no sense if this is parser code anyway.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -51,7 +52,14 @@ abstract class PackedBinaryDecimalBaseParser(
 
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    if (nBits == 0) {
+      PE(
+        start,
```
mbeckerle, line 57: This does need test coverage. It may be possible to remove this check entirely. It may not be reachable in that something else is checking for the zero case before this is called?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
mbeckerle: Yeah, I think you need to know if the decimal representation is expected to be signed or not. The error case of dfdl:decimalSigned="no" but the packed representation has a negative sign needs to be detected somehow. 
One could choose different parse primitives based on dfdl:decimalSigned yes/no or pass a boolean isSigned to a common primitive.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedDecimalParsers.scala
```diff
@@ -84,11 +84,10 @@ class PackedDecimalPrefixedLengthParser(
 
 class PackedIntegerRuntimeLengthParser(
   val e: ElementRuntimeData,
-  signed: Boolean,
   packedSignCodes: PackedSignCodes,
   val lengthEv: Evaluatable[JLong],
   val lengthUnits: LengthUnits
-) extends PackedBinaryIntegerBaseParser(e, signed)
+) extends PackedBinaryIntegerBaseParser(e)
```
mbeckerle, line 90: I think you need to pass the signed boolean.

mbeckerle (changes requested):
It may be possible to remove the check for nbits == 0 for packed decimal representations (including IBM4690Packed) because a check for 4-bit multiple has to be elsewhere and that presumably will also rule out the zero case.  At least one of these in the code does not have test coverage. 

I suggest removing the code that does check for nbits == 0 for the packed numbers and re-test. 

I also think removing all the passing of dfdl:decimalSigned as a boolean to the packed/ibm4690 primitives means you cannot detect a negative representation being parsed into an unsigned number.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
```diff
@@ -69,17 +77,30 @@ abstract class PackedBinaryDecimalBaseParser(
...
+      // context.optPrimType can be of type date/time via ConvertZonedCombinator
+      case _ => false
+    }
+  }
   protected def getBitLength(s: ParseOrUnparseState): Int
 
   def parse(start: PState): Unit = {
     val nBits = getBitLength(start)
-    if (nBits == 0) return // zero length is used for outputValueCalc often.
+    if (nBits == 0) {
```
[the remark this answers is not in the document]
olabusayoT: I can't find where we check the packed length is a multiple of 4 bits. Unless you're referring to what we do in the function below?

```scala
def bcdFromBigIntegerLength(absBigIntAsString: String, minLengthInBits: Int): (Int, Int) = {
    val numDigits = absBigIntAsString.length
    // Need to have an even number of digits to fill out a complete byte
    val requiredBitLen = if (numDigits % 2 == 0) (numDigits * 4) else ((numDigits + 1) * 4)
    val bitLen = scala.math.max(minLengthInBits, requiredBitLen)
    val numBytes = (bitLen / 8)
    val leadingZeros = bitLen / 4 - numDigits
    (numBytes, leadingZeros)
  }
```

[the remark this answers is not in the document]
olabusayoT: Op, nvm. I found it in org.apache.daffodil.core.grammar.ElementBaseGrammarMixin

```scala
 if ((binaryNumberKnownLengthInBits != -1) && (binaryNumberKnownLengthInBits % 4) != 0)
```

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedDecimalParsers.scala
stevedlawrence: Does using the signedness of the primitive type work? I think the only time xs:decimalSign should matter is with the primitive type is xs:decimal, which uses the `PackedDecimal*` parsers, not these `PackedInteger*` parsers. So if feels like we do need a signed boolean for `PackedDecimal*` parsers, but I think that's just an existing bug that could maybe be added in a separate PR, since this one is more about ranges of things.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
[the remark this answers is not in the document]
olabusayoT: This logic doesn't preclude nBits == 0, and it only checks for when nBits is known. I think we should leave in the 0 bits check as it is now, because it covers when the nBits is known and when it is calculated. Thoughts?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
olabusayoT: It is actually reachable. I've added the test. I don't think we ought to remove this check as we don't check for nBits == 0 anywhere else

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedBinaryTraits.scala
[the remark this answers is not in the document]
stevedlawrence: Yeah, you're right, we can't rely entirely compile time validation since lengths could be runtime calculate.

In fact, in addition to checking it's not zero length, we also need to check it's a multiple for 4 at runtime. Right now during runtime, we just ask for a byte array that is nbits long and the packed logic assumes the array is completely full, even though it could have partial fragment bytes.

So I think we do need to keep this 0 check, but also need to add a % 4 == 0 check.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/PackedDecimalParsers.scala
olabusayoT: So this signed boolean was replaced by the primitives signedness...as we were just passing in true/false depending on whether the primitive was signed or unsigned...the PAckedDecimalParsers never passed in decimalSigned, so that wasn't remove...as @stevedlawrence said I think that's a separate bug

olabusayoT: Created https://issues.apache.org/jira/browse/DAFFODIL-2963 to track separate bug

stevedlawrence: Also, https://github.com/apache/daffodil/pull/1391 modifies packed integers to use PrimType.fromNumber, which validates signedness according to the primitive, and the signed check in this parser goes away.

mbeckerle (approved):
+1

olabusayoT:
Note this PR causes regressions in dfdl-nitf, dfdl-vmf and vmf-pmmc-dfdl

PR #1338, opened 2024-10-15
Fix Bug with valueLength being overwritten after Trim

olabusayoT:
- currently after trimming the value of the element, we set the valueLength, and then overwrite it after returning from the parse that does the trimming. This results in the wrong value for value length. This fixes it by only setting it if it's a non-choice comlplexType and simpleTypes are handled elsewhere
- we also incorrectly use valuelength for prefixed length calculations when we ought to be using content length per the spec
- we also do not ensure valuelength isn't getting overwritten so we add asserts to setAbsStartPos0bInBits and setAbsEndPos0bInBits to verify that
- fix bug where padding is being added around prefixed length element (DAFFODIL-2943) by changing CaptureLengthRegion to wrap around contentlengthStart and padding
- fix bug where we were missing return after PE for Out of Range Binary Integers (DAFFODIL-2942)
- fix bug where we were using the main element's qname instead of the prefixed element qname in the Unparse Error message
- refactor Prefixed parsers to use state's bitLimit to get the prefix length (PrefixedLengthParserMixin2) since the specifiedLengthPrefixedParser will take care of parsing the prefix length
- refactored Prefixed unparsers to not try to unparse prefix length since that is taken care of by SpecifiedLengthPrefixedUnparser
- refactored prefixed parsers and unparsers to remove unused prefixed length parser related members
- add tests

DAFFODIL-2658

### DIFF
--- a/daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
+++ b/daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
@@ -59,6 +59,7 @@ import org.apache.daffodil.core.grammar.primitives.RightFill
 import org.apache.daffodil.core.grammar.primitives.ScalarOrderedSequenceChild
 import org.apache.daffodil.core.grammar.primitives.SpecifiedLengthExplicit
 import org.apache.daffodil.core.grammar.primitives.SpecifiedLengthImplicit
+import org.apache.daffodil.core.grammar.primitives.SpecifiedLengthPrefixed
 import org.apache.daffodil.lib.api.Diagnostic
 import org.apache.daffodil.lib.api.WarnID
 import org.apache.daffodil.lib.schema.annotation.props.gen.FailureType
@@ -301,6 +302,7 @@ object DaffodilCCodeGenerator
       case g: SeqComp => seqCompGenerateCode(g, cgState)
       case g: SpecifiedLengthExplicit => specifiedLengthExplicit(g, cgState)
       case g: SpecifiedLengthImplicit => specifiedLengthImplicit(g, cgState)
+      case g: SpecifiedLengthPrefixed => specifiedLengthPrefixed(g, cgState)
       case _ => gram.SDE("Code generation not supported for: %s", Misc.getNameFromClass(gram))
     }
   }
@@ -405,6 +407,13 @@ object DaffodilCCodeGenerator
 
   private def specifiedLengthImplicit(
     g: SpecifiedLengthImplicit,
+    cgState: CodeGeneratorState
+  ): Unit = {
+    DaffodilCCodeGenerator.generateCode(g.eGram, cgState)
+  }
+
+  private def specifiedLengthPrefixed(
+    g: SpecifiedLengthPrefixed,
     cgState: CodeGeneratorState
   ): Unit = {
     DaffodilCCodeGenerator.generateCode(g.eGram, cgState)
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
@@ -72,7 +72,9 @@ trait ElementBaseGrammarMixin
     }
   }
 
-  protected lazy val isDelimitedPrefixedPattern = {
+  lazy val isPrefixed: Boolean = lengthKind == LengthKind.Prefixed
+
+  protected lazy val isDelimitedPrefixedPattern: Boolean = {
     import LengthKind._
     lengthKind match {
       case Delimited =>
@@ -474,12 +476,16 @@ trait ElementBaseGrammarMixin
     lazy val leftPadding = leftPaddingArg
     lazy val rightPadFill = rightPadFillArg
     lazy val body = bodyArg
-    CaptureContentLengthStart(this) ~
-      leftPadding ~
-      CaptureValueLengthStart(this) ~
-      body ~
-      CaptureValueLengthEnd(this) ~
-      rightPadFill ~
+    specifiedLength(
+      CaptureContentLengthStart(this) ~
+        leftPadding ~
+        CaptureValueLengthStart(this) ~
+        body ~
+        CaptureValueLengthEnd(this) ~
+        rightPadFill
+    ) ~
+      // CaptureContentLengthEnd must be outside the specified length so it can
+      // do any skipping of bits it needs to before capturing the end of content length
       CaptureContentLengthEnd(this)
   }
 
@@ -623,14 +629,14 @@ trait ElementBaseGrammarMixin
 
   private lazy val stringPrim = {
     lengthKind match {
-      case LengthKind.Explicit => specifiedLength(StringOfSpecifiedLength(this))
-      case LengthKind.Prefixed => specifiedLength(StringOfSpecifiedLength(this))
+      case LengthKind.Explicit => StringOfSpecifiedLength(this)
+      case LengthKind.Prefixed => StringOfSpecifiedLength(this)
       case LengthKind.Delimited => stringDelimitedEndOfData
-      case LengthKind.Pattern => specifiedLength(StringOfSpecifiedLength(this))
+      case LengthKind.Pattern => StringOfSpecifiedLength(this)
       case LengthKind.Implicit => {
         val pt = this.simpleType.primType
         Assert.invariant(pt == PrimType.String)
-        specifiedLength(StringOfSpecifiedLength(this))
+        StringOfSpecifiedLength(this)
       }
       case LengthKind.EndOfParent if isComplexType =>
         notYetImplemented("lengthKind='endOfParent' for complex type")
@@ -645,7 +651,7 @@ trait ElementBaseGrammarMixin
   }
 
   private lazy val hexBinaryLengthPattern = prod("hexBinaryLengthPattern") {
-    new SpecifiedLengthPattern(this, new HexBinaryEndOfBitLimit(this))
+    new HexBinaryEndOfBitLimit(this)
   }
 
   private lazy val hexBinaryLengthPrefixed = prod("hexBinaryLengthPrefixed") {
@@ -1219,7 +1225,7 @@ trait ElementBaseGrammarMixin
   }
 
   private lazy val nilLitSimple = prod("nilLitSimple", isSimpleType) {
-    captureLengthRegions(leftPadding, specifiedLength(nilLitContent), rightPadding ~ rightFill)
+    captureLengthRegions(leftPadding, nilLitContent, rightPadding ~ rightFill)
   }
 
   private lazy val nilLitComplex = prod("nilLitComplex", isComplexType) {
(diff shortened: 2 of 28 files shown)


jadams-tresys (approved):
+1

daffodil-test/src/test/resources/org/apache/daffodil/section12/lengthKind/PrefixedTests.tdml
```diff
@@ -17,14 +17,14 @@
...
-  xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
-  defaultRoundTrip="onePass">
+                description="Section 12 - lengthKind=prefixed" xmlns:tdml="http://www.ibm.com/xmlns/dfdl/testData"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"
+                xmlns:xs="http://www.w3.org/2001/XMLSchema"
+                xmlns:fn="http://www.w3.org/2005/xpath-functions"
+                xmlns:ex="http://example.com"
+                xmlns:tns="http://example.com"
+                xmlns:dfdlx="http://www.ogf.org/dfdl/dfdl-1.0/extensions"
+                defaultRoundTrip="onePass" xmlns:x="http://www.ogf.org/dfdl/dfdl-1.0/">
```
stevedlawrence, line 27: Fix indentation.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/SpecifiedLengthParsers.scala
```diff
@@ -85,7 +85,9 @@ sealed abstract class SpecifiedLengthParserBase(eParser: Parser, erd: RuntimeDat
     if (pState.processorStatus ne Success) return
     val finalEndPos0b = startingBitPos0b + nBits
 
-    captureValueLength(pState, ULong(startingBitPos0b), ULong(dis.bitPos0b))
+    // if we haven't already set the value length, set it now
+    if (pState.infoset.valueLength.isEndUndef)
+      captureValueLength(pState, ULong(startingBitPos0b), ULong(dis.bitPos0b))
```
stevedlawrence, line 90: Is the issue here that some `eParsers` do not capture their own length and expect it to be captured here, but other parsers do capture value length (because they handle things like trimming), so sometimes there are two parsers trying to capture length for the same thing?

That makes me wonder if we shouldn't be capturing value length here at all, and instead we should be using the `CaptureValueLengthStart/End` parsers for those parsers that don't capture value length?

ElementBaseRuntime1Mixin has a `capturedByParsers` val that tries to list which parsers capture their own value length. Maybe that logic isn't completely correct, and the work around (that feels wrong) is we just always capture the length here?

olabusayoT: > Is the issue here that some `eParsers` do not capture their own length and expect it to be captured here, but other parsers do capture value length (because they handle things like trimming), so sometimes there are two parsers trying to capture length for the same thing?

That's right. For example, SpecifiedLengthExplicitParser doesn't capture its own length.

> That makes me wonder if we shouldn't be capturing value length here at all, and instead we should be using the CaptureValueLengthStart/End parsers for those parsers that don't capture value length?

I'll look into implementing this

olabusayoT: > ElementBaseRuntime1Mixin has a capturedByParsers val that tries to list which parsers capture their own value length. Maybe that logic isn't completely correct, and the work around (that feels wrong) is we just always capture the length here?

I think it's the "isReferenced" boolean that may be the issue, if we remove that, then we'll always captureValueLength if capturedByParsers is false

stevedlawrence: The `isReferenced` variable is used to avoid capturing value lengths that will never be used. For example, if we have something like this:

```xml
<element name="a" type="xs:int" ... />
```
If nothing ever references the value length of `a, then we shouldn't insert `CaptureValueLengthStart/End` parsers, regardless of the value of `capturedByParsers`. But if something does reference the length, e.g.
```xml
<element name="len" type="xs:int" dfdl:inputValueCalc="{ dfdl:valueLength(../a, 'bytes')) } />
```
The we do need to capture the value length of `a`. So `isReferenced` is needed to disabled capturing when it's not needed. I don't think we can get rid of it.

Usually we capture the length via the `CaptureValueLengthStart/End` parsers, but sometimes those parsers aren't smart enough so it must be done by a separate parser that has enough information to capture length correctly (e.g. StringOfSpecifiedLength that knows about pad/trim stuff and how to exclude it from the value length). The `capturedByParsers` val is supposed to figure out which are those special cases so we don't unnecessarily insert `CaptureValueLengthStart/End` parsers for parsers that do it themselves.

It seems like SpecifiedLengthParserBase doesn't have enough information to know how to correctly capture value length (since sometimes the subparser is the thing that knows how to do it). So that probably means it shouldn't be capturing the length at all. But that then means other parsers that relied on that behavior either need to call `captureValueLength` themselves, or `capturedByParsers` needs to be false in some cases so that we insert the `CaptureValueLengthStart/End` parsers.

mbeckerle: All the complexity around whether to insert a value-length capture parser at the start/end is about symmetry of the parser with the way the unparser works. 

Capturing length when parsing is easy. Record start position when parser is entered, ending position when it is exited.  Probably so light weight that it's not worth trying to optimize it away.

Unparsing... not so much. The unparser has to unwind and allow unparsing to continue while the value length is suspended. Definitely we want none of this overhead for elements which no expression references their length. We only want capture of length overheads when it is essential. I am also hoping that some of this is simpler for prefixed lengths, which don't work for unparsing with suspensions, etc. currently.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -24,9 +24,9 @@ import org.apache.daffodil.core.dsom.PrimitiveType
 import org.apache.daffodil.core.dsom.Root
 import org.apache.daffodil.core.dsom.SimpleTypeDefBase
```
stevedlawrence, line 25: I wonder if we should modify `setAbsStartBitPos0bInBits` so it does `Assert.invariant(maybeStartPos0bInBits.isEmpty)` and something similar for `setAbsEndBitPos0bInBits`? I think those assertion would have caught this bug since if it fails it means we have set the absolute bit position twice, which we should never do.

daffodil-core/src/main/scala/org/apache/daffodil/core/runtime1/ElementBaseRuntime1Mixin.scala
```diff
@@ -89,17 +89,19 @@ trait ElementBaseRuntime1Mixin { self: ElementBase =>
...
     // implicit or delimited.
     //
     // So for these cases we do not want to capture value length with the
     // Capture{Start,End}OfValueLengthParsers, since those lengths are captured
     // by the value parsers
-    val capturedByParsers =
-      (isSimpleType && (impliedRepresentation == Representation.Text || lengthKind == LengthKind.Delimited)) ||
+    val capturedByValueParsers =
+      (isSimpleType && (
+        primType == PrimType.String || lengthKind == LengthKind.Delimited)) ||
```
stevedlawrence, line 101: I'm wondering if this change is correct?

For example, say we have this element:
```xml
<xs:element name="foo" type="xs:int" dfdl:representation="text" dfdl:trimKind="padChar" dfdl:lengthKind="explicit" dfdl:length="10" ... />
```
So a fixed length text integer with padding. In this case I think what Daffodil does is it create a String parser to parse the fixed length string and remove padding, and then creates another parser to convert that string to an actual integer.

So in that case, even though the primType is not String, I think the String parser will still be used to capture the value length after padding is removed. So I *think* `impliedRepresentation == Text` is still needed?

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/primitives/PrimitivesBCD.scala
```diff
@@ -52,20 +52,10 @@ class BCDIntegerKnownLength(val e: ElementBase, lengthInBits: Long) extends Term
...
 class BCDIntegerPrefixedLength(val e: ElementBase) extends Terminal(e, true) {
 
-  override lazy val parser = new BCDIntegerPrefixedLengthParser(
-    e.elementRuntimeData,
-    e.prefixedLengthBody.parser,
-    e.prefixedLengthElementDecl.elementRuntimeData,
-    e.lengthUnits,
-    e.prefixedLengthAdjustmentInUnits
-  )
+  override lazy val parser = new BCDIntegerPrefixedLengthParser(e.elementRuntimeData)
```
stevedlawrence, line 55: I wonder if we want to rename these something like `BCDIntegerBitLimitLengthParser` and `BCDIntegerMinimumLengthUnparser`, making it clear that they aren't really doing anything specifically with prefixed length, and better describes the behavior of how they actuall parse/unparse things?

And when we implement things like lengthKind endOfParent, we could probably just use the same BitLimitParser, for example.

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -165,3 +165,21 @@ trait PrefixedLengthParserMixin {
     }
   }
 }
+
+/**
+ * This mixin doesn't require parsing the prefix length element and just uses
+ * the state's bitLimit and position to get the bitLength instead
+ */
+trait PrefixedLengthParserMixin2 {
```
stevedlawrence, line 173: Need a different name for this. Numbers are not descriptive enough. Maybe we renames these something like `PrefixLengthFromParserMixin` and `PrefixLengthFromBitLimitMixin`? Something to make it more clear how they differ without having to read the code.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -72,13 +72,21 @@ trait ElementBaseGrammarMixin
...
   }
 
-  protected lazy val isDelimitedPrefixedPattern = {
+  protected lazy val isPrefixed: Boolean = {
+    import LengthKind._
+    lengthKind match {
+      case Prefixed => true
+      case _ => false
+    }
+  }
```
stevedlawrence, line 81: This can just be `lengthKind eq LengthKind.Prefixed`. Don't really need a match/case if we are just going to return true/false.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -72,13 +72,21 @@ trait ElementBaseGrammarMixin
...
+
+  protected lazy val isDelimitedPrefixedPattern: Boolean = {
     import LengthKind._
     lengthKind match {
       case Delimited =>
         true // don't test for hasDelimiters because it might not be our delimiter, but a surrounding group's separator, or it's terminator, etc.
       case Pattern => true
-      case Prefixed => true
+      case Prefixed => isPrefixed
```
stevedlawrence, line 89: I would suggest just returning true here, isPrefixed doesn't do anything except check lengthKind == Prefixed, which is exactly what this match case does.

stevedlawrence (commented):
Looks reasonable and the right approach for correctly implement prefixed length. Using things like valueLength was clearly wrong. Just a few questions.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1341,7 +1356,13 @@ trait ElementBaseGrammarMixin
       case LengthKind.Delimited => body
       case LengthKind.Pattern => new SpecifiedLengthPattern(this, body)
       case LengthKind.Explicit if bitsMultiplier != 0 =>
-        new SpecifiedLengthExplicit(this, body, bitsMultiplier)
+        if (isSimpleType && primType == PrimType.HexBinary) {
+          // hexBinary has some checks that need to be done that SpecifiedLengthExplicit
+          // gets in the way of
```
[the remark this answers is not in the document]
olabusayoT: @mbeckerle , any thoughts on refactoring the code to have various specified length parser as described above, and any idea on how that would work with unparsing?

daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
```diff
@@ -289,7 +289,8 @@ object DaffodilCCodeGenerator
       case g: ElementParseAndUnspecifiedLength =>
         elementParseAndUnspecifiedLengthGenerateCode(g, cgState)
       case g: ElementUnused => noop(g)
-      case g: HexBinaryLengthPrefixed => hexBinaryLengthPrefixedGenerateCode(g.e, cgState)
+      case g: HexBinaryEndOfBitLimit if g.e.isPrefixed =>
```
stevedlawrence, line 292: Hmm, maybe my suggestion was wrong and the primitive still wants to be something like HexBInaryLengthPrefixed so that the grammar is obvious and things like code generators/etc can use the obvious grammar names? The parsers generated don't necessarily have to match the names.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1387,14 +1381,15 @@ trait ElementBaseGrammarMixin
...
-          if isSimpleType && impliedRepresentation == Representation.Binary =>
+          if isSimpleType &&
+            impliedRepresentation == Representation.Binary &&
+            primType != PrimType.HexBinary =>
         new SpecifiedLengthImplicit(this, body, implicitBinaryLengthInBits)
-      case LengthKind.Implicit if isComplexType =>
+      case LengthKind.Implicit =>
         body // for complex types, implicit means "roll up from the bottom"
+      // for simple types, the primitives have custom parsers that handle implicit length logic
+      // and don't use the limit provided by the SpecifiedLengthImplicit parser
```
stevedlawrence, line 1392: Can you move this comment above `body`? Single line comments can be on the same line, but multi-line comments should start before the line.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1387,14 +1381,15 @@ trait ElementBaseGrammarMixin
...
         new SpecifiedLengthImplicitCharacters(this, body, this.maxLength.longValue)
-
-      case LengthKind.Implicit if isSimpleType && primType == PrimType.HexBinary =>
-        new SpecifiedLengthImplicit(this, body, this.maxLength.longValue * bitsMultiplier)
       case LengthKind.Implicit
-          if isSimpleType && impliedRepresentation == Representation.Binary =>
+          if isSimpleType &&
+            impliedRepresentation == Representation.Binary &&
+            primType != PrimType.HexBinary =>
         new SpecifiedLengthImplicit(this, body, implicitBinaryLengthInBits)
```
stevedlawrence, line 1388: This condition feels like it doesn't exclude enough things. For example, this matches all binary types except hex binary. But, for example, integers with length kind implicit use something like `BinaryIntegerKnownLength` primitive/parser, which doesn't rely on SpecifiedLengthImplicit parser. It doesn't necessarily hurt, but it's just unnecessary work and might slow things down, since the BinaryIntegerKnownLengthParser is going to do that work.

I'm wonder if it's just string types that really make use of specified length stuff?

stevedlawrence (commented):
It's still not clear to me that we are correctly adding/not adding SpecifiedLength* parsers. Do we have any documentation or any easy way to tell that we are doing things correctly, and only adding those parsers when they are actually needed?

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
olabusayoT: So It look like nillable binary types need SpecifiedLengthImplicit., they are the only tests failing when I comment out that chunk of code

daffodil-codegen-c/src/main/scala/org/apache/daffodil/codegen/c/DaffodilCCodeGenerator.scala
stevedlawrence: Sorry that wasn't clear. I'm suggesting that we should keep the old grammar name `HexBinaryLengthPrefixed` so this change isn't needed. This way the grammar is a more clear for things like the code generator that examine it. But it's still fine to call the parsers `BitLimitLength/MinimumLength` etc from that grammar.

mbeckerle (approved):
+1 Code looks good. Questions on how to organize this code better long term remain, but that requires more than a PR review.

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1341,7 +1356,13 @@ trait ElementBaseGrammarMixin
...
+        if (isSimpleType && primType == PrimType.HexBinary) {
+          // hexBinary has some checks that need to be done that SpecifiedLengthExplicit
+          // gets in the way of
+          body
+        } else {
+          new SpecifiedLengthExplicit(this, body, bitsMultiplier)
+        }
       case LengthKind.Explicit => {
         Assert.invariant(!knownEncodingIsFixedWidth)
         Assert.invariant(lengthUnits eq LengthUnits.Characters)
```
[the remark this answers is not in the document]
mbeckerle: Suggest create a cleanup issue ticket for this. It's a performance (and maintainability) improvement but not specifically about this PR's primary goal is it?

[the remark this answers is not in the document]
stevedlawrence: I'm not sure this is a cleanup of existing issues. I think this change could possibly lead to adding duplicate parsers and might slow things down. So it's not that we already have duplicates, but this change could lead to new duplicates.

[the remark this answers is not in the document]
stevedlawrence: Also, I believe this bug just fixes issues with prefixed lengths, which I think has a number of other outstanding issues. So even if we fix this one, I'm not sure prefixedLength is still going to be safe to use. Feels like it's less criticial to get into this release, and due to possible regressions is worth pushing to the next release.

olabusayoT:
Created DAFFODIL-2967 to ensure we're properly adding/not adding the schemas

@mbeckerle , the piece of code you highlighted as being better off removed, removing that bit caused test failures iirc!

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
```diff
@@ -200,7 +124,7 @@ trait CalculatedPrefixedLengthUnparserMixin {
         UnparseError(
           One(state.schemaFileLocation),
           One(state.currentLocation),
-          s"The calculated value of ${elem.namedQName} ($adjustedLenInUnits) failed check due to ${check.errMsg}"
+          s"The calculated value of ${plElem.namedQName} ($adjustedLenInUnits) failed check due to ${check.errMsg}"
```
stevedlawrence, line 127: Is switching to `plElem` correct? Although the check is done against `plElem`, `plElem` is really an ephemeral quasi-element that doesn't exist in the real infoset, so it might be confusing to mention it in an error message. Maybe this instead wants to be something like "The calculated prefix length of ${elem.namedQName} ..."?

daffodil-runtime1/src/main/scala/org/apache/daffodil/runtime1/processors/parsers/BinaryNumberTraits.scala
```diff
@@ -165,3 +165,21 @@ trait PrefixedLengthParserMixin {
     }
   }
 }
+
+/**
+ * This mixin doesn't require parsing the prefix length element and just uses
+ * the state's bitLimit and position to get the bitLength instead
```
stevedlawrence, line 171: This comment is a little confusing out of context, since it isn't really clear why prefix length elements are mentioned since this doesn't really seem to be prefixed length related. This could also in theory be used on many of the parsers that rely on bit limit for length, which isn't specific to prefixed length. Maybe instead say something like
> Some parsers do not calculate their own length, but instead expect another parser to set the bit limit and then they use that bit limit as the length. An example of this is prefix length parsers. This trait can be used by those parsers to do determine the length based on the bitLimit and position.

stevedlawrence (commented):
+1, just some minor comments about questions and a little cleanup

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1322,60 +1328,75 @@ trait ElementBaseGrammarMixin
...
+          //
+          val nBits =
+            encodingInfo.knownFixedWidthEncodingInCharsToBits(this.maxLength.longValue)
+          new SpecifiedLengthImplicit(this, body, nBits)
+        }
+        case LengthKind.Implicit if isSimpleType && primType == PrimType.String =>
+          new SpecifiedLengthImplicitCharacters(this, body, this.maxLength.longValue)
+
+        case LengthKind.Implicit if isSimpleType && primType == PrimType.HexBinary =>
+          new SpecifiedLengthImplicit(this, body, this.maxLength.longValue * bitsMultiplier)
```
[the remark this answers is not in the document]
olabusayoT: hmm...im surprised it has no coverage, several tests pass that condition of isSimpleType and impliedRep == Text and isHexBinary and LK implicit!

ex: maxHexBinaryError, hexBinary_Implicit_03b etc

These tests fail when we remove that case

daffodil-core/src/main/scala/org/apache/daffodil/core/grammar/ElementBaseGrammarMixin.scala
```diff
@@ -1322,60 +1328,75 @@ trait ElementBaseGrammarMixin
...
+        case LengthKind.Implicit
+            if isSimpleType && impliedRepresentation == Representation.Binary =>
+          new SpecifiedLengthImplicit(this, body, implicitBinaryLengthInBits)
+        case LengthKind.Implicit if isComplexType =>
+          body // for complex types, implicit means "roll up from the bottom"
+        case LengthKind.EndOfParent if isComplexType =>
+          notYetImplemented("lengthKind='endOfParent' for complex type")
+        case LengthKind.EndOfParent =>
+          notYetImplemented("lengthKind='endOfParent' for simple type")
+        case _ => {
```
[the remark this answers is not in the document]
olabusayoT: Maybe there's a bug with code coverage, hexBinary_bits_be_msbf definitely hits the default case

mbeckerle:
ChatGPT thinks that yes there are issues with CodeCov and Scala code:

https://chatgpt.com/share/67a28042-0944-800f-b2e5-95c1856c91d6

PR #1339, opened 2024-10-15
Verify OVC reports data too long error as necessary

olabusayoT:
- currently we have no tests that try to OVC into a too small element. This PR adds in such a test.

(The ticket is marked verify, this test verifies that the fix is already in)

DAFFODIL-1595


jadams-tresys (approved):
+1

stevedlawrence (approved):
+1

PR #1340, opened 2024-10-15
Add SBT task to generate tunables doc

olabusayoT:
- currently we hardcode the description of tunables on daffodil-site. This list is out of date. Instead, the fix uses an sbt task 'genTunablesDoc' to generate a markdown document from the dafext.xsd and outputs it to target/tunables.md
- related daffodil-site PR https://github.com/apache/daffodil-site/pull/148

DAFFODIL-2758

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,7 @@ lazy val genProps = taskKey[Seq[File]]("Generate properties scala source")
 lazy val genSchemas = taskKey[Seq[File]]("Generate DFDL schemas")
 lazy val genCExamples = taskKey[Seq[File]]("Generate C example files")
 lazy val genVersion = taskKey[Seq[File]]("Generate VERSION file")
+lazy val genTunablesDoc = taskKey[Seq[File]]("Generate tunables doc from dafext.xsd file")
 
 lazy val daffodil = project
   .in(file("."))
@@ -52,7 +53,14 @@ lazy val daffodil = project
     tutorials,
     udf
   )
-  .settings(commonSettings, nopublish, ratSettings, unidocSettings, genCExamplesSettings)
+  .settings(
+    commonSettings,
+    nopublish,
+    ratSettings,
+    unidocSettings,
+    genTunablesDocSettings,
+    genCExamplesSettings
+  )
 
 lazy val macroLib = Project("daffodil-macro-lib", file("daffodil-macro-lib"))
   .settings(commonSettings, nopublish)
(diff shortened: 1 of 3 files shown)


jadams-tresys (approved):
+1

Generated markdown looks good

build.sbt
```diff
@@ -481,6 +489,99 @@ lazy val unidocSettings =
...
+    // build documentation
+    val sectionsSeq = tunablesSeq.flatMap { ele =>
+      val subtitle = ele \@ "name"
+      val documentation = (ele \ "annotation" \ "documentation").text
+      val sections =
+        if (documentation.nonEmpty && !documentation.trim.startsWith("Deprecated")) {
+          val s =
+            s"""
+            |#### $subtitle
+            |$documentation
```
stevedlawrence, line 510: I noticed in the daffodil-site PR that $documentation includes indentation from the dafext.xsd, which makes it show up as code. We probably want to strip that off so it shows up as normal text.

build.sbt
```diff
@@ -481,6 +489,99 @@ lazy val unidocSettings =
...
+          Some(s)
+        } else {
+          None
+        }
+      sections
+    }
+    val documentationPage =
+      s"""|---
+        |layout: page
+        |title: DFDL Tunables
```
stevedlawrence, line 521: Suggest either `Daffodil Tunables` since these aren't part of the DFDL spec, or just `Tunables`, since maybe these being for Daffodil is implied.

build.sbt
```diff
@@ -481,6 +489,106 @@ lazy val unidocSettings =
...
+      val subtitle = ele \@ "name"
+      val documentation = (ele \ "annotation" \ "documentation").text.trim.replaceAll(" +", " ")
+      val default = ele \@ "default"
+      (subtitle, documentation, default)
+    }
+    val (deprecated, nonDeprecated) = documentationTuple.partition { t =>
+      t._2.startsWith("Deprecated")
+    }
+    if (nonDeprecated.isEmpty) {
+      sys.error("tunables document generation failed as non-deprecated elements list is empty")
```
stevedlawrence, line 514: This just outputs a log message but doesn't actually fail the task. I've found the best way to create a failure with an error message is to replace `sys.error` with this:
```scala
throw new MessageOnlyException("error message here")
```
It also means you can move the following code out of the else block.

build.sbt
```diff
@@ -481,6 +489,106 @@ lazy val unidocSettings =
...
+    val outputDocFile = (Compile / target).value / "tunables.md"
+    // parse xsd file
+    val dafExtXml = scala.xml.XML.loadFile(dafExtFile)
+    // extract tunables information
+    val tunablesElements =
+      (dafExtXml \ "element").filter(_ \@ "name" == "tunables") \\ "all" \ "element"
+    // build documentation
+    val documentationTuple = tunablesElements.map { ele =>
+      val subtitle = ele \@ "name"
+      val documentation = (ele \ "annotation" \ "documentation").text.trim.replaceAll(" +", " ")
```
stevedlawrence, line 506: Suggest doing `(ele \ "annotation" \ "documentation").text.split("\n").map(_.trim).mkString("\n")`. Triming each individual line will avoid a space at the beginning of lines. I *think* the extra space shouldn't affect the markdown since it's just a single space, but for a format that is somewhat space-sensitive it's probably best to ensure there are no leading spaces.

stevedlawrence (approved):
+1, :+1:  just minor suggestions

PR #1341, opened 2024-10-15
Remove unnecessary truncation code

olabusayoT:
- we were truncating and overwriting value, but we didn't actually need to since we ended up truncating again right after. Removed the unnecessary and buggy code which changed the dataValue. StringMaybeTruncateBitsUnparser.unparse does not change the dataValue, instead it changes the DOS, resulting in the value getting truncated but string-length returning the untruncated length when it's called
- add tests

DAFFODIL-1592

### DIFF
--- a/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
+++ b/daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
@@ -20,42 +20,22 @@ package org.apache.daffodil.unparsers.runtime1
 import org.apache.daffodil.lib.exceptions.Assert
 import org.apache.daffodil.lib.schema.annotation.props.gen.LengthUnits
 import org.apache.daffodil.lib.schema.annotation.props.gen.Representation
-import org.apache.daffodil.lib.util.Maybe
 import org.apache.daffodil.lib.util.Maybe._
-import org.apache.daffodil.lib.util.MaybeJULong
-import org.apache.daffodil.runtime1.dpath.NodeInfo.PrimType
 import org.apache.daffodil.runtime1.infoset.DIElement
 import org.apache.daffodil.runtime1.infoset.DISimple
 import org.apache.daffodil.runtime1.infoset.Infoset
-import org.apache.daffodil.runtime1.infoset.RetryableException
 import org.apache.daffodil.runtime1.processors.CharsetEv
 import org.apache.daffodil.runtime1.processors.ElementRuntimeData
-import org.apache.daffodil.runtime1.processors.Evaluatable
 import org.apache.daffodil.runtime1.processors.ParseOrUnparseState
-import org.apache.daffodil.runtime1.processors.Success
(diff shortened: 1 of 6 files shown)


daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
```diff
@@ -87,122 +86,31 @@ final class SpecifiedLengthExplicitImplicitUnparser(
...
-              unparseVarWidthCharactersInBits(state)
-            case LengthUnits.Characters =>
-              unparseVarWidthCharactersInCharacters(state)
+              eUnparser.unparse1(state)
+            case LengthUnits.Characters => {
+              //
+              // variable-width encodings and lengthUnits characters, and lengthKind explicit
+              // is not supported (currently) for complex types
+              //
+              state.schemaDefinitionUnless(
```
jadams-tresys, line 95: Just trying to remember, is it still an SDE if it is something that just isn't supported yet?

jadams-tresys (approved):
+1 assuming that SDE is the correct type of error for this.

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
```diff
@@ -87,122 +86,31 @@ final class SpecifiedLengthExplicitImplicitUnparser(
...
+                getCharset(state).name,
+                lengthKind.toString,
+                lengthUnits.toString
+              )
+
+              Assert.invariant(erd.isSimpleType)
+
+              // truncation is done by StringMaybeTruncateCharactersUnparser
+              eUnparser.unparse1(state, erd)
+            }
```
stevedlawrence, line 107: It seems `SpecifiedLengthExplicitImplicitUnparser` doesn't really do anything anymore. I wonder if it can just be greatly simplified? The binary and fixed width text cases call unparseBits which just call `eUnparser.unpase1(...)` (it evaluates target length but I'm not sure if that's needed anymore?). And now the non-fixed width text case also calls `eUnparser.unparse1(...)`, just with an SDE check for complex types.

Feels like all of the real work is now moved out of this unparser and into others. Can this just be simplified to
```scala
if (text representation AND lengthUnits characters AND non fixed width AND isComplex) {
  SDE(...)
} else {
  // the child unparser inspects target length and handles truncation and padding/fill
  // nothing needs to be done in this unparser
  eUnparser.unparse1(...)
}
```
And literally everything else about targetLength and unparseBits goes away?

daffodil-runtime1-unparser/src/main/scala/org/apache/daffodil/unparsers/runtime1/SpecifiedLengthUnparsers.scala
stevedlawrence: Good point, I think we usually use `subset`/`subsetError`, which is an SDE but mentions it's a "subset".

stevedlawrence (approved):
+1, possibility for removing a lot of code?

PR #1343, opened 2024-10-17
Fix Missing File Exception

olabusayoT:
- currently we don't capture NoSuchFileException from Paths.get in our run try/catch, which results in a "This is a bug" exception. With this fix, we capture the exception and add a bit more detail to make it clear what is happening (missing file or directory), since the exception only reports the file name.
- add cli test

DAFFODIL-2944


mbeckerle (approved):
+1

PR #1344, opened 2024-10-17
fix macos install instructions

atriviality:
Updating build instructions for MacOS; libmxml bumped to 4.0 and homebrew doesn't have 3.3.1 in the recipe.

Short term fix for https://issues.apache.org/jira/browse/DAFFODIL-2889.


stevedlawrence (approved):
+1 :+1: 

Can you please squash this into a single commit?

olabusayoT (approved):
+1

PR #1354, opened 2024-10-28
Update Sonarcloud Properties

olabusayoT:
- currently we check Main.scala for duplicate code, which is a problem due to DAFFODIL-1958. So we update the path to Main.scala so we can continue to exclude it.
- we no longer have src/it in our codebase, so we remove that from tests
- Added in 3 newest modules since their absence seems to be an oversight

DAFFODIL-2949

### DIFF
--- a/.sonar-project.properties
+++ b/.sonar-project.properties
@@ -16,9 +16,9 @@
 sonar.organization=apache
 sonar.projectKey=apache-daffodil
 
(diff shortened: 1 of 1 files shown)


stevedlawrence (approved):
+1

jadams-tresys (approved):
+1

PR #1356, opened 2024-10-28
Add src/main to daffodil-test-integration for sonarcloud

olabusayoT:
- without it, sonarcloud errors out. So we add a .keep file in the directory.

(I've verified locally that sonar cloud executes successfully after this change)

DAFFODIL-2949


stevedlawrence (approved):
+1

jadams-tresys (approved):
+1

PR #1357, opened 2024-11-01
Improve diagnostic in TDML Runner

mbeckerle:
I decided to fix an ancient bug. 

Note that this change enforces the syntax of TDML better. 
The dfdlInfoset element used to tolerate having spurious text in it, so long as it also had a root element.
Two TDML files had to be fixed in Daffodil.

If TDML files in schema projects have these spurious characters that were formerly ignored, this change will break
those tests.

DAFFODIL-533

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -31,6 +31,7 @@ import java.nio.charset.StandardCharsets
 import scala.collection.mutable
 import scala.language.postfixOps
 import scala.util.Try
+import scala.xml.Elem
 import scala.xml.Node
 import scala.xml.NodeSeq
 import scala.xml.NodeSeq.seqToNodeSeq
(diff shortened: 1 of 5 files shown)


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2818,17 +2830,9 @@ case class DFDLInfoset(di: Node, parent: Infoset) {
       val newExceptions = (testSuite.loadingExceptions -- before).toSeq
       testCase.toss(TDMLException(newExceptions, None), None)
     }
-    elem
+    elem.asInstanceOf[scala.xml.Elem]
```
mbeckerle, line 2833: Can just be 'Elem'

olabusayoT (approved):
+1 Looks good, but needs a test to exercise FileNotFoundException

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2769,41 +2770,52 @@ case class Infoset(i: NodeSeq, parent: TestCase) {
...
+    val (elems, others) = di.child.partition(_.isInstanceOf[scala.xml.Elem])
+    (elems, others.text.trim) match {
+      case (Seq(elem: Elem), "") if (ty == "infoset") => Left(elem)
+      case (Seq(), str) if (ty == "file") => Right(str)
+      case _ =>
+        Assert.usageError(
+          """dfdlInfoset element must contain a single root element or a file path (when 'type="file"')."""
+        )
+    }
+  }
```
stevedlawrence, line 2790: Semi related, I wonder if we made a mistake by adding a `type` attribute to `dfdlInfoset`? Maybe instead `tdml:infoset` should be a choice of different children, so for example a file type would be specified like this instead:

```xml
<tdml:infoset>
  <tdml:file>/path/to/file.xml</tdml:file>
</tdml:infoset>
```

It's maybe too late now since lots of things use `type="file"` but maybe in the future the use of `type="foo"` to define the format of the children should be avoided? It probably would have simplified this logic a bit.

stevedlawrence (approved):
+1

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
mbeckerle: Agreed. Things in XSD should be designed so that XSD can validate them rather than more complex rules being required in code. It's a shame that XSD 1.1 has not caught on, because it can validate such co-constraints. 

The right thing would have been a different element, e.g, dfdlInfosetFile which could only have a file path as its value.

PR #1359, opened 2024-11-02
Fix always lenient calendar parsing/unparsing

olabusayoT:
- we used to override the leniency regardless of what the calendar check policy was, now this ticket fixes that by setting it based on the isLenient value of calendar.
- uncomment previously failing tests
- make newly failing tests lax
- add additional test

DAFFODIL-2951, DAFFODIL-1042


daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
```diff
@@ -6283,7 +6285,19 @@
 
     <tdml:document><![CDATA[2001-03-35]]></tdml:document>
     <tdml:errors>
-      <tdml:error/>
+      <tdml:error>Unable to parse</tdml:error>
+      <tdml:error>2001-03-35</tdml:error>
+      <tdml:error>valid range</tdml:error>
```
stevedlawrence, line 6290: It looks like this tests already failed, it's just updated to include an error message. Did `calendarCheckPolicy="strict" already correctly fail to parse this in certain cases, or was the existing error something else? We had a report that this exact kind of test would successfully parse even with a strict policy.

stevedlawrence (commented):
Change seems reasonable, but it looks like the existing tests already failed, makes me wonder if this does already work as expected? What does change formatter.setLenient actually change?

daffodil-test/src/test/resources/org/apache/daffodil/section05/simple_types/SimpleTypes.tdml
olabusayoT: This test was commented out with DAFFODIL-1042. It used to work with Saxon, but failed with the new implementation

jadams-tresys (approved):
+1

PR #1360, opened 2024-11-02
Provide diagnosticDebugName for GroupRefFactory

olabusayoT:
- currently we don't have one resulting in the class name showing up in diagnostics
- setting the diagnosticDebugNameImpl fixes the issue

DAFFODIL-2948


stevedlawrence (approved):
+1

mbeckerle (approved):
+1

PR #1361, opened 2024-11-04
Use macOS-14 for Github Actions

olabusayoT:
- which is the latest stable version. They only support the last two stable versions, so we should be good for a while
- add step to install sbt since later version of the runner image don't include it
- had to mxml build from source since macos-14 is Apple Silicon mac and by default installs to /opt/homebrew/include, which isn't on the cc search path. Attempts to add the path other ways (-I or CPATH) proved unsuccessful

DAFFODIL-2952


stevedlawrence (approved):
+1

jadams-tresys (approved):
+1

PR #1364, opened 2024-11-05
Only provide valid line/col info from Xerces

olabusayoT:
- currently we output the line/col info directly from Xerces even if it's -1. We update the code to provide none if line/col <= 0 with test
- we also fix the doublespace when there is no diagnosticDebugName provided for schema context
- we looked into implementing a locator, and it was more trouble than it was worth since Xerces doesn't provide it to us

DAFFODIL-2953

### DIFF
--- a/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLISaveParser.scala
+++ b/daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLISaveParser.scala
@@ -327,7 +327,7 @@ class TestCLISaveParser {
     withTempFile { parser =>
       runCLI(args"save-parser -s $schema $parser") { cli =>
         cli.expectErr("[error]")
-        cli.expectErr(s"Schema context:  Location line 32 column 74 in ${schema.normalize()}")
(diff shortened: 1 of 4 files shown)


daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
```diff
@@ -969,4 +969,15 @@ class TestCLIParsing {
...
+
+  @Test def test_includeDoesNotExist_Error(): Unit = {
+    val schema = path(
+      "daffodil-cli/src/test/resources/org/apache/daffodil/cli/cli_schema_invalid_include.dfdl.xsd"
+    )
+    runCLI(args"parse -s $schema") { cli =>
+      cli.sendLine("123", inputDone = true)
+      cli.expectErr("[error]")
+      cli.expectErr(s"Schema context: Location in ${schema.normalize()}")
+    }(ExitCode.UnableToCreateProcessor)
```
stevedlawrence, line 981: I imagine we already have a test that tests for failing to resolve schema locations. Maybe we can just update the expected diagnostics to be a bit more specific so we can avoid adding another test?

stevedlawrence (approved):
+1

jadams-tresys (approved):
+1

PR #1365, opened 2024-11-06
Add use of DAFFODIL_TDML_API_INFOSETS enviroment variable

olabusayoT:
- currently during testing we parse/unparse using both the sax and non-sax API, which leads to issues like trace running outputting twice for the same test which is confusing. We also run the parse for all our infoset outputters. With this enviromental variable, we default to the more efficient single infoset outputter (scalaxml) and single API (non-sax) parse/unparse.
- the DAFFODIL_TDML_API_INFOSETS env has 2 options: 'scala' and 'all'. with scala being the default and 'all' being the current implementation of running both APIs and all infoset outputters
- we convert TDMLInfosetOutputter to a trait so the All and Scala subclasses can extend it as well as TeeInfosetOutputter
- get rid of unused and inaccessible parse function
- set CI mode to all for regression testing
- add cli test showing use of scala and all mode

DAFFODIL-2904

### DIFF
--- a/daffodil-tdml-processor/src/main/scala/org/apache/daffodil/processor/tdml/DaffodilTDMLDFDLProcessor.scala
+++ b/daffodil-tdml-processor/src/main/scala/org/apache/daffodil/processor/tdml/DaffodilTDMLDFDLProcessor.scala
@@ -176,6 +176,8 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor)
   private def blobPrefix = ""
   private def blobSuffix = ".bin"
 
+  private lazy val tdmlApiInfosetsEnv = sys.env.getOrElse("DAFFODIL_TDML_API_INFOSETS", "scala")
+
   override def withDebugging(b: Boolean): DaffodilTDMLDFDLProcessor =
     copy(dp = dp.withDebugging(b))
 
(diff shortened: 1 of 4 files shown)


daffodil-tdml-processor/src/main/scala/org/apache/daffodil/processor/tdml/DaffodilTDMLDFDLProcessor.scala
```diff
@@ -252,21 +250,47 @@ class DaffodilTDMLDFDLProcessor private (private var dp: DataProcessor)
...
+      osw.flush()
+      osw.close()
+      val saxInstream = new ByteArrayInputStream(bos.toByteArray)
+      doUnparseWithBothApis(inputter, saxInstream, outStream)
+    } else {
+      doUnparseWithNonSaxApi(inputter, outStream)
+    }
+  }
+
+  def doParseWithNonSaxAPI(
```
stevedlawrence, line 266: Feels like there's some duplicate code in these doParseWith*API functions since they both implement the non-sax API. Can they be combined into one function, e.g.

```scala
private def doParse(...) {
  // always do stuff with non-sax API

  if (apiMode = full) {
    // do sax API
    // verify SAX results match non-sax results
  }
}
```
Same with the unparse functions?

daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLITdml.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Test def test_CLI_Tdml_Trace_singleTest2(): Unit = {
+    val tdml = path(
+      "daffodil-test/src/test/resources/org/apache/daffodil/section06/entities/Entities.tdml"
+    )
+
+    val envs = Map("DAFFODIL_TDML_API_INFOSETS" -> "scala")
+
+    runCLI(args"test -i -t $tdml byte_entities_6_08", envs = envs) { cli =>
```
stevedlawrence, line 49: Just a note that this doesn't really check that we are  running only the scala API. Unfortunately which APIs are used are way behind the scenes so there isn't really a good way to verify it, and there's no good way for the expect library to expect that a string *doesn't* exist. This test is probably still worthwhile since I imagine it gives us coverage and at least confirms that "scala" is legal, but if we accidentally break things so "scala" runs both APIs, this test probably won't fail.

stevedlawrence (approved):
+1 :+1:

jadams-tresys (approved):
+1

Agree with Steve that I think this came out better as an environment variable instead of tunable.

daffodil-test-integration/src/test/scala/org/apache/daffodil/cliTest/TestCLITdml.scala
olabusayoT: Yea I figured, there might be a way by checking the numbered output of the trace but that would break if we changed the output of trace. Thoughts?

```
----------------------------------------------------------------- 7
parser: <Element name='e3'><DelimiterStackParser>...</DelimiterStackParser></Element>
...
----------------------------------------------------------------- 17
parser: <Element name='e3'><DelimiterStackParser>...</DelimiterStackParser></Element>

```

stevedlawrence: I don't think there's anything that can be done. Just pointing out that we need to be careful if we ever make changes to how to run single/multiple APIs in TDML, since tests tests might not tell us if we broke it.

PR #1367, opened 2024-11-07
Fix Ignored RootNS

olabusayoT:
- currently we find the root element using only the local name of an element, this can result in ambiguous roots and an error if you have an element with no namespace and one with a namespace. This PR fixes that by searching using the qname instead
- add search for qname option if no namespace is specified, because there might still be an unambiguous root element within a namespace, the user just opted not to specify it

DAFFODIL-2947

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaComponent.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaComponent.scala
@@ -334,7 +334,7 @@ final class Schema private (
    *
    * This just scans each schema document in the schema, checking each one.
    */
-  def getGlobalElementDecl(name: String) = {
+  def getGlobalElementDecl(name: String): Option[GlobalElementDecl] = {
     val sds = schemaDocuments
     val res = sds.flatMap { sd =>
       {
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
@@ -315,12 +315,12 @@ final class SchemaSet private (
    * root element name, then this searches for a single element having that name, and if it is
    * unambiguous, it is used as the root.
    */
-  private def findRootElement(name: String) = {
+  private def findRootElement(name: String): GlobalElementDecl = {
     val candidates = schemas.flatMap {
       _.getGlobalElementDecl(name)
     }
     schemaDefinitionUnless(
-      candidates.length != 0,
+      candidates.nonEmpty,
       "No root element found for %s in any available namespace",
       name
     )
@@ -337,7 +337,7 @@ final class SchemaSet private (
       }
     )
     Assert.invariant(candidates.length == 1)
-    val ge = candidates(0)
+    val ge = candidates.head
     ge
   }
 
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -673,7 +673,14 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   lazy val tcID = (testCaseXML \ "@ID").text
   lazy val id = tcName + (if (tcID != "") "(" + tcID + ")" else "")
   lazy val rootAttrib = (testCaseXML \ "@root").text
-  lazy val rootNSAttrib = (testCaseXML \ "@rootNS").text
+  lazy val optRootNSAttrib: Option[String] = {
+    val attr = testCaseXML.asInstanceOf[Elem].attribute("rootNS")
+    if (attr.isDefined) {
+      Some(testCaseXML \@ "rootNS")
+    } else {
+      None
+    }
+  }
 
   lazy val (infosetRootName, infosetRootNamespaceString) =
     if (this.optExpectedOrInputInfoset.isDefined) {
@@ -695,12 +702,12 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   }
 
   def getRootNamespaceString() = {
-    if (optExpectedOrInputInfoset.isDefined)
+    if (this.optRootNSAttrib.isDefined) {
+      optRootNSAttrib.get
+    } else if (optExpectedOrInputInfoset.isDefined)
       infosetRootNamespaceString
     else if (optEmbeddedSchema.isDefined)
       XMLUtils.EXAMPLE_NAMESPACE.toString
-    else if (this.rootNSAttrib != "")
-      rootNSAttrib
     else {
       // For some TDML Processors, we have to provide
       // the root namespace. They don't provide a way to search
--- a/daffodil-tdml-processor/src/test/scala/org/apache/daffodil/processor/tdml/TestTDMLRunner2.scala
+++ b/daffodil-tdml-processor/src/test/scala/org/apache/daffodil/processor/tdml/TestTDMLRunner2.scala
@@ -677,4 +677,81 @@ abc # a comment
     runner.reset
   }
 
+  // DFDL-2947
+  @Test def test_rootNSSpecifiesNoNamespaceRoot(): Unit = {
+    val tdmlTestSuite =
+      <tdml:testSuite suiteName="theSuiteName" xmlns:ex={example} xmlns:tdml={tdml} xmlns:dfdl={
+        dfdl
+      } xmlns:xsd={xsd} xmlns:xs={xsd} xmlns:xsi={xsi}>
+        <tdml:parserTestCase name="test1" root="data" rootNS="" model="/test/tdml/chameleon-schema1.dfdl.xsd">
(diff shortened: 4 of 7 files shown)


daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
```diff
@@ -356,7 +356,15 @@ final class SchemaSet private (
...
       }
       case RootSpec(None, rootElementName) => {
-        findRootElement(rootElementName)
+        val possibleRoots: Seq[GlobalElementDecl] = schemaSet.schemas.flatMap {
+          _.schemaDocuments.flatMap(_.searchGlobalElementDecl(rootElementName))
+        }
+        if (possibleRoots.length == 1) {
+          possibleRoots.head
+        } else {
+          val qn = RefQName(None, rootElementName, NoNamespace)
```
mbeckerle, line 365: Add some comments explaining the rationale here. E.g., 
```
It's ambiguous what root element, and no namespace URI was provided. 
So we assume not specifying a namespace URI may mean the intention 
was to choose the root that it is in No Namespace. 
If there is no such, then the diagnostic will encourage user to specify
the namespace explicitly. 
```

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/processor/tdml/TestTDMLRunner2.scala
```diff
@@ -608,4 +608,80 @@ abc # a comment
   @Test def test_apos_test1(): Unit = { runner.runOneTest("apos_test1") }
   @Test def test_apos_test2(): Unit = { runner.runOneTest("apos_test2") }
 
+  // DFDL-2947
+  @Test def test_multipleRootCandidates1(): Unit = {
+    val tdmlTestSuite =
+      <tdml:testSuite suiteName="theSuiteName" xmlns:ex={example} xmlns:tdml={tdml} xmlns:dfdl={
+        dfdl
+      } xmlns:xsd={xsd} xmlns:xs={xsd} xmlns:xsi={xsi}>
+        <tdml:parserTestCase name="test1" root="data" rootNS="" model="/test/tdml/chameleon-schema1.dfdl.xsd">
```
mbeckerle, line 617: Maybe rename test to reflect that rootNS is being used to specify no namespace. It's not only multiple root candidates but that one of them is no namespace, and this illustrates how to specify that.

mbeckerle (approved):
+1 minor suggestion for doc to add.

daffodil-cli/src/test/scala/org/apache/daffodil/cli/cliTest/TestCLIParsing.scala
```diff
@@ -358,7 +358,7 @@ class TestCLIParsing {
 
     runCLI(args"parse -s $schema -r unknown") { cli =>
       cli.sendLine("12", inputDone = true)
-      cli.expectErr("No root element found for unknown in any available namespace")
+      cli.expectErr("No root element found for {}unknown")
```
stevedlawrence, line 361: This is potentially confusing, and maybe not correct? `unknown` and `{}unknown` are technically different things. The former means "some element called unknown in any namespace". But the later means "some element called unknown explicitly in no-namespace". The error message should probably not include the curly braces here.

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaComponent.scala
```diff
@@ -330,19 +330,31 @@ final class Schema private (
   }
 
   /**
-   * Given a name, retrieve the appropriate object.
+   * Given a qname, retrieve the appropriate object.
    *
    * This just scans each schema document in the schema, checking each one.
    */
-  def getGlobalElementDecl(name: String) = {
+  def getGlobalElementDecl(qname: RefQName): Option[GlobalElementDecl] = {
```
stevedlawrence, line 337: This function now is slightly differently than the other `getFoo` functions below. That always gives me a pause when a bunch of things look like they should be basically the same but have subtle differences. Why does this one need to be different? Why is it okay for those to still lookup based on name, but this one needs a qname?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaDocument.scala
```diff
@@ -261,9 +262,16 @@ final class SchemaDocument private (xmlSDoc: XMLSchemaDocument)
   lazy val defineVariables = annotationObjs.collect { case dv: DFDLDefineVariable => dv }
 
   /**
-   * by name getters for the global things that can be referenced.
+   * by name/qname getters for the global things that can be referenced.
    */
-  def getGlobalElementDecl(name: String) = globalElementDecls.find { _.name == name }
+  def getGlobalElementDecl(qname: RefQName): Option[GlobalElementDecl] =
```
stevedlawrence, line 267: Same question as before, why is this RefQName when the other similar functions are just a String?

daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/SchemaSet.scala
stevedlawrence: > So we assume not specifying a namespace URI may mean the intention

I'm not sure that's a safe assumption. It's possible the user just didn't realize there were two elements with the same name and they actually did want the namespaced one. Feels like it would be safer to not assume and instead error saying there's an ambiguity and just require that the user be explicit about which namespace they want. It believe it's always possible to set no-namespace if that is needed for example one can use `{}foo` in the CLI. Is there a way to do it in the TDML Runner? Maybe `rootNS=""` is explicitly no-namespace?

daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -695,12 +695,12 @@ abstract class TestCase(testCaseXML: NodeSeq, val parent: DFDLTestSuite) {
   }
 
   def getRootNamespaceString() = {
-    if (optExpectedOrInputInfoset.isDefined)
+    if (this.rootNSAttrib != "")
+      rootNSAttrib
```
stevedlawrence, line 699: I wonder if rootNSAttrib wants to be an Option? So that `rootNS=""` is `Some("")` explicitly means no-namespace, and if rootNS is missing then it's None and we just search for any element with the name `root`, and errors if there are ambiguities?

daffodil-test/src/test/resources/org/apache/daffodil/section06/namespaces/namespaces.tdml
```diff
@@ -910,7 +910,7 @@
     <tdml:document><![CDATA[333]]></tdml:document>
     <tdml:errors>
       <tdml:error>Schema Definition Error</tdml:error>
-      <tdml:error>More than one definition for name: vagueEle</tdml:error>
+      <tdml:error>More than one definition for name: {}vagueElem</tdml:error>
```
stevedlawrence, line 913: Similar question here, this feel incorrect since `{}` explicity means no-namespace, but no-namespace was not suggested as the namespace to use for this test. It's even possible that `vagueElement` doesn't even exist with no-namespace, it chould have two different namespaces, in which case {} is confusing since it implies no-namespace.

Maybe related, I noticed that `RootSpec.toString` always outputs curly braces. Maybe that's where the curly braces are coming from? Maybe it should only output curly braces if ns is defined? That would better match our "extended syntax".

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/processor/tdml/TestTDMLRunner2.scala
```diff
@@ -677,4 +676,80 @@ abc # a comment
...
+    val tdmlTestSuite =
+      <tdml:testSuite suiteName="theSuiteName" xmlns:ex={example} xmlns:tdml={tdml} xmlns:dfdl={
+        dfdl
+      } xmlns:xsd={xsd} xmlns:xs={xsd} xmlns:xsi={xsi}>
+        <tdml:parserTestCase name="test1" root="data" rootNS="doesNotExist" model="/test/tdml/chameleon-schema1.dfdl.xsd">
+          <tdml:document>37</tdml:document>
+          <tdml:errors>
+          <tdml:error>Schema Definition Error</tdml:error>
+          <tdml:error>no global element found</tdml:error>
+          <tdml:error>{"{doesNotExist}"}data</tdml:error>
```
stevedlawrence, line 747: This error message looks kind of werid, seems there is an extra set of curly braces and quotes? I would think the error should just be something like `<tdml:error>{doesNotExist}data</tdml:error>`

stevedlawrence (approved):
+1, looks good but one of the expected error messages looks not quite right?

daffodil-tdml-processor/src/test/scala/org/apache/daffodil/processor/tdml/TestTDMLRunner2.scala
olabusayoT: Oh that was done that way because we are in a scala file, and the braces in scalaxml mean interpolation, so it was done the way it was to sort of escape the braces, but I'll update to the official way to escape which is {{ and }}

PR #1368, opened 2024-11-08
Trim ErrorWarningBase Messages

olabusayoT:
- so all errors, validation errors and warning messages are stripped of leading and trailing whitespace
- currently the above are sensitive to pretty printing of the TDML, this commit aims to fix that

DAFFODIL-2955

### DIFF
--- a/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
+++ b/daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
@@ -2852,8 +2852,8 @@ abstract class ErrorWarningBase(n: NodeSeq, parent: TestCase) {
 
   protected def diagnosticNodes: Seq[Node]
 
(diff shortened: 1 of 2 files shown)


daffodil-tdml-lib/src/main/scala/org/apache/daffodil/tdml/TDMLRunner.scala
```diff
@@ -2852,8 +2852,8 @@ abstract class ErrorWarningBase(n: NodeSeq, parent: TestCase) {
 
   protected def diagnosticNodes: Seq[Node]
 
-  lazy val messages = diagnosticNodes.map {
-    _.text
+  lazy val messages: Seq[String] = diagnosticNodes.map {
+    _.text.trim
```
mbeckerle, line 2856: Just verifying, trim removes leading and trailing whitespace, but doesn't collapse interior whitespace. Am I correct?

mbeckerle (approved):
+1

PR #1369, opened 2024-11-11
Reorder DFDL statement processing to match spec

jadams-tresys:
We had been processing all asserts and discriminators before the group
content, but according to section 9.5.2 of the spec non-pattern asserts
and discriminators should be evaluated after the group content, even if
the group content fails to parse.
    
DFDL-1971

### DIFF
--- a/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
+++ b/daffodil-core/src/main/scala/org/apache/daffodil/core/dsom/DFDLStatementMixin.scala
@@ -216,7 +216,8 @@ trait ProvidesDFDLStatementMixin extends ThrowsSDE with HasTermCheck {
 
   final lazy val patternStatements: Seq[DFDLStatement] = patternAsserts ++ patternDiscrims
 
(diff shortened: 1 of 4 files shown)


jadams-tresys:
When testing I noticed that dfdl-gif and dfdl-jpeg2000 had some failures with this change.  I suspect that these schemas might need a small update as the order of when discriminators and assert statements are evaluated are changed and these schemas seem to be relying on the old/incorrect behavior

mbeckerle (changes requested):
Pretty sure this won't be correct as is. 

More cases need tests. Where the body parses successfully but the discriminator fails, where the body fails and the discriminator fails, where the body fails, but the discriminator succeeds - and references elements in the partial body infoset (infoset as it existed at the time of the body failure)

jadams-tresys:
Apparently it had been so long since I worked on DAFFODIL-1971 that I didn't realize I had a nearly complete pull request for it in #987 that seems more cleanly implemented and doesn't cause any issues with the C generator.

I've added more tests to #987 and rebased onto main and everything is looking good.  Assuming that is the way we want to implement this, I can close this PR in favor of #987,

PR #1370, opened 2024-11-12
Verify Blank OutputNewLine causes SDE

olabusayoT:
- add test to verify we SDE instead of unclear Unparser error as stated in the ticket

DAFFODIL-1589


jadams-tresys (approved):
+1

mbeckerle (approved):
+1

PR #1371, opened 2024-11-12
Add TDML Exception when there's no Expected Data

olabusayoT:
- note this only gets thrown where there is a warning as well, as UnparserTestCase requires 2 children. So if you just have an infoset and nothing else, you'll get an XSD error.

DAFFODIL-2289


mbeckerle (approved):
+1

jadams-tresys (approved):
+1

PR #1372, opened 2024-11-13
Verify overlap check doesn't occur with default format

olabusayoT:
- add test for verification

DAFFODIL-1842


mbeckerle (approved):
+1

jadams-tresys (approved):
+1

